PR #168, opened 2010-12-21
mcd function added to directories.zsh

derekprior:
mcd will make and immediately change to a directory.


1atAlcone:
Space POD container replaced positive P95 instead of P11? THE POD HAVE CALENDAR DATES BUT IT IS EMPTY.

PR #351, opened 2011-05-22
add custom completion support

gutenye:
you can define your own completion file in $ZSH/completions directory.

For example:

define a custom cd function

```
function mycd(){ cd ~/dev/$1}
```

and add mycd completion function

the $ZSH/completions/_mycd file

```
#compdef mycd

_files -W $HOME/dev
```

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -1,7 +1,7 @@
 # Initializes Oh My Zsh
 
 # add a function path
-fpath=($ZSH/functions $fpath)
+fpath=($ZSH/functions $ZSH/completions $fpath)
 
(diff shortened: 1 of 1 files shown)


apjanke:
So, the `$ZSH/completions` directory doesn't actually exist in the OMZ distribution. (I don't know if it got added at some point.) What's the intended way for this to be used? Are users supposed to put per-user custom completions in there and add them to `.gitignore`? Or is it for completions shipped with OMZ?

There's a new PR #4102 to add the corresponding `$ZSH_CUSTOM/completions` and I'm trying to figure out exactly how this is supposed to work.

PR #373, opened 2011-05-30
autosudo plugin

gutenye:
this is a convenient way to use sudo command. automatic apply the 'sudo' command at some commands if you are not root.

see the code for more information.

### DIFF
--- a/plugins/autosudo/autosudo.plugin.zsh
+++ b/plugins/autosudo/autosudo.plugin.zsh
@@ -1,7 +1,5 @@
 [ $(id -u) -eq 0 ] && return
 
-alias s="sudo"
-
 function pacman() {
 	case $1 in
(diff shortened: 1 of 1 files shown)


nextgenthemes:
@robbyrussell well I find it funny that you care about users security this way where the user would have to explicitly enable this plug-in and could be warned beforehand when he finds the plugins description, or even better inside the code while doing this commands, but then you let everyone download the install script for oh-my-zsh over and *totally insecure* HTTP connection!

https://github.com/robbyrussell/oh-my-zsh/issues/3459

mcornella:
In case someone comes across this PR, there's actually a safer alternative to this functionality in the `sudo` plugin, which sets a key binding (`ESC`+`ESC`) to prefix the last command (or current one) with `sudo `. 

Permission error? Quickly press `ESC` twice, `ENTER` and you're good to go!

PR #736, opened 2011-12-09
gen_pass: Generate a password from random and uses tr

nogoth:


ronnyhartenstein:
Thanks for this on-demand local custom plugin :)

PR #1077, opened 2012-04-18
Add option to disable status notification

Mezzle:
For certain git repositories, this slows down usage
of the shell horifically.

This option can be set with

```
git config --add oh-my-zsh.hide-status 1
```

which will disable checking/showing the status notification


mfournierca:
This is fantastic, thanks a lot.

PR #1359, opened 2012-10-15
Do not force PATH in the installer 

mpapis:
The installer forced PATH - this prevents PATH dynamically set by system, also forces any temporary user PATH that was set during installation to be recorded and used on every shell start.

It should be users conscious choice to manipulate PATH, a lot of users is not aware of this and do not even know if this good or not.

If users need to add something to PATH they should do it by adding it not overwriting system detected PATH:

```
PATH="$PATH:/user/custom/path"
```

or:

```
path+=( "/user/custom/path" )
```

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -21,9 +21,6 @@ fi
 echo "\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m"
 cp ~/.oh-my-zsh/templates/zshrc.zsh-template ~/.zshrc
 
-echo "\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m"
-echo "export PATH=$PATH" >> ~/.zshrc
-
 echo "\033[0;34mTime to change your default shell to zsh!\033[0m"
 chsh -s `which zsh`
 


a6ftcruton:
+1 to @grevolution

dantesolis:
+1 to @grevolution

GrayCoder:
Soooo, I tried commenting the PATH line and commands like "bundle" and such wouldn't work. 

However, adding "source /home/user/.rvm/scripts/rvm" to the end of my .zshrc seems to have fixed the problem.

apjanke:
This caused #3960.

On OS X, the `/etc/paths.d` directories are picked up by the systemwide `/etc/*` shell rc files and added to `$PATH` before the user's shell rc file gets called. Grabbing a static copy of the `$PATH` at OMZ installation time breaks `/etc/paths.d`; programs using `/etc/paths.d` that are installed later, or upgraded so that their path changes, are not visible on the path.

This will affect Ruby, golang, XQuartz, and TeX.

lingceng:
This problem bite me too! Why not merge the request?

jediorange:
Yeah, this is not cool. This causes Macs to ignore /etc/paths.d, which is where new apps are supposed to add paths.

jessiahr:
Looks like I am running into this issue while using RVM. Any update on a fix?

NekoNova:
Seeing as this is not getting fixed, uninstall oh-my-zsh till it actually works....

apjanke:
You don't need to uninstall Oh My Zsh. Just correct the `$PATH` in your `~/.zshrc`, possibly by just moving your old `~/.zshrc` back and copying the OMZ initialization code to it. This change only happens at installation time, and OMZ is fine running with a dynamically-set `$PATH`.

You can also avoid the problem by doing a manual installation by cloning the OMZ repo to `~/.oh-my-zsh` and adding `export ZSH=~/.oh-my-zsh; source `$ZSH/.oh-my-zsh/oh-my-zsh.sh` to your existing `~/.zshrc`.

NekoNova:
yea, or the project could you know....be fixed and support everything as it should be.

jairojunior:
+1

icechen1:
This just fixed my rails after hours of digging. +1

gtmax:
+1.

mpapis:
should I rebase?

mcornella:
No need, sorry, I just reopened this again to signal that it's not solved yet.

diegomelo182:
+1

lasernite:
+1

milos-v:
This one did it for me
[stackoverflow](http://stackoverflow.com/a/10343891/6291389)

apjanke:
> This one did it for me

`/etc/paths` is a system-level configuration file. OMZ is more of a user-level shell customization framework; it probably shouldn't be modifying `/etc/paths`, or doing anything that requires modification of it.

milos-v:
Well OMZ is not changing it. I changed the order of paths in the file like it says here: 

> open your /etc/paths file, put /usr/local/bin on top of /usr/bin

Not sure how it relates to the issue honestly but the issue is gone.

apjanke:
@milos-v Right. I'm saying you shouldn't have had to do that in the first place. And also that, even if that act worked for you, OMZ shouldn't do the same thing.

But moreover, the problem in this issue report occurs when you:
  a) run the OMZ installer, and then later:
  b) do something that changes `/etc/paths.d`
    or
  c) share your `~/.zshrc` between machines with differing `/etc/paths.d` states
If you didn't do these two things, you're probably not looking at quite the same issue as this bug report.

galarchenko:
Works for me. Thanks

hayesmaker:
+1

jeremy-bruns:
+1

lborg019:
on my OS X I used
```bash
$ vim /etc/paths.d/go
```
and the file contained:
```bash
/usr/local/go/bin
```
I added this last path to my .zshrc and it worked.
here's what my .zshrc export looked like:

```bash
export PATH="/Users/myusername/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin"
```
All I had to do was append `/usr/local/go/bin` to the end of this path, and now Go works on oh-my-zshell. Here's what my .zshrc export looks like now:

```bash
export PATH="/Users/myusername/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/go/bin"
```

CodyLeeWhite:
This problem is to much trouble. uninstalling as it appears a fix will not be made. I should not have to manually edit ohmyzsh to get my path to work with it.

xudeheng:
You should comment the rvm related lines in ~/.zshrc, it works for me.

mcornella:
This was fixed in #4925. I don't know why I kept it open so long.

PR #1478, opened 2012-12-08
Use xdg-open in Linux for the o alias, if it's available.

gokceneraslan:


mcornella:
Hi @gokceneraslan, it's been a long time but I managed to find a clean and simple solution, look at the [web-search plugin](https://github.com/robbyrussell/oh-my-zsh/blob/ef7e53a78d0e4196c2d1e6e5b268209759d51753/plugins/web-search/web-search.plugin.zsh#L17-L24) to safely open across different platforms. 
What do you say if I put that into a lib file so we can use it across the project?

gokceneraslan:
Looks very similar to my patch :) Go go go...

mcornella:
Hi @gokceneraslan, I finally implemented my standard open command. Now the `o` alias is `a -e open_command`, where `open_command` is the implemented function. 

Please test if it works correctly since I don't have `fasd` to check.
Cheers!

PR #1491, opened 2012-12-13
screen: Bugfix extracted cmd, add customizable TITLE var

scosu:
Bugfix:
zsh doc:
"If the history mechanism is active, the string that the user typed is
passed as the first argument, otherwise it is an empty string."
The title of a screen window should not depend on the history. So this
patch uses the third passed argument ("the third argument contains the
full text that is being executed.").

TITLE variable:
This combines automatic title set with customizable titles. It's often
useful in a screen session to set a manual title. By using
TITLE=YOUR_TITLE, it will replace the generated prefix but still display
the executed command. "TITLE=" will cause the screen plugin to generate
the title prefix again.

Signed-off-by: Markus Pargmann mpargmann@allfex.org

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -2,12 +2,23 @@
 #http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#ss3.1
 #Fully support screen, iterm, and probably most modern xterm and rxvt
 #Limited support for Apple Terminal (Terminal can't set window or tab separately)
+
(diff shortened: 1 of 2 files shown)


larson-carter:
@scosu Great PR, can you please rebase to our master branch?

scosu:
@larson-carter I am not using zsh anymore, please feel free to take this patch and rebase it.

larson-carter:
Ok, I am going to close this PR and wait until someone rebases it to the latest master branch or if they rebuild it.

PR #1504, opened 2012-12-19
Exit early from git plugin if not in git repo.

jediant:
Added check for .git directory in current, before wasting time querying git. This saves seconds on my pc.


ncanceill:
Yeah, as @blueyed had pointed out prior to the time of merging this, this fix does not work at all.

`git` commands exit pretty quickly when not in a repository anyway, so this was removed from #2790.

Stibbons:
Please revert this patch. This is invalid since git command work in subdirectory. You need to look for every parent of the current directory to see if these isn't a .git directory

ncanceill:
Yes, #2790 reverts this.

PR #1598, opened 2013-02-11
new plugin

zasdfgbnm:
start fbterm automatically in /dev/tty*

### DIFF
--- a/plugins/fbterm/fbterm.plugin.zsh
+++ b/plugins/fbterm/fbterm.plugin.zsh
@@ -1,0 +1,6 @@
+# start fbterm automatically in /dev/tty*
+
+if [[ $(tty|grep -o '/dev/tty') = /dev/tty ]] ; then
+	fbterm
+	exit
+fi


wzhd:
This plugin kills the the terminal if fbterm is not properly installed and configured.

Currently, this plugin doesn't check whether fbterm is installed. When fbterm is not installed and this plugin is enabled, zsh will print an error like `zsh: command not found: fbterm`, but then `exit` is executed so this error message cannot be seen. Probably this can be fixed by checking when fbterm is installed.

Even if fbterm is installed, it is still not common to encounter errors which causes fbterm to exit unexpectedly. For example, on some distros, the user need to be added to the `video` group to run fbterm as a non-root user. Otherwise, fbterm will print an error: `can't open frame buffer device!`. Then again, `exit` is executed and the user can't see what's happening. I didn't come up with a way to fix this easily. If fbterm returns an error value when it can't run properly, we can just check its return value. If it fails, just use the normal tty. (So I opened an issue on fbterm's project page https://code.google.com/p/fbterm/issues/detail?id=73)

It would be nice if the user can know what's happening easily. Once upon a time, I accidentally locked myself out of tty and had a hard time trying to figure out where the problem is. Though it was my fault to forget to configure fbterm.

mcornella:
I assume the structure `call; exit` was in place to mimic the `exec` command behavior, which runs the command in place of the current shell (_i.e._ using the same process). @zasdfgbnm probably didn't know about that command, so he used this other one instead: change `fbterm; exit` with `exec fbterm`. 

I ignore if the command not found error will still be there, so you can make some tests and submit a PR if you know how.

/cc @zasdfgbnm

wzhd:
@mcornella I just tried `exec` some non-existent commands, and the terminal exits immediately. I guess the command not found error is still there. Though I agree it may be a good idea to use `exec`.

mcornella:
Ok, so we will need checks in place as well. Maybe we can run fbterm in a subshell with the `()` syntax and see if there's a non-zero exit code, but we still need a way to run fbterm and exit immediately (otherwise we run it twice). This is what I've been able to come up:

```zsh
(fbterm --version &>/dev/null) && exec fbterm
```
[`fbterm --version`](http://manpages.ubuntu.com/manpages/lucid/man1/fbterm.1.html) just displays the fbterm version and exits, so it will fail if fbterm isn't installed _but I don't know about the rest of possible errors that might happen. What do you think?

mcornella:
I've submitted PR #4234 which fixes the command not found error and uses `exec` instead of exiting. Should I go back to just calling `fbterm` and get rid of the `exit` statement so that errors are visible?

mcornella:
Nevermind, I settled for the `fbterm && exit` construct. This will exit if and only if the exit code of fbterm is 0, _i.e._ it finished without errors. It remains to be checked that if `fbterm` encounters an error it will quit with a non-zero exit code.

PR #1624, opened 2013-02-23
Add aliases 'grb*' for git-rebase

yous:


josephj:
I use grb as an alias for git_remote_branch and I don't want use grb for git rebase. How should I stop it correctly?

mcornella:
You can `unalias grb` (put that in zshrc) and it will remove the grb alias, so the next time you run `grb` it will execute git_remote_branch

josephj:
It works! Thanks @mcornella !

PR #1629, opened 2013-02-26
Setup special keybindings using terminfo

jimhester:
The support for special keys right now is simply with hardcoded values,
this commit changes that to using terminfo information for the keys to
set the binds.  Should be much more general, and just work.  This also 
fixes a bug in the vi-mode plugin where the special keys were not bound 
because setting vi mode removes binds set in emacs mode.

Fixes the problem with 53854754e0d052 not setting the special key 
keybindings and those keys not being recognized when using vi-mode.

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -20,6 +20,9 @@ zle -N zle-keymap-select
 
 bindkey -v
 
+#rebind special key keybinds
+source $ZSH/lib/special-keys.zsh
+
 # if mode indicator wasn't setup by theme, define default
 if [[ "$MODE_INDICATOR" == "" ]]; then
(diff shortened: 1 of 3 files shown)


apjanke:
Could you rebase this PR if you'd like to pursue it, or close it? The aspect of using terminfo for the escape sequence lookups is now being used (since #1355 and the #2608 stuff). Now this PR is providing the notion of a higher-precedence layer of key mappings, which I think can be considered separately.

plugins/vi-mode/vi-mode.plugin.zsh
```diff
@@ -20,6 +20,9 @@ zle -N zle-keymap-select
 
 bindkey -v
 
+#rebind special key keybinds
+source $ZSH/lib/special-keys.zsh
```
apjanke: This doesn't respect `$ZSH_CUSTOM`, which might be trying to override `special-keys.zsh`.

mcornella:
I'm closing this PR, we're moving on a different direction and it's 3 years old now.

PR #1631, opened 2013-02-26
Disable globbing when executing `zeus rake`

blakewatters:
Enables easy use of the Zeus rake aliases with tasks that accept arguments via square brackets.


mcornella:
Confirmed to be equally necessary with or without #2058

Issue #592, opened 2011-09-09
knife plugin errors for cookbook upload


cabbiepete:
I am following the chef getting started guide specifically http://help.opscode.com/kb/start/2-setting-up-your-user-environment

I decided to install the knife plugin for my zsh.

I have created a .chef folder withing the repo as the how to suggests. This works fine on the knife side of things but the completion for local cookbooks fails.

I've tracked it down to line 166 of the _knife plugin file. It assumes that the .chef folder will be in my home folder.

I think the gist of bash completion found https://gist.github.com/1050685 (provided by opscode) allows for either case, i.e. it checks to see if we are in a chef repo if so uses the folders in there otherwise uses the knife.rb file.

Something like this would be more useful

if  ((test -d ".chef")); then
    (for i in $( grep cookbook_path .chef/knife.rb | awk 'BEGIN {FS = "[" }; {print $2}' | sed 's/\,//g' | sed "s/'//g" | sed 's/(._)]/\1/' ); do ls $i; done)
  else
    (for i in $( grep cookbook_path $HOME/.chef/knife.rb | awk 'BEGIN {FS = "[" }; {print $2}' | sed 's/\,//g' | sed "s/'//g" | sed 's/(._)]/\1/' ); do ls $i; done)
  fi

It checks to see if there is a .chef dir in the current folder similar to what the knife command does. For my particular example I need to go a bit further and also parse the #{current_dir} variable which knife allows 

I guess an alternative could be to just test for .chef in current dir and assume cookbooks folder is in the current dir also and use the folders within as the list. This seems to be what the bash script in the gist https://gist.github.com/1050685 does.

PR #1682, opened 2013-04-01
Allow you to customize the path to knife config files

maplebed:
The path to knife.rb is currently harcoded to $HOME/.chef/knife.rb.  Mine lives elsewhere.
Set KNIFE_CONF_PATH in your .zshrc to override $HOME/.chef/knife.rb.  Not setting that variable leaves it as it behaved before.
The plugin attempts to get the path to the cookbooks from the knife.rb conf file.  However, the knife.rb conf allows ruby in the cookbook_path definition (in my case cookbook_path is "#{current_dir}/../cookbooks"). Since zsh is not using ruby to parse the conf file, it fails to expand variables like #{current_dir}.  Rather than try and use ruby to interpret knife.rb (which is scary), I added an override for the cookbook directory.  Set KNIFE_COOKBOOK_PATH in your .zshrc to skip examining knife.rb for the cookbook_path.

### DIFF
--- a/plugins/knife/_knife
+++ b/plugins/knife/_knife
@@ -1,5 +1,10 @@
 #compdef knife
 
+# You can override the path to knife.rb and your cookbooks by setting
+# KNIFE_CONF_PATH=/path/to/my/.chef/knife.rb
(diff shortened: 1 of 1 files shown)


mcornella:
This fixed #592

PR #1686, opened 2013-04-03
added hub autocomplete instructions

timfjord:
See [here](https://github.com/defunkt/hub/blob/master/etc/hub.zsh_completion)


apjanke:
```
+    # Autoload _git completion functions
+    if declare -f _git > /dev/null; then
+      _git
+    fi
+    
```

is not autoloading the `_git` function; it is actually running it. Outside the context of ZLE completion, that won't work, and will produce the error @davidcelis is seeing.

The normal `_git` files supplied by the `zsh` or `git` installations are autoloadable files. That's the normal zsh convention for files named `_functionname`. Some programs supply `_foo` files that are intended to be run directly, though; I've seen this with `awscli`. Sounds like this change may have been written against a version of `_git` like that. Is this the case? Is this plugin intended to be compatible with the current `_git` files from `zsh` and `git`?

PR #1698, opened 2013-04-08
safe-paste plugin

ConradIrwin:
Pasting into a terminal can be dangerous (see http://thejh.net/misc/website-terminal-copy-paste). I certainly do it accidentally all the time.

Luckily many terminal emulators can indicate to their containing programs when pasting is happening. This plugin (shamelessly copied from http://www.zsh.org/mla/users/2011/msg00367.html) allows you to paste into zsh runnings inside xterm/urxvt/iTerm2 safely.

It does this by preventing any code from actually running while pasting, so you have a chance to review what was actually pasted before running it.

### DIFF
--- a/plugins/safe-paste/safe-paste.plugin.zsh
+++ b/plugins/safe-paste/safe-paste.plugin.zsh
@@ -1,0 +1,54 @@
+# Code from Mikael Magnusson: http://www.zsh.org/mla/users/2011/msg00367.html
+#
+# Requires xterm, urxvt, iTerm2 or any other terminal that supports bracketed
+# paste mode as documented: http://www.xfree86.org/current/ctlseqs.html
+
+# create a new keymap to use while pasting
+bindkey -N paste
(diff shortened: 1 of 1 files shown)


stephane-chazelas:
Note that this plugin is safe in the default configuration of recent versions of `xterm`, that don't allow pasting control characters other than newline (transformed to carriage-return), backspace, carriage-return, del and tab.

There's a problem only on terminals that don't do that like gnome-terminal or older versions of xterm, or xterm configured with `allowPasteControls` enabled.

See http://security.stackexchange.com/a/52655 for details.

Smasherr:
Doesn't seem to work out of the box on WSL. Has anybody got it running..?

mcornella:
Can you test https://github.com/ohmyzsh/ohmyzsh/pull/7887?

Smasherr:
@mcornella The behavior is the same, the lines get executed right after they are pasted

image

PR #1752, opened 2013-04-24
Added perms plugin

gneatgeek:
I added a new plugin for recursively setting all files' permissions to 644 recursively in a directory and for recursively setting all directories' permissions to 755.

This has been useful primarily for websites, but I've found other uses for it. I also wrote a function to do both at the same time with verification.

### DIFF
--- a/plugins/perms/perms.plugin.zsh
+++ b/plugins/perms/perms.plugin.zsh
@@ -1,47 +1,78 @@
-# http://gneatgeek.github.io
-# Rory Hardy [GneatGeek]
 # Some useful commands for setting permissions.
+#
+# Rory Hardy [GneatGeek]
+# Andrew Janke [apjanke]
 
 ### Aliases
 
 # Set all files' permissions to 644 recursively in a directory
-alias set644='find -type f -exec chmod 644 {} \;'
- 
+alias set644='find . -type f -print0 | xargs -0 chmod 644'
+
 # Set all directories' permissions to 755 recursively in a directory
-alias set755='find -type d -exec chmod 755 {} \;'
+alias set755='find . -type d -print0 | xargs -0 chmod 755'
 
 ### Functions
 
-# fixperms - useful combination of the above two aliases. Will apply to supplied dir or current dir.
+# fixperms - fix permissions on files and directories, with confirmation
+# Returns 0 on success, nonzero if any errors occurred
 fixperms () {
-  if [ "${#}" -gt "1" -o  "${1}" = "--help" ]; then
-    echo "Usage: fixperms [source]"
-  else
-    confirm=""
-    while [ "${confirm}" = "" -o "${confirm}" = "y" ]; do
-      if [ -d "${1}" ]; then
-        if [ "${confirm}" = "" ]; then
-          echo "Fixing perms on ${1}?"
-        else
-          find "${1}" -type f -exec chmod 644 {} \;
-          find "${1}" -type d -exec chmod 755 {} \;
-        fi
-      else
-        if [ "${confirm}" = "" ]; then
-          echo "Fixing perms on ${PWD##*/}?"
-        else
-          find -type f -exec chmod 644 {} \;
-          find -type d -exec chmod 755 {} \;
-        fi
-      fi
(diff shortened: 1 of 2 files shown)


apjanke:
A couple items:

* The `-exec chmod {}` form is going to be slow on large directories. How about `find | xargs chmod`, at least as a default?
* The `while` loop logic is a little hard to understand, since it looks like it's only going to make one pass regardless. This looks like a holdover from an alternate form that could process multiple target directories.
* The `if` branch for handling whether the user supplied a target or it defaulted to `$PWD` duplicates a lot of code. Might be cleaner to set the target directory based on input up front, and then apply a single path of `find`/`chmod` logic to it.
* The `-d $1` test checks whether something is a directory, so if the user supplies a target argument, but it happens to be a symlink or normal file, or a nonexistent file (like if they made a typo), it falls back to targeting `$PWD` instead. That's probably not what the user intended.
  * Since only the basename of the target dir is displayed in the confirmation message, there are even cases where the user couldn't tell if this happened by reading the message, if their intended target had a basename similar to the `$PWD` they were in. 
* Exit status doesn't reflect the exit status of the `find` and `chmod` operations.
* Since prompting is unconditional, might want to restrict this to interactive or tty-attached shells.
* OMZ typically uses zsh style `[[ ... ]]` instead of sh-compatible `[ ... ]` tests.

How about [this](https://gist.github.com/apjanke/4a4efb3367a30bd9830c)?

gneatgeek:
Thanks @apjanke, sorry for the year late response! Somehow this one slipped through the cracks. I haven't had a chance to test the code yet, but your solution looks good to me.  I'll try it out this evening and update this pull request accordingly.

gneatgeek:
@apjanke, I made some minor tweaks, but it looks like it's working out great.  @robbyrussell do find this acceptable at this time?

gneatgeek:
@robbyrussell, @apjanke, and @mcornella, is there anything else that this needs from me?

plugins/perms/perms.plugin.zsh
```diff
@@ -0,0 +1,78 @@
+# Some useful commands for setting permissions.
+#
+# Rory Hardy [GneatGeek]
+# Andrew Janke [apjanke]
+
+### Aliases
+
+# Set all files' permissions to 644 recursively in a directory
+alias set644='find . -type f -print0 | xargs -0 chmod 644'
```
mcornella: You can use zsh globbing to select all files in the current directory and subdirectories: `chmod 644 **/*(.)`

plugins/perms/perms.plugin.zsh
```diff
@@ -0,0 +1,78 @@
...
+
+# Set all directories' permissions to 755 recursively in a directory
+alias set755='find . -type d -print0 | xargs -0 chmod 755'
```
mcornella: Same as above, you can select all subdirectories with: `chmod 755 **/*(/)`.

plugins/perms/perms.plugin.zsh
```diff
@@ -0,0 +1,78 @@
...
+### Functions
+
+# fixperms - fix permissions on files and directories, with confirmation
+# Returns 0 on success, nonzero if any errors occurred
+fixperms () {
+  local opts confirm target exit_status chmod_opts use_slow_mode
+  zparseopts -E -D -a opts -help -slow v+=chmod_opts
+  if [[ $# > 1 || -n "${opts[(r)--help]}" ]]; then
+    cat <<EOF
+Usage: fixperms [-v] [--help] [--slow] [target]
```
mcornella: If you use tabs for indentation you can then use `cat <<-EOF` (notice the extra hyphen) to strip the tabs. You'd  then have to indent using spaces inside the heredoc, so it's not a clear benefit, just throwing it out there for your information. 😁

mcornella:
I've left a couple of suggestions that could improve the code, but as is the plugin is more than ok to merge. Besides, this must be the oldest one yet to be merged haha.

Thanks for your work!

gneatgeek:
Thanks for the tips and for merging the pull requests @mcornella!  :)

PR #1813, opened 2013-05-13
Play Framework 2.x support. 

GitsMcGee:
I created a quick plugin with support for the PlayFramework v2.0+. I implemented autocomplete, as well as threw in a couple of aliases for commands that are commonly used.


mcornella:
Not popular enough.

PR #1847, opened 2013-05-25
Add cdargs plugin

elcepi:
Add a plug in to use cdargs.  Basically I copy it from the bash one but work good.


robbyrussell:
Can you include a README file?

PR #1855, opened 2013-05-29
Use `vcs_info` in gentoo theme for other VCSes

freeyoung:
I love gentoo theme of oh-my-zsh, however it is somewhat too simple.

Now it can show svn/cvs info with `vcs_info`, instead of git only.

Most code of this commit is from `apple` theme, thanks to the two contributors!


ym:
:clap:

jgrar:
:+1:

mcornella:
Reposted with fixes in https://github.com/ohmyzsh/ohmyzsh/pull/9440. Go +1 that if it works for you.

PR #1856, opened 2013-05-31
check_for_upgrade.sh not executable with /bin/sh

MatthiasLohr:
check_for_upgrade.sh is not executable with /bin/sh while using bash syntax, so the execution shell should be set to /bin/bash.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env zsh
 
 function _current_epoch() {
   echo $(($(date +%s) / 60 / 60 / 24))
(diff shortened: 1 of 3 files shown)


apjanke:
Couple questions:
Is `#!/usr/bin/env zsh` the right shebang line for `install.sh`? That is run under `sh` during the install process, not `zsh`. So it needs to be kept to just `sh`-compatible syntax. Unless you're planning on changing the install process to run it through `zsh` instead?

And how about renaming `check_for_upgrade.sh` to `check_for_upgrade.zsh`? Same for `theme_chooser.sh` to `theme_chooser.zsh`, since it has a `zsh` shebang. That way all the file extensions and shebang lines are all consistent with the shell flavor the files are targeting.

mcornella:
Sorry for the late late response, this fell off the tracks. [I'm going to build upon this PR to make the transition to a full-zsh codebase](https://github.com/mcornella/oh-my-zsh/tree/use-zsh-everywhere), but for the moment i'm closing this as there are many things to take into account.

Again, sorry for losing track of this PR and thank you for your all your work.

PR #1872, opened 2013-06-06
Update the character used from powerline

gdetrez:
In the current version of the font, the fancy glyphs have been moved to the Private Use Area: https://github.com/Lokaltog/powerline/issues/4

> The code points have been updated in 53fbfe1

This updates the agnoster theme accordingly.


Mardii-D-King:
For any one having troubles with a broken prompt, follow these instructions  https://powerline.readthedocs.org/en/latest/installation/linux.html#fontconfig and it should do the trick. I have to say thanks to @emyller for the solution, thanks man. Ive been on this for hours :) :+1:

PR #1899, opened 2013-06-20
Added "jad" theme

jad:
Added a theme that includes user, hostname, directory, git status, time, and command number. The git status includes emoji to indicate dirty/clean repo status. If that's an issue with compatibility, let me know and I can update it to just use ASCII characters.

screen shot 2013-06-20 at 12 35 56 am

### DIFF
--- a/plugins/git-flow-completion
+++ b/plugins/git-flow-completion
@@ -1,1 +1,0 @@
-/Users/jad/src/home/git-flow-completion


blueyed:
A revised system should probably use Zsh's "PROMPT THEMES" mechanism (see `man zshall`), via `promptinit`.
Then themes could also be provided through a separate repo / collection, and would work with vanilla Zsh / other distributions - making the oh-my-zsh repo itself leaner/smaller.

bucaran:
Or you could do something similar to what I do in [Wahoo](https://github.com/bucaran/wahoo#use-theme).  

```
wa use <name of the theme>
# gets applied whether it exists locally or in a registered remote repo; else it fails
```

bucaran:
@robbyrussell It's been a few years since your last comment in this thread, sorry to bring up the subject again, but perhaps you could try something similar to what I did in [Wahoo](https://github.com/bucaran/wahoo#use-theme)?

```
wa use <name of the theme>
# gets applied whether it exists locally or in a registered remote repo; else it fails
```

kylebrowning:
Oh my shell is desperate for some new themes.

bucaran:
@kylebrowning You want to try fish?

christoga:
Yo try to make someone replace oh-my-zsh with fish?

mcornella:
@kylebrowning if you like this theme you can copy it from the `Files changed` tab, or you can have a look at [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) that are not shipped within OMZ.

@bucaran regarding you [wahoo comment](https://github.com/bucaran/wahoo#use-theme) which is now deleted, in zsh there's [antigen](https://github.com/zsh-users/antigen/) and [zgen](https://github.com/tarjoilija/zgen) which serve the same purpose, and we'll encourage people to use that method before reinventing the wheel.

bucaran:
@mcornella Yep, now it's gone with the wind.

PR #1951, opened 2013-07-05
Add .ipsw to the extract plugin.

Neal:
Prior to this commit, the extract plugin did not know what .ipsw file was.
This commit adds .ipsw files to the extract auto-completion list and extracts it with `unzip`.


mcornella:
:+1: Please rebase against the current upstream though

/cc @robbyrussell

PR #1972, opened 2013-07-13
Added Droplr Plugin

fabiofl:
Thought CloudApp is pretty cool, I still prefer Droplr. Unfortunately, access to it's API is pretty much closed, so the workaround was to "remote control" the desktop client itself to do the job.


robbyrussell:
Can you include a README for this?

fabiofl:
Will do.

mcornella:
Merged with some changes. Sorry for the massive delay on this one!

PR #2011, opened 2013-07-28
LS_COLORS support for completion

L0stSoul:
If LS_COLORS defined use this for zsh completion, think it may be useful :)

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -18,7 +18,11 @@ else
   zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
 fi
 
-zstyle ':completion:*' list-colors ''
+if [ "x$LS_COLORS" = "x" ]; then
+    zstyle ':completion:*' list-colors ''
+else
+    zstyle ':completion:*' list-colors ${(s.:.)LS_COLORS}
+fi
 
 # should this be in keybindings?
 bindkey -M menuselect '^o' accept-and-infer-next-history


lib/completion.zsh
```diff
@@ -18,7 +18,11 @@ else
   zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
 fi
 
-zstyle ':completion:*' list-colors ''
+if [ "x$LS_COLORS" = "x" ]; then
```
apjanke: `if [[ -z "$LS_COLORS" ]] ...` would be more idiomatic `zsh` style.

apjanke:
:+1:

I'm doing the same thing in my own custom ZSH code.

If you wanted this to automatically take in to account a theme's `$LS_COLORS`, yeah, `theme-and-appearance.zsh` would be a good place. The problem is that the initial theme loading in `oh-my-zsh.zsh` does not use `theme-and-appearance.zsh`; it just [copy-pastes its own code for finding and sourcing the theme files](https://github.com/robbyrussell/oh-my-zsh/blob/599831b85a5fd4564ab820530e2bb45f37ca3f99/oh-my-zsh.sh#L81-L98).

I have a PR #3743 out that restructures the theme loading code, so that all theme loading is done through a `theme` function, including the main `oh-my-zsh.zsh` initialization. There's a central [theme loading implementation function](https://github.com/apjanke/oh-my-zsh/blob/ad4b017018721c69dca487f0cac9a92cc8186778/lib/theme-and-appearance.zsh#L76-L91), which would be a good place to put it. Right [here](https://github.com/apjanke/oh-my-zsh/blob/ad4b017018721c69dca487f0cac9a92cc8186778/lib/theme-and-appearance.zsh#L229), in fact.

PR #2012, opened 2013-07-29
Fix for bad locale entry in lib/misc.zsh

frozen707:
LC_CTYPE was being set to $LANG, which is bad because
1. LANG expects a list, and LC_CTYPE expects a single value.
2. LC_CTYPE can be set by the environment at a different value than LANG, to cope with non-ASCII applications in a Latin environment, mainly for Asian languages.

It now checks if it is set by the environment, and if not picks the first entry of LANG or a safe default.


vincentbernat:
Just stumbled randomly on this issue.

`LANG` is not expecting a list. `LANG` is expecting a single locale. `LANGUAGE` is expecting a list of languages (this is used by GNU gettext). Setting `LC_CTYPE` to `LANG` when neither `LC_ALL` nor `LC_CTYPE` are defined is redundant. Applications are already using `LANG` in this case. See the [GNU Libc manual](http://www.gnu.org/software/libc/manual/html_node/Using-gettextized-software.html).

So #3563 is a noop.

From my point of view, people should just set `LANG` correctly. Many programs will barf otherwise. Otherwise, something like that:

```sh
(($+LC_CTYPE || $+LC_ALL || $+LANG)) || {
  local -a available
  available=("${(f)$(locale -a)}")
  export LC_CTYPE=${${${${(@M)available:#*.UTF-8}[1]}:-${${(@M)available:#*.utf8}[1]}}:-C}
}
```

mcornella:
Hi @vincentbernat, sorry for posting so late. I appreciate some clarity on this (finally! :smile:), please post a PR with your solution and we'll test if it works well.

vincentbernat:
Unfortunately, from all the bug reports, I still don't understand what we are trying to fix. From #4091, it seems that some people have `LC_ALL=C` set. So, the solution I proposed above doesn't quite fix this kind of situation. I thought that the problem was for people where no variable at all were set. Overriding a user value seems more difficult.

The current code is the following:

```sh
# only define LC_CTYPE if undefined
if [[ -z "$LC_CTYPE" && -z "$LC_ALL" ]]; then
	export LC_CTYPE=${LANG%%:*} # pick the first entry from LANG
fi
```

Two problems:
 1. `LANG` is not a list but a single locale
 2. If `LC_CTYPE` is not defined, its value is the same as for `LC_ALL`. If `LC_ALL` is not defined, its value is the same as `LANG`.

I can do a PR just removing this block of code.

Thinking a bit about it, maybe the problem solved by this code is when people use a list in `LANG` (do you do that?). Its content becomes invalid and therefore defining `LC_CTYPE` to be the first component saves the day. However, the fix is easy: don't use a list. If you want a list, use `LANGUAGE` (which is a list of languages, not a list of locales).

mcornella:
The main issue is when people use a locale that is not utf8 compatible. So by overriding `LC_CTYPE` we ensure unicode characters are handled correctly. If `LC_ALL` is set we can't do anything about it, but let's supose it's set on purpose by the user and with unicode support. That part doesn't worry me. 

The part that's important is that if `LC_CTYPE` doesn't use unicode we should set it to use unicode in the best way possible, similar to how this PR had it: setting it to locale known by the user but with UTF-8. So it's true that as is now nothing is done if, for example, `LC_CTYPE=C`. So we should change that. How does your snippet of code work exactly?

vincentbernat:
This one?

```sh
(($+LC_CTYPE || $+LC_ALL || $+LANG)) || {
  local -a available
  available=("${(f)$(locale -a)}")
  export LC_CTYPE=${${${${(@M)available:#*.UTF-8}[1]}:-${${(@M)available:#*.utf8}[1]}}:-C}
}
```

If neither `LC_CTYPE`, `LC_ALL` nor `LANG` is set, put the output of `locale -a` into the `available` array and take the first member matching `*.UTF-8` and if there is nothing, take the first member matching `*.utf8` and if there is nothing, takes `C` (but you could default to `C.UTF-8`).

The last line could be cut in several variables:

```sh
${(@M)available:#*.UTF-8}
${(@M)available:#*.utf8}
```

In `zshexpn(1)`, search for `${name:#pattern}` to understand this notation.

We could do something like that:

```sh
LC_CTYPE=${LC_CTYPE:-${LC_ALL:-${LANG:-C}}
case $LC_CTYPE in
   *.utf8|*.UTF-8)
    # All is correct
    ;;
   *)
    # Need to select an UTF-8 locale
    local -a available
    available=("${(f)$(locale -a)}")
    export LC_CTYPE=${${${${(@M)available:#*.UTF-8}[1]}:-${${(@M)available:#*.utf8}[1]}}:-C.UTF-8}
   ;;
esac
export LC_CTYPE
```

The `export LC_CTYPE` could be ignored if we only want to set it for the current shell, not any subprocess.

This way, if `LC_CTYPE` is already set correctly, we do nothing, otherwise, we select some UTF-8 locale.

I remember that C.UTF-8 is fairly recent (less than 5 years I think), so we cannot use it inconditionnaly.

Also, I absolutely don't know if you cannot just use `inexistantlocale.UTF-8` and still get something correct. That would make `C.UTF-8` always work and simplify a lot.

Also, locales can be aliased and an UTF-8 locale may not have `UTF-8` nor `utf8` in its name. However, I don't think this is reasonable to handle such special cases.

Moreover, if a user set `LANG` to `fr_FR@euro`, it is likely that their terminal doesn't support UTF-8 and they will get odd results. Generally, if the locale is not UTF-8, the user will likely run into other problems with other programs.

mcornella:
Sorry @vincentbernat, I forgot about this. Would you care to submit a PR? I like your last proposal a lot, you've made it very understandable thank you.

vincentbernat:
OK done. If people that were hit by this issue could test...

PR #2054, opened 2013-08-20
make git_prompt_status more accurate

foobacca:
The git_prompt_status function only matched some of the possible output
from `git status --porcelain --branch` - this commit should match pretty
much the whole range, defined by the git status man page as:

```
X          Y     Meaning
-------------------------------------------------
          [MD]   not updated
M        [ MD]   updated in index
A        [ MD]   added to index
D         [ M]   deleted from index
R        [ MD]   renamed in index
C        [ MD]   copied in index
[MARC]           index and work tree matches
[ MARC]     M    work tree changed since index
[ MARC]     D    deleted in work tree
-------------------------------------------------
D           D    unmerged, both deleted
A           U    unmerged, added by us
U           D    unmerged, deleted by them
U           A    unmerged, added by them
D           U    unmerged, deleted by us
A           A    unmerged, both added
U           U    unmerged, both modified
-------------------------------------------------
?           ?    untracked
!           !    ignored
-------------------------------------------------
```

Note that with the current variables it doesn't really distinguish
between changes in the staging area and changes in the working directory
- to do that would require a lot more ZSH_THEME_GIT_PROMPT_XYZ variables.

Also git status no longer seems to use the text "diverged" - instead it
says both "ahead" and "behind" if you have diverged.  The second commit
fixes that logic.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -74,32 +74,39 @@ git_prompt_status() {
   if $(echo "$INDEX" | grep -E '^\?\? ' &> /dev/null); then
     STATUS="$ZSH_THEME_GIT_PROMPT_UNTRACKED$STATUS"
   fi
-  if $(echo "$INDEX" | grep '^A  ' &> /dev/null); then
+  if $(echo "$INDEX" | grep '^A[ MDAU] ' &> /dev/null); then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Similar to other PR, @foobacca , doing some chore/housekeeping on the project.

PR #2055, opened 2013-08-22
Added a plugin that adds autocomplete to the adhearsion gem

JustinAiken:
This adds autocomplete for [adhearsion](https://github.com/adhearsion/adhearsion), a Ruby framework for creating telephone applications.

base autocomplete
2nd level autocomplete


mcornella:
Not popular enough.

PR #2057, opened 2013-08-23
add git_prompt_status2

foobacca:
More detailed git status command, with prefixes for changes in staging
and working areas, and unmerged changes.

I guess this could replace git_prompt_status, but is does use the variables in a slightly different way and use a few more.

See also pull #2054

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -114,6 +114,84 @@ git_prompt_status() {
   echo $STATUS
 }
 
+# Get the status of the working tree
+git_prompt_status2() {
+  INDEX=$(command git status --porcelain -b 2> /dev/null)
+  if [[ -z $INDEX ]]; then
+    return
+  fi
+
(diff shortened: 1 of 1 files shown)


apjanke:
46 commits in this PR now. Could it be rebased and squashed?

This PR also captured a `gitignore` plugin from your custom dir at `custom/plugins/gitignore/gitignore.plugin.zsh`. Was that intentional?

foobacca:
@apjanke thanks for looking again at this.  I had let this become my personal branch - I've now changed that so this is a single commit on top of the current master, and the gitignore file has been removed.

robbyrussell:
We've since made a lot of changes for the git prompt. Am going to close this out as a housekeeping chore. If we need to re-reference this in the future, we have everything documented. I appreciate the original help, @foobacca I'm sorry we never quite got this fully merged.

PR #2058, opened 2013-08-23
Added more levels of zeus compeletion

JustinAiken:
in action


JustinAiken:
Thanks! I still have one project left using zeus, haha

PR #2063, opened 2013-08-27
gpg-agent plugin service check changed

cdimartino:
Running on OsX 10.8:

When using gpg-agent plugin, found numerous instances of gpg-agent running on my system, growing linearly with time.  Found existing check for running services is calling a deprecated command.

gpg-agent -q has the same effect.

image

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -15,7 +15,7 @@ function start_agent_withssh {
 }
 
 # check if another agent is running
-if ! gpg-connect-agent --quiet /bye > /dev/null 2> /dev/null; then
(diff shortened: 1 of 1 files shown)


cdimartino:
This change no longer appears necessary, as gpg-agent-connect properly returns a non-zero exit code if gpg-agent is not running.  Feel free to close without merge, or go forward and merge at your discretion.  Both methods appear to be valid for checking that gpg-agent is running.

mcornella:
Ok, I'm not sure that it's the final solution, but if you say that as is it already works, let's close this. Obviously in some version it didn't work, but it seems no one is complaining now...

There are other ways to check if it's running, I know prezto uses a different one and [this guy here](http://serverfault.com/a/481312) another more complex one, but I couldn't find the ultimate best practice.

Agree to close this :+1:

PR #2075, opened 2013-09-03
update tmuxinator plugin to latest tag from project

billwanjohi:
taken directly from https://github.com/aziz/tmuxinator/blob/v0.6.5/completion/tmuxinator.zsh

This makes completion work for me, whereas the existing plugin does not.


billwanjohi:
LOL, this takes me back.

PR #2083, opened 2013-09-09
Enable completion for maven colorizer

nmikhailov:
Maven plugin introduces mvn-color, but completion for it doesn't work(and after aliasing mvn-color to mvn it doesn't work too for mvn).


fmunch:
Any news on this?

PR #2113, opened 2013-09-25
Add ability to disable theme

andyhite:
In order for the zsh flavor of lokaltog/powerline and oh-my-zsh to play nice, the prompt theming in oh-my-zsh needs to be disabled. This pull request adds a $DISABLE_THEME option, that when set to "true" will bypass loading oh-my-zsh themeing.


mcornella:
So... ready to close right?

ncanceill:
Yes, @robbyrussell please close.

Issue #2121, opened 2013-09-28
gpg-agent plugin should check existence of ssh-agent before sourcing from GPG_ENV file


eakret:
The gpg-agent plugin needs to check for the existence of an ssh-agent before sourcing the `GPG_ENV` file. The `GPG_ENV` file potentially contains `SSH_AUTH_SOCK` and `SSH_AGENT_PID` for an old gpg-agent started with `--enable-ssh-support`. In this case, sourcing the file and then checking for ssh-agent's existence results in a failure since the `ssh-add -l` check uses the stale environmental variables from `GPG_ENV` rather than the current values of the live ssh-agent.

mcornella:
Triaged.

PR #2122, opened 2013-09-28
Check ssh-agent liveness before sourcing GPG_ENV

eakret:
Resolves issue #2121

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -16,9 +16,24 @@ function start_agent_withssh {
 
 # check if another agent is running
 if ! gpg-connect-agent --quiet /bye > /dev/null 2> /dev/null; then
+    # check if ssh-agent is running
+    local SSH_RUNNING=1
+    local OLD_SSH_AUTH_SOCK=
+    local OLD_SSH_AGENT_PID=
+    if [[ -n "$SSH_AGENT_PID" ]]; then
+        kill -0 $SSH_AGENT_PID
+        SSH_RUNNING=$?
+        if [[ $SSH_RUNNING -eq 0 ]]; then
+            OLD_SSH_AUTH_SOCK="$SSH_AUTH_SOCK"
+            OLD_SSH_AGENT_PID="$SSH_AGENT_PID"
+        fi
+    fi
+
     # source settings of old agent, if applicable
     if [ -f "${GPG_ENV}" ]; then
         . ${GPG_ENV} > /dev/null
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @eakret, there are pull requests #2063 and #2497 that both change gpg-agent plugin. 
Could you guys (@cdimartino, @blueben) come together and see if you're trying to achieve the same thing? If you *are* trying to solve the same thing, divise a common strategy to solve it. If not, explain to me in layman terms what the differences are, since I don't have a good understanding of what gpg-agent and ssh-agent do or how they are intertwined...

Also, a quick start guide on gpg-agent would be great to test any of these pull requests. Thanks!

cdimartino:
My understanding is that there are two bugs in the script which are causing
similar issues.  My fix https://github.com/robbyrussell/oh-my-zsh/pull/2063
resolves an issue where multiple gpg-agent instances were being spawned due
to a faulty check to see if gpg-agent is already running.  It actually
appears though that the gpg-agent-connect command is now functioning
properly, returning a non-zero exit status if gpg-agent is not running.  My
pull req can be closed if desired.  Either method appears to be a valid way
of checking for a running process now.

blueben:
Long story short, gpg-agent can act as an SSH key agent (--enable-ssh-support) as well as a GPG key agent. That's fine if folks want to use gpg agent for that. However, gpg agent shouldn't try to be an ssh agent if an ssh agent is already running, which is the problem logged in https://github.com/robbyrussell/oh-my-zsh/issues/2496.

My fix has a somewhat more reliable test to determine if an ssh-agent is already running. Doesn't seem to collide with @cdimartino's issue or fix.

mcornella:
Ok, I've had the chance to look into it and I've learned some more. @blueben your PR is simpler and easier to merge, but there's the problem that @eakret mentions in #2121. 

Basically, if ssh-agent is running and our `GPG_ENV` file contains `SSH_*` variables other than current ones (set up by ssh-agent), the check won't return a reliable result because the `SSH_AGENT_PID` will be different than the currently running ssh-agent.

That said, I'm not satisfied with this PR either: it's too complex to understand. A simpler version would be checking for an ssh-agent running before (and after) sourcing the `GPG_ENV` file.

eakret:
Whatever patch is added does need to save away the environmental variables
for the ssh agent before sourcing the GPG_ENV file. If it checks for ssh
agent liveness before sourcing GPG_ENV and ssh agent is alive, but no
longer showing as alive after sourcing GPG_ENV, then it needs to set those
variables back to their previous values.

However, I think the cleanest answer is to remove the ssh liveness check
entirely and just have two plugins: one that starts gpg agent with the
built-in ssh agent functionality and one that just starts the standard gpg
agent.

PR #2137, opened 2013-10-02
Add exemplary custom templates folder.

Profpatsch:
One can’t see directly which parts can be overwritten in the `custom` folder.

Can you overwrite every folder? Then maybe there should be a `README` in the custom folder.

### DIFF
--- a/custom/themes/your-theme.zsh-theme
+++ b/custom/themes/your-theme.zsh-theme
@@ -1,0 +1,1 @@
+# Put your custom themes in this folder.


mcornella:
This is still relevant but I recall some other PR that used an `example` name instead of `your-theme`.
:+1:

Profpatsch:
Yes, the project is kind of slow-moving because the world stops until @robbyrussell has time to merge in PRs. I often wondered how well a system like [ZeroMQ’s](http://zguide.zeromq.org/page:all#Chapter-The-ZeroMQ-Community) would work for oh-my-zsh.

mcornella:
Pulled in 6c9af44.

PR #2154, opened 2013-10-10
Add Magento plugin

quafzi:
This plugin adds autocompletion for a common Magento development tool.


robbyrussell:
@quafzi Can you include a REDME file?

mcornella:
Hi @quafzi, we merged #2367 instead. Can you please close this?

Issue #2210, opened 2013-10-30
virtualenvwrapper plugin: Cannot find virtualenvwrapper_lazy.sh


uskudnik:
OSX Mavericks (10.9), but the issue appeared in 10.8 as well (lazy me only got to fixing it now)

Whenever I start a new terminal session or source ~/.zshrc I get the following message:

```
 zsh virtualenvwrapper plugin: Cannot find virtualenvwrapper_lazy.sh. Please install with `pip install virtualenvwrapper`.
```

Naturally, virtualenvwrapper is already installed:

```
$ pip install virtualenvwrapper
Requirement already satisfied (use --upgrade to upgrade): virtualenvwrapper in /Library/Python/2.7/site-packages
Requirement already satisfied (use --upgrade to upgrade): virtualenv in /Library/Python/2.7/site-packages (from virtualenvwrapper)
Requirement already satisfied (use --upgrade to upgrade): virtualenv-clone in /Library/Python/2.7/site-packages (from virtualenvwrapper)
Requirement already satisfied (use --upgrade to upgrade): stevedore in /Library/Python/2.7/site-packages (from virtualenvwrapper)
```

I tried specifying `/usr/local/bin/virtualenvwrapper_lazy.sh` instead of `virtualenvwrapper_lazy.sh` in `~/.oh-my-zsh/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh`, but even that didn't work. Also, ran `chown` and `chmod 755` on it, nothing helped:

```
$ ls -lh /usr/local/bin/virtualenvwrapper_lazy.sh
-rwxr-xr-x  1 username  group   1.4K Oct 30 11:17 /usr/local/bin/virtualenvwrapper_lazy.sh
```

Any ideas why this would be happening?

PR #2211, opened 2013-10-30
Fixed checking of whether virtualenvwrapper exists

uskudnik:
In response to https://github.com/robbyrussell/oh-my-zsh/issues/2210.

Do note, there may be a better/proper way of doing it, I am not a bash magician.


mcornella:
This can be closed. It's just a matter of setting the PATH right, and both versions do the same thing.

PR #2214, opened 2013-10-30
Add an option about git submodules to ignore

jokester:
Hi there

I found that when checking whether a git repo is clean, `lib/git.zsh` always ignore dirty submodules,
no matter what was specified in `.gitmodules` file.
So I added an option to enable overriding this behavior.
When this option is not set, it should work as before. See commit message for more details.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -17,6 +17,19 @@ function parse_git_dirty() {
     if [[ "$DISABLE_UNTRACKED_FILES_DIRTY" == "true" ]]; then
       FLAGS+='--untracked-files=no'
     fi
+    case "$GIT_STATUS_IGNORE_SUBMODULES" in
+      "")
+        # if unset: ignore dirty submodules
+        FLAGS+="--ignore-submodules=dirty"
+        ;;
+      "git")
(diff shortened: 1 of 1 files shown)


oryband:
@jokester in my case, i have my dotvim repo as a submodule of my dotfiles. I set dotvim's submodule's `ignore = all` because i just want dotfiels to link to dotvim, but not track its tree.

I agree with you - it is best to let `git status` follow `.gitmodule` by default - because this gives the user most control of the status.

Can someone please merge already? I've been waiting for 6 months or so for this to be merged!

jokester:
Just rebased onto current master. This patch is still conservative: only follows `.gitmodule` when `GIT_STATUS_IGNORE_SUBMODULES=git` is set. 
 
(I'm totally open on changing the default behavior, after all it's configurable now)

Darmody:
How is this going? I faced the git submodules problem recently.

jokester:
rebased onto recent `master`.

It took longer than I imagined to be either or merged/rejected :thinking:

mcornella:
Thanks for the patience

jokester:
Thanks @mcornella  🎆

PR #2224, opened 2013-11-05
Add terminal Google Translate using termit gem.

pawurb:
Hello. I would like to suggest including a [Termit](https://github.com/pawurb/termit) gem as a plugin. It allows a user to use all the Google Translate goodies (together with speech synthesis) from the terminal. It's only dependency is ruby in version from 1.9.2. Executable is aliased to translate.


mcornella:
This is more a command than a plugin.

PR #2236, opened 2013-11-08
Added function to show current Node.js version and new Bureau theme

isqua:
I've added function `nvm_prompt_info()` that will show you current Node.js version wrapped in prefix and suffix. E.g.:

``` bash
 %  ZSH_THEME_NVM_PROMPT_PREFIX="[ Node v"
 %  ZSH_THEME_NVM_PROMPT_SUFFIX=" ]"
 %  nvm_prompt_info
[ Node v0.10.15 ]
```

And added new Bureau theme:
bureau
Time, git status and Node.js version always are right-aligned.

### DIFF
--- a/themes/bureau.zsh-theme
+++ b/themes/bureau.zsh-theme
@@ -1,0 +1,113 @@
+# oh-my-zsh Bureau Theme
+
+### NVM
+
+ZSH_THEME_NVM_PROMPT_PREFIX="%B⬡%b "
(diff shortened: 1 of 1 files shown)


apjanke:
FYI, I added proposed alternate install location support in #4338. You can point it any directory by setting `$NVM_DIR` before loading OMZ. In your case, `export NVM_DIR=$HOME/dotfiles/nvm` and add `nvm` to your `$plugins` list. And the revised prompt function will recognize `nvm` loaded from anywhere.

PR #2281, opened 2013-11-27
Added CanIUse plugin

walesmd:
This plugin adds a command-line utility to easily search the [Can I Use...](http://caniuse.com) website.


mcornella:
You can use the `frontend-search` plugin for that.

PR #2312, opened 2013-12-04
Changed determination nvm in system

isqua:
As noted by @deepsweet in #2236, check if `~/.nvm/nvm.sh` exists is not the best way to determine is there an `nvm` command or not.

I replaced it with `command -v`.

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -1,3 +1,4 @@
 # The addition 'nvm install' attempts in ~/.profile
 
 [[ -s ~/.nvm/nvm.sh ]] && . ~/.nvm/nvm.sh
+[[ -r $NVM_DIR/bash_completion ]] && . $NVM_DIR/bash_completion
--- a/lib/nvm.zsh
+++ b/lib/nvm.zsh
@@ -1,6 +1,9 @@
 # get the node.js version
 function nvm_prompt_info() {
-  [ -f "$HOME/.nvm/nvm.sh" ] || return
+  if [[ "$(command -v nvm)" != "nvm" || -z $(echo $PATH | grep "$HOME/.nvm") ]];
+    then return
+  fi
+
   local nvm_prompt
   nvm_prompt=$(node -v 2>/dev/null)
   [[ "${nvm_prompt}x" == "x" ]] && return


isquariel:
I’ve just rebased the branch from master.

mcornella:
Hey @isquariel, on a second thought, maybe the best way is using [`whence -p`](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html#index-whence), in case nvm is an alias.

isquariel:
@mcornella, `whence -p nvm` returns 1 because [nvm is just a shell function](https://github.com/creationix/nvm/blob/master/nvm.sh#L542) may be it would be good to check `whence -w nvm` return 0 or `whence -w nvm | grep function` return 0? Which case is better?

mcornella:
Nevermind, in that case it's better to use `command` which will return 0 whatever the type of `nvm` is.
:+1: 

/cc @robbyrussell

plugins/nvm/nvm.plugin.zsh
```diff
@@ -1,3 +1,4 @@
 # The addition 'nvm install' attempts in ~/.profile
 
 [[ -s ~/.nvm/nvm.sh ]] && . ~/.nvm/nvm.sh
+[[ -r $NVM_DIR/bash_completion ]] && . $NVM_DIR/bash_completion
```
isquariel: Add nvm autocomplete.

mcornella: Does this work out of the box? To load bash completion files, [`bashcompinit`](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#index-bashcompinit) needs to be called.

isquariel: I use only `git`, `nvm` and `ssh-agent` plugins (and don’t use [`compleat`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/compleat/compleat.plugin.zsh) plugin). And nvm autocomplete works for me.

zsh 5.0.2 (x86_64-apple-darwin13.0), latest oh-my-zsh

But I’m not sure about no-apple zsh. How to check the plugin?

mcornella: You can check if bashcompinit is defined with `which bashcompinit`. 
You can also delete the zcomdump files (`rm ~/.zcomdump*`) and restart the terminal to see if it works. 
It can also be that nvm.sh is doing that for us, you could have a look there.

Finally, it is surprising to me how only the `compleat` plugin has a bashcompinit statement, nowhere else it is found in the oh-my-zsh repository! 
Let's see what you can find!

isqua: I just try it on fresh CentOS. Suddenly it work as is in this branch.

isqua:
Just rebase.

apjanke:
There seem to be a couple issues with this.

Using `command` to detect the `nvm` installation location won't work either in the general case. Because `nvm` is a shell function defined in `nvm.sh`, you can't use it to locate `nvm.sh`, because it is not defined until after you have already located and run `nvm.sh`. (Also, the output will just give you the body of the function, not tell you what file it was loaded from.) Using `which nvm` or `command nvm` is begging the question.

This change also seems to require the user to set `$NVM_DIR` before loading OMZ. Normally, it is `nvm.sh` which sets `$NVM_DIR` based on where you load it from. And after the plugin runs, `$NVM_DIR` will be set regardless, so the prompt info function could just use that instead of doing the search each time. As is, it'll fall back to using `~/.nvm`, regardless of where you actually loaded it from. Seems like this should be optional, to allow the user to force a particular location that isn't one of the well-known installation locations.

I think the thing to do instead is:
* Have the `nvm` plugin locate `nvm` by looking in a few well-known installation locations for `nvm.sh`
 * `$NVM_DIR`, if it is already defined
 * `$HOME/.nvm
 * `$(brew --prefix nvm)`, if `brew` is installed
 * (some other location(s) from other platforms' package systems?)
* Have the `nvm_prompt_info` function in `lib/nvm.zsh` use only `$NVM_DIR`.
 * Or have it not use the filesystem at all, and just check `which nvm` to see if nvm is loaded in this session.

Because `lib/nvm.zsh` does not express a dependency on `plugins/nvm`, I think it should probably be written to be compatible with as many different `nvm` setups as possible, not just the ones that `plugins/nvm` knows how to load. Using `which nvm` is probably the best test: it's possible that a user could define a custom `$NVM_DIR`, but not actually load `nvm` in all of their shell sessions.

plugins/nvm/nvm.plugin.zsh
apjanke: NVM does [load `bashcompinit` itself, as part of loading that bash completion file](https://github.com/creationix/nvm/blob/1e2a7a8de447c9b872fbf4cbe6c959cfc8d02a55/bash_completion#L84-L89) (but not as part of `nvm.sh`, I think).

But you should not need to use it, because `plugins/nvm` provides a zsh-specific `_nvm` completion function. The bash completion file's location under `$NVM_DIR` may change depending on the install method; e.g. Homebrew's formula relocates it to `$(brew --prefix nvm)/etc/bash_completion.d/nvm`.

isqua:
Closed in favor of #4338

PR #2351, opened 2013-12-15
A paver (Python build and configuration utility) plugin

frantzmiccoli:
This add a plugin for Paver: http://paver.github.io/paver/ it has been tested on OSX, but not on Linux (however it should work also smoothly there).


Amwam:
+1 :+1:

ncanceill:
Well, #2922 has already been merged in, but looks like this is more than ready.

@robbyrussell please merge.

PR #2354, opened 2013-12-16
local foopath disabled

Bandie:
To get rid of the error that "/home/scp1/bin/foopath" wouldn't exist.


mcornella:
Closable, this was already fixed.

PR #2356, opened 2013-12-17
Fix: __git_complete_remote_or_refspec:35: bad pattern: +*

ayanamist:
I'm using gitfast plugin instead of git.
Since i upgrade my debian sid to the latest, when i use `git push -u origin somebranch` and press tab, it will output 3 lines:
> __git_complete_remote_or_refspec:35: bad pattern: +*
__git_complete_remote_or_refspec:35: bad pattern: +*
__git_complete_remote_or_refspec:35: bad pattern: +*

So i check the source and make the modification to keep the function works.


mcornella:
Any updates @ayanamist ?

PR #2372, opened 2013-12-23
add profile auto completion for maven

rockybean:
The script will filter all profile id from pom.xml in current directory and support profile auto completion for '-P\<tab\>'.


mcornella:
Closing in favor of #2903

PR #2376, opened 2013-12-25
Updated virtualenvwrapper plugin

neiljdo:
I modified the plugin to look first in the pwd (but making sure that it is under the git project root) for the .venv config file. Previously, it looks immediately under the git project root.


mcornella:
Hi @amontalenti, did you by any chance got to it?

apjanke:
Superseded by #3918, which does a similar `.venv` check, but walks all the way up the file path. You can close this.

PR #2383, opened 2013-12-30
Fix itchy theme rvm-prompt invocation by silencing error messages

Nitrodist:
If rvm-prompt or rbenv isn't found within the system, then the
following error message is displayed:

zsh: command not found: rvm-prompt

Silencing standard error while invoking these commands fixes this
problem.


mcornella:
Fixed in master.

PR #2386, opened 2013-12-30
Fixes git_compare_version() in lib/git.zsh and refactors the whole file

LFDM:
git_compare_version() is used to determine whether dirty submodules should be ignored or not, but git has to be more recent than version 1.7.2 to know about that option.

With the former comparison logic a check if git version 1.8.1 is more recent than 1.7.2 turned out negative, which is fixed through this pull request.

UPDATE: In addition to #2401 I refactored  the whole file.
- Cleans up it's functions
- Less duplication of code and logic
- Less querying of git and the general status,
- Adds functions that where buried inside of others before (git_stash_status, git_change_status) to gain better modularity.
- Adds a big documentation section where every function and customization variable is explained.
- Adds a function to hide/show the git prompt in specific large repositories - toggle_git_prompt_visibility

Might need some more testing if anything was broken, though I don't have any troubles locally.

Possibly breaking changes:
- Cuts the duplicated variables to show the git_remote_status.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -17,7 +17,7 @@
 #   check_git_show_status below)
 #   Dirty submodules are not tracked f your git version is post 1.7.2.
 #   To ignore untracked files making your branch dirty, set
-#   DISABLE_UNTRACKED_FILES_DIRTY to 'true
+#   DISABLE_UNTRACKED_FILES_DIRTY to 'true'
 #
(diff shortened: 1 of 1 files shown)


robbyrussell:
Going to close this as it's been lingering for nearly a decade. We've made a lot of changes since then with our git prompt code, so if this is an issue now, we can re-open/reference this as necessary.

Appreciate the initial push, @LFDM!

PR #2392, opened 2014-01-01
Geeknote autocomplete and alias

s7anley:


mcornella:
Oops. This is really old. Sorry for losing track of this!

s7anley:
@mcornella thanks for merge.

Issue #2295, opened 2013-12-03
Document $ZSH_CUSTOM


robbyrussell:
If someone has some time... can you write up a short tutorial on using $ZSH_CUSTOM for custom themes/plugins and toss that in the wiki for folks? It'd be great to direct folks to when they are curious about where to stick their custom themes going forward.

EricSB:
I know this is a bit of an old issue so I apologize for adding a comment so late in the game, but wanted to say the content you added is fantastic!  The one thing I'd love to see added is perhaps an example of how to place custom aliases in the custom directory given it's mentioned in the .zshrc file.

Thanks in advance!
-Eric

PR #2401, opened 2014-01-05
Refactor of init script

LFDM:
A completely refactored oh-my-zsh script. This works on top of the commits I mentioned in #2295 and also the commit of #2113. (read first)

It's basically a big clean up, that shouldn't really break anything - it works almost completely the same, just tidier.
- Wraps each step up in an own function, which makes the code more self documenting.
- Refactored out all duplication
  - Checking twice for plugins erased. The plugin names are expanded to their full plugin path now just once, whether it resides in the custom or the default directory. They are stuffed in a variable which is reused (check comment starting on line 8).
  - Using variables instead of writing the paths out 4 times on 4 lines.
  - Sourcing of files wrapped in a function.
- Deletes the addition of $ZSH/functions and $ZSH/completions to fpath - those two directories do no exist (any more?)
- Fix of #2342 is applied.

Breaking changes:
- Plugins are only changed against the *.plugin.zsh naming convention, the variant with underscore is not checked. If that breaks plugins shipped with oh-my-zsh, I will change their names.
- Custom themes have to be in a themes folder and not at the top level of the custom dir - as it would be documented in the wiki text I wrote.
- I don't think both deprecated options are widely used anyway, but I could be wrong.

Proposal:
- Adding a changelog file, that is displayed in upgrade.sh. It might not have to all complete at all times, but in case any relevant changes are merged in, this could be a good place to inform people about was might affect them after upgrading.

Probably controversial:
- The file is mildly converted to zsh syntax. I think it's safe to assume that people will use zsh when they source this file :). Can be reverted back in a single step, however, I placed a symbolic link to avoid breaking backward compatibility. People sourcing oh-my-zsh.sh in their .zshrc will be redirected to the new oh-my-zsh.zsh file.

A question left:
- Could someone check the message of commit 7e79ed795f9d392de460efc87d1a787e2400eee7? It's regarding adding the plugin paths to fpath. The refactor does now exactly as the former script, but I am actually not sure that it's needed. Can someone explain this?

If people want to dig in further how oh-my-zsh works, this will be the file they first look at - it's easier to get an overview this way IMO.

Curious to hear your opinions.

Update: Also revampled lib/git.zsh - see #2386.

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -60,9 +60,8 @@ the "refcard":http://www.bash2zsh.com/zsh_refcard/refcard.pdf is pretty tasty fo
 
 h3. Customization
 
-If you want to override any of the default behavior, just add a new file (ending in @.zsh@) into the @custom/@ directory.
-If you have many functions which go well together you can put them as a *.plugin.zsh file in the @custom/plugins/@ directory and then enable this plugin.
-If you would like to override the functionality of a plugin distributed with oh-my-zsh, create a plugin of the same name in the @custom/plugins/@ directory and it will be loaded instead of the one in @plugins/@.
+Pretty much every single aspect of _Oh My Zsh_ is configurable to your needs - without the burden of maintaining your own fork!
+Be sure to check out our "Customization wiki":https://github.com/robbyrussell/oh-my-zsh/wiki/Customization for more info.
 
 h3. Updates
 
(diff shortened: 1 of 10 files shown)


robbyrussell:
Given that we've sat on this for over 5 years, I think the lesson for us is that big refactoring PRs are going to be seen as too risky (and too time-consuming) to migrate with confidence. (perhaps if we had a sensible and automated testing solution in place)

I'm all for making changes but would am going to advocate for incremental refactorings.

PR #2404, opened 2014-01-07
Fix permission denied error in git-prompt-info

oohlaf:
When moving into a git repo of another user, an error was shown on every
prompt:
fatal: unable to access '.git/config': Permission denied


oohlaf:
not relevant any more, closing

PR #2411, opened 2014-01-08
Updated docker plugin to use Docker container names instead of hashes.

ghost:
Since version 0.6.5 every Docker container has unique name.
http://blog.docker.io/2013/10/docker-0-6-5-links-container-naming-advanced-port-redirects-host-integration/

Currently the docker plugin provides auto-complete based on hashes.
This commit changes it to names(container hash is still visible).

Additionally plugin now shows all container not only running ones(useful for many Docker commands (start, rm))

### DIFF
--- a/plugins/docker/_docker
+++ b/plugins/docker/_docker
@@ -9,7 +9,7 @@
 # Output a selectable list of all running docker containers
 __docker_containers() {
     declare -a cont_cmd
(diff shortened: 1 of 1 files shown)


leifcr:
This should be closed in favor of #5060, as the upstream includes more fixes.

PR #2422, opened 2014-01-10
Cleaning lib

### LINKED ISSUE #2412: cd ...... : extend with global aliases
We already have the following:

```
alias cd..='cd ..'
alias cd...='cd ../..'
alias cd....='cd ../../..'
alias cd.....='cd ../../../..'
```

I suggest we could add this also:

```
alias -g ...='../..'
alias -g ....='../../..'
alias -g .....='../../../..'
```

This way we could do things like `ls ...`.

(This could be tagged with 'idea', but I don't know how to do it)

LFDM:
- Provides a more useful approach to directory switching, as suggested in #2412
- Deletes `aliases.zsh`, as almost all things defined in their can find a place in other files present
  - The directory aliases (some of the which duplicate functionality) are all in `directories.zsh`
  - The history alias is in `history.zsh`. Also refactored to a nicer case switch
  - Remaining pieces like the please alias for sudo are in `misc.zsh`
  - Deletes the .. aliases from the `common-aliases` plugin, they are in `lib` anyway
- Moves a bindkey statement from `misc.zsh` to `key-bindings.zsh`
- Avoids sourcing of `autoenv` with every call to cd. This plugin related code is also taken out of the lib directory. Its activation file is now sourced only once - if its not present, the user is informed and the plugin disabled. Overriding cd by hand is not needed, autoenv does this by itself.

Closes #2412
Closes #2601

### DIFF
--- a/lib/aliases.zsh
+++ b/lib/aliases.zsh
@@ -4,19 +4,5 @@ alias please='sudo'
 
 #alias g='grep -in'
 
-# Show history
-if [ "$HIST_STAMPS" = "mm/dd/yyyy" ]
-then
-    alias history='fc -fl 1'
-elif [ "$HIST_STAMPS" = "dd.mm.yyyy" ]
-then
-    alias history='fc -El 1'
-elif [ "$HIST_STAMPS" = "yyyy-mm-dd" ]
-then
-    alias history='fc -il 1'
-else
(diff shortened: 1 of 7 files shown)


mcornella:
I've reposted this PR at #3564 solving merge conflicts.

PR #2445, opened 2014-01-18
Fixing iTerm Tmux integration (branch off upstream/master)

andschwa:
As per issue #1903, this actually fixes the iTerm integration. No more weird crashing.

Really I should just refactor this whole plugin, but it'll have to wait until spring break probably, or for someone to ask.

As before:

Essentially, if ZSH_TMUX_ITERM2 is set to true, the tmux plugin adds the option -CC to tmux whenever it is invoked. Enabling this also disables the loading of any "$HOME/.tmux.conf" configuration file, as they are not yet supported.
Tested working with AUTOCONNECT and AUTOQUIT.

The trick was instead of tmux attach || tmux new (which would fail because tmux attach, on failing to attach, itself tries to start the server) to check of it can attach with tmux has, and if so, attach, if not new session.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -31,7 +31,7 @@ if which tmux &> /dev/null
 	# The TERM to use for 256 color terminals.
 	# Tmux states this should be screen-256color, but you may need to change it on
 	# systems without the proper terminfo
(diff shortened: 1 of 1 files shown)


andschwa:
> from `unknown repository`

So that's what happens with open PRs from long since deleted repositories.

PR #2446, opened 2014-01-19
Emoji plugin

fallwith:
This plugin makes it easy to display emoji characters in a terminal. Usage instructions are at the top of the plugin file.


apjanke:
These constants are all defined by what look like hand-coded UTF-8 sequences. It would be more readable if you used `\u' escape sequences for Unicode code point values instead. That way people could just copy the code point number from reference sites instead of doing the UTF-8 encoding yourself. And it would be easier to compare code point values against mentions of them in other material.

Because the contents of the `${emoji[*]}` values are quoted with single quotes, they actually contain the literal escape sequences for the UTF-8 characters. They don't get interpreted until the `echo` or whatever call processes them. This means that they can only be used with commands that understand `\x` escape sequences, and it is sensitive to locale settings at the time of that command, not at the time that `$emoji` is populated. Maybe it would be better to interpret the escape sequences up front ("eager evaluation"), so the actual Unicode emoji characters are stored inside `$emoji`?

These will also get mangled if the user is in a non UTF-8 locale. (E.g. if they're in C or Shift-JIS or a UCS-16 locale.) You could protect it with a `local LC_ALL="" LC_CTYPE="" inside your functions that are processing the escape sequences, or during the initial population of `$emoji` if you did eager evaluation.

Could you add the `-H` option with `typeset -AH emoji` so these don't all get spewed in the output of `set`? It's a pretty long output.

Maybe `display_emoji_list()` should take `all` as one of its list names?

apjanke:
@mcornella, @ncanceill: On OS X, most applications will silently switch to using the Apple Color Emoji font for displaying characters in the emoji ranges, regardless of what your font is set to. I.e. in Terminal or iTerm2 on OS X 10.9, these emoji Just Work, regardless of what font I have selected in my profile, and they uniformly use Apple's colorful emoji symbols.

Maybe that's what's happening here. "Droid Sans Mono patched for Powerline" does not appear to actually have glyphs for most of these characters. (I loaded it on Windows, which doesn't have Apple's emoji hack, and it shows "missing character" placeholders for the great majority of these.) @ncanceill, you don't happen to be running OS X, do you?

There's [Open Sans Emoji](https://github.com/MorbZ/OpenSansEmoji) which might fit your needs on a non-Apple system. On Windows, it displays glyphs for all these characters, and they're in a black-and-white "wingding" style.

(IMHO, current emoji font handling is kind of a disgusting hack, as a result of many of our programs having simplistic font selection where you just pick one font for a whole program or text range. HTML handles it better, with lists of fallback fonts.)

fallwith:
Thanks @apjanke. This is great feedback.

I think when I originally put this together (1.5 years ago) I felt that the single quoted UTF-8 strings that get evaluated later would make things lighter weight. Your comments regarding Unicode's better readability, avoiding the potential locale issues, and the benefits resulting from eager evaluations all make sense and would definitely improve upon my original approach. Using `typeset -AH` is also a great suggestion.

Quite a few new emoji characters have been released since the pull request was opened. If I was to write this again I'd like to include functionality that could refresh the list of characters on demand or on a schedule, perhaps by scraping http://unicode.org/emoji/charts/full-emoji-list.html

I really appreciate you taking the time to take a look at this.

apjanke:
@mcornella: Maybe you called `random_emoji_from_list` instead? If it's not given an argument, it tries checking a nonexistent unnamed array for size, and gets division by zero.

```
$ random_emoji_from_list
random_emoji_from_list:3: division by zero
```

Maybe it should check its input and either have a default list it draws from, or output an error message, possibly including valid list names.

apjanke:
@fallwith: You're welcome!

apjanke:
Yo dawg, I heard you like PRs, so I put a PR on your PR.

fallwith/oh-my-zsh#1

This is a suggested implementation for doing the things I mentioned above – switching to `\U` encoding, autogenerating definitions file from Unicode data files, and guarding the locale. Also adds a README, broader national flag support, and experimental skin tone selection.

fallwith:
Awesome work, @apjanke 

Merged.

mcornella:
We haven't pulled this yet? [_Sacrebleu!!_](https://www.youtube.com/watch?v=jECd8HgdpOo)
Let's fix that
/cc @robbyrussell

mcornella:
BTW do we have a list of fonts that support this? My current one doesn't display some of them (`Meslo LG S for Powerline`)...

apjanke:
Support is probably going to depend on OS, toolkit, or terminal emulator support for emoji more than font support. The vast majority of fonts do not have emoji characters. (I do not know of *any* fixed-width fonts which include decent emoji coverage.) In most places where emoji are supported, AFAIK, the application or toolkit-supplied UI widgets detect emoji characters and silently switch fonts for just for those characters. (E.g. on OS X and iOS, native apps will substitute "Apple Color Emoji" for whatever font is being displayed.) For example, on OS X, I can run iTerm2 or Terminal.app with any font, and it still Just Works and displays emoji using Apple's emoji font.

None of the emoji fonts have coverage for all of these characters yet, so some of them are going to display as "missing character". (The list was built from the most recent Unicode emoji spec, and the fonts are still catching up.) www.iemoji.com has good pages for comparing individual glyphs between the major emoji fonts. There's probably some web site that has comparison charts for the full set, but I don't know what it would be.

There are probably some customized versions of fonts which have an emoji character set mashed in to them, the way that the Powerline-patched fonts have the special Powerline characters mashed in.

apjanke:
Aha! http://apps.timwhitlock.info/emoji/tables/unicode has big tables comparing glyphs across the major emoji fonts. (Be prepared for it to crush your browser.)

robbyrussell:
You folks amuse me. :+1:

PR #2452, opened 2014-01-20
dnf plugin for oh-my-zsh

### LINKED ISSUE #4184: update yum plugin or a new one
Fedora 22 is the actual stable version, and is not using yum anymore, but instead dnf as a replacement.

So the need of a new plugin or change the yum plugin comes.
May be the new commands can be:
dni | dnu | dns | dnr | etc... (just thinking on commands names related to dnf)

GioMac:
--------
EDITED by @mcornella

TODO:
- [x] Add README file
- [x] Add entry to [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins)

Duplicates:
Close #3427
Close #3959
Close #4208
Close #4215

Fix #4184

### DIFF
--- a/plugins/dnf/dnf.plugin.zsh
+++ b/plugins/dnf/dnf.plugin.zsh
@@ -1,0 +1,15 @@
+## Aliases
+
+alias dnfs="dnf search"                       # search package
+alias dnfp="dnf info"                         # show package info
+alias dnfl="dnf list"                         # list packages
+alias dnfgl="dnf grouplist"                   # list package groups
+alias dnfli="dnf list installed"              # print all installed packages
+alias dnfmc="dnf makecache"                   # Generate metadata cache
+
+alias dnfu="sudo dnf update"                  # upgrate packages
+alias dnfi="sudo dnf install"                 # install package
+alias dnfgi="sudo dnf groupinstall"           # install package group
+alias dnfr="sudo dnf erase"                  # remove package
+alias dnfgr="sudo dnf groupremove"            # remove pagage group
+alias dnfc="sudo dnf clean all"               # clean cache
--- a/plugins/dnf/dnf.plugin.zsh
+++ b/plugins/dnf/dnf.plugin.zsh
@@ -10,6 +10,6 @@ alias dnfmc="dnf makecache"                   # Generate metadata cache
 alias dnfu="sudo dnf update"                  # upgrate packages
 alias dnfi="sudo dnf install"                 # install package
 alias dnfgi="sudo dnf groupinstall"           # install package group
-alias dnfr="sudo dnf erase"                  # remove package
+alias dnfr="sudo dnf erase"                   # remove package
 alias dnfgr="sudo dnf groupremove"            # remove pagage group
 alias dnfc="sudo dnf clean all"               # clean cache


penthium2:
a work around : 
cp /usr/share/zsh/5.0.7/functions/_yum /usr/share/zsh/5.0.7/functions/_dnf
 then : 
vi /usr/share/zsh/5.0.7/functions/_dnf 
and under vim type: 
:%s/yum/dnf/g
save and exit vim 
relog under zsh and enjoy .

thndgonz:
@penthium2  That was really great, thanks for sharing!

sebastianbergmann:
I think #3897 should be merged instead of this.

guizmaii:
:+1:

mcornella:
Duplicates: 
#3427 @ppapadeas 
#3959 @fjospinas
#4208 @ah450
#4215 @ZJaffee

There are some differences between the aliases these PRs make, but my suggestion is that you close the other ones and discuss here the plugin that you want. These are not that different, so it's possible that you all agree on some set of aliases in a short time. 
My gut says that aliases of the type `ds` and `dns` will clash with some command some time from now and we'll have to change them, so it's better if you stick with `dnf*` style aliases.

As per the completion, you can do 2 things:

1. Use the current yum completion. For that you just need to add the line `compdef dnf=yum` to the plugin file.

2. Use the bash completion: For that you need to:

  1. Copy the [bash completion file](https://github.com/rpm-software-management/dnf/blob/master/etc/bash_completion.d/dnf-completion.bash) inside the dnf folder with the name `_dnf`.
  2. Add this line at the top of the newly created `_dnf` file: `autoload bashcompinit && bashcompinit`

Finally, you're gonna need to add a README file documenting the plugin, as well as add an entry to the [Plugins wiki page](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins).

ZJaffee:
I think the PR I made is what we should go with mostly because the way I feel it works has the most immediate logic, none of the commands made in my most recent update to the PR have no commands that already do something else. Furthermore, they all follow the structure of the first letter of each command represents the command you are using in most cases and for the ones it doesn't it references something close to what it would be for yum if we did the same thing there.

ah450:
@GioMac `update` and `erase` are deprecated aliases in dnf for `upgrade` and `remove`. 
Maybe the will be removed from dnf in the future? 

@ZJaffee perhaps it's better if all added aliases share the same prefix for autocompletion purposes? instead of the `s` prefix for the privileged commands as in #4215

ah450:
@mcornella doesn't http://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Redhat/Command/_dnf already take care of completion?

mcornella:
@ah450 I wasn't aware of that particular implementation. Note, though, that [this was commited just 2 months ago](http://sourceforge.net/p/zsh/code/ci/72ecc2ef36b203175aad3f6dd4b76e474a82d8b6/) and therefore it hasn't made it yet into [zsh 5.0.8](http://sourceforge.net/p/zsh/code/ci/zsh-5.0.8/tree/Completion/Redhat/Command/). But we could maybe add that.

> @ZJaffee perhaps it's better if all added aliases share the same prefix for autocompletion purposes? instead of the s prefix for the privileged commands as in #4215

I agree.

What do the rest of people have to say?

mcornella:
Hi guys, I've retouched a few things, including using `upgrade` and `remove` commands and adding a README. I'd like someone to take a look at it before merging it though: [mcornella/oh-my-zsh@add-dnf-plugin:/plugins/dnf](https://github.com/mcornella/oh-my-zsh/tree/add-dnf-plugin/plugins/dnf)

mcornella:
I just merged the modified version, I hope everything's ok! :rocket:

PR #2471, opened 2014-01-26
Adds support to capistrano multistage completion

phstc:
Uses `shipit` because [cap is a reserved word](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fcap-Module), so `cap` completion doesn't work.

```bash
$ shipit |
production staging

$ shipit production |
deploy                       -- Deploy a new release
deploy:bundle                -- Bundle
deploy:check                 -- Check required files and directories exist
deploy:check:directories     -- Check shared and release directories exist
deploy:check:linked_dirs     -- Check directories to be linked exist in shared
deploy:check:linked_files    -- Check files to be linked exist in shared
deploy:cleanup               -- Clean up old releases
deploy:cleanup_rollback      -- Remove and archive rolled-back release
deploy:finished              -- Finished
deploy:finishing             -- Finish the deployment, clean up server(s)
deploy:finishing_rollback    -- Finish the rollback, clean up server(s)
deploy:log_revision          -- Log details of the deploy
deploy:published             -- Published
deploy:publishing            -- Publish the release
deploy:restart               -- Restart application
deploy:revert_release        -- Revert to previous release timestamp
deploy:reverted              -- Reverted
deploy:reverting             -- Revert server(s) to previous release
deploy:rollback              -- Rollback to previous release
deploy:started               -- Started
deploy:starting              -- Start a deployment, make sure server(s) ready
deploy:symlink:linked_dirs   -- Symlink linked directories
deploy:symlink:linked_files  -- Symlink linked files
deploy:symlink:release       -- Symlink release to current
deploy:symlink:shared        -- Symlink files and directories from shared to release
deploy:updated               -- Updated
deploy:updating              -- Update server(s) by setting up a new release
```

### DIFF
--- a/plugins/capistrano/capistrano.plugin.zsh
+++ b/plugins/capistrano/capistrano.plugin.zsh
@@ -1,7 +1,7 @@
-# Added `xcap` because `cap` is a reserved word. `cap` completion doesn't work.
+# Added `shipit` because `cap` is a reserved word. `cap` completion doesn't work.
 # http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fcap-Module
 
-func xcap() {
+func shipit() {
   if [ -f Gemfile ]
   then
     bundle exec cap $*
--- a/plugins/capistrano/capistrano.plugin.zsh
+++ b/plugins/capistrano/capistrano.plugin.zsh
@@ -1,0 +1,11 @@
+# Added `xcap` because `cap` is a reserved word. `cap` completion doesn't work.
+# http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fcap-Module
+
+func xcap() {
+  if [ -f Gemfile ]
+  then
+    bundle exec cap $*
+  else
+    cap $*
+  fi
+}
--- a/plugins/capistrano/_capistrano
+++ b/plugins/capistrano/_capistrano
@@ -1,7 +1,7 @@
-#compdef xcap cap
+#compdef shipit
 #autoload
 
-# Added `xcap` because `cap` is a reserved word. `cap` completion doesn't work.
+# Added `shipit` because `cap` is a reserved word. `cap` completion doesn't work.
 # http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fcap-Module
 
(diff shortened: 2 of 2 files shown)


adisos:
how to use this?

phstc:
@adisos it wasn't merged into oh-my-zsh yet (@robbyrussell ???). So for now you could use it as a custom plugin, something similar to:

https://github.com/phstc/dotfiles/blob/master/zshrc#L39
https://github.com/phstc/dotfiles/tree/master/oh-my-zsh/custom/plugins/capistrano_custom

mcornella:
Oops, sorry for missing this. /cc @robbyrussell

plugins/capistrano/capistrano.plugin.zsh
```diff
@@ -0,0 +1,11 @@
+# Added `shipit` because `cap` is a reserved word. `cap` completion doesn't work.
```
apjanke: Technical niggle: `cap` is not a reserved word; it is a shell builtin, supplied by the `zsh/cap` module. Won't make a difference to most users, but they interact differently with shell customization code.

phstc: @apjanke thanks :beers: so is there a way to make the completion to work using `cap`?

apjanke: I think if you rename `_capistrano` to `_shipit` to fix the `shipit` completion, then a user can just do `alias cap=shipit` and the completion will work for them. However, that will break other uses of `cap` which are not themselves protected by doing `builtin cap` or `noalias` or `autoload -U` or something similar.

apjanke: So, I *think* that means:
* rename `_capistrano` to `_shipit`
* in `shipit()`, change `cap $*` to `command cap $*` (not `noalias cap $*`, because that could pick up the `zsh/cap` builtin if the module has been loaded)
* tell the user to `alias cap=shipit`, or just do it for them in `capistrano.plugin.zsh`

I'd have to try it out to be sure; it's tricky. Maybe people would be okay with using `capi` for capistrano, instead of `cap`, since `cap` is already taken by the shell? But then again, `cap` is already taken by capistrano, too. Not sure what the Right Thing to do here is.

PR #2500, opened 2014-02-06
Added a function, that allows Cygwin users to display their git branch in PROMPT, without taking forever to process.

mikesena:
I found the default git_prompt_info far too slow to be usable within Cygwin.

I'm not sure how to do a pull request without my two earlier commits, hope that is okay.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -7,6 +7,14 @@ function git_prompt_info() {
   fi
 }
 
+git_prompt_info_quick() {
+  ref=$(git symbolic-ref HEAD) 2> /dev/null
+  if [ "${ref}" != "" ]
+  then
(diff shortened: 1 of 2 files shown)


lib/git.zsh
```diff
@@ -7,6 +7,14 @@ function git_prompt_info() {
   fi
 }
 
+git_prompt_info_quick() {
+  ref=$(git symbolic-ref HEAD) 2> /dev/null
+  if [ "${ref}" != "" ]
```
apjanke: Should use zsh style `[[...]]` instead of `[...]` to match rest of file.

mikesena:
Pull request is being handled in two locations due to my original inexperience with GIT.

https://github.com/robbyrussell/oh-my-zsh/pull/3754 is now correct.

PR #2511, opened 2014-02-09
Makes history-substring-search use term specific up and down buttons.

### LINKED ISSUE #2089: history-substring-search should  use term specific up and down buttons:
Instead of using hard coded values for the up and down key, a terminal specific version should be used using zsh terminfo:

```
--- plugins/history-substring-search/history-substring-search.zsh   2013-09-11 14:22:38.853510227 +0200
+++ plugins/history-substring-search/history-substring-search.zsh   2013-09-11 14:23:22.528461501 +0200
@@ -163,8 +163,9 @@
 zle -N history-substring-search-up
 zle -N history-substring-search-down

-bindkey '\e[A' history-substring-search-up
-bindkey '\e[B' history-substring-search-down
+zmodload zsh/terminfo
+bindkey "$terminfo[kcuu1]" history-substring-search-up
+bindkey "$terminfo[kcud1]" history-substring-search-down

 #-----------------------------------------------------------------------------
 # implementation details
```

This will also fix problems like #1433.

FlorianFranzen:
Fixes #2089 and problems with some linux distributions and/or terminals.

### DIFF
--- a/plugins/history-substring-search/history-substring-search.zsh
+++ b/plugins/history-substring-search/history-substring-search.zsh
@@ -164,8 +164,12 @@ zle -N history-substring-search-up
 zle -N history-substring-search-down
 
 zmodload zsh/terminfo
-bindkey "$terminfo[kcuu1]" history-substring-search-up
-bindkey "$terminfo[kcud1]" history-substring-search-down
+if [[ -n "$terminfo[kcuu1]" ]]; then
+  bindkey "$terminfo[kcuu1]" history-substring-search-up
+fi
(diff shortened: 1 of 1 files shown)


apjanke:
@naegelejd: `vi-mode` appears broken with respect to the terminfo usage this PR introduced. #1533 introduced a logic error in `zle-line-init` and `zle-line-finish` that leaves things in the wrong `smkx`/`rmkx` mode so the terminfo bindings don't work. See discussion under #2735.

PR #2522, opened 2014-02-12
Preserve cursor position when prepending sudo.

fladi:
This changes the default behavior of the sudo plugin to preserve the position of the cursor when the current $BUFFER is prepended with "sudo ". The original behavior can be restored by setting `sudo_preserve_cursor=0`.


robbyrussell:
Closed via #3141

PR #2526, opened 2014-02-13
expand escaped characters in osx plugins

yosida95:
Help message for feature of controlling iTunes contains escaped tab(\t). But it is not unescaped when is printed.  This PR is for fix that.

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -177,10 +177,10 @@ function itunes() {
 		""|-h|--help)
 			echo "Usage: itunes <option>"
 			echo "option:"
-			echo "\tlaunch|play|pause|stop|rewind|resume|quit"
-			echo "\tmute|unmute\tcontrol volume set"
-			echo "\tnext|previous\tplay next or previous track"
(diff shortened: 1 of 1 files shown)


apjanke:
What case is this necessary in? `-e` is the default for `zsh`'s `echo`, and `itunes -help` is working fine for me as is.

screen shot 2015-07-09 at 5 00 42 pm

mcornella:
This happens if the BSD_ECHO option is enabled: http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html#index-BSD_005fECHO_002c-use-of. Anyhow, we could maybe standardise the use of `print` instead, which doesn't depend on that option:
```zsh
$ setopt bsdecho
$ echo "\tabcde"
\tabcde
$ print "\tabcde"
	abcde
```

apjanke:
Ah, didn't even know about that option.

Preferring `print` to `echo` sounds good. We could also use `emulate -L zsh` inside functions for broader protection against portability-affecting options. Maybe doing both is best.

(I've been trying to get in the habit of using `print`/`printf` instead of `echo` myself, due to `echo`'s portability issues across different shells and platforms.)

PR #2533, opened 2014-02-13
[plugins/tmux] Support automatic session grouping

AndrewKvalheim:
The current behavior with `ZSH_TMUX_AUTOCONNECT` enabled is to attach to an existing session. This change adds `ZSH_TMUX_GROUP`, which if enabled will group with an existing session instead.

This allows multiple terminals to share the same set of windows but display different active windows.


larson-carter:
@AndrewKvalheim Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

AndrewKvalheim:
I'm personally not using this anymore so it's probably better just reimplemented by anyone who needs it.

PR #2545, opened 2014-02-19
Allow FreeBSD to correctly detect number of CPUs on host.

dbye:
Use the same scheme as Darwin - `sysctl hw.ncpu' instead of nproc, which doesn't exist in FreeBSD


mcornella:
What's the value of `$OSTYPE` for FreeBSD?

mcornella:
Nevermind, [it's `freebsd`](https://github.com/zsh-users/zsh/search?q=ostype&type=Code).

PR #2557, opened 2014-02-23
Add function to search Dash.app

tonylotts:
Plugin for searching Dash.app

### DIFF
--- a/plugins/dash/dash.plugin.zsh
+++ b/plugins/dash/dash.plugin.zsh
@@ -1,0 +1,2 @@
+# Usage: dash [keyword:]query
+dash(){open dash://"$1"}
--- a/plugins/rake/rake.plugin.zsh
+++ b/plugins/rake/rake.plugin.zsh
@@ -1,3 +1,7 @@
+# Thank you Jim for everything you contributed to the Ruby and open source community 
+# over the years. We will miss you dearly.
+alias jimweirich="rake"  
+
 alias rake="noglob rake" # allows square brackts for rake task invocation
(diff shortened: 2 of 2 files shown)


tresni:
Been using this since 2014, still :+1: for me :)

plugins/dash/dash.plugin.zsh
```diff
@@ -0,0 +1,86 @@
...
+_dash() {
+  # No sense doing this for anything except the 2nd position and if we haven't
+  # specified which docset to query against
+  if [[ $CURRENT -eq 2 && ! "$words[2]" =~ ":" ]]; then
+    local -a _all_docsets
+    _all_docsets=()
+    # Use defaults to get the array of docsets from preferences
+    # Have to smash it into one big line so that each docset is an element of
+    # our DOCSETS array
+    DOCSETS=("${(@f)$(defaults read com.kapeli.dash docsets | tr -d '\n' | grep -oE '\{.*?\}')}")
```
wrecklass: From my command line this is wrong. Should be:

**com.kapeli._dashdoc_**

Typo or did the key change in Dash recently?

wrecklass:
Added comment to the code. Seems the defaults may have changed name:

com.kapeli.dashdoc

Kapeli:
`com.kapeli.dashdoc` is used by Dash 3 and Dash 4. `com.kapeli.dash` is used by Dash 2.

mcornella:
I've merged it (about time) even if there is a mismatch between versions. Open an issue or PR (preferably) about these changes. Not sure how to query for the version of an app in macOS. Also a README for the plugin would be nice.

PR #2589, opened 2014-03-09
Respect original PATH definitions on install

avit:
This is a solution for #1359.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -23,14 +23,23 @@ fi
 
 echo "\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m"
 cp $ZSH/templates/zshrc.zsh-template ~/.zshrc
-sed -i -e "/^ZSH=/ c\\
+sed -e "/^ZSH=/ c\\
 ZSH=$ZSH
-" ~/.zshrc
+" ~/.zshrc | tee ~/.zshrc
 
-echo "\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m"
-sed -i -e "/export PATH=/ c\\
-export PATH=\"$PATH\"
-" ~/.zshrc
+ORIG_RC_FILES=`ls ~/.bashrc ~/.zshrc.pre-oh-my-zsh 2>/dev/null` || :
+if [ "$ORIG_RC_FILES" ]
+then
+  ORIG_PATHS=`grep -sh '^export PATH=' $ORIG_RC_FILES` || :
+fi
+
+if [ "$ORIG_PATHS" ]
+then
+  echo "\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m"
+  nl=$'\n'
+  sed -e '/^export PATH=/ c\
+'"${ORIG_PATHS//$nl/\\$nl}" ~/.zshrc | tee ~/.zshrc
+fi
 
 if [ "$SHELL" != "$(which zsh)" ]; then
     echo "\033[0;34mTime to change your default shell to zsh!\033[0m"


apjanke:
Looks like this fixes #3612 and #3367 as well.

ncanceill:
@avit can we please get a rebase here ?

apjanke:
Hmm. I don't think this use of `tee` works. In the case where I have an existing `~/.zshrc` on OS X 10.9.5, this version of the installer spews a copy of the new `zshrc` to the terminal output, as pictured below. And in the case where I don't have a `.zshrc` already, I end up with an empty one after OMZ installation.

This is a variation on what @mitzip said. When you do a pipeline like `sed -e '...' file | tee file`, you have a race condition. The shell will start up the programs in both sections of the pipeline. `sed` will try to read the file, and `tee` will try to write it (opening the file for output even before receiving any input, and truncating it). In my second case, it looks like `tee` is winning the race and writing `~/.zshrc` before `sed` gets a chance to read it, so `sed` reads a 0-byte file.

This needs to be replaced with temp files.

screen shot 2015-06-08 at 7 56 34 am


```
eilonwy% rm .zshrc
rm: .zshrc: No such file or directory
eilonwy% cat install-avit.sh | sh
Cloning Oh My Zsh...
Cloning into '/Users/janke/.oh-my-zsh'...
remote: Counting objects: 12497, done.
remote: Total 12497 (delta 0), reused 0 (delta 0), pack-reused 12497
Receiving objects: 100% (12497/12497), 2.40 MiB | 4.37 MiB/s, done.
Resolving deltas: 100% (6851/6851), done.
Checking connectivity... done.
         __                                     __
  ____  / /_     ____ ___  __  __   ____  _____/ /_
 / __ \/ __ \   / __ `__ \/ / / /  /_  / / ___/ __ \
/ /_/ / / / /  / / / / / / /_/ /    / /_(__  ) / / /
\____/_/ /_/  /_/ /_/ /_/\__, /    /___/____/_/ /_/
                        /____/                       ....is now installed!


 Please look over the ~/.zshrc file to select plugins, themes, and options.


 p.s. Follow us at http://twitter.com/ohmyzsh.


 p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.
eilonwy% cat .zshrc
eilonwy%
```

apjanke:
I'm not sure this change makes the $PATH situation better than before.

On the one hand, it will preserve keeping `$PATH` in the path construction variable, so system-wide path setup like `/etc/paths.d` is respected. (Either because it chooses to not replace the original template contents, which have it, or because it will have been in the `export PATH=...` statements it found in the original rc files.) That's good.

But I think it's more likely to cause immediate breakage by overlooking path elements. In rc files, the path can be modified by several `PATH=` statements. The current "capture `$PATH`" behavior will grab the effects of all of them (at least how they were evaluated for this particular session). This parsing will only pick up ones that are using `export ...` and are flush left. The rc files may not use export on all of the PATH appending statements. Or even any—it's not needed because `$PATH` is already exported.

Consider this simplified example. Here are the config files before OMZ installation.

```
# ~/.zshrc
export JAVA_HOME=/opt/java/jdk-8u45
export PATH=$PATH:$JAVA_HOME/bin
PATH=/opt/foonly/bin:$PATH
# Use local stuff first
PATH=$HOME/bin:/usr/local/bin:$PATH
PATH="$PATH:$HOME/.gem/ruby/2.0.0/bin"
$path+=/opt/bar/bin

# ~/.bashrc
export PATH=$PATH:/opt/java/jdk-7u16/bin
PATH=/opt/foonly/bin:$PATH
# Use local stuff first
PATH=$HOME/bin:/usr/local/bin:$PATH
PATH="$PATH:$HOME/.gem/ruby/2.0.0/bin"
```

I end up with this in the `.zshrc` after OMZ installation.

```
# User configuration
 
export PATH=$PATH:/opt/java/jdk-7u16/bin
export PATH=$PATH:$JAVA_HOME/bin
```

* It grabbed `$JAVA_HOME/bin` for the path, but it didn't pick up the `$JAVA_HOME` initialization it's dependent on, so that entry devolves to `/bin` when you log in.
* Aside from that, it picked up the old Java version from this stale `.bashrc` first.
* It missed `/opt/foonly`, `$HOME/bin`, and the ruby gems
* It missed `/usr/local/bin`. Which will probably still end up in the path because the system includes it by default, but it won't be ahead of /bin or /usr/bin, so programs will run but the user might get the wrong version instead of a "not found" error message.
* It misses anything using the array style `$path` syntax
* It misses anything pulled in from other files that `.zshrc` would source, etc, etc

I appreciate the problem this is trying to solve, but I don't know that it's the right approach. It makes the path-determination-on-installation behavior more complicated than before, and also introduces a couple failure points that will cause immediate breakage upon installation for some users. This might be a step backward.

IMHO the way to get full correctness with path definitions is probably to just preserve the user's entire `.zshrc` contents, and add the OMZ loading code to it. $PATH construction is complicated, dynamic, and multi-staged, done with a full general-purpose programming language. Simple parsing is not going to be able to get it right.

And failing that, just slap an explicit dynamic `:$PATH` on the end of the `PATH` the installer creates, so at least it picks up additions to the systemwide path configuration.

levous:
I'm not sure if this is the best issue to add myself to but...

After installing OMZ, rum was complaining: 
`Warning: PATH set to RVM ruby but GEM_HOME and/or GEM_PATH not set, see:
    https://github.com/wayneeseguin/rvm/issues/3212
`

Here is the offending line from my ~/.zshrc
`
# User configuration
export PATH="/Users/rustyzarse/.rvm/gems/ruby-2.2.1/bin:/Users/rustyzarse/.rvm/gems/ruby-2.2.1@global/bin:/Users/rustyzarse/.rvm/rubies/ruby-2.2.1/bin:/Users/rustyzarse/.bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/rustyzarse/.rvm/bin"
`

to fix, I simply prepended with $PATH;  All Fixed :)
`
# User configuration
export PATH="$PATH;/Users/rustyzarse/.rvm/gems/ruby-2.2.1/bin:/Users/rustyzarse/.rvm/gems/ruby-2.2.1@global/bin:/Users/rustyzarse/.rvm/rubies/ruby-2.2.1/bin:/Users/rustyzarse/.bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/rustyzarse/.rvm/bin"
`

`

Issue #1337, opened 2012-10-01
Auto-complete fails on SSH?


benmezger:
Hey guys, usually when you write ssh user@+tab so you can see all the hosts, it works fine, but unfortunatly, this are the choices which gives me:

[11.11.11.11]:2222 <- Host + port. 

Why is it giving me like this? with a bracket and \? and :port as if it's telnet?

Any fix for this? 
Regards

Issue #1009, opened 2012-03-11
Use .ssh/config for hostname auto-complete 


EspadaV8:
I created a simple patch that should check a users SSH config file for any hosts and add those to the list of possible auto completion suggestions.

``` patch
@@ -35,10 +35,12 @@ cdpath=(.)
 [ -r /etc/ssh/ssh_known_hosts ] && _global_ssh_hosts=(${${${${(f)"$(</etc/ssh/ssh_known_hosts)"}:#[\|]*}%%\ *}%%,*}) || _ssh_hosts=()
 [ -r ~/.ssh/known_hosts ] && _ssh_hosts=(${${${${(f)"$(<$HOME/.ssh/known_hosts)"}:#[\|]*}%%\ *}%%,*}) || _ssh_hosts=()
 [ -r /etc/hosts ] && : ${(A)_etc_hosts:=${(s: :)${(ps:\t:)${${(f)~~"$(</etc/hosts)"}%%\#*}##[:blank:]#[^[:blank:]]#}}} || _etc_hosts=()
+[ -r ~/.ssh/config ] && _ssh_config_hosts=(${(s: :)${(ps:\t:)${(f)"$(<$HOME/.ssh/config|grep 'Host')"}#Host}#Hostname}) || _ssh_config_hosts=()
 hosts=(
   "$_global_ssh_hosts[@]"
   "$_ssh_hosts[@]"
   "$_etc_hosts[@]"
+  "$_ssh_config_hosts[@]"
   "$HOST"
   localhost
 )
```

PR #2602, opened 2014-03-13
Use zsh's default for ':completion:*:hosts'

blueyed:
The manual configuration of ':completion:*:hosts' causes some problems
(e.g. issue #1337), and misses useful information (especially from
~/.ssh/known_hosts; issue #690 and issue #1009).

The best option appears to be using zsh's default configuration for
':completion:*:hosts'.

While rebasing this for a new pull request (#1498 got closed by accident
and then requests to re-open it got ignored), it also drops the `users
off` setting: completion of user names can be useful, and the commit
adding it (c4434d2) does not state why.

/cc @essembeh: you had introduced `users off` - do you happen to remember why?


jadjay:
I don't understand I have to put the lines back to have the good behavior...

Regression ?

PR #2606, opened 2014-03-14
rkj-repos theme: Fixed breakage in new git repo without commits

pfault:
'git rev-parse HEAD' call without existing commits in repo will fail with error message, breaking the theme


mcornella:
This works (never tested it before :grin:). Please reword the second commit though, to remove my username: github (and bitbucket) has a bad habit of notifying commits with mentions whenever they enter a repository, whether it's a fork or not. 

/cc @robbyrussell

PR #2620, opened 2014-03-17
Fixed errors if acpitool isn't installed on linux

reedriley:
Currently, when acpitool isn't installed, I get an error every time my prompt tries to call battery_pct:

```
battery_pct:1: command not found: acpi
```

This change fixes the error by guarding the call to acpi with a check that it's actually installed and executable.

(Also include a small amount of whitespace cleanup.)


Omar-Elrefaei:
WHAT, I spent > 15 min because it didn't say `Error: acpi not installed`

PR #2626, opened 2014-03-18
Save current shell when installing

jcharaoui:
- Don't change login shell to zsh if already zsh
- Restore previous shell instead of bash when uninstalling
- Fix bashism in tools/uninstall.sh


apjanke:
I have OS X (10.9) That `(dscl ...)` code works for me.

apjanke:
Should this also `rm ~/.pre-oh-my-zsh-shell` as part of the cleanup?

Issue #2608, opened 2014-03-14
key bindings for PageUp, PageDown, Home, End result in "cannot bind to an empty key sequence" on Mac OS


danchenkov:
The latest commit (7adf66f942cc143aea6ced780b249f0e989e5461) introduced these four errors on Mac OS X:
/home/username/.oh-my-zsh/lib/key-bindings.zsh:bindkey:23: cannot bind to an empty key sequence
/home/username/.oh-my-zsh/lib/key-bindings.zsh:bindkey:24: cannot bind to an empty key sequence
/home/username/.oh-my-zsh/lib/key-bindings.zsh:bindkey:30: cannot bind to an empty key sequence
/home/username/.oh-my-zsh/lib/key-bindings.zsh:bindkey:40: cannot bind to an empty key sequence

PR #2634, opened 2014-03-21
added terminfo keys verification

danchenkov:
As discussed https://github.com/robbyrussell/oh-my-zsh/issues/2608. The terminfo keys verification seems acceptable workaround to me.

### DIFF
--- a/lib/key-bindings.zsh
+++ b/lib/key-bindings.zsh
@@ -53,6 +53,10 @@ fi
 bindkey '^?' backward-delete-char                     # [Backspace] - delete backward
 if [[ "${terminfo[kdch1]}" != "" ]]; then
   bindkey "${terminfo[kdch1]}" delete-char            # [Delete] - delete forward
+else
+  bindkey "^[[3~" delete-char
+  bindkey "^[3;5~" delete-char
+  bindkey "\e[3~" delete-char
 fi
 
 # Edit the current command line in $EDITOR


apjanke:
FYI, I'm playing around with a [function-based version of this](https://github.com/apjanke/oh-my-zsh/commit/9bf0245b3bb580f217e2ab8b6babf6aee99bdbf1) on my [maybe-bindkey branch](https://github.com/apjanke/oh-my-zsh/tree/maybe-bindkey). Adds an `omz_bindkey` function that gets you back to concise, single-line calls for the conditional terminfo-based bindings.

```
omz_bindkey   kpp    up-line-or-history           # [PageUp] - Up a line of history
omz_bindkey   knp    down-line-or-history         # [PageDown] - Down a line of history
omz_bindkey   kcuu1  up-line-or-search            # start typing + [Up-Arrow] - fuzzy find history forward
omz_bindkey   kcud1  down-line-or-search          # start typing + [Down-Arrow] - fuzzy find history backward
```

I think this is more readable, gives users a way to easily do similar terminfo-based bindings in their own `.zshrc` or plugins, and gives us a place for a hook if we wanted to add debug logging (e.g tracing all the key bindings made during startup sequence so a user can track down why a particular key is acting in an unexpected way).

This might also let us extend the terminfo style approach the the oddball key sequences like Ctrl-Shift-Arrow and keypad-local sequences that are not included in the normal terminfo definitions. It would be nice to be able to say something like `omz_bindkey ctrl-kcub1 backward-word`.

PR #2653, opened 2014-03-24
Customization tweaks

### LINKED ISSUE #2295: Document $ZSH_CUSTOM

LFDM:
This goes hand in hand with the recent addition of the [Customization wiki](https://github.com/LFDM/oh-my-zsh/wiki/Customization)
- Updates the README with a link to the wiki
- Refactors the customization examples in the `custom` folder

TODO:
@ncanceill, there's also a change included that goes a bit against your PR #2645

I removed ZSH_CUSTOM from `oh-my-zsh.sh` and set it directly in the `zshrc.template`, including a comment that refers to the wiki.
We need to decide which approach we like better and add another commit to this PR before it can be merged. 

Closes #2295

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -71,9 +71,8 @@ the "refcard":http://www.bash2zsh.com/zsh_refcard/refcard.pdf is pretty tasty fo
 
 h3. Customization
 
-If you want to override any of the default behavior, just add a new file (ending in @.zsh@) into the @custom/@ directory.
-If you have many functions which go well together you can put them as a *.plugin.zsh file in the @custom/plugins/@ directory and then enable this plugin.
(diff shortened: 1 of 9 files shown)


robbyrussell:
@mcornella is this something you'd still like to merge in?

robbyrussell:
Am going to close this as we have ZSH_CUSTOM in the template, README references, and there are details in the wiki. 

https://github.com/ohmyzsh/ohmyzsh/wiki/Customization

PR #2672, opened 2014-03-30
Adding support for Coda2

drugoi:
I think, that zsh is needed support Coda2 editor under MacOS.


mcornella:
You can use this [simple alias](https://stackoverflow.com/a/28821586): `alias coda='open -a "Coda 2"'`

PR #2697, opened 2014-04-07
Fix: safe-paste.plugin.zsh does not work if $TERM == rxvt-unicode-256color 

zlobz:
Fixed bug: safe-paste.plugin.zsh doesn't work when $TERM == rxvt-unicode-256color


apjanke:
There are a lot of other variants of `xterm-` and `rxvt-unicode-` entries, too. Instead of checking for a small list of exact entries, maybe you should test against patterns like `xterm*`, `screen*`, and `rxvt-unicode*`.

Could you also change the title of this issue to be more specific about what it's doing? "Fixed bug in X" is pretty vague, and not very helpful when reading through issue lists.

PR #2699, opened 2014-04-08
Break out pull --rebase into fetch; rebase --autostash

rsrchboy:
...this should make auto-updating a little less of a perpetual fail for those 
of us with a tendency to tinker.

Not that anyone who uses zsh and oh-my-zsh would have such a tendency :)


mcornella:
Won't merge. Users should be aware of the changes they've made in their OMZ directories and act accordingly.

PR #2702, opened 2014-04-09
ssh-config hosts sections for autocompletion

elias5000:
Autocompletion plugin for people who like to make extensive use of ssh-config.
Autocompletion for clouds tool (https://github.com/cristim/clouds).

### DIFF
--- a/plugins/clouds/_clouds
+++ b/plugins/clouds/_clouds
@@ -1,0 +1,58 @@
+#compdef clouds
+#autoload
+
+# clouds zsh completion
+
+_clouds_all_stacks() {
+  all_stacks=(`clouds list | tail -n +3 | awk '{print $1}'`)
+}
+
+_clouds_local_stacks() {
+  local_stacks=(`test -d stacks && ls -1 stacks`)
+}
+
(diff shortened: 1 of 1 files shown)


plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
+############################################################
+# Take all host sections in .ssh/config and offer them for
```
apjanke: @ncanceill: I think you want the [`_ssh` completion function](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_ssh), and specifically its [`_ssh_hosts` helper function](https://github.com/zsh-users/zsh/blob/a62e96822489d7af35e1463a4de1e82ad823ba84/Completion/Unix/Command/_ssh#L612-L646) which is where it reads the ssh-config file.

I'm also running zsh 5.0.2 on Mavericks, and entries from ssh-config host section names *are* offered for completion... in some cases. Maybe we should look a bit closely at the default and target behavior before merging this in.

Switching over to #2707 for further comments.

PR #2706, opened 2014-04-09
allow . in profile names

elias5000:
Allow "." in profile names. This zsh function was originally created by me and added by a workmate I shared it with. Unfortunately he added it without my later addition to allow dots in profile names which I use broadly.


elias5000:
Cleanup orphaned PR.

mcornella:
Pulled in e8aaab5.

PR #2707, opened 2014-04-09
autocomplete hosts from hosts sections in ssh-config

elias5000:
Zsh does autocompletion for /etc/hosts and hosts from the known_hosts but not for ssh-config profiles.
This plugin adds these to the hosts completion list. It also provides a feature to remove SSH keys from the known hosts file by profile name.

### DIFF
--- a/plugins/ssh/ssh.plugin.zsh
+++ b/plugins/ssh/ssh.plugin.zsh
@@ -4,7 +4,7 @@
 # Filter out wildcard host sections.
 local hosts
 if [[ -f $HOME/.ssh/config ]]; then
-  hosts=($(cat $HOME/.ssh/config | egrep '^Host.*' | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//' | grep '\.'))
+  hosts=($(egrep '^Host.*' $HOME/.ssh/config | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//'))
   zstyle ':completion:*:hosts' hosts $hosts
 fi
 
--- a/plugins/ssh/ssh.plugin.zsh
+++ b/plugins/ssh/ssh.plugin.zsh
@@ -1,0 +1,39 @@
+############################################################
+# Take all host sections in .ssh/config and offer them for
+# completion as hosts (e.g. for ssh, rsync, scp and the like)
+# Filter out wildcard host sections.
+local hosts
+if [[ -f $HOME/.ssh/config ]]; then
+  hosts=($(cat $HOME/.ssh/config | egrep '^Host.*' | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//' | grep '\.'))
+  zstyle ':completion:*:hosts' hosts $hosts
+fi
+
+############################################################
+# Remove host key from known hosts based on a host section
+# name from .ssh/config
+function ssh_rmhkey {
+  if [[ "x$1" == "x" ]]; then return; fi
+  ssh-keygen -R $(grep -A10 $1 ~/.ssh/config | grep -i HostName | head -n 1 | awk '{print $2}')
+}
+compctl -k hosts ssh_rmhkey
+
+############################################################
+# Load SSH key into agent
+function ssh_load_key() {
+  local key=$1
+  if [[ "$key" == "" ]]; then return; fi
+  if ( ! ssh-add -l | grep -q $key ); then
+    ssh-add ~/.ssh/$key;
+  fi
+}
+
+############################################################
+# Remove SSH key from agent
+function ssh_unload_key {
+  local key=$1
+  if [[ "$key" == "" ]]; then return; fi
+  if ( ssh-add -l|grep -q $key ); then
+    local keyfile=$(ssh-add -l | grep $key | cut -d ' ' -f 3)
+    ssh-add -d $keyfile
+  fi
+}


ulope:
The shell expression used to extract the host name aliases is in my opinion needlessly complicated and too restrictive. In it's current form it excludes aliases that don't contain a dot (which in my case are all of them).

I'd suggest to use this expression instead:

```shell
hosts=$(egrep -i '^host\s+[^*]+$' $HOME/.ssh/config | cut -d ' ' -f 2)
```

elias5000:
Yours completely ignores host entries containing trailing wildcards. My overly complicated shell expression aims at also providing completion for those without giving you the asterisk.

My suggestion to cover your case (no dot), too:

    hosts=$(egrep '^Host.*' $HOME/.ssh/config | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//')

ncanceill:
Hi guys,

Please try to keep discussions on a cordial tone — we are all trying to improve on the same software here.

Concerning completion, I remember we discussed it in #2702, and it only seems necessary on certain platforms. Instead of including it in the plugin regardless, you should investigate which platforms are affected, and have the plugin detect those platforms. That way completion from your plugin would only kick in when Zsh completion does not include hosts from `ssh_config`.

elias5000:
I tried to be ironic. Sorry if that wasn't too obvious.

ulope:
Yeah, sorry I could have worded my initial comment less aggravating as well.

What I was trying to get at is that since shell startup time with oh-my-zsh installed is already a concern it seems a bit excessive to spawn a 5 process pipe to do basic string processing.

This should produce identical results and is nearly twice as fast and uses way less CPU time:

```shell
hosts=$(perl -ne 's/^host\s+(.*?)\.?\*?$/\1/i and print $1."\n"' $HOME/.ssh/config)
```

elias5000:
Which OS do you use? On my OSX 10.10.3 I get 0.002s total for my expression vs. 0.006s total for the perl command with perl being a bit more friendly with the CPU.

Perl command that would find my acceptance (no leading asterisks):

    hosts=($(perl -ne 's/^host\s+([^*]*?)\.?\*?$/\1/i and print $1."\n"' $HOME/.ssh/config))

ulope:
I'm also on OS X 10.10.3.

I used 100 iterations of each version to get more reliable results:

```
pipe: 0.38s user 0.47s system 258% CPU
perl: 0.29s user 0.26s system 83% CPU
```

apjanke:
Hi folks.

There's been a bit of thrashing in the area of SSH completion for OMZ. Check out how many issues & PRs we've had: apjanke/oh-my-zsh#4 (tracking issue). Half of those look like they're related to this topic of where host names are drawn from, and OMZ has been going back and forth on how to handle it. Maybe we should take a closer look at both Zsh's own behavior and the target behavior and get this straightened out more definitively.

Zsh's own [`_ssh`](https://github.com/zsh-users/zsh/tree/master/Completion) completion function does read `~/.ssh/config` (in [`_ssh_hosts`](https://github.com/zsh-users/zsh/blob/a62e96822489d7af35e1463a4de1e82ad823ba84/Completion/Unix/Command/_ssh#L612-L646) and use `Host` entry names from it... but only in some cases. I can't figure out exactly when. But for me:

* If I just type "`ssh `" and Tab, ssh-config entries will not be included.
* If I type "`ssh x`" (for any x), then ssh-config entries may be included, if there are any that match that prefix. But in some cases, if there are also matching entries in `/etc/hosts` or `~/.ssh/known_hosts`, those will be used in preference, and the ssh-config entries will not be included. I'm not sure exactly what controls it; using a longer `x` tends to make the ssh-config entries show up (even if it's not distinct enough to rule out all entries from the other sources).

So your behavior might vary based on what's in your `/etc/hosts` and `~/.ssh/known_hosts` files. That might be what made it look like there's a behavior difference between OS X and Ubuntu Linux in #2702.

I think the issue is that if the `_combination ... users-hosts` in `_ssh_hosts` "matches", then it skips reading the ssh-config file. (As described in #2737. I confirmed with `set -x` tracing.) But I don't know what that `users-hosts` thing really means.

apjanke:
A couple items on this implementation:

* This looks like it changes to complete hosts *only* from `~/.ssh/config`, instead of using them in addition to `/etc/hosts` and `known_hosts`. Could it combine them instead to show the union?
* The `ssh -F` option can point it to a different config file. Could that be respected?
* The broad `zstyle ':completion:*:hosts'` pattern makes this affect hosts completion for non-ssh programs, too, such as `telnet`, `lynx`, and `wget`.
* There's a global `/etc/sshd_config` file that entries could be drawn from too.
* This reads the config file once at OMZ startup, so it's going to miss any changes to it during a session.

Here's an alternate single-`grep` implementation you could use too. Might be lighter than `perl`, or be useful on systems which don't have Perl installed. And it will grab both user and global ssh_config files, including old ones with "2" suffixes, and gracefully handle their absence.

```
# Lists hosts defined in ssh config files
function _omz_list_ssh_config_hosts() {
  echo ${(F)${(f)"$(grep -E -i -h "^\s*Host [^\*]" {/etc/ssh_,~/.ssh/}config(|2)(N) /dev/null)"}##Host }
}
```

Zsh's `_ssh_hosts()` has [a pure-zsh implementation, too](https://github.com/zsh-users/zsh/blob/a62e96822489d7af35e1463a4de1e82ad823ba84/Completion/Unix/Command/_ssh#L624-L640). That might be fastest of all, especially on Cygwin, where shellouts are more expensive.

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
+############################################################
+# Take all host sections in .ssh/config and offer them for
+# completion as hosts (e.g. for ssh, rsync, scp and the like)
+# Filter out wildcard host sections.
+local hosts
```
apjanke: `local` doesn't work at the top level of plugin or theme files (currently). It needs to be wrapped in an anonymous function to provide a scope for it.

apjanke:
FYI, the `common-aliases` plugin will conflict with this because it also defines zsh host completion style; see #4257. Since it uses a narrower pattern in its `zstyle` definition, I think it will always take precedence over this, regardless of load order. So you'd need to make sure not to use them together.

Gandem:
Hi folks,

May I be of any help for this PR ? We were planning to implement auto-completion for ssh hosts based on ssh-config at work, but it would be awesome to have it in oh-my-zsh !

nickcmaynard:
I'm suffering from this as well - stuff in `.ssh/config` is ignored completely if there's a matching host in known_hosts - which there will almost always will be.  I would suggest that if there are manually configured entries in `.ssh/config` they should always take priority, as I understand (not sure?) this PR implements.

This appears to have been open for *3.5 years*, and the "tracking issue" #4345 for *2 years*.  I respectfully submit that it would be appropriate for somebody (@apjanke?) to grasp this particular nettle and adjust behaviour to be less surprising.

Alexandero89:
Is there any reason this PR is still not merged?

I really dont want to be rude,but there is a good solution for a problem that only has to be approved.
Why is this laying around for almost 5,5 years?

apjanke:
@virajkulkarni14 Hi! Sorry, I'm not an admin for ohmyzsh; I don't have commit rights. I just hang out here a lot. :)

virajkulkarni14:
@apjanke, I'm so sorry. I just assumed you were the repo owner or one of the admins since you gave some really useful pointers above.

I dug around and found @ModdedGamers, @robbyrussell & @mcornella to be relevant. Folks, if you could please approve & merge this PR, it would be really helpful! Thank you so much! 🙏

carlosala:
Let's try to unblock this! Sorry to everyone for being so late reviewing this.
@elias5000 could you add a README.md file explaining how to enable the plugin, what does it add, etc? You can refer to any other plugin, as an example [this](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/rbw).
Thanks, and sorry!

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+if [[ -f $HOME/.ssh/config ]]; then
```
carlosala, line 6: Just to be REALLY picky, could you quote the expression (`"$HOME/.ssh/config"`)

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+function ssh_rmhkey {
+  if [[ "x$1" == "x" ]]; then return; fi
+  ssh-keygen -R $(grep -A10 $1 ~/.ssh/config | grep -i HostName | head -n 1 | awk '{print $2}')
+}
+compctl -k hosts ssh_rmhkey
+
+############################################################
+# Load SSH key into agent
+function ssh_load_key() {
+  local key=$1
```
carlosala, line 23: Same here, let's quote `"$1"`

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+  if [[ "$key" == "" ]]; then return; fi
```
carlosala, line 24: Here we could use `[[ -z "$key" ]]` it's maybe more standard, but it's also ok to leave this like that.

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+  if ( ! ssh-add -l | grep -q $key ); then
```
carlosala, line 25: Same quoting stuff

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+    ssh-add ~/.ssh/$key;
```
carlosala, line 26: and quote the whole `"~/.ssh/$key"`

plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+  fi
+}
+
+############################################################
+# Remove SSH key from agent
+function ssh_unload_key {
+  local key=$1
```
carlosala, line 33: same quoting stuff as before in the whole function

carlosala (changes requested):

plugins/ssh/ssh.plugin.zsh
mcornella: This won't work, as quoting `~` makes it not expand to `$HOME`.

mcornella:
@carlosala I think for a PR of this age it's better to redo it, as the OP is probably no longer interested in it. Can you redo the commits in a separate PR (keeping authoring) and let's discuss it there?

plugins/ssh/ssh.plugin.zsh
carlosala: Wow, that's true. I'd expect to expand in double quotes, and do not in single quotes, which is the rationale of not expanding in any quotes?

mcornella: Tilde expansion is a separate kind that works different than parameter expansion, the rule about double and single quotes only applies to dollar ($) strings. You can look up `tilde expansion` on the zsh docs.

ferdnyc:
@mcornella:
> @carlosala I think for a PR of this age it's better to redo it, as the OP is probably no longer interested in it. Can you redo the commits in a separate PR (keeping authoring) and let's discuss it there?

Was this ever done? I see a thumbs-up from @carlosala on that comment, but no mention of or reference to a replacement PR.

plugins/ssh/ssh.plugin.zsh
ferdnyc: Personally I prefer the `-z` form and consider it more standard by far. If using this form, I'd expect to see it in old-school Bourne style, as

```bash
if [[ x"$key" == x"" ]]
```

...to avoid testing any genuinely empty strings.

plugins/ssh/ssh.plugin.zsh
ferdnyc: The bigger problem here is, this won't _work_. If `$key` is a private key filename under `$HOME/.ssh/`, its path won't be visible in the output of `ssh-add -l` if that fingerprint has already been unlocked. The output consists of lines that look like this:

```text
2048 SHA256:fingerprintfingerprint comment (RSA)
```

(Where "comment" is frequently "user@remotehost".)

So, this is going to be doing a lot of unnecessary scanning of `ssh-add -l` output, looking for filenames that aren't displayed.

I'd say it would be better to just remove the conditional, save one invocation each of `ssh-add` and `grep`, and just always call `ssh-add` unconditionally, except that `ssh-add /path/to/private_key` for a private key file that requires a passphrase will **always** prompt for that passphrase, even if the key is already added and unlocked.

But the right way to get the fingerprint for a private key is to run `ssh-keygen -l -f /path/to/private_key`. The output of **that** command is the same fingerprint line that will be listed in `ssh-add -l` if the key is added and unlocked.

carlosala:
@ferdnyc I never did that. Do you mind redoing the PR? We can review it!

PR #2730, opened 2014-04-18
Common Perl functions and prompt helpers.

rsrchboy:
...at the moment, just `perl_prompt_info()`, modeled after the other
similar functions for other languages.


Marcholio:
This could be either merged or closed

PR #2736, opened 2014-04-22
gitfast: update to upstream v1.9.2

felipec:
Lots of changes from upstream.


mcornella:
Should solve most problems with gitfast plugin...
@felipec are there any more updates available (it's been a long time since you posted this, sorry for not seeing this pull request...)

/cc @robbyrussell

felipec:
I'll check if there are more.

PR #2738, opened 2014-04-22
Fix dirhistory when the directory contains spaces

UnrealQuester:
pwd output is not escaped, so when changing to a directory like "space dir", only "space" gets saved in the history. This fix wraps `pwd` in double quotes to mitigate that.


mcornella:
Did you close by mistake?

UnrealQuester:
Sorry for not stating the reason. The fix is no longer relevant as the current version of the plugin uses $PDW and thus works with spaces.

mcornella:
Oh right I wasn't aware of that. Thanks!

PR #2739, opened 2014-04-23
Added setting of the window title under Cygwin

ferenczy:
Common method for setting window title using escape sequences works under Cygwin on Windows too (for example in mintty, ConEmu and even in plain cmd.exe). But Cygwin sets TERM environment variable to "cygwin" by default. Therefore I have added "cygwin" to the TERM testing condition. Tested in ConEmu, mintty and cmd.exe.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -2,13 +2,14 @@
 #http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#ss3.1
 #Fully support screen, iterm, and probably most modern xterm and rxvt
 #Limited support for Apple Terminal (Terminal can't set window or tab separately)
+#Should work with Cygwin in e.g. mintty, ConEmu or plain cmd.exe
 function title {
   if [[ "$DISABLE_AUTO_TITLE" == "true" ]] || [[ "$EMACS" == *term* ]]; then
     return
   fi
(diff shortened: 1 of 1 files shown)


apjanke:
I tested on Cygwin under Windows 7. It works. It's largely unneeded with current Cygwin versions, because MinTTY declares itself as `xterm` and that's the main "prompt" option installed for users now. But this makes it work in cmd.exe/regular Windows console windows too (what's known as the "plain" Cygwin prompt), and it's good to have for completeness.

:+1:

ferenczy:
@apjanke Not only regular Windows console, but mainly ConEmu. I think that ConEmu is used much more than MinTTY, which even doesn't support multiple tabs. I have ConEmu running constantly with a lot of opened tabs (the most of them are Cygwin, sometimes Git-Bash, PowerShell, or even plain cmd), everything inside one window. Cygwin is designed to work with `TERM` environment variable set to "cygwin", so would be great, if oh-my-zsh would support it.

apjanke:
Oh, cool. I didn't know about ConEmu. Checked it out today. This looks a *lot* nicer than MinTTY. Thanks @ferenczy !

Tested on ConEmu on Windows 7 with Cygwin zsh. Works there too.

This looks good to merge after a rebase.

apjanke:
Ping... want to rebase this so we can merge?

PR #2751, opened 2014-04-25
add check for colored ls on openbsd. (retry of #2225)

drm00:
Retry of PR#2225, with help from ncanceill.  Will close the old PR now.


mcornella:
Sorry for the long wait! :sweat_smile: 

/cc @robbyrussell

PR #2782, opened 2014-05-04
Fixed gallois to support virtualenv.

guitarpoet:
Gallois them works perfect for me, but it didn't work with [virtualenv](http://www.virtualenv.org) by default. So I patched it to support virtualenv.


larson-carter:
@guitarpoet Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

PR #2788, opened 2014-05-06
add cygwin support to jira.plugin

gschrader:


ncanceill:
Well, #2922 has been merged already.

Anyway, the `-o` option is not available in all distributions of `uname`, so we need a proper way to detect Cygwin first.

apjanke:
There's an `open_command()` in `lib/functions.zsh` that already supports cygwin. You can just switch to using that, and remove the OS detection logic.

```
  if [ -z "$1" ]; then
    echo "Opening new issue"
    open_command "$jira_url/secure/CreateIssue!default.jspa"
  else
    echo "Opening issue #$1"
    if [[ "x$JIRA_RAPID_BOARD" = "xtrue" ]]; then
      open_command  "$jira_url/issues/$1"
    else
      open_command  "$jira_url/browse/$1"
    fi
  fi
```

apjanke:
Looks like that `open_command` stuff was already done in #4217, and this PR can be closed.

mcornella:
Oh shoot, I didn't remember about this one. I could have autoclosed it... Please @gschrader close this PR.

PR #2805, opened 2014-05-13
Add iTerm2 tab color plugin

e1senh0rn:
Whenever you do SSH, iTerm2 tab changes color to custom one. It greatly helps figuring out what environment you are in.

Idea and code by Wade Simmons.
https://github.com/wadey/dotfiles/blob/master/zsh/iterm2.zsh

### DIFF
--- a/plugins/iterm2-tab-color/iterm2-tab-color.plugin.zsh
+++ b/plugins/iterm2-tab-color/iterm2-tab-color.plugin.zsh
@@ -1,0 +1,43 @@
+# Written by Wade Simmons
+# https://github.com/wadey
+#
+# iTerm2 tab color commands
+# http://code.google.com/p/iterm2/wiki/ProprietaryEscapeCodes
+
+if [[ -n "$ITERM_SESSION_ID" ]]; then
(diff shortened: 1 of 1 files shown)


apjanke:
Could this be combined with #3790 into a common `iterm2` plugin, instead of splitting each function in to a separate plugin?

plugins/iterm2-tab-color/iterm2-tab-color.plugin.zsh
```diff
@@ -0,0 +1,43 @@
+# Written by Wade Simmons
+# https://github.com/wadey
+#
+# iTerm2 tab color commands
+# http://code.google.com/p/iterm2/wiki/ProprietaryEscapeCodes
```
apjanke: This URL is out of date. The current one is https://iterm2.com/documentation-escape-codes.html.

avivrosenberg:
Hi,
I've added functionality to change tab color in PR #3790.

robbyrussell:
Closing as this was handled in #3790

PR #2807, opened 2014-05-15
Mercurial: faster prompt, auto-workon

grimborg:
I've added auto-workon for virtualenvs for mercurial repositories like it works on git.

Fixed speed problem of the agnoster theme when inside mercurial repositories, which was slow to the point of being unusable: it was calling hg many times, now it calls hg only once.


grimborg:
Closing so that I don't see it anymore on my open PRs.

PR #2820, opened 2014-05-20
Steeef.theme: Simplifying the git check for untracked files

newmaniese:
Right now on large projects, ls-files takes a long time to loop through the entire tree. Git status shows all the information that we are looking for in the command prompt and is much quicker.


mcornella:
I can't reproduce your perfomance results. It looks in my tests that the current command is 2x faster on a big test suite (100000+ untracked files).

PR #2824, opened 2014-05-22
Change reference of `Bundle` to `Plugin`

andriykohut:
Change reference of `Bundle` to `Plugin` + new repo name


belak:
It looks like these changes made it into the repo in 7f0b577b142fad91ad04f03112498304c9f27dbb

Issue #1115, opened 2012-05-12
Automatically escape HEAD^ in git commands


aantn:
Any change this can be included?

https://github.com/knu/zsh-git-escape-magic

PR #2847, opened 2014-06-01
Plugin to magically escape zsh meta-characters in git commands

pho3nixf1re:
The plugin simply autoloads git-escape-magic which is included here and
copied from https://github.com/knu/zsh-git-escape-magic .

This addresses #1115

### DIFF
--- a/plugins/git-escape-magic/README.md
+++ b/plugins/git-escape-magic/README.md
@@ -1,0 +1,17 @@
+# Git Escape Magic
+
+This plugin is copied from the original at
+https://github.com/knu/zsh-git-escape-magic. All credit for the
+functionality enabled by this plugin should go to @knu.
+
+An excerpt from that project's readme explains it's purpose.
+
+> It eliminates the need for manually escaping those meta-characters. The zle function it provides is context aware and recognizes the characteristics of each subcommand of git. Every time you type one of these meta-characters on a git command line, it automatically escapes the meta-character with a backslash as necessary and as appropriate.
+
+## Useage
+
+To use this plugin add it to your list of plugins in your `.zshrc` file.
+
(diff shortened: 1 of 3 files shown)


Konfekt:
Looks convenient. How about taking aliases of `git` into account, like `g` ? Could this not serve as a general escape-magic plugin ? For example `zcalc` also would be handier if `*` is auto escaped.

Konfekt:
Never mind. Just found out at

http://stackoverflow.com/questions/11456403/stop-shell-wildcard-character-expansion/22945024#22945024

about `alias g='noglob git'`

Konfekt:
Any comments on why this has not been merged? This is a convenient plugin for users of zsh and git  that otherwise needs manual fiddling to install.

mcornella:
I thought there wasn't any more interest in this plugin.

pho3nixf1re:
I am still interested in this. I have it in the custom plugins folder currently and have used it since this was opened.

mcornella:
Ok, I'll gladly merge this. Can you add a README with instructions on how to use it (for instance, that it doesn't work with git aliases) and update to the latest version of the git-escape-magic file?

pho3nixf1re:
Sure, I'll get to be by the end of the week. Thanks!

pho3nixf1re:
I'm cleaning house on my old PR's. I just updated this with a README and updated the script. I have also verified that this plugin still works. If there are no other concerns it should be ready for merge now. Thanks for the insane patience on this!

pho3nixf1re:
Just pinging on this, I'm still using it myself and I'm sure others would like to have access to it. Anything left to get it merged?

mcornella:
Congratulations on being probably the oldest PR merged (sorry)

PR #2866, opened 2014-06-06
Fixed wd plugin bug on zsh <= 4.3.10

hamiltont:
See http://www.zsh.org/mla/workers/2014/msg00501.html for details.
At time of writing, some RedHat and CentOS still have 4.3.10 as
the most up-to-date zsh in their package managers, so this should
allow wd to work on those systems properly.

Not sure if you only want to pull in full versions and not just minor 
patches. If that's the case just close this out please :-)

### DIFF
--- a/plugins/wd/wd.sh
+++ b/plugins/wd/wd.sh
@@ -64,9 +64,9 @@ wd_warp()
 
 wd_add()
 {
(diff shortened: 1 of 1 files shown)


mcornella:
Ping @hamiltont, we merged #3108. Please close this one

hamiltont:
Thanks!

PR #2875, opened 2014-06-11
Removes the support for screen TERMs in termsupport.zsh

### LINKED ISSUE #2015: Screen plugin does not work because duplicate hook already defined in lib/termsupport.zsh
`lib/termsupport.zsh` already defined `precmd` and `preexec` hooks to updated window title and terminal title, declaring `plugins=(screen)` does not provide the expected behavior.

Unloading them could be a wrokaround.

```
if [[ "$TERM" == screen* ]]; then
  autoload -U add-zsh-hook
  add-zsh-hook -d precmd  omz_termsupport_precmd
  add-zsh-hook -d preexec omz_termsupport_preexec
  ...
```

court-jus:
Without this fix, it is not possible to have fixed window titles in
screen, zsh always makes automatic titles.

This fixes #2015 and #1491

Users should now use the screen plugin to get an automatic windowtitle.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -6,9 +6,7 @@ function title {
   if [[ "$DISABLE_AUTO_TITLE" == "true" ]] || [[ "$EMACS" == *term* ]]; then
     return
   fi
-  if [[ "$TERM" == screen* ]]; then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Closing this one out as I don't believe it's still an issue. Please let us know if I'm wrong

PR #2884, opened 2014-06-16
Validate if its a git repository or not before triggering git commands.

### LINKED ISSUE #2883: fatal: Invalid gitfile format: .git
After upgrading to the latest version, I see the error `fatal: Invalid gitfile format: .git` whenever I get to a non-git repository directory.

screenshot

rhoml:
closes #2883


webknjaz:
In fact, this check doesn't take into account that user may have a `--separate-git-dir`. And in this case it would be a file containing a reference, like: `gitdir: /path/to/real/repo`

https://git-scm.com/docs/git-clone#git-clone---separate-git-dirltgitdirgt

PR #2888, opened 2014-06-16
vi-mode: Allow an insert mode indicator to be set.

ajbw:
Set INS_MODE_INDICATOR to set what will be shown in the prompt during
insert mode.


alx741:
+1

PR #2889, opened 2014-06-19
add percol completion plug-in

dweinstein:
- Source: https://github.com/mooz/percol/blob/master/tools/zsh/_percol


ncanceill:
Wait — this comes from the official source, let us trust them.

Could you add a short readme to indicate that it adds completion and give a link to the source?

PR #2903, opened 2014-06-23
Maven plugin - dynamic profiles

kubamarchwicki:
Lists all available profiles for maven: from current pom.xml, hierarchy of parent pom.xml (until no pom is available - external parent dependency) and the root .m2/settings.xml

This is similar stuff I've pushed to juven bash completion https://github.com/juven/maven-bash-completion/pull/44 but as I've changed the shell... ;)

As I'm not a bash / zsh master I'm keen on feedback and review :)
Thx

### DIFF
--- a/plugins/mvn/mvn.plugin.zsh
+++ b/plugins/mvn/mvn.plugin.zsh
@@ -236,6 +236,9 @@ function listMavenCompletions {
         -P $profiles
     )
 
+    unset POM_HIERARCHY
+    unset profiles
+
 }
 
 compctl -K listMavenCompletions mvn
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
I would love to have this feature added and I can test it if one tester is enough :)

bartoszmajsak:
Technically speaking this code is ported from [bash completion](https://github.com/juven/maven-bash-completion) project, hence I believe if it eventually will be merged we should think about adding attribution of some sort?

mcornella:
I'm opening this because this is the implementation that I'll merge, but first I want to rebase it and clean up a bit the code. You don't have to do anything, I'll work on it this week hopefully.

PR #2924, opened 2014-07-02
Respect all git versions

nicorevin:
Since 1.8.5 git doesn't use `# ` in output. See http://blogs.atlassian.com/2013/12/whats-new-git-1-8-5/

### DIFF
--- a/plugins/git-prompt/gitstatus.py
+++ b/plugins/git-prompt/gitstatus.py
@@ -24,14 +24,14 @@ if error:
 lines = output.splitlines()
 
 behead_re = re.compile(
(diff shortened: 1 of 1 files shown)


apjanke:
If you add the `-b` option to `--porcelain` you'll get the branch information (ahead/behind/diverged). That's where `lib/git.zsh` gets it from for the git-enabled theme prompts.

ncanceill:
@nicorevin what about what @apjanke just [suggested](https://github.com/robbyrussell/oh-my-zsh/pull/2924#issuecomment-88395220)?

nicorevin:
I would better get rid of it at all as it is way too expensive for a shell prompt

PR #2927, opened 2014-07-03
Add pyenv-virtualenvwrapper

defcube:
Based on virtualenvwrapper, but useful if using pyenv

### DIFF
--- a/plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
+++ b/plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
@@ -1,0 +1,51 @@
+eval "$(pyenv init -)"
+eval "$(pyenv-sh-virtualenvwrapper -)"
+
+  if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
+    # Automatically activate Git projects's virtual environments based on the
+    # directory name of the project. Virtual environment name can be overridden
+    # by placing a .venv file in the project root with a virtualenv name in it
+    function workon_cwd {
+        if [ ! $WORKON_CWD ]; then
+            WORKON_CWD=1
+            # Check if this is a Git repo
+            PROJECT_ROOT=`git rev-parse --show-toplevel 2> /dev/null`
+            if (( $? != 0 )); then
+                PROJECT_ROOT="."
+            fi
+            # Check for virtualenv name override
+            if [[ -f "$PROJECT_ROOT/.venv" ]]; then
+                ENV_NAME=`cat "$PROJECT_ROOT/.venv"`
+            elif [[ "$PROJECT_ROOT" != "." ]]; then
+                ENV_NAME=`basename "$PROJECT_ROOT"`
+            else
+                ENV_NAME=""
+            fi
+            if [[ "$ENV_NAME" != "" ]]; then
+                # Activate the environment only if it is not already active
+                if [[ "$VIRTUAL_ENV" != "$WORKON_HOME/$ENV_NAME" ]]; then
+                    if [[ -e "$WORKON_HOME/$ENV_NAME/bin/activate" ]]; then
+                        workon "$ENV_NAME" && export CD_VIRTUAL_ENV="$ENV_NAME"
+                    fi
+                fi
+            elif [ $CD_VIRTUAL_ENV ]; then
+                # We've just left the repo, deactivate the environment
+                # Note: this only happens if the virtualenv was activated automatically
+                deactivate && unset CD_VIRTUAL_ENV
+            fi
+            unset PROJECT_ROOT
+            unset WORKON_CWD
+        fi
+    }
+
+    # Append workon_cwd to the chpwd_functions array, so it will be called on cd
+    # http://zsh.sourceforge.net/Doc/Release/Functions.html
+    # TODO: replace with 'add-zsh-hook chpwd workon_cwd' when oh-my-zsh min version is raised above 4.3.4
+    if (( ${+chpwd_functions} )); then
+        if (( $chpwd_functions[(I)workon_cwd] == 0 )); then
+            set -A chpwd_functions $chpwd_functions workon_cwd
+        fi
+    else
+        set -A chpwd_functions workon_cwd
+    fi
+  fi


jonafato:
I've been using this recently and am happy with it so far. +1, would like to see it merged.

jonafato:
It looks like as of [this commit](https://github.com/robbyrussell/oh-my-zsh/commit/7b478d75a3a0db466e78b6e46a0e468d64354618), the virtualenvwrapper plugin no longer depends on the directory being a git repository. This seems like an improvement. @defcube If you're on board with this change, I'm happy to send you a patch.

Kriechi:
I think pyenv should not be activated in this plugin. There is a separate plugin to handle pyenv, and doing it again here might cause double-loading issues and take more time.

@defcube #6165 recently introduced a simplified pyenv plugin which also loads `pyenv virtualenv-init`. Would it make sense to integrate a simplified version of you virtualenvwrapper, similar to virtualenv-init?
I'm not deeply familiar with both tools, but I guess they kinda do similar things or might even provide identical functionality?

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
+eval "$(pyenv init -)"
```
cyphus: Pyenv initialization is handled in the pyenv plugin. I'd recommend loading it directly from here. I think there's an oh-my-zsh root directory variable that you can reference for portability.

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+eval "$(pyenv-sh-virtualenvwrapper -)"
```
cyphus: Check that the command is available before loading.

cyphus: Also, any thoughts on using pyenv-sh-virtualenvwrapper_lazy as the default or at least allowing it as an option?

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+
+  if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
```
cyphus: Rather than using `!` to negate and `-eq` to test for equality, use `-ne`:

`if [[ $DISABLE_VENV_CD -ne 1 ]];`

It's more readable that way.

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+    # Automatically activate Git projects's virtual environments based on the
+    # directory name of the project. Virtual environment name can be overridden
+    # by placing a .venv file in the project root with a virtualenv name in it
+    function workon_cwd {
+        if [ ! $WORKON_CWD ]; then
```
cyphus: What is this conditional `WORKON_CWD` protecting against?

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+                        workon "$ENV_NAME" && export CD_VIRTUAL_ENV="$ENV_NAME"
+                    fi
+                fi
+            elif [ $CD_VIRTUAL_ENV ]; then
+                # We've just left the repo, deactivate the environment
+                # Note: this only happens if the virtualenv was activated automatically
+                deactivate && unset CD_VIRTUAL_ENV
+            fi
+            unset PROJECT_ROOT
+            unset WORKON_CWD
```
cyphus: Don't forget to unset `ENV_NAME` too.

cyphus: The control flow of this function could be simplified a bit. For example, the code for if the current directory is outside a git repo is spread out across multiple conditionals.

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+    # http://zsh.sourceforge.net/Doc/Release/Functions.html
+    # TODO: replace with 'add-zsh-hook chpwd workon_cwd' when oh-my-zsh min version is raised above 4.3.4
+    if (( ${+chpwd_functions} )); then
+        if (( $chpwd_functions[(I)workon_cwd] == 0 )); then
+            set -A chpwd_functions $chpwd_functions workon_cwd
+        fi
+    else
+        set -A chpwd_functions workon_cwd
+    fi
+  fi
```
cyphus: Most zsh plugins load tools, aliases and statusbar functions into the environment. This plugin does a bit more in that it automatically activates and deactivates environments based on the working directory. I think that is fine to do so long as users of this plugin understand that. Can you put a comment near the top that describes what this plugin does and any options (like DISABLE_VENV_CD) that users can set in their `.zshrc` to control how this plugin works?

I'm a little bit wary of adding this to oh-my-zsh based on the fact that the latest pyenv-virtualenvwrapper version is almost 4 years old at this point. The readme for pyenv-virtualenvwrapper even starts with a link to pyenv-virtualenv, saying it has more convenient commands. It does look like this plugin already has some happy users though.

Overall this is a good PR. Thanks for the contribution!

plugins/pyenv-virtualenvwrapper/pyenv-virtualenvwrapper.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+                deactivate && unset CD_VIRTUAL_ENV
+            fi
+            unset PROJECT_ROOT
+            unset WORKON_CWD
+        fi
+    }
+
+    # Append workon_cwd to the chpwd_functions array, so it will be called on cd
+    # http://zsh.sourceforge.net/Doc/Release/Functions.html
+    # TODO: replace with 'add-zsh-hook chpwd workon_cwd' when oh-my-zsh min version is raised above 4.3.4
```
cyphus: If this were to be merged into oh-my-zsh, the plugin would only be shipped with oh-my-zsh > 4.3.4. You can add a conditional to check the OMZ version, but it should be safe to assume that you can run `add-zsh-hook` here.

maxbrunet:
Hi there,
I opened the PR #6911 to incorporate `pyenv-virtualenvwrapper` support to the exiting [`virtualenvwrapper` plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/virtualenvwrapper).
It would avoid code duplication. Please, check it out and leave a :+1: if it's good for you.
And feedback is always welcome.

PR #2931, opened 2014-07-05
Added helper for checking NPM package versions

bfricka:
This is very helpful b/c it will check the NPM registry version (default behavior), but also the local and global versions of the package, and will print these versions if found.

Before:

```
→ npm v gulp version
3.8.5
```

After:

```
→ nvv gulp
NPM registry package: 3.8.5
Local package: 3.8.4
Global package: 3.8.5
```


mcornella:
Closing due to age and lack of interest. Thanks for your contribution though!

Issue #2646, opened 2014-03-23
New tabs don't open with same directory


KevinBongart:
I remember this feature to work, but stopped working for me a while ago. Now, on a clean OS X Mavericks install, up-to-date oh-my-zsh with the `osx` plugin, I can see it doesn't work either.

Am I the only one? In which case I would investigate my config.

I disabled all plugins and only left `git` and `osx`, still the same issue. I also tried with and without the "New tabs open with same working directory" in the Terminal.app preferences, no luck.

Note: I also have `autojump` installed. Could it interfere? (disabled it, no change)

PR #2932, opened 2014-07-05
Fix terminalapp plugin #2416

### LINKED ISSUE #2416: terminalapp plugin is broken & obsolete in Mavericks
terminalapp plugin is broken. It's causing the following behaviour: http://stackoverflow.com/questions/11146306/zsh-tab-completion-bug-on-ubuntu

Also, the whole functionality has been replaced by Mavericks natively.

outpunk:
Fix #2416 in terminalapp plugin

### DIFF
--- a/plugins/terminalapp/terminalapp.plugin.zsh
+++ b/plugins/terminalapp/terminalapp.plugin.zsh
@@ -2,38 +2,45 @@
 # based on this answer: http://superuser.com/a/315029
 # 2012-10-26: (javageek) Changed code using the updated answer
 
-# Tell the terminal about the working directory whenever it changes.
-if [[ "$TERM_PROGRAM" == "Apple_Terminal" ]] && [[ -z "$INSIDE_EMACS" ]]; then
-  update_terminal_cwd() {
-        # Identify the directory using a "file:" scheme URL, including
-        # the host name to disambiguate local vs. remote paths.
+# Run only on OS X < 10.9 because the whole functionality has been replaced by Mavericks natively.
+# https://github.com/robbyrussell/oh-my-zsh/issues/2416
+if [[ "$(uname)" == "Darwin" ]] && [[ "$(uname -r | cut -d . -f 1)" -lt 13 ]]; then
 
-        # Percent-encode the pathname.
-        local URL_PATH=''
-        {
-            # Use LANG=C to process text byte-by-byte.
-            local i ch hexch LANG=C
-            for ((i = 1; i <= ${#PWD}; ++i)); do
-                ch="$PWD[i]"
-                if [[ "$ch" =~ [/._~A-Za-z0-9-] ]]; then
-                    URL_PATH+="$ch"
-                else
-                    hexch=$(printf "%02X" "'$ch")
-                    URL_PATH+="%$hexch"
-                fi
-            done
+    # Tell the terminal about the working directory whenever it changes.
+    if [[ "$TERM_PROGRAM" == "Apple_Terminal" ]] && [[ -z "$INSIDE_EMACS" ]]; then
+      update_terminal_cwd() {
+            # Identify the directory using a "file:" scheme URL, including
+            # the host name to disambiguate local vs. remote paths.
+
+            # Percent-encode the pathname.
+            local URL_PATH=''
+            {
+                # Use LANG=C to process text byte-by-byte.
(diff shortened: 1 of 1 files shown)


mcornella:
Just realised we pushed a duplicate version of Terminal.app plugin in `lib/termsupport.zsh`, though I'm not sure about it causing #2416. Gonna have to ask @apjanke about this.

apjanke:
I, uhhh, was not aware of the existence of the terminalapp plugin. (And neither was the poster of #2646, which prompted its addition.)

I read through the plugin's source. Yeah, the stuff I added to `lib/termsupport.zsh` in #3429 is just a less-fleshed-out dup of the existing functionality in the terminalapp plugin.

FWIW, I don't think it's causing #2416. (Either the plugin or #3429.) For one, looking through the StackOverflow post, it looks like it's related to color/bold control stuff and not the pwd reporting stuff. Also, I'm on Mavericks, and I'm not seeing the bad behavior described in #2416, with or without the terminalapp plugin enabled.

Anyway, it sounds like we should just remove the #3429 Terminal.app stuff from `termsupport.zsh` and tell #2646 (and me) to turn on the terminalapp plugin instead.

As an aside, I'm curious about what #2416 means by "the whole functionality has been replaced by Mavericks natively". I'm on Mavericks, and when I don't have the oh-my-zsh pwd-updating behavior enabled, Terminal.app running zsh is not aware of pwd changes. AFAIK, this is implemented in OS X via `/etc/bashrc` which doesn't apply to zsh invoked through normal means. (Didn't work for me on Yosemite back when I was running it, either.)

mcornella:
Thank you Andrew, I also thought the linked SO thread in #2416 wasn't related to this functionality. I went a little further and a quick google search returned [this other SO thread](http://stackoverflow.com/questions/23090976/zsh-new-terminal-tab-opens-in-home-folder-instead-of-cwd-if-foreground-process) which complains about being unable to open a new tab on the same folder **on Mavericks**. Other related links also mention the same problem and the same solution, see [superuser.com](http://superuser.com/questions/313650/resume-zsh-terminal-os-x-lion), [russellfinn.com](http://russellfinn.com/2013/10/25/current-working-directory-restoration-in-mavericks-terminal-with-tcsh/), and PR #522.

FWIW, I think the way to go is keeping the functionality in `lib/termsupport.zsh`: we want to maintain every type of terminal emulator as a first class citizen and out of the box, without the need to enable a plugin. I'll be following closely the current implemented solution and how it was arrived to, since it seems more complex than the one we pushed. 

If we could also hear how exactly this breaks Ubuntu or other non-OSX platforms we can maybe provide a better fix. Work-in-progress will be, it seems.

apjanke:
The implementation in `plugins/terminalapp` URL-escapes all special characters. The simpler one in `lib/termsupport.zsh` escapes only spaces. The more complex one in the plugin is better; dirs with special characters in them can break the simpler one we pushed in `termsupport`. Here's an example.

    $ mkdir '$#@%^&*%$'
    $ cd \$\#@%\^\&\*%\$ 

Also it looks like the `termsupport.zsh` version is using the incorrect variable `$HOSTNAME` instead of `$HOST`. (I think `$HOSTNAME` is a bash-ism.)

I also like the idea of pulling it in to `lib/termsupport.zsh` so it's supported out of the box.

apjanke:
@mcornella - I made PR #3582 that will go ahead and fold terminalapp in to termsupport. Went ahead and did it because we need to switch over to their implementation to fix the escaping and host issues anyway.

Anybody else we should ask about merging the plugin? Do plugins have notional owners/maintainers?

mcornella:
> Do plugins have notional owners/maintainers?

Plugin mantainers **in theory** are all documented [in the wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins#terminalapp), below each plugin (ping @OutPunk). 
Another method we used with @ncanceill was looking at the history of the plugin files using `git log -- path/to/plugin/directory`, or [github itself](https://github.com/robbyrussell/oh-my-zsh/commits/master/plugins/terminalapp/terminalapp.plugin.zsh). 

@apjanke regarding this issue, it should be auto-closed too with one of the PRs you made.

apjanke:
Thanks. I'll read through the rest of the wiki, and I'm getting up to speed on `git log`/`git blame`.

PR #2933, opened 2014-07-06
Integra plugin 'pro'

jonDotsoy:


mcornella:
BTW, I never got to say: bonne espagnole, @ncanceill :grin:

ncanceill:
Thanks ;-)

Sorry, I am unable to return the compliment: in French, the Spanish language is masculine, thus you should have written _"bon espagnol"_.

mcornella:
Haha my bad! :grin:

PR #2938, opened 2014-07-07
Added case insensitive substring completion

yuri-rage:
Existing matcher-list patterns make case insensitive and substring completion mutually exclusive.  By adding a few patterns, we can have both!

e.g.,
Before: `cat adm [tab]` = `cat adm` (no result)

After: `cat adm [tab]` = `cat README.md`

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -15,7 +15,17 @@ if [ "x$CASE_SENSITIVE" = "xtrue" ]; then
   zstyle ':completion:*' matcher-list 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
   unset CASE_SENSITIVE
 else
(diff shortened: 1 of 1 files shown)


varenc:
this is quite old...but I found it when I was searching for how to get case insensitive substring completion! Right now the with `CASE_SENSITIVE="false"` set to, oh-my-zsh configures zsh in such a way that it's case-insensitive when matching the beginning on a file, but case-sensitive when matching a substring.

varenc:
this is quite old...but I found it when I was searching for how to get case insensitive substring completion! It's exactly what I want.

Right now the with`CASE_SENSITIVE="false"`, oh-my-zsh configures zsh in such that it's case-insensitive when matching the beginning on a f ile, but case-sensitive when matching a substring.  I find this confusing a flow-breaking at times.  I love this fix!  

I bet this is the expected behavior for most users.

robbyrussell:
My apologies for us never getting around to this. Am closing to clean up our backlog a bit given how long this has lingered.

PR #2944, opened 2014-07-10
Add gem providing tab completion for the spring gem

JustinAiken:
Adds tab completion for [Spring](https://github.com/rails/spring), which is now part of Rails itself.

..because whenever I type `spring` and hit tab and nothing happens... I get sad


robbyrussell:
Closing for reasons mentioned in #3020.

Thanks for the pitch, @JustinAiken!

Issue #2942, opened 2014-07-10
Enable colored man on illumos/Solaris


moinakg:
The following patch enables proper colored man support on Solaris and illumos platforms.

diff --git a/plugins/colored-man/colored-man.plugin.zsh b/plugins/colored-man/colored-man.plugin.zsh
index 5605628..8877446 100644
--- a/plugins/colored-man/colored-man.plugin.zsh
+++ b/plugins/colored-man/colored-man.plugin.zsh
@@ -1,3 +1,22 @@
+un=`/usr/bin/uname -s`
+if [ "$un" = "SunOS" ]
+then
+	if [ ! -x ${HOME}/bin/nroff ]
+	then
+		mkdir -p ${HOME}/bin
+		cat > ${HOME}/bin/nroff <<EOF
+#!/bin/sh
+if [ -n "\$_NROFF_U" -a "\$1,\$2,\$3" = "-u0,-Tlp,-man" ]; then
+	shift
+	exec /usr/bin/nroff -u\${_NROFF_U} "\$@"
+fi
+#-- Some other invocation of nroff
+exec /usr/bin/nroff "\$@"
+EOF
+	chmod +x ${HOME}/bin/nroff
+	fi
+fi
+
 man() {
       env \
       	  LESS_TERMCAP_mb=$(printf "\e[1;31m") \
@@ -7,5 +26,8 @@ man() {
 	  LESS_TERMCAP_so=$(printf "\e[1;44;33m") \
 	  LESS_TERMCAP_ue=$(printf "\e[0m") \
 	  LESS_TERMCAP_us=$(printf "\e[1;32m") \
+	  PAGER=/usr/bin/less \
+	  _NROFF_U=1 \
+	  PATH=${HOME}/bin:${PATH} \
 	  			   man "$@"
 }

PR #2946, opened 2014-07-11
Solaris portability tweaks.

moinakg:
As a follow-up to issue #2942 here's a patch that fixes a few things on Solaris.
- Enable colored man
- Fix an error from check_for_upgrade.sh (_current_epoch:8: bad math expression: operand expected at `%s / 60 / ...')
- Enable kill command completion

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -3,7 +3,7 @@
 zmodload zsh/datetime
 
 function _current_epoch() {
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @moinakg, I'm in the middle of refactoring the `colored-man-pages` plugin and I was wondering, why are we doing the `nroff -u1` thing? Is it still needed or can we get rid of it?

PR #2962, opened 2014-07-17
Random theme functionality encapsulated in a theme.

DArtagan:
Hello!
The statements for selecting a random theme in oh-my-zsh.sh and the themes plugin are duplicate. Most people eventually settle on a theme, making those lines in oh-my-zsh.sh superfluous. To address those points, it may make sense to put the random theme functionality into a theme of its own (since themes are just zsh-script).  I understand this could conflict with other PRs like #2902, I'm happy to edit this PR as needed if other changes happen first.  Thoughts?

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -70,21 +70,12 @@ done
 unset config_file
 
 # Load the theme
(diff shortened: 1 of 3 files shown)


Marcholio:
This could be closed due to inactivity

PR #2975, opened 2014-07-22
add storm plugin

halfelf:
Plugin for [apache-storm](http://storm.incubator.apache.org)'s command line client.


robbyrussell:
Given that this has been lingering for 7 years and nobody else has upvoted the plugin, I'll close this for not and if people rise up and ask for it, we can bring it in then. :-)

PR #2980, opened 2014-07-25
Add a newline to the prompt for convenience

ikhthiandor:


apjanke:
I'd go with #2984, where it's optional, and current default behavior is preserved. Given that `agnoster` was written without a newline at the end, and it would have been easy to add, we can probably assume that its omission was intentional, and its author and users are used to the current format.

ikhthiandor:
Closing this PR since it doesn't seem to bother other users much and I've settled into bira theme for a while.

mcornella:
We'll halt all new changes to the agnoster theme for now, unless they fix a bug, to see if we can make it converge with the [main agnoster theme](https://github.com/agnoster/agnoster-zsh-theme). So you can still submit your modification there.

Another option is to put the custom `prompt_build` and `prompt_newline` functions in your zshrc file, which will override the default ones and get you the same behavior without having to get this merged.

PR #2982, opened 2014-07-27
Makepkg plugin

AlexTalker:
Plugin provide aliases for usually [makepkg](https://wiki.archlinux.org/index.php/Makepkg) commands.


mcornella:
Closed for lack of support. You can add it as a local plugin by looking at the [Customization wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#adding-a-new-plugin).

PR #2984, opened 2014-07-28
Give agnoster an option to provide prompt on newline

wenlock:
This change adds a new environment export called PROMPT_ON_NEWLINE.
For example, when configured in ~/.zshrc as export PROMPT_ON_NEWLINE=true,
the prompt will always show on a new line so that you have more room to type.
Use full when working in deep folders.   Should be backwards compatible
with previous default behavior.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -178,4 +178,9 @@ build_prompt() {
   prompt_end
 }
 
-PROMPT='%{%f%b%k%}$(build_prompt) '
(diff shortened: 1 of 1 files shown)


mcornella:
Halting all modifications to the agnoster theme for now. You can still submit this to the [main agnoster repo](https://github.com/agnoster/agnoster-zsh-theme). See https://github.com/robbyrussell/oh-my-zsh/pull/2980#issuecomment-172630273.

PR #2985, opened 2014-07-28
Fix the wrong agnoster.zsh-theme git symbol " " to "⭠"

shengwen1997:
Hello,

I am a Taiwanese who speak Chinese.
I recently setup the oh-my-zsh on my Ubuntu notebook and I found this error. (14.04)

Fixed

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -100,7 +100,7 @@ prompt_git() {
     zstyle ':vcs_info:*' formats ' %u%c'
     zstyle ':vcs_info:*' actionformats ' %u%c'
     vcs_info
-    echo -n "${ref/refs\/heads\// }${vcs_info_msg_0_%% }${mode}"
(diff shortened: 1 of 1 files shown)


mcornella:
@shengwen1997 you need to install the right ones from the repository I linked. The instructions are in this link: https://github.com/powerline/fonts/blob/master/README.rst
Then, depending on your terminal emulator, you'll have to set it up so that it uses the updated fonts.

apjanke:
Ran in to this same issue recently with #3713. Installing the updated font that mocrnella linked to fixed it. 

It turned out that the `agnoster` section in the Themes wiki page was linking to the outdated version of the font, and causing some confusion for users. We updated the wiki page to point to the current, correct font.

Did installing updated fonts fix it for you too, @shengwen1997?

apjanke:
I think this can be closed now. @mcornella, can this be added to one of the big cleanup lists?

mcornella:
It sure can

PR #2999, opened 2014-08-01
fix paclist() for non-english locales

nicorevin:


mcornella:
:+1: 
The command comes from [this exact forum post](https://bbs.archlinux.org/viewtopic.php?pid=730933#p730933), see that it's provided without a call to `sudo`, probably because no actual change is made.

/cc @robbyrussell

nicorevin:
no need for `sudo`. LC_ALL added because of filtering by "Description" which is locale sensitive.

PR #3006, opened 2014-08-08
Add an option to the tmux plugin to force 256 colors. 

sh4wn:
Tmux has a command line option to force 256 colors wether tmux thinks the terminal supports it or not. For some reason tmux thinks gnome-terminal doesn't support 256 colors, so this option makes life a bit easier. 

This pull request adds an option for the oh-my-zsh tmux plugin, so tmux can be called with the '-2' command line argument, forcing 256 colors.


sh4wn:
As can be read in pull request #3128, this can be easily fixed by setting the TERM environment variable.

PR #3007, opened 2014-08-08
Added simple support for Rust Cargo

dahankzter:
The would be Rust build system.


andradei:
@dahankzter Have you seen this [zsh plugin](https://github.com/rust-lang/zsh-config)?

You could make a PR there too to add the cargo autocompletion, whoever wants to use it should have to manually download and configure it, or use [antigen](https://github.com/zsh-users/antigen) which is currently a more active project.

PR #3008, opened 2014-08-08
A new plugin for oh-my-zsh: easy zsh named directory management

MikeDacre:
This plugin allows the user to create, change, and delete named directories easily. It also provides an autocompletion function, and a function called folder_name() that can be included in PROMPT to show the user all the names of the current folder (not including default names such as '~' or '~user').

It stores the bookmarks in the file `$HOME/.zshbookmarks` by default, but this is customizable.

I have used this function for years since I first wrote it, and I think the community could benefit from it :)

Happy to change anything.

### DIFF
--- a/plugins/cdbk/cdbk.plugin.zsh
+++ b/plugins/cdbk/cdbk.plugin.zsh
@@ -1,5 +1,5 @@
 # cdbk.plugin.zsh (cd bookmarking for zsh)
-# Last modified: 2014-08-31 21:25
+# Last modified: 2014-08-31 21:51
 
 # Define location of bookmark file and source it every time this file is sourced
 ZSH_BOOKMARKS="$HOME/.zshbookmarks";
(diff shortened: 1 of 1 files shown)


popstas:
@MikeDacre, what is the difference between your plugin and [jump](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/jump/jump.plugin.zsh) plugin?

MikeDacre:
@popstas Wow. I thought no one cared about this at all, it has been 18 months since I submitted this.

From looking at jump it seems that it is essentially the same idea, cdbk just has a nicer ui and is able to parse out hardcoded named directories from ~/.zshrc.

I didn't know jump existed when I submitted cdbk 18 months ago, feel free to drop the pull request, I don't really care that much. I wrote cdbk six years ago, and I am not switching to jump, particularly given the extra features of cdbk and the fact I use it every day. However, considering the liveliness of this pull request, my assumption is that interest in cdbk is essentially zero, so just delete the pull request if you want.

mcornella:
You can add your plugin to the [external plugins list](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins).

PR #3018, opened 2014-08-11
feat(plugins): add `term_tab` plugin to complete other zsh sessions' directories

deathjest3r:
This plugin allows the user to complete other zsh session pathes by hitting 'CTRL+v'.


popstas:
Works for me on Ubuntu 15.10, also it complete zsh session in tmux.

deathjest3r:
Thank you for testing

PR #3020, opened 2014-08-12
add spring aliases

passalini:
Just some aliases to spring (Rails application preloader)


robbyrussell:
Similar to what @Kriechi conveyed, I don't think that we should be running spring directly all that often. Am going to close this for now and assume we're mostly doing things via bin/rails and/or via bundler in 2021. 

Sorry for the delayed response on this!

PR #3021, opened 2014-08-12
Clean up and add new vundle commands

steve0hh:
Vundle [quick start readme](https://github.com/gmarik/Vundle.vim#quick-start) is recommending users use 

    $ git clone https://github.com/gmarik/Vundle.vim.git ~/.vim/bundle/Vundle.vim
instead of 

    $ git clone http://github.com/gmarik/vundle.git ~/.vim/bundle/vundle

So I've changed them accordingly in this pull request.

Vundle is also currently undergoing an [interface change](https://github.com/gmarik/Vundle.vim/blob/v0.10.2/doc/vundle.txt#L372-L396).

Changed to the new `PluginInstall` command instead of `BundleInstall` so on and so forth.


mcornella:
This is solved in the current version, please close.

PR #3031, opened 2014-08-16
Updated cabal completion to leverage standard bash approach.

carymrobbins:
Also cleaned up cabal sandbox prompt info.


carymrobbins:
Closing since this is old and likely not to be merged.

PR #3033, opened 2014-08-16
Auto completion for some 'forever' commands

RolfKoenders:
Some auto completion for the program 'forever'. 
My first try on a zsh plugin, so maybe some optimizing is needed?

### DIFF
--- a/plugins/forever/README
+++ b/plugins/forever/README
@@ -1,0 +1,10 @@
+# Forever
+
+----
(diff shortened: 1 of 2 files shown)


RolfKoenders:
I think we can close this one. I don't even know if it still works or not.. I can check later. Im also not using forever anymore (almost 2 years later) 😌

PR #3034, opened 2014-08-17
Add zbell plugin

bullno1:
This is a modified version of this script https://gist.github.com/oknowton/8346801, which notifies users when a long running command has finished.


Renrhaf:
+1

PR #3037, opened 2014-08-18
[docker plugin] Add support for containers names and stopped containers

pchaussalet:
Since version 0.6.5, docker containers can be named. The zsh plugin should be ale to use them, too.


leifcr:
#5060 solves that by using upstream completion.

PR #3059, opened 2014-08-26
Adding BBEdit plugin (OS X only)

chrisfsmith:


Calorion:
The readme needs updated; BBEdit no longer requires downloading a separate command line tool (just execute BBEdit→Install Command Line Tools…), and…there’s no text to add to the `.zshrc` file!

PR #3063, opened 2014-08-28
Add postgres utils command completion

felixbuenemann:
This allows to tab complete eg. database names.

Source: https://github.com/zsh-users/zsh-completions/blob/master/src/_pgsql_utils


mcornella:
[There's a postgres completion file shipped within zsh](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_postgresql).

PR #3066, opened 2014-08-29
Improved kclean command

la-magra:
The original kclean returned 'zsh: no matches found: '  in some machines
and in others requested to remove the current headers to resolve unmet
dependencies.
So I optimized it a little, now it should work seamlessly in most debian/ubuntu 
machines.


ajpaulson:
I know this is old - but it works great for me :+1:

eclectic-coding:
What is the status? I would like this shortcut to work for me.

Issue #2587, opened 2014-03-08
Remove all copyright information in themes/plugins


robbyrussell:
We're using Git so that we can all contribute to the project... together. If someone has some time, I would like to remove all copyright information from each theme/plugin that someone has added themselves to. 

Thanks to version control--we will all know who contributed what to each and every line of code. My big concern is that someone might contribute a plugin with a copyright and then others might not feel like they can modify it. 

Lastly, it's just not a convention across all plugins--so let's tidy this up going forward.

PR #3078, opened 2014-09-01
gulp.js plugin

akoenig:


mmrko:
FWIW as per [gulp docs](https://github.com/gulpjs/gulp-cli/tree/master/completion) you can just add `eval "$(gulp --completion=zsh)"` to your `.zshrc`. The plugin should probably use this shorthand as well (in the same vein as the [npm plugin](https://github.com/robbyrussell/oh-my-zsh/blob/c78277fd8bda5fec87504469afdf121355876006/plugins/npm/npm.plugin.zsh)).

akoenig:
@mmrko True story. I was aware of this solution while building my thingy. This one is starting gulp (and therefore a whole Node.js process) which causes performance problems on low-end systems like the RPi.

mmrko:
Nice work! Yours is indeed way snappier. Maybe this could be a PR over at [gulp-cli](/gulpjs/gulp-cli). What do you think? Would help tackle gulpjs/gulp/issues/851 as well :)

dafstone:
+1 - quick and handy.

mmrko:
Something to take into account if you tend to split your gulp tasks into multiple files is that this will only pick up tasks defined in gulpfile.js.

fundon:
+1, Nice

carueda:
:+1:

davidbakertv:
+1

mcornella:
Hi @akoenig, please rename your completion function to `__gulp_completion` or similar, it's not good practice to use `$` on function names. 

You should also put the information of the plugin in a separate `README.md` file so that the `plugin.zsh` one only has the functionality (for example, the [sublime text](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/sublime/README.md) plugin). The copyright info is redundant, too (see #2587). 

Lastly, you can put completion in a file named `_gulp` and zsh will read it by default. That's a common and better practice for plugins that only provide autocompletion. See [adb](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/adb/_adb), [nvm](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/nvm/_nvm), [pass](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/pass/_pass) or [pip](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/pip/_pip) plugins to get the basic structure of completion files.

macgyver:
I'm not qualified to code review an oh-my-zsh plugin but I added this to my own `~/.zshrc` and it works perfectly for all my gulp tasks, thanks!

aforty:
+1

ScreenDriver:
+1

PR #3091, opened 2014-09-03
Check if there is a git repository dir.

harveydf:
This pull request fix problem when you are on child dir of the repository. https://github.com/robbyrussell/oh-my-zsh/pull/1504/


mcornella:
This may still be a good idea to merge...

yograterol:
yep :)

PR #3098, opened 2014-09-05
Vagrant improvements

lexinator:


mcornella:
The completion script has diverged significantly since then. Open a new PR if the changes are still relevant today.

PR #3101, opened 2014-09-05
removed trailing "; " of default $command from tab function.

drq883:
If it is there, iterm puts the command there, but does not execute it.

I made this mod and the function tab() works on iTerm as it should.

Thanks.


apjanke:
PR #4093 includes the same fix, plus some other stuff. Sorry for stepping on your PR; I didn't see it. I'll see if I can fiddle with the Git history to include your commit so you get credit if it's merged.

PR #3107, opened 2014-09-07
add --depth operand

Nefelim4ag:
git clone --depth can speed up zsh installing, because only latest N commit will be downloaded


mcornella:
Implemented in #3493. Please close

PR #3109, opened 2014-09-07
Remove duplicate `git diff` alias

erikjansson:
Removed the extra `gd` alias from `git.plugin.zsh`!


mcornella:
Not relevant anymore.

PR #3113, opened 2014-09-08
Fix unquoted var borking with "setopt shwordsplit" & whitespace

rowanthorpe:
When I needed zsh to do wordsplitting Bourne-style (using "setopt shwordsplit") oh-my-zsh borked on an unquoted var. This fixes it.


rowanthorpe:
Does this still need to be open here, given that the pull-request was merged in per-directory-history years ago, or should I close it now? @jimhester

rowanthorpe:
I can see the plugin has been updated far enough to incorporate the change from there, so I am closing this.

PR #3123, opened 2014-09-11
`test -d` to make sure awscli is actually installed

quodlibetor:
brew --prefix shows the location that a package is or would be
installed:

https://github.com/Homebrew/homebrew/blob/a8582c58/share/man/man1/brew.1#L448-L449


cristim:
@quodlibetor Is this still needed on the current version? 

Also, it looks like if you squash these three commits the first two are actually useless.

quodlibetor:
Seems like latest plugin version is sane in the face of non-brew, but I am spending most of my days on OSX these days so I can't actually check.

PR #3133, opened 2014-09-13
last-working-dir working with spaces in dirnames

choener:
This commit should make the last-working-dir plugin work with directories that contain spaces (at least for me).


0xC0FFEE:
This fixed it for me too! Thanks!

mcornella:
:+1:, but please rebase to keep only the one relevant commit. 

/cc @robbyrussell

PR #3135, opened 2014-09-14
Add a plugin: tig

adamme:
This plugin adds tig command completion for zsh. The core functionality is taken from the official tig bash completion and it works just fine with zsh.

[Tig](http://jonas.nitro.dk/tig/) is an awesome text-mode interface for git, if you don't know about it yet.

### DIFF
--- a/plugins/tig/tig-completion.bash
+++ b/plugins/tig/tig-completion.bash
@@ -1,0 +1,291 @@
+##
+# bash completion support for tig
+#
+# Copyright (C) 2007-2010 Jonas fonseca
+# Copyright (C) 2006,2007 Shawn Pearce
+#
+# Based git's git-completion.sh: http://repo.or.cz/w/git/fastimport.git
+#
+# The contained completion routines provide support for completing:
+#
+#    *) local and remote branch names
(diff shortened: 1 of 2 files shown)


zhugexiaobo:
+1

ncanceill:
Hi @adamme, and thanks for contributing.

You must realize that the file you want to include bears a copyright notice at the top. This is a clear indication that some rules may apply to copying the source.

Indeed, the repository where I assume you copied this from contains a [COPYING](https://github.com/jonas/tig/blob/master/COPYING) file, indicating the source distributes under GPLv2.

But GPL is less permissive than [the MIT license](https://github.com/robbyrussell/oh-my-zsh/blob/master/MIT-LICENSE.txt) under which OMZ distributes, and thus cannot be simply integrated as such.

I advise you close this PR and put those files in `$ZSH_CUSTOM`.

adamme:
@ncanceill Hi there,

Thanks for bringing up the license compatibility issue to my attention. I'll close this PR, even though I'd like very much to see it merged, which is clearly not feasible for the moment.

PR #3136, opened 2014-09-14
[!]: fix for aliases-completion

voku:
e.g. for -> alias gcc='colour gcc' <- https://github.com/voku/dotfiles/blob/master/.aliases


mcornella:
See https://github.com/robbyrussell/oh-my-zsh/pull/241#commitcomment-327732. Won't implement because it breaks stuff. Please close.

PR #3139, opened 2014-09-15
added powerline plugin

typekpb:
this makes powerline + zsh work in my setup. 
might be usable for others too.

Please note: manual powerline installation is still required


mcornella:
This is the same as not having a plugin at all. Please close or improve further

typekpb:
closing due to lack of motivation from my side (not a powerline user anymore)

PR #3141, opened 2014-09-15
sudo: maintain cursor position

muggenhor:
I.e. when prefixing the current command-line with 'sudo ' maintain the
current cursor position instead of jumping to the end of the line.

This takes advantage of the fact that LBUFFER always contains all text left of the cursor and also happens to contain all text at the start of the buffer currently being edited.


mcornella:
Let's ship this then
/cc @robbyrussell

mcornella:
Hi @muggenhor, can you add `Close #2522, close #3281, close #3554` so those are automatically closed?

PR #3155, opened 2014-09-17
Fix AWS plugin overriding currently set RPROMPT

ebr:
If RPROMPT is set by a theme, this plugin will rudely override it once an AWS profile is loaded.
This PR adds a function, following the git_prompt_info() pattern, that can be used in a theme to display the currently loaded AWS profile wherever suitable.


cristim:
@ebr, is this still needed on the current code? If so, please update the commit and let's try to have it merged.

mcornella:
Obsoleted by #4320

PR #3157, opened 2014-09-19
Refactored michelebologna's theme

### LINKED ISSUE #2665: Theme michelebologna breaks with (at least) wd plugin
I've been using the michelebologna theme for around a month now and noticed, occasionally, that the timestamp at the end of the line was sometimes prepended with the ascii code for green. I then found a way to reproduce it reliably, using wd to change to any warp point, like so:

image

Having looked at the wd script, there is a variable called $GREEN that has \033[92m as it's contents, but it should not have been called by the command in the screenshot above.

As far as I can see, the theme doesn't have the term \033[92m in it either, so I'm not sure how it's come about.

Having now switched to the "clean" theme I do not see this issue.

I can't be sure, but I think this behaviour is apparent when running commands other than wd.

mbologna:
Theme michelebologna refactored following Google's guidelines for scripts [ http://goo.gl/oJSXH2 ]:
* Using local variables [avoids name clashing with other scripts http://goo.gl/6Tv7r2 ] where possible (except for exports).
* Added comments on top of file to describe the features of the theme.

This is to resolve #2665

### DIFF
--- a/themes/michelebologna.zsh-theme
+++ b/themes/michelebologna.zsh-theme
@@ -1,44 +1,74 @@
-# reference colors
-GREEN="%{$fg_bold[green]%}"
-RED="%{$fg_bold[red]%}"
-CYAN="%{$fg_bold[cyan]%}"
-YELLOW="%{$fg_bold[yellow]%}"
-BLUE="%{$fg_bold[blue]%}"
-MAGENTA="%{$fg_bold[magenta]%}"
-WHITE="%{$fg_bold[white]%}"
+# Michele Bologna's theme
+# http://michelebologna.net
+#
+# This a theme for oh-my-zsh. Features a colored prompt with:
+# * username@host: [jobs] [git] workdir % 
+# * hostname color is based on hostname characters. When using as root, the 
+# prompt shows only the hostname in red color.
+# * [jobs], if applicable, counts the number of suspended jobs tty
+# * [git], if applicable, represents the status of your git repo (more on that
+# later)
+# * '%' prompt will be green if last command return value is 0, yellow otherwise.
+# 
+# git prompt is inspired by official git contrib prompt: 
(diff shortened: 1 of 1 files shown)


robbyrussell:
@mbologna can you rebase for me?

mbologna:
@robbyrussell done

mbologna:
@robbyrussell rebased on current master.

ncanceill:
So is this ready for merge?

mbologna:
Yes.

mbologna:
@ncanceill @robbyrussell there is a problem with michelebologna's theme in current robby's master.
At the moment the prompt complains that it cannot find a function which is defined in another PR: #4014

themes/michelebologna.zsh-theme
```diff
@@ -1,44 +1,74 @@
...
+ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE="="
+ZSH_THEME_GIT_PROMPT_AHEAD_REMOTE=">"
+ZSH_THEME_GIT_PROMPT_BEHIND_REMOTE="<"
+ZSH_THEME_GIT_PROMPT_DIVERGED_REMOTE="<>"
 
-# wrap all together
-PROMPT='$USERNAME_OUTPUT$HOSTNAME_OUTPUT:$CURRENT_DIR_OUTPUT $LAST_COMMAND_OUTPUT%#$RESET_COLOR '
-RPROMPT='%1(j.fg: [%j].) $GREEN$(git_prompt_info)$RESET_COLOR [%@]'
+PROMPT='$username_output$hostname_output:$current_dir_output%1(j. [$jobs_bg].)'
+PROMPT+='$(__git_ps1)'
```
ncanceill: This is the line at fault and should be fixed by #4014.

PR #3164, opened 2014-09-22
virtualenvwrapper: Deactivate only if in virtualenv

### LINKED ISSUE #2185: virtualenvwrapper plugin: deactivate returns error
I've enabled the `virtualenvwrapper` plugin and added a `.venv` file to my project folder. When I cd to it my venv is activated successfully and when I cd out of it it is deactivated successfully. So far so good. However when I manually run `deactivate` while inside the directory I get the following:
```bash
deactivate:12: command not found: virtualenv_deactivate
```

Issuing `which deactivate` (before getting above error) returns the following:

```bash
deactivate () {
  typeset env_postdeactivate_hook
  typeset old_env
  virtualenvwrapper_run_hook "pre_deactivate"
  env_postdeactivate_hook="$VIRTUAL_ENV/$VIRTUALENVWRAPPER_ENV_BIN_DIR/postdeactivate"
  old_env=$(basename "$VIRTUAL_ENV")
  virtualenv_deactivate $1
  virtualenvwrapper_run_hook "post_deactivate" "$old_env"
  if [ ! "$1" = "nondestructive" ]
  then
    unset -f virtualenv_deactivate > /dev/null 2>&1
    unset -f deactivate > /dev/null 2>&1
  fi
}
```

Some notes:

* I am not sourcing `virtualenvwrapper.sh` in my `.zshrc` (as it sourced by the plugin)
* I have installed `virtualenvwrapper==4.1.1` through `pip`.
* My Python (2.7.5) was installed with `homebrew`

nailor:
If user manually deactivates the virtualenv when using this mode, zsh
will produce following error:

    deactivate:12: command not found: virtualenv_deactivate

To avoid this, check that the VIRTUAL_ENV flag is set before trying to
automatically deactivate the virtual environment.

Fixes #2185

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -38,7 +38,7 @@ if (( $+commands[$virtualenvwrapper] )); then
                 source $ENV_NAME/bin/activate && export CD_VIRTUAL_ENV="$ENV_NAME"
               fi
             fi
-          elif [ $CD_VIRTUAL_ENV ]; then
+          elif [[ -n $CD_VIRTUAL_ENV && -n $VIRTUAL_ENV ]]; then
             # We've just left the repo, deactivate the environment
             # Note: this only happens if the virtualenv was activated automatically
             deactivate && unset CD_VIRTUAL_ENV


mcornella:
This is just a comment to show that **2** testers so far +1'd this PR:
* @toxsick [+1](https://github.com/robbyrussell/oh-my-zsh/issues/2185#issuecomment-61071537)
* @lucn [+1](https://github.com/robbyrussell/oh-my-zsh/issues/2185#issuecomment-63005300)

mcornella:
Some testing info from https://github.com/robbyrussell/oh-my-zsh/issues/2185#issuecomment-60320007:
> Well, you need to install the `virtualenvwrapper`. If you run a Linux distribution, they probably have it in a package manager (for Ubuntu and Debian the package is called virtualenvwrapper). If you run something else, like Os X, you need to install it via pip (`pip install virtualenvwrapper`) and make it available in the $PATH.
> 
> After that, make a directory called `foo` and create a virtualenv called foo with the wrapper (`mkvirtualenv foo`). After the creation is done, you can disable it by running deactivate. Now, if you cd into foo with the plugin enabled, it should change to virtualenv foo, which is shown on the prompt (depending on the theme I'd guess). Now, run deactivate and change the directory and you'll get the error. With this PR the error is gone.

@nailor: unfortunately, I can't reproduce the error you mention **nor the appropriate behavior of virtualenvwrapper**. Here's what I did (without this PR checked-out):

1. Installed virtualenvwrapper package because I use Debian.
2. Enabled `virtualenv` and `virtualenvwrapper` plugins. 
3. Realized `mkvirtualenv` wasn't available. Saw [this](http://askubuntu.com/questions/251378/where-is-virtualenvwrapper-sh) :disappointed: 
4. Purged virtualenvwrapper package; installed via pip.
5. Run `md foo; mkvirtualenv foo`. `(foo)` appeared in my prompt without cd-ing into foo directory (using agnoster theme). I rebooted the terminal and entered the foo directory, but no prompt changes. 
Only when I run `mkvirtualenv` the prompt changes, I don't know if the agnoster theme has a bug or there is something else in the `virtualenv` / `virtualenvwrapper` plugin.

The same behavior reproduced with your PR checked-out. What theme are you using?

nailor:
@mcornella This aims to fix the bug when you manually deactivate (run `deactivate`) and then change the directory outside the project, which causes following error:

    deactivate:12: command not found: virtualenv_deactivate

I'm running default theme in my test instance and the problem persists there. The bug can also be seen in the patch: Running `deactivate` removes virtualenv's `bin`-directory from the `PATH` and unsets the `VIRTUAL_ENV` variable. However, in the upstream code, we don't test if the `VIRTUAL_ENV` is still set, thus failing the `deactivate` call when exiting the directory.

mcornella:
Ok, I managed to have a working test, but the error I get is a little different:
```
workon_cwd:30: command not found: deactivate
```
I had to also create a `.venv` file inside the `foo/` folder with the content _"foo"_ to get it to work.

I can confirm, too, that your PR solves the error.
/cc @robbyrussell

PR #3167, opened 2014-09-23
Add alias for git add --verbose

### LINKED ISSUE #6793: Add `gav` as git-alias for `git add -v`
I find it helpful when Git tells me which files are added when I run `git add`. So I can be sure I have no unexpected files added to the staging area. However, [I don't see a git-alias](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L43) for that in this project.

Would it be okay to add a `gav` alias (i.e. `gav` = `ga -v` = `git add -v`), would you accept such a PR?

tushart91:
Add alias `gav='git add -v'`
fixes #6793

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -44,6 +44,7 @@ alias ga='git add'
 alias gaa='git add --all'
 alias gapa='git add --patch'
 alias gau='git add --update'
+alias gav='git add --verbose'
 alias gap='git apply'
 
 alias gb='git branch'


bittner:
Interestingly, `gaa = 'git add --all'` is [already included](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L44) in the aliases.

therelaxist:
Updated based on your comment.
All my requirements are already in upstream so just adding fix for your requirement.

bittner:
Nice, thank you!

I see the `git diff --staged` is implemented as `gdc` = `git diff --cached`. Would it be too much to ask to add also a `gds` = `git diff --staged` alias (which does the same)? What do you think?

bittner:
@tushart91 Can you add `gds = git diff --staged` as an official alias to `gdc`?

`git diff` supports both options, according to the [git-diff man page](https://linux.die.net/man/1/git-diff) ("`--staged is a synonym of --cached.`"), hence it should be acceptable to also support both short aliases.

mcornella:
@bittner use `gdc` or add the `gds` alias in your zshrc. Having an alias means using up another name in the namespace; having two is just wasteful.

bittner:
@mcornella :+1: 

Interesting. I thought to have seen `--staged` used more widely with `git diff` (that's why I cared). But the manpage seems to favor `--cached` anyway. Fine with me.

bittner:
BTW, `gdc` seems to have been removed (by error?) for some reason. It's [not in the aliases file](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L88-L93) anymore. Is there a reason for that, @mcornella?

PR #3178, opened 2014-09-26
jenv completion plugin

kubamarchwicki:
jenv is Java environment manager, and it is a clone of RVM for Java. jenv can manage parallel versions of Java development kits on any Unix based system. It provides a convenient command line interface for installing, switching, removing and listing Candidates (http://jenv.io/)

This plugin is a copy of the original jenv completion script from the original distribution, put within oh-my-zsh context


mcornella:
Thanks for your contribution, closing due to age and the completion version being old.

PR #3184, opened 2014-09-29
Add gerrit plugin

takac:
Use gerrit plugin to look up references based of change ids, update
patch sets and more.


mcornella:
Has no interest.

PR #3186, opened 2014-10-03
add warp plugin

hanguofeng:
add the [warp](https://github.com/jpalardy/warp) plugin


mcornella:
Rejected: enough plugins, and it doesn't seem to be that popular.

PR #3188, opened 2014-10-04
Fix completion for Laravel 5 (previously 4.3)

jeremyworboys:
Laravel changed the output out when listing commands.

This change makes the `laravel4` plugin compatible with both Laravel 4 and Laravel 5.


lucacri:
+1 , love this plugin!

ncanceill:
This needs to be rebased before merge.

mcornella:
Fixed in #3594

PR #3189, opened 2014-10-04
adding 'haxe' and 'haxelib' zsh plug-in and auto-completion

haysclark:
adding hxKraut's haxe and haxelib zsh plug-in and auto-completion to oh-my-zsh.
[http://haxe.org/](http://haxe.org/)


mcornella:
Has no interest.

PR #3191, opened 2014-10-04
added zirpot theme

cirpo:
Hi, I added a new theme
maybe it's useful for other ppl too

alt tag


robbyrussell:
Thanks for this, but we're not currently accepting any new themes per our README.

PR #3193, opened 2014-10-06
Update alias of for Atom

92bondstreet:
### Refactor at alias for Atom

*at* command is already use on MacOS for jobs handling.

I just change at to *a* alias.


mcornella:
Obsolete and not need, if you want to use `a` instead of `at` you can put `alias a='at'` in your zshrc file.
Please close.

PR #3197, opened 2014-10-07
Dropbox CLI completion

petrbel:
With this completion function one can use `dropbox <tab>` to see and select all possible options. According to [this source](http://askql.wordpress.com/2011/01/11/zsh-writing-own-completion/), completion function stored in `completions` is the right way it should be implemented, nevertheless maybe it should be plugin instead.

Anyway, this works fine and it's useful.

### DIFF
--- a/completions/_dropbox
+++ b/completions/_dropbox
@@ -1,0 +1,18 @@
+#compdef dropbox
+
+# Dropbox CLI completion
+ 
+_arguments "1:Command:((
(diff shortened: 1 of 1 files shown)


mcornella:
In its day this was useful, now there's dbxcli which is different and has [another completion file](https://github.com/dropbox/dbxcli/blob/master/contrib/zsh-completion/_dbxcli).

PR #3201, opened 2014-10-08
Fixed ssh-agent plugin when ssh-add is not in /usr/bin/ssh-add

carymrobbins:
On NixOS, for example, **ssh-add** might reside in **/run/current-system/sw/bin/ssh-add**.  Better that we try to grab it by name and allow for **PATH** resolution before resorting to an absolute path.


mcornella:
Fixed on master.

PR #3211, opened 2014-10-11
add ipban plugins - plugins to check attacks under iptables

ghost:
It's a little script can use some iptables functions. You can check how many ip attacks you, and look at the ip bannished. Add or remove manually an Ip.

How to use it ? simply type in your zsh shell

ipban count to check the number if ip banned
ipban list : list the ip banned

### DIFF
--- a/plugins/ipban/ipban.plugin.zsh
+++ b/plugins/ipban/ipban.plugin.zsh
@@ -1,0 +1,37 @@
+CSI="\033["
+CEND="${CSI}0m"
+CRED="${CSI}1;31m"
+CCYAN="${CSI}1;36m"
+ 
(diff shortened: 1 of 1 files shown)


mcornella:
Too much plugins come bundled with OMZ and this particular one doesn't seem to be useful to many people. It could also use some style cleanup, but that's not here nor there.

You can make a separate repo with your plugin and link to it from the [external plugins](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins) wiki page.

You can also add your plugin as a custom plugin. See [customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization) for instructions.

PR #3213, opened 2014-10-12
added support for GNU version of `date`

nikhgupta:
`emoji-clock` used `date -v` which is only supported on OSX version of `date`. Added support for GNU version, so that `emoji-clock` plugin can, now, be used everywhere.


apjanke:
This is pretty cool.

What do you folks think about making `emoji-clock()` take an optional argument in the form `HH:MM`, to be used instead of the current time? That would make testing easier.

(And using `local` to avoid leaking the variables to the caller's space.)

PR #3220, opened 2014-10-14
avoided unnecessary git call from `git_prompt_status`

nikhgupta:
The primary use of `git_prompt_status` is in prompts, which demands that the function should be as fast as possible. Taking an early exit from this function on non-git directories saves around `60ms` on my machine for every prompt that is rendered.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -79,7 +79,7 @@ function git_prompt_long_sha() {
 
 # Get the status of the working tree
 git_prompt_status() {
-  INDEX=$(command git status --porcelain -b 2> /dev/null)
+  INDEX=$(command git status --porcelain -b 2> /dev/null) || return 1
   STATUS=""
   if $(echo "$INDEX" | grep -E '^\?\? ' &> /dev/null); then
     STATUS="$ZSH_THEME_GIT_PROMPT_UNTRACKED$STATUS"


apjanke:
I'm working on a related PR to add a timeout to the `git status` calls, so slow repos don't cause the zsh prompt to hang in the status check functions (like in the case of a large repo directory tree, or repos on a network filesystem). It's #3725 (branch https://github.com/apjanke/oh-my-zsh/tree/git-status-timebox). I could merge in your short-circuit change so they can be tested and merged at the same time.

It looks like a lot of the time in `git_prompt_status` is spent on repeated shell-outs to `grep`. Parsing the status results using built-in `zsh` features could make it faster even in the case where you can't short-circuit. I took a stab at this (https://github.com/apjanke/oh-my-zsh/commit/f26f5014883deca0e56c94365994dab87e29d011) and the basic parser I came up with got it from around 45 ms down to 20 ms on the small repos I tested it on.

apjanke:
The difference between `grep` shell-outs and zsh builtins or the short-circuit is particularly pronounced on Cygwin on Windows, where starting a new process is more expensive. Saves nearly half a second. Though it's still a bit too slow for my taste at 110 ms when in a repo.

Here's a non-repo dir. This reflects your early-exit optimization, which saves *half a second* per prompt on Windows.

```
[@ in ~/local/opp]
$ time (_git_prompt_status_grep)

( _git_prompt_status_grep; )  0.10s user 0.36s system 85% cpu 0.546 total
[@ in ~/local/opp]
$ time (_git_prompt_status_zsh_parse)
( _git_prompt_status_zsh_parse; )  0.00s user 0.03s system 96% cpu 0.031 total
```

And here's a repo dir. About 400 ms speedup.

```
[@ in ~/local/opp/omz/oh-my-zsh on ⇄ git-status-timebox ±]
$ time (_git_prompt_status_grep)
M
( _git_prompt_status_grep; )  0.04s user 0.36s system 81% cpu 0.499 total
[@ in ~/local/opp/omz/oh-my-zsh on ⇄ git-status-timebox ±]
$ time (_git_prompt_status_zsh_parse)
M
( _git_prompt_status_zsh_parse; )  0.03s user 0.06s system 84% cpu 0.109 total
```

(This is on a repo on a local SSD, after multiple calls to each variant to warm up the cache, on a 2-core Windows 7 VM in VMware Fusion on a MacBook Pro.)

PR #3222, opened 2014-10-14
gitstatus.py incorrectly parse git status for some config

autra:
On my machine (Ubuntu 14.04 with zsh and git 2.0), there is no '#' at the start of each lines of git status.
But the regexp defined in gitstatus.py expect them to know the state of the repo. The regex should also accept output without these.

 to test, you can use dogenpunk theme for exemple.


mcornella:
Pushed a better update in #4205

PR #3228, opened 2014-10-15
Display AWS profile info in agnoster theme

ebr:
Adds display of AWS profile to the theme. follows patterns already established in the theme. Will only display profile info if one is loaded.


mcornella:
Obsolete.

Issue #3238, opened 2014-10-20
Shell prompts break and show RVM_PROJECT_PATH when used with rvm


clarkewd:
With the latest version of `rvm` it seems there is a bug with the `PROMPT` in some themes that causes the text `RVM_PROJECT_PATH` to show.

Please see https://github.com/wayneeseguin/rvm/issues/3091 for more details.

PR #3244, opened 2014-10-20
[#3238] swapping out %~ for %1/ in a few themes to fix rvm_project_path

ballPointPenguin:
I tested each of these theme fixes on my machine and they seem to be working fine. I only touched a handful of themes that use rvm. If this feels like an acceptable solution, it would be trivial to fix the rest.


jclif:
I saw this when I added 'setopt auto_name_dirs'

Issue #3208, opened 2014-10-11
Warp point should not contain whitespace while is just  contains "s"


krisdante:
I don't know why but on one of my OSX I am not able to add any warp point with "s" letter in name.
Seems that \s matches not only space but also "s".
This happens only on one of my computers. What can be miss-configured?

```
git:(master) wd add ssite
 * Warp point should not contain whitespace
```

mcornella:
Already solved in the current version, please close

PR #3249, opened 2014-10-21
fix whitespace regular expression

itsthatguy:
Referenced here: https://github.com/robbyrussell/oh-my-zsh/issues/3208


mcornella:
Already in the codebase, please close

PR #3251, opened 2014-10-22
Set git to quiet mode

sippsolutions:
Set git to quiet mode so you can use it in an automated environment installer


mcornella:
I decided against this because then the user doesn't know what's happening. For example, the clone may be really slow due to connectivity issues, and if the user can't see that they may either cancel (when there is really progress) or not cancel (when the download has really stopped).

As per using it in an automated environment I don't see why it would be an issue. Perhaps I'm mistaken.

PR #3258, opened 2014-10-23
Add the environment variable DISABLE_CMD_AUTO_TITLE 

exavatar:
This will be handy for those that only want to see pwd in their window and tab titles.


mcornella:
Already implemented

Issue #3093, opened 2014-09-03
rbenv plugin not working with homebrew


jweaver60:
I can't get the rbenv plugin to work with homebrew installed rbenv. It looks to be setting `$RBENV_ROOT` to `/usr/local/opt/rbenv` instead of `~/.rbenv`. When I enable the rbenv plugin then rbenv reports only the system ruby as installed and available.

bartdorsey:
Same problem here. If it's supposed to work with the brew installed rbenv then it should be /usr/local/var/rbenv or .rbenv.  It should never be /usr/local/opt/rbenv

caesarsol:
same problem.

I have a solution, but I'm not sure if it works for everyone.

the problem is in rbenv.plugin.zsh, line 13.
the script selects the RBENV_HOME dir based on the presence of the `bin` directory, but my

mcornella:
See #3808, +1 if it works for you.

caesarsol:
@mcornella reversing the order doesn't work for me, since the bin directory doesn't exist in my `~/.rbenv`...

caesarsol:
oops, sorry I haven't read the pull request from @mbologna which is exactly my patch.

mbologna:
Hey @jweaver60 @bartdorsey @caesarsol @mcornella: please check with #4384 if it solves your problem

tanelsuurhans:
Is there any hope for this getting merged in, I still see this issue present on OSX

PR #3275, opened 2014-10-29
rbenv plugin now works on linux too

mbologna:
Previously, rbenv plugin was testing if there is something in `~/.rbenv/bin`.
However, rbenv puts binaries in `~/.rbenv/versions` so testing the earlier mentioned directory will not work.
In addition, PATH is already handled by rbenv.

### DIFF
--- a/plugins/rbenv/rbenv.plugin.zsh
+++ b/plugins/rbenv/rbenv.plugin.zsh
@@ -13,13 +13,12 @@ if _homebrew-installed && _rbenv-from-homebrew-installed ; then
 fi
 
 for rbenvdir in "${rbenvdirs[@]}" ; do
-  if [ -d $rbenvdir/bin -a $FOUND_RBENV -eq 0 ] ; then
+  if [ -d $rbenvdir -a $FOUND_RBENV -eq 0 ] ; then
     FOUND_RBENV=1
     if [[ $RBENV_ROOT = '' ]]; then
       RBENV_ROOT=$rbenvdir
     fi
     export RBENV_ROOT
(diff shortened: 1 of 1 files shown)


mbologna:
@robbyrussell rebased on current master.

mcornella:
@ka8725 @Bounga @cap10morgan @sheerun @jbhannah @timsly @dipth @peterhoeg:
can you guys comment on this?

timsly:
but what if someone has old rbenv?

mbologna:
@timsly what do you mean? old version of rbenv install it in another directory?

timsly:
@mbologna you can ignore my comment, because i forgot what i mean there :-)

mcornella:
I found a reference to the `.rbenv/bin` on [the current master README](https://github.com/sstephenson/rbenv/blob/3b6faa85319d6ec42902a4498786ace29b80baaa/README.md#basic-github-checkout). Can you explain how did you install `rbenv` and how it handles the `PATH` exactly?

mbologna:
Looked at the current code of oh my zsh and it seems that my PR is no longer needed. Closing!

mcornella:
Thanks!

mcornella:
Hi @mbologna, let's open this one again because it seems it is the right one: https://github.com/robbyrussell/oh-my-zsh/issues/3093#issuecomment-142259502 https://github.com/robbyrussell/oh-my-zsh/pull/3808#issuecomment-139057226

mbologna:
Ok, I figured out why I opened this. Closing this one as it tries to solve two different issues.
Opening #4389 with a complete description and a solution.

PR #3278, opened 2014-10-31
command-not-found for OS X with brew

jansegre:
I have adapted the script at https://github.com/bfontaine/brew-command-not-found to work on zsh to enhance the __command-not-found__ plugin.

What it does:

```
$ ztelnet
The program 'ztelnet' is currently not installed. You can install it by typing:
  brew install zssh
```

And when there is no match:

```
$ asdf
zsh: command not found: asdf
```


mcornella:
Closed as obsolete as per https://github.com/robbyrussell/oh-my-zsh/pull/4537#issuecomment-152771547.

PR #3281, opened 2014-11-02
Fixed: sudo plugin interfering with vim-mode + cursor position #3141

conrado:
This patch alters the sudo-plugin's bindkey for sudo-command-line, which is eating up the `ESC` key used on vim-mode to enter command mode.

The *bug* only affects people using vim-mode; the *fix* changes the bindkey to `C-_`. 

Guess this needs design decision, for keeping compatibility between the two modes.

incorporates #3141
replaces #2522

### DIFF
--- a/plugins/sudo/sudo.plugin.zsh
+++ b/plugins/sudo/sudo.plugin.zsh
@@ -13,9 +13,8 @@
 # ------------------------------------------------------------------------------
 
 sudo-command-line() {
-[[ -z $BUFFER ]] && zle up-history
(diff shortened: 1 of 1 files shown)


mcornella:
I like `ESC`+`ESC` for the sudo plugin, it's easy to remember and fast. Let's postpone this one until we have more people to comment.

mcornella:
We'll merge #3141 instead. You can keep the bindkey change when that other one is merged.

robbyrussell:
Closed via #3141

PR #3290, opened 2014-11-04
add a prompt function in vi-mode plugin

metaphox:
When using vi key binding, it is useful to have both the right prompt and the left one showing some visual clue for the current status. So I added this function in the vi-mode plugin that can be used for the left prompt.


ikhthiandor:
How can I make this work with the zsh theme bira?

apjanke:
Looks like this conflicts/duplicates #2888. Could they be reconciled, or one of them chosen over the other?

PR #3293, opened 2014-11-06
This adds two functions to modify your PATH.

wontruefree:
Both functions check to see if the directories that will be inserted
into your PATH are currently on the system.  This is good for having
one zshrc file and deploying it to myltiple enviroments.

example:
```
prepend_path "${HOME}/bin"
postpend_path "/usr/local/bin/heroku"
```

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -75,7 +75,7 @@ function env_default() {
 }
 
 #
-# Prepends the PATH varible with the input path
+# Prepends the PATH variable with the input path
 #
 # Arguments:
 #    1. path - The path that will be prepened
@@ -87,7 +87,7 @@ function prepend_path() {
 }
 
 #
-# Prepends the PATH varible with the input path
+# Prepends the PATH variable with the input path
 #
 # Arguments:
 #    1. path - The path that will be prepened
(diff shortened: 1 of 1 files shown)


lib/functions.zsh
```diff
@@ -73,3 +73,27 @@ function env_default() {
...
+#    1. path - The path that will be prepened
+# Return value:
+#    0 if PATH is exported to ENV
+#
+function prepend_path() {
+    [[ -d "$1" ]] && export PATH="$1":$PATH && return 0
+}
+
+#
+# Prepends the PATH varible with the input path
```
apjanke: Should this be "Append" or "Postpend" instead of "Prepend"?

lib/functions.zsh
```diff
@@ -73,3 +73,27 @@ function env_default() {
     env | grep -q "^$1=" && return 0 
     export "$1=$2"       && return 3
 }
+
+#
+# Prepends the PATH varible with the input path
```
apjanke: "varible" misspelled.

javajax:
@apjanke Thanks for the feedback.
I updated the PR

PR #3297, opened 2014-11-06
Added mrm alias to rails

allcentury:
A common workflow in a rails database migrations is to migrate and check
rollback for up/down migration issues.  This helps keep the schema in
check regardless if any part of a migration fails.


robbyrussell:
⌛️ Given how slow we've been to consider this, am closing for now.

PR #3301, opened 2014-11-08
Background notifications on long-running commands

t413:
Get background notifications for long running commands! *Running a build? A deploy? Moving files?* Anything that takes more than 6 seconds (easily configurable) sends you a slick and unobtrusive system notification: 

screenshot 2014-11-08 14 15 12

It works with osx notification center (using terminal-notifier), ubuntu's native `notify-send`, and cygwin:

**Linux**
screenshot from 2014-11-07 15 58 36

**Windows**
screenshot from 2014-11-07 15 55 00

It supports having a custom message hook (in case you need it to say 'holy smokes batman' for example).

There are a few hacks to do the same thing-- often making heavy use of AppleScript and being very platform-spesefic. This plugin makes a point to run on all three major platforms.

### DIFF
--- a/plugins/bgnotify/README.md
+++ b/plugins/bgnotify/README.md
@@ -13,6 +13,7 @@ Just add bgnotify to your plugins list in your `.zshrc`
 - On OS X you'll need [terminal-notifer](https://github.com/alloy/terminal-notifier)
   * `brew install terminal-notifier` (or `gem install terminal-notifier`)
 - On ubuntu you're already all set!
+- On windows you can use [notifu](http://www.paralint.com/projects/notifu/) or the Cygwin Ports libnotify package
 
 
 ## Configuration
@@ -36,4 +37,3 @@ function bgnotify_formatted {
 plugins=(git bgnotify)  ## add to plugins list
 source $ZSH/oh-my-zsh.sh  ## existing source call
 ~~~
-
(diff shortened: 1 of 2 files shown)


mcornella:
@t413 thank you for this very well-crafted PR. Do you have a notion of another program that works on non-Unity Linux setups?

t413:
Thanks! Researching *notify-send* on linux shows that it's actually a relatively standard utility. Here's the [Arch linux wiki page](https://wiki.archlinux.org/index.php/Desktop_notifications) on the topic. Are you running a particular setup you'd like me to test against?

mcornella:
> Thanks! Researching notify-send on linux shows that it's actually a relatively standard utility. Here's the Arch linux wiki page on the topic. Are you running a particular setup you'd like me to test against?

Thank you for that initial push. I actually looked it up a little bit and found [this post](http://www.webupd8.org/2011/05/how-to-use-standard-gnome-notification.html) that got me in the right direction. TL;DR: on Debian you need to install `libnotify-bin`, as well as in Ubuntu (in which it's actually installed by default). It looks a little different on gnome-shell though:

Debian with Gnome 3

robbyrussell:
This looks really cool!

aL3xa:
+1

PR #3306, opened 2014-11-10
do not reset LC_CTYPE to LANG

shtouff:
under osx, where LANG is null, it leads to LC_CTYPE='C', which drop some issues which utf8 files.


mcornella:
Already in the codebase

PR #3310, opened 2014-11-11
Fixing the link to the install script

bhardin:


mcornella:
Dropped for the moment until https support is implemented in ohmyz.sh

PR #3311, opened 2014-11-11
Add auto-completion and aliases for Puppet:

marksmall:
The plugin auto-completes the puppet sub-commands when you
press <TAB>, this displays the sub-commands and a little help
text.  Once you select the sub-command, if you press <TAB>
again, then options are displayed, again with a little help
text.

Added some common aliases for puppet commands.


mcornella:
Has no interest.

PR #3313, opened 2014-11-11
Add puppet-lint aliases:

marksmall:
Added a file to contain puppet-lint aliases, only one exists
at the moment.


mcornella:
Has no interest.

Issue #2737, opened 2014-04-22
ssh completion broken in last version (takes always an entry in /etc/hosts if present)


ibc:
Since a few days (last update) my oh-my-zsh fails when typing the beginning of a Hostname value in .config/ssh. For example, in my .ssh/config I have:

```
Host home-dev-1
    hostname    X.X.X.X
    user              xxxxx
```

and in my /etc/hosts I have:

```
1.2.3.4   home-test
```

If I type in the terminal `ssh home` and press TAB then I get `home-test` (which is an entry in my /etc/hosts).

If there is no entries in my /etc/hosts beginning with "home" then it would get entries in my .ssh/config. But obviously this is not the expected behavior and it is a clear regression.

apjanke:
FYI, the `common-aliases` plugin will conflict with this because it also defines zsh host completion style; see #4257. Since it uses a narrower pattern in its `zstyle` definition, I think it will always take precedence over this, regardless of load order. So you'd need to make sure not to use them together.

meeech:
Still seeing this behaviour. Any movement or advice on this?  Different ticket to look at?

Someguy123:
**It's now 2019, this is still a problem.**

A somewhat related issue - I also experience a ghost hosts file problem (not sure if that's been reported): 

On my Macbook I once had a hosts file entry for `someserver.dev` (name adjusted for privacy), while on my newer iMac (some various dot files copied over from my Macbook) that entry does not exist in the hosts file, not even as a commented out line; but still when I type `some` and press tab, it'll immediately complete `someserver.dev` regardless of how many other servers I have starting with `some` in .ssh/config.

I've tried to remove any instances of `someserver.dev` from my zsh history file, but still no luck.

To workaround the "ghost hosts file" problem, I've ended up having to just add additional aliases for those servers that don't start with `some`, but still my old habits are hard to break, and at least once every few days I'll run into this completion problem. 

For host entries that first existed on my iMac, simply commenting them out, and then opening a new terminal session is enough to restore my tab completion, but not exactly a "solution"...

**Is there any _real_ solution for this?** This issue has been open for almost **_5 years_** now, and there's not so much as a workaround.

aurelijusrozenas:
Hello 2021, still a problem? :/
I have some hosts in .ssh/config and in hosts, autocomplete always suggest me just from hosts file. Autocompletion on bash works as desired: suggest all matches from hosts and ssh config.  
Still no solution for this?

[...]

PR #3321, opened 2014-11-13
[tmux-cssh plugin] Initial version

moul:


aimxhaisse:
:+1:

moul:
Is there a way I can help to merge this pull-request ?
Thank you

moul:
up

davizucon:
@moul thank you, I tried, and worked! Its auto complete from hosts file, there is some way to auto-complete from ~/.ssh/config? 
  
I think its ok reading config file from ssh command, because  tmux-cssh depends directly from ssh, and if we config a host (user,  alias) on  ~/.ssh/config, we can use as [host1 host2 host3 host4]  parameter in tmux-cssh command.

Thanks for your help, make my day more productive and fun!

apjanke:
As a hack, you could piggyback on Zsh's own ssh-aware host completion by changing `':hosts:_hosts'` to `':hosts:_ssh_hosts'`. Then it would pick up `~/.ssh/config` entries (sometimes; see #2737).

You'd need to call the `_ssh` function once during the plugin's initialization to force it to autoload so the `_ssh_hosts()` definition is available; it's a helper function defined inside `_ssh`. It's not documented so it probably isn't public or stable, but it seems to be around in all the recent `zsh` versions, so it'll probably work.

davizucon:
Hi @apjanke, thanks, worked very well.
Now I am trying to add auto-complete to several hosts, like: tmux-cssh host1 host2 host3 so on.

apjanke:
@davizucon: You might need to swap the `hosts` and `subcommands` around in the completion, or combine them using `_alternative`. The `*` indicates that it applies to all arguments; without it, it just applies to the next argument.

Have a look at [this tutorial on writing zsh completion functions](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org). I found it a lot more of a gentle introduction than the main zsh documentation, which is more of a complete reference.

moul:
@apjanke what do you think about merging this (now 1-year old PR) that is actually working well, so we can discuss about improvements in dedicated issues ?

I'm personally using this plugin on a daily-base without encountering any issue

apjanke:
Sounds fine to me. I don't have commit rights, though; only @robbyrussell and @mcornella do.

mcornella:
Sorry for the delay @moul. We'll want a README as well though.

PR #3329, opened 2014-11-18
Improved ant completion

capeterson:
Fix regenerating .ant_targets on most tab completions (which made ant support sloooow). 
Also make ant do the work of enumerating targets which fixes edge cases.

Note that this change does make it so that "internal" targets (those with no description) are no longer shown, which could possibly be considered a breaking change. I'd argue this is good though, as `ant -projecthelp` also doesn't output these.

closes #1827

### DIFF
--- a/plugins/ant/ant.plugin.zsh
+++ b/plugins/ant/ant.plugin.zsh
@@ -1,6 +1,6 @@
 _ant_does_target_list_need_generating () {
   [ ! -f .ant_targets ] && return 0;
-  [ .ant_targets -nt build.xml ] && return 0;
+  [ build.xml -nt .ant_targets ] && return 0;
   return 1;
 }
 
--- a/plugins/ant/ant.plugin.zsh
+++ b/plugins/ant/ant.plugin.zsh
@@ -9,7 +9,7 @@ _ant () {
     if _ant_does_target_list_need_generating; then
     	ant -p | awk -F " " 'NR > 5 { print lastTarget }{lastTarget = $1}' > .ant_targets
     fi
-    compadd `cat .ant_targets`
+    compadd -- `cat .ant_targets`
   fi
 }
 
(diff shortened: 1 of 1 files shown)


capeterson:
Just a quick check in to see if there's anything outstanding on this?

gogoprog:
I agree with this PR and mine can be closed ( #1827 )

mcornella:
Nothing outstanding really, just a little bit of delay in the merging process. I'll gladly contact Robby and we'll surely have it merged soon
/cc @robbyrussell

arloherrine:
I made this change on my local system an I'm getting errors when I try to complete:
`_ant:compadd:5: bad option: -c`
Maybe I'm missing something, but I think it might have to be 
```
compadd -- `cat .ant_targets`
```
When I make that change it works great for me.

mcornella:
It's quite possible that you have a target that starts with a dash. Can you post the output of `cat .ant_targets` ?

capeterson:
Sure, happy to. Since it was never reading `.ant_targets` before this is a "new" issue. Ant targets with a leading dash are a common way of denoting private targets anyways, so it really should work!

mcornella:
@capeterson it doesn't work because of the way zsh runs that line: first it runs the `cat` part, then passes its output to `compadd`. So if `.ant_targets` contains, for example, a `-compile` target, zsh will run `compadd -compile`. `-compile` is obviously not a valid `compadd` option, so it will complain. 

Those double dashes (`--`) tell `compadd` to stop looking for an option and treat the rest of the command line as plain text. It's standard behavior, actually, so the same trick will work in other commands (like `git`, for example).

Hopefully I cleared up things.

mcornella:
Hey we'll get this one merged soon, can you apply @arloherrine's solution?

capeterson:
Done, looks to be working properly from some quick testing, thanks @arloherrine!

mcornella:
Cool, though you've polluted the PR. Let's fix it: 

1. `git reset --hard upstream/master`, where `upstream/master` is the reference to the master branch of the robbyrussell remote (if you've named it other than upstream, then use that).
2. `git cherry-pick 926d8f8a fae545cb 07f3ed6c`
3. `git push --force`

capeterson:
Sorry - can't believe I did that. Fixed now.

arloherrine:
My pleasure!

PR #3337, opened 2014-11-22
respect the convention when loading themes from $ZSH_CUSTOM

chazzbg:
Respect the naming convention and search for theme in $ZSH_CUSTOM/themes/ also


apjanke:
Oops. I forgot I had seen this PR earlier and duplicated its functionality in #3587. I also rewrote it to not have `pwd` or leaked variable side effects.

PR #3347, opened 2014-11-26
Do not use GREP_OPTIONS to set grep parameters

mcornella:
## STATUS: will merge #3341 for now

From grep version 2.21, using environment variable `GREP_OPTIONS` to set common parameters to grep calls is deprecated.

As an alternative, we use a custom function to pass in our defined `GREP_OPTIONS`, while at the same time preserving any aliases that called grep.

If we choose the alias way as suggested in the grep documentation, we'll overwrite any already defined grep aliases, or get overwritten by any subsequently defined user aliases.

See discussion in #3341.

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -3,6 +3,7 @@
 # Examples: http://rubyurl.com/ZXv
 #
 
+# color grep output
 GREP_OPTIONS="--color=auto"
 
 # avoid VCS folders (if the necessary grep flags are available)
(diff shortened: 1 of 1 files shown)


nichochar:
This new alias workaround has stopped me from being able to set custom `--exclude=*.pyc` and I get an error:
`zsh: no matches found: --exclude=*.pyc`

xuhdev:
@nichochar have you tried `'--exclude=*.pyc'`

zmwangx:
@nichochar As suggested by @xuhdev, you have to quote the globbing pattern, otherwise it is expanded by the shell, and unless you unset the zsh option `nomatch`, you'll get an error whenever there's no filename matching the globbing pattern (moreover, even if there are filenames matching the pattern, the behavior is not what you want).

PR #3359, opened 2014-12-01
Don't clobber $HISTSIZE and $SAVEHIST

unixorn:
The existing code was clobbering $HISTSIZE and $SAVEHIST, switch to only setting them if they're unset


CyberShadow:
> I think the best approach is defining those variables after oh-my-zsh is loaded.

That doesn't work, if you want to increase OMZ's limit:

```
$ history | wc -l
26640

$ HISTSIZE=100

$ HISTSIZE=10000

$ history | wc -l
100
```

OMZ's `history.zsh` truncates the history when it sets `HISTSIZE`, and currently there is no way to override it.

PR #3373, opened 2014-12-04
use zsh completion cache for costly python module completions

heptal:
zsh completion is very slow when completing on Python modules:

`python -m <TAB>` or `python -m <partial_match><TAB>` invokes

``` python
from pydoc import help; help("modules")
```

on every `<tab>` which is quite slow and prone to spewing messages to stderr. Using the zsh completion cache functions speeds it up immensely.


mcornella:
The best way forward is deleting our python completion and letting the official one do its job.

PR #3374, opened 2014-12-04
[RDY] Remove (outdated) compdefs for `git branch` and `git remote`

blueyed:
I was wondering why the completion for `git branch` was outdated (it
knows about `--set-upstream` only, and not `-u` etc), and found these
files.

I have not compared them to Zsh's upstream version, but in case there
are any additional features/fixes, they should get forwarded to them.

@Stibbons 
Any objections? - asking you, since you've added them.


ncanceill:
Sorry it has been five months, no idea how I missed this one. Thanks again @blueyed.

blueyed:
It won't be closed by your commit message from there?
If so, please close it as a duplicate.

PR #3376, opened 2014-12-05
Update regex to make it works

amenophis:
just one space instead two


mcornella:
Implemented already. Please close

PR #3378, opened 2014-12-05
transfer.sh support

clawfire:
Implement a quick script to use the services from [transfer.sh](http://transfert.sh) that let user upload up to 5go for 14 days and share easily files.

The script is made by [nl5887](https://github.com/nl5887).

### DIFF
--- a/plugins/transfer/transfer.plugin.zsh
+++ b/plugins/transfer/transfer.plugin.zsh
@@ -1,0 +1,22 @@
+transfer() {
+    if [ $# -eq 0 ]; then
+              echo "No arguments specified. Usage:"
+              echo "transfer /tmp/test.md"
+              echo "cat /tmp/test.md | transfer test.md"
+              return 1
+    fi
+ 
+    # write to output to tmpfile because of progress bar
+    tmpfile=$( mktemp -t transferXXX );
+ 
+    if tty -s; then
+        basefile=$(basename "$1" | sed -e 's/[^a-zA-Z0-9._-]/-/g');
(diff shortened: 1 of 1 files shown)


plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,22 @@
+transfer() {
+    if [ $# -eq 0 ]; then
+              echo "No arguments specified. Usage:"
+              echo "transfer /tmp/test.md"
+              echo "cat /tmp/test.md | transfer test.md"
+              return 1
+    fi
+ 
+    # write to output to tmpfile because of progress bar
+    tmpfile=$( mktemp -t transferXXX );
```
apjanke: Could `$tmpfile` and `$basefile` be made `local` to avoid polluting the caller's variable space?

apjanke:
Looks like there's an updated version of this script at https://gist.github.com/nl5887/a511f172d3fb3cd0e42d with support for directory uploads and more error handling.

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,22 @@
...
+ 
+    if tty -s; then
+        basefile=$(basename "$1" | sed -e 's/[^a-zA-Z0-9._-]/-/g');
+        curl --progress-bar --upload-file "$1" "https://transfer.sh/$basefile" >> $tmpfile;
+    else
+        curl --progress-bar --upload-file "-" "https://transfer.sh/$1"
```
apjanke: This branch still uses `--progress-bar` but isn't writing to `$tmpfile`. Intentional?

PR #3383, opened 2014-12-07
Switch the directory with the vim key-binding

MartyJiang:


mcornella:
You have commited other modifications. Please close and submit another PR (use a dedicated branch other than `master`)

PR #3400, opened 2014-12-13
Print current chroot (created by schroot)

fshp:
Print only if chroot active.
Print current chroot


robbyrussell:
Yeah, that might be a better approach

PR #3405, opened 2014-12-15
Modified battery gauge to allow custom chars

jackspirou:
This request adds a couple lines to the battery gauge function allowing theme creators to add their own unicode characters for full and empty.  This is useful if you want squares instead of triangles, like I did for the Jarvis theme.  

It is important to note that this change doesn't affect any other themes.  The default battery gauge character values are still triangles. 

I removed the theme from this pull request after seeing the ask in the project README to do so.


jackspirou:
This PR is old. Closing it.

PR #2134, opened 2013-10-02
-0 ↦ -1

Profpatsch:
-0 wont do much, and indeed, it doesn’t. ;)

PR #3413, opened 2014-12-16
dircycle plugin: clean and fix logic once and for all

### LINKED ISSUE #1566: dircycle plugin doesn't work on OS X
I'm not sure if I'm doing something dumb, but the dircycle plugin simply doesn't seem to work on OS X; even after enabled, Ctrl+Shift+Left/Right doesn't do anything. I assume it should?

### LINKED ISSUE #2101: dircycle plugin works incorrect
My problem is in pushd number arguments.
If I change dircycle commands from this:
```bash
pushd -q +1 # Left
pushd -q -0 # Right
```
to this:
```bash
pushd -q -1 # Left
pushd -q +0 # Right
```
then plugin works well.
Before I could use dircycle without any correction, so I think it is only my problem.

mcornella:
#### Description

This pull request fixes the error introduced in #3395, which changes `pushd -q -0` to `pushd -q +0`, which doesn't do anything unless you have the `pushd_minus` option set. But, if that's the case, then `pushd -q +1` doesn't work correctly. 

To that end I standardised that option since it seems to change between different zsh setups. You can read more in the [`PUSHD_MINUS` section of the `zshoptions` manpage](http://zsh.sourceforge.net/Doc/Release/Options.html#index-PUSHDMINUS).

I also removed the evals and cleaned up the plugin a little bit. 

##### 1. Fast switch
I added in commit d63120501ed924b801774359fb3e8be902cbd0e5 the hability to switch between directories and redraw the prompt without using a new line to run the `pushd` command. 
Here are a couple of screencasts showcasing the difference:
  - [Current switch (**old** version)](https://asciinema.org/a/15264): 
dircycle slow-switch

  - [Fast switch (**new** version)](https://asciinema.org/a/15265): 
dircycle fast-switch

**Please comment what you think about this new behavior**.

##### 2. Left/right switch direction
Finally the chosen behavior for left/right is (see https://github.com/robbyrussell/oh-my-zsh/pull/3413#issuecomment-67987390): 
  
  - `CTRL`+`SHIFT`+`LEFT `: moves to last visited directory
  - `CTRL`+`SHIFT`+`RIGHT`: moves to next visited directory

the previous behavior seemed a little odd an unintuitive to me:
  
  - `CTRL`+`SHIFT`+`LEFT `: moves to directory on the left as appears in `dirs` output
  - `CTRL`+`SHIFT`+`RIGHT`: moves to directory on the right
  
**Please comment which direction you think is better**.

----
Close #1566
Close #2134
Close #2101
Close #2474

### DIFF
--- a/plugins/dircycle/dircycle.plugin.zsh
+++ b/plugins/dircycle/dircycle.plugin.zsh
@@ -1,10 +1,27 @@
-##
-# dircycle plugin: enables cycling through the directory
-# stack using Ctrl+Shift+Left/Right
+# enables cycling through the directory stack using
+# Ctrl+Shift+Left/Right
+#
+# left/right direction follows the order in which directories
+# were visited, like left/right arrows do in a browser
 
-eval "insert-cycledleft () { zle push-line; LBUFFER='pushd -q +1'; zle accept-line }"
+# NO_PUSHD_MINUS syntax:
+#  pushd +N: start counting from left of `dirs' output
+#  pushd -N: start counting from right of `dirs' output
+setopt nopushdminus
+
+insert-cycledleft () {
+	zle push-line
+	LBUFFER='pushd -q +1'
+	zle accept-line
+}
 zle -N insert-cycledleft
-bindkey "\e[1;6D" insert-cycledleft
-eval "insert-cycledright () { zle push-line; LBUFFER='pushd -q +0'; zle accept-line }"
+
+insert-cycledright () {
+	zle push-line
+	LBUFFER='pushd -q -0'
+	zle accept-line
+}
 zle -N insert-cycledright
+
+bindkey "\e[1;6D" insert-cycledleft
 bindkey "\e[1;6C" insert-cycledright
--- a/plugins/dircycle/dircycle.plugin.zsh
+++ b/plugins/dircycle/dircycle.plugin.zsh
@@ -7,15 +7,20 @@
 # NO_PUSHD_MINUS syntax:
 #  pushd +N: start counting from left of `dirs' output
 #  pushd -N: start counting from right of `dirs' output
-setopt nopushdminus
 
 insert-cycledleft () {
+	emulate -L zsh
+	setopt nopushdminus
+
 	builtin pushd -q +1 &>/dev/null || true
 	zle reset-prompt
 }
 zle -N insert-cycledleft
 
 insert-cycledright () {
+	emulate -L zsh
+	setopt nopushdminus
+
 	builtin pushd -q -0 &>/dev/null || true
 	zle reset-prompt
 }
--- a/plugins/dircycle/dircycle.plugin.zsh
+++ b/plugins/dircycle/dircycle.plugin.zsh
@@ -10,16 +10,14 @@
 setopt nopushdminus
 
 insert-cycledleft () {
-	zle push-line
-	LBUFFER='pushd -q +1'
-	zle accept-line
+	builtin pushd -q +1 &>/dev/null || true
+	zle reset-prompt
 }
 zle -N insert-cycledleft
 
 insert-cycledright () {
-	zle push-line
-	LBUFFER='pushd -q -0'
-	zle accept-line
+	builtin pushd -q -0 &>/dev/null || true
+	zle reset-prompt
 }
 zle -N insert-cycledright
 
--- a/plugins/dircycle/dircycle.plugin.zsh
+++ b/plugins/dircycle/dircycle.plugin.zsh
@@ -26,5 +26,12 @@ insert-cycledright () {
 }
 zle -N insert-cycledright
 
-bindkey "\e[1;6D" insert-cycledleft
-bindkey "\e[1;6C" insert-cycledright
+
+# add key bindings for iTerm2
+if [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
+	bindkey "^[[1;6D" insert-cycledleft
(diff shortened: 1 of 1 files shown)


mcornella:
I have added the iTerm2 modification to the pull-request branch. I believe it is ready to merge, with at least 2 testers. Let's see if I can get someone else to chime in though.

MyChaOS87:
Checked it out on Linux, works perfectly, would be great if it will be merged soon.

mcornella:
Thanks for confirming, @MyChaOS87. Im cc-ing a bunch of other dircycle users: @dpaola2 @jonbrenner.

apjanke:
Hi folks.

I don't think this iTerm-specific code is correct or necessary.

```
# add key bindings for iTerm2
if [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
	bindkey "^[[1;6D" insert-cycledleft
	bindkey "^[[1;6C" insert-cycledright
else
	bindkey "\e[1;6D" insert-cycledleft
	bindkey "\e[1;6C" insert-cycledright
fi
```

The two branches of that `if` statement do exactly the same thing. In a `bindkey` call, `\e` and `^[` both mean `ESC`; they're just alternate ways of writing it. You can test it by using both in a row.

```
$ bindkey "^[[1;6D" insert-cycledleft
$ bindkey "\e[1;6D" backward-word
$ bindkey | grep 6D
"^[[1;6D" backward-word
```

I think the issue isn't that iTerm uses different escape sequences from other terminals for Ctrl-Shift-arrows, but that by default, iTerm doesn't send *anything* for Ctrl-Shift-arrows. (Maybe because it's a Mac idiom that Ctrl-Shift-stuff does things at the OS/window management level, not inside a terminal.) The OS X Terminal.app has a similar issue, in that Ctrl-Shift-arrow just sends the same thing as a the normal, unmodified arrow key.

This is a configuration thing that needs to be done at the level of iTerm or Terminal.app if you want to use Ctrl-Shift-arrow inside `zsh`. I don't think we can detect or change those settings from within `zsh` (well, without resorting to AppleScript in the OMZ startup sequence, which is just getting ridiculous).

Having the iTerm.app test in the code might be causing confusion or false confidence. Maybe it should be removed, since it's not doing anything?

The escape sequences here match what `xterm` produces. Other terminal emulators, in particular `rxvt`/`urxvt`, emit different sequences for Ctrl-Shift-arrow. Unfortunately, AFAIK `terminfo` doesn't provide info on Ctrl-Shift-arrow as a capability, so we can't use it for portability. Maybe this is something that could be hacked in to `lib/termsupport.zsh` as an extension to `$terminfo`, so at least within OMZ we have access to portable mappings for at least the most popular terminals. (Or maybe there's a third party database for this that I'm not aware of?)

@andrewferrier – are you sure you never did customizations for your iTerm2? Specifically, in the Keys section of your profile – did you maybe "Load Preset..." with something like "xterm defaults"? That will load a bunch of extra key settings that include all the Ctrl-Shift-arrows, with escape sequences that match `xterm`'s. But this isn't the default for iTerm2's initial setting or new profiles it creates. (I don't know why, especially since the default `$TERM` it reports is `xterm-256color`.)

This is the default set of Keys for a new profile in iTerm2 2.1.1 for me on OS X 10.9.5. (Curiously, it's not available as a preset either; you can only get it with a new profile.)

screen shot 2015-07-06 at 10 33 44 pm

mcornella:
You're absolutely right @apjanke, `'\e'` is the same as `'^['`. From ["Function keys and so on" (zsh docs)](http://zsh.sourceforge.net/Guide/zshguide04.html#fkeys):

> If you move the cursor back over that `^[`, you'll find it's a single character --- you can position the cursor over the `^`, but not the `[`. This is zsh's way of inserting a real, live escape character into the line. 
> ...
> You can also use the two characters `^[`, which is just another way of entering an escape key. However, the kosher thing to do is to turn it into `\e`.

And if I try it running both commands the keyboard shortcuts still work ok.

I think the problem was that vi-mode was clobbering the key bindings, as he explained in https://github.com/robbyrussell/oh-my-zsh/pull/3413#issuecomment-68036042. I overlooked that fact and kept the unnecessary if as a consequence. 
So I'm :+1: for removing that piece of logic and I'll use this opportunity to also a README to the dircycle plugin.

apjanke:
Cool.

You might also be able to fix the load-order dependency wrt `vi-mode` now. I'm looking in to it for `history-substring-search` over at #3913. It looks like using `bindkey -m` to bind keys in specific keymaps (instead of the current keymap), and adding bindings for both `emacs` and `viins` maps (and possibly `vicmd`, too, depending on preference), could take care of that issue.

plugins/dircycle/dircycle.plugin.zsh
```diff
@@ -1,10 +1,37 @@
...
+
+	builtin pushd -q -0 &>/dev/null || true
+	zle reset-prompt
+}
 zle -N insert-cycledright
-bindkey "\e[1;6C" insert-cycledright
+
+
+# add key bindings for iTerm2
+if [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
```
apjanke: These two binding sets are redundant: `^[` and `\e` mean the same thing in bindkey; they're both "escape". Submitted #4623 to collapse this and add some notes.

apjanke:
I put in PR #4623 to de-redundantize these. (Ran in to it today, and submitted the PR before I looked back and saw I'd already commented on this, sorry.)

Issue #789, opened 2011-12-30
when typing history -r the result is ➜ ~ history -r fc: event not found: -r


greg0ire:
I created this bug report because of this comment :
https://github.com/robbyrussell/oh-my-zsh/issues/739#issuecomment-3283206

gkop:
Not sure why this is closed. Anyhow, I found it best to just forget about the history alias and drop to fc:

    $ fc -lr 1

Or for my purposes:

    $ fc -li 1

nsarafa:
One workaround is to modify your ~/.zsh_history file directly and save your changes. Restart your terminal, and you should find your history has been modified per your changes to your ~/.zsh_history file

Issue #739, opened 2011-12-10
when typing history -c  the result is ➜  ~  history -c fc: event not found: -c


googya:
when typing history -c  the result is 
➜  ~  history -c fc: event not found: -c

I donot understand the result . Is my oh-my-zsh setting wrong? 
I install the command with ➜  ~  wget --no-check-certificate https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh -O - | sh  and change nothing...

It may be a bug!!

mcornella:
Hi everyone, I pushed a solution at #3422, hopefully you can take a look and +1 if it works well

miguelpalazzo:
I know this is very old, but had the same issue with `history -c` and login to `bash`, typing it and exit was more than enough for me.

crsuarez:
you could to this as a workaround.
```
[csuarez@centosdev:~] % truncate -s 0 $HISTFILE 
[csuarez@centosdev:~] % echo $HISTFILE         
/home/csuarez/.zsh_history
[csuarez@centosdev:~] % cat /home/csuarez/.zsh_history
: 1443547857:0;echo $HISTFILE
: 1443547861:0;cat /home/csuarez/.zsh_history
[csuarez@centosdev:~] % 

```

atooy:
fc: event not found :-c

atooy:
fc: event not found: -c            and i can't delete the file .zsh_history...(didn't use sudo permission, afraid there will be something wrong if i delete this file.)

mcornella:
You can safely delete the file, it will be recreated next time you start the terminal.

atooy:
Appreciate ur reply! It did recreated and the history was cleared!

nsarafa:
this bug still persists. anybody find a workaround?

atooy:
Just delete the file .zsh_history and the history shall be gone.    (not so much a bug as a difference between bash and zsh...i think)

KANGOD:
OMG, from Dec 11, 2011 to Nov 6, 2016, 5 years but the problem is still here!
At least, could you please provide equivalent arguments for `fc`? Something like `fc --reset` for `history --reset`, `fc --clear` for `history --clear`.
Must we deal with the .zsh_history file? This may be a solution for `history -c`, but obviously not for `history -r`.

mcornella:
[...]

Issue #1010, opened 2012-03-12
Error when use history -d


favadi:
I cannot use any of history option like -d, -D, -E, -f.
I always get this message.

```
fc: event not found: -d
```

WilsonLopezRomero:
Where I can find that file nowadays ?

mcornella:
It's in `lib/history.zsh` but you shouldn't change it there. Instead, you should redefine the history alias in your zshrc file. 

There's also a PR open to fix it, see #3422. If you test it and you find that `history -d` please +1 the PR. Thanks!

PR #3422, opened 2014-12-18
Enhance history alias

mcornella:
Allow passing in other `fc` arguments (complain in #789) and provides a simple workaround to the lack of `history -c`-like command (as seen in #739).

I also reorganized the script a little bit and removed the `append_history` ~~and `hist_verify` which seem to not be good defaults.~~ I've put it back in because at this point that's the expected behavior. Use `unsetopt hist_verify` in your zshrc files if you want to disable it.

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -32,7 +32,6 @@ HISTSIZE=10000
 SAVEHIST=10000
 
 ## History command configuration
-setopt append_history         # append history to HISTFILE on session exit
 setopt extended_history       # record timestamp of command in HISTFILE
 setopt hist_expire_dups_first # delete duplicates first when HISTFILE size exceeds HISTSIZE
 setopt hist_ignore_dups       # ignore duplicated commands history list
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -6,7 +6,7 @@ function omz_history {
     echo -n >| "$HISTFILE"
     echo >&2 History file deleted. Reload the session to see its effects.
   else
-    fc $@ -l 1
+    builtin fc "$@" -l 1
   fi
 }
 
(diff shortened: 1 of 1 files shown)


mcornella:
TODO: Test whether #1010 works.

mcornella:
Tested `history -d` as per #1010 and it works ok :+1:

Shir0kamii:
Hello !

I don't know if it's specific to my setup, but simply changing aliases of history to "fc -l 1" to "fc -l" is enough to get all functionnalities of fc and is what the [documentation](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html) seems to say (is it official though?)

I get all the functionnalities you're implementing in this PR for almost no change, I think it's worth trying, isn't it ? I don't have another computer so could someone else try it to be sure it's not specific to my environment because of some manipulations ?

Hopes it helps!

neclimdul:
WFM

mcornella:
@Shir0kamii said:
> I don't know if it's specific to my setup, but simply changing aliases of history to "fc -l 1" to "fc -l" is enough to get all functionnalities of fc and is what the documentation seems to say (is it official though?)

That only shows the 16 last entries of command history. The reason behind `fc -l 1` is to show all of it. This has been this way since the start of OMZ.

Regardless, with this PR you'll be able to change that as well by doing `history -l` if you prefer (same as `fc -l`, but whatever).

mcornella:
TODO: check #5694 and #5573 before merging.

ajasmin:
`history -5` doesn't work with this.

mcornella:
> history -5 doesn't work with this.

I can't find a way to make it work. It doesn't work with the old version though, so I don't see it as a blocker. IMO this PR can be merged after 10 +1s.

dikiaap:
I have a problem here after updating. When I type any command in Terminal, it says `omz_history:fc:15: event not found: -l`

PR #3426, opened 2014-12-20
termsupport: Move DISABLE_AUTO_TITLE check to hooks

apjanke:
This moves the DISABLE_AUTO_TITLE check from title() itself to the precmd/preexec hook functions that call it. It also allows single-argument use by defaulting the long window title ($2) to the short icon title ($1).

I'd like to do this because it would allow the title() function to be used directly by the user or other callers. In my workflow, I like to set the terminal tab title manually, and had my own settitle() function set up to do that. The OMZ title() function is already out there in the global function namespace, so it makes more sense to do that.

This arrangement of the check in the hooks was discussed in https://github.com/robbyrussell/oh-my-zsh/pull/266 but dismissed as DRY/not needed for performance. But I think it's worth reconsidering: this isn't for performance, but an interface/functionality change that makes title() more useful.

Also reformatted the comments to match comments in other OMZ lib/*.zsh files.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -1,9 +1,16 @@
-#usage: title short_tab_title looooooooooooooooooooooggggggg_windows_title
-#http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#ss3.1
-#Fully support screen, iterm, and probably most modern xterm and rxvt
-#Limited support for Apple Terminal (Terminal can't set window or tab separately)
+# Set terminal window and tab/icon title
+#
+# usage: title short_tab_title [long_window_title]
+#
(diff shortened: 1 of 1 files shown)


mcornella:
Hey can you squash this PR to keep just the relevant commits? Once this one or the other are merged then we'll rebase.
/cc @robbyrussell

mcornella:
Robby just merged the other PR, you can rebase to keep only commit 7a955ed906b967c6aac68a1561ce2283e198481c.

apjanke:
Sorry I didn't get to squashing this before it got merged – I'm a Git newbie and still figuring out how that works.

mcornella:
Don't worry Andrew, we're all noobs at something (I was just half a year ago!). What's important is this made it in :smile:

PR #3427, opened 2014-12-20
Add aliases for DNF package manager.

ppapadeas:
DNF is the new YUM replacement as the default package manager in Fedora moving forward.


sebastianbergmann:
This is the same as #2452 with the only difference being that the common alias prefix is just `dn` instead of `dnf`.

mcornella:

PR #3428, opened 2014-12-21
add new meteor commands

acemtp:
remove 'add' since there're more than 3000 packages, we cannot display them


mcornella:
Won't merge: the completion is optional, some people will want it and some won't. For those that don't, they just need to not press tab after `meteor add`.

PR #3429, opened 2014-12-21
termsupport: add chpwd hook for setting pwd in Apple Terminal.app

apjanke:
In termsupport, add a chpwd hook that notifies Apple Terminal.app of current working directory changes using the undocumented '\e]7;%s\a' sequence, instead of the normal window/icon title controls. 

Fixes https://github.com/robbyrussell/oh-my-zsh/issues/2646, letting new tabs always open in the correct directory.

This is done using chpwd_functions instead of the prompt/preexec hooks and is intentionally not controlled by DISABLE_AUTO_TITLE, so the directory is properly updated even when auto-titling is off (e.g. when user is setting window and icon titles themself).


robbyrussell:
Thanks team!

PR #3443, opened 2014-12-27
Refactor and improve web-search plugin

### LINKED ISSUE #3441: stderr redirection in web-search plugin
Issue: web_search function creating file `1` each time it called. Incorrect redirection of stderr.
found on:175b4a807383530aa75145b5b6fdedb3ce1f11f2
file: .oh-my-zsh/plugins/web-search/web-search.plugin.zsh
line 40:
-- nohup $open_cmd "$url" >/dev/null 2&>1
++ nohup $open_cmd "$url" >/dev/null 2>&1

mcornella:
### STATUS: ready!
----
This pull request implements simpler logic that enables adding more search engines in the future, while mantaining ease-of-use to define aliases.

It also implements the following very solicited features:

- Silence nohup:
  Fix #1970, fix #3097, fix #3183, fix #3218, fix #3302, fix #3441.

- Add Yandex support:
  Fix #1845, fix #3119.

- Add Cygwin support to open command:
  Fix #2831, fix #3103.

In total, **10 issues** can be closed with this PR.
Feedback is greatly appreciated.

### DIFF
--- a/plugins/web-search/web-search.plugin.zsh
+++ b/plugins/web-search/web-search.plugin.zsh
@@ -10,6 +10,7 @@ function web_search() {
     bing        "https://www.bing.com/search?q="
     yahoo       "https://www.yahoo.com/search?q="
     duckduckgo  "https://www.duckduckgo.com/?q="
+    yandex      "http://yandex.ru/yandsearch?text="
   )
 
   # define the open command
@@ -46,6 +47,8 @@ alias bing='web_search bing'
 alias google='web_search google'
 alias yahoo='web_search yahoo'
 alias ddg='web_search duckduckgo'
+alias yandex='web_search yandex'
+
 #add your own !bang searches here
 alias wiki='web_search duckduckgo \!w'
 alias news='web_search duckduckgo \!n'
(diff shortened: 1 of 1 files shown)


mcornella:
@ajdle @Xenopathic @alessandro-pezzato @bchretien @takac @mkolosick @varunkvv @ibnHatab @Biriukov @ai81 @mmounirou @friederbluemle @peteches @Guerki @stibinator @stelim @xuhdev 
Can you guys test this out?

varunkvv:
:+1: . Tested on ubuntu with zsh 5.0.2.

xuhdev:
No problem on 5.0.7 on Arch Linux. There is one thing noticeable: the yandex search is using http but others are using https. I tried yandex with https which also works. Don't know if there is a specific reason to use http in this case.

mcornella:
Huh, that's weird. I never checked the search page, but the home page redirects to the http URL. 
I'll make it to use https then, thanks for your feedback!

takac:
LGTM. No issues here on 5.0.2 on Ubuntu.

mcornella:
I don't know, I left it like I found it. Notice that, as it is, it's backwards compatible with previously defined aliases. That said, we could add a `ddg` key and keep the `duckduckgo` as well, but I'd like some consensus on that before I do anything.

PR #3444, opened 2014-12-27
Add completion for httpie

lululau:
HTTPie is a command line HTTP client, a user-friendly cURL replacement. 
http://httpie.org


robbyrussell:
Can you include a README file in another PR?

PR #3445, opened 2014-12-28
Removing trash command from osx plugin

### LINKED ISSUE #3421: OSX: move files to the proper Trash
The command:
```sh
trash /Volumes/something/abc
```
moves the file/directory to `~/.Trash` on OS X, i.e. moves it to a different drive. It should just move the file/directory to the trash proper for the volume the file is on.

Tested on OS X Yosemite with oh-my-zsh from commit 1978a0923c864cacdc520adff83ed63f62c91073 (current `master`).

Note that Apple provides an API to move files to a proper trash via AppleScript; modules like https://www.npmjs.com/package/trash do it correctly because of that.

Originally erroneously reported at https://github.com/sindresorhus/trash/issues/17.

MoLLIa:
Current **trash** command implementation from osx plugin has a bunch of issues:

1. It does not have “Put Back” feature (Fixes #3038).
2. It does not rename files correctly whether we already have file with the same name in the Trash (Fixes #3195, #3038).
3. It does not move file to proper trash on multiple drives systems (Closes #3421).
4. It overrides trash command from http://hasseg.org/trash/ implementation and requires additional workaround to work with osx plugin.

@Kriechi has suggested https://github.com/robbyrussell/oh-my-zsh/pull/3038#issuecomment-52826550 to remove trash command from plugin in favor of binary trash and afair you have agreed to remove it, but it still exists.

So here is the magic PR. :blush:

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -137,23 +137,6 @@ function man-preview() {
   man -t "$@" | open -f -a Preview
 }
 
-function trash() {
-  local trash_dir="${HOME}/.Trash"
-  local temp_ifs="$IFS"
-  IFS=$'\n'
-  for item in "$@"; do
-    if [[ -e "$item" ]]; then
-      item_name="$(basename $item)"
-      if [[ -e "${trash_dir}/${item_name}" ]]; then
-        mv -f "$item" "${trash_dir}/${item_name} $(date "+%H-%M-%S")"
-      else
-        mv -f "$item" "${trash_dir}/"
-      fi
-    fi
-  done
-  IFS=$temp_ifs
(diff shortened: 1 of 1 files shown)


mcornella:
#3038 won't be removed because it's not preceded by a `fixes`. A comma isn't read by @github's parser

mcornella:
#3195 is also closed by this PR.

ncanceill:
Thanks!

henvic:
Just to make sure... By binary trash you mean this http://hasseg.org/trash/ binary program, right?

Kriechi:
@henvic yes, http://hasseg.org/trash/ should be much more reliable - and is easily available via homebrew.

PR #3451, opened 2014-12-29
sublime: Make subl a function so it can be called from functions in other modules

paulmelnikow:
Without this change, whether a function can invoke 'subl' depends on the order in which modules are loaded. With this change, the order doesn't matter.

zsh resolves aliases at parse time not run time, which means if you invoke an alias from a function, the alias must be defined before the function.

See http://stackoverflow.com/a/25532276/893113 which refers to an explanation of this limitation.

For code which may be used non-interactively, the recommend workaround is to use functions instead of aliases.


paulmelnikow:
Just wanted to bump this up. Without this change, users sometimes can't invoke the command from functions in other modules.

ncanceill:
This cannot hurt, thanks.

Issue #3447, opened 2014-12-29
oh-my-zsh uninstall.sh issues - switch back to bash is unsatisfactory


sr-murthy:
Hi

I recently installed `oh-my-zsh` on my system (running OS X Yosemite, v. 10.10.1, with `bash` (v. 3.2.53) as default, to test it out.  It was really interesting, but wanting to get back to `bash` I ran the uninstall script and noticed two minor problems with it (with easy fixes):
1. it does not make the change to the user visible in the _current_ shell window - this is because the although the relevant line `chsh -s /bin/bash` (line 24 in `uninstall.sh`) will certainly change the `UserShell` property in the user's record in the OS X Directory Services database (for the local domain node "`.`") from `bin/zsh` to `/bin/bash`, the change will not be applied to the user's current (presumably `bash`) shell window - they have to open a new window (after possibly closing the current one, or from the existing one).  Of course the user may know this, but also they may not, but it would seem a nice to idea to have this done cleanly so that the uninstallation appears painless.  I propose a simple fix - do `exec su - $USER`, which replaces the current shell instance with a new one, immediately after `chsh -s /bin/bash`.  Some prompting, pre- and post-execution checking may be required.
2. it sources `/etc/profile` (after doing `chsh -s /bin/bash`).  This seems unnecessary according to the [OS X bash man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/bash.1.html) `/etc/profile` is sourced first after `bash` is invoked as an interactive login shell (or a non-interactive login shell with the `--login` option).  So having done `chsh -s /bin/bash` the user would be expected to open either a new window from their terminal app or close and reopen their terminal app, which will start an interactive login `bash` shell that will source `/etc/profile` first.  Thus this is unnecessary in the uninstall script.

I have forked this repo, applied the changes and made a [pull request here](https://github.com/robbyrussell/oh-my-zsh/pull/3454).

mcornella:
Hey @sr-murthy, #5169 deals with this and it's now ready to test. If you want to help speed up the testing and merge (which should happen early next week), you can wget or curl it from the URL `https://raw.githubusercontent.com/mcornella/oh-my-zsh/refactor-installer/tools/install.sh`. Read the comments at the top of the script to know how to call it.

Thanks!

PR #3454, opened 2014-12-30
Smoother way of changing default login shell back to bash on uninstall

sr-murthy:
The current uninstall script attempts to `chsh -s /bin/bash` to change the default login shell for the user back to `bash`.  But `chsh -s /bin/bash` requires user login, and the script cannot deal with login failure or any disruption to the command.  The main change, in general terms, is to check the exit code of `chsh -s /bin/bash` and to prompt the user more appropriately in case of success or failure of the login.  A minor change is to remove the command `source /etc/profile`.  This is because, as [explained here](https://github.com/robbyrussell/oh-my-zsh/issues/3447), according to the [OS X bash man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/bash.1.html) `/etc/profile` is sourced first after `bash` is invoked as an interactive login shell (or a non-interactive login shell with the `--login` option).  So having done `chsh -s /bin/bash` the user would be expected to open either a new window from their terminal app or close and reopen their terminal app, which will start an interactive login `bash` shell that will source `/etc/profile` first.  Thus this is unnecessary in the uninstall script.

In the case of success `chsh -s /bin/bash` will reset the `UserShell` attribute of the user’s record in the OS X Directory Services database for their local domain node `.` back to `/bin/bash`, but this change is not immediately reflected in the working shell window (in which the uninstall script was executed).  So the change here specifically involves asking the user whether they would like to do this, and dealing with the two different possibilities.

These changes are self-explanatory, but could be improved even further - for example, an assumption is made that the user wants to default back to `/bin/bash` but perhaps they do not, or perhaps they do not care because they have specified a specific shell type and path in their terminal app, or they do not wish to default to `bash` but another one of the accepted shells in `/etc/shells`.  For the moment thought these changes should make the uninstallation process less painful and the script easier to maintain and debug.

Note: the default login shell for a local domain user, from the point of OS X, is determined not by environment variables like `SHELL` or the shell preferences in Terminal.app but the value of the `UserShell` attribute of the user's record in the OS X Directory Services database for the their local domain node `.`.  Moreover executing `chsh -s <shell path>` in the active shell window will in fact change this attribute directly but the change will not be visible immediately unless a new shell window is opened or the session restarted or with a command like `exec su - $USER`.

### DIFF
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -20,9 +20,24 @@ then
 
   source ~/.zshrc;
 else
-  echo "Switching back to bash"
+  echo "Changing default login shell from zsh back to bash ... login required."
   chsh -s /bin/bash
-  source /etc/profile
+  if [ "$?" = "0" ]; then
+    echo "Done.  All new shell windows will be bash!"
+    printf "Replace this zsh window with a bash window right now? (yes/no): "
+    read CHOICE
+    if [ "$CHOICE" = "yes" ]; then
+      echo "OK, please login again below ... if something goes wrong you can \
+always close this window and start a new bash window."
+      exec su - $USER
+    else
+      break
+    fi
+  else
+    echo "Oops, something went wrong with the change to bash ... \
+you can execute chsh -s /bin/bash manually and start a new window"
+    break
+  fi
 fi
 
-echo "Thanks for trying out Oh My Zsh. It's been uninstalled."
+echo -e "\nThanks for trying out Oh My Zsh. It's been uninstalled."


apjanke:
I don't know if the shell replacement code does quite what you want. It runs a `bash` shell, but there's still a parent `zsh` process running underneath it.

Take a look at how `uninstall_oh_my_zsh()` is implemented. It runs `uninstall.sh` in a new child `sh` process.

```
$ which uninstall_oh_my_zsh
uninstall_oh_my_zsh () {
	env ZSH=$ZSH /bin/sh $ZSH/tools/uninstall.sh
}
```

So when `uninstall.sh` does that `exec`, it replaces the child `sh` with a `bash`. But the parent `zsh` is still running. And when you exit the child `bash`, it drops you back to `zsh`. I don't know if that behavior is less confusing than the old behavior.

For example, here's what I saw when testing this out on Fedora 21. (I tested this by doing a normal oh-my-zsh installation, and then moving ~/.oh-my-zsh away and cloning sr-murthy's repo to replace it.) Note that at the end, I'm back in an oh-my-zsh themed `zsh` process.

```
➜  ~  uninstall_oh_my_zsh 
Removing ~/.oh-my-zsh
Looking for original zsh config...
Changing default login shell from zsh back to bash ... login required.
Changing shell for janke.
Password: 
Shell changed.
Done.  All new shell windows will be bash!
Replace this zsh window with a bash window right now? (yes/no): yes
OK, please login again below ... if something goes wrong you can always close this window and start a new bash window.
Password: 
[janke@dora ~]$ ps
   PID TTY          TIME CMD
  4035 pts/1    00:00:00 zsh
  4232 pts/1    00:00:00 bash
  4271 pts/1    00:00:00 ps
[janke@dora ~]$ exit
logout
➜  ~  ps
   PID TTY          TIME CMD
  4035 pts/1    00:00:00 zsh
  4280 pts/1    00:00:00 ps
➜  ~  
```

I agree that removing `source /etc/profile` makes sense. Because `uninstall.sh` was running in a subshell, it never would have had any visible effect in the first place.

tools/uninstall.sh
```diff
@@ -20,9 +20,24 @@ then
 
   source ~/.zshrc;
 else
-  echo "Switching back to bash"
+  echo "Changing default login shell from zsh back to bash ... login required."
   chsh -s /bin/bash
-  source /etc/profile
+  if [ "$?" = "0" ]; then
+    echo "Done.  All new shell windows will be bash!"
```
apjanke: Is "window" the right term? The user might not be running a windowing system. Maybe "session"?

apjanke:
See also #2626. Looks like there's an existing PR for saving the user's old default shell.

sr-murthy:
Thanks, you're right, I should have checked this.  I'll try and close this pull request.

PR #3460, opened 2014-12-31
Fix to make the gitstatus.py works with any language

c4software:
Hi,

The script fail when the language of the system is not set at "en_US.UTF-8" . I have made change to force the lang to en_US.UTF-8 when this script run.

I also notice the behead_re and  diverge_re regexp are wrong. (extra # and space at the begining)

Now the gitstatus.py return the correct informations.

### DIFF
--- a/plugins/git-prompt/gitstatus.py
+++ b/plugins/git-prompt/gitstatus.py
@@ -2,6 +2,7 @@
 # -*- coding: UTF-8 -*-
 from subprocess import Popen, PIPE
 import re
+import os
 
 # change those symbols to whatever you prefer
 symbols = {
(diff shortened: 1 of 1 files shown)


c4software:
Anyone can review it?

mcornella:
A better way would be using `LANG=en_US.utf8` before python in [line 46 of `git-prompt.plugin.zsh`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git-prompt/git-prompt.plugin.zsh#L46).
Can you test that out?

A better language, though, would be using `C`, which picks the default language in which git is written (in this case, English). In some systems, mainly those that are in languages other than English, "en_US" locale is not available.

c4software:
Nop its doesn't work.

mcornella:
Sorry, I meant to say `C.utf8`, can you try it out? Also, post the output of `locale -a`

mcornella:
Updates @c4software ?

mcornella:
Ok, I finally reviewed this and it seems to be worse than expected. Let's pull in the latest changes from [olivierverdier/zsh-git-prompt](https://github.com/olivierverdier/zsh-git-prompt) instead.

mcornella:
My recent PR #4205 solves this problem, as well as the `#` regexps.

c4software:
Thanks for the feedback (sorry for the delay…)

mcornella:
No problem! Do you mind checking #4205 and +1 there if it works?

PR #3461, opened 2014-12-31
Provide downloaded scripts checksums for security

### LINKED ISSUE #3459: Users are instructed to run a script downloaded via HTTP
On ohmyz.sh and the project's README file, users are instructed to run commands such as

``curl -L http://install.ohmyz.sh | sh`` and
``wget --no-check-certificate http://install.ohmyz.sh -O - | sh``

While this may be practical, it's totally unsafe and dangerous. The script is susceptible to multiple kinds of man-in-the-middle attacks. This problem could be solved by one of the following methods:

* Using HTTPS on ohmyz.sh.
* Using the direct link to the file in question (https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh) and getting rid of the install.ohmyz.sh link.
* Having some kind of checksum check in place.

dionyziz:
As reported in #3459, the README file suggests an insecure download and
execution via HTTP instead of HTTPS.

However, before this essential migration to HTTPS is completed, this
patch adds checksums that will allow users to verify the authenticity of
their download.

The authenticity of the checksums is assured because it will be viewed
on an HTTPS-served GitHub page on the correct author username
(robbyrussell).

I have verified this checksum by downloading the HTTP script from two
different network topologies:

1. The Google corporate network
2. The Linode network

I also visually verified that the script code is not malicious. I have
also compared it with the HTTPS-served script at this location and found
it to be identical:

https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh

This comes from an HTTPS connection from a valid GitHub domain which
ensures that the owner (robbyrussell) has pushed this file as long as we
trust GitHub and PKI.

This commit is signed with my personal GPG key for authenticity.

The reviewer should also verify the checksum is correct.

Thanks for helping keep users secure.

Fixes #3459

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -18,11 +18,25 @@ You can install this via the command-line with either @curl@ or @wget@.
 
 h4. via @curl@:
 
-@curl -L http://install.ohmyz.sh | sh@
+@curl -L http://install.ohmyz.sh > ohmyz.sh@
+@openssl sha256 ohmyz.sh@
+
+Verify that the checksum is correct:
+@SHA256(ohmyz.sh)= ad4290f1234ced4f9f344868422e194764d076255df806d90a6bb5bc96c77fb9@
+
+Then run:
+@sh ohmyz.sh@
 
 h4. via @wget@:
 
-@wget --no-check-certificate http://install.ohmyz.sh -O - | sh@
+@wget --no-check-certificate http://install.ohmyz.sh -O ohmyz.sh@
+@openssl sha256 ohmyz.sh@
+
+Verify that the checksum is correct:
+@SHA256(ohmyz.sh)= ad4290f1234ced4f9f344868422e194764d076255df806d90a6bb5bc96c77fb9@
+
+Then run:
+@sh ohmyz.sh@
 
 h4. *Optionally*, change the install directory:
 


nextgenthemes:
I do not see this suggestion on http://ohmyz.sh/ all I see is that everybody should just copy and paste this insecure code and hope nothing happens, seriously getting annoyed by this lack if caring.

dionyziz:
Yes, it would be nice to have this either merged or receive a response from the author :)

robbyrussell:
"seriously getting annoyed by this lack if caring"

Don't mistake silence for a lack of caring. I haven't read most of the open pull requests at this point in time. Low priority != Does not care.

nextgenthemes:
putting the security of thousads of users at risk by declaring it low priority = lack of careing. WTF has more priority then security you have been warned about for ages and should actually never do it this way in the first place.

mcornella:
For that matter, it's easier for the user and safer to just use the github url directly, so I don't think we should keep this one open. @robbyrussell do you agree on going back to the old, long URL while we're implementing SSL?

mcornella:
BTW, the full `https` URL is just below in the README file, but I understand the `average user` complaint. I've submitted #3607 for the time being.

corinnekunze:
http://ohmyz.sh/ has now been updated with the old, longer URL for the time being.

robbyrussell:
@corinnekunze  Can you remove the ' --no-check-certificate' bit on the ohmyz.sh page for us?

corinnekunze:
@robbyrussell Done and done!

dionyziz:
Thanks folks!

PR #3462, opened 2014-12-31
[colored-man] squeeze blank lines

valeriangalliat:
Like **man**(1) does by default, use `less -s` as pager to squeeze blank
lines.


mcornella:
I know this is old but this can be achieved by setting `$LESS` to also have `-s`.

PR #3465, opened 2015-01-04
fcd does a partial search and cds into the result

ArjenSchwarz:
You can define your own search root (or roots) and it will use that as a
base for the results. It also supports autocompletion for the results it
finds.


mcornella:
Many plugins like this one already exist in the repository. You can add it as [an external plugin](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins).

PR #3467, opened 2015-01-04
[Plugin composer] Multiple fixes

mcornella:
### STATUS: ready!
-------
This pull request is a compilation of the following:

- #2000: use `$_comp_command1` instead of `composer` to allow using `composer.phar` command; also complete `composer.phar` commands as well.
- #2049: add `composer global` aliases.
- #3375: little change in awk regular expression to parse composer commands in newer versions.

Also, I removed the requisite of having a `composer.json` file in the working directory to show composer commands. This I did because, as it was before, commands in the else statement weren't all that were available to run. Now, all commands will be shown instead (deprecates #2143).
**If no one agrees with this change I can get rid of it**.

Fix #2000, fix #2049, fix #3375.
Close #2143.
Close #3514, #3555.

### DIFF
--- a/plugins/composer/composer.plugin.zsh
+++ b/plugins/composer/composer.plugin.zsh
@@ -41,9 +41,12 @@ compdef _composer composer.phar
 alias c='composer'
 alias csu='composer self-update'
 alias cu='composer update'
+alias cr='composer require'
 alias ci='composer install'
 alias ccp='composer create-project'
 alias cdu='composer dump-autoload'
+alias cgu='composer global update'
+alias cgr='composer global require'
 
 # install composer in the current directory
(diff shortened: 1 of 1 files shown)


mcornella:
@zvirusz @Anahkiasen @komuta: what do you think?

Anahkiasen:
Looks good to me

komuta:
Testing it from now on, and so far everything seems ok for me too.
I have just one request: I realized that I messed up with my author name in #3375. I just fixed it up, so could you please upgrade my patch in your branch ? Thanks.

mcornella:
Done!

komuta:
That was fast! Thanks!

irnnr:
looks good to me

PR #3474, opened 2015-01-07
Setup `grep` alias lazily through a function

blueyed:
This saves an extra / initial call to `grep` during shell startup.

/cc @mcornella

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -7,13 +7,14 @@ _setup_grep_alias() {
         echo | grep $1 "" >/dev/null 2>&1
     }
 
+    local GREP_OPTIONS
     # Color grep results.
-    local GREP_OPTIONS="--color=auto"
+    GREP_OPTIONS=(--color=auto)
 
     if grep-flag-available --exclude-dir=.cvs; then
-        GREP_OPTIONS+=" --exclude-dir=$VCS_FOLDERS"
+        GREP_OPTIONS+=(--exclude-dir=$VCS_FOLDERS)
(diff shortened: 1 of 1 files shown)


mcornella:
It doesn't work on first try though, what's the motivation of this?

blueyed:
> It doesn't work on first try though

Fixed that.

I've also made it a function now, and auto-add `-r` if called on a dir. That might be too much magic though. What do you think?

mcornella:
Yeah I'm not crazy about the recursive flag. Now at least it works all the time, but see that you removed the alias redefinition, so every time we run grep we'll be executing the setup function. 

Also, see my proposed PR where I turn grep into a function for further discussion regarding alias/function definition: #3347

blueyed:
> but see that you removed the alias redefinition, so every time we run grep we'll be executing the setup function

It uses `unalias grep` and then sets up a function, when the alias is called for the first time.

This might also use the autoload mechanism provided by zsh, where the function would be defined in a file, which then gets sourced on the first invocation.

mcornella:
Yeah sorry, I missed it. It's on the very [top of the function](https://github.com/robbyrussell/oh-my-zsh/pull/3474/files#diff-bfc27536871d24498f9e19c8e54b0628R22)

mcornella:
Have you had a chance to look into the alias vs. function discussion? It was a little scattered into https://github.com/robbyrussell/oh-my-zsh/pull/3347#issuecomment-64924377 and https://github.com/robbyrussell/oh-my-zsh/pull/3341#issuecomment-64830311

PR #3476, opened 2015-01-08
Add simplified theme haichong20

jackqt:
Simplified theme for oh-my-zsh. 
Show out current time/user name/current path/git branch and latest commit ID
The input command will be on new line instead of follow after path as one line


robbyrussell:
Thanks for this, but we're not currently accepting any new themes per our README.

PR #3478, opened 2015-01-08
Allow custom cache dir

stephenrjohnson:
We are running oh-my-zsh on a shared install and the only issue is each
user having a custom cache dir. This will allow us to set the cache dir
to ZSH_CACHE_DIR=~/oh-my-zsh/cache/ .

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -38,7 +38,7 @@ cdpath=(.)
 
 # Use caching so that commands like apt and dpkg complete are useable
 zstyle ':completion::complete:*' use-cache 1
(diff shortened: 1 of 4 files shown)


pdcawley:
`plugins/last-working-dir/last-working-dir.plugin.zsh` also makes reference to `$ZSH/cache`, so that could probably stand to respect $ZSH_CACHE_DIR as well.

stephenrjohnson:
@pdcawley Good catch, updated the reference to $ZSH/cache in a few places.

robbyrussell:
Thanks for this!

PR #3490, opened 2015-01-12
Fix Plugin commands

lmartins:
Is seems the plugin commands were missing the "wp plugin" prefix.


mcornella:
Obvious PR :+1: 

/cc @joshmedeski
/cc @robbyrussell

joshmedeski:
good catch @lmartins :smile:

hsingh23:
+1

PR #3493, opened 2015-01-14
Fast installation time by shallow copying git repo

NicholasTD07:
Use `--depth=1` to clone just the lastest version rather than every version in git.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -10,7 +10,7 @@ if [ -d "$ZSH" ]; then
 fi
 
 echo "\033[0;34mCloning Oh My Zsh...\033[0m"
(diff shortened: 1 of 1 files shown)


mcornella:
I presume doing a `git fetch` afterwards would be enough to download the whole repo?

NicholasTD07:
@mcornella I found the 'unshallow' option in git-pull, which is documented in this page: http://git-scm.com/docs/git-pull. As expected, running `git pull --unshallow` will convert shallow repo to a complete one.

I agree that as long as it is documented, it should'n be a problem. And IMHO, for general usage, complete git history is not necessary.

robbyrussell:
Nice, this will speed things up for new folks.

PR #3494, opened 2015-01-15
Fix echo nothing if $SHORT_HOST doesn't exist.

AlexTalker:
`echo` doesn't care about empty argument, so, if $SHORT_HOST doesn't exist, it's just print space or so.

### DIFF
--- a/themes/candy-kingdom.zsh-theme
+++ b/themes/candy-kingdom.zsh-theme
@@ -13,7 +13,7 @@ patches: <patches|join( → )|pre_applied(%{$fg[yellow]%})|post_applied(%{$reset
 }
 
 function box_name {
-    [ -f ~/.box-name ] && cat ~/.box-name || echo $SHORT_HOST || echo $HOST
+    [ -f ~/.box-name ] && cat ~/.box-name || echo ${SHORT_HOST:-HOST}
 }
 
 PROMPT='
--- a/themes/fino-time.zsh-theme
+++ b/themes/fino-time.zsh-theme
@@ -21,7 +21,7 @@ function prompt_char {
 }
 
 function box_name {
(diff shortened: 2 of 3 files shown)


mcornella:
`$SHORT_HOST` is defined on initialization, see [`oh-my-zsh.sh`](https://github.com/robbyrussell/oh-my-zsh/blob/master/oh-my-zsh.sh#L42-L48). 
That said, the line _is_ wrong, but here's what it should be:
```patch
-    [ -f ~/.box-name ] && cat ~/.box-name || echo $SHORT_HOST || echo $HOST
+    [ -f ~/.box-name ] && cat ~/.box-name || echo ${SHORT_HOST:-HOST}
```
What that does is, if `$SHORT_HOST` is undefined or empty, use `$HOST` (it's a common construction in all major shells, see [POSIX spec](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02)).

AlexTalker:
@mcornella oh, interesting.
First time I even wanted did it via test(as you know,  [[ and ]] is alias for test), but I thought, it's not very portable :smile: 
Let me fix it...

mcornella:
Hey can you squash your commits? 

Also, please don't mention me (or anyone) in a commit because github notifies the user each time the commit enters a repository (including forks). I've received a lot of emails because of it :grin: 
It's fine if you don't include the at sign `@`, which is the character to trigger a mention. 

Let me know if you have doubts regarding how to do it.

AlexTalker:
@mcornella , hey, I did it first time, sorry if something is wrong :smile: 
The only doubts I had is force pushing into my repo, but folks from #git helpful :+1:

mcornella:
Good job man! 
/cc @robbyrussell: good to go!

PR #3496, opened 2015-01-15
fixes plugin aliases, adds completion

hsingh23:
Fixes all plugin aliases:
alias wppa='activate'
 goes to =>
alias wppa='wp plugin activate'

adds completions


mcornella:
Duplicate of #3490. Please close this and go +1 the other one.

PR #3497, opened 2015-01-15
Adds completions to wp-cli plugin

hsingh23:


mcornella:
Oh right! I missed that you also added completion! Sorry :grin: 

@lmartins @joshmedeski can you try this?

lmartins:
Looks good here.

PR #3501, opened 2015-01-16
Git improved

FrankGruellich:
Hi,

this adds two features:
1. the prompt configuration is much more flexible and enables one to configure it in pretty much any aspect by modifying the theme and not messing in the code files,
2. the git status is updated less frequently, only when needed:
- when a git command is invoked,
- when the directory is changed.

I'd appreciate if you could add it.  Let me know if there are any issues.  Thanks a lot.

Kind regards,
 Frank.


apjanke:
Wouldn't this end up with stale git info if you were working with a repo in multiple programs (e.g. multiple zsh windows, or zsh and a Git GUI, or an external text editor), had git commands run inside functions or programs that didn't start with the string "git", or were otherwise modifiying the repo state outside of a single zsh process?

FrankGruellich:
Short answer: yes.

mcornella:
This is more of a feature to be implemented on a per-theme basis. I know there are some themes already in OMZ that do this, like `steeef` and `half-life`.

PR #3517, opened 2015-01-25
debian plugin for alzheimer and typo people

casaper:
# debian-simple

To me the short to letter aliases the [debian](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/debian) plugin has are simply not memorable and possibly even dangerous, because my tab key is in extensive use. I bet the maker's of the original debian plugin will love theyre plugin to be with those short aliases, wich is of course totally fine.
But I also kind of think that I'm not the only debian/ubuntu oh-my-zsh user that has a terrible memory ;), so I thought it would be nicer to share my modification of **debian** in stead of just making my _custom/plugins/debian_.


mcornella:
There seems to be no love for this plugin. You can set it up as a custom plugin, see [adding a new plugin](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#adding-a-new-plugin).

PR #3522, opened 2015-01-27
Do not use -i (insecure) with compinit

blueyed:
This was added in 99d72ca, to work around warnings with `sudo -s`.

`sudo -s` seems to cause a lot of other issues anyway, where files get
owned by root etc.

I think it's better to be safe by default.

From zsh's man page:

       For security reasons compinit also checks if the completion
       system would use files not owned by root or by the cur‐ rent
       user, or files in directories that are world- or group-writable
       or that are not owned by root or by  the  cur‐ rent  user.   If
       such  files or directories are found, compinit will ask if the
       completion system should really be used.  To avoid these tests
       and make all files found be used without  asking,  use  the
       option  -u,  and  to  make compinit  silently  ignore  all
       insecure  files and directories use the option -i.  This security
       check is skipped entirely when the -C option is given.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -54,7 +54,7 @@ fi
 
 # Load and run compinit
 autoload -U compinit
-compinit -i -d "${ZSH_COMPDUMP}"
+compinit -d "${ZSH_COMPDUMP}"
 
 # Load all of the plugins that were defined in ~/.zshrc
 for plugin ($plugins); do


ncanceill:
Well, I guess this makes sense since we cannot know what environments people deploy OMZ in...

However, I believe you should not have to worry about `-i` being _"insecure"_ unless you already have serious security issues, and by that time I am afraid that those warnings would get more annoying than helpful.

Anyone care to weigh in?

apjanke:
I think the `-i` option is safe, too. "Ignoring" those insecure files means excluding them from the completion lookup, not skipping the security check on them, so it's not a security hole. The `compinit -u` option is the one that would be insecure. `compinit -i` is probably safest because it doesn't even give you a chance to include the insecure paths, which an uninformed or rushed user might.

IMHO, people shouldn't run OMZ as `root` anyway; it's not exactly heavily audited.

There's a problem with simply removing `-i` on Cygwin, because it's common that there are insecure paths in its setup (from OMZ itself), and I don't think we have a good handle on how to get rid of them easily. So Cygwin users might see some annoying warnings even as non-root users. E.g. here's what I get when I run plain `compinit` on Cygwin after loading OMZ.

```
[@ in ~]
$ compinit
zsh compinit: insecure directories, run compaudit for list.
Ignore insecure directories and continue [y] or abort compinit [n]? y%                                           [@ in ~]
```

Not sure what the right thing to do here is.

blueyed:
Thanks for your investigation @apjanke!

It sounds like this PR should be closed then / is void for the common use case?!

apjanke:
Yeah, I believe so: this isn't needed to close a security hole, and the `-i` behavior is better than the prompting, IMHO.

There are some other completion-related issues caused by `compinit -i`, but they require a more complicated fix, and are being discussed in #3889. They're more about making `compinit -i` play well with the rest of OMZ, and not removing the `-i`.

PR #3523, opened 2015-01-28
Changed hard-coded vim to use EDITOR variable

gaulinmp:
previously zshrc alias called 'vim...'. Changed it to use ${EDITOR:-vim} so it looks for the EDITOR variable (set by in the default .zshrc file) and defaults to vim. Using single quote so updates to EDITOR environmental variable will be used automatically.


apjanke:
Works for me on OS X 10.9, Fedora 21, and Windows 7/Cygwin. Changing `$EDITOR` is respected.

gaulinmp:
Fixed by 9a5e1a055201409ebe9e1ab97cd8c3fe52b990f5. Closing.

PR #3525, opened 2015-01-28
Move pwd update output to precmd to avoid contamination

### LINKED ISSUE #3524: Output from chpwd hook contaminates script output
The recent change for Apple Terminal style pwd notification https://github.com/robbyrussell/oh-my-zsh/pull/3429 causes terminal control characters to be emitted upon each chpwd. This happens even if it's not about to display a prompt; e.g. if a cd is done inside a function or script.

This can contaminate the output of the function or script if it's being captured by anything. In particular, David Riley reports:

-------------------------------

The recent change breaks Android development on OSX because Android
has a function:
get_build_var () {
T=$(gettop)
if [ ! "$T" ]
then
echo "Couldn't locate the top of the tree.  Try setting TOP." >&2
return
fi
(
\cd $T
CALLED_FROM_SETUP=true BUILD_SYSTEM=build/core command make
--no-print-directory -f build/core/config.mk dumpvar-$1
)
}

The cd commands ends up printing some terminal commands which then get
returned from get_build_var which causes problems later on since it's
expecting a simple string instead of some terminal codes.

------------------------------

The pwd reporting mechanism needs to be changed to emit output only when it's safe to do so, i.e. when displaying an interactive prompt. (It might be convenient to do it on any cd change, so interactive scripts reflect their current location, but I don't think it can be done safely since the terminal control mechanism isn't out-of-band.)

This probably means moving the pwd reporting output back to a precmd hook.

apjanke:
Fixes #3524, where terminal control characters emitted from a chpwd hook can contaminate results of scripts or functions. Moves the control output to the precmd hook so it only shows at a prompt, when it's safe to do so. 

The pwd update code is done in the omz_termsupport_precmd function, instead of in title(), so it works even when DISABLE_AUTO_TITLE is set, so #2646 is still taken care of.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -16,7 +16,7 @@ function title {
 ZSH_THEME_TERM_TAB_TITLE_IDLE="%15<..<%~%<<" #15 char left truncated PWD
 ZSH_THEME_TERM_TITLE_IDLE="%n@%m: %~"
 
-# Runs when you have the prompt
+# Runs before showing the prompt
 function omz_termsupport_precmd {
   title $ZSH_THEME_TERM_TAB_TITLE_IDLE $ZSH_THEME_TERM_TITLE_IDLE
 
@@ -28,7 +28,7 @@ function omz_termsupport_precmd {
   fi
 }
 
-# Runs at the beginning of (and during) of command execution
+# Runs before executing the command
 function omz_termsupport_preexec {
   emulate -L zsh
   setopt extended_glob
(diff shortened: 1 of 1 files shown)


mcornella:
Agree 100% with your diagnosis and your solution :+1: 
Since you've changed those `Appears` comments, I'd also completely reword them to be more accurate:
the first one should be `Runs before showing the prompt`, and the second one `Runs before executing the command`

/cc @robbyrussell

apjanke:
Reworded as suggested. I've also reworded the other comments in termsupport.zsh on separate pull request (https://github.com/robbyrussell/oh-my-zsh/pull/3426).

PR #2790, opened 2014-05-07
[UPDATE] Plugin: git

ncanceill:
## Update branch

**Plugin:** `git`

This branch contains updates for the `git` plugin. If you want to add a commit, please [submit a PR](https://github.com/ncanceill/oh-my-zsh/pull/new/plugin-git) to [my branch](https://github.com/ncanceill/oh-my-zsh/tree/plugin-git).

## Status

**READY FOR MERGE**

This is now rebased against latest upstream (ff4663a6b83b8dd205da3968de4bc458abc08b2c).

There is still some stuff to do before merge:
- [x] Test completion now that most `compdef`s were removed
- [x] Settle the debate on `gp/gl` versus `gP/gp`
- [x] Do another sweep to include any pending PRs related to this plugin
- [ ] Update the [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugin:git)
- [x] Squash commits, because atomicity is out of the window already

## Changelog

The auto-closing feature [should kick in](#issuecomment-43131152) when this is merged, which will ease maintenance.

* Fixes #1500
* Fixes #1504
* Fixes #1546
* Fixes #1601
* Fixes #1800
* Fixes #1966
* Fixes #1982
* Fixes #2124
* Fixes #2338
* Fixes #2407
* Fixes #2433
* Fixes #2463
* Fixes #2493
* Uses #2494, fixes #1846
* Fixes #2546
* Fixes #2684
* Fixes #2714
* Fixes #2921
* Fixes #3085
* Fixes #3091
* Fixes #3122
* Fixes #3276
* Fixes #3374
* Fixes #3543
* Fixes #3680, fixes #3708
* Fixes #3711
* Fixes #3726
* Fixes #3749
* Fixes #3763
* Fixes #3836

PR #3530, opened 2015-01-29
[plugins/git] Add grm and gcp

jedcn:
### Overview

This is a small change to the main git plugin.

It adds `grm` and `gmv` as aliases for `git  rm` and `git mv`.

This is in keeping with the aliases that are already defined-- they are small and save a few keystrokes.

### Things to consider?

* I am new to zsh, and I added the `compdef` stuff because it seemed like the right thing to do, but I don't know how to test it.


Kriechi:
I personally use `grm` for `git rebase master`.
(and `grc` for `git rebase --continue`).
Not sure what the current naming schema is... (prefix `g`, etc.)

I think there a various other PRs trying to cleanup and refactor the git plugin: #3501 #3374 and most importantly #2790

jedcn:
Ahh. Interesting. I didn't realize so much work was outstanding in this area. Good luck to all involved. There are probably many people who will be very excited if the core git plugin changes.

For my part, I will continue to customize my environment using whatever the git plugin provides. I'm OK with any alias changing, really. It's just the matter of a few moments to redefine things I don't like, or I think are missing.

---

@Kriechi Thanks for the thoughts on `grm` as `git rebase master`.. I don't use `git rm` that often-- yet I rebase quite frequently. I'll consider it for my personal config. Take care.

PR #3536, opened 2015-02-02
Only add `--color` to grep's alias if that option is valid.

### LINKED ISSUE #3470: grep options under OpenBSD
Under OpenBSD '--color' is not a valid option for grep. The alias for grep in https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/common-aliases/common-aliases.plugin.zsh#L18 causes grep commands to fail always.

hobarrera:
The option `--color` is not available on all versions of grep. This change verifies that the option is valid before adding it to the alias used by default.

Fixes #3470.

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -3,8 +3,12 @@ grep-flag-available() {
     echo | grep $1 "" >/dev/null 2>&1
 }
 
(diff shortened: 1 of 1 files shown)


hobarrera:
Closed by the identical #3538.

PR #3541, opened 2015-02-03
Tmuxinator plugin improvements

agios:
Plugin find works with symlinked configuration folders (eg when using some kind of dotfiles management)

@phstc is it possible to check that it works well on OSX too?

Also completion now supports the shorthand syntax `mux project_name`

### DIFF
--- a/plugins/tmuxinator/_tmuxinator
+++ b/plugins/tmuxinator/_tmuxinator
@@ -8,6 +8,8 @@ _arguments -C \
   '1: :->cmds' \
   '2:: :->args' && ret=0
 
+_configs=(${$(echo ~/.tmuxinator/*.yml):r:t})
+
 case $state in
   cmds)
     _values "tmuxinator command" \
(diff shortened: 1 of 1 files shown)


mcornella:
See my comment in #3540

agios:
* As far as the zsh builtins go, never occured to me, but yes, this should be more portable, will change accordingly.

* Concerning the $_configs in the cmds part, mux supports a shorthand syntax for start, where instead of `mux start name` you can just type `mux name`

mcornella:
Perfect. Okay then, you have my 👍
—
Marc

plugins/tmuxinator/_tmuxinator
```diff
@@ -8,6 +8,8 @@ _arguments -C \
   '1: :->cmds' \
   '2:: :->args' && ret=0
 
+_configs=(${$(echo ~/.tmuxinator/*.yml):r:t})
```
phstc: works here on my OS X :+1:

mcornella:
#2812 also tries to achieve this but still using the `find` command. Add `Fix #2812` in the description to auto-close that one.

agios:
All set

mcornella:
Perfect, this will probably be merged later on this week. Thanks!

robbyrussell:
Thanks @agios !

PR #3545, opened 2015-02-04
Plugin for Ember-cli

BilalBudhani:
This PR adds plugin for ember-cli and adds required aliases for the commands respectively. 

Feedback would be appreciated. :shipit:


tejasbubane:
:+1:

Issue #3537, opened 2015-02-02
vi-mode does not update prompt correctly when window resized


martong:
When a window is resized, the prompt is not updated correctly:
$KEYMAP has an empty value.

To fix it we need this in vi-mode plugin:
```
TRAPWINCH() {
  [[ -o zle ]] && zle reset-prompt; zle -R
}
```

mcornella:
Hi @martong, can you link some references and post a pull request?

martong:
Okay, so here is how to reproduce the issue:
1) Have a multiline `PROMPT` , the second line containes `$(vi_mode_prompt_info)`
2) Change into vicmd mode ( vi_mode_prompt_info draws some red `<<<`)
3) Then resize the window. The result is that the red `<<<` is disappeared.

PR is on the way ...

martong:
@mcornella Before the PR, could you check whether you have this issue as well?

mcornella:
Sure, I'll check it tonight. I'll let you know what I find.

mcornella:
I can confirm that this behavior occurs even on a single line prompt, having `$(vi_mode_prompt_info)` in `RPS1` prompt variable.
Your solution works. The check of `zle` option is since it's always enabled by default in interactive sessions, so the code would look like this:
```sh
TRAPWINCH() {
  zle reset-prompt
  zle -R
}
```
Thanks for your contribution, let's get the PR ready and see if more people can confirm this works and doesn't introduce any bugs.
Cheers!

martong:
I am using iTerm2 and during resizing the window I get the following error messages:
TRAPWINCH:zle:1: widgets can only be called when ZLE is active
TRAPWINCH:zle:1: widgets can only be called when ZLE is active

Probably zle is not always active?

martong:
Here is the PR:
https://github.com/robbyrussell/oh-my-zsh/pull/3547

97-109-107:
I got ``TRAPWINCH:zle:2: widgets can only be called when ZLE is active`` whin on commit``5ee5403``

Upon removing 
````
TRAPWINCH() {
  if [[ -o zle ]]; then
    zle reset-prompt
    zle -R
  fi
}
````

I stopped getting the error; my prompt consists of:
````
PROMPT='%{$fg_no_bold[white]%}%n%{$fg_no_bold[white]%} %{$fg_no_bold[red]%}%3~$(git_prompt_info)%{$reset_color%} '

# git theming
[...]

PR #3547, opened 2015-02-04
Added trapwinch to vi-mode plugin.

martong:
Fixes https://github.com/robbyrussell/oh-my-zsh/issues/3537


mcornella:
Excelent, so you have my :+1: Let's see if we can get some other vi-mode users to test as well
/cc @robbyrussell

PR #3549, opened 2015-02-04
Make copydir plugin work on Linux

staeff:
pbcopy is not available on Linux. Make the plugin use xclip instaed, when it is not a OS X.

### DIFF
--- a/plugins/copydir/copydir.plugin.zsh
+++ b/plugins/copydir/copydir.plugin.zsh
@@ -1,3 +1,7 @@
+if [[ $(uname -s) != 'Darwin' ]]; then
+  alias pbcopy='xclip -selection clipboard'
+fi
+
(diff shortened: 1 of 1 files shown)


apjanke:
Works for me on Fedora 21.

Maybe the test should be a positive test for `[[ $OSTYPE == linux-gnu ]]` instead of the broader "not Darwin". There's no `pbcopy` on Windows/Cygwin, either, but I don't think `xclip` is available there. `clip.exe` or `echo > /dev/clipboard` might be a better substitute over there anyway since they give you access to the system clipboard instead of the X clipboard.

As an aside, maybe `echo -n $PWD | pbcopy` would be tidier than the current `pwd | tr ` implementation?

mcornella:
master uses `clipcopy`

PR #3551, opened 2015-02-04
Let autocomplete follow symlinks

do3cc:
Else, tab completion throws errors if ~/.tmuxinator is a symlink
This is actually the PR https://github.com/robbyrussell/oh-my-zsh/pull/2979
which I somehow managed to close via a rebase


mcornella:
Duplicate of #3541, please close this one and +1 the other if it works for you.

Issue #3535, opened 2015-02-02
add `git add --all` alias


itsNikolay:
It would be great to have alias something like this:
`gaa` to `git add --all`

tejasbubane:
:+1: I think this is handy.

Newman101:
Done, please refer to #3556.

MarcoSero:
Has this been removed? I just updated to the latest version and I don't see it anymore.

tejasbubane:
Ya even I don't see it in the [file](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh).

ncanceill:
It was removed in the latest update, but I will gladly bring it back for you if you really want it. In the meantime you can use `ga -A` which is only two characters longer.

PR #3556, opened 2015-02-06
Added alias for git add --all, solves #3544

Newman101:
Solves #3544 :)

Have a great weekend and thanks for a great shell!


tejasbubane:
#3544 already is a PR that solves #3535. This is a duplicate.

PR #3563, opened 2015-02-09
Fix Unicode problems by only setting LC_CTYPE if not set

mcornella:
This temporarily closes #2012 since it provides a complex solution that is possibly not useful. If we still have unicode issues with just this solution, we'll open it again and test it.

In most systems, `LC_CTYPE` will be correctly set while `LANG` will not be set (since it's optional). This generates a lot of messed up prompts and other problems. A cleanup of unicode-related issues will be needed after this gets merged.

/cc @robbyrussell

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -9,11 +9,21 @@ setopt long_list_jobs
 export PAGER="less"
 export LESS="-R"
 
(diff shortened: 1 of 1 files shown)


frozen707:
I approve, should work well for most cases. :+1:

PR #3564, opened 2015-02-09
Repost #2422: cleaning lib

### LINKED ISSUE #2412: cd ...... : extend with global aliases

### LINKED ISSUE #3491: Cd command report
```bash
➜  ~  cd lala
cd:cd:13: no such file or directory: lala
➜  ~  cat lala
cat: lala: No such file or directory
```
cd command present an annoying double ```cd:cd:``` message info

mcornella:
Repost of #2422 by @LFDM: 

> - Provides a more useful approach to directory switching, as suggested in #2412
> - Deletes aliases.zsh, as almost all things defined in their can find a place in other files present
>   - The directory aliases (some of the which duplicate functionality) are all in directories.zsh
>   - The history alias is in history.zsh. Also refactored to a nicer case switch
>   - Remaining pieces like the please alias for sudo are in misc.zsh
>   - Deletes the .. aliases from the common-aliases plugin, they are in lib anyway
> - Moves a bindkey statement from misc.zsh to key-bindings.zsh
> - Avoids sourcing of autoenv with every call to cd. This plugin related code is also taken out of the lib directory. Its activation file is now sourced only once - if its not present, the user is informed and the plugin disabled.  Overriding cd by hand is not needed, autoenv does this by itself.
> 
> Closes #2412
> Closes #2601

Closes #2422 
Fixes #3491

This also contains the removal of `sl` alias (as per https://github.com/robbyrussell/oh-my-zsh/pull/2422#issuecomment-53995324), and a small cleanup of the `theme-and-appearance.zsh` file as well as redundant `colors` calls in plugins and themes. 

/cc @robbyrussell

### DIFF
--- a/lib/aliases.zsh
+++ b/lib/aliases.zsh
@@ -4,19 +4,5 @@ alias please='sudo'
 
 #alias g='grep -in'
 
-# Show history
-if [ "$HIST_STAMPS" = "mm/dd/yyyy" ]
-then
-    alias history='fc -fl 1'
-elif [ "$HIST_STAMPS" = "dd.mm.yyyy" ]
-then
-    alias history='fc -El 1'
-elif [ "$HIST_STAMPS" = "yyyy-mm-dd" ]
-then
-    alias history='fc -il 1'
-else
-    alias history='fc -l 1'
-fi
-
 alias afind='ack-grep -il'
 
--- a/lib/aliases.zsh
+++ b/lib/aliases.zsh
@@ -1,8 +1,0 @@
-# Super user
-alias _='sudo'
-alias please='sudo'
-
-#alias g='grep -in'
-
-alias afind='ack-grep -il'
-
--- a/lib/aliases.zsh
+++ b/lib/aliases.zsh
@@ -1,11 +1,3 @@
-# Push and pop directories on directory stack
-alias pu='pushd'
-alias po='popd'
-
-# Basic directory operations
-alias ...='cd ../..'
-alias -- -='cd -'
-
 # Super user
 alias _='sudo'
 alias please='sudo'
@@ -25,11 +17,6 @@ then
 else
     alias history='fc -l 1'
 fi
-# List direcory contents
-alias lsa='ls -lah'
-alias l='ls -lah'
-alias ll='ls -lh'
-alias la='ls -lAh'
 
 alias afind='ack-grep -il'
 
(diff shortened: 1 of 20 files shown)


robbyrussell:
Wow, thanks folks!

mcornella:
Yay!! Thanks Robby :smile:

LFDM:
Thank you guys, especially to you @mcornella - I am sorry I haven't paid more attention to this lately. As you seem to follow this closely: Is there anything you want me to work on? I might find some time to contribute again.

mcornella:
Hi @LFDM, don't worry it's normal not to have spare time to waste :wink: I'm currently working on a better history alias: why don't you take a look, I could use an extra pair of eyes! #3422

lib/aliases.zsh
```diff
@@ -1,35 +0,0 @@
-# Push and pop directories on directory stack
-alias pu='pushd'
-alias po='popd'
-
-# Basic directory operations
-alias ...='cd ../..'
-alias -- -='cd -'
```
soimort: Is there any special reason for removing `alias -- -='cd -'` while not adding it back to `directories.zsh
`?
(I got used to enter a single `-` instead of `cd -`, and it's no longer there now.)

mcornella: It's the same as `1`: https://github.com/robbyrussell/oh-my-zsh/pull/3564/files#diff-f7d0a3ef1bc4893bc37a320a5c593e70L13

You can add the alias at the end of your zshrc file, or if more people want it back, we can add it again to `directories.zsh`

mcornella: Hi @soimort, I've added back the alias because another user was using that as well: https://github.com/robbyrussell/oh-my-zsh/commit/4fbfb149d3edcfe18b7b008c3a4a8e7523d4b811

soimort: It's nice to have the `-` alias out-of-the-box, thanks :+1:

PR #2819, opened 2014-05-20
Fixing virtualenvwrapper plugin for Ubuntu (and Debian)

bsandrow:
Ubuntu and Debian store the system-installed `virtualenvwrapper` in
`/etc/bash_completion.d/virtualenvwrapper,` so that it gets automatically sourced
at startup in Bash. By not putting it somewhere in `$PATH`, they end up excluding
others (e.g. Zsh) that might want to use that file. Oops!

The virtualenvwrapper plugin should account for this so that Ubuntu (or Debian)
users don't end up with this message:
```
zsh virtualenvwrapper plugin: Cannot find virtualenvwrapper.sh. Please install with `pip install virtualenvwrapper`.
```
even when they have a virtualenvwrapper installed to a known location.

PR #3567, opened 2015-02-10
Repost #2819: fix virtualenvwrapper for Ubuntu and Debian

mcornella:
From #2819 by @bsandrow:

> Ubuntu and Debian store the system-installed virtualenvwrapper in
/etc/bash_completion.d/virtualenvwrapper, so that it gets automatically sourced
at startup in Bash. By not putting it somewhere in $PATH, they end up excluding
others (e.g. Zsh) that might want to use that file. Oops!

> The virtualenvwrapper plugin should account for this so that Ubuntu (or Debian)
users don't end up with this message:
> ```
zsh virtualenvwrapper plugin: Cannot find virtualenvwrapper.sh. Please install with `pip install virtualenvwrapper`.
```
> even when they have a virtualenvwrapper installed to a known location.

I also fix indentation ([whitespace agnostic diff](https://github.com/robbyrussell/oh-my-zsh/pull/3567/files?w=0)) and adding to chpwd hook.

Close #2819.
/cc @robbyrussell


mcornella:
Rebased and fixed chpwd setup.

PR #3568, opened 2015-02-10
Repost #3426: move DISABLE_AUTO_TITLE check to hooks

mcornella:
From #3426 by @apjanke:
> This moves the DISABLE_AUTO_TITLE check from title() itself to the precmd/preexec hook functions that call it. It also allows single-argument use by defaulting the long window title ($2) to the short icon title ($1).

> I'd like to do this because it would allow the title() function to be used directly by the user or other callers. In my workflow, I like to set the terminal tab title manually, and had my own settitle() function set up to do that. The OMZ title() function is already out there in the global function namespace, so it makes more sense to do that.

> This arrangement of the check in the hooks was discussed in #266 but dismissed as DRY/not needed for performance. But I think it's worth reconsidering: this isn't for performance, but an interface/functionality change that makes title() more useful.
> 
> Also reformatted the comments to match comments in other OMZ lib/*.zsh files.

Close #3426

It also reformats some code to suite better zsh style and quotes `$TERM` every place it's used.

/cc @robbyrussell

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -1,8 +1,16 @@
-#usage: title short_tab_title looooooooooooooooooooooggggggg_windows_title
-#http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#ss3.1
-#Fully support screen, iterm, and probably most modern xterm and rxvt
+# Set terminal window and tab/icon title
+#
+# usage: title short_tab_title [long_window_title]
+#
+# See: http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#ss3.1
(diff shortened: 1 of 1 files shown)


apjanke:
I have a question about quoting $TERM and similar variables as done here. AFAIK, Zsh doesn't do automatic word splitting on scalar variable substitution like other shells, so it doesn't seem like these quotes are necessary unless the SH_WORD_SPLIT option is on. Are we quoting to support that option, or ksh/sh emulation, or for other reasons?

mcornella:
Just for consistency since I saw the first 2 checks using quotes. I didn't look into it much and decided to use all quotes. That said, I have no idea whether it was needed or not, but now I'll have to take a look :wink:

PR #3571, opened 2015-02-11
Extract CWD notifier back out of auto-title hooks

### LINKED ISSUE #3570: New Terminal tabs no longer default to the current working directory
I just ran the update to `oh-my-zsh` and now when I open new tabs in Terminal they default to my home directory (`~`) instead of the current working directory. Do you know what might be causing this?

mcornella:
This fixes the bug introduced in #3568 combined with #3525, which prevents the routine to notify current working directory to Apple Terminal _**if**_ the auto-title functionality is disabled (if `DISABLE_AUTO_TITLE == true`). 

Fixes #3570.


mcornella:
Closing because of typo in branch name.

PR #3575, opened 2015-02-12
Add nmap plugin and update Vundle plugin

uberspot:
Just some aliases that could be useful for people frequently scanning with nmap. 
Most of these are easy to type while autocompleting with <TAB>. 
I'm not sure if this is the proper way to contribute this to the repo. 
If you have any suggestions do let me know. 

Also I updated the vundle plugin with the latest vundle repo name and commands.


robbyrussell:
@uberspot Can you re-submit these as two unique pull requests?

uberspot:
Done :)

Issue #2937, opened 2014-07-07
terminalapp plugin breaks the prompt


gabrielecirulli:
After noticing that zsh on OS X was not remembering the current directory when I open a new tab as a command is running (in this case, the Rails server), I found a solution [here](http://superuser.com/questions/232835/zsh-open-a-new-tab-in-the-same-directory) and I added `plugins=(git osx terminalapp)` in `.zshrc`.

Even though this made the terminal remember the directory even when a command is running, the problem I'm facing right now is that whenever I'm in a git repository, and I type a command, the command text will overwrite my prompt:

![2048 gabriele mercury_ _work_web_development_projects_personal_2048 personal_2048 zsh zsh 10024](https://cloud.githubusercontent.com/assets/886011/3499879/57615ab0-0605-11e4-94aa-bff4983b0436.png)

When I type, the text just overwrites the `$` prompt:

![2048 gabriele mercury_ _work_web_development_projects_personal_2048 personal_2048 zsh zsh 10024-1](https://cloud.githubusercontent.com/assets/886011/3499886/692d0f8c-0605-11e4-90c8-a3630e13c023.png)

This is how it should actually behave (it behaves correctly in directories that are not git repos):

![gabriele gabriele mercury_ zsh zsh 10024](https://cloud.githubusercontent.com/assets/886011/3499895/89758da0-0605-11e4-9198-14dd035e4453.png)

I pinpointed the problem to the `terminalapp` plugin (`plugins/terminalapp/terminalapp.plugin.zsh`), but I couldn't find a straightforward way to fix it (I am also new to zsh).

Is there a solution to this issue?

**EDIT:** I am using a slightly modified version of the Avit theme, [found here](https://gist.github.com/gabrielecirulli/e297da3e8a318d76fcd3). I noticed that the Avit theme also displays this issue, but other themes don't. Could it be something in the theme itself?

**EDIT 2:** The `bureau` theme seems to display this as well.

mcornella:
See #3582 
@apjanke can you guess what's going on here? I know the terminalapp plugin uses the `chpwd` hook still, can it be that?

apjanke:
I can't reproduce this either, including with that custom theme. I'm on OS X 10.9, oh-my-zsh up to date (commit 77140a2).

[...]

PR #3582, opened 2015-02-15
Fold terminalapp plugin in to lib/termsupport

### LINKED ISSUE #2233: SVN Plugin messes up Prompt when URL contains Spaces

### LINKED ISSUE #3581: termsupport Terminal.app stuff fails for dirs with special chars and remote hosts
Couple bugs in the `lib/termsupport.zsh` implementation of Apple Terminal.app cwd-update support.

It fails on directories with special characters in their name, because only spaces are being URL-escaped. For example, if I do this, Terminal.app ends up incorrectly showing `tmp` as the current directory.

    $ cd tmp
    $ mkdir '$#@%^&*%$'
    $ cd '$#@%^&*%$/'

There's an alternate implementation in the `terminalapp` plugin that URL-escapes all special characters, and works for directories like this.

Also, it's using $HOSTNAME instead of $HOST. $HOSTNAME is a bash-ism and isn't necessarily set under zsh. If $HOSTNAME is unset, it gets replaced with a blank, so the reported `file:` URL always looks like it's on the local host. Fine if you're running zsh locally, but if you're logged in to a remote machine, it'll think the dirs are on localhost instead, so they'll be displayed incorrectly.

Discovered this stuff during the discussion on #2932.

### LINKED ISSUE #3583: Terminal.app cwd update is garbled for multibyte characters
The Terminal.app cwd updating code in the terminalapp plugin (or in lib/termsupport.zsh if #3582 has been merged) has problems with dir names that contain multibyte characters. The multibyte characters are not properly escaped byte-by-byte; it looks like it's trying to use the Unicode code point value for the whole character instead.

To reproduce:

    mkdir -p tmp
    cd tmp
    mkdir ԱԲԳ
    cd ԱԲԳ
    # And then open a new tab with Cmd-T

When I do that, the window title shows "S1S2S3" as the directory, and new tabs open in the home dir, displaying what looks like a warning about a bad dir.

    ## cd /Users/janke/tmp/S1S2S3 ##
    Last login: Sat Feb 14 20:21:46 on ttys007
    [@ in ~]

The URL it's producing for that dir is: file://eilonwy.local/Users/janke/tmp/%531%532%533
But it should have multiple %-escapes per character. (I got that raw URL by adding a diagnostic `echo $PWD_URL >> ~/omz_debug_terminalapp_cwd` statement to `update_terminalapp_cwd()`.)

apjanke:
Removes the terminalapp plugin and folds its implementation in to lib/termsupport.zsh, so Terminal.app is supported without requiring a plugin to be specified.
This replaces the redundant Terminal.app support that was recently added to termsupport in #3429.
Adds `urlencode`/`urldecode` to core `lib/functions`.
Has `svn` plugin decode repo URLs before using in prompt.

Fixes #3581 because the terminalapp plugin implementation uses better escaping.
Addresses the discussion in #2932. 
Closes #2932.
Fixes #3583.
Fixes #2233.
Closes #4129 (because this contains the same urlencode/urldecode functions in core lib).


apjanke:
Downside: this is a potentially breaking change for users that have custom overrides for the `terminalapp` plugin or `lib/termsupport.zsh` in their $ZSH_CUSTOM.

How big a deal is that? Is there a way to notify users of the change? Maybe an oh-my-zsh users mailing list? Or a mechanism in the upgrade script to display some change notes when they do an upgrade that picks this up?

mcornella:
Ok I reviewed this:

- :white_check_mark: Putting function definition and hook-up inside an `if`
- :white_check_mark: Deleting the plugin and putting functionality in `termsupport.zsh`
- :x: Using the old algorithm to convert to url. **I think I found a better way:** [zsh urlencode \#1](https://gist.github.com/nykac/6474224), [zsh urlencode \#2](http://www.zsh.org/mla/users/2005/msg01328.html), [zsh urlencode \#3](http://www.zsh.org/mla/users/2008/msg01153.html).

Using urlencode to build URLs is normally the appropriate way to do it in almost every language possible, the RFC is too complicated to be messing around and zsh doesn't help. So using a common solution is the best way to go. I'll be testing more of this later on.

apjanke:
Those urlencode implementations are not locale-aware and will fail on high value (multibyte) UTF-8 Unicode characters in ways similar to the old algorithm. For example:

```
$ . ./define_urlencode.zsh
[@ in ~/tmp/zsh-urlencode]
$ urlencode1 ԱԲԳ
%31%32%33
[@ in ~/tmp/zsh-urlencode]
$ urlencode2 ԱԲԳ
%531%532%533
[@ in ~/tmp/zsh-urlencode]
$ urlencode3 ԱԲԳ
%31%32%33
[@ in ~/tmp/zsh-urlencode]
$
```

URL encoding percent-escapes must be done byte by byte, not character by character. The new algorithm I grabbed from SO and checked in is just a multibyte-correct version of this urlencode function. I don't know of a standard URL encoding lib or command that's widely available. Maybe we should grab the new one and wrap it in an `omz_urlencode` function? (Though really, to be correct, it should check the current locale, grab the string and convert it to UTF-8, and percent-escape that. The new one I checked in assumes the current locale is either UTF-8 or an ASCII-compatible single-byte encoding.)

Combined test script I used for reference:

```
# define_urlencode.zsh
#
# A few different impls of urlencode found on the web (not necessarily correct!)

urlencode1() {
    setopt localoptions extendedglob
    input=( ${(s::)1} )
    print ${(j::)input/(#b)([^A-Za-z0-9_.\!~*\'\(\)-])/%${(l:2::0:)$(([##16]#match))}}
} 

urlencode2() {
    setopt localoptions extendedglob
    input=( ${(s::)1} )
    print ${(j::)input/(#b)([^A-Za-z0-9_.!~*\'\(\)-])/%$(([##16]#match))}
}

urlencode3() {
    setopt localoptions extendedglob
    input=( ${(s::)1} )
    print ${(j::)input/(#b)([^A-Za-z0-9_.!~*\'\(\)-])/%${(l:2::0:)$(([##16]#match))}}
}
```

apjanke:
Pinging @outpunk: What do you think about the terminalapp plugin functionality being folded in to core OMZ?

apjanke:
I see there's also a `urltools` plugin that provides `urlencode` and `urldecode` functions or aliases by shelling out to Node, Python, Ruby, PHP, Perl, etc, depending on what's installed or requested. Can plugins depend on other plugins? (If we kept this stuff in `terminalapp` instead of folding it in to `lib/termsupport`?) Though I think it would be nice to have a single implementation for it, instead of multiplexing a whole bunch of different ones with external language dependencies; that could make debugging things for other users difficult.

outpunk:
@apjanke I think it would be nice. Also there is no necessity in this stuff on Yosemite.

apjanke:
Are you sure Yosemite has built-in support? I just tested this on a fresh, updated OS X 10.10 Yosemite install, and I get the same behavior as on Mavericks: `bash` will update the terminal directory using code in `/etc/bashrc`, but when running `zsh` in a Terminal, the directory is not displayed in the title and new tabs open in `$HOME` instead of the prior tab's current working directory.

mcornella:
> The new one I checked in assumes the current locale is either UTF-8 or an ASCII-compatible single-byte encoding.

When you refer to the "new one" you mean the one in #3584 right? Please keep terminal.app functionality only in this PR and close the rest. It's easier to follow progress; the other way it puts too much a strain in my brain (pun intended). The way I mean is, if you notice something that's missing, post a comment detailing it, then work on the code and push a commit to this same branch. 

Also, add `close #2932` in your PR description to auto-close that one too. Look at #3467, #3413, #3443 for examples.

----
Now, about url encoding, I'm assuming this is the last version of the one you made:
```sh
  # URL-encodes a string
  # Outputs the encoded string on stdout
  # Returns nonzero if encoding failed
  function _omz_urlencode() {
    local url_str=''
    {
      local str=$1

      # URLs must use UTF-8 encoding; convert if required
      local encoding=${LC_CTYPE/*./}
      if [[ -n $encoding && $encoding != UTF-8 ]]; then
        str=$(iconv -f $encoding -t UTF-8)
        if [[ $? != 0 ]]; then
          echo "Error converting string from $encoding to UTF-8" >&2
          return 1
        fi
      fi

      # Use LC_CTYPE=C to process text byte-by-byte
      local i ch hexch LC_CTYPE=C
      for ((i = 1; i <= ${#str}; ++i)); do
        ch="$str[i]"
        if [[ "$ch" =~ [/._~A-Za-z0-9-] ]]; then
          url_str+="$ch"
        else
          hexch=$(printf "%02X" "'$ch")
          url_str+="%$hexch"
        fi
      done
      echo $url_str
    }
  }
```
I'll comment just 2 things for now: 
1. Those extra curly braces do something or is it just a mistake?
2. Locales can also be in the form `XX.utf8` so you might want to test for that too.

I'll be testing this out later this weekend.

----
@outpunk we need more information on how it works, why, and some documentation or references

apjanke:
Gotcha. I've closed #2937 and folded its changes in to this PR so you can just deal with this one. Added the auto-closing reference to #2932.

The braces weren't doing anything. I had just picked them up from the original code on [this SO answer](http://superuser.com/a/328148/7788). Removed now.

I'll add code to address those variant `utf8` locales. Thanks.

mcornella:
Ok! Progress being made makes me happy :smile:

apjanke:
Added the test for the `utf8` variant and fixed a dumb bug in the `iconv` call that would break it for non-UTF8 locales. I think it's ready to go now, if you want to have another look at it.

outpunk:
@apjanke here is the plugins I use:  .
And here is what I get after opening a new tab http://img.madyankin.name/PBdRn.png

apjanke:
I really can't reproduce that on stock Yosemite. I don't think Yosemite has native support for this in zsh.

Here's the behavior I see on a fresh, stock install of Yosemite 10.10.2. I've used `chsh` to change my shell to zsh, but have not installed oh-my-zsh yet. I fire up Terminal and `cd` to a new directory. Note that there's no directory name or icon in the window title bar.

screen shot 2015-02-22 at 10 03 47 am

Then I open a new tab with ⌘-T. No dir icon, and the shell starts in the $HOME instead of ~/foo.

screen shot 2015-02-22 at 10 05 30 am

Now if I switch to `bash`, it works: the dir is noted in the title bar and new tabs open there. But that's only for bash, and it uses the same `/etc/bashrc` support as found in Mavericks.

screen shot 2015-02-22 at 10 07 00 am

screen shot 2015-02-22 at 10 07 09 am

If I install oh-my-zsh, then the directory starts updating under zsh for me. The current `lib/termsupport.zsh` has [Terminal.app directory change support in it](https://github.com/robbyrussell/oh-my-zsh/blob/3704fbb71eb4d886aff11cdaff79a1a23a895cd6/lib/termsupport.zsh#L57-66). I suspect that might be what's getting the directory state to update for you, and one way or another that feature needs to stay in oh-my-zsh.

apjanke:
Any other comments or issues? I think this is ready to merge.

mcornella:
I think this is ok to merge. 

I'd like a set of tests for those `url{en,de}code` functions and a clarification of the difference to #4129. I'd go ahead as well and put them directly into `lib/functions.zsh` instead of making them separate PRs.
Oh, and a deprecation notice in the `terminalapp` plugin would be preferable as well, saying that this plugin no longer contains any functionality.

apjanke:
The ones in #4192 are just newer, better versions of these. If #4192 gets approved to merge, the implementations here should be removed and replaced with calls to the #4192 functions. If #4192 does not get approved, the `omz_urlencode` implementation there should be copied back here to replace the existing `_omz_urlencode` function.

For the deprecation notice: should it just be a comment in an otherwise-empty plugin file, or a warning message output to the user when the plugin is loaded?

mcornella:
An warning message. The idea is that users remove the terminalapp plugin from their configs. After that, a PR removing that entirely should be made in a month or two.

apjanke:
Okay, I've pulled in the core lib urlencode/decode functions from #4129 to this PR. They're now exactly the same so you don't have to worry about comparing them. This PR now auto-closes #4129.

I'd like to hold off on adding a deprecation warning message until a couple weeks after this PR is merged and has had a couple weeks to roll out through updates to all users. Otherwise, there wouldn't be a clean way to share a `~/.zshrc` between different systems without manually updating them all: a Terminal.app user would need to keep `terminalapp` in their `$plugins` list until all their systems had been updated, and would be nagged by warnings on systems which were updated. (And because of stuff like #4072, some users may be stuck on older versions on some of their systems.) For now, I left a comment in the plugin file.

There might be a reason to resurrect the `terminalapp` plugin in the near future, too. The OS X 10.11 El Capitan beta has greatly expanded `/etc/*rc` files that describe several new or previously-undocumented proprietary escape sequences and variables for Terminal.app. Since that goes beyond the broadly-supported functionality in `lib/termsupport.zsh` to advanced program-specific stuff, someone might want to put support for that new stuff in `plugins/terminalapp`.

mcornella:
> I'd like to hold off on adding a deprecation warning message until a couple weeks after this PR is merged and has had a couple weeks to roll out through updates to all users. Otherwise, there wouldn't be a clean way to share a ~/.zshrc between different systems without manually updating them all: a Terminal.app user would need to keep terminalapp in their $plugins list until all their systems had been updated, and would be nagged by warnings on systems which were updated. (And because of stuff like #4072, some users may be stuck on older versions on some of their systems.) For now, I left a comment in the plugin file.

I agree with your reasoning :+1: 

I wanted to comment this on the other PR but I forgot: can you put the `zmodload zsh/langinfo` line inside each function? This way we make sure it is really loaded in case some user has `zmodload -u zsh/langinfo` or similar (which is extremely unlikely I guess, but still). This is just OCD stuff.

lib/termsupport.zsh
```diff
@@ -53,14 +53,27 @@ precmd_functions+=(omz_termsupport_precmd)
...
+
+  # Emits the control sequence to notify Terminal.app of the cwd
+  function update_terminalapp_cwd() {
+    # Identify the directory using a "file:" scheme URL, including
+    # the host name to disambiguate local vs. remote paths.
+
+    # Percent-encode the pathname.
+    local URL_PATH=$(omz_urlencode -P $PWD)
+    [[ $? != 0 ]] && return 1
+    local PWD_URL="file://$HOST$URL_PATH"
```
mcornella: Is the `$HOST` part really necessary? I thought a `file:///...` URL just assumed localhost.

lib/termsupport.zsh
```diff
@@ -53,14 +53,27 @@ precmd_functions+=(omz_termsupport_precmd)
...
+
+if [[ "$TERM_PROGRAM" == "Apple_Terminal" ]] && [[ -z "$INSIDE_EMACS" ]]; then
+
+  # Emits the control sequence to notify Terminal.app of the cwd
+  function update_terminalapp_cwd() {
+    # Identify the directory using a "file:" scheme URL, including
+    # the host name to disambiguate local vs. remote paths.
+
+    # Percent-encode the pathname.
+    local URL_PATH=$(omz_urlencode -P $PWD)
```
mcornella: I haven't tested this, but perhaps `$PWD` should be quoted? Last review stuff only.

lib/functions.zsh
```diff
@@ -73,3 +73,137 @@ function env_default() {
...
+  local LC_ALL=C
+  export LC_ALL
+  
+  # Change + back to ' '
+  local tmp=${encoded_url:gs/+/ /}
+  # Protect other escapes to pass through the printf unchanged
+  tmp=${tmp:gs/\\/\\\\/}
+  # Handle %-escapes by turning them into `\xXX` printf escapes
+  tmp=${tmp:gs/%/\\x/}
+  echo -E "before decode $tmp"
```
mcornella: Should we get rid of these debugging statements as well? Or perhaps we can only output them if there's an error.

mcornella:
I'd say the urlencode/urldecode functions pass the test (even if it's not thorough). Here's two weirdly named folders that are correctly encoded:
```
κόσμε
٩(-̮̮̃-̃)۶ ٩(●̮̮̃•̃)۶ ٩(͡๏̯͡๏)۶ ٩(-̮̮̃•̃).
```
If I then run `open_command $(omz_urlencode -P file://$PWD)` (no quotes needed) the folder is opened with the explorer of my system. I'll also test this on cygwin when I have a minute.

Great job Andrew! :+1: :clap:

lib/termsupport.zsh
apjanke: But the might not be on localhost (from the terminal emulator's perspective) if you are SSHed or otherwise connected to a remote machine.

lib/functions.zsh
apjanke: Ugh, yes. Sorry.

lib/termsupport.zsh
apjanke: The function should have an `emulate -L zsh` to protect it against options affecting word splitting and the like. If we add that, then quotes are probably not needed. I'll put in the `emulate`.

apjanke:
Thanks!

I've added the defensive `emulate`s, removed the debugging output, and fixed a merge conflict with `master`. I think it's ready to go.

WRT putting `zmodload` inside the fuctions, I dunno: I don't think we can cleanly isolate the module loading state from the caller. Unlike `emulate -L` and `local` variables, I think `zmodload` is an inherently global operation. If the caller had done a `zmodload -u`, once the function returns, we will have clobbered their module state. And I think module loading is a relatively expensive operation: you're loading and linking a binary file and running a boot function (which may itself have global side effects). Even if we could restore the module load state, I don't know that you'd want to be doing that at the level of function calls, for performance reasons.

I'd almost lean the other way, and consider gathering all of the `zmodload` calls in core OMZ in one place, so it can serve as a central list of module dependencies for OMZ.

lib/termsupport.zsh
mcornella: Didn't think of that :+1: I can't actually test if this works on a remote machine though.

mcornella:
> If I then run open_command $(omz_urlencode -P file://$PWD) (no quotes needed) the folder is opened with the explorer of my system. **I'll also test this on cygwin when I have a minute.**

I've tested the same command on cygwin and the bad news is that Windows doesn't open those folders if I do `open_command file:///[FOLDER]` (the one with emoticons is actually an invalid folder). The good news is that the encoding that `omz_urlencode` generates is the right one, checked against the Linux output and what Chrome copies if I navigate to that folder and copy the URL.

So that's a big fat :+1: from Windows as well.

> WRT putting zmodload inside the fuctions, I dunno: I don't think we can cleanly isolate the module loading state from the caller. Unlike emulate -L and local variables, I think zmodload is an inherently global operation. If the caller had done a zmodload -u, once the function returns, we will have clobbered their module state. And I think module loading is a relatively expensive operation: you're loading and linking a binary file and running a boot function (which may itself have global side effects). Even if we could restore the module load state, I don't know that you'd want to be doing that at the level of function calls, for performance reasons.
> 
> I'd almost lean the other way, and consider gathering all of the zmodload calls in core OMZ in one place, so it can serve as a central list of module dependencies for OMZ.

Agree :+1:

Good job on the rest, this is ready to rock!
/cc @robbyrussell

lib/termsupport.zsh
apjanke: I've tested it. 

The effect is that when you're ssh'ed in to a remote machine, the remote path is still displayed, and with the correct machine name at the root of the path (when you right-click the folder to display its path), but the folder icons are displayed with "?" on them, and won't open a Finder window to that folder when you click on it. Not very useful, but it prevents the bad case of where you're on a remote machine, and there happens to be a local folder with the same path, and you click on it to open the folder but now you're looking at different places in the terminal and Finder but the UI indicators suggest they're the same place. That could cause confusion or errors if you're not aware of exactly what's going on, and assume that you're looking at the same files. (E.g. if you want to clean up some files on the remote machine, but you end up blowing them away on the local machine, or vice versa.) Super bad if you're ssh'ing from a dev to prod box.

I could see iTerm2 getting creative with ssh use to implement the open-remote-folder-in-Finder functionality. Or maybe even doing it in OMZ with some creative path detection in zsh on the remote side to convert the `file://` URL to an `afp://` one, if there's one available.

Apple's own `/etc/bashrc` includes the hostname too, so this keeps things consistent between `bash` and `zsh`.

In theory, if file sharing is turned on for the other machines, Terminal could be able to open the remote folder when requested. But I haven't been able to get this to work. Probably because AFP shares don't necessarily correspond to the root of the remote filesystem and OS X doesn't have enough info to do that mapping, and OS X (AFAIK) doesn't have the single-sign-on and conventional `\\host\c$` shares that Windows does to make that mapping easy.

apjanke:
> I've tested the same command on cygwin and the bad news is that Windows doesn't open those folders if I do open_command file:///[FOLDER] (the one with emoticons is actually an invalid folder). 

Too many slashes? $PWD includes the leading slash. I got around to testing on cygwin, and if I do `open_command file://$PWD`, it works. But the `open_command $(omz_urlencode -P file://$PWD)` will fail with "file not found" and %s in the error message.

It looks to me like the issue is with `cygstart`: it appears to be interpreting the URL itself and converting it to a Windows file path to pass to Windows Explorer. (Going file URL -> Cygwin/Unix file path -> Windows file path before handing it off to `start` or whatever Windows thing it's calling.)  But it is not interpreting %-escape sequences in the file URL. If you stick the non-ASCII characters in the "URL" directly (which makes it an not a valid URL), it works. I'll see if I can report this as an upstream bug to Cygwin.

Thanks!

apjanke:
Or it might be a Windows thing, too: Windows' own `start` doesn't seem to respect escape sequences in `file://` URLs, and does accept non-ASCII characters. So maybe `cygstart` is converting it back in to a `file://` URL, and we're just seeing normal Windows behavior.

apjanke:
Woohoo, merge party today!

PR #3584, opened 2015-02-15
Fix multibyte escaping in Terminal.app URLs with LC_CTYPE

### LINKED ISSUE #3583: Terminal.app cwd update is garbled for multibyte characters

apjanke:
Fix #3583 - garbled multibyte characters in Terminal.app cwd update control sequences by using LC_CTYPE=C to do byte-by-byte processing, instead of LANG=C, which doesn't seem to work.

This branch is based off #3582's branch, since the LC_CTYPE fix is inside the code that got moved in that PR.


apjanke:
Realized there's another issue: URL-escaping requires the string be in UTF-8 encoding. If the caller's locale is a different encoding, it'll be messed up. Added code to test for the locale's encoding and convert it to UTF-8 if it's a different encoding.

apjanke:
Folded these changes in to PR #3582; closing this one as a duplicate.

PR #3587, opened 2015-02-15
Misc theme plugin cleanup (var and pwd hygiene)

### LINKED ISSUE #3585: theme plugin side effects (cwd & variables) and inconsistency
There are a few related issues in the `themes` plugin. I'm listing them all in one Issue because they're interrelated and I plan on fixing them all in one PR.

* lstheme() changes the pwd as a side effect
* lstheme() does not include custom themes
* theme() leaks variables `$N`, `$themes`, and `$RANDOM_THEME` when setting random theme
* mixed tabs and spaces for indentation
* theme() reads custom themes from `$ZSH_CUSTOM` but not `$ZSH_CUSTOM/themes`.

### lstheme pwd side effect

If you call lstheme, it dumps you in to the themes dir. It shouldn't change the directory.

```
[@ in ~]
~ » pwd
/Users/janke
~ » lstheme > /dev/null
.oh-my-zsh/themes [master] » pwd
/Users/janke/.oh-my-zsh/themes
.oh-my-zsh/themes [master] »
```

### Custom theme locations

It seems like the dir structure of $ZSH_CUSTOM should mirror the directory structure of the main oh-my-zsh directory. Theme() looks under the themes/ subdir of the main dir, but the root dir under $ZSH_CUSTOM. This is also inconsistent with `oh-my-zsh.sh`, which looks in both `/` and `themes/` under $ZSH_CUSTOM.

Really, most of the theme plugin implementation is duplicated by the theme loading section in the `oh-my-zsh.zsh` script. The theme plugin should probably be moved in to `lib/theme-and-appearance.zsh` and `oh-my-zsh.zsh` should call it.

apjanke:
Rewrite of the themes plugin to fix some hygiene and consistency issues.

Fixes #3585.
Close #3337.

There are a few related issues in the `themes` plugin. I'm listing them all in one Issue because they're interrelated and I plan on fixing them all in one PR.

* lstheme() changes the pwd as a side effect
* lstheme() does not include custom themes
* theme() leaks variables `$N`, `$themes`, and `$RANDOM_THEME` when setting random theme
* mixed tabs and spaces for indentation
* theme() reads custom themes from `$ZSH_CUSTOM` but not `$ZSH_CUSTOM/themes`.

### lstheme pwd side effect

If you call lstheme, it dumps you in to the themes dir. It shouldn't change the directory.

```
[@ in ~]
~ » pwd
/Users/janke
~ » lstheme > /dev/null
.oh-my-zsh/themes [master] » pwd
/Users/janke/.oh-my-zsh/themes
.oh-my-zsh/themes [master] »
```

### Custom theme locations

It seems like the dir structure of $ZSH_CUSTOM should mirror the directory structure of the main oh-my-zsh directory. Theme() looks under the themes/ subdir of the main dir, but the root dir under $ZSH_CUSTOM. This is also inconsistent with `oh-my-zsh.sh`, which looks in both `/` and `themes/` under $ZSH_CUSTOM.

Really, most of the theme plugin implementation is duplicated by the theme loading section in the `oh-my-zsh.zsh` script. The theme plugin should probably be moved in to `lib/theme-and-appearance.zsh` and `oh-my-zsh.zsh` should call it.

### DIFF
--- a/lib/_theme
+++ b/lib/_theme
@@ -1,0 +1,3 @@
+#compdef theme
+
+_arguments "1: :($(lstheme | tr "\n" " "))"
--- a/lib/_theme
+++ b/lib/_theme
@@ -1,3 +1,0 @@
-#compdef theme
-
-_arguments "1: :($(lstheme | tr "\n" " "))"
--- a/plugins/themes/_theme
+++ b/plugins/themes/_theme
@@ -1,3 +1,3 @@
 #compdef theme
 
-_arguments "1: :($(lstheme | tr "\n" " "))"
+_arguments "1: :($(lstheme))"
--- a/plugins/themes/themes.plugin.zsh
+++ b/plugins/themes/themes.plugin.zsh
@@ -70,8 +70,8 @@ function _omz_load_theme_from_file() {
     values_before[$param]=${(P)param}
   done
 
-  # Actually load the theme
-  source $file
+  # Actually load the theme, using an indirection function
+  _omz_source_theme_file $file
 
   # Debugging stuff
   if [[ $ZSH_THEME_DEBUG == true ]]; then
@@ -126,6 +126,14 @@ function _omz_load_theme_from_file() {
       echo "WARNING: Theme changed precmd()"
     fi
   fi
+}
+
+# Sources the given file
+# The only reason this function exists is to provide a layer of
+# indirection so that the theme file runs in its own function call stack frame
+# and "local" statements in the theme definitions work as intended.
+function _omz_source_theme_file() {
+  source $1
 }
 
 # Resets all theme settings to their default state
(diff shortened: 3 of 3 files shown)


mcornella:
:ok_hand:

robbyrussell:
Can we get a few testers on this?

apjanke:
I've added some code that attempts to cleanly unload themes before loading a new one. It resets theme-related variables and removes hook functions that the themes install. Makes it more likely that switching between themes inside a session will be correct.

Also added debugging code that lets you see what a theme has changed upon loading. This really is only for debugging use: it seems to cause instability upon loading certain themes like `agnoster` and `dstufft`, including causing the shell to exit or crash. Introduces `$ZSH_THEME_DEBUG` parameter for enabling it; off by default.

apjanke:
The new code also adds another function call of indirection before sourcing the theme definition, so `local` declarations inside the theme definitions work like (I think) they are intended to.

apjanke:
Okay, this Pull Request is getting out of hand. Sorry. I've rebased and squashed the commits along with some new code, and re-submitted it as #3743. Closing this PR in favor of that one.

PR #3590, opened 2015-02-15
Fix composer completion: "\s" is a gawk-specific regexp operator.

voronkovich:
See http://www.gnu.org/software/gawk/manual/gawk.html#Regexp, section: 3.5 gawk-Specific Regexp Operators

### DIFF
--- a/plugins/composer/composer.plugin.zsh
+++ b/plugins/composer/composer.plugin.zsh
@@ -7,7 +7,7 @@
 
 # Composer basic command completion
 _composer_get_command_list () {
-    $_comp_command1 --no-ansi | sed "1,/Available commands/d" | awk '/^\s*[a-z]+/ { print $1 }'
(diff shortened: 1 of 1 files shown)


mcornella:
More specifically, here's the [link to the gawk-specific part](http://www.gnu.org/software/gawk/manual/gawk.html#GNU-Regexp-Operators). 

Can we use a [bracket expression](http://www.gnu.org/software/gawk/manual/gawk.html#Bracket-Expressions) instead? It says those are a POSIX standard

voronkovich:
We can't use character classes because mawk (default awk in Ubuntu http://manpages.ubuntu.com/manpages/lucid/man1/mawk.1.html) doesn't support it. [ \t] would be proper solution. Thanks.

mcornella:
Ok, thanks for taking the time to ensure total compatibility and finding the appropriate documentation, that's what we should all strive for! :sparkles: 

Remember to keep the commit history clean via squash and force push ([example](http://davidwalsh.name/squash-commits-git)).

:+1: 
/cc @robbyrussell

PR #3591, opened 2015-02-15
Bring back -h (--human-readable) option to ls variants

zmwangx:
The `-h` options in `l`, `ll`, and `la` were lost in 25b1cd6 when `ls` variants were factored from `aliases.zsh` into `directories.zsh`. There's no reason to remove the `--human-readable` option.

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -24,9 +24,9 @@ alias d='dirs -v | head -10'
 
 # List directory contents
 alias lsa='ls -lah'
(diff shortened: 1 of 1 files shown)


mcornella:
Agreed, thank you Zhiming. From [this conversation](https://github.com/robbyrussell/oh-my-zsh/commit/25b1cd687b28cb8d05229130ee5a0e2494276894#commitcomment-9747208).
/cc @robbyrussell

PR #3593, opened 2015-02-16
add clear screen and function to create directories and cd into them

codematix:
- Add `c` alias to `clear`
- Add function `mkcd` which makes a directory and cd into it


robbyrussell:
@codematix Why not use the command control-l to clear?

codematix:
`c` still seems easier to get to than `ctrl+i`. Also what about `mkcd`?

PR #3597, opened 2015-02-16
Remove '.' from $cdpath to avoid Cygwin breakage

### LINKED ISSUE #3550: cd:cd:13 in Cygwin root
If I do the following commands:

    cd /
    cd home

I get

    cd:cd:13: no such file or directory: home

If I run

    unfunction cd

I get

    cd: no such file or directory: home

I can do

    cd /home
    cd ./home
    cd ..

and I can `cd` elsewhere

This is on Cygwin x86_64 and zsh 5.0.6. Even with `bin`, `lib` in root this doesn't work.

I will test later on Fedora 21 x86_64.

apjanke:
Fixes #3550. Leave $cdpath at its default (`()`) instead of adding explicit `.`, which runs in to a file normalization bug on cygwin.

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -34,7 +34,6 @@ fi
 
 # disable named-directories autocompletion
 zstyle ':completion:*:cd:*' tag-order local-directories directory-stack path-directories
-cdpath=(.)
 
 # Use caching so that commands like apt and dpkg complete are useable
 zstyle ':completion::complete:*' use-cache 1


mcornella:
One little detail, auto-closing statements like `fix`, `solve`, or `close` should go on the PR description. Otherwise Github doesn't pick it up if it's in the title.
Haven't looked yet at the whole discussion, will comment later.

apjanke:
That explains why it wasn't showing up as a reference in the other Issue. Thanks!

mcornella:
> That explains why it wasn't showing up as a reference in the other Issue. Thanks!

Yeah, I made a little study about that: https://github.com/robbyrussell/oh-my-zsh/pull/2790#issuecomment-43131152

Issue #3579, opened 2015-02-14
AUTO_TITLE functionality breaks in tmux


gotyaoi:
In my tmux config, I have it set up to handle the window title like so:

```
set-option -g set-titles on
set-option -g set-titles-string '#{pane_title}'
set-option -g default-terminal 'screen-256color'
```

That is, get the title from the active pane's title. The function in oh my zsh for setting the window title looks like this:

```
function title {
   if [[ "$DISABLE_AUTO_TITLE" == "true" ]] || [[ "$EMACS" == *term* ]]; then
     return
   fi
   if [[ "$TERM" == screen* ]]; then
     print -Pn "\ek$1:q\e\\" #set screen hardstatus, usually truncated at 20 chars
   elif [[ "$TERM" == xterm* ]] || [[ $TERM == rxvt* ]] || [[ $TERM == ansi ]] || [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
     print -Pn "\e]2;$2:q\a" #set window name
     print -Pn "\e]1;$1:q\a" #set icon (=tab) name
   fi
 }
```

since tmux advises using `screen` or `screen-256color` for your TERM variable, this causes the first if condition to match, and the screen hardstatus doesn't seem to work for tmux, leading to a garbled or unchanging title. For tmux, I think  the statements in the elif branch makes sense, since then things just act like any other shell, which tmux can easily to hook into. Since tmux sets the $TMUX variable, that would probably be an easy way to detect this case, perhaps something like:

```
function title {
   if [[ "$DISABLE_AUTO_TITLE" == "true" ]] || [[ "$EMACS" == *term* ]]; then
     return
   fi
   if [[ "$TERM" == screen* ]] && [[ -z "$TMUX" ]]; then
     print -Pn "\ek$1:q\e\\" #set screen hardstatus, usually truncated at 20 chars
   elif [[ "$TERM" == xterm* ]] || [[ $TERM == rxvt* ]] || [[ $TERM == ansi ]] || [[ "$TERM_PROGRAM" == "iTerm.app" ]] || [[ -n "$TMUX" ]]; then
     print -Pn "\e]2;$2:q\a" #set window name
     print -Pn "\e]1;$1:q\a" #set icon (=tab) name
   fi
 }
```

mcornella:
Hi @gotyaoi, first of all the title functionality has been updated and it's no longer the one you show. That said, I believe the same problem exists. 
[...]

PR #3599, opened 2015-02-16
Update lib/termsupport.zsh to account for tmux

gotyaoi:
Since the default and recommended config for tmux sets TERM to screen or screen-256color, this file was attempting to set tmux's title using screen hardstatus. While tmux can be configured to accept these codes, it makes more sense to simply set the window name normally, so tmux can get the title from the active pane. This commit avoids using screen hardstatus when tmux is detected, and instead uses normal window codes.

Fixes https://github.com/robbyrussell/oh-my-zsh/issues/3579

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -13,9 +13,9 @@ function title {
   # if it is set and empty, leave it as is
   : ${2=$1}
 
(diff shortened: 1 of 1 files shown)


gotyaoi:
This needs a bit more thought actually.

PR #3601, opened 2015-02-17
Create Ubuntu/Debian/Mint/... plugin, based on Debian plugin

nextgenthemes:
* Easy to remember consistent aliases
* Simplified, no more switch between apt-get and aptitude, just uses apt-get ans assumes sudo is installed
* Added aar() function for `sudo apt-add-repository xxxx && apt-get upgrade && apt-get install xxxx`
* More complete autocompletion ;)


mcornella:
Hi @nextgenthemes, in the README file, use capital letters instead of bold parameters in the part where you describe aliases to be more readable in the terminal or text editors without markdown support. 
Will review the rest later this week.

nextgenthemes:
@mcornella Done.

PR #3606, opened 2015-02-19
Adding autoloaded completion plugin for homeshick

arlopurcell:


robbyrussell:
Hi @arloherrine, can you please include a README file with some usage examples?

arloherrine:
Looking through other comparable completion plugins, I couldn't find any with whole README files that I could use as a template. Maybe adding a comment like this with a link to the project's homepage (which itself links to tutorials with usage examples etc.) is enough?

arloherrine:
ping?

TheTechRobo:
If you just added a readme.md file with the link to the homepage, with a short summary of what it does, I think it would be fine.

mcornella:
The upstream project [already has completion](https://github.com/andsens/homeshick/blob/master/completions/_homeshick). Closing this one, but thanks for the contribution!

PR #3608, opened 2015-02-19
Adding ssh-config to vagrant plugin autocomplete. Was missing

idjaw:
A command I personally end up having to use in my day-to-day. I thought maybe others could benefit from it. I found that ssh-config was not autocompleting the list of available vagrant machines. Found it was missing from the list of supported autocomplete commands in the _vagrant script.


idjaw:
Thank you.

Issue #3605, opened 2015-02-19
vi-mode broken since last auto-update?


Cutuchiqueno:
Since last autoupdate I get the following message when I open a terminal: `TRAPWINCH:zle:2: widgets can only be called when ZLE is active`. I figured out that the TRAPWINCH function comes from the vi-mode plugin. To test I deactivated this plugin and the messages disappeared.

I haven't changed anything manually in in my .zshrc after the autoupdate. Before this update there was no error message.

apjanke:
I'm not getting that error when I run `vi-mode`. (OS X 10.9, Terminal or iTerm, robbyrussell or custom theme.) But that TRAPWINCH was indeed added recently to vi-mode (Feb 4). Let's track this down. What OS, terminal program, oh-my-zsh theme, and plugins are you running? Mind posting your .zshrc as a gist?

mcornella:
We're looking at this in https://github.com/robbyrussell/oh-my-zsh/issues/3537#issuecomment-75052757.

scottkidder:
Same issue since upgrade. mcornella's fix in #3537 works for me.
```
TRAPWINCH() {
  [[ -o zle ]] && zle && zle reset-prompt && zle -R
}
```

mcornella:
Can you try the same without the `[[ -o zle]]` part? (line should start
with `zle &&`)

mcornella:
Everybody, please try PR #3610 so we can get it merged.

PR #3610, opened 2015-02-19
Verify zle is active before trying to reset-prompt.

bderrly:
Fix the problem with running zle commands when it is not active. This is to fix https://github.com/robbyrussell/oh-my-zsh/issues/3605.

Relevant documentation (http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html):
With no options and no arguments, only the return status will be set. It is zero if ZLE is currently active and widgets could be invoked using this builtin command and non-zero otherwise. Note that even if non-zero status is returned, zle may still be active as part of the completion system; this does not allow direct calls to ZLE widgets.

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -16,14 +16,7 @@ function zle-keymap-select zle-line-init zle-line-finish {
 
 # Ensure that the prompt is redrawn when the terminal size changes.
 TRAPWINCH() {
-  if [[ -o zle ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
Hey @bderrly, can you keep it short like in https://github.com/robbyrussell/oh-my-zsh/issues/3605#issuecomment-75271013:
```sh
TRAPWINCH() {
  zle && { zle reset-prompt; zle -R }
}
```

bderrly:
Done.

simendsjo:
This fixed my problem too (issue #3620)

PR #3611, opened 2015-02-20
chucknorris: Keep Chuck's fortunes fresh

apjanke:
Adds auto-updating of the compiled fortune data cache file for when the input Chuck string file is updated. Otherwise, if the fortunes are updated, the stale cache file will prevent any new sayings from Mr. Norris from appearing. The source string file would get updated during an oh-my-zsh update, but since the data file was already built, it wouldn't attempt to build it unless user manually removed the old cache file. Better to do it automatically.

To test this, just `touch chucknorris` and open a new zsh login session.

```
$ pwd
/Users/janke/.oh-my-zsh/plugins/chucknorris/fortunes
$ touch chucknorris
$ zsh -l
"/Users/janke/.oh-my-zsh/plugins/chucknorris/fortunes/chucknorris.dat" created
There were 447 strings
Longest string: 374 bytes
Shortest string: 1 byte
$
```

I threw in an anonymous function so it could use local variables, and switched to holding the fortune directory in a local variable for readability, now that more file paths are being used. Spelled-out files everywhere would have made for a very long conditional expression.

### DIFF
--- a/plugins/chucknorris/chucknorris.plugin.zsh
+++ b/plugins/chucknorris/chucknorris.plugin.zsh
@@ -1,6 +1,12 @@
-if [ ! -f $ZSH/plugins/chucknorris/fortunes/chucknorris.dat ]; then
-    strfile $ZSH/plugins/chucknorris/fortunes/chucknorris $ZSH/plugins/chucknorris/fortunes/chucknorris.dat
-fi
+() {
+  # Automatically generate or update Chuck's compiled fortune data file
+  local fdir=$ZSH/plugins/chucknorris/fortunes
+  if [[ ! -f $fdir/chucknorris.dat ]] || [[ $fdir/chucknorris.dat -ot $fdir/chucknorris ]]; then
+    strfile $fdir/chucknorris $fdir/chucknorris.dat
+  fi
 
-alias chuck="fortune -a $ZSH/plugins/chucknorris/fortunes"
-alias chuck_cow="chuck | cowthink"
+  # Aliases
+  alias chuck="fortune -a $fdir"
+  alias chuck_cow="chuck | cowthink"
+}
+
--- a/plugins/chucknorris/chucknorris.plugin.zsh
+++ b/plugins/chucknorris/chucknorris.plugin.zsh
@@ -1,12 +1,11 @@
-() {
-  # Automatically generate or update Chuck's compiled fortune data file
-  local fdir=$ZSH/plugins/chucknorris/fortunes
-  if [[ ! -f $fdir/chucknorris.dat ]] || [[ $fdir/chucknorris.dat -ot $fdir/chucknorris ]]; then
-    strfile $fdir/chucknorris $fdir/chucknorris.dat
-  fi
+# Automatically generate or update Chuck's compiled fortune data file
+DIR=${0:h}/fortunes
+if [[ ! -f $DIR/chucknorris.dat ]] || [[ $DIR/chucknorris.dat -ot $DIR/chucknorris ]]; then
+  strfile $DIR/chucknorris $DIR/chucknorris.dat
+fi
 
-  # Aliases
-  alias chuck="fortune -a $fdir"
-  alias chuck_cow="chuck | cowthink"
-}
+# Aliases
+alias chuck="fortune -a $DIR"
+alias chuck_cow="chuck | cowthink"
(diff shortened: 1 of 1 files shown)


apjanke:
Hi @maff, I put together auto-updating for your `chucknorris` plugin in case any new facts about Mr. Norris come to light. Any comments on this code change?

mcornella:
Things I'd do to the chucknorris plugin:
1. Put the `.dat` file in `$ZSH/cache`.
2. Put the chucknorris fortune file at the same level as the `plugin.zsh` file.
3. Delete `.gitignore` and `LICENSE` files and put the content of the latter as comments in the `plugin.zsh` file.
4. Use `DIR="${0:h}"` to get the directory of the plugin instead of using the hardcoded path from `$ZSH`. This is tied to your proposal of using `$fdir` for better reuse.

apjanke:
Those all sound good to me. I'll try to add item 4 (using `$0` with `$fdir`) to my PR. Going to defer to @maff on the rest since they're design changes.

I can't use `$0` along with the anonymous function (which provides scope for the `local`); `$0` gets set to `(anon)`. Know a similar structure that'd work, and avoid polluting the caller's namespace?

mcornella:
You can use `unset` to remove it from the global scope. In this case, it's easy :+1: 
BTW, I had in mind doing a similar modification on the rest of the plugins that use relative paths, but I never got to it I think.

apjanke:
The issue with using `unset` is that it would clobber any existing `$fdir` that the user or caller had defined. I was trying to keep it in local scope to completely avoid altering the caller's variable namespace.

mcornella:
You're right, but there's certain things that we just can't avoid. There's a whole bunch of plugins that use the `$DIR` variable, so if you feel more comfortable using that...

apjanke:
I see. Darn it. Okay, going with `$DIR` and `unset`.

I like moving the `.dat` file to `$ZSH_CACHE_DIR`. As is, when I switch to a branch for a pending pull request or check out an older version of the source, the dat file shows up as a new untracked file because the `.gitignore` for it doesn't exist on the other branches or in the past.

apjanke:
Question on the license: can we use GPL code in oh-my-zsh? According to the root dir, oh-my-zsh is under the more liberal MIT license.

mcornella:
> I like moving the `.dat` file to `$ZSH_CACHE_DIR`

Thanks for remembering to say it's `$ZSH_CACHE_DIR`, not `$ZSH/cache`. My brain is probably in liquid state right now. 

> Question on the license: can we use GPL code in oh-my-zsh? According to the root dir, oh-my-zsh is under the more liberal MIT license.

There was some controversy around that. #2587 was the original discussion, #2588 where we ended up removing all the copyright and licensing stuff. You can see in #2588 that there was an Apache license and what I believe to be GPL. 
We will probably remove this one too but I'd like @maff's take on this.

maff:
@apjanke +1 to this change. Originally I just included the .dat file together with the source, but that didn't work for different systems (the .dat file created on OSX didn't work on linux). I guess it's something encoding-related, however autogenerating the file seems cleaner to me.

@mcornella Thanks for your suggestions, they all look good to me. As I never submitted anything to oh-my-zsh any input is appreciated :)

Regarding the GPL issue: I'm fine with removing the LICENSE file and the GPL reference - I just inherited the license from the Gentoo package I got the fortune file from. What do you think? Can we just remove the license from the plugin and reference the original package in a comment in the plugin file?

apjanke:
Well, the Gentoo package originates from http://www.k-lug.org/~kessler/projects.html, and the original download there is GPL licensed. We're only using the data file from it, which is a compilation of text from http://www.chucknorrisfacts.com/, which has no licensing info at all, and who knows where they sourced it from. I'm not a lawyer, so I don't know what the impact of this is, or what we should do about it. I'd say contact the Paul Kessler who made that original package, but there's no contact info in the README or on his website, so...???

apjanke:
Okay, I have an idea. I wrote a little scraper script to rebuild the fortune file from chucknorris.com directly, and built a new Chuck fortune file using that, in #3642. Now we have a fortune file of known provenance and there's nothing left from the original GPL-ed licensed distribution. I think if you merge that, we can safely remove the GPL licensing notices because we have nothing left from that package.

Issue #2586, opened 2014-03-08
install.sh: `PATH=` with no `$PATH` inside


Jaspur:
The https://raw.github.com/robbyrussell/oh-my-zsh/master/tools/install.sh file gives an error:

```
Cloning Oh My Zsh...
Cloning into '/home/vagrant/.oh-my-zsh'...
Looking for an existing zsh config...
Using the Oh My Zsh template file and adding it to ~/.zshrc
Copying your current PATH and adding it to the end of ~/.zshrc for you.
Time to change your default shell to zsh!
Password: chsh: PAM authentication failed
         __                                     __
  ____  / /_     ____ ___  __  __   ____  _____/ /_
 / __ \/ __ \   / __ `__ \/ / / /  /_  / / ___/ __ \

  * WARNING: Above files contains `PATH=` with no `$PATH` inside, this can break RVM,
    for details check https://github.com/wayneeseguin/rvm/issues/1351#issuecomment-10939525
    to avoid this warning append #PATH.

No binary rubies available for: ubuntu/12.04/x86_64/ruby-2.1.1.
Continuing with compilation. Please read 'rvm help mount' to get more information on binary rubies.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11.4M  100 11.4M    0     0  3841k      0  0:00:03  0:00:03 --:--:-- 6780k
/home/vagrant/.oh-my-zsh/oh-my-zsh.sh: line 13: syntax error near unexpected token `('
/home/vagrant/.oh-my-zsh/oh-my-zsh.sh: line 13: `for config_file ($ZSH/lib/*.zsh); do'
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions into the /opt/vagrant_ruby/lib/ruby/gems/1.8 directory.

```

It refers to https://github.com/wayneeseguin/rvm/issues/1351#issuecomment-10939525 , but from what I can see that is already done at line 30 of `install.sh`

PR #3613, opened 2015-02-20
Installer: Remove "sed -i" for portability and hygiene

### LINKED ISSUE #3367: Installation fails on BSD
The installation fails on OpenBSD. Maybe there's an assumption that sed is GNU sed?

```
$ curl -L http://install.ohmyz.sh | sh
100   115  100   115    0     0    628      0 --:--:-- --:--:-- --:--:--   858
100  1906  100  1906    0     0   5015      0 --:--:-- --:--:-- --:--:-- 12878
Cloning Oh My Zsh...
Cloning into '/home/hugo/.oh-my-zsh'...
remote: Counting objects: 11280, done.
remote: Total 11280 (delta 0), reused 0 (delta 0)
Receiving objects: 100% (11280/11280), 2.05 MiB | 950.00 KiB/s, done.
Resolving deltas: 100% (6173/6173), done.
Checking connectivity... done.
sed: unknown option -- i
usage: sed [-aEnru] command [file ...]
       sed [-aEnru] [-e command] [-f command_file] [file ...]
```

```
$ uname -a
OpenBSD elysion.barrera.io 5.6 GENERIC.MP#632 amd64
$ zsh  --version
zsh 5.0.5 (x86_64-unknown-openbsd5.6)
```

### LINKED ISSUE #3612: Installer leaks "~/.zshrc-e" file, maybe from bad "-i" option
The oh-my-zsh installer is leaving behind apparently unused files named `~/.zhsrc-e` that appear to be copies of the zshrc template, only partially processed. I think this is due to an issue with the options passed to `sed` by the installer.

Consider the `sed` calls at lines 26 and 31.
```shell
sed -i -e "/export PATH=/ c\\
export PATH=\"$PATH\"
" ~/.zshrc
```

That's broken over multiple lines, but it's of the form `sed -i -e "blah_blah" ~/.zshrc`. The problem is that the `-i` option takes an argument, which is the suffix to use for creating backup files. Here, that `-e` is interpreted as the argument to `-i`, not a separate option. So it's creating a backup file ending in `-e` every time it's called. For in-place editing with no backup, you need to give a zero-length suffix with an explicit `""` empty string: `sed -i -e "blah-blah" ~/.zshrc`.

It might be better to avoid `-i` altogether and use explicit temp files, because `-i` is a nonstandard extension that's not supported on all systems. (See #3367.)

apjanke:
Fixes #3367 
Fixes #3612
Closes #4186

The installer uses the nonstandard `sed -i` extension for in-place editing. This makes it not portable to systems whose `sed` doesn't have the `-i` extension, in particular OpenBSD (#3367). This replaces the `sed -i` with explicitly-managed temp files.

This also fixes the problem of leaking `~/.zshrc-e` files due to incorrect use of the `-i` option in the current installer (#3612).

###   Testing   ###

To test this, run the following in a user account that does not already have OMZ installed. (Careful... it seems a little difficult to copy and paste from this little field in some browsers.)

```
sh -c "$(curl -fsSL https://raw.githubusercontent.com/apjanke/oh-my-zsh/remove-sed-i/tools/install.sh)"
```

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -23,14 +23,16 @@ fi
 
 echo "\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m"
 cp $ZSH/templates/zshrc.zsh-template ~/.zshrc
-sed -i -e "/^export ZSH=/ c\\
+sed "/^export ZSH=/ c\\
 export ZSH=$ZSH
-" ~/.zshrc
+" ~/.zshrc > ~/.zshrc-omztemp
+mv -f ~/.zshrc-omztemp ~/.zshrc
 
 echo "\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m"
-sed -i -e "/export PATH=/ c\\
+sed "/export PATH=/ c\\
 export PATH=\"$PATH\"
-" ~/.zshrc
+" ~/.zshrc > ~/.zshrc-omztemp
+mv -f ~/.zshrc-omztemp ~/.zshrc
 
 if [ "$SHELL" != "$(which zsh)" ]; then
     echo "\033[0;34mTime to change your default shell to zsh!\033[0m"


hobarrera:
LGTM. Thanks.

mcornella:
see #2589
/cc @avit

apjanke:
Thanks. I hadn't noticed that because I had only been searching in issues and not PRs. It fixes this issue.

Closing as a duplicate of #2589.

apjanke:
I'm re-opening this, because the more extensive #2589 is probably never going to get merged in. I think we should fix the `sed` portability issue in the mean time.

hobarrera:
FWIW, OpenBSD's sed now implements `-i`.

apjanke:
I think we should still avoid using it, so it'll work on older systems, or other ones we're not aware of. It's still a nonstandard extension, and apparently easy to misuse.

mcornella:
Ok this has been open for too long, let's pull this one. Please autoclose #4186 
I'm also considering either:

a) commenting out the PATH export and keep it there for people to modify it manually.
b) putting literal `$PATH` instead of the expanded variable to solve #1359, #2586 and #3960.
c) a combination of both.

The PATH export should go at the top as well. As is now, the PATH is modified after oh-my-zsh is sourced which makes some plugins not work (for instance, npm). 

This is a long standing issue and we should solve it promptly once and for all. What do you think?

/cc @robbyrussell

apjanke:
Added the autoclose reference.

I think it would make more sense to handle the `$PATH` stuff under a different PR, since it's unrelated to `sed` portability and it's a significant and contentious change (based on #1359's 3-year comment history) that doesn't IMHO have an obvious correct or best fix. I think the `$PATH` handling could be improved, but I'd rather not to tie a trivial `sed` usage change to all that.

mcornella:
> I think it would make more sense to handle the `$PATH` stuff under a different PR.

Yes, sorry that wasn't clear. I'm not satisfied with any of the currently proposed changes and I'd like to make one that fixes all currently open issues.

> Added the autoclose reference.

Thanks, this is now a merge candidate :+1: 
/cc @robbyrussell

PR #3616, opened 2015-02-20
Plugins: vi-mode: extra vi-like bindings, vi-like commands clipboard 

alx741:
- Added standard `ctrl-r` for incremental history searching
- Added `G` `gg` `u` `^-R` vim commands behavior
- Added X clipboard interaction with standard vim commands ( y/p/d/c )
- From zsh wiki vim bindings [examples|suggestions?]

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -63,3 +63,45 @@ function vi_mode_prompt_info() {
 if [[ "$RPS1" == "" && "$RPROMPT" == "" ]]; then
   RPS1='$(vi_mode_prompt_info)'
 fi
+
+
+# Allow Copy/Paste with the system clipboard
+# behave as expected with vim commands ( y/p/d/c/s )
+[[ -n $DISPLAY ]] && (( $+commands[xclip] )) && {
+
+  function cutbuffer() {
+    zle .$WIDGET
+    echo $CUTBUFFER | xclip -selection clipboard
+  }
+
+  zle_cut_widgets=(
+    vi-backward-delete-char
+    vi-change
+    vi-change-eol
+    vi-change-whole-line
+    vi-delete
+    vi-delete-char
+    vi-kill-eol
+    vi-substitute
+    vi-yank
+    vi-yank-eol
+  )
+  for widget in $zle_cut_widgets
+  do
+    zle -N $widget cutbuffer
+  done
+
+  function putbuffer() {
+    zle copy-region-as-kill "$(xclip -o)"
+    zle .$WIDGET
+  }
+
+  zle_put_widgets=(
+    vi-put-after
+    vi-put-before
+  )
+  for widget in $zle_put_widgets
+  do
+    zle -N $widget putbuffer
+  done
+}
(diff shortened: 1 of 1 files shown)


regagain:
ctrl-r for history searching & the extra vim-like bindings work well and are much appreciated (I have not tested the copy/paste from the clipboard).

alx741:
A way copy/paste can be used/tested (that is actually useful) is by copy an URL from the browser and then in the terminal (<ESC> is used to get into normal mode -of course-):

    $ wget <ESC>p

Another example for youtube video downloading:

    $ youtube-dl <ESC>p

regagain:
OK! I've tested it now and it's really cool. One strange thing, though: it seems `yy` cease to properly update the vi-like buffer once something has been copied from the clipboard.
For example:
`ctrl+c` from the browser or anything else, then:
```
$ <ESC>p
```
results in
```
$ from_browser
```
But after this:
```
$ from_zsh<ESC>yy
$ <ESC>p
```
still results in
```
$ from_browser
```
(but `ctrl+v` is still working as expected, i.e. gives `from_zsh`).

alx741:
Thank you very much for testing,

I didn't notice that, it's strange indeed... I'll check it out!

djpresent:
Excelente ... +1

Alibru:
Interesting and useful! +1

alx741:
@0arthur  Added in pull request message, but.. do you mean inside the file?

0arthur:
Daniel,
This is fine.

Cheers,
Arthur

On Sun, Mar 29, 2015 at 16:33 -0700, Daniel Campoverde Carri�n wrote:
>@0arthur  Added in pull request message, but.. do you mean inside the file?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/robbyrussell/oh-my-zsh/pull/3616#issuecomment-87497578

-- 
Arthur Noel
 arthur@0compute.net
 +44 (0)7967 385514

regagain:
Hello there!

I've been using this for almost two months now, and find it useful & reliable. Can we please merge it?

Thanks,

rosshadden:
The X clipboard functionality should be broken out into a separate merge request, so as not to prevent this one from being merged.  There are many things that need to be discussed and ironed out.

For instance, it is unexpected that `y, p, d, c` would alter the system clipboard.  For one thing this adds an `xclip` dependency, but more importantly that's just not how `vim` itself works.  When you use these operations in `vim`, it uses the internal register just like `zsh`'s vi-mode does natively.  To get X integration you tell it to use the `"` register (for example).  Having `vim` use the system clipboard for all change/delete operations does not sound like a good idea, as you would be unnecessarily polluting the clipboard.

Now that being said, it would be very useful to make `vicmd` commands that yank/put with the system clipboard, which users could bind themselves, or this project could bind to a different set of keys.

afwlehmann:
The latest changes seem to have happened quite some time ago. Since I'm very much interested in vim-like copy/paste functionality, may I ask whether there's any progress or related solution?

felipe-vieira0:
@alx741  ,

Unfortunately, this is interfering with navigation (the up and down keys do not search correctly anymore). I would love to see this implemented.

Thanks!

alx741:
@felipe-vieira0   yeah... I suppose at this point its impossible to make it work without pretty much re-implementing it, I stopped using oh-my-zh years ago though and went with just a personal .zshrc for improved performance. So I guess there will not be further progress from my side on this... sorry about that.

Issue #3282, opened 2014-11-02
Load custom configs from $ZSH_CUSTOM before loading plugins defined in ~/.zshrc?


rhbecker:
I'm new to oh-my-zsh. One of the first plugins I'm messing with is ssh-agent.

At first blush, my intuition is to keep the setting of zstyles - e.g. `zstyle :omz:plugins:ssh-agent identities example-key` - out of my main `~/.zshrc` file, rather putting them into `*.zsh` files in my `$ZSH_CUSTOM` directory.

However, for that to work, I need the `$ZSH/.oh-my-zsh.sh` file to load those custom configs before loading plugins.

Am I stupidly breaking something by doing that? Or would that be a good idea for everyone - i.e. something I should submit as a PR?

mcornella:
Hey @rhbecker can you close this issue now that your use case is solved?

rhbecker:
When you say "your use case is solved", are you referencing the workaround you suggested in your November 7th comment?

mcornella:
Yeah exactly. It's conceptually an aberration but I think it might work... unless the init script calls it again after sourcing oh-my-zsh. I don't know...

I think we merged quite recently a PR to allow using custom lib files, so you could have a particular `$ZSH_CUSTOM/lib/something.zsh` sourced before the rest. Take a look at the init script for more info.

In the end, I think none of these solutions is good enough. Maybe we could standardise some file in the custom directory to run before everything else.. like a pre-init.zsh file. But I'm not a 100% on that either. You might want to reconsider putting it in the zshrc file.

BTW your applied solution of loading custom files first breaks the use case of overriding plugin or libs settings (I'm talking aliases and such).

Sorry for the big reply :grin:

apjanke:
This was merged in #3617 / #3809. Custom configs are now loaded before plugins.

Could you close this issue?

rhbecker:
Cool! I haven't confirmed this solves my issue yet, but I'll assume so and re-open if not. Thanks!

PR #3617, opened 2015-02-21
Move custom *.zsh file sourcing up so that they can load plugins

MatthewHager:
I want use custom/*.zsh files to load plugins so that I can create one file that both loads and customizes a plugin. For example:

```
plugins+=(vi-mode)
bindkey -M viins 'jj' vi-cmd-mode
```

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -29,6 +29,11 @@ for config_file ($ZSH/lib/*.zsh); do
   source $config_file
 done
 
+# Load all of your custom configurations from custom/
+for config_file ($ZSH_CUSTOM/*.zsh(N)); do
+  source $config_file
+done
+unset config_file
 
 is_plugin() {
   local base_dir=$1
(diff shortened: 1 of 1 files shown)


gotyaoi:
Works for me. Sorta nice in that everything doesn't just have to be placed in .zshrc.

DanyHunter:
Can confirm it works and it is a nice change. :+1:

MatthewHager:
Please accept this PR. We're running our project off a fork because it isn't usable without this simple (but useful) change. Thanks!

ncanceill:
I cannot think of any case where that would break anyone's config, so if no one else can this should be good to go.

@robbyrussell can we merge this?

apjanke:
BTW, the [old Customization wiki page](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization/4108a7c747d9b6586d90c1e01d26c6d9a4b5a54b) had documented the load order as being "custom code after plugins". If changing other stuff in the oh-my-zsh load sequence, please update the doco on the Customization and/or Design wiki pages (and maybe the main `README.md`) to match.

apjanke:
One of the use cases for this: #3282.

PR #3607, opened 2015-02-19
Revert usage of install.ohmyz.sh URL, until HTTPS support.

mcornella:
This reverts commit b6cbba9dfee5985232b473e94595c2fb1ab83715.

Use full Github URL until support for HTTPS in [ohmyz.sh](http://ohmyz.sh) is available.

Close #3459
Close #3461

PR #3618, opened 2015-02-21
Disable --no-check-certificate flag on wget

mcornella:
It was implemented in #258 to solve github.com's use of a wildcard certificate but it's no longer needed and it poses a security risk as commented on https://github.com/robbyrussell/oh-my-zsh/pull/3607#issuecomment-75392512.

/cc @robbyrussell

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -22,7 +22,7 @@ h4. via @curl@:
 
 h4. via @wget@:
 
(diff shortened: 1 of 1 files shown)


robbyrussell:
I handled this in the new markdown file

PR #3622, opened 2015-02-22
Re-order OS detection if..else statement in bgnotify plugin.

koddsson:
This patch will re-order the if statements when checking what operating system the user is running.

This is useful for people that already have a `notify-send` binary in their $PATH but are running OSX. An example of such requirements is anyone running `vagrant-notify` is required to have a `notify-send` binary on their $PATH.

See [link](https://github.com/fgrehm/vagrant-notify#os-x).

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -1,6 +1,5 @@
 #!/usr/bin/env zsh
 
-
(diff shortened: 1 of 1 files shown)


robbyrussell:
@koddsson Can you rebase real quick for me? I just pulled in another change to bgnotify

koddsson:
@robbyrussell Done!

PR #3623, opened 2015-02-22
Add nmap plugin with some aliases for scan modes

uberspot:
Just some aliases that could be useful for people frequently scanning with nmap.
Most of these are easy to type while autocompleting with TAB.
I'm not sure if this is the proper way to contribute this to the repo.
If you have any suggestions do let me know.


robbyrussell:
@uberspot could you include a README with this?

uberspot:
Done

robbyrussell:
Beautiful. Thanks!

PR #3627, opened 2015-02-22
Better research history with arrow keys

GNUletik:
Enable to research history with arrow keys. 
For example, you can do a "git clone [UP-ARROW]" to get all the previous "git clone" commands. And not only the "git" commands as previously.


apjanke:
Why is this commenting out the `zle-line-init` and `zle-line-finish` widgets? Those are required for proper terminfo support.

lib/key-bindings.zsh
```diff
@@ -23,16 +23,26 @@ bindkey '^r' history-incremental-search-backward      # [Ctrl-r] - Search backwa
...
-fi
-if [[ "${terminfo[kcud1]}" != "" ]]; then
-  bindkey "${terminfo[kcud1]}" down-line-or-search    # start typing + [Down-Arrow] - fuzzy find history backward
-fi
+# New functionality !
+autoload -U up-line-or-beginning-search
+autoload -U down-line-or-beginning-search
+zle -N up-line-or-beginning-search
+zle -N down-line-or-beginning-search
+bindkey "^[[A" up-line-or-beginning-search # Up
```
apjanke: This should use `${terminfo[xxxx]}` lookups instead of hardcoded escape sequences so it is portable across multiple terminal types. The previous up/down-arrow bindings were using a good form for that.

lib/key-bindings.zsh
```diff
@@ -23,16 +23,26 @@ bindkey '^r' history-incremental-search-backward      # [Ctrl-r] - Search backwa
 if [[ "${terminfo[kpp]}" != "" ]]; then
   bindkey "${terminfo[kpp]}" up-line-or-history       # [PageUp] - Up a line of history
 fi
-if [[ "${terminfo[knp]}" != "" ]]; then
-  bindkey "${terminfo[knp]}" down-line-or-history     # [PageDown] - Down a line of history
```
apjanke: The PageDown binding is commented out but no new one provided?

lib/key-bindings.zsh
GNUletik: If someone know how.. :)

GNUletik:
Hi ! I was busy, and then forget.. But here it is !

- [x] The PageDown bindings is not anymore commented
- [x] I'm now using ```${terminfo[xxxx]}``` instead of hardcoded escape sequencies

You can now do (for example)
```git clone [UP arrow]``` and search for your last ```git clone``` commands instead of your last ```git``` commands !

PR #3632, opened 2015-02-24
Add match for `MM` in git prompt status

rummik:
Add match for `MM` (file modified, some modifications added, but modifications still exist) to git_prompt_status.

Steps to repro:

```
cd /tmp
mkdir foo$$
cd foo$$
git init
touch foo
git add foo
git commit -m foo
echo 'asdf' > foo
git add foo
echo 'asdf' >> foo
```

Expected:
Shows modifications added, and modifications exist.

Actual:
The repo is shown as dirty, but no reason for it being dirty.


jalada:
How come this has never been merged in?

Ma27:
@robbyrussell could you merge this please? I stumble upon this issue quite often and fixing it would be a great improvement :-)

PR #3636, opened 2015-02-24
cask: check for Homebrew path for completion file

### LINKED ISSUE #3629: cask.plugin.zsh sourcing wrong path
`cask.plugin.zsh` tries to source `/usr/local/etc/cask_completion.zsh`, but in my homebrew, it's located at `/usr/local/share/zsh/site-functions/cask_completion.zsh`.

Perhaps because I have `zsh-completions` installed?

At any rate, it'd be cool if the plugin checked that the file exists before sourcing it.

apjanke:
Fixes #3629.

Mac Homebrew installs the Cask zsh completion files to a different location (relative to the `cask` binary) than the plain Cask installation does. Add code to detect homebrew and check its cask completion location, too.

I've tested on OS X 10.9, and it works for me for regular Cask, Homebrew Cask, and having both installed. For the regular Cask, I had to grab version 0.7.2 instead of the head of master, because they're recently removed the zsh completions from the main Cask project. To install non-homebrew Cask version 0.7.2 for testing, do this:

```
cd
git clone https://github.com/cask/cask.git .cask
cd .cask
git checkout tags/v0.7.1
export PATH="~/.cask/bin:$PATH"
```

And then do `zsh -l` from that session. Or add that PATH statement in your `~/.zshrc` so that all new sessions pick it up.

### DIFF
--- a/plugins/cask/cask.plugin.zsh
+++ b/plugins/cask/cask.plugin.zsh
@@ -1,5 +1,21 @@
-if which cask &> /dev/null; then
-  source $(dirname $(which cask))/../etc/cask_completion.zsh
-else
-  print "zsh cask plugin: cask not found"
-fi
+() {
+  if which cask &> /dev/null; then
+    local cask_bin cask_base f comp_files
+    cask_bin=$(which cask)
+    cask_base=${cask_bin:h:h}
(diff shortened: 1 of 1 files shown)


mcornella:
Merged on f258083.

PR #3637, opened 2015-02-24
Plugin to create bookmarks in the file system.

amilaperera:
This plugin can primarily be used to add bookmarks which
makes it easier to navigate through the file system.
The added bookmarks can also be listed, deleted, renamed and modified.

After enabling this plugin, '**bm**' command is created.
Use '**bm**' to create bookmarks in the file system, manage those
bookmarks and navigate through the file system using the created bookmarks.
To invoke the basic usage of the command, type the following in command line.

> % bm -h

### DIFF
--- a/plugins/bookmark/bookmark.plugin.zsh
+++ b/plugins/bookmark/bookmark.plugin.zsh
@@ -1,0 +1,223 @@
+#############################################################
+# Author: Amila Perera
+# File Name: bookmark.plugin.zsh
+#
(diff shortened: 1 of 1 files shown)


mcornella:
We already have [a bunch of those](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins-Overview#fs-jumping), if you want you can add your plugin to the [external plugins list](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins).

PR #3639, opened 2015-02-25
Add curl plugin

ismaelpuerto:
# Curl aliases plugin
## Functions explained
- curlj - pretty json (require Python 2.6)
## Aliases explained
- curlh - show header only
- curls - download filename
- curlt - show response time in milliseconds
- curlrc - show http code

### DIFF
--- a/plugins/curl/README.md
+++ b/plugins/curl/README.md
@@ -1,0 +1,5 @@
+# Curl aliases plugin
+
+## Functions explained
+
+## Aliases explained
--- a/plugins/curl/README.md
+++ b/plugins/curl/README.md
@@ -2,4 +2,11 @@
 
 ## Functions explained
 
+ * curlj - pretty json (require Python 2.6)
+
 ## Aliases explained
+
+ * curlh - show header only
+ * curls - download filename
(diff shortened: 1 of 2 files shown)


tomasnorre:
I would like to have the Arguments completion too..
Should I push a commit afterwards when this gets merged or how to do this ?

robbyrussell:
⌛️ Would like to see more upvote/interest in this before we consider a merge

plugins/curl/README.md
```diff
@@ -0,0 +1,12 @@
...
+## Functions explained
+
+ * curlj - pretty json (require Python 2.6)
+
+## Aliases explained
+
+ * curlh - show header only
+ * curls - download filename
+ * curlt - show response time in milliseconds
+ * curlrc - show http code
```
TheTechRobo, line 12: I think it would be better if you mentioned ONLY show http code and discard the rest. People might be expecting otherwise. Otherwise, LGTM!

TheTechRobo (commented):
I just have a tinyyyyyyy documentation suggestion.

PR #3642, opened 2015-02-27
Make a personal connection with chucknorris

apjanke:
Add script to re-scrape chucknorris fortunes from chucknorris.com.
Rebuild the chucknorris fortune file with it, using 20 pages worth.

This replaces the original chucknorris fortune file which was drawn from a distribution with a GPL license. With this change, there are no files left from the GPL-ed package, and I think we can safely remove the GPL license notice from the oh-my-zsh code and not worry about GPL contamination.

### DIFF
--- a/plugins/chucknorris/LICENSE
+++ b/plugins/chucknorris/LICENSE
@@ -1,2 +1,0 @@
-License: GPL v2
-Thanks to http://www.k-lug.org/~kessler/projects.html for the fortune file.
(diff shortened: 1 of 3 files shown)


apjanke:
Note: When testing this, be aware that until #3611 is merged, you must manually delete the `fortunes/chucknorris.dat` file to get it to be regenerated from the updated fortunes source file that `scrape-norris.zsh` writes out.

apjanke:
This PR now removes the GPL LICENSE file itself, for convenience and to make it clear what's going on.

Also added a data validation pass to the scraper script, to ensure the integrity of the generated Facts database file.

maff:
Sorry for the late response...looks good to me :)

mcornella:
I resolved the conflicts and finally merged this puppy. I had to remove the scraper because chucknorrisfacts.com is no longer up.

PR #3646, opened 2015-02-28
Add Cygwin support to installer.

### LINKED ISSUE #3588: Installer fails on cygwin with git and chsh errors
I tried doing a fresh install of oh-my-zsh in Cygwin on Windows 7 x64, and it failed to install, with kind of cryptic messages. Both my Windows and cygwin installations are up to date as of 2/14/2015.

If you have Git for Windows installed and on the path, but not Cygwin's git installed, `install.sh` will use the Windows git, and for me it failed to clone the OMZ repo. Looks like Cygwin's git must be installed for `install.sh` to work.

If `install.sh` makes it past the repo cloning, it will then fail when trying to change the shell, because `chsh` does not exist on Cygwin.

##  What the failure looks like

Here's me running install when Windows Git 1.9.4 but not Cygwin git is installed.

```
gwydion% which git
/cygdrive/c/Program Files (x86)/Git/cmd/git
gwydion% git --version
git version 1.9.4.msysgit.0
gwydion% curl -L http://install.ohmyz.sh | sh
100   115  100   115    0     0    149      0 --:--:-- --:--:-- --:--:--   393
100  1930  100  1930    0     0   2014      0 --:--:-- --:--:-- --:--:--  2014
\033[0;34mCloning Oh My Zsh...\033[0m
Cloning into '/home/janke/.oh-my-zsh'...
\033[0;34mLooking for an existing zsh config...\033[0m
\033[0;33mFound ~/.zshrc.\033[0m \033[0;32mBacking up to ~/.zshrc.pre-oh-my-zsh\033[0m
\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m
cp: cannot stat ‘/home/janke/.oh-my-zsh/templates/zshrc.zsh-template’: No such file or directory
gwydion%
```

With Windows Git 1.9.5 installed, it fails at the same point but you get a different error message.

```
gwydion% curl -L http://install.ohmyz.sh | sh
100   115  100   115    0     0     80      0  0:00:01  0:00:01 --:--:--   155
100  1930  100  1930    0     0    643      0  0:00:03  0:00:03 --:--:--  1783
\033[0;34mCloning Oh My Zsh...\033[0m
fatal: destination path '/home/janke/.oh-my-zsh' already exists and is not an empty directory.
git not installed
gwydion%
```

Despite what the error message says, '/home/janke/.oh-my-zsh' did not in fact exist.

I think what's going on here is that git 1.9.5 is now returning an exit status in this case. The code in install.zsh is lumping any error from the git invocation together and reporting it as not being installed. This should probably be two separate checks.
```
hash git >/dev/null 2>&1 && env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git $ZSH || {
  echo "git not installed"
  exit
}
```

I was able to get the clone to work using Windows git if I gave it a relative path to clone to. May be something with how Windows git is handling Unix style paths.
```
gwydion% env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git $ZSH
fatal: destination path '/home/janke/.oh-my-zsh' already exists and is not an empty directory.
gwydion% file /home/janke/.oh-my-zsh
/home/janke/.oh-my-zsh: cannot open `/home/janke/.oh-my-zsh' (No such file or directory)
gwydion% env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git .oh-my-zsh
Cloning into '.oh-my-zsh'...
gwydion%
```

Installing Cygwin git fixes that issue. After that, it will get most of the way through the install, and then break when trying to call chsh.
```
gwydion% which git
/usr/bin/git
gwydion% curl -L http://install.ohmyz.sh | sh
...
\033[0;34mCloning Oh My Zsh...\033[0m
Cloning into '/home/janke/.oh-my-zsh'...
...
\033[0;34mLooking for an existing zsh config...\033[0m
\033[0;33mFound ~/.zshrc.\033[0m \033[0;32mBacking up to ~/.zshrc.pre-oh-my-zsh\033[0m
\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m
\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m
\033[0;34mTime to change your default shell to zsh!\033[0m
sh: line 37: chsh: command not found
gwydion%

## What to do?

Since running the installer on cygwin has some prerequisites, we should probably document them in the installation instructions. And maybe we could change `install.zsh` to handle the cygwin cases by:
* Detecting if the `git` it found is the Windows/MSYS Git and throwing an error message asking for Cygwin git to be installed
* Testing whether `chsh` exists on the system, and giving a "please change your shell manually" message if it doesn't

Also, it seems the installer should exit with a nonzero status when it's aborting due to an error, like git missing or OMZ already being installed.

### PR DESCRIPTION
**Pull request by apjanke**
Adds support to the installer for Cygwin quirks.
* Balk at incompatible Windows/MSYS git
* Test for chsh presence before trying to use it
* Replace non-portable `[[ ... ]]` and `[ x = *pattern* ]` constructs with `[ ... ]` and grep calls

Fixes #3588

I've tested this on Cygwin Win 7, OS X 10.9, Debian 7, and FreeBSD 10.1, and it runs without errors on them, including when the installing user was starting with a different shell. One exception: `uninstall.sh` errors out on FreeBSD because it's using a hardcoded `/bin/bash` path for the shell to revert to. Bash is not installed by default on FreeBSD, and when it is present, it's at `/usr/local/bin/bash`.  Looks like #2836 or #2626 could take care of that.

### CODE REVIEW

**Comment by juanojeda** [comment]
Will this ever get merged and implemented? I'm trying to install oh-my-zsh (due to an awesome writeup from Smashing Magazine) and I've hit a roadblock because the installer keeps creating `c:/home` and putting the initial download, then looking for files in `/home/`.

**Comment by apjanke** [comment]
@juanojeda: Eventually, yes. Though I don't think it would fix your `C:\home` issue. I haven't seen that before, and this PR is not intended to address that. (But it probably should, and if we can figure this out, I'll add something for it.) This PR fixes some minor problems; the installer is already basically working on Cygwin, and lots of people are using Oh My Zsh there (including myself).

Are you running the installer command from the Cygwin prompt (mintty/shell) or the Windows command prompt? Even if you have the Cygwin commands on your Windows path, the OMZ installation won't work from the Windows command prompt. It's a Unix thing, so it needs to be run from the Cygwin prompt.

File: tools/install.sh [hunk 1]
```diff
@@ -32,10 +44,14 @@ sed -i -e "/export PATH=/ c\\
 export PATH=\"$PATH\"
 " ~/.zshrc
 
-TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
-if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
```
mcornella: Why is this removed?

apjanke: Oops – bad merge when rebasing. I'll add it back in.

mcornella:
:+1: 
/cc @robbyrussell

robbyrussell:
Sorry for the latency, folks!

PR #3647, opened 2015-02-28
Installer: More thorough cleanup of files and unneeded `source` calls

apjanke:
This makes the installer and uninstaller a bit better about hygiene, removing some extra data files and reducing the chance of missing `.oh-my-zsh` at a custom location.

* Change `uninstall.sh` to run under `zsh`, so we have access to better file globbing, string substitution, and $ZDOTDIR. (We can assume zsh is available at this point, becuase the `uninstall_oh_my_zsh` function that calls it is a `zsh` function.
* Have `uninstall.zsh` respect `$ZSH` for finding the `.oh-my-zsh` dir
* Remove additional oh-my-zsh data files in `uninstall.zsh`: `~/.zsh-update` and `~/.zcompdump-${SHORT_HOST}-*`
* Remove unneeded `source ~/.zshrc` in `uninstall.zsh`
* remove bogus `. ~/.zshrc` at end of install.sh: it would just read it in the `sh` process running `install.sh` right before exiting, so it would have no effect aside from sometimes kicking out syntax errors.

We could probably go a bit further with the principle of not deleting user-produced data during the uninstallation process, and preserve any files in their `$ZSH_CUSTOM` dir. As is, the default location for the custom dir is under `$ZSH`, so it gets blown away during the uninstall.

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -3,7 +3,7 @@ function zsh_stats() {
 }
 
 function uninstall_oh_my_zsh() {
-  env ZSH=$ZSH /bin/sh $ZSH/tools/uninstall.sh
+  env ZSH=$ZSH SHORT_HOST=$SHORT_HOST zsh $ZSH/tools/uninstall.zsh
 }
 
 function upgrade_oh_my_zsh() {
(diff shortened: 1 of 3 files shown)


apjanke:
Just noticed: this PR also fixes a portability problem with `uninstall.sh`. Some systems have a `sh` flavor that doesn't recognize the `[[ ... ]]` syntax, so the test in the current code for existence of `~/.oh-my-zsh` will error out, and it won't get deleted. This happens to me on FreeBSD and PC-BSD.

apjanke:
Closing this; will re-post a different PR. After working more with OMZ devs, it sounds like OMZ wants to keep the scripts in `tools` running under `sh` and not `zsh`. Will post a revised PR that keeps them running under `sh`.

PR #3648, opened 2015-02-28
Installer: add ability to install from forked repos

apjanke:
Give install.sh ability to install from forked & branched repos, instead of just main robbyrussel repo.

This is a developer tool to support testing of changes to the core installation code: you'll be able to do a roundtrip test of install and uninstall using the working code on your branch. For example, I was working on some changes to `install.sh`, `uninstall_oh_my_zsh()`, and `uninstall.sh` together, and it took a lot of manual fussing with my omz install to get it all lined up for a decent test.

It's controlled by passing `$REPO` and `$BRANCH` environment variables to install.sh. If these aren't given, it has the same behavior as now, cloning master from robbyrussell/oh-my-zsh.

So with this, you could just do this to test out my Cygwin-related changes to the installer from #3646.

```
export REPO=apjanke/oh-my-zsh BRANCH=installer-cygwin-support-3; sh -c "$(curl -fsSL https://raw.githubusercontent.com/$REPO/$BRANCH/tools/install.sh)"
```


mcornella:
I don't think we want to pull this: you can easily have a custom installer that does this without making the main installer more complex.

mcornella:
On second thought this is not that big a change. I'm pulling a version of it in #5169.

mcornella:
#5169 has added these options, as well as the possibility to set the full remote URL with the variable `REMOTE`. This will come in handy when wanting an ssh-based URL or in a fork from Bitbucket or GitLab.

Thanks!

mcornella:
Closed by #5169.

PR #3656, opened 2015-03-02
Remove virtualenv code from autoenv plugin

RobLoach:
It seems like this virtualenv code is not related to autoenv.

cc @serdardalgic @mcornella

### DIFF
--- a/plugins/autoenv/autoenv.plugin.zsh
+++ b/plugins/autoenv/autoenv.plugin.zsh
@@ -7,23 +7,3 @@ if ! source $HOME/.autoenv/activate.sh 2>/dev/null; then
   echo '--------------------------'
   return 1
 fi
-
-# The use_env call below is a reusable command to activate/create a new Python
-# virtualenv, requiring only a single declarative line of code in your .env files.
-# It only performs an action if the requested virtualenv is not the current one.
-
-use_env() {
-    typeset venv
-    venv="$1"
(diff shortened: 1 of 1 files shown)


apjanke:
Autoenv is itself a virtualenv management tool, like virtualenvwrapper. Sounds like it could be related. Especially because the comment says "requiring only a single declarative line of code in your .env files." An `.env` file is an autoenv file. This seems to be a function intended for use in autoenv files.

RobLoach:
It talks about creating a Python virtualenv though.... Should it then be moved to [virtualenv.plugin.zsh](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/virtualenv/virtualenv.plugin.zsh), or [virtualenvwrapper](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh) then?

apjanke:
Probably not. It is a virtualenv, but autoenv is a layer *on top of* virtualenv, for managing virtualenvs and relating them to projects in the file system. and this virtualenv is specifically one that is manage by autoenv.

virtualenvwrapper is an alternative product that sits on top of virtualenv and also manages them. But it does things slightly differently.

There are several tools that go with Python virtualenv functionality, some with overlapping or conflicting functionality. And there are corresponding OMZ plugins for several of them. This function looks like it just happens to be one used by the autoenv folks, so it should probably stay with that plugin.

PR #3657, opened 2015-03-02
Source rvm from the rvm plugin

RobLoach:
Have the plugin source rvm


mcornella:
`rvm` appends this same lines in `.zshrc`, which would load rvm after it's been loaded by the OMZ plugin. Thanks though (sorry for the delay).

PR #3660, opened 2015-03-03
added myip-plugin

Laird-Dave:
Plugin to get current public IPv4 address, probably obsolete as soon as we all use v6 addresses ;-)

This plugin is nothing more than a convenience wrapper for dig +short and opendns.

### DIFF
--- a/plugins/myip/README.md
+++ b/plugins/myip/README.md
@@ -1,0 +1,15 @@
+# Goal
+This adds a simple command to find out your current public IPv4 address
+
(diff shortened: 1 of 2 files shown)


mcornella:
Functionality present in the [systemadmin plugin](https://github.com/robbyrussell/oh-my-zsh/blob/3705d47bb3f3229234cba992320eadc97a221caf/plugins/systemadmin/systemadmin.plugin.zsh#L142).

PR #3664, opened 2015-03-04
installer: Use printf and terminfo for portability

### LINKED ISSUE #3658: Non-portable terminal control sequences in installer
The oh-my-zsh installer (`install.sh` as run through `curl <installer_url> | sh`) uses some terminal control sequences to change the color of the text it outputs. Something about these is not portable to some systems. On some systems, instead of changing the text color, it outputs some non-human readable "garbage" characters. These look like the original escape sequences as printed in the `install.sh` source.

This varies with some combination of the OS, terminal emulator, and what flavor of shell `/bin/sh` is on that system. Here's a table of what I've seen.

* Shell = the shell I invoked `install.sh` under, exactly as typed on the command line
* /bin/sh flavor = the kind of shell that `/bin/sh` is linked to in this system, determined by `sh --version` or reading `man sh`

OS | Terminal | $TERM | Shell | /bin/sh flavor | Result
--- | ------ | --------- | ------ | ------ | ------
Cygwin/Win 7 | `mintty` | `xterm` | `sh` | `bash` 4.x | garbage
Cygwin/Win 7 | Windows console | `cygwin` | `sh` | `bash` 4.x | garbage
Debian 7 | console | `linux` | `sh` | `dash` | OK
Debian 7 | `gnome-terminal` | `xterm` | `sh` | `dash` |  OK
Fedora 21 | console | `linux` | `sh` | `bash` 4.x |  garbage
Fedora 21 | `gnome-terminal` | `xterm-256color` | `sh` | `bash` 4.x |  garbage
Fedora 21 | `rxvt-unicode` | `rxvt-unicode` | `sh` | `bash` 4.x | garbage
PC-BSD 1.01 | console | `xterm` | `sh` | `sh` | garbage
PC-BSD 10.1 | `gnome-terminal` | `xterm` | `sh` | `sh` | garbage
OpenBSD 5.6 | console | `vt220` | `sh` | `pdksh` | OK
OS X 10.9.5 | Terminal.app | `xterm-256color` | `sh` | `bash` 3.x | OK
PC-BSD 10.1 | `gnome-terminal` | `xterm` | `ksh` | `sh` | OK

From looking at this, it seems like the main correlate is the flavor of shell that's being used. `bash` 4.x and some more POSIX-y shells look like they produce garbage, regardless of the terminal they're running under.

I think this is also supported by the fact that the original backslashes appear in the output. Not just for the terminal control escape sequences, but for the `"\n"` LF characters later on. So maybe the quote and backslash string syntax used in the installer is not portable. Or the behavior of the `echo` command is different between the shells.

-------------------------------

Here are screenshots of the runs I did for figuring out that table. If anybody else has test cases under other combinations of system shells and terminals that would be enlightening, feel free to post and I'll add them to the chart. If you can, include the output of `echo $TERM`, `uname -a`, and (if your `sh` supports it) `sh --version`.

(Note the screenshots that say "FreeBSD" are actually PC-BSD, which is built on FreeBSD.)

cygwin win 7 - mintty - sh is bash 4
cygwin win 7 - windows console - sh is bash 4
debian 7 - console - sh is dash
debian 7 - gnome-terminal - sh is dash
fedora 21 - console - sh is bash 4
fedora 21 - gnome-terminal 3 - sh is bash 4
fedora 21 - rxvt-unicode - sh is bash 4
freebsd 10 1 - console - sh is sh
freebsd 10 1 - gnome-terminal - sh is sh
openbsd 5 6 - console - sh is pdksh
os x 10 9 5 - terminal app - sh is bash 3

apjanke:
Updates the installer for portability to different systems and terminals.

Replaces uses of `echo` with `printf` in the installer scripts for portability. `echo`'s handling of escape characters and options is system-dependent. This makes the terminal escape sequences for setting colors portable to Cygwin, FreeBSD, and other platforms whose `sh`'s `echo` doesn't behave like OS X, and causes the escape sequences to be output literally.

Replaces hardcoded `xterm` style terminal control sequences with `terminfo` lookups, for portability to different terminals.

Tests whether it is running in a terminal, and doesn't use terminal visual effect escape sequences if it's not in a terminal.

Fixes #3658 
Closes #4249 because it includes the same changes (manually copied over here due to potential merge conflict)

Has minor conflict with #3646. Merging either will require a rebase and manual adjustment of the other.


mcornella:
I've tested this on cygwin with bash and it works correctly, as Andrew said. Ok to merge :+1: 
/cc @robbyrussell

tools/install.sh
```diff
@@ -1,52 +1,78 @@
 set -e
 
+# Use colors, but only if connected to a terminal, and that terminal
+# supports them.
+ncolors=$(tput colors)
+if [ -t 1 ] && [ -n "$ncolors" ] && [ "$ncolors" -ge 8 ]; then
+  RED=="$(tput setaf 1)"
```
mcornella: Typo: `==` instead of `=`

apjanke: Fixed. Thanks.

apjanke:
I've merged in the changes from #4249 and #4241, so this is ready to go again. Tested on OS X.

tools/upgrade.sh
```diff
@@ -1,16 +1,37 @@
-printf '\033[0;34m%s\033[0m\n' "Upgrading Oh My Zsh"
+
+# Use colors, but only if connected to a terminal, and that terminal
+# supports them.
+ncolors=$(tput colors)
+if [ -t 1 ] && [ -n "$ncolors" ] && [ "$ncolors" -ge 8 ]; then
+  RED=="$(tput setaf 1)"
```
mcornella: This one hit you again :grin:

mcornella:
This is ready to go again after that double equal sign is gone :+1:

tools/upgrade.sh
apjanke: Stupid copy-pasta error. Thanks for the catch.

apjanke:
Double equal sign is fixed now.

tools/install.sh
```diff
@@ -43,19 +62,25 @@ mv -f ~/.zshrc-omztemp ~/.zshrc
...
+echo ''
+echo ''
+echo 'Please look over the ~/.zshrc file to select plugins, themes, and options.'
+echo ''
+echo 'p.s. Follow us at https://twitter.com/ohmyzsh.'
+echo ''
+echo 'p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.'
+echo ''
+printf "${NORMAL}"
 env zsh
```
mcornella: Why `env` and not `exec`? I think this final part has been broken for a long time; now at least it doesn't source `.zshrc` after the zsh session finishes, but maybe we can change this to `exec zsh` as well? I don't know, what are your thoughts?

apjanke: I'm just preserving existing behavior here. (The deletion of the following `. ~/.zshrc` line just reflects the recent change from `wget ... | sh` to `sh -c $(wget ...)`, which changes the installer's `stdin` orientation.)

IMHO, both `env zsh` and `exec zsh` are bad: I would never expect an installer to take over a running shell. I understand this is trying to be user friendly, but there's no way to do it "cleanly", and it's unusual for an installer. `env zsh` is weirder than `exec zsh` bc it just launches a subshell, and when you exit that you're back in the old shell without OMZ loaded, which is going back in time and esp confusing for new users. On the other hand, any installer running `exec` (esp one that tells you to `wget somewhere.com | sh`) smells of malware, or at least presumption. Replacing a user's running shell without telling them in advance you're going to do that is rude at best.

I would rather this just printed a message saying something like "Open a new terminal window to enjoy Oh My Zsh!"

mcornella: Actually the `. ~/.zshrc` was wrong as well with the previous method of installation (`curl | sh`), since it would be read by `sh` and possibly lead to syntax errors.

As is now, here's the behavior we get:

##### `env zsh`:
1. A subshell is created running zsh.
2. User finishes session, runs `exit` or `CTRL`+`D`.
3. Goes back to `sh -c "$(wget)"`.
4. Has no more instructions to run, the `sh` process finishes.
5. Back to initial shell where the user ran `sh -c ...`

##### `exec zsh`:
1. The current `sh -c ...` process is substituted by `zsh`
2. User finishes session, runs `exit` or `CTRL`+`D`.
3. Back to initial shell where the user ran `sh -c ...`

So, in terms of behavior, both are equally flawed, and I think this is as good a behavior as we're gonna get.

Regarding executing directly or printing a message, I think it's ok that we're at least showing what they'll get once they restart the terminal / [logout and log back in](https://github.com/robbyrussell/oh-my-zsh/issues/4298). Perhaps we should make it explicit that they need to restart/logout once they're done? 

I don't know, every time I try to think in terms of what the user might do or think I fail miserably. It's probably good as it is now.

apjanke: The old `. ~/.zshrc` wasn't actually wrong in that manner with the old method, just confusing. In the case of `curl | sh`, `sh` doesn't process the entire input file as a script at once; it reads and evaluates it line by line. When that `| sh` spawns a new `zsh`, that `zsh` session *takes over* stdin (which is attached to `curl |` and not the user's tty/keyboard) and reads the rest of the input. So all the lines after `env zsh` are actually processed by that `zsh` subshell. But then `zsh` runs out of input and exits, returning control to `sh`, so the user never got to actually work in that new `zsh` session. I think people just didn't realize this was happening because the installer was commonly run from a `zsh` session anyway, at least by developers.

This is the same reason that `chsh -s` was failing on systems where it read from `stdin` instead of directly from `/dev/tty`: it would consume the next line of `install.sh` instead of new input from the keyboard, and use that as the password.

Here's a simple script to verify it. Save this as `foo.sh`.
```
echo Hello $$ $0
env zsh
echo Another hello $$ $0
```

I get the following.
```
eilonwy:~ janke$ cat foo.sh | sh
Hello 3903 sh
Another hello 3904 zsh
eilonwy:~ janke$
```

Anyway, moot point, since we're not using `curl | sh` any more.

###   What to do now   ###

Oh yeah. You're right. I didn't see that `exec` would be replacing the `sh -c` subshell that was running `install.sh`, and not the interactive shell that you ran it from. Sorry. Maybe I've been thinking in terms of shell functions too much lately.

I still think it's a bit of an annoyance and would personally prefer just a printed "restart your terminal to get OMZ", because I wouldn't expect an installer to go right ahead and run a program, and you've still got that original non-OMZ shell parent session running, which seems a little messy. But it's not as bad as I said in my previous comment. I'm fine with whichever way you guys want to do it.

mcornella: > Here's a simple script to verify it. Save this as foo.sh.
>``` 
> echo Hello $$ $0
> env zsh
> echo Another hello $$ $0
>```
>
> I get the following.
> 
>```
eilonwy:~ janke$ cat foo.sh | sh
Hello 3903 sh
Another hello 3904 zsh
eilonwy:~ janke$
>```

Huh, you're right. I've been thinking it wasn't like that this whole time... And I confirm that with `sh -c` that's not the behavior we get, as you correctly knew before removing the source line:
```
$ sh -c 'env zsh; echo "$0: this is not actually read by zsh!"'
zsh> exit
sh: this is not actually read by zsh!
```

Frankly, I'm ok with the current system, let's leave this discussion for a subsequent PR.
```

apjanke: Sounds good to me.

apjanke: One other data point: #4261 is a user who was confused about OMZ's normal startup sequence by the "preview" subshell in the installer, and would also like an option to avoid it, to support scripted installs.

Is there an existing issue for discussion of what the installer behavior should be here?

mcornella: That's a good point, let's open an RFC issue to decide what the behavior of the installer should be.

robbyrussell:
@apjanke I've pulled in #3889 ahead of this, would you be able to rebase for this?

apjanke:
Woah, it's a veritable merge party today!

I've merged in `master` here but couldn't get the rebase to work (either due to the manually-resolved conflicts, or my `git` skills aren't strong enough). Would you like me to just open a new branch/PR with the commits manually squashed (as in https://github.com/apjanke/oh-my-zsh/commits/installer-portable-colors-work), or is this good?

mcornella:
A quick diff of both `install.sh` and `upgrade.sh` show that both branches are equal, so it looks good from here :wink: :+1:

tools/install.sh
apjanke: One other thing: if you run the installer with `sh install.sh` instead of `sh -c "$(cat install.sh)"`, then, depending on your system, it looks like it might not slurp and buffer the entire script. So if you modify the `install.sh` file while that interactive `zsh` subshell is running, it may have unpredictable results when it returns from the subshell and tries to run the rest of the script. It might be best to find a way to run this such that the installer script finishes completely upon its invocation.

apjanke: Opened a discussion/RFC issue: #4393 What should installer behavior be, esp wrt launching new shells?

apjanke:
Caused #4404: This change breaks the installer on systems that don't have `tput`, or whose `tput` only supports `termcap` style capabilities, instead of `terminfo` style ones. This includes FreeBSD.

PR #3673, opened 2015-03-06
Use camelCase for npm aliases

### LINKED ISSUE #3640: npmd alias is messing with the fantastic npmd module
Alias at https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/npm/npm.plugin.zsh#L7 should be removed or changed.

This in favor of this cool module https://github.com/dominictarr/npmd ...

jameswomack:
A member of the community contacted me, letting me know they use an npm module that is globally installed as npmd. This module does not have very wide adoption, but it does highlight that I could have done better to prevent conflicts. As these aliases are for devs that use npm, and npm enforces lowercase names, a potential solve is to use camelCased aliases. 

Fixes #3640

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,8 +1,14 @@
 eval "$(npm completion 2>/dev/null)"
 
-# Install and save to dependencies
-alias npms="npm i -S "
 
-# Install and save to dev-dependencies
-# Skipped if npmd module installed
-[[ ! -e $(which npmd) ]] && alias npmd="npm i -D "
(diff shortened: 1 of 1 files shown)


jameswomack:
Additionally, I don't believe these aliases have seen enough adoption yet to prevent this update.

mcornella:
Very well! Please squash your commits into one and add `Fixes #3640` to the PR description to autoclose that other one when this PR is merged. 

/cc @robbyrussell

mcornella:
Also, for future pull requests, the best practice is using a separate branch other than master to submit.

PR #3677, opened 2015-03-08
Added automatic detection of OSX/iTerm2, fixed the wrapper to append -CC...

bwithem:
... (iTerm2) to commands with arguments.


larson-carter:
@bwithem Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

mcornella:
This was left as a configuration option because this has [a few consequences when using tmux](https://iterm2.com/documentation-tmux-integration.html), therefore it would be a breaking change. Thanks for the contribution though

PR #3683, opened 2015-03-11
Use camelCase for npm aliases

### LINKED ISSUE #3640: npmd alias is messing with the fantastic npmd module

jameswomack:
A member of the community contacted me, letting me know they use an npm module that is globally installed as npmd. This module does not have very wide adoption, but it does highlight that I could have done better to prevent conflicts. As these aliases are for devs that use npm, and npm enforces lowercase names, a potential solve is to use camelCased aliases.

Fixes #3640

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,7 +1,14 @@
 eval "$(npm completion 2>/dev/null)"
 
-# Install and save to dependencies
-alias npms="npm i -S "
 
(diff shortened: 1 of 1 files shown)


mcornella:
:+1: /cc @robbyrussell

PR #3685, opened 2015-03-12
Add alias for db:drop db:migrate db:seed command

IgorMarques:
Makes easier to do a full reset on database.

Idea by @lucasbibiano (https://github.com/lucasbibiano)


mcornella:
I believe this is now acomplished with `db:reset` so this is obsolete. Apologies for finding this now, I'm going through `rails` plugin PRs.

PR #3695, opened 2015-03-18
themes/adben: Fix typo to correct check mark color

apjanke:
Corrects the "clean" VCS check mark color in theme `adben`, author @adben.

In `themes/adben.zsh-theme`, there's a typo in the initialization of `$VCS_CLEAN_COLOR` that defines it with lower-case `cs`. Line 45, currently.

```
Vcs_CLEAN_COLOR="${PR_RESET}${PR_GREEN}"
```

Later uses of this variable are in all caps, and since zsh is case sensitive with variable names, the get replaced with the empty string instead of the reset and green color control sequences.

```
ZSH_THEME_SVN_PROMPT_CLEAN="${VCS_CLEAN_COLOR} ✔${VCS_SUFIX_COLOR}"
...
ZSH_THEME_GIT_PROMPT_CLEAN="${VCS_CLEAN_COLOR} ✔${VCS_SUFIX_COLOR}"
```

This is causing the "clean" checkmark in the vcs prompt to show as red instead of green.

screen shot 2015-03-18 at 1 01 08 am

This pull request fixes the typo.


adben:
Thank you @apjanke!!

Issue #2337, opened 2013-12-11
compdef: unknown command or service: rails after changing "rails3" plugin to "rails"


drakmail:
After changing rails3 to rails in my zshrc, I got this error:

```
compdef: unknown command or service: rails
```

PS. I'm using rbenv.

jasonm23:
I'm getting this issue, have done 

```
rm -f ~/.zcompdump
compinit 
```

`compaudit` returns no output.

jasonm23:
The offending lines are in `rails.plugin.zsh`

```
alias rails='_rails_command'
compdef _rails_command=rails
```

I don't know why though... (investigating.)

magicalbanana:
@jasonm23 even figured it out? Trying to find out too but can't seem to make it work. Perhaps we should just remove it. :dancer:

jasonm23:
I ended up removing it from my init.

avstudio:
fix for me (rbenv)
rbenv rehash

lynnegithub:
Thanks! @cmejo

AlexVentura:
I'm working with `Rails 5.2.3` on `OS X Mojave 10.14.4` and this answer from @cmejo fixed the issue!

---

```
Fixed it by doing:

compaudit | xargs chmod g-w
compaudit | xargs chown root
rm -f ~/.zcompdump*; compinit

Any downsides to this method; am I okay permissions wise? Thanks!
``

trejo08:
> On my machine at work `rm -f ~/.zcompdump*; compinit` fixed the issue. On my machine at home and on a colleagues machine that didn't help.

This command works fine for me

PR #3705, opened 2015-03-20
compdef: unknown command or service

magicalbanana:
Commented line should eliminate this error until fix is found or proper investigation is made.

### DIFF
--- a/plugins/rails/rails.plugin.zsh
+++ b/plugins/rails/rails.plugin.zsh
@@ -18,8 +18,8 @@ function _rake_command () {
   fi
 }
 
-alias rails='_rails_command'
-compdef _rails_command=rails
+#alias rails='_rails_command'
+#compdef _rails_command=rails
 
 alias rake='_rake_command'
 compdef _rake_command=rake


Kriechi:
@magicalbanana you mean a proper investigation by searching through existing issues on GitHub?
https://github.com/robbyrussell/oh-my-zsh/issues/2337#issuecomment-38144387

Please close this PR.

apjanke:
Instead of merging this PR, I'd try to diagnose and fix the issue on the machine(s) where you're getting it.

The error `compdef: unknown command or service` occurs if you call `compdef` before initializing the zsh completion system by calling `autoload -U compinit; compinit`. Normally the `oh-my-zsh.sh` file will do this for you as part of oh-my-zsh startup. It's at [line 62](https://github.com/robbyrussell/oh-my-zsh/blob/bf7b9165fd725d2bdf45ca7e639b2c91f5ef65bb/oh-my-zsh.sh#L62-L64).

This could be caused by a messed-up `oh-my-zsh.sh`. Or a bad `.zcompdump` file. Try deleting your zcompdump files with `rm ~/.zcompdump*` and using `git` to check that your oh-my-zsh install is up to date and doesn't have any local changes, and then try again.

If you're still getting the error, you can smooth over it by patching `plugins/rails/rails.plugin.zsh` in your `$ZSH_CUSTOM` directory instead of disabling it for all OMZ users.

apjanke:
Sorry, I misread the error message: it's `command not found` that happens when `compinit` hasn't run; this looks like it's coming from `compdef` itself. I still think it's worth trying to diagnose and fix, or disabling via `$ZSH_CUSTOM` before doing a PR to turn it off globally.

ncanceill:
@magicalbanana this issue probably occurs because the version of Zsh you are running does no support completion for rails. Please upgrade your Zsh distribution.

ncanceill:
... and please keep in mind that arbitrarily commenting out a line printing an error does not make said error go away.

apjanke:
Is this issue still happening for you?

Issue #2869, opened 2014-06-09
Don't display Powerline icons after upgrading to OS X 10.10 Yosemite


takahiro47:
After upgrading to OS X 10.10 Yosemite all of the themes which use "Powerline" don't display powerline icons.
Of cause, I tested to change iTerm2.app's font to 'Ricty Regular for Powerline', 'Inconsolata-dz for Powerline', 'Source Code Pro for Powerline', '' and so on.

※ on tmux, powerline icons has displayed that icons with no problems!

![2014-06-09 17 11 51](https://cloud.githubusercontent.com/assets/1147724/3214388/c9c17630-efad-11e3-9495-339ff996eecb.png)

jonlambert:
Did this ever get resolved? I'm experiencing a similar issue at the moment on Yosemite 10.10.1 in Terminal. 

![w0t6k412lpgpup9zs5wlhm9gncruydobhcbq6ne79cw](https://cloud.githubusercontent.com/assets/856225/5795047/5a70721e-9f7e-11e4-861d-5c68f0490da7.png)

schrepfler:
My zsh theme was actually corrupted, the special characters got encoded in something else (no idea why Yosemite would do that). Putting a new checkout of the theme solved the issue.

jonlambert:
Hmm that's a good idea, but I just checked out the theme (and even re-cloned the .oh-my-zsh directory) to no avail! 
![screenshot 2015-01-19 12 35 07](https://cloud.githubusercontent.com/assets/856225/5800537/a9dc0594-9fd7-11e4-8a96-b8b5fb4b30c6.png)

MartinRogalla:
Encountering the same problems with Iterm2 and default Terminal in Yosemite. Should be a Powerline/OSX specific problem since it also occurs when using vim-airline. Has anyone found a solution for this?

![screen shot 2015-01-23 at 14 59 19](https://cloud.githubusercontent.com/assets/1066157/5875716/73d57bec-a310-11e4-823f-ae2f6241b51a.png)

seawatts:
Also seeing this same error

jkenneydaniel:
Was this resolved? Having same problem - have tried reinstalling powerline fonts a million times... On Yosemite, with iTerm2.

mclearc:
I've had similar problems. The only solution I've found is to use the unicode numbers. 

Symbol  |  powerline | vim-powerline
-----------|----------------|---------------|-
[...]

Issue #1906, opened 2013-06-24
A few invalid symbols in 'agnoster.zsh-theme'


hitigon:
Like line 29. Please fix it. Thank you.

xinput:
hello. 

i got a problem with the agnoster theme.
i want to use agnoster bit it looks like this 
![bildschirmfoto 2015-04-18 um 23 29 34](https://cloud.githubusercontent.com/assets/11652924/7217313/0423c2da-e623-11e4-9584-5e88d287e416.png)

when i tried a git checkout to an older agnoster front i get this error:
![bildschirmfoto 2015-04-18 um 23 29 53](https://cloud.githubusercontent.com/assets/11652924/7217317/21e0ea46-e623-11e4-8618-2fd15d3af1f5.png)


someone knows why i get an "too many arguments" error?

mjbrender:
FYI -- seems like this is back in El Capitan: 

<img width="372" alt="screen shot 2015-10-01 at 8 06 23 pm" src="https://cloud.githubusercontent.com/assets/1744971/10236778/e95b7024-6877-11e5-9599-a857eaa8d2ef.png">

DanielGGordon:
Kind of a side question, but has anyone ever gotten Agnoster/oh-my-zsh to work properly on Cygwin (Windows)? I'm wondering if it is even worth trying to accomplish.

CyberLight:
If you has a problem on **El Capitan 10.11 El Capitan 10.11.1**,  please use fixed version of theme from my repo 
https://github.com/CyberLight/agnoster.zsh-theme

scboffspring:
@DanielGGordon yup It works fine wiht cygwin :)

mcduder:
@scboffspring what did you change to make it work?

mcornella:
Use an updated powerline font. See github.com/powerline/fonts

DanielGGordon:
@scboffspring So I didn't get the font working - but even if I got that working, another problem was it was slooooow. I'm talking about 5-7 seconds just to `cd` into my repo.

mcduder:
@DanielGGordon I got it to work by using a [mod](https://gist.github.com/mcduder/d767808a2aa2942dd524) for agnoster that I tinkered around with a little bit. This should be faster and try to look for fonts in here https://github.com/powerline/fonts. Install the font and then restart cygwin and check in the options.

DanielGGordon:
[...]

PR #3713, opened 2015-03-23
Modify agnoster theme for Mac OSX Yosemite

honghsien5:
Agnoster theme for OSX Yosemite stopped working since the update.
Changed the unicode character for separator as well as adding the ± symbol for git so the theme would display correctly on Yosemite.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -26,7 +26,7 @@
 # A few utility functions to make it easy and re-usable to draw segmented prompts
 
 CURRENT_BG='NONE'
-SEGMENT_SEPARATOR='⮀'
+SEGMENT_SEPARATOR='\ue0b0'
 
 
 # Begin a segment
@@ -76,7 +76,7 @@ prompt_git() {
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git show-ref --head -s --abbrev |head -n1 2> /dev/null)"
     if [[ -n $dirty ]]; then
       prompt_segment red black 
-      st=' ±'
+      st=' \u00b1'
     else
       prompt_segment green black
 
@@ -101,7 +101,7 @@ prompt_git() {
     zstyle ':vcs_info:*' formats ' %u%c'
     zstyle ':vcs_info:*' actionformats ' %u%c'
     vcs_info
-    echo -n "${ref/refs\/heads\//⭠ }${vcs_info_msg_0_%% }${mode}"
+    echo -n "${ref/refs\/heads\//\ue0a0 }${vcs_info_msg_0_%% }${mode}"
   fi
 }
 
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -76,7 +76,6 @@ prompt_git() {
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git show-ref --head -s --abbrev |head -n1 2> /dev/null)"
     if [[ -n $dirty ]]; then
       prompt_segment red black 
-      st=' \u00b1'
     else
       prompt_segment green black
 
@@ -89,7 +88,6 @@ prompt_git() {
     elif [[ -e "${repo_path}/rebase" || -e "${repo_path}/rebase-apply" || -e "${repo_path}/rebase-merge" || -e "${repo_path}/../.dotest" ]]; then
       mode=" >R>"
     fi
-    mode=$st;
     setopt promptsubst
     autoload -Uz vcs_info
 
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -26,7 +26,8 @@
 # A few utility functions to make it easy and re-usable to draw segmented prompts
 
 CURRENT_BG='NONE'
-SEGMENT_SEPARATOR=''
+SEGMENT_SEPARATOR='⮀'
+
 
 # Begin a segment
 # Takes two arguments, background and foreground. Both can be omitted,
@@ -74,9 +75,11 @@ prompt_git() {
     dirty=$(parse_git_dirty)
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git show-ref --head -s --abbrev |head -n1 2> /dev/null)"
     if [[ -n $dirty ]]; then
-      prompt_segment yellow black
+      prompt_segment red black 
+      st=' ±'
     else
       prompt_segment green black
+
     fi
 
     if [[ -e "${repo_path}/BISECT_LOG" ]]; then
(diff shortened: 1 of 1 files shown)


apjanke:
Out of curiosity, what font(s) are you using for your terminal? Neither the old (U+E0B0) or new (U+2B80) character display for me in any of the fonts I have installed on my OS X Mavericks machine.

Do you know what the specific issue was that caused this theme to break for you on Yosemite? Agnoster is designed to use some non-standard characters that require ["Powerline-patched" fonts](https://github.com/Lokaltog/powerline-fonts) that include those non-standard glyphs to work properly. This PR looks like it's switching away from one of those special Powerline characters. Perhaps your Yosemite machine needs a patched font installed, or its terminal configured to use one.

honghsien5:
I am currently using the powerline patched Menlo 14pt on Iterm 2 which is the font that the guide suggested. Before I make the modification on the theme file, the symbols including seperator and branch were not able to displayed incorrectly and showed up as the empty square symbol. There are other users who use Yosemite that apply this theme have the same issues. I essentially followed one of the solution to make the changes in this PR. Refer to this link for the issue: https://github.com/robbyrussell/oh-my-zsh/issues/2869

The compatibility was actually part of my concern with this pull-request since I wasn't sure if the new edit would break the theme for other OSes.

apjanke:
Okay. I'll grab one of the powerline patched fonts and see if I can reproduce.

apjanke:
I've installed the current version of the powerline-patched fonts from the [powerline repo](https://github.com/powerline/fonts) and the characters output okay for me on both 10.9 and 10.10, with U+E0B0 (the one currently used by the agnoster theme) resolving to that pair of triangles that look like a right-pointing triangle.

Menlo 12 (regular font) on 10.9:

menlo 12 - 10 9 - terminal

Meslo LG M for Powerline on 10.9:

meslo lg m for powerline - 10 9 - terminal

Meslo LG M for Powerline on 10.10:

meslo lg m for powerline - 10 10 - terminal

I think what's going on is that the Powerline project has [changed which of the private-use code points they use over time](https://github.com/powerline/powerline/commit/53fbfe15fead8cc7598bcb4ee9714a221ab7e446). This means that the powerline-patched fonts have had to change, and *actually change their code point to glyph mappings*.

This is confusing to work with because the source code for powerline and the agnoster themes use the non-standard powerline Unicode characters directly in their source code, and their appearance *in the source code* will depend on which version of the patched powerline fonts you have installed. So you can't tell what the source code is unless you know which version of the font you have installed.

From viewing the underlying code points used in the source code, it looks like the current agnoster theme definition is using the current powerline code points (U+E0B0 for the arrow), and your patch would change it to use the old code points. This suggests that you may have older versions of the powerline patched fonts installed on your system.

Try re-installing your powerline fonts using the current versions from the powerline repo, or a current version of their patcher.

IMHO, it would make things a lot easier if the source code was changed to use numeric Unicode code point escape sequences instead of the special characters. That would be correctly and unambiguosly readable regardless of what font the developer was using, and would allow users without powerline-patched fonts to still read the source. It would also allow diffs to be correctly readable when switching the characters around; if I understand this correctly, there's no font that would make both the old and new glyphs legible in a diff output.

apjanke:
And here's what agnoster itself looks like on my OS X 10.10 Yosemite machine using the newly downloaded patched powerline fonts. This is in iTerm2, using the Solarized Dark color scheme.

Current agnoster definition:

agnoter - meslo powerline - upstream

With your patch:

agnoster - meslo powerline - honghsien5 patch

I think this is consistent with you having an outdated version of the patched powerline fonts that is using the old glyphs. (Unless the [powerline/fonts](https://github.com/powerline/fonts) version itself never got updated to use the new glyphs and is out of date. It's not clear from their commit history if and when that happened, and powerline/fonts#44 never got closed. But it is an official powerline repo, so I'd tend to defer to it.)

apjanke:
Looks like this is a dupe of #1906, which says that outdated patched powerline fonts were the issue.

honghsien5:
I see. Thank you for putting the work in analyzing the issue!

I have revert the unicode value on my commit. I also change some of the special character into the numeric unicode format for easier reading in the future. 
Since I followed the guide on the theme page for my installation, I think it would be a good idea to change the powerline font download link to the right one.

Should I create another PR for this change?

honghsien5:
Forgot to clarify which download link that I was referring to.
Under the theme page https://github.com/robbyrussell/oh-my-zsh/wiki/themes, the download link for patched Menlo 14pt within the parentheses is the link that I used to download the font. It should be modified or deleted to prevent future confusion.

apjanke:
Oh, yep. That explains it. The "patched Menlo 14pt" link from the agnoster Themes section is to a font file posted in January 2012, and it would have the old code point mappings.

I've updated the wiki page to remove the obsolete link and update the name of the current powerline font repo. Thanks.

apjanke:
Yeah, I think it would be appropriate to just close this one and make a new Pull Request, since the remaining change doesn't relate to this PR's original title or most of the discussion. And probably ping Agnoster's author or maintainer, since this significantly changes the appearance of the theme. Red is often used to indicate an error condition, instead of an informative notice, so it bears discussing with the theme's existing users and developers.

honghsien5:
Good idea. I changed it to Red because the original Green and Yellow are indistinguishable due to my colorblindness. I have sent an email to the theme's author regarding the color. On the other hand, what is usually a good way to discuss with the theme's existing users and developer?

apjanke:
A good way is to post a new Issue here on oh-my-zsh, and reference the theme's author or maintainer(s) with an @-mention in the issue text. That way they'll get a notification that their attention is needed. I don't know a good way of getting users' attention, aside from including the theme name in the issue title.

For you and other users specifically, you may also be able to change your terminal profile's color configuration to make the yellow color map to a different color that's distinguishable from green, and maybe from red as well. That would fix the problem for any theme or terminal app without upstream code changes. There's even a discussion a discussion about a colorblind-friendly variation of the Solarized color scheme over at altercation/solarized#91. That might be ideal, because then you can fix it at the client color configuration level, and not have to worry about support from individual themes.

honghsien5:
Sounds good. Changing the color on my client side seems to be the best solution. I'll go ahead and create another pull request for the numeric unicode modification.

apjanke:
Cool. And on second thought, the unicode escape change might not be worthwhile: judging from the other issues we looked at, it seems like the powerline code points have been stable for a while and you can expect all users to be on the current fonts. Maybe it would be better to leave them as characters, and figure out how to use tools to dump the underlying code point values if it becomes an issue again.

honghsien5:
Yeah. Since the outdated link has been removed, I also thought the chance of people having the same issue would be pretty low.

i2:
Make sure you change the font in your Terminal app. Also you don't need to install all the Powerline fonts. Installing just this one does the job too: https://github.com/Lokaltog/powerline/raw/develop/font/PowerlineSymbols.otf

PR #3727, opened 2015-03-27
Add support for choosing the method of seting terminal titles.

gotyaoi:
This patch preserves the current behavior, but also allows a user to set
the method they want used. This is useful in certain tmux setups that
use both screen and xterm methods, at the very least.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -13,13 +13,37 @@ function title {
   # if it is set and empty, leave it as is
   : ${2=$1}
 
-  if [[ "$TERM" == screen* ]]; then
-    print -Pn "\ek$1:q\e\\" #set screen hardstatus, usually truncated at 20 chars
-  elif [[ "$TERM" == xterm* ]] || [[ "$TERM" == rxvt* ]] || [[ "$TERM" == ansi ]] || [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
-    print -Pn "\e]2;$2:q\a" #set window name
-    print -Pn "\e]1;$1:q\a" #set icon (=tab) name
-  fi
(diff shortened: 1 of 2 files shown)


gotyaoi:
see https://github.com/robbyrussell/oh-my-zsh/issues/3579

winny-:
:+1:

lib/termsupport.zsh
```diff
@@ -13,13 +13,37 @@ function title {
...
+      print -Pn "\ek$1:q\e\\"
+      print -Pn "\e]2;$2:q\a"
+      print -Pn "\e]1;$1:q\a"
+      ;;
+    bothlong)
+      print -Pn "\ek$2:q\e\\"
+      print -Pn "\e]2;$2:q\a"
+      print -Pn "\e]1;$1:q\a"
+      ;;
+  esac
```
ncanceill: Are you sure we do not need a default case here?

gotyaoi:
Well, in the case that the variable is unset, or is set but does not match one of the cases, it simply executes none of the branches and no title setting occurs. Pretty much the same as an empty default case, I think. Did you have some specific default behavior in mind?

gotyaoi:
In addition, in the case where they didn't enter anything, the variable should be set automatically, per the previous behavior.

ncanceill:
Not necessarily, I was just checking.

Looks OK for me, but you will need a few more testers for this to get merged.

larson-carter:
@gotyaoi Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

gotyaoi:
Unfortunately I haven't thought about this in at least 3 years and I don't actually use oh-my-zsh anymore, so I cannot.

Issue #3379, opened 2014-12-05
Feature request: split plugins into submodules


adamwong246:
I really like oh-my-zsh but I feel that it's not right to have to download all the plugins, when I only use a few. 

I think the maintainer of each plugin should use the git subtree split technique to split plugins into separate repos and then add references to those repos as submodules. Oh-My-Zsh can continue to operate as usual while allowing those outside the oh-my-zsh ecosystem to use (and contribute to!) the project. 

I'm still working out the kinks but here's an example.
https://github.com/adamwong246/heck-yeah-zsh-colored-man - the colored-man plugin, split into a seperate repo. Ideally this would be maintained by the originator, in this case @moinakg. 

https://github.com/adamwong246/dotfiles - my dotfiles, which references the split repo as a submodule. 

Of course, this may introduce some extra cognitive overhead when merging and such but this approach allows for a more "federated" system of plugins.

dritter:
:+1: The same applies to templates. I think it would unburden the OMZ-team from a lot of tasks and the vast amount of pull requests will decrease.. As a bonus: pull requests will be merged faster..

ncanceill:
Hi, and thanks for contributing.

There is an ongoing discussion on project management and maintenance, but for now it has been temporarily chosen to keep everything in a single repository.

Besides, we do not have a dedicated maintainer for each plugin.

dritter:
@ncanceill Can you say where the discussion is happening? Can't find anything in the wiki nor in the issues..

ncanceill:
Some of it is in #2771.

ncanceill:
Also some of it is in #1035.

PR #3734, opened 2015-03-30
Added theme "powerlevel9k"

dritter:
Here is a new theme "powerlevel9k" which may be a substitute for the "agnoster" theme. It is more flexible and has some additional features..

Screenshot is in the original repository: https://github.com/bhilburn/powerlevel9k

It may also showcase the use of git submodules as requested in #3379


dritter:
@bhilburn FYI

ncanceill:
[Please do NOT send us themes](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes)

Issue #377, opened 2011-05-30
Proposal to Simplify OH-MY-ZSH


sorin-ionescu:
Organising code inside of OH-MY-ZSH is becoming unnecessarily complex. For a while, there has been the _/custom_ directory to dump files that override or append to the functionality of _/lib_, _/plugins_, and _/themes_.

Then the ability to override _/plugins_ via _/custom/plugins_ was added (#89).  Now, a new proposal has been submitted to allow the overriding of _/themes_ (#376) and _/lib_ (in the comments to #376) via _/custom/themes_ and _/custom/lib_ respectively.

Additionally, there is the _/functions_ directory (b865f2a) for dumping miscellaneous functions and the _/completions_ directory (#351) for dumping miscellaneous completions.

These directories are unnecessary for we already have a software configuration management (SCM) tool called Git. We should use Git exclusively.

I propose that we get rid of _/custom_, and edit _/lib_, _/plugins_, and _/themes_ directly. We also get rid of _/functions_, and _/completions_ to encourage the creation of plugins instead of dumping random code in those directories.

Pulling updates while retaining customisations is easy by recursively merging conflicting hunks by favouring **our** version. 

```
git pull --no-commit --strategy=recursive -X ours robbyrussell master
```

Updates should always be checked for dangerous code (`rm -rf /`). Hence we use `pull --no-commit` to stop Git from auto-committing to let us check the files even when there is a clean merge.

```
git status
*** Check the files listed. ***
git commit
```

Lastly, if we use Git exclusively, the _/tools_ directory becomes unnecessary and should be removed.

Keep It Simple!

PR #3736, opened 2015-03-30
add help to install zsh - fix #3715

### LINKED ISSUE #3715: add help to install zsh
I want to install ohmyzsh but I need to install zsh first. A brief paragraph to install zsh would be welcome. I'll be happy to add it but any help would be appreciated. Shall I do it in the README or the wiki?

MartinDelille:


MartinDelille:
this would fix #3715

ncanceill:
As discussed over there, can you make it a simple link to `http://zsh.sourceforge.net/FAQ/zshfaq01.html` instead of the list of steps?

ncanceill:
Also, can you literally include "fix #3715" in the description of this PR, in order to auto-close #3715 when this gets merged?

MartinDelille:
I check your link and it is very verbose compared to my 4 lines instruction.

ncanceill:
Let us look at this objectively. I originally thought everybody installing OMZ knew about Zsh, but apparently I was wrong, since there are two kind of OMZ newcomers in the world:
* those who did know of Zsh;
* and those who did not.

Now,
* if you want OMZ and you know about Zsh, you probably already have installed, even if it is not your default shell, by this point you should know how to change shells anyway;
* if you want OMZ but you do not know about Zsh, it is probably a good time to start reading some verbose information about it, before even installing.

We are proud of OMZ being (quite) easy to set up, but this should not include Zsh. Adding a few aliases and functions is way less change than changing your shell. So it does not surprise me that there is a bunch of stuff you should read before doing the Bash->Zsh switch.

Moreover, the goal of OMZ is not to _make installing Zsh easier_.

In the end, I would agree if your simple 4-steps instructions were able to cover all use-cases, but of course it does not. You would have to take into consideration distros missing Zsh packages or having out of date versions, or even without any package manager at all and then you need to install from source; on some lightweight Linux distros, `chsh` itself can be missing and shell switching is done by hand in `/etc/passwd`.

Granted, the [link](http://zsh.sourceforge.net/FAQ/zshfaq01.html) I was suggesting would be better if it could directly point to question **1.6: Where do I get it?** on the page, but I cannot find a proper `href` to point to.

***

Another thing, in case you still disagree: your four steps are inconsistent.

Steps 3 and 4 do not take care of installing Zsh, they just make it your default shell. Suppose Zsh is already installed — maybe even some other user has it as his default shell, but you do not. Then `zsh --version` will give the proper output, and steps 1 and 2 become superfluous, but steps 3 and 4 remain necessary.

MartinDelille:
Ok no problem I will close this bug and this PR. I totally understand you point of view: I use OMZ since a friend advice it to me but in the end I don't really know the difference between bash, zsh and OMZ. My proposal was a little bit naive, sorry for the disturbance ;-)

ncanceill:
Hey, easy there tiger! My criticism was not meaning that you should close this.

As you pointed out in #3715, it can happen that people try to install OMZ without knowing about Zsh, and we need to address that. Also, the dichotomy between installing and setting Zsh is already missing from the current Readme, so there is stuff to do here.

Please consider re-opening and pursuing the discussion, so we can all agree on a renewed Prerequisites section of the Readme.

mcornella:
> Granted, the link I was suggesting would be better if it could directly point to question 1.6: Where do I get it? on the page, but I cannot find a proper href to point to.

http://zsh.sourceforge.net/FAQ/zshfaq01.html#l7

See [named anchors](http://www.w3schools.com/tags/att_a_name.asp) image

----

That said, I agree with @MartinDelille in that it's too verbose for first-time users of zsh, specially because we should assume they want it working in *as little time as possible*. But I also agree that the steps need to be more generic. The `chsh` part is not needed though, if we're fixing it in the `install.sh` script. 

What about a simple:

> Zsh should be installed (a version greater than *4.3.9*). Check that you have it by running the command `zsh --version 2>/dev/null || echo 'zsh not installed'`.
> If you don't have zsh installed yet, go to [link on the wiki](#) to get installation instructions for your system.

We could also integrate this check in the installation script, as well as point people to the appropriate wiki page to know what to do next.

Let me know what you think.

ncanceill:
Yes, a wiki page is a good option. This will offer the opportunity to give a simple 3-steps process and then go into the details and give the link to the Zsh FAQ.

jgrar:
I disagree that further help is needed in the README. If this is a problem for first time users, as it seems to be, then it should be cleared up within the project description itself. This is an issue with the project language and how it is presented to first time users. Stress language such as '*for the ZSH Shell*'. Make a clearer distinction between OMZ, ZSH and other Shells.

ncanceill:
OK, can you write down a couple sentences so we see what you have in mind?

ncanceill:
I just stumbled on [a 3-year-old post](https://github.com/robbyrussell/oh-my-zsh/issues/377#issuecomment-4204013) by @robbyrussell and he kind of contradicts my point:
> oh-my-zsh needs to be focused on helping people adopt zsh as easily and painless as possible. We should assume that an overwhelming majority of people who use OMZ will be coming from bash/sh and this will be the first introduction to zsh. We shouldn't assume that they:
> 
> a. know git
> b. know too much bash/zsh/sh (they've probably barely scratched the surface here)
> 
> Knowing these are nice-to-have, not pre-requisites. I want to take care of these people, first and foremost. OMZ needs to hold their hand as long as we can. When they feel comfortable to run further ahead (advanced shit), we can try our best to accommodate.

So scratch what I said above about OMZ not needing to make installing Zsh easy.

I still believe that the README should simply state that you need to install Zsh and set it as your default shell, with a link to the wiki page explaining how to do so in details, with different sections for different OSes or setups.

MartinDelille:
What do you think of this wiki page? https://github.com/robbyrussell/oh-my-zsh/wiki/Install-ZSH

ncanceill:
Nice, I updated it with some details, can you change the PR into a link to that page?

MartinDelille:
Done

ncanceill:
Excellent.

Can you please keep it where the original instructions were? (_ie_ above `curl` and `wget`, not below)

Also can you please squash your commits?

MartinDelille:
I'm not a squash expert and if I was able to merge my commits, 3 new appear during the operation...

jgrar:
@MartinDelille re squashing commits, does this help? https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

MartinDelille:
Done: I made an error during the first interactive rebase and removed the additional commits! I learned squashing my commits in the end ;-)

PR #3739, opened 2015-03-30
Fix in theme plugin

horosgrisa:


apjanke:
This is redundant with #3587, which removes the pwd side-effect.

Using `cd` and `$OLDPWD` still has a side effect: When something runs `lstheme()`, it changes their `$OLDPWD` to `$ZSH/themes`. So if the caller was planning on using `$OLDPWD` for something (e.g. they had done a `cd` themselves and were planning on undoing using `$OLDPWD`) it'll no longer work.

PR #3741, opened 2015-03-31
rr theme: keep $ret_status really local with anonymous function

apjanke:
Changes the `robbyrussell` theme for variable hygiene. This wraps the use of "local" variable `$ret_status` in an anonymous function so it's local to just the theme setup, and doesn't leak `$ret_status` in to the global namespace.

Theme definitions are `source`d, and not autoloaded and called as functions, so they do not have their own scope for variable definitions aside from the caller's. Without the anonymous function, the `local ret_status` happens at the top level of the `zsh` session, and so is in the global namespace and still visible after the theme is loaded.

```
➜  ~  zsh -l
➜  ~  echo $ret_status
%(?:%{%}➜ :%{%}➜ %s)
➜  ~
```

Caused by #3742, which has a longer explanation.


apjanke:
Closing in favor of #3743, which has a wider-reaching solution.

PR #3703, opened 2015-03-19
Allow user to blacklist themes from random selector.

FranGM:
If user adds themes they don't want to see to ZSH_BLACKLISTED_THEMES they won't be picked by the random theme selector.

Fixes #3704

PR #3743, opened 2015-04-01
Unify theme support code

### LINKED ISSUE #3585: theme plugin side effects (cwd & variables) and inconsistency

### LINKED ISSUE #3586: lib/theme-and-appearance.zsh sets non-appearance options
The `lib/theme-and-appearance.zsh` script sets some shell options that don't look related to cosmetic theme and appearance stuff. At line 22:

```
#setopt no_beep
setopt auto_cd
setopt multios
setopt cdablevars
```

Shouldn't these be set somewhere else? They have significant semantic effects, especially `multios`.

### LINKED ISSUE #3704: Feature request: Allow user to blacklist themes from being picked by the random selector
I had a similar issue as https://github.com/robbyrussell/oh-my-zsh/issues/3036, so I wanted to blacklist some themes from being chosen by the picker, maybe because they have dependencies I don't have or because they're not pretty enough ;)

I created a small PR that accomplishes this: https://github.com/robbyrussell/oh-my-zsh/pull/3703

### LINKED ISSUE #3742: "local" doesn't work in theme definitions
A lot of the theme definition files (`themes/*.zsh-theme`) are using the `local` declaration to make their variables local to the theme definition. This doesn't actually work, because `local` variables in `zsh` are dynamically scoped, not lexically scoped, and the theme definition files are run using `source` so they're evaluated in the caller's function stack frame. (A sourced file is not a scope, unlike autoloaded functions.) In the current code's design, that's the global namespace. So those `local` declarations are having no effect.

This could be fixed by doing the `source` for theme loading inside a function call to provide a level of indirection for the dynamic scoping. Could be done with an anonymous function or a specially-provided function (which could be more readable for debugging).

Same thing probably applies to plugins and `lib/*.zsh` files if they're using `local` at the top level. You could wrap most of `oh-my-zsh.sh` in a function call to provide a scope frame if you wanted to take care of it all in one fell swoop.

### Affected themes

Here's a quick list of themes that look like they're using `local`. It's not fully correct, because they could be using it inside functions. But I spot-checked several of them and they just had bare `local` declarations at the top level.

```
$ grep -rl local themes  | wc -l
      51
$ grep -rl local themes
themes/af-magic.zsh-theme
themes/afowler.zsh-theme
themes/agnoster.zsh-theme
themes/alanpeabody.zsh-theme
themes/avit.zsh-theme
themes/bira.zsh-theme
themes/bureau.zsh-theme
themes/candy-kingdom.zsh-theme
themes/dieter.zsh-theme
themes/dogenpunk.zsh-theme
themes/dpoggi.zsh-theme
themes/eastwood.zsh-theme
themes/essembeh.zsh-theme
themes/fino-time.zsh-theme
themes/fino.zsh-theme
themes/fishy.zsh-theme
themes/flazz.zsh-theme
themes/frontcube.zsh-theme
themes/funky.zsh-theme
themes/gallifrey.zsh-theme
themes/gallois.zsh-theme
themes/gnzh.zsh-theme
themes/intheloop.zsh-theme
themes/itchy.zsh-theme
themes/jnrowe.zsh-theme
themes/jonathan.zsh-theme
themes/jreese.zsh-theme
themes/kennethreitz.zsh-theme
themes/linuxonly.zsh-theme
themes/lukerandall.zsh-theme
themes/macovsky-ruby.zsh-theme
themes/macovsky.zsh-theme
themes/mira.zsh-theme
themes/mortalscumbag.zsh-theme
themes/murilasso.zsh-theme
themes/muse.zsh-theme
themes/obraun.zsh-theme
themes/peepcode.zsh-theme
themes/pure.zsh-theme
themes/pygmalion.zsh-theme
themes/re5et.zsh-theme
themes/robbyrussell.zsh-theme
themes/simonoff.zsh-theme
themes/smt.zsh-theme
themes/sonicradish.zsh-theme
themes/sorin.zsh-theme
themes/sunaku.zsh-theme
themes/sunrise.zsh-theme
themes/trapd00r.zsh-theme
themes/wedisagree.zsh-theme
themes/ys.zsh-theme
```

### LINKED ISSUE #8889: Perform environment reset between themes in theme-chooser
## Problem

When running the `/tools/theme-chooser.sh` script to evaluate and look at different themes, there are many occasions when some configuration elements from one theme will "bleed" over to the next one or more until a theme overrides said configuration element. For example, the `clean-theme.zsh` configures an `$RPROMPT` value, however, the next theme to be displayed, `cloud-theme.zsh`, does not set a value for `$RPROMPT`, so previous value of `$RPROMPT` remains. The theme after `could-theme.zsh` is `crcandy-theme.zsh`, which also does not set a value for `$RPROMPT`, so the previous value from `clean-theme.zsh` still carries over.

This can (and does) provide a very skewed and potentially incorrect sampling of the theme from the quick preview we get through `theme-chooser.sh`.

## Desired Outcome

Optimally, the proverbial slate would be wiped clean of all configuration elements between the sampling of themes. A significant level of difficulty comes in when plugins are involved which add their own theming variable(s). 

## Possible steps to remediation

1. Reset a small and pre-defined subset of variables most common among themes. Those would include:
* `$PROMPT`
* `$RPROMPT`
* `$ZSH_THEME_GIT_PROMPT_(PREFIX|SUFFIX|CLEAN|DIRTY)`
* Possibly other `$ZSH_THEME_GIT_*` values, from the core *git* plugin

2. After loading and displaying a theme, extract any environment variables that might have been set, such as a grep for `ZSH_THEME_[^=]`, extract the variable name, and unset it.

3. Get the currently defined environment variables that begin with `ZSH_` (or `ZSH_THEME`), and unset any which are not part of the core ZSH/OMZ functionality, such as `ZSH_NAME`, `ZSH_VERSION`, etc.

4. Have all themes included in the base deployment follow the `prompt_cleanup` pattern from the ZSH Docs' ["Writing Themes"](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Writing-Themes) section

5. Invoking a new instance of the shell, using only the specified theme file. This would be the "most thorough" solution, but also the most time and resource-intensive due to the need of starting and exiting a new shell instance for every single theme.

6. Do nothing, leave it as is and close this issue as "Won't fix". Not ideal (from my position and opinion), but still an option none-the-less.

## What I've tried

I've already gone down the path of option 1 above, however, as I was looking for the more common "core' variables within the themes, I saw a few RVM and VENV among others. My list quickly grew large, somewhere around 20, which is why I mention having only a few in that option.

## Conclusion

Given that depending on the path representing the most favourable option, I would be willing to take on this issue, or perhaps put together a POC of the top 2-3 options which would require a code change.

apjanke:
### Overview

This PR unifies the theme support code between the `themes` plugin and core `oh-my-zsh` by folding the `themes` plugin in to `lib/themes-and-appearance.zsh` and having `oh-my-zsh.sh` use its functions for the initial theme loading. It adds code for robustness, including the support for unloading themes, debugging the theme loading process, and fixing `local` declarations inside themes.

Fixes #3742.
Closes #3703.
Fixes #3704.
Fixes #3586.
Fixes #3585.
Fixes #8889.
Closes #3587.
Closes #3337.
Closes #3739.
Closes #3741.
Closes #3703.
Closes #2962. ("random" functionality provided via a keyword arg instead of defined theme.)
Closes #2113. (`$ZSH_THEME=""` will disable theming.)
Closes #4471 (which changes search logic in `theme` plugin; this PR effectively includes that change.)
Closes #4605 (this also removes the `lstheme` `cd` and adds looking in `$ZSH_CUSTOM/themes`

### What this accomplishes

Behavior of themes loaded by `oh-my-zsh.sh` during the initialization process and the `theme()` function is now consistent.

Themes can be cleanly switched during a zsh session. Previously, there was no way to cleanly unload a theme, so loading a new theme using the `themes` plugin could end up with a mix of settings from the new theme and previously loaded themes. This adds a mechanism to "unload" a theme by resetting theme-related variables to defaults and removing theme-installed `chpwd`/`preexec`/`precmd` hooks.

Themes defined in `$ZSH_CUSTOM` are supported everywhere now, so they can be used in initial startup or `theme()` calls, and are included in random theme selection.

Adds new `theme next` and `theme <n>` calling forms to let you step through the list of all defined themes. Adds new `theme off` command for unloading the theme and reverting to ZSH default values.

Fixes leakage of `local` variables defined inside themes. Previously, they were sourced in the top-level shell namespace, so `local` declarations had no effect.

Adds "blacklist" support to exclude certain themes from being picked by `random` or `next`, as requested in #3704. (Also important for working with the theme debugging code.)

Adds optional debugging output to the theme loading process so you can evaluate themes for correctness and side effects.
### Implementation

Fixes the `local` definitions in themes by sourcing the theme definition from inside a function call, to provide a layer of indirection. The additional function call stack frame becomes the context for `local` declarations.

Fixes themes that were relying on not-really-local behavior for parameters declared as `local` inside the theme definitions. Mostly these were of the pattern `local ret_status="..."; SOME_PROMPT='${ret_status}'` and were fixed by moving `${ret_status}` outside the single-quotes or replacing them with double-quotes, so the parameter substitution happens at theme loading time instead of prompt evaluation time.

Fixes the `lstheme()` function to remove the `pwd` side effects, and has all code that wants to get a list of themes call `lstheme` instead of looking through the theme directories themselves. Now everything has a consistent view of what themes are defined. The total ordering of themes is well-defined and stable, so it can be used as a reference for loading themes by index.

The `theme next` command uses that `lstheme` ordering to allow you to iterate through the whole list of themes. This is mostly useful for debugging and development work.

The non-existent `$ZSH/completions` directory added to `$fpath` by `oh-my-zsh.sh` is created, as a place for the `_theme` completion definition to go. This avoids adding `$ZSH/lib` to `$fpath`.

The parts of the code that track hook functions and debugging state do set difference operations on arrays. To be compatible with ZSH 4.x, this is done with a custom `_omz_array_setdiff` function instead of using the `${foo:|bar}` substitution, which is new in ZSH 5.x.
### Possible conflicts

The `theme` and `lstheme` functions are now always defined when oh-my-zsh is loaded, instead of just when the `themes` plugin is loaded.

If any themes declare parameters as `local` but are actually relying on them behaving as globals, this could break those themes, because `local` parameters really will be local now. (That is a bug in those theme definitions, since they're misusing `local`.) This PR includes fixes to all the themes in the repo that were doing so.

Conflicts with #2401 since the code diverges significantly.


lib/theme-and-appearance.zsh
```diff
@@ -1,10 +1,429 @@
...
+# Return value:
+#   0 on success
+#   Nonzero on failure, such as requested theme not being found or theme index
+#      out of range.
+#
+# The "random" and "next" forms will skip over themes in the ZSH_BLACKLISTED_THEMES 
+# variable.
+#
+# As a consequence of this argument syntax, themes may not be named "random", "next", 
+# or an integer number.
```
ncanceill: We want to remember to document that somewhere in the wiki.

ncanceill:
I am up for testing this, can you find a few more?

apjanke:
@mcornella, maybe you'd be up for reviewing and testing this too? I don't really know any of the other OMZ maintainers since I'm new here.

apjanke:
@avit, @agnoster, any chance you'd be interested in testing this, since you've worked with some of the theme functionality?

agnoster:
I'm afraid my knowledge of zsh is limited to what I needed to kludge together my theme. Sounds cool though!

mcornella:
Sorry @apjanke, adding this to my list of pending PRs to test (which is getting a little bigger now :wink:).

apjanke:
Note to follow up on: some users are defining themes inside subdirectories under `custom/themes` so Git submodules can be used. [PowerLevel9K](https://github.com/bhilburn/powerlevel9k) does this. This has the effect of creating themes with `/` in the name. Need to decide whether this is actually supported.

Both the new theme code and the old stuff will not include these "nested" themes in the theme name listing. But they will be loaded if requested specifically. Need to decide if this is actually supported, and if it is, add recursive search to the theme enumeration code.

apjanke:
Rebased on master to get a clean merge.

robbyrussell:
@apjanke Hey there -- can you rebase one more time? I feel like the changes in the `lib/theme-and-appearance.zsh` require some careful attention

apjanke:
Sure. Going to be a bit – I'm in a heavy work and travel period right now. Feel free to ping me in a couple weeks if I forget.

apjanke:
Rebased. I looked closely at `lib/theme-and-appearance.zsh`, and I think the rebase merge is good. It was kind of a pain, though. Maybe we could merge this soon to avoid further rebases?

PR #3745, opened 2015-04-04
Fixing a problem of displaying special characters

lebensterben:
It's required to install a patched font to correctly display some special characters in this theme. (https://github.com/robbyrussell/oh-my-zsh/wiki/themes)
But I found it doesn't work until I change the theme file to the original one, which is in its author's gist. (https://gist.github.com/agnoster/3712874)

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -6,7 +6,7 @@
 # # README
 #
 # In order for this theme to render correctly, you will need a
-# [Powerline-patched font](https://github.com/Lokaltog/powerline-fonts).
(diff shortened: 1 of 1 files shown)


ncanceill:

Most problems with the Agnoster symbols are due either to your Powerline fonts not being up to date, or to your terminal emulator being misconfigured. It is very not likely to be an OMZ issue.

Your PR itself is a mess because you did not start it from a clean codebase.

Please close this issue, and for the special characters look for solutions in #1906 or the other issues mentioned over there.

ncanceill:
Thank you, this helps us maintain issues to a manageable level.

Feel free to re-open with more details if you do not find a solution to your problem.

PR #3748, opened 2015-04-05
Configuration of pre-exec terminal title

hugoh:
Added ZSH_THEME_TERM_TAB_TITLE_EXEC and ZSH_THEME_TERM_TITLE_EXEC configuration variables to allow pre-exec title configuration

Right now, only idle titles are customizable. Defaulted to their current values.


mcornella:
That is too complicated to make it into the project. A better option is to redefine the whole function, or make a separate preexec function and add it to `precmd_functions`. 

Thank you for your contribution though

PR #3749, opened 2015-04-05
Add git-fetch aliases to git plugin

pcurry:
Adds two aliases for git fetch activities to keep updated with remote repositories.


ncanceill:
This is now fixed in #2790.

Please do not merge this PR, since it has become polluted by empty merge commits.

pcurry:
Verified @ncanceill 's branch had the aliases I use.  This PR doesn't need to go forward.

Issue #3751, opened 2015-04-06
add `git bisect` aliases


dgirgenti:
I use these aliases in my environment:

<code>gbs</code> = <code>git bisect</code>
<code>gbss</code> = <code>git bisect start</code>
<code>gbsg</code> = <code>git bisect good</code>
<code>gbsb</code> = <code>git bisect bad</code>
<code>gbsr</code> = <code>git bisect reset</code>

PR #3752, opened 2015-04-06
Add git bisect aliases to git plugin

dgirgenti:
solution for #3751


dgirgenti:
I screwed this one up

PR #3754, opened 2015-04-07
Added a function to the core GIT library to allow use for Cygwin users

mikesena:
(Designed to be used by custom themes)

The prompt info for GIT takes an extensive time to return in Cygwin, due to the way Cygwin has to handle c++ forks.

This small addition means that the basic status is able to be displayed, but not taking an extensive amount of time to do so.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -7,6 +7,13 @@ function git_prompt_info() {
   fi
 }
 
+function git_cygwin_prompt_info_quick() {
+  ref=$(command git symbolic-ref -q --short HEAD 2> /dev/null)
+  if [[ "${ref}" != "" ]]; then
+    echo "git:${ref}"
+  fi
+}
+
 
 # Checks if working tree is dirty
 parse_git_dirty() {


lib/git.zsh
```diff
@@ -7,6 +7,14 @@ function git_prompt_info() {
   fi
 }
 
+function git_cygwin_prompt_info_quick() {
+  ref=$(git symbolic-ref HEAD) 2> /dev/null
```
ncanceill: You probably mean:
```zsh
ref=$(git symbolic-ref HEAD 2> /dev/null)
```

mikesena: Correct; I'll amend the pull request tomorrow.




Michael

mikesena:
Pushed an update.

ncanceill:
OK, now please squash your commits together in order to keep the history clean and atomic.

Then find some testers and ask for merge.

apjanke:
FYI, if you're interested in a broader fix for this, there are a couple other work-in-progress Pull Requests (#3725 #3220 #2386) that speed up `git_prompt_status` on Cygwin by replacing the `grep` shellouts with builtin `zsh` parsing support, or short-circuiting unnecessary `git` calls. Those will apply to any theme that uses the normal `git_prompt_status`, not just custom themes.

But it'll probably be a while before they get merged in, so you should go ahead with this PR.

mikesena:
I have squashed those two commits together now.  @apjanke, would you mind testing this, or be able to test it?

mikesena:
@robbyrussell; are you fine with this merge?

mcornella:
This should use the `$ZSH_THEME_GIT_PROMPT_PREFIX/SUFFIX` variables instead of hardcoding `git:`. 

Also, use `command git` instead of just `git`, and you can simplify further by using `git symbolic-ref -q --short HEAD`, which returns the short version of the current branch (without `refs/heads/`) and won't show error messages as well.

mikesena:
@mcornella - I've modified it in the 3 ways you suggested above.

 - Used the shorter form of the git command suggested
 - Appended 'command ' to git. (why do we do that anyway?)
 - Used the `$ZSH_THEME_GIT_PROMPT_PREFIX/SUFFIX` variables

mcornella:
> Used the shorter form of the git command suggested

:+1: 

> Appended 'command' to git (why do we do that anyway?)

That's to prevent aliases or functions being used: "we want the pure git command"..

> Used the $ZSH_THEME_GIT_PROMPT_PREFIX/SUFFIX variables.

Can't see it, did you commit it?

PR #3761, opened 2015-04-09
Better vi-mode

lingnand:


apjanke:
This looks like a useful change.

I think you could modify the code a bit to be portable to more terminal types.

```
if [ "$TERM" = "xterm-256color" ] || [ "$TERM" = "screen-256color" ]; then
```

These cursor control sequences don't require 256 colors, so they would work with lower-color or b&w `xterm` and `screen` variants. And modern `urxvt` supports them, too. You might want to test for the patterns `xterm*`, `screen*`, and `rxvt-unicode*` instead of just `xterm-256color` and `screen-256color`. Especially since several terminal emulators seem to default to declaring `$TERM` as plain `xterm` even though they support 256 colors.

```
bindkey '^[[Z' reverse-menu-complete
```

The terminfo capability `kcbt` stands for "back-tab", and shift-tab gets bound to it. So you could do `bindkey "$terminfo[kcbt]" reverse-menu-complete` instead, for portability (and readability).

lynnard:
I've modified the `if` statement to match the terminals as you suggested. Unfortunately your other suggestion doesn't seem to work - `kcbt` is not found in my `$terminfo` variable...

apjanke:
Darn; I thought `kcbt` was pretty widely supported. What `$TERM` value are you using and what OS are you on? (And what distribution/version if Linux?) I found it in most of the entries (and all the ones for common emulators) I checked in OS X 10.9 and Debian 7. (Though they all use the same escape sequence, so there's probably no actual portability impact to not using terminfo.)

I guess you could still use something like `bindkey ${terminfo[kcbt]:-^\[\[Z} reverse-menu-complete`, which would give you portability to terminals with different sequences, and fall back to the hardcoded `xterm` style sequence if the `terminfo` entry is missing `kcbt`.

mcornella:
This needs to be rebased after the recent `vi-mode` key-binding changes.

dragonxlwang:
this PR fix issues with ``shift-tab`` bug with vim-mode. Anytime to get merged soon?

mcornella:
This PR should be rebased and contextualized to the new PR #5113 by @apjanke.
@dragonxlwang in the meantime, you can add the bindkey command to your zshrc file to get the proper `shift-tab` behavior:
```
bindkey '^[[Z' reverse-menu-complete
```

CallumHoward:
Consider adding `TERM=tmux` and `TERM=tmux256color` terminal types:
https://wiki.archlinux.org/index.php/tmux#256_colors

PR #3763, opened 2015-04-10
git.plugin: Fix typos in comments

mc10:


ncanceill:
Thank you, this will be fixed in #2790.

PR #3764, opened 2015-04-10
Fix `rbenv_prompt_info` prefix/suffix

chieping:
Result of rbenv_prompt_info should be the same as rvm_prompt_info.

before:

after:

### DIFF
--- a/plugins/rbenv/rbenv.plugin.zsh
+++ b/plugins/rbenv/rbenv.plugin.zsh
@@ -44,9 +44,9 @@ for rbenvdir in "${rbenvdirs[@]}" ; do
 
     function rbenv_prompt_info() {
       if [[ -n $(current_gemset) ]] ; then
-        echo "$(current_ruby)@$(current_gemset)"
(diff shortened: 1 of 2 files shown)


ncanceill:
cc @peterhoeg

chieping:
Sorry, I rebased for in case rbenv is not installed.

apjanke:
Instead of doing `${ZSH_THEME_RVM_PROMPT_PREFIX:=(}`, it might be better to just initialize them in `lib/theme-and-appearance.zsh`

```
ZSH_THEME_RVM_PROMPT_PREFIX="("
ZSH_THEME_RVM_PROMPT_SUFFIX=")"
```

and just use their current values in the prompt_info function itself. That would let users configure it to have blank prefixes and suffixes if they really wanted. Same with the original `rvm_prompt_info` function.

apjanke:
I went and made a PR #4588 which does that, and changes all the themes using one-off `rvm-prompt` calls to use `ruby_prompt_info()` instead. Looks like a lot of them were written before `ruby_prompt_info()` was available. Also changes the names to `$ZSH_THEME_RUBY_PROMPT_PREFIX/SUFFIX`, since they apply to Ruby environment managers besides RVM.

If both these PRs get merged, we'll need to reconcile the names between them.

PR #3767, opened 2015-04-10
meteor plugin updated

udnaan:
meteor commands updated to reflect changes in meteor cli 1+. 
"meteor remove" grabs the package list from local .meteor/packages in the project rather than doing a "meteor list" which doesn't work most of the time and is way slower.


ncanceill:

Also when your PR concerns a plugin please copy to the maintainer like so: @jorge-d

And please do not send PRs from your master branch — I can elaborate on why this is bad practice, shoot me an email.

francocatena:
+1

jorge-d:
+1

PR #3768, opened 2015-04-11
Plugins: gitfast: show only local branches when autocomplete for "git checkout"

eugeny-dementev:
In my work i'm very often use git checkout for switching between branches, and for this help gitfast plugin. But, when i type git checkout and press tab, he show me all branches, local, remote, all. I don't need all, i need only my local (for my work) branches.
Here i make some little changes for this and then think it can be good not only for me.

### DIFF
--- a/plugins/gitfast/git-completion.bash
+++ b/plugins/gitfast/git-completion.bash
@@ -1033,7 +1033,11 @@ _git_checkout ()
 		if [ -n "$(__git_find_on_cmdline "$flags")" ]; then
 			track=''
 		fi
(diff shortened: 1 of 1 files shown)


ncanceill:

I understand you want to complete only local branches, but the thing is `git checkout` is not only dedicated to local branches. And since there is a possibility to checkout remote branches as well, they definitely should be offered for completion.

The only way I see this could be acceptable is with a setting, for instance `$GITFAST_CHECKOUT_LOCAL_ONLY`, that you can set somewhere in your `~/.zshrc`.

PR #3772, opened 2015-04-14
hg branch command is sometimes slow, run it just once

vegai:


vegai:
Ehh, nope :D What was I thinking, the second command is hg status. Gotta think a bit more, going to a mercurial dir over sshfs gives me a 1-2 second delay before every prompt. Could things be cached a bit, perhaps

PR #3773, opened 2015-04-14
Add scu-* aliases for systemctl --user commands.

razor-x:
Plugin: systemd.

Lets you do `scu-status`, `scu-start awesome`, for `systemctl --user status`, `systemctl --user start awesome`, etc.


DoctorJellyface:
What about e.g. `sc-u-status` for better tab completion?

razor-x:
@DoctorJellyface Sure, I'm just unaware in what way that specifically helps tab completion :smile:

DoctorJellyface:
You don't get scu stuff when you want just sc.

PR #3774, opened 2015-04-15
quick and dirty fix for deprecated grep_options env var

kallerotmg:
might want to fix it properly, alas, i do not have the time.

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -26,3 +26,6 @@ alias grep="grep $GREP_OPTIONS"
 unset GREP_OPTIONS
 unset VCS_FOLDERS
 unfunction grep-flag-available
+
+alias grep="/usr/bin/grep $GREP_OPTIONS"
+unset GREP_OPTIONS


jgrar:
Forgive me if i'm wrong, but isn't GREP_OPTIONS unset on line 26? Your patch uses it in the alias definition on line 30 and then unsets it again on line 31

ncanceill:
Hi @kallerotmg, and thanks for contributing.

We know of the deprecation, and we are trying to reach consensus on a solution. In the meantime, as you wrote it this is not a proper fix, and will not be merged.

Please provide a proper fix or close this, and try to follow the discussion in #3341 and #3347 as indicated by @jgrar.

mcornella:
This won't work because `$GREP_OPTIONS` is already unsetted when you do the extra alias (see [line 26](https://github.com/robbyrussell/oh-my-zsh/pull/3774/files#diff-bfc27536871d24498f9e19c8e54b0628R26)). Also, a hardcoded route is not the way to go: instead you want to use the `command` builtin like so `alias grep="command grep $GREP_OPTIONS"` (in [line 23](https://github.com/robbyrussell/oh-my-zsh/pull/3774/files#diff-bfc27536871d24498f9e19c8e54b0628R23)).

That said, **the current implementation should already work**. Please explain first what issue you have with it so we can properly address your problem.

kallerotmg:
as previously stated, it is a duplicate of #3341, closing.

mcornella:
:+1:

PR #3780, opened 2015-04-15
update chsh command call

### LINKED ISSUE #3779: Error in chsh in Centos 7
/etc/shells may not point to /usr/bin/zsh (which is the result when using 'which zsh') 
using chsh -l and choosing zsh from there will guarantee chsh to work without failure.

To replicate:
Boot Centos 7 mimimal,
yum install git, zsh
git clone this repo
run install script

Error msg:
\033[0;34mLooking for an existing zsh config...\033[0m
\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m
\033[0;34mCopying your current PATH and adding it to the end of ~/.zshrc for you.\033[0m
\033[0;34mTime to change your default shell to zsh!\033[0m
Changing shell for vagrant.
chsh: "/usr/bin/zsh" is not listed in /etc/shells.
Use chsh -l to see list.

jellyjellyrobot:
/etc/shells may not point to /usr/bin/zsh (which is the result when using 'which zsh') 
using chsh -l and choosing zsh from there will guarantee chsh to work without failure.

Fixes #3779

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -34,7 +34,7 @@ export PATH=\"$PATH\"
 
 if [ "$SHELL" != "$(which zsh)" ]; then
     echo "\033[0;34mTime to change your default shell to zsh!\033[0m"
-    chsh -s `which zsh`
+    chsh -s  $(chsh -l | grep "zsh" -m 1)
 fi
 
 echo "\033[0;32m"'         __                                     __   '"\033[0m"


ncanceill:

This should be fixed by #2836 which uses:
```zsh
```
Please confirm that this version works for you too.

***

Off-topic, instead of _"Ref #3779"_ you should have written _"Fixes #3779"_, in order to auto-close that issue at merge time.

ncanceill:
Thanks.

Note that you actually need to edit the PR's body since it does not work in comments.

PR #3781, opened 2015-04-15
Add bbda for quick branch deletion

sberrevoets:
Add the `gbda` ("git branch delete all") alias for quickly deleting all branches that have been merged into the current branch.

As far as I could tell, this is the only change that was necessary.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -55,6 +55,7 @@ compdef _git gb=git-branch
 alias gba='git branch -a'
 compdef _git gba=git-branch
 alias gbr='git branch --remote'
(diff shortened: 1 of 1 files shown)


ncanceill:

First of all, I am working on a big update to the git plugin: #2790. Could you please close this PR and send it to [my `plugin-git` branch](https://github.com/ncanceill/oh-my-zsh/tree/plugin-git) instead? You will be kind enough to notice that the aliases are sorted alphabetically, so please insert yours where appropriate.

Then, I have two technical comments:
- the backslash before the asterisk is not needed since it is already quoted and there is no regex flag
- you should use `command grep` and `command xargs` in order to bypass any alias that could possibly mess up things

sberrevoets:
Thanks for the feedback. I'll make the changes and open a new PR against the branch you mentioned.

PR #3783, opened 2015-04-17
Porting _adb from zsh-users/zsh-completions

zh0122:
Porting _adb from zsh-users/zsh-completions

### DIFF
--- a/plugins/adb/_adb
+++ b/plugins/adb/_adb
@@ -200,6 +200,7 @@ _adb_device_cmds() {
     'push:copy file/dir to device'
     'pull:copy file/dir from device'
     'sync:copy host->device only if changed'
+    'sideload:flash a package with sideload'
     'shell:run remote shell interactively or command'
     'emu:run emulator console command'
     'logcat:view device log'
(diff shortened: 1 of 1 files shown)


luca020400:
Great :+1:

jgrar:
I don't know about this. Doing it this way won't track changes to zsh-users/zsh-completions, you've really just copied it out of that repository and into this one. It shouldn't be OMZ's responsibility to maintain someone else's work.

It would be a lot better for the user if they managed a repo that was seperate from OMZ *by themself* under custom/. Something like:

```
git clone https://github.com/zsh-users/zsh-completions.git
ln -s ./zsh-completions/src/_adb ~/.oh-my-zsh/custom/adb.zsh
```

jgrar:
@zh0122 why did this get closed?

zh0122:
Hi @jgrar ,

Because you said as below... 
> It shouldn't be OMZ's responsibility to maintain someone else's work.
It would be a lot better for the user if they managed a repo that was seperate from OMZ by themself under custom/, something like:

If you will merge it, I will reopen it

Thanks

PR #3793, opened 2015-04-20
Grep bug in Agnoster

mrkmndz:
In Agnoster theme, in order to detect a modified file, hg status is grepped
Currently it uses regex "^(M|A)" to look for a M or A after a line break, when instead it should be "^[MA]"


ncanceill:
Thanks, that is a quick and easy fix.

This is ready to merge.

PR #3794, opened 2015-04-20
No input for Update Check defaults to Yes

### LINKED ISSUE #3735: Add [Y/n] question in update dialog to make Yes the default
At the moment the update dialogue looks like that:

```
[Oh My Zsh] Would you like to check for updates?
Type Y to update oh-my-zsh: 
```

When I hit enter, it closes.

Wouldn't it be cool, to prefer `Yes` over `No` as the default action?

There are a lot of console applications asking "Type [Y/n] to ..." where the capital letter is the default that is performed when the user hits `Return`.

HeroCC:
This fixes issue #3735 by adding a [Y/n] to the update request, and defaults to continue if there is no input.

Fixes #3735

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -41,10 +41,9 @@ then
     then
       _upgrade_zsh
     else
-      echo "[Oh My Zsh] Would you like to check for updates?"
-      echo "Type Y to update oh-my-zsh: \c"
+      echo "[Oh My Zsh] Would you like to check for updates? [Y/n]: \c"
       read line
-      if [ "$line" = Y ] || [ "$line" = y ]; then
+      if [ "$line" = Y ] || [ "$line" = y ] || [ -z "$line" ]; then
         _upgrade_zsh
       else
         _update_zsh_update


ncanceill:

I fruitlessly looked for a builtin to do that, but your fix will do.

However, please literally add _"Fixes #3735"_ to your above comment so that Issue auto-closes at merge time.

HeroCC:
I edited the comment to include the Fixes part, thanks for the feedback!

ncanceill:
Thanks!

@webmaster128 would you be so kind and test this please?

webmaster128:
@ncanceill Sure. How do I manually trigger the upgrade check? Can I somehow reset the time when the last check was done?

ncanceill:
You can run:
```zsh
zsh $ZSH/tools/check_for_upgrade.sh
```

webmaster128:
This does not help for testing since it finds out that no check is necessary.

This one does the job: `echo "LAST_EPOCH=0" > ~/.zsh-update`

webmaster128:
Patch works as expected.

ncanceill:
Oh yes sorry, anyway thanks for reporting.

This is an easy merge, I will create a new super-PR to bring them all together.

Pending...

webmaster128:
What is expected behaviour when user presses `Ctrl+C` in the 'check for update' question? I end up with a broken shell. But I guess this has never been different.

ncanceill:
Yes, we should set a trap for interrupt, but this is another story.

webmaster128:
Great work! Thanks @HeroCC @ncanceill

PR #3795, opened 2015-04-20
Reduce prompt delay in large git repos

corhere:
The avit theme's _git_time_since_commit function was running git up to
three times, one of which piped the entire git log for the current
branch to /dev/null. Replace it with a single call to `git log`,
checking the exit code for success.

### DIFF
--- a/themes/avit.zsh-theme
+++ b/themes/avit.zsh-theme
@@ -39,9 +39,7 @@ function _ruby_version() {
 # use a neutral color, otherwise colors will vary according to time.
 function _git_time_since_commit() {
 # Only proceed if there is actually a commit.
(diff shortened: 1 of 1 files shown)


ncanceill:
Hi, and thanks for contributing!

apjanke:
Looks good. I tested on OS X 10.9.5. I can't confirm the speedup (it's pretty fast on my box either way), but the new code works: the RPROMPT still has time since last commit, and values match the old implementation.

ncanceill:
Not thoroughly tested yet, but looks good to me.

robbyrussell:
@corhere would it be possible to rebase this?

corhere:
@robbyrussell rebased.

PR #3797, opened 2015-04-21
Fix for gpg-agent on OSX

tacgnol:
If the native osx ssh-agent isn't running (e.g. if ssh hasn't yet been run on the mac since boot), then gpg-agent will start with ssh support when a new shell window is opened, precluding the use of native ssh-agent support (inc. keychain access).


mcornella:
Obsolete.

Issue #3360, opened 2014-12-01
First-time users run into `chsh: option requires an argument -- 's'`


dandv:
I've just tried to install oh-my-zsh for the first time on a new machine:

```sh
$ curl -L http://install.ohmyz.sh | sh
[...]
Cloning Oh My Zsh...
Cloning into '/home/dan/.oh-my-zsh'...
[...]
Looking for an existing zsh config...
Using the Oh My Zsh template file and adding it to ~/.zshrc
Copying your current PATH and adding it to the end of ~/.zshrc for you.
Time to change your default shell to zsh!
chsh: option requires an argument -- 's'
Usage: chsh [options] [LOGIN]

Options:
  -h, --help                    display this help message and exit
  -s, --shell SHELL             new login shell for the user account

$ chsh -s zsh
Password:
chsh: zsh is an invalid shell
$ chsh -s /bin/zsh
Password:
chsh: /bin/zsh is an invalid shell
````

Perhaps the automated installer should complain that `zsh` wasn't install, and offer to install it?

**Comment by brunobatista** [comment]
+1

**Comment by tobiasmuecksch** [comment]
+1

**Comment by mharmuth** [comment]
+1

**Comment by drawcard** [comment]
+1

**Comment by linux-modder** [comment]
+1 to non issue  zsh is an expected depsolv of oh-my-zsh

**Comment by chelogui** [comment]
+1

**Comment by wjwwood** [comment]
It should at least print out the command it meant to run so that users can recover easily from it.

It could be as easy as:

```sh
zsh --version > /dev/null 2>&1
if [ $? -eq 0 ]; then
  echo "running: chsh -s zsh"
  chsh -s zsh
else
  echo "zsh does not appear to be installed."
  echo "please install it and then run: chsh -s zsh"
fi
```

Would a pull request like this be considered (obviously after being cleaned up)?

drawcard:
@wjwwood Great idea :+1:

wjwwood:
> @wjwwood Great idea :+1:

https://github.com/robbyrussell/oh-my-zsh/pull/3799

mcornella:
Fixed in #3809.

PR #3799, opened 2015-04-22
install.sh: fail gracefully when zsh is not installed

wjwwood:
This is in response to #3360.

Basically this changes two things:

- It prints out `Running: chsh -s /bin/zsh` right before running `chsh` so if that fails the user can fix the problem and copy-paste, try again.
- If zsh is not installed it tells the user this and then instructs the user to install zsh then run `` chsh -s `which zsh` `` afterwards.

In the latter case it exits afterwards to prevent the next step from failing, i.e. `env zsh`.

This could be improved on by restructuring the logic further and bailing out earlier when `zsh` is not installed, similar to how it handles `git` not being available. However, this is still better than the current behavior of `chsh` failing and the install stopping in the middle:

```
$ curl -L http://install.ohmyz.sh | sh
[...]
Cloning Oh My Zsh...
Cloning into '/home/user/.oh-my-zsh'...
[...]
Copying your current PATH and adding it to the end of ~/.zshrc for you.
Time to change your default shell to zsh!
chsh: option requires an argument -- 's'
Usage: chsh [options] [LOGIN]

Options:
  -h, --help                    display this help message and exit
  -s, --shell SHELL             new login shell for the user account

```

Here's a screen show of the new lines:

screen shot 2015-04-21 at 8 03 40 pm

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -33,8 +33,16 @@ export PATH=\"$PATH\"
 " ~/.zshrc
 
 if [ "$SHELL" != "$(which zsh)" ]; then
+  zsh --version > /dev/null 2>&1
+  if [ $? -eq 0 ]; then
     echo "\033[0;34mTime to change your default shell to zsh!\033[0m"
+    echo "\033[0;34mRunning:\033[0;1m chsh -s `which zsh`\033[0m"
     chsh -s `which zsh`
+  else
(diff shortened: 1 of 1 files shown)


apjanke:
Overlaps with #2836 and #3780, which also revise the `chsh` shell selection logic. Maybe these can be unified?

mcornella:
Fixed in #3809.

PR #3806, opened 2015-04-24
Fix currentWindowId

duerrp:
On my system xmonad on Ubuntu 14.04, the xprop -root call has an earlier ocurence of NET_ACTIVE_WINDOW that makes this function fail. I guess someone who knows more about xprop should rewrite this function in a less brittle way, but for now my fix should do.


ncanceill:

Can we get a few testers to be sure it does not break anyone's config?

robbyrussell:
We'll close this given how long it's been pending testers for quite some time.

PR #3808, opened 2015-04-24
Rbenv directory search order with brew

mmorga:
The bug is that as-written, rbenv (if installed by brew) always uses the brew prefix directory and the RBENV_ROOT. But the default desired behavior of rbenv is to use the $HOME/.rbenv to store installed rubies. As a result, you get the system ruby only.

My change builds upon the #3275 pull request from @mbologna and reverses the order of the search directories for rbenv to put the brew prefix last.

Since this bug only occurs for brew users, non-OSX users will not be affected.

### DIFF
--- a/plugins/rbenv/rbenv.plugin.zsh
+++ b/plugins/rbenv/rbenv.plugin.zsh
@@ -9,7 +9,7 @@ _rbenv-from-homebrew-installed() {
 FOUND_RBENV=0
 rbenvdirs=("$HOME/.rbenv" "/usr/local/rbenv" "/opt/rbenv" "/usr/local/opt/rbenv")
 if _homebrew-installed && _rbenv-from-homebrew-installed ; then
-    rbenvdirs=($(brew --prefix rbenv) "${rbenvdirs[@]}")
+    rbenvdirs=("${rbenvdirs[@]}" $(brew --prefix rbenv))
 fi
 
 for rbenvdir in "${rbenvdirs[@]}" ; do


ncanceill:

You will have to `git rebase -i @~2` in order to get rid of the empty merge commit.

Then please find a few people to test and +1 this.

gchan:
+1

I haven't tested this but I have encountered the same problem on OSX and brew installed rbenv. Traced the issue with the brew directory rather than $HOME directories for rubies.

mbologna:
Is the bug still present? @mmorga @mcornella

mcornella:
I have no idea... @gchan ?

gchan:
Yes. I just enabled the rbenv plugin and this still happens.

```
$ zsh --version
zsh 5.0.5 (x86_64-apple-darwin14.0)
```

mbologna:
Probably this is why I opened #3275

gchan:
Yes @mbologna, the code changes in #3275 are very similar to this PR. 

Unsure about the reason for line 12 (https://github.com/robbyrussell/oh-my-zsh/pull/3808/files#diff-22fdd1fcfb5b98380f344442393e1f6fL12) however.

mbologna:
@gchan @mcornella @mmorga please check with #4384 if it solves your problem

caesarsol:
@mbologna I don't think that would fix for me, since I have no `bin` directory in `~/.rbenv/`.

I'm getting a bit lost between these PRs... 
my edit would just be like the one in your https://github.com/robbyrussell/oh-my-zsh/pull/3275, specifically [this line](https://github.com/robbyrussell/oh-my-zsh/pull/3275/files#diff-22fdd1fcfb5b98380f344442393e1f6fR16).

mbologna:
@caesarsol #4384 *does not* infer anything from `~/.rbenv/bin`. Please check the diff.

caesarsol:
I'm sorry, isn't the diff only deleting lines 7~10?
I still see `$rbenvdir/bin` in the file...

mbologna:
Nope. See https://github.com/mbologna/oh-my-zsh/commit/f3e357d71253a8255f1a1ccece9426f37da47d00#diff-22fdd1fcfb5b98380f344442393e1f6fL7-10

caesarsol:
I'd reopen https://github.com/robbyrussell/oh-my-zsh/pull/3275/files, that's the issue I'm having.
This PR doesn't solve the problem.

caesarsol:
@mbologna sorry for the misunderstanding, what I'm saying is: 
this PR doesn't solve the issue for me.
I didn't realize you were splitting two different issues.

mbologna:
@caesarsol this PR is closed, see above.

eugenesvk:
Thank for the fix!
Although in my case I had to make __both__ changes (one by @caesarsol and the other one by @mbologna) to make it work. Whenever I revert to implementing only one of the changes, it all breaks down again.

PR #3813, opened 2015-04-25
Fixes gradle autocomplete so it includes subproject tasks

ziegs:
This fixes the gradle plugin. The regular expression used to pull task names didn't include a colon. Subprojects in gradle are of the form foo:fooTask, so they weren't being included. This fixes the bug.


ncanceill:
Can we get a few +1s here to confirm this works?

PR #3815, opened 2015-04-26
Add new plugin for virtual host management

willmendesneto:
Add new plugin `vhost` for virtual host management

### DIFF
--- a/plugins/vhost/README.md
+++ b/plugins/vhost/README.md
@@ -1,0 +1,54 @@
+# VHOST
+
+> Making virtual host maniputation tasks more easier
+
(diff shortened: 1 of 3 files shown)


gbrennon:
plugin tested and working!

willmendesneto:
@ncanceill, @robbyrussell

robbyrussell:
This is an interesting one. Most of our plugins are trying to make autocomplete features on existing tools rather than a collection of unix shell scripts. 

I'd prefer that this sort of tool be it's own external project and we could have a plugin that references it.

PR #3816, opened 2015-04-26
Update Homebrew plugin as `brew upgrade` soon will require `--all` parameter

izzyleung:
In the latest version of Homebrew, when you try to run `brew upgrade`, it will prompt you

```
Warning: brew upgrade with no arguments will change behaviour soon!
It currently upgrades all formula but this will soon change to require '--all'.
Please update any workflows, documentation and scripts!
```

So, add this parameter to avoid errors when upgrading Homebrew in the future (possibly).

PR #3818, opened 2015-04-27
brew: brew upgrade now requires --all

xu-cheng:


ncanceill:

This is fixed in #3816.

PR #3814, opened 2015-04-26
Add autocomplete for frontend-search

willmendesneto:

PR #3819, opened 2015-04-27
remove old installation instruction

alfredbez:


ncanceill:
Thanks, but this is already taken care of in #3814.

PR #3809, opened 2015-04-25
Easy-to-Merge

ncanceill:
## Easy to Merge

This branch collects PRs that are easy to merge. If you want to add a commit, please [submit a PR](https://github.com/ncanceill/oh-my-zsh/pull/new/easymerge) to [my branch](https://github.com/ncanceill/oh-my-zsh/tree/easymerge).

**Rule:** either the PR is an extremely simple fix, or it has been discussed–tested–and–signed-off.

### Status

**READY for MERGE**

This is the third Easy-to-Merge PR: see #2827 #2922

### Changelog

* Fixes #1631
* Fixes #2351
* Fixes #2723
* Fixes #2777
* Fixes #2821
* Fixes #2829
* Fixes #2856
* Fixes #2898
* Fixes #2926
* Fixes #3038, fixes #3195, fixes #3421, fixes #3445
* Fixes #3125
* Fixes #3133
* Fixes #3157
* Fixes #3158
* Fixes #3194
* Fixes #3214
* Fixes #3245
* Fixes #3360, fixes #3715, fixes #3736, fixes #3799
* Fixes #3364
* Fixes #3451
* Fixes #3516, fixes #3872
* Fixes #3542
* Fixes #3617
* Fixes #3638
* Fixes #3644
* Fixes #3663
* Fixes #3693, fixes #3923
* Fixes #3695
* Fixes #3732
* Fixes #3758
* Fixes #3787
* Fixes #3791
* Fixes #3793
* Fixes #3794
* Fixes #3816, fixes #3818, fixes #3829, fixes #3834, fixes #3840, fixes #3853, fixes #3855, fixes #3856, fixes #3883, fixes #3905
* Fixes #3837
* Fixes #3842, fixes #3898
* Fixes #3851
* Fixes #3864
* Fixes #3870
* Fixes #3871
* Fixes #3875
* Fixes #3878, fixes #3961, fixes #3984
* Fixes #3893, fixes #3981
* Fixes #3901
* Fixes #3908
* Fixes #3925
* Fixes #3955
* Fixes #3970

PR #3820, opened 2015-04-28
add argument to brew upgrade, reasons down below:

timfeirg:
from brew:

```
Warning: brew upgrade with no arguments will change behaviour soon!
It currently upgrades all formula but this will soon change to require '--all'.
Please update any workflows, documentation and scripts!
```


ncanceill:

Your PR duplicates #3816 and #3818, and will be fixed in #3809.

Please close.

PR #3821, opened 2015-04-28
Fix minimal.zsh-theme's check for in_svn and add support for mercurial

tresni:
6ce08acb27 broke minimal.zsh-theme's check for SVN repos.  I've updated minimal.zsh-theme to use the return code instead of output value.  Additionally took the time to add support for mercurial as well.

### DIFF
--- a/themes/minimal.zsh-theme
+++ b/themes/minimal.zsh-theme
@@ -6,10 +6,16 @@ ZSH_THEME_SVN_PROMPT_PREFIX=$ZSH_THEME_GIT_PROMPT_PREFIX
 ZSH_THEME_SVN_PROMPT_SUFFIX=$ZSH_THEME_GIT_PROMPT_SUFFIX
 ZSH_THEME_SVN_PROMPT_DIRTY=$ZSH_THEME_GIT_PROMPT_DIRTY
 ZSH_THEME_SVN_PROMPT_CLEAN=$ZSH_THEME_GIT_PROMPT_CLEAN
+ZSH_THEME_HG_PROMPT_PREFIX=$ZSH_THEME_GIT_PROMPT_PREFIX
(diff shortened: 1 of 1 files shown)


ncanceill:
Thanks for contributing. This looks nice.

tresni:
I can confirm that it is working on 5 different linux boxes and my OSX primary hosts.  The change isn't that complicated and is fixing something that is broken (wouldn't display anything in svn repos due to the change noted above, now it does again.)  Given I see no open bug reports on this I'm not sure how many people noticed/cared.  But figured I'd fix it since I use the theme (or more correctly did.)

If you want me to separate out the bug fix from the improvement (ie adding hg_prompt_info) I'm happy to do so.

Telling:
This update broke the git promt on my machine. I'm on OS X 10.10.3, with iTerm2 2.9.20150628-nightly and zsh 5.0.5. I have reverted back to https://github.com/robbyrussell/oh-my-zsh/commit/663d43241cd890c2fa78040a612c6d424674c5f9 and it's working again. Let me know if I can help with debugging.

tresni:
Was there a specific error you were getting?

Telling:
Sorry I should've been more clear. It's broke in the sense that the git prompt is simply not there when it should be. I have not seen any error messages.

tresni:
Interesting.  did this happen on all git repositories?  Do you have SVN or Mercurial installed as well?

Telling:
This happened on all the git repos i tried. I've just tried with a mercurial repo, and there it seems to work (however with the mercurial prompt on another line), see below.

```sh
~ » cd git-repo
~/git-repo » cd ..
~ » cd mercurial-repo
~/mercurial-repo 1
[default] »
```

PR #3822, opened 2015-04-28
Allow setting path to nvm install dir.

wtcross:
This PR is a very small change to the `nvm` plugin.

The plugin is now closer to the code that the `nvm` install script puts in the `.bashrc` file by default. This allows for specifying where `nvm` is installed so that global installs can be achieved. The plugin now looks for the `NVM_DIR` environment variable that should contain the `nvm` install path. The default value is `~/.nvm` to not break config of people using this plugin.

### DIFF
--- a/plugins/nvm/_nvm
+++ b/plugins/nvm/_nvm
@@ -1,7 +1,7 @@
 #compdef nvm
 #autoload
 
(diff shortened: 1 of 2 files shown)


mcornella:
Fixed on master.

Issue #2030, opened 2013-08-08
completion of files with a common suffix fails


v76591wC:
I have problem completing files with a common suffix.

Steps to repeat:
* Clone a fresh oh-my-zsh
* create directories to test: mkdir 1test 2test
* try to complete them: ls test[TAB][TAB]

First [TAB]: I see Ttest with the cursor behind the capital T.
Second [TAB]: Ttest, cursor behind the t at the end

Expected behavior: Best: complete/iterate through the *test dirs. Still good: Don't complete.

Interestingly enough: If I create an additional 1_test directory it works.

What do I need to tweak to get my expected behavior?

PR #3826, opened 2015-04-30
Add hyphen insensitivity

forivall:
I find it quite useful to be able to type a hyphen ('-') and then have it be completed to an underscore. 

This could be bundled in with case insensitivity, but that might be an annoying change for some users.


ncanceill:

This seems to work, but it will require a lo-o-ot of testing. See #2030 for an example of undesirable stuff that could happen.

Also you should add `$HYPHEN_INSENSITIVE` to the [template `.zshrc`](https://github.com/robbyrussell/oh-my-zsh/blob/master/templates/zshrc.zsh-template).

Finally, please do not send PRs from your master branch — instead you should create a dedicated branch for each new PR.

PR #3833, opened 2015-05-01
coffeescript aliases for easy compiling

coopermaruyama:
__cf__: compile and show output
__cfc__: compile & copy
__cfp__: compile from pasteboard & print


robbyrussell:
Hi, can you please include a README?

coopermaruyama:
Done. Does that suffice?

PR #3829, opened 2015-04-30
Update bubu command in brew plugin

walesmd:
`brew upgrade` functionality will be changing soon. This PR adds the `--all` flag to the `bubu` alias provided in the brew plugin to handle this change.

Issue #3824, opened 2015-04-29
Examples for tmux.plugin.zsh?


binarykitchen:
Are there any examples out there how to use tmux.plugin.zsh in conjunction with an SSH session?

Just the basic stuff, thanks!

ncanceill:
Please see the [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins#tmux) on how to use the plugin.

Also note that, when you SSH to a remote machine, you open a shell on that machine, so Zsh, OMZ and the tmux plugin should be installed on that remote machine.

Maybe I did not understand what you were asking, can you try to be a bit clearer?

binarykitchen:
yes, i saw that but there are no examples and i didn't know the plugin must be enabled on that remote machine.

all i need is some examples how you are using the plugin. i mean, how you ignite a ssh session from localhost to the remote machine that never dies by using that plugin?

ncanceill:
As I explained above, when you open an SSH session, you are actually using a shell on the remote machine. Your local shell configuration is thus irrelevant. If you want Zsh and OMZ when you SSH to the remote machine, you need to install Zsh and OMZ on the remote machine.

A bit more about SSH: when you connect to a remote machine, it is acting as SSH server. With a default configuration, the server will automatically disconnect clients after a certain period of inactivity. When you get disconnected, the SSH session is closed, and programs you have launched from that shell session get interrupted.

***

Also, before we get to the plugin, you need to understand what the [`tmux`](http://tmux.sourceforge.net) program is: a terminal multiplexer, kind of like `screen`. When you launch it from a shell, it allows you to open multiple programs within the same terminal; also, the programs will not be interrupted when the shell session is closed, instead they will be detached and waiting to be re-attached to a `tmux` session.

So, technically, `tmux` will not _"ignite a ssh session_ [...] _that never dies"_. What it allows you to do is:
1. Use SSH to open a shell on the remote machine
2. From that shell, launch `tmux`
3. From that `tmux` session, launch your programs and do your stuff

Then, if you get disconnected, your programs are not interrupted. You can simply:
1. Use SSH to connect again
2. Use `tmux` to re-attach to your programs
3. Continue working like it never happened

Instead, if you really want to keep you SSH sessions from disconnecting entirely, you should use the [`*AliveInterval` config options](http://www.openssh.com/faq.html#2.12).

***

[...]

PR #3840, opened 2015-05-02
Update brew.plugin.zsh

szhousun:
add argument to brew upgrade


Kriechi:
Thanks, but I think we already got this one covered...
#3816 #3829 #3824 and of course #3809 already contain the same fix.
Please close this PR.

PR #3842, opened 2015-05-02
Updating README

matheusrocha89:
This Change remove the .zsh on the plugin activate example, cause with the .zsh at the end the plugin won't be activated


ncanceill:
Thanks!

PR #3846, opened 2015-05-03
Plugins: add a plugin helper plugin

KhasMek:
The purpose of this plugin is to assist with the use
  and interaction of other plugins.

  Current features are as follows
  - list all plugins
  - list all enabled plugins
  - list all plugins with README files
  - print the README file of specified plugin (similar
    to the man function for other CLI programs)
  - print all aliases of specified plugin(s)
  - enable a specified plugin
  - disable a specified plugin

### DIFF
--- a/plugins/plugin-helper/LICENSE
+++ b/plugins/plugin-helper/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 KhasMek
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


ncanceill:

The MIT license file is superfluous since the whole project is already under MIT license.

More importantly though, we already have #2690 for that purpose, but:
> We've yet to come to a clear consensus on how to manage plugins
(see [#1035(comment)](https://github.com/robbyrussell/oh-my-zsh/pull/1035#issuecomment-53995503))

KhasMek:
Hi, @ncanceill and thanks for taking the time to respond.

Thanks for the reminder about the redundant licensing. In regards to the patch, I submitted this before I started looking too far into the existing Issues and PR's. Once I did I soon realized that this was most likely unnecessary. But, I'm glad it's already in the works!

I'll pay closer attention with my future PR's as to not cluster up this section and will keep looking at current issues and PR's and help out where I can.

Thanks again!

ncanceill:
Great! You might want to check the [wiki page on volunteering](https://github.com/robbyrussell/oh-my-zsh/wiki/Volunteers).

PR #3849, opened 2015-05-04
Add git alias for force push

natalia-mentormate:


ncanceill:

This would be taken care of in #2790 if there were not [some voiceful concerns](https://github.com/robbyrussell/oh-my-zsh/pull/2790#discussion_r29489811) about aliasing such a destructive command. You can close this PR and contribute to the discussion over there.

PR #3853, opened 2015-05-05
Added --all flag to brew update

wlonk:
This will stop that annoying warning.

*For your time, here's a cute animal picture…*

otter-and-badger


Philosoft:
we've got this covered https://github.com/robbyrussell/oh-my-zsh/pull/3840

wlonk:
Hah, thanks. Looks like it's covered all over the place.

PR #3859, opened 2015-05-06
gradle plugin: also list tasks in subprojects

gnagy:
With this fix gradle autocomplete also lists tasks in subprojects (separated by ':') and task and project names containing '-'. I also simplified the mathing by using awk.


ncanceill:
cc @matthewmccullough

mcornella:
Same as in #3813, but also allows for hyphens and simplifies the math.
If this works, you have my :+1:

mcornella:
Hi, Robby went with #3813 instead, but this is still relevant since it reads hyphens and simplifies the number of commands used. Can you rebase so we can merge this as well?

PR #2949, opened 2014-07-11
AWS plugin: alias for easier update of access keys

cristim:
This change initially included support for MFA authentication, which was later moved to a dedicated otp plugin.

Now it includes an alias for acces key updates and a few code style changes.

PR #3862, opened 2015-05-06
otp: new plugin for MFA one-time passwords using oathtool, integrated with the clipboard

cristim:
This plugin allows you to create one-time passwords using oathtool, able to replace MFA devices.

The oathtool key is kept in a GPG-encrypted file so the codes can only be generated by an user able to decrypt it.

Provided aliases:
- otp_add_device - creates a new encrypted storage for an oathtool key and stores it on the disk. For encrypting the key, it will ask for a GPG user ID(your GPG key's email address). Then the OTP key needs to be pasted, followed by a ^D character inserted on an empty line.
- ot - generates a MFA code based on the given key and copies it to the clipboard(on Linux it relies on xsel, on MacOS X it uses pbcopy instead)

### DIFF
--- a/plugins/otp/otp.plugin.zsh
+++ b/plugins/otp/otp.plugin.zsh
@@ -1,0 +1,54 @@
+export OTP_HOME=~/.otp
+mkdir -p $OTP_HOME
+
+function ot () {
+  if ! command -v oathtool > /dev/null 2>&1; then
+    echo "Note: you need to install oathtool or oath-toolkit, depending on your OS or distribution."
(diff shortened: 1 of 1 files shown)


cristim:
@mcornella do you have any updates on this?

I have a few local branches for such PRs which I'm having a hard time keeping up-to-date if my PRs aren't merged.

mcornella:
Merged based on comments liking it at #2949. A README would be great to have, you can look up other plugin READMEs to get a sense of the general structure.

PR #3865, opened 2015-05-07
Schroot plugin

fshp:
Plugin for get current name of chroot (with help schroot)


fshp:
Example

fshp:
Moved to external [plugin](https://github.com/fshp/schroot.plugin.zsh)

Issue #3844, opened 2015-05-03
debian plugin 'ag' alias overrides silversearcher app


oryband:
Many people, including myself, are using [the silver searcher](http://geoff.greer.fm/ag/) as a better grep/ack. debian plugin overrides it. You should probably disable this alias if the application is installed, or just find a different alias.

Is there a way to turn this alias off instead?

apjanke:
Duplicate of #3633.

KhasMek:
Please test out that patch and let me know how it works for you @apjanke 
Maybe my implementation isn't the cleanest/best idea for the project, but it's working just fine on my end.

oryband:
closing because duplicate of #3633 , @KhasMek i've replied on the original issue

Issue #3633, opened 2015-02-24
The nuptial problem about "silversearcher - ag". 


jachinpy:
One small problem, in the ".oh-my-zsh/plugin/ubuntu" alias system,  use ubuntu "ag" command and search tool name repetition problems "silversearcher-ag".

apjanke:
I'm having a hard time understanding this, but it sounds like the problem is that the Silver Searcher program defines the command `ag`, and the Ubuntu plugin redefines `ag` with `alias ag='sudo apt-get' `, which clobbers the Silver Searcher's `ag` command. So when you run `ag`, you intend to call silversearcher, but end up running apt-get.

As a workaround, you can add `unalias ag` in your `~/.zshrc` file, after loading oh-my-zsh. This will remove the ubuntu plugin's alias and make the Silver Searcher `ag` command visible again.

Caused by #3601, which added the ubuntu plugin. @nextgenthemes, what do you think about the issue of `alias ag` masking the Silver Searcher command?

nextgenthemes:
I do not know how things are handled here on oh-my-zsh, this was my first contribution. I do not know Silver Searcher and I do not know how many users will want to use it and the ubuntu plugin at the same time. I would say its not to much to ask to add `unalias ag` if your running into this issue.

Feel free to make a pull request to my plugins readme that warns people about this. Or add some check if silverseacher is active and if so use another alias or something but I would like to keep that alias the way it is. But its not up to me.

oryband:
duplicate or origin of #3866

akaihola:
`unalias ag` solves this only partially. Command line completion is still broken. If I type `ag <tab>`, I get this:
```bash
$ ag (eval):1: command not found: _ag
(eval):1: command not found: _ag
(eval):1: command not found: _ag
```

Issue #3682, opened 2015-03-11
'gap' is a program


ChrisJefferson:
Since installing oh-my-zsh, I kept finding that running 'gap' would start some weird git related thing. 

I have now discovered this is a git alias added by the git package. Perhaps if there is a program called 'gap', that should be used instead of the alias? Or choose a different one?

apjanke:
Sounds like you're referring to `gap` from GAP (https://github.com/gap-system/gap)?

@Stibbons and @ncanceill are the maintainers of the git plugin. Can you folks comment?

@ChrisJefferson, as a workaround, you can `unalias gap` after sourcing `$ZSH/oh-my-zsh.sh` in your `~/.zshrc`.

ChrisJefferson:
Yes, it is that `gap`. I didn't know about unalias (probably should have done), I can at least use that for now to get my machine working again. Don't know if there will always be an argument over these short names.

ncanceill:
Hi @ChrisJefferson,

I had never heard of the [GAP system](http://www.gap-system.org), and I did not know about a GNU program called `gap`. This is obviously a conflict, and there are two choices when it comes to this:
- changing the `gap` alias to something else
- not doing anything

***

We do not like modifying aliases, because it may break backwards-compatibility for existing users of the plugin. This means we need a really good reason for doing so: proof that this conflict affects many users.

This is where you come in. Since you want this problem to be solved, your task is now to find other people like you: people using both GAP and the Git plugin. When you find them, get them to +1 this issue, and we will consider changing the `gap` alias to something else.

If you choose this option, please keep this issue open until you find enough of them or give up.

***

In case you do not complete the above task, the alias will definitely not be modified. This would leave you with a few options:
- use `unalias` in your `~/.zshrc` as [suggested above](#issuecomment-78300671)
- use another alias for your `gap` command, for instance:
```
alias the_real_gap="/path/to/the/real/gap"
```
- maintain your own fork of OMZ and edit `$ZSH/plugins/git/git.plugin.zsh` to your needs

If you choose any of these options, please close this issue.

***

Thank you @apjanke for the ping. Next time @ChrisJefferson please notify the maintainers of the affected plugins.

ChrisJefferson:
I'll just stop using the git plugin.

ncanceill:
OK, this was not in the options I listed, but suit yourself.

Sorry but for the sake of compatibility we cannot easily change plugins aliases. Any other suggestion is welcome.

[...]

PR #3866, opened 2015-05-07
Plugins: ubuntu - account for silversearcher

KhasMek:
The alias 'ag' in the ubuntu plugin causes conflicts
  if the user also has [The Silver Searcher](http://geoff.greer.fm/ag/) installed. This patch checks
  to see if the silversearcher binary is available, if so,
  the aliases is not made.

This addresses issues #3633 and #3844

### DIFF
--- a/plugins/ubuntu/ubuntu.plugin.zsh
+++ b/plugins/ubuntu/ubuntu.plugin.zsh
@@ -26,7 +26,9 @@ compdef _afu afu='sudo apt-file update'
 alias ppap='sudo ppa-purge'
 compdef _ppap ppap='sudo ppa-purge'
 
-alias ag='sudo apt-get'            # age - but without sudo
+if [[ -z $(command -v ag) ]]; then
+  alias ag='sudo apt-get'          # age - but without sudo
+fi
 alias aga='sudo apt-get autoclean' # aac
 alias agb='sudo apt-get build-dep' # abd
 alias agc='sudo apt-get clean'     # adc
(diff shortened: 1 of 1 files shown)


oryband:
@KhasMek you should probably add a "toggle-on" variable to apply this alias: when an environment variable named `APPLY_AG_ALIAS` is `"true"`, apply the alias. Otherwise don't (also when *ag-silversearcher* is installed). This way, people will be given the choice whether to use this or not in cases where they do or don't have ag-silversearcher.

wdyt?

KhasMek:
@oryband that is certainly doable, nor am I really opposed to it. However, IMHO, all the extra characters for the flag check in the plugin, then setting the flag either via CLI or in a dotfile seems like a bit of a waste when the user could just manually set the alias via CLI or a dotfile for the occasions they wish to use it over Silver Searcher.

I think it may be best to wait to see the overall opinion of the project/plugin maintainers before any further modification to the PR is made. Sound good to you?

oryband:
@KhasMek yeah sure. Alternatively we can just document this "feature" and explain how to override it. :)

ncanceill:
@blueyed and I are working on a safe aliasing system which should take care of that kind of stuff.

oryband:
ping?

ncanceill:
Ouch it has been a while... sorry guys.

For reference, the discussion I was referring to began with https://github.com/robbyrussell/oh-my-zsh/issues/3682#issuecomment-78437079

I have to read that again and get started.

PR #3869, opened 2015-05-08
Don't clobber default LSCOLORS, if DISABLE_LS_COLORS is set to true.

shreevatsa:
This change makes oh-my-zsh touch the LSCOLORS variable only if DISABLE_LS_COLORS is not set.

Background:
I don't want colors by default, and I manually run `ls -G` (on Mac OS X) when I want colors.

For users who've grown accustomed to the default `ls` colors, it's a bit jarring to have oh-my-zsh show everything in new colors, even when they think they've turned off this setting with `DISABLE_LS_COLORS=true`.

Anyway, there is no good reason to `export LSCOLORS` when `DISABLE_LS_COLORS` is set, so this change should be strictly better.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -1,10 +1,10 @@
 # ls colors
 autoload -U colors && colors
-export LSCOLORS="Gxfxcxdxbxegedabagacad"
 
 # Enable ls colors
 if [ "$DISABLE_LS_COLORS" != "true" ]
 then
+  export LSCOLORS="Gxfxcxdxbxegedabagacad"
   # Find the option for using colors in ls, depending on the version: Linux or BSD
   if [[ "$(uname -s)" == "NetBSD" ]]; then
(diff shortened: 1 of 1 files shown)


apjanke:
Looks good to me.

While you're in here, do you want to reformat the `if` statements to uniformly use `[[ ... ]]` double brackets and `; then` on the same line as the `if`?

shreevatsa:
I actually deleted the repo since this was ignored for so long, and I gave up. :-)
To make those additional changes I'll have to create another fork and make a new pull request, but really I have given up on trying to use oh-my-zsh altogether after a few traumatic experiences like losing some of my years-long history. :-)

apjanke:
Oh well.

Would you mind closing this PR then? I'll just fold the change in to another active PR I have open, like #3743.

shreevatsa:
(Done, closed this... just adding a comment in case there was no notification earlier.)

Issue #3516, opened 2015-01-23
Authentication failure at "Changing shell for x" 


bitcasual:
Authentication failure when running both the wget and curl installers. It occurs both when I'm using su -c and sudo in latest Fedora, as well as when I'm running without root privileges.

apjanke:
I'm having trouble reproducing. The installer is working fine for me on Fedora 21 as a normal user or as root. (Though I wouldn't recommend setting up oh-my-zsh for root.) 

You shouldn't need `su -c` or `sudo`, since oh-my-zsh is a per-user installation and installs in to your home directory. It does not require root privileges to install.

Is this still happening for you? Which version of Fedora are you running?

apjanke:
Okay, I can reproduce a similar error now. Does this look like the error you were seeing?

Here's what I see on Debian 7, as a newly created user whose current default shell is `/bin/bash`.

```
alice@debbie:~$ curl -L https://raw.github.com/robbyrussell/oh-my-zsh/master/tools/install.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1930  100  1930    0     0   5369      0 --:--:-- --:--:-- --:--:--  5369
Cloning Oh My Zsh...
Cloning into '/home/alice/.oh-my-zsh'...
remote: Counting objects: 636, done.
remote: Compressing objects: 100% (500/500), done.
remote: Total 636 (delta 14), reused 501 (delta 4), pack-reused 0
Receiving objects: 100% (636/636), 358.32 KiB, done.
Resolving deltas: 100% (14/14), done.
Looking for an existing zsh config...
Using the Oh My Zsh template file and adding it to ~/.zshrc
Copying your current PATH and adding it to the end of ~/.zshrc for you.
Time to change your default shell to zsh!
Password: chsh: PAM: Authentication failure
alice@debbie:~$ 
```
(I see basically the same thing on Fedora 21, too, with a fresh user account. I think the reason I couldn't reproduce before is that my user's shell was already `zsh` so the installer wasn't calling `chsh`.)

I'm guessing this is happening because the subshell that the installation script is running in is non-interactive, so `chsh` isn't prompting for a password (or it's prompting and reading an empty string instead of giving me a chance to type anything). I can reproduce that by just running `chsh` in a `sh` subshell.

```
[...]

PR #3872, opened 2015-05-09
Update README.markdown

diraol:
Updating the install instructions as proposed in: https://github.com/robbyrussell/oh-my-zsh/issues/3516#issuecomment-76506726


apjanke:
This fixed the README in the Git repo, but the http://ohmyz.sh/ front page is still using the old form. Some people may still be seeing that and getting the `chsh` errors, like #4131. @robbyrussell, can the web site be updated too?

PR #3874, opened 2015-05-09
Debian Plugin: Differing upgrade and safe-upgrade

diraol:
Improving with "safe-upgrade" for aptitude and "upgrade" for apt-get on debian plugin


blueyed:
LGTM.

Apart from the `ag` alias shadowing the `ag` program (but that's not added in this PR).

PR #3879, opened 2015-05-11
Make friendly to `set -u`

steshaw:
I like to have `set -u` turned on in my shell. These changes fix the problems I ran into with oh-my-zsh.

Environment variables (that may not be set) are accessed via ${FOO:-} instead of just ${FOO}.


ncanceill:
This seems like an awful hassle, is it really necessary?

Can you not simply `set -u` after sourcing OMZ?

steshaw:
I like scripts to be `set -u` safe.

ncanceill:
No need to worry, as you can see with prepending "x" to the string we are safe _vis-à-vis_ empty variables.

I advise you trust OMZ to deal with unset variables, and then `set -u` after sourcing OMZ, so you are safe with any other scripts you may run.

steshaw:
Prepending "x" to the string is an old Borne Shell trick that is not
required when you use the modern [[ ]]. I had thought about changing that
too while I was in there but wanted to keep the change as small as possible.

I noticed that only in a few places the older [ ] style test is used. These
could all be replaced with [[ ]] and then any double quoting and "x"
prepending tricks removed.

See https://gist.github.com/steshaw/16218d663f39cc6e64d3 for an example.

I can understand you wanting to reject my patch for the `set -u` safety. It
does help prevent errors (with misspelled variable names) but doesn't seem
to be a very popular technique amongst shell programmers. I don't
understand why though. I like it :).

ncanceill:
@robbyrussell @mcornella @apjanke what do you think?

blueyed:
:+1:

apjanke:
Hmm. Nothing personal, but this extra defensive/hygienic stuff using fine-grained syntax (as opposed to higher-level design) kind of makes my eyes ache, and is one of the reasons I switched to `zsh` from `bash`. I would be bummed to see it showing up much. But it's not my framework.

There's a *lot* of different shell options in `zsh` that can change its behavior, and if we try to defend against many of them on a line-by-line basis like this, OMZ is going to end up looking like a Perl script. That's a bummer, because one of the nice things about OMZ is that it provides an easy on-ramp for users who are new to Zsh and shell in general.

If you want to code defensively against `set -u` and the like, instead of making pervasive syntax changes against particular errors, how about using `emulate -L zsh` inside functions to locally restore all options to a state that's known to work with the syntax of the code you're writing? That would be both more robust and IMHO easier on the eyes. Simply adding anonymous function wrappers using `() { ... }` to the library files would give you a context for `emulate -L` to work in.

plugins/coffee/coffee.plugin.zsh
```diff
@@ -0,0 +1,16 @@
+#!/bin/zsh
+
+# compile a string of coffeescript and print to output
+cf () {
```
georgkreimer: conflicts with cloud foundry-cli tool!

mcornella:
I agree with the general feeling here that this type of defensive programming looks and feels wrong. That undefined variables are treated as empty is not a bad thing, and I don't think we'll merge this one.

If you want to have it enabled in your shell, you can set it after oh-my-zsh has been sourced so you don't get warnings for the oh-my-zsh codebase.

steshaw:
No worries, I think I'm eventually giving up on shell scripting anyhow :)

mcornella:
@steshaw hey can you close this PR?

PR #3883, opened 2015-05-12
added the --all argument

vassyz:
brew upgrade with no arguments will change behaviour soon


mcornella:
Hi! The homebrew team backtracked on the --all parameter decision: Homebrew/homebrew@a45a34d.
Please close this

PR #3884, opened 2015-05-12
Add basic bzr support in the avit theme

rpadovani:
I love oh-my-zsh, thanks for your awesome work!

I use avit theme (with some different colors) and I use often bzr, so I added a very basic support to vcs_info in the avit theme, trying to stick the avit style. 

It does only two things: 
- print repo name and number of commit
- report if there are local changes

I plan to investigate further to see if it's possible to add more things, but at the moment it's the best I can do, so I propose to merge this.

Thanks,
R.
screenshot from 2015-05-12 14 20 26

### DIFF
--- a/themes/avit.zsh-theme
+++ b/themes/avit.zsh-theme
@@ -1,7 +1,7 @@
 # AVIT ZSH Theme
 
 PROMPT='
-$(_user_host)${_current_dir} $(git_prompt_info) $(_ruby_version)
+$(_user_host)${_current_dir}''${vcs_info_msg_0_}'' $(git_prompt_info) $(_ruby_version)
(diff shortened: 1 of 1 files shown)


apjanke:
This is a substantial change, since it's now using zsh's vcs_info support. Probably should be tested first.

`local` does not work in this context, since it's not inside a function call. If you really want them to be local, wrap that section in an anonymous function.

CC @avit; you're the theme's author. Comments?

mcornella:
You can extend the theme copying it as a custom theme. See [Customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes).

PR #3891, opened 2015-05-13
Use existing ssh-agent when invoking a sudo shell

stucki:
When invoking a shell as root using `sudo -s`, the ssh-agent plugin starts a new agent although it already exists.

The problem boils down to a check if ssh-agent is running using `ps x`. If that is extended to `ps ax` for root, then the existing ssh-agent will still work.


stucki:
Status: Issue still exists, and the fix still works fine.

stucki:
Updated this PR. I still find this quite annoying.

mcornella:
Thanks!

PR #3892, opened 2015-05-13
Added environment based shortcuts in symfony2 plugin

delphiki:
I've been using the "dev" and "prod" aliases for a while.
They've proven to be really useful when you type in commands based on symfony environment.
Credit for this tip goes to @hhamon 

The only downside I see is that the names of these aliases might not be specific enough.

### DIFF
--- a/plugins/symfony2/symfony2.plugin.zsh
+++ b/plugins/symfony2/symfony2.plugin.zsh
@@ -25,5 +25,5 @@ alias sfcw='sf cache:warmup'
 alias sfroute='sf router:debug'
 alias sfcontainer='sf container:debug'
 alias sfgb='sf generate:bundle'
-alias dev="sf --env=dev"
-alias prod="sf --env=prod"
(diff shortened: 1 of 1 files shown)


ncanceill:
cc @andrewtch

andrewtch:
Wow, guys, me no expert here, although I wrote the original version of this plugin (if I recall it correctly XD)

Well, to be serios, the only advice would be to follow original ``sfsomething`` notation and use, for example, ``sfdev`` and ``sfprod`` aliases as ``dev`` and ``prod`` aliases are way too generic.

delphiki:
That might be better now :)

mcornella:
Merged. Care to submit a README with the list of aliases like in [this example](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/dnf/README.md)? :wink:

delphiki:
@mcornella Thanks! I'll add the README asap.

PR #3901, opened 2015-05-17
Update taskwarrior.plugin.zsh

regagain:
Corrected documentation to avoid confusion (see https://github.com/robbyrussell/oh-my-zsh/issues/3796)


ncanceill:
Thanks!

PR #3911, opened 2015-05-20
added ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE

mbologna:
In addition to the already existing variables ZSH_THEME_GIT_PROMPT for behind, ahead and diverged remote, I added ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -36,7 +36,10 @@ git_remote_status() {
         ahead=$(command git rev-list ${hook_com[branch]}@{upstream}..HEAD 2>/dev/null | wc -l)
         behind=$(command git rev-list HEAD..${hook_com[branch]}@{upstream} 2>/dev/null | wc -l)
 
-        if [ $ahead -gt 0 ] && [ $behind -eq 0 ]
+        if [ $ahead -eq 0 ] && [ $behind -eq 0 ]
+        then
+          echo "$ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE"
+        elif [ $ahead -gt 0 ] && [ $behind -eq 0 ]
         then
             git_remote_status="$ZSH_THEME_GIT_PROMPT_AHEAD_REMOTE"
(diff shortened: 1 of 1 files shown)


mbologna:
@mcornella could you please check this out?

mcornella:
A demonstration of this PR:
remote-equal-sign
On themes which do not use the `$ ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE` variable the theme will look the same.

This has my approval, but there are other PRs that want to change the way `lib/git.zsh` works and I don't think we'll get this merged anytime soon. We have to compile them and triage them and choose the best way to move forward while maintaining backwards compatibilty. Of course, we could just merge this right away and change whatever we need to later.

Let's merge this yes
:+1: /cc @robbyrussell

mbologna:
This won't change the default behavior, just adds this additional property. It does not break existing themes :)

lib/git.zsh
```diff
@@ -36,7 +36,10 @@ git_remote_status() {
         ahead=$(command git rev-list ${hook_com[branch]}@{upstream}..HEAD 2>/dev/null | wc -l)
         behind=$(command git rev-list HEAD..${hook_com[branch]}@{upstream} 2>/dev/null | wc -l)
 
-        if [ $ahead -gt 0 ] && [ $behind -eq 0 ]
+        if [ $ahead -eq 0 ] && [ $behind -eq 0 ]
+        then
+          echo "$ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE"
```
sgrell: Unfortunately this change did break existing themes. I'm using the strug theme and the echo output causes some wired line break. Changing the echo output to a variable assignment fixed it for me. ```git_remote_status="$ZSH_THEME_GIT_PROMPT_EQUAL_REMOTE"```

mbologna: I tried to reproduce the problem, without success. Looks good to me:
screen shot 2015-09-22 at 9 51 30 pm

sgrell:
Your first prompt shows the problem with (origin/master) on the second line. When you checkout a commit it doesn't seem to output this part, I don't know why, I didn't dig that deep. 

I can reproduce that behaviour when I only change that one line.

oh-my-zsh

mbologna:
@sgrell you're right, corrected in #4387

PR #3914, opened 2015-05-20
Timebox git calls in git_prompt_dirty

### LINKED ISSUE #3009: Why is the git plugin slow for large repos?

### LINKED ISSUE #3706: Weird problem on a Vagrant VM with NFS shared folders

apjanke:
This "timeboxes" the calls to `git` inside `git_prompt_dirty` to prevent the prompt from locking up when you're in a slow git repo. If the `git` call takes too long, it aborts it and just puts "?" for the `git_prompt_dirty` status. The timeout length and "timed-out" prompt text are user-configurable.

This only timeboxes the `git` call in `git_prompt_dirty` (aka `parse_git_dirty`, renamed for consistency with the other prompt info functions). `git_prompt_info` uses `git_prompt_dirty`, so this will help its slow `git` calls, too. This does not affect the more detailed `git_prompt_status`. I couldn't figure out how to cleanly timebox its `git` calls, which use more than one line of output.

Fixes #3009.
Fixes #3706.
Supersedes #3725. This is the same logic, rebased and squashed, and only for `git_prompt_dirty`.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -1,4 +1,9 @@
-# get the name of the branch we are on
+# lib/git.zsh - Various functions for incorporating git status info 
+# into the zsh PROMPT.
+
+
+# Constructs the git info section of the prompt
+# This prompt section includes the current branch and a short clean/dirty indicator
 function git_prompt_info() {
   if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
     ref=$(command git symbolic-ref HEAD 2> /dev/null) || \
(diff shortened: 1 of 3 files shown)


robbyrussell:
@apjanke Would you be up for helping us get this rebased and pushed out?

justin808:
Hi @robbyrussell, I just tried _any_  command on a network drive with git directories copied using CCC and every file is modified:

```
old mode 100644
new mode 100755
```

so `git status` takes like 30 seconds!

Why is this PR closed?

Sounds useful!

robbyrussell:
@justin808 I closed it as we hadn't had anybody raising this as a need despite it being an open PR for several years. We'd welcome a fresh PR if someone is keen to explore this, again

PR #3920, opened 2015-05-22
added $OSTYPE check for cygwin

igleyy:
It makes possible to use jira plugin since xdg-open is not available on Cygwin.

### DIFF
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -14,6 +14,8 @@ open_jira_issue () {
   local open_cmd
   if [[ "$OSTYPE" = darwin* ]]; then
     open_cmd='open'
+  elif [[ "$OSTYPE" = cygwin ]]; then
+    open_cmd='cygstart'
   else
     open_cmd='xdg-open'
   fi


apjanke:
:+1:

apjanke:
Conflicts with #2988. This PR should probably auto-close it, since it uses the preferred `$OSTYPE` instead of shelling out to `uname`.

mcornella:
Agreed. Though I'd go even further and standardise these sparse and duplicated logic into one unique library function / alias as I did in the [web-search plugin](https://github.com/robbyrussell/oh-my-zsh/blob/ef7e53a78d0e4196c2d1e6e5b268209759d51753/plugins/web-search/web-search.plugin.zsh#L17-L24). 
The `frontend-search` and `node` plugins have these types of branches, and then there are other plugins that just use `open` but could use this standardised function instead: `bwana`, `fasd` (#1478), `lighthouse`, as well as `marked2`, `osx`, `textmate` and `xcode` that are OSX-specific so we don't bother with those.

apjanke:
Yeah, centralizing it in a shared lib function makes a lot of sense.

mcornella:
I finally implemented a working solution: #4217. Please test if it works correctly

PR #3921, opened 2015-05-22
Add Marker plugin

pindexis:
Marker is a simple tool that enhances the productivity when working on the command-line.
It allows "bookmarking" commands and quickly fetching them without using shell aliases/functions .
The Project [ReadMe](https://github.com/pindexis/marker) contains more details.

Hoping this gets merged :)


mcornella:
Your marker function looks really cool, I specially like the fuzzy matching part (I fell in love with it first in Sublime Text), but I don't think this belongs here: it's essentially adding a source statement in your zshrc file, which you already indicate when running the installer. I believe having this plugin will not make it easier than it is now for your users.

I liked the idea though (and sorry for the late response)!

PR #3925, opened 2015-05-24
Added references to the external themes wiki page in README

dritter:
Added references to the `external themes` wiki page.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -67,6 +67,8 @@ To use a different theme, simple change the value to match the name of your desi
 
 Open up a new terminal window and your prompt should look something like...
 
+In case you did not find a suitable theme for your needs, please have a look at the wiki for [more themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes).
+
 ## Advanced Topics
 
 If you're the type that likes to get their hands dirty... these sections might resonate.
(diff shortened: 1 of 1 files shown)


dritter:
Hmm.. Three mentions of one wiki page in the same file.. Not the best solution, but if it helps to minimize the amount of pull requests.. Hopefully, this is all right.

README.markdown
```diff
@@ -122,6 +124,8 @@ If you have many functions that go well together, you can put them as a `abcyzea
 
 If you would like to override the functionality of a plugin distributed with Oh My Zsh, create a plugin of the same name in the `custom/plugins/` directory and it will be loaded instead of the one in `plugins/`.
 
+Feel free to add your theme to the [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes), if you want to share it with others.
```
ncanceill: I believe this reference is not really necessary, which would cut it down to two.

ncanceill:
Also could you please squash your commits to keep the history clean and atomic?

dritter:
@ncanceill All done. 
Now there are just two mentions of the "external themes" wiki page and this is just one commit.

ncanceill:
Neat!

PR #3926, opened 2015-05-25
Alias gdtr to git diff-tree

marshall-lee:
`gdt` cannot be used for `git diff-tree` because this alias is being overriding later for `git difftool` so there is a conflict in names.
I propose to do renaming for `git diff-tree` alias but not for `git difftool` alias because:
1) `gdt` alias may already been used by someone for `git difftol`.
2) We already have a `gmt` for `git mergetool` so having a `gdt` for `git difftool` is less surprising.


marshall-lee:
cc @Stibbons

mcornella:
@marshall-lee we recently pushed PR #2790 that made a lot of changes to the git plugin, one of them being the removal of the duplicate `gdt` alias, in favor of the `git diff-tree` version. 
If you think we should change that you need to rebase this PR against the current master and make the change you want to see.

ncanceill:
Also, @Stibbons is not longer maintaining the Git plugin, I am.

Anyway, @mcornella put it perfectly, just follow what he wrote.

Stibbons:
+1

marshall-lee:
Things are more complicated than I thought. There's no sense in this PR anymore.

PR #3931, opened 2015-05-26
Fixes detection of homebrew installed rbenv

elektronaut:
This test doesn't work at all, `brew --prefix rbenv` simply returns `/usr/local/Cellar/rbenv/0.4.0` if rbenv isn't installed. Changing it to `brew ls rbenv` fixes it.

It's a performance improvement if rbenv isn't installed through homebrew. On my machine, zsh startup time is reduced from 0.59s to 0.46s. 

Ideally, I'd love to get rid of both calls to `brew ls rbenv` and `brew --prefix rbenv` if possible, they're quite expensive. Removing both cuts my zsh startup time by almost half. 

This change only affects Homebrew users on OS X.

### DIFF
--- a/plugins/rbenv/rbenv.plugin.zsh
+++ b/plugins/rbenv/rbenv.plugin.zsh
@@ -3,7 +3,7 @@ _homebrew-installed() {
 }
 
 _rbenv-from-homebrew-installed() {
-  brew --prefix rbenv &> /dev/null
+  brew ls rbenv &> /dev/null
 }
 
 FOUND_RBENV=0
--- a/plugins/rbenv/rbenv.plugin.zsh
+++ b/plugins/rbenv/rbenv.plugin.zsh
@@ -3,7 +3,7 @@ _homebrew-installed() {
 }
 
 _rbenv-from-homebrew-installed() {
-  brew ls rbenv &> /dev/null
+  [ -e "$(brew --prefix)/opt/rbenv" ]
 }
 
 FOUND_RBENV=0


elektronaut:
I found a faster way of detecting rbenv through homebrew, this brings zsh startup time down to 0.36s on my machine (down from 0.59s). 

It should also be a slight improvement if rbenv is installed through homebrew, I timed it to 0.51s.

mcornella:
This needs testing. Anyone with a OSX machine can confirm this works?

elektronaut:
Looks like there was some work done in #4248 to address part of the problem. Here's the relevant code as it is today:

```zsh
rbenvdirs=("$HOME/.rbenv" "/usr/local/rbenv" "/opt/rbenv" "/usr/local/opt/rbenv")
if _homebrew-installed && rbenv_homebrew_path=$(brew --prefix rbenv 2>/dev/null); then
    rbenvdirs=($rbenv_homebrew_path "${rbenvdirs[@]}")
    unset rbenv_homebrew_path
fi
```

The issue here is with the second clause of the `if` statement:

```zsh
rbenv_homebrew_path=$(brew --prefix rbenv 2>/dev/null)
```

It's trying to detect if rbenv is installed through homebrew, and adds it to the list of paths if it is. Problem is, the test doesn't actually do anything. `brew --prefix rbenv` always returns true.

With rbenv installed:

```
$ brew --prefix rbenv
/usr/local/opt/rbenv
$ echo $?  
0
```

Without rbenv installed:

```
$ brew --prefix rbenv
/usr/local/Cellar/rbenv/0.4.0
$ echo $?  
0
```

Functionally, this is not a big deal. It's just a list of places where rbenv *might* live, adding one extra bogus path doesn't break anything. 

In any case, the logic is flat out wrong, so the test should probably either be fixed or simply eliminated. This code would be functionally equivalent:

```zsh
rbenvdirs=("$HOME/.rbenv" "/usr/local/rbenv" "/opt/rbenv" "/usr/local/opt/rbenv")
if _homebrew-installed; then
    rbenvdirs=($(brew --prefix rbenv) "${rbenvdirs[@]}")
fi
```

mcornella:
Hi @elektronaut, #3808 also tries to fix the logic. Can you check it out?

PR #3932, opened 2015-05-27
some characters of the command repeated on the screen when completing

dominikwinter:
need to be included within a percent-braces construct %{…%}
see: http://unix.stackexchange.com/questions/90772/first-characters-of-the-command-repeated-in-the-display-when-completing


apjanke:
What OS and terminal emulator are you running on? The current code works for me, but this change seems to introduce an error, where the text backs up over the last couple characters of the prompt when doing a complete.

Here's a screenshot of me running with this change on OS X 10.9.5, zsh 5.0.2, in iTerm2. The first line shows text where it normally appears after the prompt. I typed `ls Do` and then hit Ctrl-C to get a fresh line. On the second line, I typed `ls Do` and hit tab. When I hit tab, it jumped back two characters. I don't think that's the right behavior.

screen shot 2015-06-08 at 5 23 54 am

That `${ret_status}` is not zero-width. It includes two printing characters, an arrow and a space. And if you look inside the text of `${ret_status}`, you'll see that the non-printing control characters inside it are already enclused in `${...}`.

What behavior are you seeing that this is attempting to fix?

dominikwinter:
I had the exact same behavior without this change. It looks like there is somewhere an utf8 configuration error in my terminal chain:

(windows) -> putty -> (centos 7 docker container) -> tmux -> zsh

Actually you can close this pull request. I will analyse the problem later.

apjanke:
Only the submitter and the repo owner have permissions to close issues and PRs. Could you close it yourself? Should be a "comment & close" button available on this screen.

Issue #3913, opened 2015-05-20
history-substring-search defines function that collides with beep alias


jbulchergist:
I would love to activate ZLE and get rid of this:

```
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
echo:zle: widgets can only be called when ZLE is active
```

... every time I hit enter. I changed systems from Ubuntu to openSUSE, and now zsh complains about my custom theme. How can I activate ZLE? There doesn't seem to be any documentation on google or any of the zsh, oh-my-zsh projects. I'm not even sure if this is a zsh problem or an .oh-my-zsh problem.

I'd also like to know if there is a way to check to see if the option is set; the documentation for zshzle indicates the option is set by default in interactive shells, so I don't know why it wouldn't be set in mine.

I have no idea where to start troubleshooting this. Help!

-Thanks

mcornella:
You're probably using a zle widget inside a function that gets called everytime you update your prompt. Please post your full custom theme to see what we're dealing with.

jbulchergist:
Thanks for helping with this! Here you go:

```sh
# Name:   trapd00r zsh theme
# Author: Magnus Woldrich <m@japh.se>
#
# This theme needs a terminal supporting 256 colors as well as unicode. It also
# needs the script that splits up the current path and makes it fancy as located
# here: https://github.com/trapd00r/utils/blob/master/zsh_path
#
# By default it spans over two lines like so:
#
[...]

Issue #3724, opened 2015-03-26
af-magic breaks line wrapping on ARM


5paceToast:
Running on ARMv7, using af-magic breaks line wrapping.

Expected behavior: after the cursor reaches the end of the line, it passes a \n and continues on the next line.
Actual behavior: after the cursor reaches the end of the line, it seems to pass a \r (erasing the line from view) and reusing the existing line. If you backspace back unto the present line, the characters are still missing (though in-buffer).

This seems to be an af-magic specific issue, since after testing with a few other teams, that issue would not reproduce (but is 100% reproducible with it, in this environment).

Environment:
- Raspberry Pi 2
- Arch Linux ARM (up to date)
- Version: zsh 5.0.7 (armv7l-unknown-linux-gnueabi)

I might look into af-magic source and/or try compiling zsh manually to see if that changes anything, but for now submitting this (in case it's something obvious I'm missing).
What I do find strange is this only happens on my ARM system...

jgrar:
@5paceToast I've seen similar behaviour but unrelated to ARM and zsh, usually being an issue with PS1, does changing it solve the issue?

Could you provide more information about your current configuration? There's a diagnostic script here: https://github.com/apjanke/oh-my-zsh-custom/blob/master/diagnostics.zsh

paste the output of  `source diagnostics.zsh; omz_diagnostic_dump` to a gist

5paceToast:
Output with af-magic (broken): http://sprunge.us/KJge
Output with gallois (works fine): http://sprunge.us/LEXH
(if you want it in gist specifically, I can do that, I'm just used to sprunge ( `command | sprunge` )

apjanke:
That's a little odd. There's not a whole lot of difference between `af-magic` and `gallois`. Does it happen with other themes that have a multi-line `PS1`, like `amuse` or `adben`?

What terminal are you working in? Are you on toast-dpi locally or ssh'ed in to it?

I've also updated that diagnostic dump function to grab additional information, if you want to do a rerun of it.

apjanke:
Oh, and I think we would prefer gists: that way all of the information is available on GitHub indefinitely for future reference. Sprunge eventually deletes its posts, so we'd lose that data and the bug reports would be less useful.

5paceToast:
[...]

Issue #3356, opened 2014-11-29
complete:13: command not found: compdef


jprice:
I am getting 2 instances of this error when I launch a new terminal window

```
complete:13: command not found: compdef
complete:13: command not found: compdef
complete:13: command not found: compdef
complete:13: command not found: compdef
jprice@xxx:pts/3-> /home » jprice  (0) 
```

I found this: #2315 and I have commented out line 25 in the check_for_upgrades.sh. 

I have also backed up all my dotfiles and removed them, and used a clean install of oh-my-zsh and I still seem to get it.

I'm lost as to what else to do? Any suggestions?

mcornella:
Any updates?

jcrben:
I had this issue for a frustrating hour or so today. There are a lot of old issues and varied suggestions which came up in Google, but the only thing that worked for me was [this](https://github.com/robbyrussell/oh-my-zsh/issues/630#issuecomment-70291622) from #630  

    #This will perform chmod g-w for each file returned by compaudit to remove write access for group
    compaudit | xargs -I % chmod g-w "%"
    #This will perform chown to current user (Windows and Linux) for each file returned by compaudit
    compaudit | xargs -I % chown $USER "%"
    #Remove all dump files (which normally speed up initialization)
    rm ~/.zcompdump*
    #Regenerate completions file
    compinit

apjanke:
Looks like the `rm ~/.zcompdump*` which mcornella suggested. (With a typo fix.)

Next time this happens to someone, could they grab your `~/.zcompdump*` files and post the as gists or a tarball for analysis?

Here's a lead: This looks like an interaction between ZSH's bashcompinit support for bash style completions, and the order in which we load stuff up. Have a look at the `bashcompinit` function in ZSH (ZSH itself, not oh-my-zsh; on my system it's at `/usr/share/zsh/5.0.2/functions/bashcompinit`). It defines the function `complete()`, which AFAIK is not otherwise defined in `zsh`. And on line 13 of complete's function body, it calls `compdef ...`. But it doesn't itself define compdef. So if you load up bashcompinit and call complete without loading up the rest of the zsh completion system, you get that exact `complete:13: command not found: compdef` error message.

Here's me reproducing it in a session with an empty .zshrc.

```
eilonwy% autoload -U +X bashcompinit; bashcompinit
eilonwy% complete foo
complete:13: command not found: compdef
eilonwy%
```

[...]

PR #3940, opened 2015-05-30
Add omz_diagnostic_dump() to help with bug reporting

### LINKED ISSUE #3915: Diagnostic dump function for bug reporting?
Hi, OMZ folks,

A lot of the issues we get are problem reports about OMZ on a particular user's machine, and since there's complex configuration for both the system and OMZ, we end up doing a lot of back and forth to elicit more info from the user. I think it would make bug reports easier if we had a function in OMZ that would just dump out all the relevant state and configuration for OMZ and ZSH, so the user could just submit that instead of answering a series of questions in comments.

I've put one together at [diagnostics.zsh](https://github.com/apjanke/oh-my-zsh-custom/blob/master/diagnostics.zsh) in my oh-my-zsh-custom tree. It dumps out information about the user's system and shells, zsh and environment state, OMZ installation, and grabs copies of all the relevant zsh configuration files.

Any interest in bringing this in to the main OMZ distribution, maybe as `lib/diagnostics.zsh`?

apjanke:
This adds a new `lib/diagnostics.zsh` library file for OMZ/ZSH diagnostic functions, and defines the `omz_diagnostic_dump()` function that will dump a lot of info about the user's current shell state and configuration files to a dump file suitable for posting as a gist.

The intention is that this could be made part of the normal debugging and support workflow by asking bug reporters to just "run `omz_diagnostic_dump()` and post the results" instead of having to ask a series of questions and request individual files.

I've been using this in my custom dir for a while and it's been helpful with diagnosing some issue reports already (#3913, #3356, #3724, #3350).

Closes #3915

### DIFF
--- a/lib/diagnostics.zsh
+++ b/lib/diagnostics.zsh
@@ -104,12 +104,12 @@ function _omz_diag_dump_one_big_text() {
   command date
   command uname -a
   builtin echo OSTYPE=$OSTYPE
-  command zsh --version
+  builtin echo ZSH_VERSION=$ZSH_VERSION
   builtin echo User: $USER
   builtin echo
 
   # Installed programs
-  programs=(sh zsh ksh bash sed cat grep find git posh)
+  programs=(sh zsh ksh bash sed cat grep ls find git posh)
   for program in $programs; do
     local md5_str="" md5="" link_str="" extra_str=""
     progfile=$(builtin which $program)
@@ -119,7 +119,7 @@ function _omz_diag_dump_one_big_text() {
           extra_str+=" $(md5 -q $progfile)"
         fi
         if [[ -h "$progfile" ]]; then
-          extra_str+=" ( -> ${file:A} )"
+          extra_str+=" ( -> ${progfile:A} )"
         fi
       fi
       builtin printf '%-9s %-20s %s\n' "$program is" "$progfile" "$extra_str"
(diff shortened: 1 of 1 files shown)


mcornella:
One more thing: set the locale to default so that everything is in English. I achieved that through `local -x LC_ALL=C`, which is the same as `local LC_ALL=C; export LC_ALL` that you did in #4129.

apjanke:
That's probably a good idea. I'll look at how to structure that: it's a little tricky since this function runs in the same shell session that it's examining, and part of the output is to capture the user's existing `locale` settings. And we'd want any command or file resolution etc that could be affected by locale settings to be done under the user's locale so we can see what they're seeing.

mcornella:
Oh, right. Hadn't thought of that. In my test run this only happened in the `grep` version output, but version number was totally readable after all. On further thought, I don't think there's a necessity to do that.

In `systeminfo`, however, the output in spanish doesn't have `OS Name/OS Version` but since it's the very first thing that's printed you can get it with `systeminfo | head -n 4 | tail -n 2`.

Finally, in Linux and cygwin as I have checked, `md5sum` is generally available so you can try that if `md5` is not found.

apjanke:
Ah, I didn't know that about `systeminfo`. Would `LC_ALL` even fix that? `systeminfo` is a native Windows command, not a Cygwin one, so I don't know that it would respect Unix style locale variables.

apjanke:
How about `shasum`? That seems pretty widely available, and then we don't have to worry about the interface differences between `md5` and `md5sum`.

Switched the `systeminfo` output filtering to your line-based method.

mcornella:
> Ah, I didn't know that about systeminfo. Would LC_ALL even fix that? systeminfo is a native Windows command, not a Cygwin one, so I don't know that it would respect Unix style locale variables.

That's exactly right, forgot to add that but I see you thought about it anyway :smile:

It's perfect now, I think it's ok to merge :+1:
/cc @robbyrussell

apjanke:
Thanks!

robbyrussell:
Thanks @apjanke !

apjanke:
You're welcome! Thanks for getting this merged.

PR #3942, opened 2015-05-31
Plugins: git - be less verbose with commits

KhasMek:
Remove the -v flag by default with the standard aliases gc, gc!, gca, and gca!, while adding new aliases for those that choose to put the entire diffs in their commit message.

IMHO, and from my understanding of the creators of git and many others - showing a full diff in the commit message not only completely breaks the acceptable/expected format of _good_ commit messages, but is also completely redundant and unnecessary. (If you have access to the commit message, more than likely you can access the diff of the commit via a multitude of other methods (CLI tools, GUI tools, web interfaces in a much nicer looking format.)

Am I missing some vital logic behind the reasoning to showing the entire diff in a commit message???


mcornella:
I like it better showing the diff, for the main reason that sometimes I don't know what I'm commiting and stare at the blank line thinking. If I have the diff right there I can quickly see what the changes are. But this PR allows both to coexist, so if there's support for the non-verbose option I'll happily change my workflow to use the `gcv` variants.

If it helps understanding, here's an example of what I see when I verbosely commit:
```diff
    
  
# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
# On branch master
# Your branch is up-to-date with 'origin/master'.
#
# Changes to be committed:
#       modified:   nanorc
#
# ------------------------ >8 ------------------------
# Do not touch the line above.
# Everything below will be removed.
diff --git a/nanorc b/nanorc
index d4c20ae..dce00da 100644
--- a/nanorc
+++ b/nanorc
@@ -1,4 +1,4 @@
-include ~/.nano/syntax/ALL.nanorc
+#include ~/.nano/syntax/ALL.nanorc

 set autoindent
 set const
```

KhasMek:
Sounds good. I'll rebase off of the #2790 and update.

blueyed:
@KhasMek the diff is commented in the commit message - it does not get added to the message itself.

mcornella:
Yeah that's exactly right. I overlooked that part of your description.
For completeness, the equivalent commit of my example above is the following:
```diff
commit b134751b654a70d5d7d0deffa8d8147f7e92c160
Author: Marc Cornellà <marc.cornella@live.com>
Date:   Fri Jun 5 18:28:42 2015 +0200

    Disable nano syntax highlight, for now.

diff --git a/nanorc b/nanorc
index d4c20ae..dce00da 100644
--- a/nanorc
+++ b/nanorc
@@ -1,4 +1,4 @@
-include ~/.nano/syntax/ALL.nanorc
+#include ~/.nano/syntax/ALL.nanorc
 
 set autoindent
 set const
```
You can see that the message doesn't contain the diff. 
Also, this is shown above the diff when doing `git commit -v`:
```diff
# ------------------------ >8 ------------------------
# Do not touch the line above.
# Everything below will be removed.
```

KhasMek:
That certainly is the way it's _supposed_ to function. However, I have noticed several occasions where git freaks out and adds the diff, even though it's below that line. Most recently was when working with a lot of markdown. I'm assuming it's actually a bug in git and need to check their bug tracker as well to see if anything has been filed about it. Seems like it's usually from diffs that show a lot of comments in shell scripts or any other large use of #'s in a commit. You can see one example [here](https://github.com/KhasMek/omz-wiki/commit/da94830c93c9f32f428f3097e4d85e4bc4c858f2).

mcornella:
Hmmm that's quite weird yes. Thanks for providing such a clear sample. Anyway, go ahead with the PR and let's see if more people have the same problem.

blueyed:
I can see that huge diffs might be bad/unnecessary sometimes.

Apart from that it still seems like a misunderstanding.

By default, only comments are being removed from the commit message, not
everything after the scissors line.

See the help for `git commit --cleanup`.  The default appears to be "strip":

    --cleanup=<mode>
        This option determines how the supplied commit message should be
        cleaned up before committing. The <mode> can be strip, whitespace,
        verbatim, scissors or default.

        strip
            Strip leading and trailing empty lines, trailing whitespace, and
            #commentary and collapse consecutive empty lines.

        whitespace
            Same as strip except #commentary is not removed.

        verbatim
            Do not change the message at all.

        scissors
            Same as whitespace, except that everything from (and including)
            the line "# ------------------------ >8 ------------------------"
            is truncated if the message is to be edited. "#" can be customized
            with core.commentChar.

        default
            Same as strip if the message is to be edited. Otherwise
            whitespace.

        The default can be changed by the commit.cleanup configuration
        variable (see git-config(1)).

blueyed:
Given that `-v` would be removed, which I am not in favour of, it's probably not worth it to "duplicate" them with a `-v` variant.  You do not save much by `gcv` instead of `gc -v`, and when you use it regularly, change it in your fork/customization instead (for `gc`).

ncanceill:
Thanks guys for contributing once more ;-)

The puzzle here is that there are many Git versions and configurations that can happen in the wild. @KhasMek mentioned that Git might still include the diff from the verbose message.

Another example is my configuration. I have instructions in the prepared message (in French, which is my locale) saying that both comments and everything below the scissors-line will be (and indeed gets) removed. However, my own manual says that the `scissors` setting works as `whitespace` (but with the scissors-line), and `whitespace` works like `strip` but does not remove comments. All my Git config regarding this is default, so the manual says comments with "#" should not be removed but the prepared message says otherwise, and comments are indeed removed.

I am not in favor of removing the `-v` flag either, and I think we need more people to report problems before we consider removing it.

anpol:
I have a problem with `git commit -v` due to a pre-commit hook defined by my project administrator, i.e. I'm unable to modify it. The pre-commit hook fails on commit messages that have lines longer than 80 characters. Such lines could occur due to the diff appended at the end.

Those who need the diff appended by default could proceed with the following:
```
git config --global commit.verbose 1
```

mcornella:
I've decided to close this. These are very few edge cases, and can be solved by redefining the aliases.

PR #3651, opened 2015-02-28
Round 3 of batch-closing obsolete issues

mcornella:
## 62 issues closed

The reason for closing each issue is explained in the issue itself

Close #1514
Close #2113
Close #2178
Close #2354
Close #2356
Close #2755
Close #2933
Close #2942
Close #2972
Close #2985
Close #2987
Close #3013
Close #3015
Close #3016
Close #3024
Close #3064
Close #3096
Close #3107
Close #3136
Close #3179
Close #3190
Close #3207
Close #3208
Close #3212
Close #3246
Close #3249
Close #3258
Close #3285
Close #3288
Close #3306
Close #3325
Close #3346
Close #3356
Close #3360
Close #3370
Close #3379
Close #3383
Close #3384
Close #3398
Close #3437
Close #3439
Close #3446
Close #3487
Close #3492
Close #3529
Close #3539
Close #3621
Close #3634
Close #3720
Close #3722
Close #3734
Close #3769
Close #3799
Close #3816
Close #3828
Close #3829
Close #3834
Close #3855
Close #3873
Close #3883
Close #3904
Close #3949

PR #3943, opened 2015-05-31
Add --all flag to brew update command

robertd:


Kriechi:
Thanks - I think we already got like 42 PRs fixing this very same thing.
The most promising one would be #3809 (@ncanceill)

mcornella:
This will be fixed when #3809, which will be imminent. I'm adding it to #3651.

DomT4:
You may wish to hold off on that for now. Reversal is under discussion in https://github.com/Homebrew/homebrew/pull/40372.

mcornella:
You're absolutely right. Let's make sure that's not pushed then: https://github.com/Homebrew/homebrew/commit/a45a34db0b2c66f626f3724ae9a52563b4aa94b5

robertd:
Looks like they reverted it. Closing the issue.

mcornella:
Thanks @robertd and @DomT4

PR #3950, opened 2015-06-03
Xcode plugin: use regexp to find projects or workspaces

Palleas:
In a project I'm working on, there are `xcconfig`s files in the project's root directory, which means that 
using the `xc` alias finds and opens the first one instead of the workspace or the project. 
Using `grep` with a regexp fixes this.

### DIFF
--- a/plugins/xcode/xcode.plugin.zsh
+++ b/plugins/xcode/xcode.plugin.zsh
@@ -1,6 +1,6 @@
 #xc function courtesy of http://gist.github.com/subdigital/5420709
 function xc {
-  xcode_proj=`ls | grep "\.xc" | sort -r | head -1`
+  xcode_proj=`ls | grep -E "\.(xcworkspace|xcodeproj)" | sort -r | head -1`
   if [[ `echo -n $xcode_proj | wc -m` == 0 ]]
   then
     echo "No xcworkspace/xcodeproj file found in the current directory."


Palleas:
Is there anything I can do to get this PR merged?

mcornella:
As is now, the `xc` function shouldn't even be working, since `ls` only lists by default non-hidden files (those beginning with a dot are hidden). Are you sure this works?

Palleas:
Yes I'm using it right now. Be don't care about hidden files, the one we care about (project/workspace) aren't hidden.

mcornella:
Ok, this must be a particularity of OSX then, in Linux **all** files and folders beginning with a dot are hidden. Ok then, I can confirm that using this fix the `.xcodeproj` file will be selected instead of the `.xconfig` file:
```
$ ls -A
.xcodeproj  .xconfig
$ ls -A | grep "\.xc" | sort -r | head -1
.xconfig
$ ls -A | grep -E "\.(xcworkspace|xcodeproj)" | sort -r | head -1
.xcodeproj
```
**and** it will also prevent other `.xc*` files from being selected instead.

That said, a native zsh solution is preferred and if we're gonna fix it, let's fix it right. We can substitute the whole thing for just: 
```zsh
print -nl .{xcworkspace,xcodeproj}(N) | head -1
```

mcornella:
UPDATE: we could do it better by removing the `print` and using zsh arrays, like in [this answer from stackoverflow](http://unix.stackexchange.com/a/26825). And while we're at it we will `local`ize the `xcode_proj` variable.

Here's a diff ready to [apply](http://git-scm.com/docs/git-apply):
```diff
diff --git a/plugins/xcode/xcode.plugin.zsh b/plugins/xcode/xcode.plugin.zsh
index 1d1205b..82fe0e9 100644
--- a/plugins/xcode/xcode.plugin.zsh
+++ b/plugins/xcode/xcode.plugin.zsh
@@ -1,12 +1,12 @@
 #xc function courtesy of http://gist.github.com/subdigital/5420709
 function xc {
-  xcode_proj=`ls | grep "\.xc" | sort -r | head -1`
-  if [[ `echo -n $xcode_proj | wc -m` == 0 ]]
-  then
+  local xcode_proj
+  xcode_proj=(.{xcworkspace,xcodeproj}(N))
+  if [[ ${#xcode_proj} -eq 0 ]]; then
     echo "No xcworkspace/xcodeproj file found in the current directory."
   else
-    echo "Found $xcode_proj"
-    open "$xcode_proj"
+    echo "Found ${xcode_proj[1]}"
+    open "${xcode_proj[1]}"
   fi
 }
```

mcornella:
Also, you can go ahead and put `Fixes #3259` in your PR description to autoclose PR #3259.

tresni:
Shouldn't this be closed due to #4235 being merged?

mcornella:
Yes thank you.

PR #3959, opened 2015-06-06
  Add plugin for dnf (New package manager and package management library for Fedora) based in yum plugin

fjospinas:


octagonal:
Wouldn't it be better if they were aliased as `di` instead of `dnfi`?

mcornella:

PR #3961, opened 2015-06-07
Update aliases in laravel5.plugin.zsh file

scrubmx:
Some minor changes to the aliases.
* removed php artisan dump-autoload since is no longer a command in laravel 5
* updated the la5routes alias to the new version of the command
* added a la5vendor alias to publish assets from vendor packages


ncanceill:
Thanks!

PR #3962, opened 2015-06-07
Use local command properly in plugins/git/git.plugin.zsh

devunt:
zsh `local` command will print out pair or variable name and value if it has value. [1]

What supposed to happen is:
screenshot from 2015-06-07 16 49 57

What actually happend is:
screenshot from 2015-06-07 16 38 18

[1] http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -1,5 +1,4 @@
 # Query/use custom command for `git`.
-local git_cmd=
 zstyle -s ":vcs_info:git:*:-all-" "command" git_cmd
 : ${git_cmd:=git}
 
@@ -12,7 +11,7 @@ zstyle -s ":vcs_info:git:*:-all-" "command" git_cmd
 # Using '--quiet' with 'symbolic-ref' will not cause a fatal error (128) if
 # it's not a symbolic ref, but in a Git repo.
 function current_branch() {
-  local ref=
+  local ref
   ref=$($git_cmd symbolic-ref --quiet HEAD 2> /dev/null)
   local ret=$?
   if [[ $ret != 0 ]]; then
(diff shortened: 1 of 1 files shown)


devunt:
CC: @ncanceill

apjanke:
The real issue is that this code is using `local` outside a function call. When used inside a function call, `local <varname>` does not produce output. That `local git_cmd` that's sitting outside a function doesn't actually localize the variable: `zsh` variables are dynamically, not lexically scoped. You may as well just remove that `local`: the way plugins are loaded now, it's not doing anything, and if the plugin were loaded from inside a function call, localizing it would probably break things.

devunt:
I know that the bash `local` command can't be used from the outside of function, but is the zsh same?

apjanke:
Local variables in `zsh` work the same as local variables in `bash`. But the `local` command in `zsh` is slightly different: you *can* call it outside a function. It just probably won't do what you want. Basically, you should only use local variables in `zsh` the same places where you would use them in `bash`. Inside functions, or in code that's only going to get called from inside a function.

apjanke:
The presence of this `local` may be causing #3963. The error message sounds like it's related.

ncanceill:
@devunt would you be so kind as to fetch my fork and cherry-pick 6c29041 to see if it fixes your issue?

devunt:
@ncanceill it works like a charm.

ncanceill:
Awesome! this will be merged as soon as the other issues are solved.

devunt:
Then should I close this pr now?

ncanceill:
You can, but even if you omit to do so, this will close when #3990 gets merged.

PR #3966, opened 2015-06-08
Make spectrum_ls implementation consistent with $FG and spectrum_bls

### LINKED ISSUE #3964: Does spectrum_ls() work?
I've been looking at `lib/spectrum.zsh` since I've been working on prompt color stuff lately.

Does the `spectrum_ls()` function work? It says it displays all 256 colors, but when I run it, only the first 8 or so lines are colored, and the rest are in the default color. This happens on OS X 10.9, Windows 7, and Debian 7.

screen shot 2015-06-08 at 6 38 56 am

Here's the implementation.

```shell
# Show all 256 colors with color number
function spectrum_ls() {
  for code in {000..255}; do
    print -P -- "$code: %F{$code}$ZSH_SPECTRUM_TEXT%f"
  done
}
```

Instead of using the `${FG[n]}` control sequences it defined itself, it's using zsh's `%F...%f` prompt color controls. I don't think those support 256-color numeric codes; only the basic ANSI colors.

If I rewrite it using the control sequences instead of prompt color controls, then all 256 colors are shown as being colored.

```zsh
# Show all 256 colors with color number
function spectrum_ls2() {
  for code in {000..255}; do
    print -P "$code: ${FG[$code]}$ZSH_SPECTRUM_TEXT${FX[reset]}"
  done
}
```

screen shot 2015-06-08 at 6 46 04 am

Am I missing something here?

apjanke:
Fixes #3964 

The current `spectrum_ls` is using `%F/%f` prompt codes. These don't seem to support 256 colors. This changes it to use the control sequences it's defined in `$FG`.

Also adjusts the spectrum_bls implementation so the leading number is not included in the coloration, to ensure that it's always readable, even when the background color being displayed is close to the foreground text color.

### DIFF
--- a/lib/spectrum.zsh
+++ b/lib/spectrum.zsh
@@ -25,13 +25,13 @@ ZSH_SPECTRUM_TEXT=${ZSH_SPECTRUM_TEXT:-Arma virumque cano Troiae qui primus ab o
 # Show all 256 colors with color number
 function spectrum_ls() {
   for code in {000..255}; do
-    print -P -- "$code: %F{$code}$ZSH_SPECTRUM_TEXT%f"
+    print -P -- "$code: %{$FG[$code]%}$ZSH_SPECTRUM_TEXT%{$reset_color%}"
   done
 }
 
 # Show all 256 colors where the background is set to specific color
 function spectrum_bls() {
   for code in {000..255}; do
-    print -P -- "$BG[$code]$code: $ZSH_SPECTRUM_TEXT %{$reset_color%}"
+    print -P -- "$code: %{$BG[$code]%}$ZSH_SPECTRUM_TEXT%{$reset_color%}"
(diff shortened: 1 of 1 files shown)


apjanke:
Oops... looks like problem was just with my terminal configuration, not `spectrum_ls`. Closing without merging.

apjanke:
I'm reopening this PR, because I think this is still the right implementation. 

Even though I could get the current implementation to display all 256 colors by tweaking my `$TERM` setting, `spectrum_ls` is part of `spectrum`, and it should expose exactly what `$FG` contains, not what Zsh's `%F` supports, even though in many cases they should be the same thing.

This change gets `spectrum_ls` to match `$FG` exactly, and to be consistent with the behavior of `spectrum_bls`.

PR #3967, opened 2015-06-08
Allow the systemd plugin to work when sudo is not installed.

xaqq:
Problem: The systemd plugin requires sudo to be installed.
The plugin set-up aliases sc-\* for each systemctl command, and
prefix the alias with sudo so they can be used from non-root
account. However, this doesn't work if "sudo" is not installed.

Solution: Do not prefix the alias with "sudo" when they
are created for the root account.

This doesn't change anything for non-root user, but allows
root user to use the plugin even when sudo is not available
on the machine.

### DIFF
--- a/plugins/systemd/systemd.plugin.zsh
+++ b/plugins/systemd/systemd.plugin.zsh
@@ -8,4 +8,9 @@ sudo_commands=(
   link load cancel set-environment unset-environment)
 
 for c in $user_commands; do; alias sc-$c="systemctl $c"; done
-for c in $sudo_commands; do; alias sc-$c="sudo systemctl $c"; done
+# Do not use sudo for root user
+if [ "`id -u`" = "0" ]; then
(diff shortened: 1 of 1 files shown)


plugins/systemd/systemd.plugin.zsh
```diff
@@ -8,4 +8,9 @@ sudo_commands=(
   link load cancel set-environment unset-environment)
 
 for c in $user_commands; do; alias sc-$c="systemctl $c"; done
-for c in $sudo_commands; do; alias sc-$c="sudo systemctl $c"; done
+# Do not use sudo for root user
+if [ "`id -u`" = "0" ]; then
```
blueyed: You should use `$UID` here instead - avoids a subshell: `if [[ $UID == 0 ]]; then`

mcornella:
I don't see a demand for this, sudo is normally installed. Reopen if I'm wrong. In the meantime, you can use `alias sudo=''` to be able to use `sudo` commands anyway.

PR #3969, opened 2015-06-08
web-search.plugin.zsh: add stackoverflow searches for common languages

vitorgalvao:
Not sure if desirable, but could come in handy, and show off a bit more of the usefulness of the plugin.


mcornella:
Hi @vitorgalvao, thank you for that, I didn't know this functionality of duckduckgo. That said, I wouldn't put all those aliases in it. Maybe just define the one `stackoverflow` alias and put the rest as examples in a README for the web-search plugin (which is still pending to be written).

Also, keep in mind that the `frontend-search` already has a stackoverflow alias but it uses the stackoverflow website directly.

Thanks again

vitorgalvao:
That sounds more like the appropriate place.

PR #3977, opened 2015-06-09
Git plugin conflict with gdc (GNU D Compiler)

CorvinM:
Added check for gdc compiler before adding the conflicting gdc alias in the git plugin


plugins/git/git.plugin.zsh
```diff
@@ -80,7 +80,12 @@ alias gcp='git cherry-pick'
 alias gcs='git commit -S'
 
 alias gd='git diff'
+
+#Dont add alias if gdc (GCC D Compiler) is present
+command -v gdc >/dev/null 2>&1 || {
```
blueyed: You can use zsh's `$commands` array, e.g. `if (( $+commands[gdc] )); then …`, or `if [[ -n "${commands[gdc]}" ]]`.

That should be provided through a `safe_alias` function though (see https://github.com/robbyrussell/oh-my-zsh/issues/3682#issuecomment-79344414).

blueyed:
I think this PR can be closed in favour of #3990.

ncanceill:
Wait, we yet have to agree on a replacement alias and include it in #3990.

ncanceill:
But yes, this will be closed by #3990 anyway.

juanpastas:
I don't understand, this means that if I have if gdc is not showing git diff it should show some thing about gcc d compiler?

I don't have that compiler, but still I need to use gdca, is that correct?

blueyed:
@juanpastas 
The idea is to not setup a `gdc` alias if there is a `gdc` program.

(your question is not really clear to me however)

juanpastas:
What I meant is that gdc says command not found, while instead of that it
should output cached diff.

But this does not happen because it was just changed to gdca, but not
actually detecting if gdc overlaps other command.

blueyed:
Yes, it was changed for now, but might come back.
See https://github.com/robbyrussell/oh-my-zsh/pull/3990#issuecomment-111175759 and following comments.

juanpastas:
Thanks.

PR #3980, opened 2015-06-10
Added alias gaa = git add --all

eysteinbye:
Added alias gaa
``` gaa = git add --all ```
It has been here before I think

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -49,7 +49,7 @@ function work_in_progress() {
 alias g='git'
 
 alias ga='git add'
-
+alias gaa='git add --all'
 alias gb='git branch'
 alias gba='git branch -a'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jthagesk** [comment]
+1

**Comment by jthagesk** [comment]
A lot of other previous git commands is also missing, eg. ggpull, ggpush.

File: plugins/git/git.plugin.zsh [hunk 1]
```diff
@@ -49,7 +49,7 @@ function work_in_progress() {
 alias g='git'
 
 alias ga='git add'
-
+alias gaa='git add --all'
```
blueyed: The newline was probably meant to separate `git add ..` from `git branch ..`.
But there's no need to update the PR probably, as it will be added in the hotfix-PR probably.

ncanceill: I could not have put it better.

Issue #3893, opened 2015-05-14
virtualenvwrapper plugin fails on ubuntu with EQUALS option set


stuffy-the-dragon:
Refering to code from #2819.
On my system I got: <pre>"/etc/bash_completion.d/virtualenvwrapper:5: = not found" </pre> when the virtualenvwrapper plugin was loaded.

The problem is that /etc/bash_completion.d/virtualenvwrapper is bash script that is run by zsh with the EQUALS option set (default I think). 
When zsh hits the <pre>if [ "$USE_FULL" == "yes" ]; then</pre> line in /etc/bash_completion.d/virtualenvwrapper it tries to do '=' expansion as described in 14.7.3 [here](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Filename-Expansion)

I managed to fix the problem by adding <pre>unsetopt EQUALS</pre> to the top of the virtualenvwrapper.plugin.zsh file but there is probably a better way to do this.

ncanceill:
CC: @mispy

PR #3981, opened 2015-06-10
use local options to set no_equals before sourcing bash completion sc…

twang817:
use local options to set no_equals before sourcing bash completion script

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -1,10 +1,18 @@
 virtualenvwrapper='virtualenvwrapper.sh'
 
 if (( $+commands[$virtualenvwrapper] )); then
-  source ${${virtualenvwrapper}:c}
+  function {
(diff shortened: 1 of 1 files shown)


mcornella:
Can you elaborate on what this accomplishes? In my system sourcing the virtualenvwrapper.sh script already works well.

ncanceill:
As I understand it, @twang817 wants `setopt equals` to be specifically unset (in case it was set) while the virtualenvwrapper script is sourced. However this simply disables `=` expansion, which prevents words starting with `=` to be expanded to the full path of the corresponding command.

However I fail to see where a word starting with `=` could happen here.

twang817:
Sorry, this is to address this issue:

https://github.com/robbyrussell/oh-my-zsh/issues/3893

ncanceill:
OK thanks! You should have referenced that in your PR description — as mentioned in the [project of contribution guidelines](https://github.com/mcornella/oh-my-zsh/blob/add-contributing-file/CONTRIBUTING.md#getting-started).

PR #3983, opened 2015-06-10
Fix #3982

### LINKED ISSUE #3982: 'gap' alias removed (accidentely) from git plugin
the commit 5b75cc7880de1931a0507800d90eb77ed659fa46 removes `gap` alias on line 77

groolot:


doomspork:
@groolot it doesn't look like `gap` will be fixed, this is discussed in the PR comments: https://github.com/robbyrussell/oh-my-zsh/pull/2790#issuecomment-110689393.  You'll likely need to define that alias on your own.

PR #3878, opened 2015-05-10
Reflecting changes in artisan commands

tibdex:

PR #3984, opened 2015-06-10
correct la5routes alias

thomasRag:
the command "php artisan routes" as been deleted in laravel 5. "php artisan route:list" is now doing the same job.


ncanceill:

Please be so kind as to close this PR since it is a duplicate of #3878.

PR #3986, opened 2015-06-10
Only load url-quote-magic if it is available.

### LINKED ISSUE #3614: Consider spring cleanup of non-essentials from lib/
Currently, there is a large number of non-essential scripts in the default lib/ directory (e.g. bzr.zsh). While I understand that most of these only define functions, I see simply no reason to load them. Please consider making these into plugins and keep lib/ folder clutter-free.

At this point, it is necessary to keep an empty file of the same name in custom/ for each script provided  in lib/ so they are not loaded automatically. I am running oh-my-zsh on multiple machines and some do not support to full feature set used (e.g. url-quote-magic) making this use case quite inconvenient.

evilhamsterman:
Partially fixes #3614 by checking to see if url-quote-magic exists before loading it.

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -1,6 +1,10 @@
-## smart urls
-autoload -U url-quote-magic
-zle -N self-insert url-quote-magic
+## Load smart urls if available
+for d in $fpath; do
+	if [[ -e "$d/url-quote-magic"]]; then
+		autoload -U url-quote-magic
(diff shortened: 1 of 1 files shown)


mbologna:
+1, works with 2fb3265 @webmaster128

takano32:
Sorry for dup.
I closed my PR.
https://github.com/robbyrussell/oh-my-zsh/pull/4038

PR #3756, opened 2015-04-07
add git bisect aliases to git plugin

dgirgenti:
fixes #3751

PR #3990, opened 2015-06-11
[HotFix] Git plugin

### LINKED ISSUE #3751: add `git bisect` aliases

### LINKED ISSUE #3972: What happened with ggpush, gap and others?
I can see [here](https://github.com/robbyrussell/oh-my-zsh/commit/5b75cc7880de1931a0507800d90eb77ed659fa46#diff-4ad2773a26526d0e361e66139c283099L145) a lot of aliases been removed... this breaks my workflow and I guess others too, could you explain the rationale behind those removals?

### LINKED ISSUE #3991: Strange `ggpnp` behaviour
As discussed with @ncanceill [in the comments](https://github.com/robbyrussell/oh-my-zsh/commit/5b75cc7880de1931a0507800d90eb77ed659fa46#commitcomment-11623364), `ggpnp` has a weird behaviour when being launched for the first time in a zsh session on a repository.
The first part runs (`ggl`) runs well but the second part (`ggp`) fails with a "fatal: remote part of refspec is not a valid name in" error. If I run `ggl` or `ggp` alone, then I run `ggpnp` - everything works fine.

See the comment thread in the code for more details and attempts to debug the issue.

ncanceill:
This is a hot fix for the last Git plugin update #2790.

**STATUS: READY for MERGE**

***

**TODO:**
* [x] Fix the `local` call — fixes #3962
* [x] Bring back the old `gg*` aliases — fixes #3972
* [x] Recreate aliases for `git add` — see #3535 and #3682 and #3972, fixes #3980
* [x] Change the `gdc` alias — fixes #3977
* [x] Fix the error messages in the `gwip` alias — see [this comment](https://github.com/robbyrussell/oh-my-zsh/pull/2790#issuecomment-110963824)
* [x] Investigate the `current_branch` function — see [this comment](https://github.com/robbyrussell/oh-my-zsh/pull/2790#issuecomment-111154245)
* [x] Fix the `gg*` functions to accept additional arguments — see [this comment](https://github.com/robbyrussell/oh-my-zsh/issues/3972#issuecomment-110882984)
* [x] Fix the `ggpnp` function — fixes #3991

***

Closing forgotten issues: fixes #3751, fixes #3756

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -104,11 +104,15 @@ ggl() {
 git pull origin "${b:=$1}"
 }
 compdef _git ggl=git-checkout
+alias ggpull='ggl'
+compdef _git ggpull=git-checkout
 ggp() {
 [[ "$#" != 1 ]] && b="$(current_branch)"
 git push origin "${b:=$1}"
 }
 compdef _git ggp=git-checkout
+alias ggpush='ggp'
+compdef _git ggpush=git-checkout
 ggpnp() {
 ggl "$1" && ggp "$1"
 }
@@ -119,6 +123,8 @@ ggu() {
 git pull --rebase origin "${b:=$1}"
 }
 compdef _git ggu=git-checkout
+alias ggpur='ggu'
+compdef _git ggpur=git-checkout
 
 alias gignore='git update-index --assume-unchanged'
 alias gignored='git ls-files -v | grep "^[[:lower:]]"'
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -204,4 +204,4 @@ alias gupv='git pull --rebase -v'
 alias gvt='git verify-tag'
 
 alias gwch='git whatchanged -p --abbrev-commit --pretty=medium'
-alias gwip='git add -A; git rm $(git ls-files --deleted); git commit -m "--wip--"'
+alias gwip='git add -A; git rm $(git ls-files --deleted) 2> /dev/null; git commit -m "--wip--"'
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -94,19 +94,19 @@ alias gfo='git fetch origin'
 alias gg='git gui citool'
 alias gga='git gui citool --amend'
 ggf() {
-[[ "$#" != 1 ]] && b="$(current_branch)"
+[[ "$#" != 1 ]] && local b="$(current_branch)"
 git push --force origin "${b:=$1}"
 }
 compdef _git ggf=git-checkout
 ggl() {
-[[ "$#" != 1 ]] && b="$(current_branch)"
+[[ "$#" != 1 ]] && local b="$(current_branch)"
 git pull origin "${b:=$1}"
 }
 compdef _git ggl=git-checkout
 alias ggpull='ggl'
 compdef _git ggpull=git-checkout
 ggp() {
-[[ "$#" != 1 ]] && b="$(current_branch)"
+[[ "$#" != 1 ]] && local b="$(current_branch)"
 git push origin "${b:=$1}"
 }
 compdef _git ggp=git-checkout
@@ -118,7 +118,7 @@ ggl "$1" && ggp "$1"
 compdef _git ggpnp=git-checkout
 alias ggsup='git branch --set-upstream-to=origin/$(current_branch)'
 ggu() {
-[[ "$#" != 1 ]] && b="$(current_branch)"
+[[ "$#" != 1 ]] && local b="$(current_branch)"
 git pull --rebase origin "${b:=$1}"
 }
 compdef _git ggu=git-checkout
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -99,21 +99,29 @@ git push --force origin "${b:=$1}"
 }
 compdef _git ggf=git-checkout
 ggl() {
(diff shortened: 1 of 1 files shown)


apjanke:
I don't think there's a way to make that top-level `git_cmd` work as a local variable. It's not the plugin loading system's fault, it's a `zsh` limitation: Zsh only supports dynamic scoping, not lexical scoping, and doesn't have namespaces or packages. (Same for other shells.) You want that `$git_cmd` to be shared between independent functions and persist between calls (and, it looks like, make it visible to the Zsh completion system). That needs a regular global variable. 

If the plugin loading system changed such that the top-level `local` really was local because the plugin was loaded inside a function, then it would break, because it would only be local to the initial run of the plugin definition file itself. Subsequent calls to the functions defined in the plugin would not see the "local" value.

A global `$git_cmd` isn't just the easy fix, it's the only fix AFAIK.

Maybe just rename it to `$_omz_git_git_cmd` to make it clear where it's coming from and avoid collision with variable names that the user or other scripts might use; `$git_cmd` is pretty generic and I could see other things picking that. 

You could probably do something like declare one big global `$_OMZ_VARS` associative array for various OMZ-defined variables, and have all the functions that want to use it temporarily pop it in and out of truly `local` variables inside their functions. You'd still be using global vars, but you'd reduce the number of names it was taking. *However*, since they're still global names, you could still collide with those other variable names inside the OMZ function calls, e.g. if you're using callbacks, or calling other commands that use the collided variables. So the code would be a lot uglier, and still brittle in subtle ways. I'm not a fan of this hack.

ncanceill:
Thanks a bunch @apjanke! I will rename the variable and remove the `local` keyword, until we figure out a variable handling system.

ncanceill:
@itsNikolay @tejasbubane @Newman101 @MarcoSero @opsidao @tehmaspc @ChrisJefferson @blueyed @Stibbons

Do we agree on `gaa` for `git add --all`?

What alias would you suggest to replace `gap` for `git add --patch`?

What alias would you suggest to replace `gdc` for `git diff --cached`?

MarcoSero:
@ncanceill `gaa` for `git add --all` is good for me :+1:

blueyed:
> Until a safe aliasing system is written, easy fix is to use something else than gap.

I would introduce the `safe_alias` method for this:

    omz_safe_alias() {
        # With `safe_alias foo='bar baz'`, define the alias only if it is not a
        # command, function, or alias already.
        local alias lhs rhs
        alias=${(s:=:)1}
        lhs=$alias[1]
        rhs=$alias[2,-1]
        if (( $+commands[$lhs] || $+functions[$lhs] || $+aliases[$lhs] )); then
            return 1
        fi
        alias $lhs="$rhs"
    }

Otherwise the next complains would be about this alias being removed / not added again.

ncanceill:
@blueyed looks fine, can you make a PR? also we will need to echo a warning to advise the user that the alias was not created, and maybe a setting in `.zshrc` to change the behavior to instead erase the command with the alias and print an appropriate warning.

blueyed:
@ncanceill 
> looks fine, can you make a PR? 

I'd say to include/use it in the git plugin first (with this PR), and then move it to a central place later (with the settings etc).

I've changed the `return 1` to `return 73` (according to Python's `os.EX_CANTCREAT`), to make it different than a failure from the `alias` itself (if you try `alias foo=bar baz`).

ncanceill:
If we want to have this used throughout OMZ entirely, why would we first implement it in the Git plugin? I think this belongs either in [`oh-my-zsh.sh`](https://github.com/robbyrussell/oh-my-zsh/blob/master/oh-my-zsh.sh) or in a [`lib`](https://github.com/robbyrussell/oh-my-zsh/tree/master/lib) file.

Nevertheless, I agree on testing it in the Git plugin first.

ncanceill:
OK all problems should now be fixed.

Last thing to do is pick aliases for `git add --patch` and `git diff --cached`. Please send suggestions!

I want to get this PR out fast, so without any suggestion by tomorrow I will have to pick the aliases myself.

blueyed:
> Nevertheless, I agree on testing it in the Git plugin first.

Then you could use it for `gap` and `gdc`, can't you?

nikklassen:
+1 for gdca

ncanceill:
OK that does it, time is up for hot fix suggestions.

@blueyed I think pushing a new, non-tested feature in a hot fix would be irresponsible, I will be working on it in a new PR.

***

**New/changed aliases:**
* `gaa = git add --all`
* `gapa = git add --patch`
* `gdca = git diff --cached`

***

@robbyrussell please pull this hot fix in ASAP

iammichiel:
Can't wait until this gets merged.

robbyrussell:
Thanks for jumping on this, folks!

Stibbons:
+1 (late review)

ncanceill:
Thanks @robbyrussell, and sorry to have caused this mess in the first place ;-)

oshybystyi:
@ncanceill no one knows future - so don't judge yourself :+1:

ChrisJefferson:
Speaking as the original complainer about the 'gap' alias, having 3 letter aliases is (a) useful for most people most of the time, but was (b) always going to clash with something, some day :) I'm glad things are now improved :)

PR #3993, opened 2015-06-12
[theme] Fix color problem of steeef theme

honnix:
steeef theme sets colors depending on $TERM, however if %F is
used, it must be closed by %f instead of %{$reset_color%}.

### DIFF
--- a/themes/steeef.zsh-theme
+++ b/themes/steeef.zsh-theme
@@ -26,12 +26,14 @@ if [[ $TERM = *256color* || $TERM = *rxvt* ]]; then
     purple="%F{135}"
     hotpink="%F{161}"
     limegreen="%F{118}"
+    PR_RST="%f"
 else
     turquoise="$fg[cyan]"
     orange="$fg[yellow]"
     purple="$fg[magenta]"
     hotpink="$fg[red]"
     limegreen="$fg[green]"
+    PR_RST="%{${reset_color}%}"
 fi
 
 # enable VCS systems you use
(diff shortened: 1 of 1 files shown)


ncanceill:
cc @steeef

apjanke:
You could simplify the code by just using `%F`/`%f` everywhere.

```
    turquoise="%F{cyan}"
    orange="%F{yellow}"
    purple="%F{magenta}"
...

PROMPT=$'
purple%n%f at $orange%m%f in $limegreen%~%f $vcs_info_msg_0_$(virtualenv_info)%f
$ '

honnix:
I think the initiative was to support different TERM. That's why I'm keeping the code unchanged. 



Best regards,
Hongxin

apjanke:
The `%F` codes support various terminals; they look up their escape sequences in `terminfo` so they automatically use the `$TERM` you have set. They're just like `$fg`, except you don't have to put extra `%{...%}` around them.

honnix:
This may not be the case if `$TERM` is set to for example `xterm` and that's why a check against 256-color terminal is done there. If the terminal does support 256-color, this theme will offer more accurate colors, otherwise it's approximation.

apjanke:
I'm not sure I follow. A non-256color `xterm` still supports the basic ANSI colors, so the basic `%F{cyan}`, `%F{yellow}`, and so on will still work for it, as long as you don't use the high color values for them. I'm not suggesting that you remove the different branches for low color and 256-color, just that the `%F` syntax be used for the low-color case as well.

(And that test for 256-color support could be done more portably by testing `$terminfo[colors]` instead of doing pattern matching on the `$TERM` name.)

honnix:
WelI I think the author intended to use high color value when possible. I agree to use %F for lower value as well but that really goes beyond this pull request. I think I will close it for now because it meant to fix a bug not refactoring. Thanks a lot for the discussion. 



Best regards,
Hongxin

apjanke:
I think it would be in scope for this PR: the problem this PR is fixing is the mismatch between the `%{${reset_color}%}` in the `$PROMPT` string and the `%F` syntax used in the 256-color case, right? This PR fixes the mismatch by conditionalizing `${PR_RST}` to be either `%{${reset_color}%}` or `%f`. But you could fix the mismatch by going the other way, and making both high and low color cases use `%F` uniformly.

honnix:
Sure. I'll go for the other way.

honnix:
a new PR submitted #4155

PR #3994, opened 2015-06-12
Magento CLI Plugin

rafaelstz:
The best alias and tools to easy development Magento


robbyrussell:
There's a bunch of hard coded paths in this, so I don't know this will work well for others.

rafaelstz:
The base path is defined as follows:

```bash 
# Core
 base="~/.magento"
```

The installation commands are executed in the defined folder. These commands only create a folder to store and copy the application to ```/usr/local/bin/```, this folder is used to store applications on OSX and Ubuntu.
I use the original install code of applications with complements for my folder and test in two system.

https://github.com/netz98/n98-magerun
https://github.com/jreinke/modgit

PR #3997, opened 2015-06-13
Add pyenv-virtualenv support to pyenv plugin.

razor-x:


Kriechi:
@mcornella I believe this PR can be closed because it is superseeded by #6165, which already contains similar code.

PR #4000, opened 2015-06-14
Replace `.` aliases with as-you-type expansion

dimitriye98:
Replace aliases:
- `...`
- `....`
- `.....`
- `......`

with as-you-type auto-expansion.

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -3,10 +3,16 @@ setopt auto_pushd
 setopt pushd_ignore_dups
 setopt pushdminus
 
-alias -g ...='../..'
-alias -g ....='../../..'
-alias -g .....='../../../..'
-alias -g ......='../../../../..'
+resolve-dot() {
+	if [[ $LBUFFER = *.. ]]; then
+		LBUFFER+=/..
+	else
+		LBUFFER+=.
+	fi
+}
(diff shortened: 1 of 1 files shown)


robbyrussell:
Out of curiosity, how much quicker/slower is this?

dimitriye98:
Latency-wise, there doesn't seem to be any. It's definitely easier to grok how many levels you'll be jumping up, but admittedly, the one downside is that you need to delete 3 characters to undo a level rather than 1. I personally prefer it to aliases, since it gives more visual feedback.

dimitriye98:
If you want I can pull request it as a plugin instead of a mod to the core.

mcornella:
Hi @dimitriye98, this is cool but it doesn't take into account the type of dots being typed: if it's a route it's ok, but everything else should be expanded. So for instance if I'm typing `echo to be continued...` it will be expanded into `echo to be continued../..` which is not what I'd expect at all. 

The good news is that I have an alternative proposal: use the logic in [this blog post](http://blog.patshead.com/2011/07/automatically-expanding-zsh-global-aliases-as-you-type.html) and adapt it to only expand `...` aliases.

robbyrussell:
We'll close this given how long it's been pending more discussion for quite some time.

PR #4001, opened 2015-06-14
Fixed odd chars on mac for agnoster theme

ivanfoo:
Quick fix to solve odd separator char when using agnoster theme on Mac

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -26,7 +26,13 @@
 # A few utility functions to make it easy and re-usable to draw segmented prompts
 
 CURRENT_BG='NONE'
-SEGMENT_SEPARATOR=''
+
+# Fix odd char on mac
+if [[ `uname` == 'Darwin' ]]; then
+    SEGMENT_SEPARATOR='\ue0b0'
+else
+    SEGMENT_SEPARATOR=''
+fi
 
 # Begin a segment
 # Takes two arguments, background and foreground. Both can be omitted,


ncanceill:
Was this really necessary?

I had Agnoster's theme working seamlessly on Mac before this change. Did you have a properly-patched powerline font set up?

ivanfoo:
Hi,

Yes, I have a properly-patched powerline font set up and I couldn't get it
working. In fact, there is an issue about it here:
https://github.com/robbyrussell/oh-my-zsh/issues/1906


2015-06-15 8:56 GMT+02:00 Nicolas Canceill <notifications@github.com>:

ncanceill:
OK apparently this was reported in [issue 2869](https://github.com/robbyrussell/oh-my-zsh/issues/2869#issuecomment-108913226). Anyway this should not break anything.

themes/agnoster.zsh-theme
```diff
@@ -26,7 +26,13 @@
 # A few utility functions to make it easy and re-usable to draw segmented prompts
 
 CURRENT_BG='NONE'
-SEGMENT_SEPARATOR=''
+
+# Fix odd char on mac
+if [[ `uname` == 'Darwin' ]]; then
+    SEGMENT_SEPARATOR='\ue0b0'
```
ujjwalt: We want ⮀and not \ue0b0 right? It doesn't work on my machine with a power line font.

ivanfoo:
We want \ue0b0 on Mac boxes and ⮀ on the rest...I have a Linux box too and the fix is working

ujjwalt:
The \ue0b0 does not work on my Mac. I have the Menlo power line font. What do you suggest?

themes/agnoster.zsh-theme
apjanke: How old is your Powerline font? `\ue0b0` is the segment separator that Powerline currently uses. They changed the code points they were using a while back, and `\ue0b0` is the new one. (I can't tell which other character you're referring to, because my browser isn't using a powerline-patched font. And since there are two versions of the font, pasting the characters directly as single Unicode characters leads to ambiguity, because their appearance is going to depend on which versions of the font people in the conversation have installed.)

apjanke:
I suspect it should actually be \ue0b0 everywhere, and this issue is caused by outdated Powerline fonts being installed (see #3713). AFAICT, Powerline doesn't use different Unicode characters on different OSes (which would be weird). But they did change the code points they were using back in 2012, so some machines may have old versions of the fonts installed.

@ivanfoo: When you say "and ⮀ on the rest", I can't tell what character that is, because I don't use Powerline-patched fonts in my browser. What character (Unicode code point) are you referring to? What version (date or SHA) of the patched font do you have installed on your Linux box? What's the behavior if you `ssh` from a terminal on your Mac to your Linux box and run `zsh` with the `agnoster` theme there, or vice versa?

@ujjwalt: What exactly does "not work" mean? What are you seeing? Try re-downloading the Menlo Powerline font from https://github.com/powerline/fonts. If \ue0b0 does not work for you, it probably means you have an obsolete version (with old code points) of the font installed.

To test which version of the font you're using (in your *local* terminal emulator), you can do this.

```
printf "\ue0b0 \ue0a0 \u2b80 \u00b1 \u27a6 \u2718 \u26a1 \u2699 \n"
``` 

If you see a triangle "segment separator" followed by a branch symbol, you are using a current powerline-patched font. (I *think*.) If you see something else, you're looking at an old, non-patched, or otherwise bad font.

I've put in #4091 to clarify this and make it consistent across platforms.

ujjwalt:
powerline fail
This is what I see

ujjwalt:
So it does work if I use your font.

apjanke:
That means your "Menlo for Powerline" font was an obsolete one with the old code points.

I'd recommend switching to one of the "Meslo LG X for Powerline" variants. (The "X" just indicates different line spacing.) Meslo is a minor variant on Menlo with a few tweaks for on-screen use. You probably won't even notice the difference between them.

ujjwalt:
Works perfectly now. You may close this issue.

apjanke:
It's already closed. :) ("Merged" counts as closed.)

PR #4002, opened 2015-06-15
Add missing space causing parse error.

aliafshar:

PR #4005, opened 2015-06-15
lib/misc.zsh:3: parse error near `then'

kraiz:
Fixes: ~.oh-my-zsh/lib/misc.zsh:3: parse error near `then'


kraiz:
Duplciate of #4002

PR #4007, opened 2015-06-15
fixed lib misc

m42e:
Fixed lib/misc.zsh to work on cygwin without errors.


ncanceill:
Thanks but this is already covered by #4002 which will be pulled ASAP, so please close this one.

m42e:
Already fixed

PR #4010, opened 2015-06-15
Fix error "oh-my-zsh/lib/misc.zsh:3: parse error near `then'"

alessandro-pezzato:
Fix the missing space in `"]]; then` (syntax error)


ncanceill:
Thanks this is already fixed in #4002 and multiple other PRs.

jankeesvw:
Got the same issue, even after upgrading to the latest version.

ncanceill:
... and now #4002 has been merge, can you please close this?

PR #4012, opened 2015-06-15
Fix missing space in misc.zsh

Taytay:
Without this fix, I am seeing the error:
`parse error near `then'`


Taytay:
Nevermind - others beat me to it :)

PR #4014, opened 2015-06-15
Added __git_ps1() from official git repository

mbologna:
From git repository, I added the file:
  https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh
  which offer the __git_ps1() function to customize the prompt.

  Basically, it keeps track of what changes are
  made in the current git repo (if available) and
  returns some symbols to put in your prompt (e.g. %
  if there are untracked files).

* In the original version, flags (e.g. SHOWDIRTYSTATE)
  must be enabled to return symbols described above.
  Changed to enable these features by default (removed if)

* Returned symbols can be customized using OMZ default
  variables (e.g. ZSH_THEME_GIT_PROMPT_STASHED) defined in
  your theme. I modified __git_ps1() in order to use the symbols you specified in those variables      instead of git ones'


apjanke:
I don't think we can use this in OMZ, since this new code [is GPL](https://github.com/git/git/blob/v2.4.3/contrib/completion/git-prompt.sh#L3-L4), and Oh My Zsh is [MIT licensed](https://github.com/robbyrussell/oh-my-zsh/blob/master/MIT-LICENSE.txt). These two licenses are not compatible.

ncanceill:
Indeed they are not.

@mbologna can you work on a workaround?

mbologna:
Closing this, as #4050 solves the problem.

mbologna:
I would like to point out that the same file from git repo is already included in omz:
https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/gitfast/git-prompt.sh

apjanke:
Looks like most of the files in the gitfast plugin are GPL. Their provenance should be checked and the plugin may need to be revised or removed.

PageVPetty:
Mine kept saying, " -bash: __git_ps1: command not found  " until I realized that the file " git-prompt.sh " was misspelled. It said " git-promt.sh ". As soon as I changed that everything worked fine.

PR #4013, opened 2015-06-15
fixes a parse error in misc.zsh

tobsan:

PR #4015, opened 2015-06-15
FIX a parse error with zsh 5.0.2

naholyr:
The error: ``…/.oh-my-zsh/lib/misc.zsh:3: parse error near `then'``

```
$ zsh --version

zsh 5.0.2 (x86_64-pc-linux-gnu)
```

The added space fixes the parse error.


naholyr:
Duplicates #4013

PR #4016, opened 2015-06-15
Fix typo & Improve README in coffee plugin.

coopermaruyama:
* `cfp` alias was referencing an old version of the `cf` command which
no longer exists. This has been updated to reference a valid command.

* Fixed typo in README and made the example more readable.


mcornella:
Thank you for your contribution!

Issue #4018, opened 2015-06-15
 parse error near `then' after update


Javix:
Just updated to the latest version by taping in 'y' in my terminal and got the below error:
```
ast-forwarded master to 8921902388bf357ad0ab6ec198f4ba1f14b94fe4.

/Users/XXXXX/.oh-my-zsh/lib/misc.zsh:3: parse error near `then'
```
It reappears when quitting and reopening the terminal.

OS: OS X 10.10.3

Regards

Kevin-Lee:
I got the same issue and the pull request I've just created fixes it.

Javix:
Perfect, waiting forward to get the merge.

Javix:
Seems like the same has already been reported: https://github.com/robbyrussell/oh-my-zsh/issues/4037

oshybystyi:
yeah, @Javix can you close this, #4002 is in the master already...

Javix:
Great, no problem.

PR #4019, opened 2015-06-15
fixed syntax

vinc3nt:

PR #4021, opened 2015-06-15
Fixes robbyrussell/oh-my-zsh#4018 - parse error near 'then' after update

Kevin-Lee:
This pull request fixes robbyrussell/oh-my-zsh#4018 - parse error near `then' after update which is caused by [lib/misc.zsh:3](/robbyrussell/oh-my-zsh/blob/422db48e37beadf31fd9081b7590f9498f7194b2/lib/misc.zsh#L3).

```shell
	if [[ -e "$d/url-quote-magic"]]; then
```
There must be a space char between `...magic"` and `]]; ...` so it must be
```shell
	if [[ -e "$d/url-quote-magic" ]]; then
```
I've tested with mine and it works fine after adding the space.

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -1,6 +1,6 @@
 ## Load smart urls if available
 for d in $fpath; do
-	if [[ -e "$d/url-quote-magic"]]; then
+	if [[ -e "$d/url-quote-magic" ]]; then
 		autoload -U url-quote-magic
 		zle -N self-insert url-quote-magic
 	fi


Kevin-Lee:
Oops! While I was working on this, someone else already created a pull request to fix it. :smile: So this pull request is duplicate of #4019.

Kevin-Lee:
This might be the first one? #4002 Sorry I didn't know it. :sweat_smile:

PR #3918, opened 2015-05-21
Detect the .venv in any parent directory of the current directory

cbazin:
This patch allow the autoactivation of virtualenv in any parent directory.

This is useful when the .venv directory is not at the root of the git repository.

In addition, it works even if the .venv is not in a git repository.

PR #4025, opened 2015-06-15
Fix for the fact that realpath is not installed on OS X by default (fixes #4030)

### LINKED ISSUE #4030: New dependency (realpath) breaks when switching to a directory present with a virtualenv
```bash
sudo apt-get install realpath
```

SandyRogers:
…so plugin would not work

Issue was introduced by pull request #3918.
This fixes #4030.

This is a bit of a work-around for now (just checks if realpath is
installed…) but at least the plugin works again :+1: 

Better would be to install it, e.g.
https://github.com/harto/realpath-osx

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -36,14 +36,25 @@ if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
   function workon_cwd {
     if [ ! $WORKON_CWD ]; then
       WORKON_CWD=1
-      # Check if this is a Git repo
-      PROJECT_ROOT=`pwd`
-      while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" ]]; do
-        PROJECT_ROOT=`realpath $PROJECT_ROOT/..`
-      done
-      if [[ "$PROJECT_ROOT" == "/" ]]; then
-        PROJECT_ROOT="."
+      
+      # Check whether realpath is installed (it isn't on a mac...)
+      if type "$realpath" > /dev/null; then
+        # If we have realpath, can auto switch virtualenv based on git name
+        # Check if this is a Git repo
+        PROJECT_ROOT=`pwd`
+        while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" ]]; do
+          PROJECT_ROOT=`realpath $PROJECT_ROOT/..`
+        done
+        if [[ "$PROJECT_ROOT" == "/" ]]; then
+          PROJECT_ROOT="."
+        fi
+      else
+        PROJECT_ROOT=`git rev-parse --show-toplevel 2> /dev/null`
+        if (( $? != 0 )); then
+          PROJECT_ROOT="."
+        fi
       fi
+      
       # Check for virtualenv name override
       if [[ -f "$PROJECT_ROOT/.venv" ]]; then
         ENV_NAME=`cat "$PROJECT_ROOT/.venv"`


ncanceill:
Looks good.

This fixes #4030, please include `fixes #4030` in the PR description text.

@robbyrussell this is a hot fix for the last upgrade, please pull it along with #4002

joeghodsi:
From #4030. Does this handle the reporter's issue:
`kill it and then install realpath. And now the error is gone, but it's not automatically activating the virtualenv`

agonzalezro:
Tested locally, works great.

Yes @joeghodsi it seems that #4030 is fixed as well.

Tested in a mac with and without `coreutils` package installed.

SandyRogers:
Yes I believe the fix here addresses problems mentioned in #4030 so should be good to go.

davidetruffo:
hi! 
is this available via "upgrade_oh_my_zsh" yet? :)

ncanceill:
Not yet, but @robbyrussell should be merging this soon.

MrHalfman:
Why aren't you usign `readlink -f` instead of `realpath`?
It seems to be a better solution :wink:.

SandyRogers:
@MrHalfman Cool, I will try readlink instead (this pull request was just meant as a super quick hotfix).

SandyRogers:
@MrHalfman OS X supplied `readlink` doesn't have `-f` option, unless user has installed `greadlink` via e.g. homebrew. There are workarounds I guess, but I'd suggest we merge this pull request for now to at least get things operational again for OS X users...

cbazin:
Given stackexchange and stackoverflow  "readlink -f" might be a good alternative for Linux systems but for osx the problem is still the same.

Maybe we could try something like this : 
PROJECT_ROOT=$(cd "$PROJECT_ROOT/.."; pwd)
Instead of :
PROJECT_ROOT=`realpath $PROJECT_ROOT/..`

It might work (except if the macosx users doesn't have pwd ;-)

Source : 
http://stackoverflow.com/questions/5756524/how-to-get-absolute-path-name-of-shell-script-on-macos

mail6543210:
What about only adding `cd` approach as a fallback of `realpath` approach?

As an Ubuntu user, since debian is adding `realpath` from `coreutils` in `v8.23-1`, it'll be available in debian-based distros soon.
(Currently, it's available for debian>=jessie and ubuntu>=utopic.)

But I'm not sure about the availability of `realpath` in other Linux distros.

Or maybe using `alias realpath=readlink [-f]` and `unalias realpath` as a fallback of `realpath`?

agonzalezro:
I don't fully understand the problem here.

If I understand the problem properly using `realpath` will "fix" a path from: `/Users/agonzalezro/this/is/something/../.venv` to `/Users/agonzalezro/this/is/something/.venv`.

Are those `..` so problematic? I would bother too much about getting the realpath of that path.

mail6543210:
Another minor question, why is the dollar sign in line 41?

Is realpath a variable in Mac?

On my ubuntu, `type "$realpath"` is always evaluated as false, which totally disables the effect of PR #3918 .

I guess it should be `type realpath`, and can be replaced by 
`if (( $+commands[realpath] )); then`
that is, using zsh's builtin variable to detect executable.

agonzalezro:
@mail6543210 `realpath` is part of the `coreutils` package. It's not installed by default.

mail6543210:
@agonzalezro but `readlink` is also in `coreutils`?

agonzalezro:
@mail6543210 it is. But as @SandyRogers mentioned before you will need to have install greadlink (the GNU version).

It's becoming a little bit messy since none of those are installed by default.

Would it be possible to just don't clean that path as I asked before? Of course, this is up to the creator of the PR & the maintainers, but it could be a "good" solution.

MrHalfman:
The original problem was that the venv plugin does not automatically "enables" venv if its directory is not situated in the root path of the project.
So, if I understand well your suggestion @agonzalezro the result will cancel PR #3918.

I think that the solution suggested by @cbazin is the good one. (Use cd for OSX)

agonzalezro:
@MrHalfman I don't want to change the loop so #3918 will still be working. But inside the loop instead `realpath`ing everything you will be appending `/..` to the final path.

cbazin:
@agonzalezro

If you don't use the absolute path of the file, you can't compare the path with "/" which is the exit condition of the loop in case there is no .venv in the superdirectories. 
 
Again, the following line should solve the problem for both osx and linux : 

PROJECT_ROOT=$(cd "$PROJECT_ROOT/.."; pwd) 

(The problem is that it calls "cd" and "pwd" many times.)

mail6543210:
@agonzalezro I've got your point. Thanks!

@cbazin The problem of using `cd` is that `cd` doesn't follow the symbolic links.
So a symlink in `DirA` pointing to a subdir `SubDirA` of a project `DirPrj` will not workon `DirPrj/.venv` but `DirA/.venv`.

cbazin:
@mail6543210 Isn't it odd (error prone) to allow this behaviour? 

Imagine you follow 1, 2, 3 or even more symlinks, it becomes very hard for the programmer to understand which virtualenv is currently active.

If you don't follow the symlink, the behaviour seems to be more straightforward.

Do you know how it worked before the patch?

mail6543210:
@cbazin Yes, it's odd. But if someone use this behavior, which venv might they expect to workon? `DirPrj`(`realpath`/`readlink`) or `DirA`(`cd`)?
Actually, we'll never know:) I guess they most likely expect the former, but maybe in some cases the later(?).

I just find out that `cd` can follow symlic too, by using the option `-P`.
Which becomes:
`PROJECT_ROOT=$(cd -P "$PROJECT_ROOT/.."; pwd)`

cbazin:
Ok. I still feel the behaviour is error prone but the backward compatibility is more important.
I checked with a prior version of oh-my-zsh and the git command does follow the symbolic links.

Thus It seems the solution proposed by @mail6543210 is the best so far.

    PROJECT_ROOT=$(cd -P "$PROJECT_ROOT/.."; pwd)

Maybe a global option could be a good idea (to follow symlink or not).

timmow:
zsh allows [parameter expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion) - would it work to replace

```
PROJECT_ROOT=`realpath $PROJECT_ROOT/..`
```

with

```
PROJECT_ROOT=${PROJECT_ROOT:A:h}
```

Where the A modifier "Turns a file name into an absolute path, resolving symlinks"
and the h modifer "Removes a trailing pathname component, leaving the head. This works like ‘dirname’."

This seems to work for me locally

apjanke:
@timmow: Yeah, that `:A:h` is standard Zsh, and should work portably with no external dependencies. That's how I'd do it. Works for me on OS X 10.9.

Here's a better link to the [Parameter Expansion Modifiers subsection](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers) that documents `:A` and `:h`.

oryband:
:+1: can we mergergerz this already plz k thx bai?

cbazin:
It seems to be the best answer so far. 

This solution is less explicit (readable) than the 'cd' solution but it uses the zsh syntax.

tklovett:
Please please please please please merge this or another fix as soon as possible. The first `cd` in every new terminal window infinite loops and its so frustrating :(

SandyRogers:
Sorry for silence on my part. I will try and implement one of these alternative suggestions this weekend, since my original solution seems out of favour.

apjanke:
Sorry for stepping on your PR, @SandyRogers, but it seems like you're busy at the moment, and this is breaking stuff out in the wild. I've put in a simpler PR #4106 using `zsh`-native `:A`, which does the same thing as `realpath` with no external dependencies. It supersedes this PR.

If folks could try out #4106, I'd appreciate that. It's a simple change and should be an easy merge.

SandyRogers:
Nice, thanks @apjanke! Yep I've been busy, and it was perfect weather this weekend cycling/being on the river...! Will try out your PR now.

mcornella:
@SandyRogers try #4074 instead, there's a slight difference in the algorithm of #4106.

apjanke:
I don't think there is, actually: #4074 is just making the same mistake as this #4025. A single call to `realpath` or `:A` resolves the symlinks at *all* levels of the file path, so the additional calls to it as you're walking up the path are unnecessary. The end result is the same, but with more I/O.

I've included a test case on #4074 that demonstrates this behavior.

PR #4026, opened 2015-06-15
Fix parse error in misc.zsh

dzen:


ncanceill:
Thanks for contributing, this is fixed in #4002 and multiple other new PRs.

dzen:
sorry :(

PR #4028, opened 2015-06-15
Update misc.zsh

dsdolzhenko:
On OS X 10.10.3 with zsh 5.0.7 (x86_64-apple-darwin14.0.0),  while opening new session the following error message occurs:

> ~/.oh-my-zsh/lib/misc.zsh:3: parse error near `then'

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -1,6 +1,6 @@
 ## Load smart urls if available
 for d in $fpath; do
-	if [[ -e "$d/url-quote-magic"]]; then
+	if [[ -e "$d/url-quote-magic" ]]; then
 		autoload -U url-quote-magic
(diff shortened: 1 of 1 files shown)


jirutka:
The same problem here.

mrloop:
Same problem on Ubuntu 15.04, commit resolves problem.

oshybystyi:
sorry, folks there are already a [lot of prs](https://github.com/robbyrussell/oh-my-zsh/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+misc) that fix same issue. #4002 is probably the first one.

apjanke:
#4002 has been merged. Are you still having this problem? If not, please close.

PR #4032, opened 2015-06-15
Add missing space causing error when ran

dummymael:
Added missing space to the if condition which caused an error when
the script was ran during an auto-update.


dummymael:
Lots of PRs already opened for this :D

PR #4038, opened 2015-06-15
lib/misc.zsh: fix parse error near `then' error

takano32:
fix parse error
$HOME/.oh-my-zsh/lib/misc.zsh:3: parse error near `then'`


ncanceill:
Thanks for contributing, however we already have a gazillion pending PRs for that.

takano32:
Sorry for my duplicate PR.

qstosh:
thanks!

PR #4040, opened 2015-06-15
Make fine-time theme work with chruby

esambo:
by using [prompt_info_functions](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/prompt_info_functions.zsh#L31)


apjanke:
@a-b, looks like you're the author/maintainer for this theme. Comments?

PR #4043, opened 2015-06-15
Adds ubuntu/linux support to atom plugin

bananenmannfrau:
Checks if the OS is linux and in that case starts atom from /usr/bin/atom. If not the previous way is used.


ncanceill:
cc @aforty

mcornella:
Superseeded by #4342.

robobenklein:
Atom plugin has been removed: https://github.com/robbyrussell/oh-my-zsh/commit/0516a39dec64568f6db194d143a0ae2969c3cf9b

PR #3289, opened 2014-11-04
Add docker useful alias

ColinHuang:
Docker is a tool that can be used to build applications, provide standardized environments for development, testing and production, isolate dependencies, and more. More and more people are using docker as their daily testing or develop. Add more alias help to save time to type commands.

PR #4054, opened 2015-06-17
Added useful docker aliases

kdubovikov:
Docker alias for deleting exited containers.


leifcr:
@kernelmode Can you look at #3289 and see if they can be merged to one PR?

Aliases should be in docker.plugin.zsh and not _docker, since _docker is pulled from upstream docker repo.

robbyrussell:
⌛️ Given how much time has passed, am going to close this as we needed to refactor where the alias is set.

Issue #4053, opened 2015-06-17
After upgrade, ggpull breaks local git repo


anneveling:
After upgrading oh-my-zsh to latest version, my daily routine of doing "ggpull" on a branch tracked remotely, resulted in an additional output line like this:

`
➜  project-42 git:(master) ggpull
From github.com:gust/project-42
 * branch            master     -> FETCH_HEAD
 * branch            HEAD      -> FETCH_HEAD
`
This happens even on a clean folder, just after a clone.

Because of this double fetch-head (or whatever it is called), simple pulls will then result in complex merge conflicts.

We tracked this to https://github.com/robbyrussell/oh-my-zsh/issues/3991
and commit https://github.com/robbyrussell/oh-my-zsh/commit/f513999a4dbfea3a15287368848a718543fcaf7b

Reverting oh-my-zsh to version 6ff96dab36865ab208b08a004903338bfd405eac does not have the issue.

It seems that the latest changes to the git plugin making the current-branch variable local has broken this. Took me some time to figure this out, will stay on older version for now but it would be great if you could fix this? Do you need more info to reproduce? Simple enough: just clone any github repo into a clean folder and do a "ggpull".

ncanceill:
Hi, and thanks for reporting.

This was already identified and I will send a fix today.

For your curiosity, the cause of this is a try to allow additional arguments to `ggpull` and the like; when called with `""` as an argument `git pull` will recognize it as an empty branch name, and default to the remote HEAD.

anneveling:
Ooh that was quick. Thanks. Looked at the issue list but didn't see the duplicate there. Will upgrade again to latest version when you say it's fixed. Keep up the good work. oh-my-zsh makes my life so much easier

PR #4056, opened 2015-06-17
Fix plugin/git ggpull and ggpush backward compatibility

### LINKED ISSUE #3995: Git function for current branch is not working
The following fails : 

    ggpush --force

The alias is using the last git commit hash as current branch name. 
This problem appeared after this change 5b75cc7880de1931a0507800d90eb77ed659fa46. 

Stacktrace :

    error: src refspec c3acafc does not match any.
    error: failed to push some refs to 'ssh://git@whateverrepo.git'

### LINKED ISSUE #4055: Old habit to `ggpush -f` is super dangerous now
A reasonable amount of people use `git push --force` on a daily basis, I do when I rebase an already pushed branch.

I am, and I think many people are (hello @Calyhre), used to do it using `ggpush -f` while assuming a controlled `git push origin $(current_branch)`.

Recent changes have made `ggpush -f` equivalent to `git push origin -f` which arguably is the worst git command to run.

Can you imagine how many git repos will be fucked up by people not knowing about this change? (hello again @Calyhre).

I am not arguing about how new aliases should work or how using `--force` is bad, I am arguing about how dangerous the current behaviour is compared to expectations.

ArnaudRinquin:
Current `ggpush` is `git push origin` instead of `git push origin $(current_branch)`
Same for `ggpull`

Closes #4055
Closes #3995
And probably many others.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -105,7 +105,7 @@ ggl() {
 git pull origin "${b:=$1}" "${*[2,-1]}"
 }
 compdef _git ggl=git-checkout
-alias ggpull='ggl'
+alias ggpull='git pull origin $(current_branch)'
 compdef _git ggpull=git-checkout
 ggp() {
 if [[ "$#" != 0 ]] && [[ "$#" != 1 ]]; then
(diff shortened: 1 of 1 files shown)


ArnaudRinquin:
Closes #4053

ncanceill:
Thanks, I was about to do it.

PR #4057, opened 2015-06-17
allow customization of colors for colored-man

twang817:
allows customization of colored in colored-man

### DIFF
--- a/plugins/colored-man/colored-man.plugin.zsh
+++ b/plugins/colored-man/colored-man.plugin.zsh
@@ -1,32 +1,50 @@
 if [ "$OSTYPE[0,7]" = "solaris" ]
 then
-	if [ ! -x ${HOME}/bin/nroff ]
-	then
-		mkdir -p ${HOME}/bin
-		cat > ${HOME}/bin/nroff <<EOF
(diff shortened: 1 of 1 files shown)


apjanke:
How about using terminfo sequences for portability and readability, like in https://github.com/apjanke/oh-my-zsh/tree/colored-man-terminfo?

```
  local context=:omz:plugins:colored-man
  zstyle $context mb '$fg[red]'          # begin "blink"
  zstyle $context md '$fg[red]'          # begin "bold"
  zstyle $context so '$fg_bold[yellow]$bg[blue]'  # begin standout
  zstyle $context se '$reset_color'      # end standout
  zstyle $context us '$fg_bold[green]$termcap[us]'   # start "underline"
  zstyle $context ue '$reset_color'      # end "underline"
  zstyle $context me '$reset_color'      # end all modes
```

That also adds a delayed evaluation to the style definitions, to keep the output of `zstyle` readable. Otherwise, literal terminal control sequences are emitted as part of its output, which ends up coloring the output but being pretty hard to read.

screen shot 2015-09-28 at 10 55 56 am

vs:

screen shot 2015-09-28 at 10 56 13 am

PR #4061, opened 2015-06-18
archlinux: Remove yasu / yaourt --sucre

aeons:
The option has been removed from yaourt:  archlinuxfr/yaourt@0cedad581a35e66b670853dc3be3077258d8eab7


hosaka:
Wouldn't it be better to add the --noconfirm option instead? It does exactly what you would expect.

mcornella:
I agree with @hosaka, better make it `yaourt -Syua --no-confirm` instead of removing it. Thanks

mcornella:
Fixed in 3b8d0ad.

PR #4066, opened 2015-06-19
Symfony2 plugin: Add doctrine:schema:update aliases

AntoineLemaire:
Just basic aliases very useful !


robbyrussell:
Closing as there are similar aliases in the plugin, already

PR #4070, opened 2015-06-19
[Hotfix] Search for .git rather than .venv inside loop.

oskargicast:
Fix automatic activate for virtualenvs once a .git file is found. As the comment said "Check if this is a Git repo" and not .venv


apjanke:
Attempt to fix #4110.

This change is incorrect; it's too simple. That checking for the `.venv` directory was introduced by #3918, and was an intentional change.

We do need to add back support for treating git repos as virtualenvs, but it shouldn't be done by just changing to only use git repos. This simple change also would break some of the `.venv` customization logic.

apjanke:
I've put in  #4122, which restores the pre-#3918 functionality of treating Git repos as project roots, but keeps the `.venv` customization support it introduced. Give that one a test and see if it fixes your issue.

PR #3090, opened 2014-09-03
Added .gitattributes to preserve LF in shell scripts for Cygwin users

andrewpsy:
fixes #3079

Issue #3963, opened 2015-06-08
.oh-my-zsh/plugins/git/git.plugin.zsh:local:2: not valid in this context: git_cmd^M


shiduanguang:
After the latest upgrade_oh_my_zsh, I got this error and the git plugin didn't work.

apjanke:
Can you be more specific about the errors you're seeing? When do you see that error message?

As an experiment, delete line 2 of `plugins/git/git.plugin.zsh`, where it says `local git_cmd`. Does that fix it?

apjanke:
Also, check to make sure you don't have a DOS-mode file for `plugins/git/git.plugin.zsh` for some reason, if you're on a non-DOS system. I think `^M` is `\r`, and I can reproduce this error by running some bogus DOS-mode text on my OS X 10.9 system.

```
$ eval $(printf 'local git_cmd\r')
local: not valid in this context: git_cmd^M
```

razor-x:
Not sure if this is related, but I just updated and now whenever I open a new shell I get `git_cmd=git` and then my normal prompt, e.g.,

```
git_cmd=git
$
```

apjanke:
It's related; that top-level `local` was added in [a recent update to the `git` plugin](https://github.com/robbyrussell/oh-my-zsh/commit/5b75cc7880de1931a0507800d90eb77ed659fa46). There's an outstanding PR #3962 to fix it.

shiduanguang:
This problem is solved now, the reason is that I set my git line endings incorrectly.
On Mac OS X 10.10.3:
```shell
$ git reset --hard cffeefde903149df8ef48224fedf5205cb2497c3
$ git config --global core.autocrlf input
$ git pull
```

Issue #3687, opened 2015-03-12
Latest update breaks zsh


mikelyons:
the latest update fails with 

```
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:2: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:unsetopt:4: no such option: flowcontrol^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:setopt:6: no such option: complete_in_word^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:setopt:7: no such option: always_to_end^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:8: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:10: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/completion.zsh:11: invalid module name `zsh/complist^M'
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:setopt:2: no such option: auto_pushd^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:setopt:3: no such option: pushd_ignore_dups^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:setopt:4: no such option: pushdminus^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:5: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:10: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:20: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:24: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/directories.zsh:30: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/grep.zsh:18: parse error near `elif'
/Users/lyonm028/.oh-my-zsh/lib/history.zsh:10: parse error near `in^M'
/Users/lyonm028/.oh-my-zsh/lib/key-bindings.zsh:4: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/key-bindings.zsh:88: parse error near `\n'
/Users/lyonm028/.oh-my-zsh/lib/misc.zsh:4: command not found: ^M
/Users/lyonm028/.oh-my-zsh/lib/misc.zsh:setopt:6: no such option: long_list_jobs^M
[...]

Issue #3595, opened 2015-02-16
Everything broken after auto-update


kevinansfield:
Each time I open a new tab I see the following:

```
Last login: Mon Feb 16 10:36:55 on ttys014
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:setopt:2: no such option: auto_pushd^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:setopt:3: no such option: pushd_ignore_dups^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:setopt:4: no such option: pushdminus^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:5: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:10: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:20: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:24: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/directories.zsh:30: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/history.zsh:10: parse error near `in^M'
/Users/kevinansfield/.oh-my-zsh/lib/key-bindings.zsh:4: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/key-bindings.zsh:88: parse error near `\n'
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:4: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:setopt:6: no such option: long_list_jobs^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:7: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:11: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:15: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:18: command not found: ^M
/Users/kevinansfield/.oh-my-zsh/lib/misc.zsh:23: parse error near `\n'
/Users/kevinansfield/.oh-my-zsh/lib/termsupport.zsh:18: parse error near `elif'
/Users/kevinansfield/.oh-my-zsh/lib/theme-and-appearance.zsh:2: command not found: colors^M
[...]

PR #4071, opened 2015-06-20
Add .gitattributes file to use the right line endings

### LINKED ISSUE #4069: Git clone errors if .gitconfig specifies autocrlf = true
This pertains to issues #1730 #3294, #3304 and perhaps others.  Cloning the repository fails if `core.autocrlf = true`.  A workaround is OK but I don't understand why the issue is considered resolved; autocrlf is a standard Git setting and users should be able to clone the repository successfully whichever setting they choose.

---
### WORKAROUND:

Run the following:

``` sh
cd $ZSH
git config core.autocrlf false
git rm --cached -r .
git reset --hard
```

mcornella:
#### STATUS: not ready

~~This enforces the following EOL character in the following files:~~
- ~~**LF** in zsh related files: `.sh`, `.zsh`, `.zsh-theme` and `_*` (completion). This setting forces `\n` as a line ending, given it is the only acceptable one parsed by zsh. This is important for cygwin users, for instance.~~
- ~~**Native EOL** (default in `core.eol` setting). This will use the native EOL in all text files except above, regardless of the core.autocrlf setting which is considered harmful at this point. This means files like READMEs and such. _I'm not 100% sure about this point, feel free to discuss it._~~
- ~~Binary files (as interpreted by git) **will not be EOL-converted**.~~

The current choice is to enforce `LF` on all files by default. This is probably wrong but it is temporary until we figure out the best possible choice.

Closes #3090.
Closes #4069.

Related issues:
#1363, #1483, #2770, #3079, #3595, #3687, #3963, #4402, #4665, #4783.

Related links:
- http://blog.subgit.com/line-endings-handling-in-svn-git-and-subgit/
- http://git-scm.com/docs/gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# force LF ending by default
+* eol=lf


apjanke:
i think this is a good idea. 

But I'm not a fan of setting native EOL. If I understand it correctly, that'll turn on line ending conversion for some files in the Cygwin environment, even if autocrlf is off. That'll mess stuff up, bc a lot of Cygwin tools expect Unix style line endings. I think this should be removed. 

IMHO, enabling autocrlf in Cygwin is a user mistake, since Cygwin is a little schizophrenic about line endings. The best situation is for line conversion to not happen for any file under Cygwin. We're just working around that here, not supporting an environment where line ending normalization is needed or expected.

Since there's no non-Cygwin "native" Windows zsh AFAIK, we don't have to worry about supporting that situation, which is where cold conversion would actually be appropriate.

I'd consider going so far as to skip the *.sh pattern matching, and make .gitattributes a one-liner that disables text for *every* file in the repo. I think that's the most correct behavior under Cygwin.

mcornella:
The native default is for only READMEs and other txt files. This way Windows users could still correctly view and edit those files (for instance, if they are using notepad). This is a really fringe situation though, and I'll go back on my decision if needed.
Can you elaborate on how that will break stuff?

apjanke:
I had misunderstood the behavior at least partially. I think I get it better now after rereading the Git documentation some more.

On plain Windows, Git considers the "native" line ending to be CRLF, but under Cygwin it's LF. That makes me less worried; I had mistakenly thought it was using CRLF for "native" under Cygwin.

Sorry in advance for the long-windedness. Here goes...

First off, I don't think `!eol` works. The `gitattributes` syntax for unsetting a value is `-name`, not `!name`. But unsetting a value isn't the same as making it "unspecified"; it's the same as setting it to false. So I don't know if you can get the "use native line endings" behavior by unsetting `eol`. Though I'm not sure if that's because the `-name` syntax doesn't work to revert `eol` to "unspecified", or if it's because in the normal case, an "unspecified" `eol` is overridden by `autocrlf=true`. I've tested it out with some local repos on Cygwin, and neither `!eol` nor `-eol` seem to have any effect. `eol=lf` does, but that would force line normalization to Unix style under plain Windows too.

###   Potential Issues   ###

There are a couple scenarios where I see potential problems:

1. Text processing of natural language text using Unix tools
2. Computer-readable text with other extensions
3. Files that really do need to be CRLF everywhere
4. It just complicates things

None of these are big cases, but they're a consideration. Probably as much as the case for CRLF conversion: any decent Windows text editor works fine with Unix mode files, and users advanced enough to be running Cygwin, zsh, and OMZ are advanced enough to have something better than Notepad.

#### Text processing of natural language text using Unix tools  ####

This is when someone on Cygwin wanted to work with the human-readable markdown or txt files as text, but wanted to use `grep`, `sed`, `zsh`, `pico` or other Unix tools to do the work.

I don't think this is as big a problem as it used to be. I tested several of the text editors and text processing tools in current Cygwin, and they all seem to handle CRLF mode text just fine. Users may have to adjust the regular expressions they use with the tools, though.

But there are still edge cases here, and while it sounds exotic, I think I even have a live example: in #3913 (https://github.com/apjanke/oh-my-zsh/tree/histsubstr-update-2015_04_26), I wrote a `zsh` script to programmatically update a plugin from its upstream repo. Part of this involves adding OMZ-specific stuff to its README.markdown file. If normalization and "text file only" CRLF checkouts were happening, I think this means that I'd have a CRLF README file, and then the `zsh` script would append a few paragraphs of LF text to it, so you'd have a mixed-mode file. What happens then? Do I just have a locally-messed-up file that gets fixed upon checkin? Does it affect how the normalization happens?

Stuff involving asciidoc, ebook, and docbook processing might fall under this too, if we have plugins for those.

####  Computer-readable text with other extensions   ####

* Other shells: `.bash`, `.ksh`, `.csh` files
* Scripts in other languages like `.py`, `.pl`, and `.m` files
* Files with no extension that are external commands written as scripts with shebang lines
* Data files like `.1` roff-formatted man pages
* Data files with no extension
* Other file types we've overlooked

These are computer-readable scripts and should probably not be normalized, for the same reasons the `.zsh` and `.sh` files aren't. People could add other unforeseen extensions too.

The way the `.gitattributes` in the PR is written, it makes `text=auto` the default, and only excludes a couple specific `.zsh`-related extensions. This means any script-like stuff or text-based data format is going to be detected as text. This could be addressed by making "not text" the default, and only positively identifying known-text extensions like `.txt` and `.markdown` as text.

Examples:
* The current PR doesn't include `*.zsh-template`, so I think the installer running under `autocrlf` would produce a broken `~/.zshrc`.
* `plugins/gitfast/git-completion.bash`
* The chucknorris plugin has a text data file named `chucknorris` to be processed with `strfile`. `strfile` uses LF-only, so it produces a bad data file and the plugin breaks.
* .conf files like `plugins/tmux/tmux.extra.conf` – I don't know if `tmux` likes CRLF files, but I wouldn't bet on it

####  Files that really do need to be CRLF everywhere   ####

Some files really should be in DOS mode, regardless of your platform or conversion settings.

* Windows-related files for plugins that integrate with native Windows applications might need config or script files to be in CRLF mode.
* If we ever write tests, we might want test data to see if things work on CRLF data, like DOS mode text, or traffic from internet protocols like SMTP or HTTP.
* There's probably some screwy X.509 certificate-related file format that specifies CRLF.

If text normalization is on, these will be CRLF in the author's working copy and test out okay, get normalized to LF on checkin (even if they're not running `autocrlf`), and then be in incorrect LF mode when checked out for other users who aren't normalizing.

####   It just complicates things   ####

Aside from specific breakage, enabling text normalization just complicates things, and makes it harder to reason about what's going to happen between different users. This is exacerbated because we're not just using Git as source control for developers; we're using it as a distribution and deployment mechanism for users, too.

How do things behave when you're using a mix of Cygwin and native Git tools to work with a repo? That's not too fringe: I use both Cygwin command-line `git` and the Windows GUI SourceTree for working with OMZ repos.

What's the behavior under various configurations when users pick up an update with the new `.gitattributes`? What happens when we eventually revise `.gitattributes` and it goes out in an update? Which version controls the normalization of files in an update that include an update to the normalization definitions in `.gitattributes`? Are there cases where CRLF files in the working copy might be seen as local changes after a change to `.gitattributes` alters how they should be normalized, thus preventing future updates from running?

For files that need exceptions to this (e.g. as more computer-readable formats are found, or a particular file needs to be CRLF), will changes to plugins then be entangled with changes to the root `.gitattributes`?

What happens if you have your repo on a shared drive that's mounted across multiple platforms? What if it's in your Dropbox or Google Drive? What if you use VM file sharing to move files between your Windows and OS X VMs directly (say, while testing cross-platform behavior of a potential PR) instead of going through an intermediate Git repo? What if it's under a roaming profile on Windows? What about [downstream repackagers who redistribute OMZ outside Git](http://babun.github.io/) and [aren't very sophisticated about this stuff](https://github.com/robbyrussell/oh-my-zsh/pull/3889)?

And philosophically, actively supporting normalization like this just encourages people to use `autocrlf` with this or other projects, which imho complicates things for everyone.

apjanke:
Should this auto-close #3090?

mcornella:
I switched to LF on all files by default. Will refine when I have the time. I also added references to issues which would have been solved by this, to show that a solution is needed even if we have to check for all the edge cases you indicated. Notice that on Windows, the Github clients set `core.autocrlf` by default (though babun doesn't ship with it, which makes it not have that problem).

ordonezalex:
Hello, I just updated oh-my-zsh and got some CRLF errors for the second time this month.

Can we get this merged?

Going through the changed files and removing the CRs is really annoying.

Is there anything preventing this from being merged?

mcornella:
@a-p-o extensive testing and a check of all the edge cases made by @apjanke. I understand this is a very frequent issue but there isn't enough data to merge it. I'd rather instruct users on how to solve their errors than corrupting their repositories and forcing a reinstall.

Hopefully I can make a big enough test suite soon now that I'm [testing docker images](https://github.com/mcornella/dockerfiles).

mcornella:
I've refrained from this approach due to the concerns expressed by Andrew above. A better approach is to fix the git configuration of the repository, setting up the appropriate value for the conflicting settings that users encounter.

These include:
```
core.eol = lf
core.autocrlf = false
```

PR #4074, opened 2015-06-22
Use param expansion instead of `realpath` (fixes #4030)

### LINKED ISSUE #4030: New dependency (realpath) breaks when switching to a directory present with a virtualenv

mail6543210:
This fix #4030, just like what #4025 does, but using different approach.

This fix use `cd -P` instead of detecting the existence of `realpath` or `readlink -f`.


cebor:
:+1:

mcornella:
A couple of fixes:
1. Use `$PWD` instead of calling `pwd` (avoids an extra call).
2. Wrap both `PROJECT_ROOT=....` between `"` to support folders with spaces, like this: `PROJECT_ROOT="..."`.

apjanke:
I missed this issue and put in #4106 recently, which also uses `:A` but is coded slightly differently (only does one `:A` resolution, which I think is all that's needed), and does the `$PWD` and quoting. Sorry about that. This should "closes #4106" so whichever PR is merged cleans up the other one automatically.

mcornella:
Well, #4106 is actually better as you explained there, so if you can address my latest comment I'll suggest we merge that one instead.

PR #2495, opened 2014-02-04
Added path and shortcut for Cygwin and Sublime Text 3

deepakhj:
Added path and shortcut for Cygwin and Sublime Text 3. Tested on Windows 7, Cygwin 64 (2.831), and Sublime Text 3 (Build 3059).

PR #4076, opened 2015-06-22
Sublime plugin : Add cygwin support

ototu:
Here is simple patch to add cygwin support for sublime plugin

Some more location may need to be added but these are standard ones

Fixes #2495


apjanke:
+1

For portability, this should use  `cygpath "$PROGRAMFILES"` instead of a hardcoded `'/cygdrive/c/Program Files'`. Something like this.

```
    local cygprogfiles=$(cygpath "$PROGRAMFILES")
    local cygprogfiles_x86=""
    cygprogfiles_x86=$(cmd /c "echo %ProgramFiles(x86)%")
    cygprogfiles_x86=${cygprogfiles_x86/[\\r\\n]//}
    cygprogfiles_x86=$(cygpath "$cygprogfiles_x86")
    _sublime_cygwin_paths=(
        "$cygprogfiles/Sublime Text 2/sublime_text.exe"
        "$cygprogfiles_x86/Sublime Text 2/sublime_text.exe"
    )
```

(The `cmd` call is necessary because Cygwin doesn't expose both the x86 versions of `%ProgramFiles%` etc, due to syntax issues or it being a 32-bit program or something.) (And the extra `local=""` is a hack because `local` doesn't actually work at the top level of plugin definitions because they're not inside a function.)

And could you include Sublime Text 3, too, since it's out in beta now? The other platforms include it, and put it ahead of Sublime Text 2. It's at these paths by default.

```
        "$cygprogfiles/Sublime Text 3/sublime_text.exe"
        "$cygprogfiles_x86/Sublime Text 3/sublime_text.exe"
```

apjanke:
See #4086 for some info on those `local` calls.

apjanke:
This supersedes the existing #2495 for Sublime Cygwin support. If you put "Fixes #2495" in the description, that'll add an auto-closing link to it.

apjanke:
To get those `local` statements to work, you need to wrap the whole thing in an anonymous function, [like this](https://github.com/apjanke/oh-my-zsh/commit/76cd57530b535ad04bd196487eae62ec3e856d03). And then you don't need the `2> /dev/null` after any of the `local`s. If you have to redirect the output of `local`, it's not working right.

apjanke:
While we're in here, could we change

```
if [[ $('uname') == 'Linux' ]]; then
```

to

```
if [[ "$OSTYPE" == 'linux-gnu' ]]; then
```

to save the cost of a shell-out in the startup sequence?

plugins/sublime/sublime.plugin.zsh
```diff
@@ -41,6 +41,30 @@ elif  [[ "$OSTYPE" = darwin* ]]; then
...
+    cygprogfiles_x86=$(cygpath "$cygprogfiles_x86")
+    local _sublime_cygwin_paths > /dev/null 2>&1
+    _sublime_cygwin_paths=(
+        "$cygprogfiles/Sublime Text 2/sublime_text.exe"
+        "$cygprogfiles/Sublime Text 3/sublime_text.exe"
+        "$cygprogfiles_x86/Sublime Text 2/sublime_text.exe"
+        "$cygprogfiles_x86/Sublime Text 3/sublime_text.exe"
+    )
+
+    IFS=$'\n'
```
apjanke: This change to `$IFS` really ought to be `local`, since it could affect other input processing if it leaks.

ototu:
New patches : 
* Use $OSTYPE for linux
* Save/Restore IFS for cygwin
* Use a default value "." for $* if no file/dir is provided (cygwin) : It made an help message appear
* Merge apjanke's patch
* Remove now unneeded '>/dev/null 2>&1' stuff from local declaration

PR #4083, opened 2015-06-23
Proper utf-8 symbol for SEGMENT_SEPARATOR

pmoust:


apjanke:
Are you sure this is necessary? For me, this patch replaces the "right-pointing-triangle" separator character with a "centered ring" character.

This sounds like you might be seeing an issue like #3713, which wanted to make the same change. It is caused by using old versions of the patched Powerline fonts. Powerline changed the private-use code points they use a while back.  U+E0B0 is the one currently in use. This patch reverts it to the old one.

You may need to just download the current version of the powerline-patched fonts and install them.

pmoust:
Closing in favor of https://github.com/robbyrussell/oh-my-zsh/pull/4091

PR #4086, opened 2015-06-23
sublime plugin: make "local" actually work, using anonymous function

apjanke:
The "local" variables in the `sublime` plugin are not successfully localized. This PR fixes that by wrapping the plugin's code in an anonymous function.

The  `sublime` plugin uses `local` in an attempt to localize its variables. But it doesn't actually work, because it's done at the top level of the file and not inside a function, and plugins are not loaded inside a function call. So those "local" variables end up "global" and persisting in the caller's shell anyway.

```
[@ in ~]
$ echo $plugins
osx themes nyan sublime
[@ in ~]
$ set | grep -a subl | grep _path
_sublime_darwin_paths=(/usr/local/bin/subl '/Applications/Sublime Text.app/Contents/SharedSupport/bin/subl' '/Applications/Sublime Text 3.app/Contents/SharedSupport/bin/subl' '/Applications/Sublime Text 2.app/Contents/SharedSupport/bin/subl' '/Users/janke/Applications/Sublime Text.app/Contents/SharedSupport/bin/subl' '/Users/janke/Applications/Sublime Text 3.app/Contents/SharedSupport/bin/subl' '/Users/janke/Applications/Sublime Text 2.app/Contents/SharedSupport/bin/subl')
_sublime_path=/usr/local/bin/subl
[@ in ~]
$
```

And there are some `>/dev/null` hacks in the existing `local` statements to deal with the fact that when you call `local` at the top level, it outputs the current variable value instead of actually localizing it.

This PR fixes it by wrapping the entire `sublime` plugin in an anonymous function. This makes `local` actually work, prevents variable leakage, and removes the need for the `>/dev/null` hack. And for good measure, it localizes `$_sublime_path`, too, which seems to have been missed.


apjanke:
@paulmelnikow, @GunfighterJ, @pid1 – you're recent authors on the sublime plugin. Care to review this change? It's just a simple fix for correct `local` usage.

GunfighterJ:
Looks good to me :)

paulmelnikow:
:+1:

PR #4041, opened 2015-06-15
Fix jira plugin syntax error due to merge conflict

apjanke:
Fixes a syntax error due to an apparent badly-resolved merge conflict between #2898 and #2955.

Fixes #4027 
Closes #4089

PR #4089, opened 2015-06-24
Fixes parse error caused by extraneous else

### LINKED ISSUE #4027: jira.plugin.zsh:52: parse error near `else'
Two else branches in this if......

ladyborg:
Resolves #4027

### DIFF
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -49,7 +49,7 @@ open_jira_issue () {
     else
       $open_cmd  "$jira_url/browse/$jira_prefix$1"
     fi
-  else 
+
     local addcomment=''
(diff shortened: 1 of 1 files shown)


apjanke:
Conflicts with #4041. 

This fixes the syntax error, but leaves a logic error from the bad merge in place. It'll end up double-invoking stuff. I think #4041 is the right resolution because it removes the entire extra else block, not just the `else` keyword.

If you want to get this fix in quickly, could you test and +1 #4041 to encourage the maintainers to merge it?

ladyborg:
Yep, totally right. Didn't even see that pr before submitting this one. Feel free to close.

ncanceill:
Actually, only you and @robbyrussel can do that.

PR #4090, opened 2015-06-25
Add Lotus.rb bin to bundled commands

Bounga:
[Lotus Web Framework](http://lotusrb.org) is provided with a `lotus` command which has to be run using `bundle exec` when used in bundler context.

This PR adds `lotus` bin to auto bundled commands.

### DIFF
--- a/plugins/bundler/bundler.plugin.zsh
+++ b/plugins/bundler/bundler.plugin.zsh
@@ -17,6 +17,7 @@ bundled_commands=(
   jekyll
   kitchen
   knife
+  lotus
   middleman
   nanoc
   pry


mcornella:
What happens if you wrap it (`lotus='bundle exec lotus'`) and then try to create a new Lotus project with `lotus new project`? Does it work correctly or does `bundle` get in the way?

rickenharp:
I just tried it with `alias lotus='bundle exec lotus'`, which resulted in
```
 $ lotus new foobar --database=postgres --test=rspec
Could not locate Gemfile or .bundle/ directory
```
Adding lotus to bundled_commands instead works like a charm.

mcornella:
Sorry, time passed and I forgot about this one. [Lotus is now Hanami](http://hanamirb.org/blog/2016/01/22/lotus-is-now-hanami.html), can you verify that this is still relevant?

Also, @rickenharp it's weird because the `bundled_commands` logic will resort to creating that same alias (unless lotus is binstubbed). So can you make sure that you have it binstubbed, and that every lotus user will have it as well?

@Bounga can you back your claim that it "has to be run using bundle exec when used in bundler context". Is there some written documentation?

robbyrussell:
We'll close this given how long it's been pending more details for quite some time.

PR #4091, opened 2015-06-26
agnoster: Use current Powerline chars everywhere, and clarify behavior

apjanke:
Updates the agnoster theme to define its special Powerline characters using Unicode escape sequences, and use the current Powerline code point for SEGMENT_SEPARATOR on all platforms.

This makes the theme consistent with the current definition of Powerline, and removes the spurious Mac special case. There is no special handling necessary for Mac; that's just a result of users with old fonts misdiagnosing it.

In 2012, Powerline changed which Unicode code points are used for its special prompt characters, which has resulted in confusion. There are old versions of the Powerline-patched fonts out there that have the glyphs in the old code point slots. This means that whether Powerline looks "right" depends on which version of the fonts you have installed, which version of the code points your theme/prompt definition is using, and whether they are consistent.

We periodically get bug reports or PRs asking to change this character, and they're often because a user has an old version of the Powerline font installed. (#1906, #2869, #2985, #3713, #4001, #4065, #4083) See #3713 for a good long discussion of the issue.

These bugs are hard to diagnose and discuss because when literal Powerline characters are included in the source code, the appearance *of the source code itself* depends on which version of the powerline fonts you have installed and which you're using in your editor.

Switching to using escape sequences in the theme definition source code means that it is unambiguously readable by anybody, regardless of whether they are using Powerline-patched fonts in the editor, and whether they have the new or obsolete version of them.

This change also adds comments to the source code explaining the situation, and discouraging users from making further PRs for the segment separator character.

I've tested on OS X with current versions of the Powerline fonts.

Testers needed. 

Attn: @agnoster – sorry for being presumptuous and updating your theme, but I think this change will really help some users out and cut down on noise in the bug tracker. Look good to you?
Attn: @mrbfrank, @ivanfoo, @gdetrez, since you've done work on the Powerline-related stuff in Agnoster. Thoughts?

Closes #4083

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -32,17 +32,19 @@ CURRENT_BG='NONE'
 
 # Special Powerline characters
 
-# NOTE: This segment separator character is correct.  In 2012, Powerline changed
-# the code points they use for their special characters. This is the new code point.
-# If this is not working for you, you probably have an old version of the 
-# Powerline-patched fonts installed. Download and install the new version.
-# Do not submit PRs to change this unless you have reviewed the Powerline code point
-# history and have new information.
-# This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
-# what font the user is viewing this source code in. Do not replace the
-# escape sequence with a single literal character.
-SEGMENT_SEPARATOR='\ue0b0' # 
-
+() {
+  local LC_ALL="" LC_CTYPE="en_US.UTF-8"
+  # NOTE: This segment separator character is correct.  In 2012, Powerline changed
+  # the code points they use for their special characters. This is the new code point.
+  # If this is not working for you, you probably have an old version of the 
+  # Powerline-patched fonts installed. Download and install the new version.
+  # Do not submit PRs to change this unless you have reviewed the Powerline code point
+  # history and have new information.
+  # This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
+  # what font the user is viewing this source code in. Do not replace the
+  # escape sequence with a single literal character.
+  SEGMENT_SEPARATOR=$'\ue0b0' # 
+}
 
 # Begin a segment
 # Takes two arguments, background and foreground. Both can be omitted,
@@ -83,7 +85,12 @@ prompt_context() {
 
 # Git: branch/detached head, dirty status
 prompt_git() {
-  local PL_BRANCH_CHAR='\ue0a0'         # 
+
+  local PL_BRANCH_CHAR
+  () {
+    local LC_ALL="" LC_CTYPE="en_US.UTF-8"
+    PL_BRANCH_CHAR=$'\ue0a0'         # 
+  }
   local ref dirty mode repo_path
   repo_path=$(git rev-parse --git-dir 2>/dev/null)
 
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -7,6 +7,9 @@
 #
 # In order for this theme to render correctly, you will need a
 # [Powerline-patched font](https://github.com/Lokaltog/powerline-fonts).
+# Make sure you have a recent version: the code points that Powerline
+# uses changed in 2012, and older versions will display incorrectly,
+# in confusing ways.
 #
 # In addition, I recommend the
 # [Solarized theme](https://github.com/altercation/solarized/) and, if you're
@@ -27,12 +30,19 @@
 
 CURRENT_BG='NONE'
 
-# Fix odd char on mac
-if [[ `uname` == 'Darwin' ]]; then
-    SEGMENT_SEPARATOR='\ue0b0'
-else
-    SEGMENT_SEPARATOR=''
-fi
+# Special Powerline characters
+
+# NOTE: This segment separator character is correct.  In 2012, Powerline changed
+# the code points they use for their special characters. This is the new code point.
+# If this is not working for you, you probably have an old version of the 
+# Powerline-patched fonts installed. Download and install the new version.
+# Do not submit PRs to change this unless you have reviewed the Powerline code point
+# history and have new information.
+# This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
+# what font the user is viewing this source code in. Do not replace the
+# escape sequence with a single literal character.
+SEGMENT_SEPARATOR='\ue0b0' # 
+
 
 # Begin a segment
 # Takes two arguments, background and foreground. Both can be omitted,
@@ -73,6 +83,7 @@ prompt_context() {
 
 # Git: branch/detached head, dirty status
 prompt_git() {
+  local PL_BRANCH_CHAR='\ue0a0'         # 
   local ref dirty mode repo_path
   repo_path=$(git rev-parse --git-dir 2>/dev/null)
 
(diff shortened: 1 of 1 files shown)


agnoster:
Yeah, this is definitely the way to go. My official version's *main* divergence from the current OMZ version is (I think) that I use defines for each special character and define them all in terms of unicode escape sequences: https://gist.github.com/agnoster/3712874

agnoster:
(I don't actually use OMZ currently, so I'm not 100% sure if the latest version works there - I retooled it to work out of the box on vanilla ZSH so I could trim down dependencies. It's possible that just "merging" that latest version would work, if anyone's adventurous enough to try that out.)

mrbfrank:
:+1: 
@apjanke great comments in #3713 
@agnoster great theme I use it every day, thanks for sharing

apjanke:
Thanks!

@agnoster: The code is rearranged enough that `diff` doesn't say much, but I tried out your current version briefly under OMZ and it seems to work fine. I think you're right, and the only real changes are those Unicode escape sequences, and you're `print` instead of `echo` now. (Which we should be too, for portability.)

I'd be hesitant to just merge it in, though, because the new version is using some global variables like `$BRANCH`, `$DETACHED`, and `$PLUSMINUS`, and those are simple-enough names they seem like they could collide with user variables. (OMZ sets enough global variables as it is; there's been some effort to reduce its global namespace footprint lately.)

apjanke:
I have noticed one portability glitch: These `\uXXXX` escape sequences depend on being in a Unicode locale, like `xx_YY.UTF-8`. If the user is in a non-Unicode locale, like with `LC_ALL=C`, those escape sequences [will fail with a `character not in range` error](https://github.com/robbyrussell/oh-my-zsh/issues/4065) as in #4065. I don't know what the Right Thing to do is here. Use of those escape sequences could be protected with a `local LC_ALL` in the functions where they're used, but I don't know if that sort of hack is an appropriate use of locale settings.

For now, I'm leaving them unprotected, assuming `agnoster` is only intended to work in Unicode locales, and a `LC_ALL=C` at the interactive level is a misconfiguration of some sort.

But there may be users using Shift-JIS or other non-Unicode encodings where a non-Unicode interactive locale makes sense. And most of the non-Powerline characters probably have encodings in those other fonts. So maybe locale protection upon character definition is the right thing to do, and then rely on character set encoding on prompt presentation to make the symbols usable or degrade gracefully (instead of emitting confusing errors).

mcornella:
The unicode locale is (sort of) implemented in `lib/misc.zsh`. We "_fixed_" it on #3563 but [it wasn't perfect](https://github.com/robbyrussell/oh-my-zsh/pull/2012#issuecomment-75542823). So this technically should be set by default. We just need `LC_TYPE` though (I think).

Your suggestion of wrapping those characters with a locale redefinition is very interesting and one that seems worthy of consideration. 

A couple fixes to your suggested code though: you can use `$'\char'` syntax to define a special character (_e.g._ `$'\n'` or `$'\ue0b0'`) without the need of `printf`; `C.UTF-8` should be more widely available (it just requires unicode support for it to be available, while `en_US` requires the English locale); and using `LC_TYPE` instead of `LC_ALL` should do the trick as well. 

Let's test this further and see if it works. Great job!

apjanke:
Thank you!

`$'\char'` and `C.UTF-8` sound great. I'll switch to those. (I'll see if I can find a reference on `C.UTF-8`'s availability.) Avoiding an `en` language dependency would be great, and it would make the code intent clearer too. Thanks for the suggestion.

This stuff from `lib/misch.zsh`?

```
# only define LC_CTYPE if undefined
if [[ -z "$LC_CTYPE" && -z "$LC_ALL" ]]; then
	export LC_CTYPE=${LANG%%:*} # pick the first entry from LANG
fi
```

I don't think that addresses this at all. It only handles it if the user has undefined LC_* settings. (And I don't think it should be necessary, because `locale` implementations should handle that defaulting in the client.) The case we need to worry about is where the LC_* stuff *is* set, and possibly "correctly", but it just isn't UTF-8. (I'm not even sure what that stuff in `lib/misc.zsh` *is* addressing. I need to review the issue history and locale API doco on that to really understand it.)

It's not just `LC_CTYPE`, though that's what I had expected myself, too, before testing it. We also need to protect against `LC_ALL` specifically, because `LC_ALL` *overrides* the setting for any of the individual locale classes, instead of providing a default value for them. Maybe unsetting `LC_ALL` and setting just `LC_CTYPE` would be more readable; I dunno.

```
➜  ~  LC_ALL=C zsh -l
➜  ~  foo='\u2603'; echo $foo
zsh: character not in range

➜  ~  () { local LC_CTYPE=en_US.UTF-8; foo='\u2603'; echo $foo }
(anon): character not in range

➜  ~  () { local LC_ALL=en_US.UTF-8; foo='\u2603'; echo $foo }
☃
➜  ~  () { local LC_ALL="" LC_CTYPE=en_US.UTF-8; foo='\u2603'; echo $foo }
☃
```

apjanke:
I've updated the branch with "local LC_*" protection for the escape sequences, and am using that `$'\u...'` syntax for eager evaluation. Using the `en_US.UTF-8` locale for now.

I dunno about this `C.UTF-8` locale. How widely available is it? My OS X 10.9.5 machine and Cygwin 
machines don't have it. Debian 7 does have it. And it looks like as recently as this year, [glibc is discussing whether to include it as a default locale](http://sourceware.org/ml/libc-alpha/2015-02/msg00477.html).

```
$ locale -a | grep C.UTF-8
$
```
If I set `LC_CTYPE=C.UTF-8` on OS X, then it seems to fall back to plain `C`.

```
$ LC_CTYPE=C.UTF-8
[~]
$ locale
LANG="en_US.UTF-8"
LC_COLLATE="C"
LC_CTYPE="C"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```

In this case, if I use the `\u` escape sequences, they *appear* to work, and the right characters appear in the terminal. But I don't know if that's because everything's getting processed correctly, or if raw bytes are being passed around that happen to line up with my terminal settings, or maybe OS X components that are always assuming UTF-8. I'm not at all sure that it's the correct, portable thing to do. It seems like `C.UTF-8` would do what we want, but it's a relatively new thing, and doesn't have a universal install base.

Since English is so widespread in the computing world, are there really many systems that wouldn't have `en_US.UTF-8` installed, even if it's not used as the default?

Do any of our users have an "exotic" non-Western system that would be set up with a default non-UTF-8 locale, and maybe not even have English installed, that could be used as a test?

apjanke:
@agnoster: Oh... also, OMZ's `agnoster` has Mercurial support (`prompt_hg()`). Is that something we added in ourselves, or did you remove it at some point?

agnoster:
I believe it got added in OMZ and I never ported it over. I should really
make mine a proper repo to make it easier to merge changes, I just never
thought it would get popular >.<

pmoust:
I believe this is the way to go . 
I 'd say let's have this merged in, if an exotic non-utf8 locale user pops in, it can be reviewed when he/she raises an issue. 
I *feel* this PR will satisfy more peeps that use `oh-my-zsh` with a slight chance that it might get funky for very few.

mcornella:
We should also change the reference URL, it has now changed to https://github.com/powerline/fonts.

PR #4093, opened 2015-06-26
osx: fix tab() functions so the cd works in iTerm2, and status/output are better

### LINKED ISSUE #3670: Commit ce607d3fcb4c730c991563b4c62725cd89ce0ced breaks "tab" functionality

apjanke:
The `osx` plugin's `tab` functions are not working quite right. I think this is what #3670 is referring to.

Fixes #3670
Closes #3101 (includes the same fix)

###  iTerm issues

In iTerm, when you open a new tab, the command to `cd` to the directory you launched it from is displayed, but not executed.

screen shot 2015-06-26 at 2 07 45 am

This seems to be because a trailing semicolon in the `write text` argument prevents it from being executed. The `keystroke return` from ce607d3fcb4c730c991563b4c62725cd89ce0ced appears to be an attempt to fix this, but it didn't work, and was just producing an error. (I think because `keystroke` is a verb for System Events, not for iTerm.)

In my theme, a blue "✘" at the beginning indicates an error exit status (nonzero `$?`).

This PR fixes `tab` for iTerm by removing the trailing semicolon from the commands it runs in the new tab.

###  Terminal.app issues

screen shot 2015-06-26 at 1 50 50 am

In Terminal.app, `tab()` works to create a new tab, but it outputs a useless-seeming "tab N of window id XXXX". I'm assuming people don't actually want that output. And `tab` gives an error exit status even when it succeeded. This is because `tab` goes on to test `[[ "$the_app" == 'iTerm' ]] && {` after the code for Terminal.app. That `[[ ... ]] && { ... }` syntax leaves `$?` as nonzero when the test fails, and since that was the last statement in the `tab` function, that becomes the exit status of `tab`. This PR fixes that by replacing the `[[ ... ]] && { ... }` syntax with `if [[ ... ]]; then ...` tests, which don't have the `$?` side effects.

`split_tab` and `vsplit_tab` are not supported under Terminal.app. Calling them gives an error exit status, but no error message to indicate why they failed. (And if the user doesn't have an exit status indicator in their prompt, they may not know what happened.) This PR adds explanatory error messages when you call a `split` function from within an unsupported terminal app. Here's what it looks like under this PR.

screen shot 2015-06-26 at 2 04 29 am

###  Miscellaneous

The app detection test, which was the same for all the `*tab` functions, is factored out to a function.

The existing code is discarding error messages from `osascript` to `/dev/null`. Those error messages are meaningful, and might have helped diagnose this. I restored them.

The `split_tab` and `vsplit_tab` have the unfortunate side effect of bringing all of the iTerm windows to the front. Probably due to the `tell application "iTerm" to activate`. This PR does not fix that.


apjanke:
@Marco-Lindsay – this should fix the `tab()` problem in #3670 for you. Could you test it out?

Marco-Lindsay:
+1

mcornella:
Needs to be rebased, but :+1: 
/cc @robbyrussell

apjanke:
Merged to catch up with master. I can't figure out how to rebase it "cleanly" at this point, but it's only one extra commit; hopefully that's fine.

PR #4095, opened 2015-06-26
Use 'quit all' instead of counting needed quits

kd35a:
Instead of quiting each window manually, quit all directly.

### DIFF
--- a/plugins/vundle/vundle.plugin.zsh
+++ b/plugins/vundle/vundle.plugin.zsh
@@ -13,15 +13,15 @@ function vundle-init () {
 
 function vundle () {
   vundle-init
-  vim -c "execute \"PluginInstall\" | q | q"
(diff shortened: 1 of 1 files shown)


mcornella:
I don't use vundle nor vim, but I assume that `qa` is needed to close any number of tabs, while the current `q | q` solution just closes 2 (which is exactly what is needed, but `qa` is shorter). I'm getting this from http://stackoverflow.com/a/510333.

I'd like at least one independent tester though, if you can find one please direct him this way.
cc @dnixx @kokx @bertabus @Gonzih: can you guys test this?

kokx:
@mcornella It is cleaner / 'nicer' this way, but in functionality or speed it doesn't matter at all (it's not any faster or slower). It still works exactly the same way.

mcornella:
Ok thanks for confirming. 
/cc @robbyrussell

PR #4098, opened 2015-06-27
'la -> ls -A' shows all files in column format; 'l -> ls' shows output of 'ls'

garyduong:
Three of the aliases, `l`, `la`, `ll`, had nearly the same output when executed. I think this implementation will differentiate more clearly what each of its intended purposes are.

Here are the changes I have proposed.
- `l`: `ls` output. Trivial.
- `la`: `ls -A` output. Lists *all* files and directories, in column format. Much easier to view at a glance.

I had also considered `alias -='cd -'`, but there is a rule against aliases for quoted and escaped characters.

Let me know what you think!

screen shot 2015-06-27 at 15 05 17

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -24,9 +24,9 @@ alias d='dirs -v | head -10'
 
 # List directory contents
 alias lsa='ls -lah'
-alias l='ls -lah'
+alias l='ls'
 alias ll='ls -lh'
-alias la='ls -lAh'
+alias la='ls -A'
 
 # Push and pop directories on directory stack
 alias pu='pushd'


mcornella:
I like these. In fact, I have [quite the array of ls aliases](https://github.com/mcornella/dotfiles/blob/master/aliases#L16-L23) in my dotfiles. But I don't know what the general feeling is regarding this aliases, we should get more people to discuss them.

@apjanke @ncanceill what do you think?

apjanke:
These aliases have been around with basically their current meanings [since at least Dec 2012](https://github.com/robbyrussell/oh-my-zsh/commit/11ff3b642db87012ed8a19d6a0c5baa5dcd6a004#diff-4bf9d92828fd9c95c73d458dfb947ab5), that's a good three years. We don't know how many users are using them or how many prefer the current behavior, and we don't have a mechanism for notifying users they will be changing behavior, aside from just changing it and letting them notice next time they use them. I think the presumption should be in terms of keeping the existing behavior of aliases that have been around for a while. Regardless of what I think of the particular aliases, it's a matter of what the general userbase prefers.

I do [my own `ls*` aliasing too](https://github.com/apjanke/dotfiles/blob/308dd67c90767950d056ac672950317857401de7/generic/dots/dotfiles/zshbashrc.zsh#L18-L22). 

> I had also considered alias -='cd -', but there is a rule against aliases for quoted and escaped characters.

That restriction on plain unquoted strings for aliases is only in effect if the `POSIX_ALIASES` option is on. It is off by default. The command `alias -='cd -'` fails because the `-` looks like it's introducing an option to the alias command. To make it work, use `--` to indicate the end of options.

```
$ alias -- -=foo
$ -
zsh: command not found: foo
$
```

Using this form, `-` is [already aliased to `cd -` in OMZ](https://github.com/robbyrussell/oh-my-zsh/blob/5de1704a0a5d483d15528a72a3be60d44356b686/lib/directories.zsh#L11).

mcornella:
Let's wait on this until we have a safe aliasing infrastructure (#4585) in place then.

ncanceill:
Sorry guys, been idle for some time.

Personally I am used to using `ls` when I want basic `ls` output. When I feel like using `l`, I am already in an active state of mind, and probably wiggling around in the system, so I like having the full column output including hidden files.

Anyhow, @mcornella said it best, as always.

I fear _“let us wait for #4585”_ is going to become the routine answer to many issues and PRs, right along _“please do not send more themes”_ and _“this is a duplicate of #XXXX”_.

Issue #3557, opened 2015-02-07
[feature] add 'docker-composer' plugin bundler


itsNikolay:
https://github.com/docker/fig
Docker team took the `fig` tool to it's support, and renamed to `docker-composer`
could somebody add aliases for it, it's very usefull tool which will replace Vagrant in further.
many thanks

robbyrussell:
@itsNikolay Want to submit a PR for this? You can look at existing plugins for examples

itsNikolay:
@robbyrussell sure, I will try it

wernight:
bashcompinit gives some errors with the Bash autocomplete provided:

````
_docker-compose:20: command not found: _get_comp_words_by_ref
_docker-compose:27: parse error: condition expected: 1
_docker-compose:declare:48: not valid in this context: _docker-compose_docker-compose
...
````

Suggesting to use argparse in [docker-compose](https://github.com/docker/compose/issues/1150)

paulodiovani:
Could this one be used? https://github.com/sdurrheimer/docker-compose-zsh-completion.git

I added in `plugins` and it's working fine.

wernight:
Using another projet's auto-complete will lead to out of sync auto-complete.

It should be supported natively.

mcornella:
Solved in #4099, please close this issue.

PR #4099, opened 2015-06-28
Plugin now uses completion script from docker-compose repo.

moncho:
It removes changes done in #3848 but it solves #3557.


apjanke:
Is this license-compatible with oh-my-zsh? OMZ [is MIT licensed](https://github.com/robbyrussell/oh-my-zsh/blob/master/MIT-LICENSE.txt), but docker-compose [says it's Apache licensed](https://github.com/docker/compose/blob/8346186469eb046ebd619727ac101c83322e698b/LICENSE), which requires distribution under the same terms and with a copy of the Apache license.

moncho:
I did not consider the original project license, sorry.

Actually the script origin is a different repo: https://github.com/sdurrheimer/docker-compose-zsh-completion , but it has been merged into docker-compose repo.

Using the original repo was already suggested on #3557, but was discarded. So maybe this PR is not  such a good idea, but it works :)

apjanke:
Sounds like they're considering it to be part of the main project and using that license – sdurrheimer/docker-compose-zsh-completion#1. I'm not sure what the upshot of this is, but since the you know where the original code came from, maybe if you ask nicely the author would dual-license the original source under the MIT license and then you could just copy it in to OMZ and maintain it as a forked plugin (instead of a downstream synced copy).

moncho:
@apjanke Following your suggestion, I have contacted the author and asked him about adding the MIT license. Now we wait :)

apjanke:
Upstream has added the MIT license, per sdurrheimer/docker-compose-zsh-completion#1. You're good to go.

apjanke:
:+1:

jonafato:
+1 would love to see this merged

PR #4102, opened 2015-06-29
Load completions from custom config folder

marknsikora:
The present setup loads custom completions from `$ZSH/completions` which is against what everything else is doing. Plugins and profiles are both loaded from the custom folder. The user should be able to store their all custom modifications in `$ZSH_CUSTOM`.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -5,14 +5,14 @@ fi
 
 # Initializes Oh My Zsh
 
-# add a function path
-fpath=($ZSH/functions $ZSH/completions $fpath)
-
 # Set ZSH_CUSTOM to the path where your custom config files
 # and plugins exists, or else we will use the default custom/
 if [[ -z "$ZSH_CUSTOM" ]]; then
     ZSH_CUSTOM="$ZSH/custom"
 fi
+
+# add a function path
+fpath=($ZSH/functions $ZSH/completions $ZSH_CUSTOM/completions $fpath)
 
 # Set ZSH_CACHE_DIR to the path where cache files should be created
 # or else we will use the default cache/


apjanke:
This whole step might be vestigial, since OMZ no longer contains `$ZSH/completions` or `$ZSH/functions` folders.

Should `$ZSH_CUSTOM/completions` come before `$ZSH/completions` in the `$fpath` order? Typically, stuff in the custom folder overrides stuff in the main `$ZSH` hierarchy.

Should `$ZSH_CUSTOM/functions` also be added, for consistency?

marknsikora:
My goal was to avoid any breaking changes, I'm not sure what the projects policy is on those. 

As far as adding functions I'm slightly against that. Having all user completions consolidated in one place seems like the better option.

apjanke:
In theory, any change here has the potential to be breaking. Users might have broken code in them under `$ZSH_CUSTOM` and be relying on them not being used, or be doing other wacky stuff with them (`zsh` runs arbitrary code, after all). But since neither `$ZSH_CUSTOM/completions` nor `$ZSH_CUSTOM/functions` is currently in use, I think the risk is pretty low here, especially with regards to changing precedence order, because there are currently no files in the main OMZ directory to resolve precedence for. And if users have added them, they've probably only put them in one place or the other.

What's the reasoning for excluding `functions`? The `$ZSH_CUSTOM` mechanism is there to allow users to override parts of the core, and if `functions` isn't included, there's no way for them to do that if `$ZSH/functions` gets restored. It seems like the omission of both `$ZSH_CUSTOM/functions` and `$ZSH_CUSTOM/completions` here was just an oversight. E.g. here's the original discussion on these directories: completions #351, functions b865f2ab38156c33bdb1d6ffc6df56b3f2d53588. It sounds like the relationship to `$ZSH_CUSTOM` wasn't fully considered, as the `custom` dir was still in the process of being added.

robbyrussell:
We'll close this given how long it's been pending further discussion for quite some time.
.

PR #4105, opened 2015-06-29
simple theme matching RedHat bash looks

hellozt:
most zsh themes are well departed from the basic bash themes.  This adds a minimal theme that tries to mimic the bash theme used on most RHEL based distros


mcornella:
Hi @hellozt, we have lots of themes already. You can add yours as an [external theme in the wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes).

PR #4106, opened 2015-06-29
virtualenvwrapper: Remove `realpath` dependency using zsh `:A`

### LINKED ISSUE #4030: New dependency (realpath) breaks when switching to a directory present with a virtualenv

### LINKED ISSUE #4052: Hanging at startup with `virtualenvwrapper` and `last-working-dir` plugins
I have to ^C at the start of zsh if I have both these plugins.

I found that 7b478d75a3a0db466e78b6e46a0e468d64354618 is the bad commit after bisecting.

### LINKED ISSUE #4079: workon_cwd:6: command not found: realpath
I always open the virtualenv plugin,but I upgrade oh-my-zsh today, when I cd to a project dir, it tell me `workon_cwd:6: command not found: realpath`

apjanke:
PR #3918 introduced a dependency on the command `realpath` to the virtualenvwrapper plugin. This is a problem because `realpath` is not installed on many systems by default. (It's not even available in the main Homebrew tap for OS X.) This caused problems for some users (#4079).

This PR removes the `realpath` dependency by replacing `realpath` calls with `zsh`'s native `:A` [parameter expansion modifier](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers), which does the same thing as `realpath`, except portably and with no external dependency. This is also a performance win because it avoids the need to shell out to an external command.

The original code also called `realpath` at every step of the directory tree it was walking up. This is unnecessary because the first `realpath` or `:A` call resolves intermediate symlinks. This PR removes the extra calls.

Uses `$PWD` instead of `$(pwd)`, to avoid an unnecessary shell-out (for performance) and to make the `:A` modifier call a one-liner.

Fixes #4079.
Fixes #4030.
Fixes #4052. (I think.)
Closes #4025, which attempted to fix the issue by checking for the existence of `realpath`. Using `:A` results in simpler, faster code. (Sorry for stepping on your PR, @SandyRogers, but people are asking for a fix for immediate breakage.)
Closes #4074, which is basically the same thing. (Sorry for stepping on your PR, too, @mail6543210. I didn't see it when I searched for existing fixes.)
Closes #2376 (Really, #3918 should have auto-closed #2376, but since this follows on #3918's commits, may as well do it here to avoid the extra clerical work.)

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -30,16 +30,17 @@ if [[ "$WORKON_HOME" == "" ]]; then
 fi
 
 if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
-  # Automatically activate Git projects's virtual environments based on the
+  # Automatically activate Git projects' virtual environments based on the
   # directory name of the project. Virtual environment name can be overridden
   # by placing a .venv file in the project root with a virtualenv name in it
   function workon_cwd {
     if [ ! $WORKON_CWD ]; then
       WORKON_CWD=1
       # Check if this is a Git repo
-      PROJECT_ROOT=`pwd`
+      # Get absolute path, resolving symlinks
+      PROJECT_ROOT="${PWD:A}"
       while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" ]]; do
-        PROJECT_ROOT=`realpath $PROJECT_ROOT/..`
+        PROJECT_ROOT="${PROJECT_ROOT:h}"
       done
       if [[ "$PROJECT_ROOT" == "/" ]]; then
         PROJECT_ROOT="."
@@ -50,7 +51,7 @@ if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
       elif [[ -f "$PROJECT_ROOT/.venv/bin/activate" ]];then
         ENV_NAME="$PROJECT_ROOT/.venv"
       elif [[ "$PROJECT_ROOT" != "." ]]; then
-        ENV_NAME=`basename "$PROJECT_ROOT"`
+        ENV_NAME="${PROJECT_ROOT:t}"
       else
         ENV_NAME=""
       fi


davidetruffo:
Hi! 
I just tried out this solution and:
- it removes the error on first "cd" 
- it still DOES NOT activate the virtualenv for the current directory (if any).

Am i missing something?
Thanks for your time :)
d.

mcornella:
@davidetruffo can you try #4074 ?

davidetruffo:
Of course i can @mcornella 
But unfortunately no venv active still..

i'm no backend/programmer at all, but from my understanding the main issue is that the ENV_NAME is always an empty string, and this makes the function to never get into the activation condition... 
````
while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" ]]; do
        PROJECT_ROOT=${PROJECT_ROOT:A:h}
      done
      if [[ "$PROJECT_ROOT" == "/" ]]; then
        PROJECT_ROOT="."
      fi
      # Check for virtualenv name override
      if [[ -f "$PROJECT_ROOT/.venv" ]]; then
        ENV_NAME=`cat "$PROJECT_ROOT/.venv"`
      elif [[ -f "$PROJECT_ROOT/.venv/bin/activate" ]];then
        ENV_NAME="$PROJECT_ROOT/.venv"
      elif [[ "$PROJECT_ROOT" != "." ]]; then
        ENV_NAME=`basename "$PROJECT_ROOT"`
      else
        ENV_NAME=""
      fi
````

and the result is always that 
````
ENV_NAME == ""
````
so it doesn't satisfy this
`````
if [[ "$ENV_NAME" != "" ]]; then
````

**Comment by mcornella** [comment]
`$ENV_NAME` will only be set if there's a `.venv` file *or* folder in the current directory or any of the parents. So you have to create a `.venv` file with the `ENV_NAME` you want as content, and put it somewhere in one of the parent folders to test this.

@apjanke this version of #4074 doesn't translate exactly as the realpath alternative. Here only the `$PWD` is translated into the symlink target, not the rest of the parents. The realpath alternative follows exactly what the [`git rev-parse`](https://github.com/robbyrussell/oh-my-zsh/pull/3918/files#diff-357c21979357d891f4cb0e48d2cab6f2L35) command did before (see https://github.com/robbyrussell/oh-my-zsh/pull/4025#issuecomment-112897449), so I'll support #4074 instead. BTW, there's a small detail to polish, but I'll propose it there instead.

**Comment by apjanke** [comment]
@mcornella: Are you basing that on a test, or examining the code? I believe the behavior of this `:A` version is the same as the `realpath` version. Both `realpath` and `zsh`'s `:A` modifier resolve symlinks all the way up the path, not just at the current leaf file like `realink` does. So the calls to `realpath` at every step up the tree were redundant in the first place; only the one call on `$PWD` was necessary. I removed it for efficiency and clarity. I've done some basic tests on `:A` vs `realpath` on OS X in this situation, and they behaved the same. (Using the `realpath` from `brew tap iveney/mocha; brew install realpath`.)

Here's a test case.
```
cd ~
mkdir -p tmp
cd tmp
mkdir -p foo/bar
mkdir -p baz/real/path
ln -s ~/tmp/baz/real/path foo/bar/link
cd foo/bar/link
mkdir -p dirs/under/link
cd dirs/under/link
echo $PWD
echo ${PWD:A}
```
I get this on OS X 10.9.5.
```
$ echo $PWD
/Users/janke/tmp/foo/bar/link/dirs/under/link
$ echo ${PWD:A}
/Users/janke/tmp/baz/real/path/dirs/under/link
```

Note that it resolved the symlink three levels up. You can take it another step, and make a symlink to a dir under the original symlink, so you have multiple levels of symlink indirection at different levels in the file path.

```
cd ~/tmp/foo
ln -s ~/tmp/foo/bar/link/dirs another_symlink
cd another_symlink/under/link
echo $PWD; echo ${PWD:A}
```

Both get resolved, and the canonical non-symlinked path is the result.
```
$ echo $PWD; echo ${PWD:A}
/Users/janke/tmp/foo/another_symlink/under/link
/Users/janke/tmp/baz/real/path/dirs/under/link
```

Since there are no longer any symlinks in the file path, there is no need to do symlink resolution with `:A` or `readlink` again as you're walking up the path.

**Comment by apjanke** [comment]
Either way, how did I miss the existence of #4074? Sorry about that; I need to search harder next time.

I miss real bug tracking systems where you can see dependency and other relationship graphs. Anybody know how to view that for GitHub issues?

**Comment by davidetruffo** [comment]
@mcornella thanks for the explanation!
It still doesn't activate the venv for me..
Here's a sample of the structure i used to have (and it used to work)
````
Projects
  |- project_1
  |- project_2
  |- venvs
       |- project_1 (venv for project_1)
       |- project_2
````
so typing
`````
cd Projects/project_1
``````
would activate Projects/venvs/project_1

I have now updated my structure to test (i hope properly :) ) to this:
````
Projects
  |- project_1
       |- .venv (a file, containing "project_1" as a string)
  |- project_2
  |- venvs
       |- project_1 (venv for project_1)
       |- project_2
````
But still nothing (or better, i get $ENV_NAME set, but the virtualenv is not activated

I'm wondering if i misunderstood it all, meaning that this fix just removes the dependency on `realpath`or it should ALSO activate/deactivate the virtualenv just cd-ing in/out

Again, thanks for your time :)

**Comment by mcornella** [comment]
@apjanke: wow I hadn't thought of that possibility, so one `:A` is all we need. Thanks for the extensive response and test suite. Can you do one more test? Make a git repository inside one of these folders and see what `git rev-parse --show-toplevel` returns (if it's the canonical path or also contains symlinks).

@davidetruffo: 

1. Without `.venv` doesn't work as it used to be due to the way the [new algorithm is done](https://github.com/robbyrussell/oh-my-zsh/blob/599831b85a5fd4564ab820530e2bb45f37ca3f99/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh#L41-L46): `PROJECT_ROOT` is used to traverse the path up to the root element, but if no `.venv` is found, `PROJECT_ROOT` stays `/` and then `PROJECT_ROOT=.` in the next if. Therefore in the [following if chain](https://github.com/robbyrussell/oh-my-zsh/blob/599831b85a5fd4564ab820530e2bb45f37ca3f99/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh#L52-L56) the execution doesn't enter the proper if branch. 
The proper way to fix this would be to use a temporary variable holding the current folder being traversed, and only set `PROJECT_ROOT` if `.venv` was found. Can you implement this @apjanke ?

2. The automatic activation is done [in this part of the code](https://github.com/robbyrussell/oh-my-zsh/blob/599831b85a5fd4564ab820530e2bb45f37ca3f99/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh#L59-L65). So it either runs `workon $ENV_NAME` or `source $PROJECT_ROOT/.venv/bin/activate`, but it will not activate if neither of those exist.

**Comment by davidetruffo** [comment]
@mcornella thanks for the sweet and clear explanation!
So to make it work (activating/deactivating the right venv) i need to update my structure, am i correct? (just to be super-duper clear)

**Comment by mcornella** [comment]
It depends, **(1)** will work when it is fixed. **(2)** should work as long as you have `$WORKON_HOME` set to `Projects/venv`, can you post the output of `echo $WORKON_HOME`?

**Comment by apjanke** [comment]
You're welcoeme.

Here's the `git` part of the test. The `git rev-parse --show-toplevel` is identical to the `:A` implementation: canonical path, with all symlinks removed.

```
cd ~/tmp/foo/another_symlink/under/link/
mkdir git-repo; cd git-repo; git init
echo ${PWD:A}; git rev-parse --show-toplevel;
if [[ "$(git rev-parse --show-toplevel)" == "${PWD:A}" ]]; then echo SAME; fi
```
My results:

```
➜  git-repo git:(master) echo ${PWD:A}; git rev-parse --show-toplevel;
/Users/janke/tmp/baz/real/path/dirs/under/link/git-repo
/Users/janke/tmp/baz/real/path/dirs/under/link/git-repo
➜  git-repo git:(master) if [[ "$(git rev-parse --show-toplevel)" == "${PWD:A}" ]]; then echo SAME; fi
SAME
```

And about that other bit with the "no-`.venv`" case – sure, I'll add that in. This PR was originally just to fix the `realpath` dependency, but it should really take care of any bugs introduced by the recent changes that added in the `realpath` stuff. Give me an hour or two...

mcornella:
Perfect, so git behavior confirmed. Thanks for the hard work!

benjaoming:
:+1: for merging this ...and if there is an issue with `:A`, then that can be a separate issue.

davidetruffo:
@mcornella 
here's the output for `echo $WORKON_HOME` ->  `/Users/davidetruffo/Projects/venvs`
it is set in my .zshrc

apjanke:
Uh oh. So, I reviewed the [old version of virtualenvwrapper](https://github.com/mcornella/oh-my-zsh/blob/b62911660d3743d3dfb0116f5b05aabd67a24a06/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh) and #3918, and the expected behavior in edge cases isn't clear to me.

Originally, before #3918, it looks like your "project root" was the root dir of the lowermost `git` repo you were in, and the "$ENV_NAME" was the basename of that project root directory, and it could be overridden by a ".venv" dir directly under the Git repo. With #3918, the `.venv` could be located anywhere on the path above you, and the lowermost one would be taken as the project root. But the semantics of using the git repo's root dir basename as the default `$ENV_NAME` were removed. It would *only* recognize a project if it was explicitly customized with a `.venv` dir. It's sounding like that was unintentional, and the git repo root should still be taken as the project root, in the absence of a `.venv`.

(That test for the Git behavior I did only tested that it treated symlinks the same way `:A` does. It didn't test for the old semantics of "a git repo defines a project root".)

I can restore that. The implementation is easy enough. But I need to know how the two are supposed to interact. It's not documented in the code or PR, and I don't know virtualenv/virtualenvwrapper conventions well enough to answer that myself.
* It seems clear that a `.venv` at or below the git repo root takes precedence, and defines the `$ENV_NAME`.
* What about a `.venv` located *above* a git repo root which doesn't have its own `.venv`? Should the git repo basename be used? Or should the `.venv` higher up in the path take precedence? To me, it seems like the Git repo should take precedence, but I don't understand the workflow intentions behind #3918, so maybe that's not the case.
* @davidetruffo: I'm assuming in your example that those projects are git repos, right?  That if you're not under a Git repo or a dir with `.venv`, then you are not in a virtualenv and it should be deactivated.

benjaoming:
Tested that this works as intended, great and conventional fix! I would recommend merging this as it solves the issue in case and try moving the discussion regarding #3918 to separate issue.

It's already plenty difficult to catch up with the issue and the PRs floating around :)

apjanke:
Yeah, that makes sense to me. (There's enough out there that I missed a PR on this myself.) Let's clear up this immediate `realpath` breakage with this, and resolve the "git repo base dir vs explcit `.venv` only" problem under a different PR. They are distinct questions, and it doesn't affect the correctness of this PR for the "`realpath` dependency" issue.

I'm :+1: for merging this now, and we can open a separate issue to discuss the git vs `.venv` semantics.

davidetruffo:
@apjanke yes, they are git repos (in my example: project_1, project_2 are git repos). My bad, i gave it for granted but it's not that granted :)

mcornella:
Ok, I've proposed this PR to Robby. Hopefully this one and #4041 will be merged during the day.

cebor:
:+1:

apjanke:
Looks like some other people are expecting the old behavior of "treat git repos without `.venv` as project roots", too: #4110. I'll put together a separate PR for that.

apjanke:
I put in a follow-up fix for treating Git repos as project roots: #4122.
@davidetruffo, that should fix the problem with your Git projects that you described. Could you try that one out, and if it works, give a +1 in the comments on #4122?

davidetruffo:
@apjanke sorry for the late reply!
#4122 works like a charm with my example (so it activates/deactivates the right virtualenv when you cd in/out from a git repo)

IanLee1521:
:+1: This fixes #4030 for me.

freeyoung:
+1, please merge.

tklovett:
:+1: This PR fixes the `workon_cwd:6: command not found: realpath` infinite loop for me, please merge.

PR #4111, opened 2015-07-01
New alias zwip that launch cucumber with the :wip profile

CharlonTank:


CharlonTank:
Why this is not added to oh_my_zsh ? I can get it :/

PR #4113, opened 2015-07-01
gnzh theme: fix "eval" and related scope problems.

### LINKED ISSUE #3425: Error message when logging in from a tty
I have a working oh-my-zsh setup in my Arch Linux machine (default config, I only added some aliases for pacman and other programs), but every time I log in from a **tty** I get the following error message:

image

apjanke:
Fixes #3425.

This clears up issues related to `eval` and the use of eager vs lazy evaulation in the `gnzh` theme. The `eval` was the one causing user-visible problems, but there's a small tangle of related issues, and this addresses them all.

The outcome of this PR is that:
* Prompt effect escapes are all lazy-evaluated at prompt time, using `%F`/`%f`
* Variables used to build the `$PROMPT` string are all eager-evaluated at theme loading time
 * Except for ones that deliberately use `$(...)` to call functions at prompt time
* Variables are all `local`-ized

This PR removes the problematic `eval` usage by switching to [normal `zsh` %F/%f prompt escapes](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Visual-effects) to avoid "eval" and the associated global variables.

There were several `local` statements that weren't having any effect, because the theme is not loaded in a function, so they were globals. This PR wraps whole thing in an anonymous function so the existing `local` statements actually work. 

But that breaks the theme, because a lot of stuff was done with lazy-evaluated `'...'` quotes, which relied on those actually being persistent global variables. So this PR switches '...' to eager "..." where needed, so stuff actually works with local variables. 

And it `local`-izes the remaining variables that are used only in prompt construction.

I've tested this on the following, exercising the git, return status, and root detection features, and it looks good.
* Mac OS X 10.9.5 with iTerm2 and Terminal.app
* Debian 7 using Konsole, `xterm`, and `urxvt`
* Cygwin/Windows 7 using mintty
Could use some more testers for `urxvt` and the console on Linux distros that were having this problem, and for RVM.

### DIFF
--- a/themes/gnzh.zsh-theme
+++ b/themes/gnzh.zsh-theme
@@ -1,53 +1,50 @@
 # ZSH Theme - Preview: http://dl.dropbox.com/u/4109351/pics/gnzh-zsh-theme.png
 # Based on bira theme
 
-# load some modules
-autoload -U zsh/terminfo # Used in the colour alias below
 setopt prompt_subst
 
-# make some aliases for the colours: (could use normal escape sequences too)
-for color in RED GREEN YELLOW BLUE MAGENTA CYAN WHITE; do
-  eval PR_$color='%{$fg[${(L)color}]%}'
-done
-eval PR_NO_COLOR="%{$terminfo[sgr0]%}"
-eval PR_BOLD="%{$terminfo[bold]%}"
+() {
+
+local PR_USER PR_USER_OP PR_PROMPT PR_HOST
 
 # Check the UID
 if [[ $UID -ne 0 ]]; then # normal user
-  eval PR_USER='${PR_GREEN}%n${PR_NO_COLOR}'
-  eval PR_USER_OP='${PR_GREEN}%#${PR_NO_COLOR}'
-  local PR_PROMPT='$PR_NO_COLOR➤ $PR_NO_COLOR'
+  PR_USER='%F{green}%n%f'
+  PR_USER_OP='%F{green}%#%f'
+  PR_PROMPT='%f➤ %f'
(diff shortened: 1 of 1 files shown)


mcornella:
:+1: /cc @robbyrussell

maehne:
I tried out this pull request to fix issue https://github.com/robbyrussell/oh-my-zsh/issues/3425 on Debian Jessie. It fixes indeed the eval error message when logging directly into a text console. However, the fix has the side effect that now the arrow part on the second line and the command you enter after it is displayed in bright white (and in KDE Konsole additionally with a thicker font). The start of the arrow on the first line is displayed in the normal gray tone of the text console.

apjanke:
Looks like an unclosed `%B` in the directory display was leaking bold text in to the rest of the prompt. `%f` won't reset `%B` like the old $PR_NO_COLOR would reset bold; it needs to be closed separately with a `%b`. 

Some terminals are set to display bold text in a bright color, which is why it looked brighter on your terminal. I didn't notice this while testing because I have the bold=bright setting off in my terminals, and bold doesn't look all that different from normal text to me. But it's clearly visible when toggling the bold font on and off in the profile.

I've amended the commit to fix this. Give it another try.

maehne:
Thank you for your quick reaction to my report! I just tested your fix and now the gnzh theme works nicely without any error messages or coloring problems. :+1:

apjanke:
Great. I think this is ready to merge.

mcornella:
:+1: /cc @robbyrussell

PR #4115, opened 2015-07-02
Change upstream to --set-upstream

juanpastas:


apjanke:
This is a significant change to the semantics of `gpu` (which was added in #2790). What's the reasoning behind this?

CC @ncanceill, since you were the one to add the current `gpu` alias.

juanpastas:
`git push upstream` is to push to a remote called upstream, right? is not a special name. If that is the case, I thought a lot of people do not have a remote with upstream name.

So it made more sense to me having `--set-upstream`, and if you want to push setting the upstream you do `gpu origin master` for example. It would be even better to have `gpu` and `gpup`, `gpu` for `git push -u origin master` and `gpup` for `git push -u`.

`-u` is the same as `--set-upstream`.

Am I making sense?

apjanke:
That explanation makes sense, but the original author probably knew what they were doing and wanted the current behavior. (E.g. in OMZ's workflow, it's common for some people to have an `upstream` remote. So it's not a special name to `git` itself, but has meaning for developers using a certain workflow.) 

If you want to change behavior like that, please open an issue to discuss it and contact the original author; don't just put in a PR to alter it, especially without explanation. Names of commands are part of peoples' workflow, and changing the behavior of established commands can break things.

If you want to alter the aliases just for yourself, you can redefine them in your `~/.zshrc` after OMZ has been initialized.

juanpastas:
get it, thanks.

PR #4121, opened 2015-07-03
chucknorris: check for fortune/strfile dependency

### LINKED ISSUE #4112: chucknorris plugin depends on fortune
chucknorris plugin should also have dependency on fortune.

 Story everytime I setup omz on a new Mac:
* I have to google "command not found: strfile", 
* Land on some page, 
* Realise I need *fortune* package

If not more, then at least issue might help somebody google the similar error :)

### LINKED ISSUE #4334: Chucknorris strfile
Hi,

when i add the chucknorris plugin i get this

/.oh-my-zsh/plugins/chucknorris/chucknorris.plugin.zsh:2: command not found: strfile

See you

apjanke:
Fixes #4112
Fixes #4334
Depends on #3611 

This plugin depends on fortune and strfile, which is part of fortune. But that's not immediately obvious to many users, per #4112.

This PR has chucknorris check for `strfile` when it needs it, and if not found, issue a more helpful error message that suggests a "fortune" package, and doesn't attempt to use strfile to update the fortunes dir. Still puts the aliases in place, since they'll produce more obvious error messages if used when `fortune` is not installed.

Also handles Cygwin installing `strfile` to `/usr/sbin`, which is not on the path by default, by directly checking for `/usr/sbin/strfile` if the plain `strfile` command is not found.

Adds an anonymous function for hygiene, to avoid clobbering the global `$DIR` variable, and uses a less-likely-to-collide `$CHUCKNORRIS_PLUGIN_DIR` temporary global for where it needs to access `$0` outside the anonymous function.

CC @maff, since you're the original chucknorris plugin author.

### DIFF
--- a/plugins/chucknorris/chucknorris.plugin.zsh
+++ b/plugins/chucknorris/chucknorris.plugin.zsh
@@ -1,6 +1,12 @@
-if [ ! -f $ZSH/plugins/chucknorris/fortunes/chucknorris.dat ]; then
-    strfile $ZSH/plugins/chucknorris/fortunes/chucknorris $ZSH/plugins/chucknorris/fortunes/chucknorris.dat
-fi
+() {
+  # Automatically generate or update Chuck's compiled fortune data file
+  local fdir=$ZSH/plugins/chucknorris/fortunes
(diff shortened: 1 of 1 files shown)


maff:
Looks good to me. @apjanke I've added you to the maintainers list for this plugin on the wiki as you've done way more work on this than I did with my original hack ;)

apjanke:
Thanks! :)

PR #4122, opened 2015-07-03
virtualenvwrapper: Treat git repos as project roots again

### LINKED ISSUE #4110: virtualenvwrapper not activatin the venv according to directory
After one of the updates virtualenvwrapper stopped to activate env via directory name.
Testcase:
Clean config from fresh install of oh my zsh from git
In plugins only virtualenvwrapper
```
cd Projects && mkdir new_proj && cd new_proj && git init && mkvirtualenv -p /usr/bin/python2 new_proj
cd ..
cd new_proj
```
Expecting:
Virtuale environment "new_proj" should be activated
Actual result - there is even no this tiny delay I used to had before when CDing into directories and no venv activated.

```
λ echo $chpwd_functions
autojump_chpwd workon_cwd steeef_chpwd
λ echo $ZSH_VERSION 
5.0.8
```

apjanke:
Makes `virtualenvwrapper` see git repos as project roots again, which was lost in the #3918 change.

Fixes #4110
Closes #4070 
Depends on #4106 

###  Details

In #3918, the virtualenvwrapper plugin logic was changed so that `.venv` dirs at any point created a virtualenvwrapper project root, but Git repos without a custom `.venv` were no longer treated as project roots. This was unexpected by some users, who are using non-customized Git repos as project roots. (See #4110, #4070, discussion in #4106, discussion in #3918).

This PR restores the old behavior of treating any Git repo as a project root, in addition to #3918's use of `.venv` as defining a project root. Combined with #3918, this means that the lowermost Git repo or customized `.venv` project will be treated as the project root.

There appear to be some other issues in detection of the venv name and activating venvs, but this PR does not address them. (E.g. invalid venv names are silently ignored and result in th previous venv being left activated.) This PR only undoes the loss of git repo functionality from #3918.

### Implementation

This uses `git rev-parse --show-toplevel` instead of `[[ -d $DIR/.git ]]` to detect git repos, because it is more robust. The presence of a `.git` subdir alone doesn't mean it's an actual git repo. (Try it yourself: just `mkdir .git` and see if it works as a repo.) And the `git` executable should be the source of truth about what is a git repo.

Degrades such that if `git` is not installed, then git repos will be silently ignored and not treated as project roots.

It seems like there would be an inconsistency between walking up the absolute path vs asking `git rev-parse show-toplevel` what the repo root is, in the case where you are inside a repo which has a symlink to a directory outside the repo, and you have cd'ed down through that directory. However, I tested it, and `git rev-parse show-toplevel` works the same way: it considers your absolute path. I'm not sure if this is the intended behavior for `git` wrt symlinks outside a repo, but we are consistent with it.

[Here's a test script](https://gist.github.com/apjanke/275ee69f3749ee157da5) that demonstrates the `--show-toplevel` behavior. On OS X 10.9.5, I get the following, which is consistent with this PR's code. It looks like `git` evaluates symlinks before locating the top of the repo. (I have Homebrew installed, so `/usr/local` is itself a Git repo.)

```
$ zsh ./test_rev_parse.zsh
/Users/janke/tmp/a_git_repo/a_dir/link_to_local_bin
show-toplevel: /usr/local
/Users/janke/tmp/a_git_repo/a_dir/under_tmp
fatal: Not a git repository (or any of the parent directories): .git
show-toplevel:
```

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -14,32 +14,38 @@ elif [[ -f "/etc/bash_completion.d/virtualenvwrapper" ]]; then
     source "/etc/bash_completion.d/virtualenvwrapper"
   }
 else
-  print "zsh virtualenvwrapper plugin: Cannot find ${virtualenvwrapper}.\n"\
+  print "[oh-my-zsh] virtualenvwrapper plugin: Cannot find ${virtualenvwrapper}.\n"\
         "Please install with \`pip install virtualenvwrapper\`" >&2
   return
 fi
 if ! type workon &>/dev/null; then
-  print "zsh virtualenvwrapper plugin: shell function 'workon' not defined.\n"\
+  print "[oh-my-zsh] virtualenvwrapper plugin: shell function 'workon' not defined.\n"\
         "Please check ${virtualenvwrapper}" >&2
   return
 fi
 
 if [[ "$WORKON_HOME" == "" ]]; then
-  print "\$WORKON_HOME is not defined so ZSH plugin virtualenvwrapper will not work" >&2
+  print "[oh-my-zsh] \$WORKON_HOME is not defined so plugin virtualenvwrapper will not work" >&2
   return
 fi
 
 if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
-  # Automatically activate Git projects' virtual environments based on the
+  # Automatically activate Git projects or other customized virtualenvwrapper projects based on the
   # directory name of the project. Virtual environment name can be overridden
-  # by placing a .venv file in the project root with a virtualenv name in it
+  # by placing a .venv file in the project root with a virtualenv name in it.
   function workon_cwd {
-    if [ ! $WORKON_CWD ]; then
-      WORKON_CWD=1
+    if [[ -z "$WORKON_CWD" ]]; then
+      local WORKON_CWD=1
       # Check if this is a Git repo
+      local GIT_REPO_ROOT=""
+      local GIT_TOPLEVEL=$(git rev-parse --show-toplevel 2> /dev/null)
+      if [[ $? == 0 ]]; then
+        GIT_REPO_ROOT="$GIT_TOPLEVEL"
+      fi
       # Get absolute path, resolving symlinks
-      PROJECT_ROOT="${PWD:A}"
-      while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" ]]; do
+      local PROJECT_ROOT="${PWD:A}"
+      while [[ "$PROJECT_ROOT" != "/" && ! -e "$PROJECT_ROOT/.venv" \
+               && ! -d "$PROJECT_ROOT/.git"  && "$PROJECT_ROOT" != "$GIT_REPO_ROOT" ]]; do
         PROJECT_ROOT="${PROJECT_ROOT:h}"
       done
       if [[ "$PROJECT_ROOT" == "/" ]]; then
@@ -47,7 +53,7 @@ if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
       fi
       # Check for virtualenv name override
       if [[ -f "$PROJECT_ROOT/.venv" ]]; then
-        ENV_NAME=`cat "$PROJECT_ROOT/.venv"`
+        ENV_NAME=`cat "$PROJECT_ROOT/.venv"`        
       elif [[ -f "$PROJECT_ROOT/.venv/bin/activate" ]];then
         ENV_NAME="$PROJECT_ROOT/.venv"
       elif [[ "$PROJECT_ROOT" != "." ]]; then
@@ -69,8 +75,6 @@ if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
         # Note: this only happens if the virtualenv was activated automatically
         deactivate && unset CD_VIRTUAL_ENV
       fi
-      unset PROJECT_ROOT
-      unset WORKON_CWD
     fi
   }
 
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -30,16 +30,17 @@ if [[ "$WORKON_HOME" == "" ]]; then
 fi
 
 if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
-  # Automatically activate Git projects's virtual environments based on the
+  # Automatically activate Git projects' virtual environments based on the
   # directory name of the project. Virtual environment name can be overridden
   # by placing a .venv file in the project root with a virtualenv name in it
   function workon_cwd {
     if [ ! $WORKON_CWD ]; then
       WORKON_CWD=1
       # Check if this is a Git repo
-      PROJECT_ROOT=`pwd`
+      # Get absolute path, resolving symlinks
+      PROJECT_ROOT="${PWD:A}"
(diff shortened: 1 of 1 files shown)


davidetruffo:
lovely +1

mcornella:
Very nice, but my suggestion of using a temporary variable instead of PROJECT_ROOT (https://github.com/robbyrussell/oh-my-zsh/pull/4106#issuecomment-117578928) still needs to be addressed. 
Other than that, kudos!

apjanke:
I could add that in. (Sorry, I had forgot about that comment.) I wanted to do the same myself. But I don't think it's necessary for correctness, just style, so I left it out to keep this to a minimal set of changes, and not mess with the original author's style too much.

I think it is still entering the correct branch in the `if` statement. When it temporarily sets `$PROJECT_ROOT` to be `.`, it's not actually using it to mean "this directory". It's just a placeholder to indicate "no project root found". The [pre-#3918 logic worked the same way](https://github.com/cbazin/oh-my-zsh/blob/732c7b8070a75ac5ad5384b4d22d19f499e8424f/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh#L41-L44).

I think that's the intended behavior, with the one exception that this means you can't put a `.venv` under root at `/.venv` to make a default venv wrapper for the entire filesystem. But re-using the `$PROJECT_ROOT` as a temporary and `.` as a placeholder that way makes it confusing and hard to read.

Similar issue with `ENV_NAME="$PROJECT_ROOT/.venv"`. When that's done, `$ENV_NAME` isn't holding a name, it's holding a path. But then later on, there's a corresponding `elif [[ -e "$ENV_NAME/bin/activate" ]]` which treats it as a path instead of a name, too. So it eventually works out, though it's hard to read.

If it were my code, I'd like to substantially refactor the logic to use additional temporary variables and keep it so the values matched their names and they didn't use `.` for what looked like a path but really isn't and so on. But that's not needed to fix the problem this PR is addressing, and it was the case pre-#3918, and it's not my plugin. So I think maybe that should go in a separate PR.

What do you folks think? Keep this PR simple, or go for a bigger refactor?

apjanke:
@laithalissa: Heheh. You're more observant than I am. I overlooked 4074 when I made 4106 in the first place. I wish Git had more of a real bug tracking system where it was easy to see pending fixes, duplications, and other relationships between bugs.

cebor:
yeah pls merge :+1:

cldershem:
This fixed the issue for me.  Thank you.

oryband:
:+1: please merge. i am using virtualenvs with the same names as my git repos and have been experiencing this implicit "feature" for a long time now.

osleg:
Works for me with both .venv and bare .git projects

guillaumevincent:
hi @apjanke,

your pull request doesn't work with python3 virtualenv, any idea why ?

```
 > ls .virtualenvs
postactivate        postrmvirtualenv    premkvirtualenv 
akemail             get_env_details     postdeactivate      preactivate         prermvirtualenv     initialize          postmkproject       predeactivate       rangevoting
okiwi               postmkvirtualenv    premkproject        rtfdoc
 > jump akemail
(akemail) akemail (master)
 > jump rtfdoc
 rtfdoc 
 > workon rtfdoc
(rtfdoc) rtfdoc
```

virtualenv command :
`
mkvirtualenv --python=/usr/local/bin/python3 rtfdoc
`
well done for the fix for python2.7

osleg:
Hm, for me it never worked with jump, it should?

guillaumevincent:
@osleg jump is a wrapper to cd I guess

but change directory (cd) doesn't work either

```
 > cd workspace/oslab/rtfdoc
 rtfdoc
> cd
> cd workspace/akemail/akemail
(akemail) akemail (master)
```

apjanke:
I don't know why it's not working for Python 3 for you. (You're referring to [the new `venv` support that's part of Python 3 itself](https://www.python.org/dev/peps/pep-0405/), right?)

Did `virtualenvwrapper` used to work for you pre-#3918, that is, as of d0b690b584fa48946cfd93014aa6033ec0cc7fc3?

Your `rtfdoc` directory: Does it have a `.venv` file or subdirectory? How is the name of the virtualenv being detected there?

What is your `$WORKON_HOME`?

guillaumevincent:
@apjanke akemail is in `workspace/akemail/akemail` or rtfdoc in `workspace/oslab/rtfdoc`don't have a `.venv`file or subdirectory in any ot them. My `$WORKON_HOME = .virtualenvs`as listed in my previous post

`akemail` is a python 2.7 and `rtfdoc` is a python 3.4

I don't use `venv` to create my virtualenv python 3.4 because I use virtualenvwrapper and the  `mkvirtualenv`command (see my previous post)

I thought that virtualenvwrapper activate virtualenvs automatically when cd or jump in directory with the same virtualenv's name. It works for my python 2.7 interpreters (all of them) not for my any python 3.4 interpreters

apjanke:
Do you mean `$WORKON_HOME` is literally `.virtualenvs`, or is it `~/.virtualenvs` or some other absolute path? Using a relative path in `$WORKON_HOME` will probably cause problems.

Virtualenvwrapper won't activate a venv for just any directory based on the name. It will activate it for:

* A git repo, based on the name of the root directory of the repo
* Any directory tree with a `.venv` file or directory defining the name of a virtualenv

Unless your `rtfdoc` directory is a Git repo or contains a `.venv`, then `virtualenvwrapper` won't recognize it as a project root, and won't try to activate a venv for it. This is normal `virtualenvwrapper` behavior.

It appears that your `akemail` directory is a Git repo (because it says "master" in the prompt), and that's probably why `virtualenvwrapper` is recognizing it. Maybe it's a coincidence that it's working for your 2.7 stuff and not the 3.x stuff, because your 2.7-related project directories happen to have Git repos initialized in them?

guillaumevincent:
@apjanke my `$WORKON_HOME` is `~/.virtualenvs`

sorry for the extra work, I added a git repository, and it works as expected !

Thank you for the clarification

:+1: 

The pull 4122 works for me too

apjanke:
Hi @robbyrussell, @mcornella: Could we get this merged? It's been tested, and some users are waiting on it to fix a bug in the current version.

KenMercusLai:
My change contains 2 main logical modification according to previous commit.

the logic is shown as below:


* It uses the current folder name is the virutalenv name when .venv file is not found recursively. As we cannot assume all virtualenv has .venv file, I think use the folder name as the virtualenv name would be better, which is also behaviour when creating a virtualenv.
* there is no need to check ENV_NAME is empty or not because it will never be empty after previous change.
* there are subfolders in virtualenv root folder and the environment should not be deactivated when entering them. So the second main change was checking whether current folder belongs to the root folder of activated environment.

```bash
       elif [[ `pwd` != "$CD_VIRTUAL_ENV"* ]]; then
          # We've just left the repo, deactivate the environment
          # Note: this only happens if the virtualenv was activated automatically
          deactivate && unset CD_VIRTUAL_ENV
        fi
```

I have posted my idea on my blog and tried to explain this idea, you can check it out here:
http://kenmlai.me/better-process-for-zsh-virtualenvwrapper-plugin/

sorry for my clumsy English. :disappointed:

apjanke:
I'm going to take a bit to read and digest this. Thank you for documenting this well! It's a lot easier to understand a code change when there's human-readable doco and not just code.

For other readers: this is wrt the proposed change in #4178. I asked that it be run by the `virtualenvwrapper` community for discussion before merging.

PavloKovalov:
@apjanke :+1:  works well
```
pavel at macbook-pro in ~/Projects/Work 
$ echo $ZSH_VERSION 
5.0.8
```
```
pavel at macbook-pro in ~/Projects/Work 
$ echo $chpwd_functions
workon_cwd steeef_chpwd
```
Tested on my OSX 10.10.4 and Ubuntu 14.04.2

PR #4128, opened 2015-07-05
svn: decode URLs before display, to avoid breaking prompt

### LINKED ISSUE #2233: SVN Plugin messes up Prompt when URL contains Spaces

apjanke:
Subversion repos are identified by URLs, which may be %-escaped if they have reserved characters in them. Those %s will be interpreted as prompt escapes if an escaped URL fragment is included directly in the prompt. This PR has it decode the URLs before using them in the prompt.

Fixes #2233.

### DIFF
--- a/plugins/svn/svn.plugin.zsh
+++ b/plugins/svn/svn.plugin.zsh
@@ -1,17 +1,65 @@
 # vim:ft=zsh ts=2 sw=2 sts=2
 #
 function svn_prompt_info() {
+  local _DISPLAY
   if in_svn; then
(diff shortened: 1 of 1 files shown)


mcornella:
Ok, this also needs to be incorporated into #4129. I'm not sure what the merge procedure will be but we should coordinate those 3 PRs to get it right.

apjanke:
I can just fold this fix into #4129. That'll make the merge simpler, and people will be able to review the full set of changes by viewing a single branch diff.

apjanke:
Folded this in to #4129. Closing in favor of #4129.

Issue #3482, opened 2015-01-10
Fedora support for command-not-found plugin


ai:
Fedora 21 has command not found behaviour for bash. But this plugin didn’t work in Fedora.

/cc @toctan @lorrin @sargas

PR #4129, opened 2015-07-05
Add URL encode/decode functions to core lib

### LINKED ISSUE #2233: SVN Plugin messes up Prompt when URL contains Spaces

apjanke:
OMZ isn't a web-oriented tool per se, but there are several places where URLs are used in it: reporting pwd to terminal applications, working with `svn` repo URLs, and so on. URL encoding and decoding of strings comes in to play there. This is nontrivial, so it would be nice to have functions for it available in the core library, so plugins and other code can use them. Could be useful for users building their own scripts and working on the command line, too, since a lot of development work is web stuff.

This PR adds `omz_urlencode()` and `omz_urldecode()` functions to `lib.functions.zsh`. Their implementations are pure native `zsh`, for performance and consistency across platforms. (Well, except for a call to `iconv` when needed for non-UTF-8 locales.) They provide:
* A pretty full implementation of RFC 2396, handling all the character classes
* Handling of locale's character set
* Safe with respect to literal '\' and '%' in encoded strings
* Options to encode "mark" and "reserved" characters a %-escapes
* Minimal subshell calls, to avoid performance overhead, especially on Windows/Cygwin
* No external dependencies except `iconv` for non-UTF-8

If this PR is accepted, then the single-use urlencode and urldecode functions in pending PRs #3582 and #4128 can be changed to use these central implementations. I'm not making that switch part of this PR because I don't want to intertwine them until it's determined which PRs will be accepted.

There is an existing `urlencode`/`urldecode` implementation in the `urltools` plugin. But that's not suitable for use by core OMZ code or other plugins, because it would introduce core -> plugin or plugin -> plugin dependencies. They don't have options to control encoding of mark or reserved characters. Plus, those implementations rely on a variety of external components, and some of the definitions appear to be incomplete or inconsistent with each other, and don't have locale handling. Having a single native `zsh` implementation would ensure the same behavior on all systems and make things easier to debug.

Fixes #2233

UPDATE: These same functions are now included in #3582. If it is merged, this PR is no longer needed.

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -73,3 +73,134 @@ function env_default() {
     env | grep -q "^$1=" && return 0 
     export "$1=$2"       && return 3
 }
+
+# Required for $langinfo
+zmodload zsh/langinfo
+
+# URL-encode a string
+#
+# Encodes a string using RFC 2396 URL-encoding (%-escaped).
+# See: https://www.ietf.org/rfc/rfc2396.txt
+#
+# By default, reserved characters and unreserved "mark" characters are
+# not escaped by this function. This allows the common usage of passing
+# an entire URL in, and encoding just special characters in it, with 
+# the expectation that reserved and mark characters are used appropriately.
+# The -r and -m options turn on escaping of the reserved and mark characters,
+# respectively, which allows arbitrary strings to be fully escaped for
+# embedding inside URLs, where reserved characters might be misinterpreted.
+#
+# Prints the encoded string on stdout.
+# Returns nonzero if encoding failed.
+#
+# Usage:
+#  omz_urlencode [-r] [-m] <string>
+#  
+#    -r causes reserved characters (;/?:@&=+$,) to be escaped
+#
+#    -m causes "mark" characters (_.!~*''()-) to be escaped
+#
+#    -P causes spaces to be encoded as '%20' instead of '+'
+function omz_urlencode() {
+  emulate -L zsh
+  zparseopts -D -E -a opts r m P
+
+  local in_str=$1
+  local url_str=""
+  local spaces_as_plus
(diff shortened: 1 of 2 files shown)


mcornella:
What's the difference between these versions and ones in #3582 ?

apjanke:
They serve the same purpose, but have several small improvements in implementation and defensive coding now that I know the related features a bit better. 

* #3482 only had encode; this one adds a decode
* Options (`-r`, `-m`, `-P`) to allow caller to control extent of escaping
* Better coverage of different types of characters that need to be escaped, so it can now handle escaping of arbitrary strings for embedding inside URLs safely. (But this is an optional behavior, so the casual "Escape this whole URL string" case still works.)
* Adds documentation
* More thorough `local`-ization of variables
* Avoids subshells (by using in-shell arithmetic operations) for performance, especially on Windows
* Uses local options to defend against shell state
 * (though this should actually go a step further and do `emulate -L zsh`; I'll fix that
* Protects against `LC_ALL` (which would defeat the `local LC_CTYPE` in the old implementation
* Works with US-ASCII locales
* Uses `$langinfo` instead of trying to parse the locale name itself, for more robust encoding detection

And they're good enough now that I think they're suitable for reuse, so they're now named `omz_...` instead of `_omz_...` to indicate they're not "private use" functions.

The implementation here is better, and I should probably just copy it into #3582 so it still has the better code, if it gets merged on its own. (Though I'd rather save work by getting both of them merged, and then just changing the stuff from #3582 to call these functions.)

apjanke:
Okay, I amended it to use `emulate -L zsh`.

apjanke:
Closing this in favor of #3582, which includes the exact same urlencode/urldecode functions. If the larger terminal-related changes from #3582 get rejected, then we can re-open this to get just the urlencoding functions in.

PR #4132, opened 2015-07-06
Adding ability to add JIRA_PREFIX as an env variable (e.g. in .zshrc)…

pmalek:
… and minor refactor to bash curly variable braces

### DIFF
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -57,9 +57,9 @@ open_jira_issue () {
     fi
 
     if [[ "$JIRA_RAPID_BOARD" == "true" ]]; then
-      $open_cmd "$jira_url/issues/$jira_prefix$1$addcomment"
+      $open_cmd "${jira_url}/issues/${jira_prefix}${1}${addcomment}"
     else
-      $open_cmd "$jira_url/browse/$jira_prefix$1$addcomment"
+      $open_cmd "${jira_url}/browse/${jira_prefix}${1}${addcomment}"
     fi
   fi
 }
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -76,24 +76,24 @@ jira_name () {
 }
 
 jira_query () {
-    verb="$1"
-    if [[ "${verb}" = "reported" ]]; then
-      lookup=reporter
-      preposition=by
-    elif [[ "${verb}" = "assigned" ]]; then
-      lookup=assignee
-      preposition=to
-    else
-      echo "not a valid lookup $verb"
-      return 1
-    fi
-    shift 1
-    jira_name $@
-    if [[ $? = 1 ]]; then
-        return 1
-    fi
-    echo "Browsing issues ${verb} ${preposition} ${jira_name}"
-    $open_cmd "${jira_url}/secure/IssueNavigator.jspa?reset=true&jqlQuery=${lookup}+%3D+%22${jira_name}%22+AND+resolution+%3D+unresolved+ORDER+BY+priority+DESC%2C+created+ASC"
+  verb="$1"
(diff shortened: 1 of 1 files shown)


plugins/jira/jira.plugin.zsh
```diff
@@ -51,7 +54,7 @@ open_jira_issue () {
       echo "Opening issue #$1"
     fi
     
-    if [[ "x$JIRA_RAPID_BOARD" = "xtrue" ]]; then
+    if [[ "$JIRA_RAPID_BOARD" = "true" ]]; then
```
apjanke: Could use `==` instead of `=` here for more zsh-like style.

pmalek: I will do that when I'm back home.

pmalek: Done. Can we merge:)?

apjanke:
While we're in here: the functions are a little inconsistent about using `local`. The variables `jira_url`, `jira_prefix`, `verb`, `lookup`, `preposition`, and `jira_name` look like they are for this plugin's internal use and could be made local. Is that the case? (`jira_name` would need to be `local`-ized in `jira_query()` and not `jira_name()`)

pmalek:
Can you take a look? I am not a bash expert in bash so there might be some error although I have tested it and seems to work fine with all the functionalities.

apjanke:
This is `zsh`, not `bash`, so there may be a couple minor differences.

The `local`s are looking good now, based on a read-through of the code. I'll run some tests to verify it when I have a chance later today.

For brevity, you can put multiple variables on one `local` line. This:
```
  local jira_name
  local verb
  verb="$1"
  local lookup
  local preposition
```
can just be this:
```
  local verb="$1"
  local jira_name lookup preposition
```
(You could even stick them all on a single line, but I like to separate out function arguments from internal variables, just for readability.)

mcornella:
Ok, is this ready to be merged?

pmalek:
I think yes. @apjanke do you agree?

apjanke:
I haven't tested it, but the changes look good based on inspecting the code.

Since we've touched a lot of lines for style & refactoring, how about looking at two other minor changes while we're in here? (No need to if you don't want to; these are just additional improvements that I noticed while reviewing the changes, and the current changes look good by me.)

The [`open_jira_issue()` function uses `$1` and `$2` directly](https://github.com/pmalek/oh-my-zsh/blob/ad4675cb0a8f732abef7b9a2b3df2141afbf79aa/plugins/jira/jira.plugin.zsh#L45-L64) throughout the function body instead of assigning them to readable local variables like `jira_query()` does with `$verb`. How about doing so here too, for readability? And documenting the second argument and special `"assigned"` and `"reported"` calling forms in the function's comments?

Minor nitpick: the return value of `jira_name()` is [tested here with ` = 1` to detect errors](https://github.com/pmalek/oh-my-zsh/blob/ad4675cb0a8f732abef7b9a2b3df2141afbf79aa/plugins/jira/jira.plugin.zsh#L94-L97). But conventionally, 0 is used for success and any nonzero value an error. While we know that this particular implementation can (I think) only return 1 or 0, `if [[ $? != 0 ]]` might be a more readable and robust way of checking for errors, especially if `jira_name()` grows in the future.

Might also be nice to add an explicit reserved `"new"` keyword for opening a new issue, and having the zero-arg form default to the `"new"` action, instead of requiring `""` for it, to make it easier to call from other code with fewer readability or quoting issues. But that may be going beyond the scope of this PR.

apjanke:
Since we had a couple other outstanding JIRA PRs that were related and had conflicts, I went and pulled this and the others' commits into a combined PR #4259, which also addresses my last comment, and adds completion support. I'd say we merge that one instead, since it has the manual conflict resolution taken care of.

pmalek:
Ok cool :)

PR #4135, opened 2015-07-07
rails rake test

skv-headless:
http://guides.rubyonrails.org/testing.html#running-tests


mcornella:
:+1: /cc @robbyrussell

PR #4136, opened 2015-07-08
new module for node: auto-set PATH on prompt

mightyiam:


plugins/node_modules-bin/node_modules-bin.plugin.zsh
```diff
@@ -0,0 +1,33 @@
...
+    # remove old bin path from $path
+    path[$path[(i)$OLD_NPM_BIN_PATH]]=()
+  fi
+
+  # add new bin path to $path
+  path=($NPM_BIN_PATH $path)
+  export PATH
+
+  # for next run
+  export OLD_NPM_BIN_PATH=$NPM_BIN_PATH
```
mightyiam: I'm not sure what the policy is in this project regarding exporting stuff. Perhaps there's a better way to do it, which doesn't involve exporting?

mightyiam:
This is sort of my first shell script.

Also, since this is small, I didn't bother reading the style guidelines, but I looked at an existing plugin's code.

plugins/node_modules-bin/node_modules-bin.plugin.zsh
mightyiam: Actually, I think that exporting this means that this plugin is broke in child shells.

I think I've just experienced that.

I'd be happy to hear of another way to keep this bit of data between calls of the same shell only.

mightyiam:
After a few commits, this looks cleaner and works much faster. Fast enough for me on an old laptop.

mightyiam:
This is very wrong. Use [`npm-run`](https://www.npmjs.com/package/npm-run).

PR #4137, opened 2015-07-08
coffee: protect args with double-quote

troter:


apjanke:
Could you put the plugin name at the start of this PR's name, so it's apparent what it applies to? E.g. "coffee: protect args with double-quote".

mcornella:
:+1: /cc @robbyrussell

PR #4141, opened 2015-07-09
Add new plugin 'superb-aliases' to Oh-My-Zsh

nntoan:
Add a new plugin which contains a lot of useful aliases for the *NIX enviroment.


mcornella:
We already have the `common-aliases` plugin. Can you explain what's different on this one?

PR #4145, opened 2015-07-09
Updating Vundle.vim url and path

daviddavis:
Per the Vundle.vim README, the new Github url is https://github.com/VundleVim/Vundle.vim. Moreover, the README states that the Vundle.vim repo should be cloned to `~/.vim/bundle/Vundle.vim`. For more info, check out [the Vundle.vim README](https://github.com/VundleVim/Vundle.vim).

### DIFF
--- a/plugins/vundle/vundle.plugin.zsh
+++ b/plugins/vundle/vundle.plugin.zsh
@@ -1,13 +1,12 @@
 function vundle-init () {
-  if [ ! -d ~/.vim/bundle/vundle/ ]
-  then
-    mkdir -p ~/.vim/bundle/vundle/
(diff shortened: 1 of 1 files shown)


apjanke:
Can you do a couple style fixes while updating this?

* OMZ uses double-bracket `[[ ... ]]` tests and puts the `; then` on the same line as the `if`
* Could the "Read about..." message include a note that the user will have to edit their `~/.vimrc` to actually load Vundle and list plugins?

That aside, +1. Works for me on OS X 10.9.5, as long as I already have the Vundle commands in my `.vimrc`

daviddavis:
Done. :+1:

apjanke:
Lookin' good to me.

:+1:

PR #4148, opened 2015-07-10
Get docker ip

lrkwz:
function to get a docker container ip


mcornella:
Can we merge this into the `docker` plugin?

lrkwz:
of course

leifcr:
@lrkwz Can you see if this can be added to #3289? I do think it's useful, but it would fit better to be part of the docker plugin instead of a separate plugin.

PR #4152, opened 2015-07-12
Add new plugin js-repl for javascript REPL

deepak:
only has safari on OS X


mcornella:
Not suitable for as a whole plugin.

PR #4154, opened 2015-07-12
Esc-esc toggles sudo

sfragis:
This change improves the behaviour of the sudo plugin. By pressing esc-esc once the sudo command is added at the beginning as usual; pressing it once more removes "sudo " from the beginning of the line.


mcornella:
Works well and it's very useful when you've typed that by mistake
:+1: /cc @robbyrussell

PR #4155, opened 2015-07-13
[theme] Fix color problem of steeef theme

honnix:
steeef theme sets colors depending on $TERM, however if %F is
used, it must be closed by %f instead of %{$reset_color%}.

Further on, all %f directives have been replaced by %F and color
support of current TERM is checked more properly.

### DIFF
--- a/themes/steeef.zsh-theme
+++ b/themes/steeef.zsh-theme
@@ -10,7 +10,7 @@
 export VIRTUAL_ENV_DISABLE_PROMPT=1
 
 function virtualenv_info {
-    [ $VIRTUAL_ENV ] && echo '('$fg[blue]`basename $VIRTUAL_ENV`%{$reset_color%}') '
+    [ $VIRTUAL_ENV ] && echo '('%F{blue}`basename $VIRTUAL_ENV`%f') '
 }
 PR_GIT_UPDATE=1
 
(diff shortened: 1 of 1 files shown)


apjanke:
:+1:

I'm in favor of the switch to `%F` everywhere.

I tested this on OS X 10.9.5 under iTerm2, Terminal.app, `xterm`, and `urxvt`; looks good in all of them.

themes/steeef.zsh-theme
```diff
@@ -10,7 +10,7 @@
 export VIRTUAL_ENV_DISABLE_PROMPT=1
 
 function virtualenv_info {
-    [ $VIRTUAL_ENV ] && echo '('$fg[blue]`basename $VIRTUAL_ENV`%{$reset_color%}') '
+    [ $VIRTUAL_ENV ] && echo '('%F{blue}`basename $VIRTUAL_ENV`%f') '
```
mcornella: One other fix here: use `${VIRTUAL_ENV:h}` instead of the call to `basename`.
Also `[ $VIRTUAL_ENV ]` could be switch for `[[ -n "$VIRTUAL_ENV" ]]`.

mcornella:
With my suggestions this has my :+1:

themes/steeef.zsh-theme
honnix: Well, I agree, but shall we merge this PR first? I can fix these with a new PR because they are not really related with colors.

mcornella: Sure, let's merge this one then.

mcornella:
Good to go!
/cc @robbyrussell

PR #4156, opened 2015-07-13
Make man colors similar to Amazon and Ubuntu distributions

a-b:


plugins/colored-man/colored-man.plugin.zsh
```diff
@@ -19,12 +19,13 @@ fi
 man() {
       env \
       	  LESS_TERMCAP_mb=$(printf "\e[1;31m") \
-	  LESS_TERMCAP_md=$(printf "\e[1;31m") \
+	  LESS_TERMCAP_mb=$(printf "\e[1;31m") \
```
apjanke: Looks like this is redundantly defining LESS_TERMCAP_mb, which is also set on line 21.

a-b:
Good catch, fixed. Thank you!

mcornella:
We could use `echoti` (zsh's `tput`) or zsh's builtin `%F{color}` and `%K{color}` sequences instead of hardcoded values.
Can you post pre and post screenshots of this change?

PR #4158, opened 2015-07-13
Add jira_url_help function

andrewstuart:
This gives a bit more information on what to do if none of the JIRA url locations are present.


plugins/jira/jira.plugin.zsh
```diff
@@ -59,6 +59,16 @@ open_jira_issue () {
   fi
 }
 
+jira_url_help() {
+  cat << EOF
+JIRA url is not specified anywhere.
+Valid options, in order of preference:
```
apjanke: Should this say "precedence" instead of "preference"? This list is the order in which they're checked, not in order of which is "better", right?

plugins/jira/jira.plugin.zsh
```diff
@@ -59,6 +59,16 @@ open_jira_issue () {
...
+  .jira-url file
+  $HOME/.jira-url file
```
apjanke: Would it be clearer to escape the `$HOME` so it shows up as a literal `$HOME` in the output, instead of the current user's particular path? For example, I get this when I run it.

➜  ~  jira
JIRA url is not specified anywhere.
Valid options, in order of preference:
  .jira-url file
  /Users/janke/.jira-url file
  JIRA_URL environment variable

andrewstuart: Good catch. I forgot to run it after I made that change, so I was expecting a literal "$HOME."

plugins/jira/jira.plugin.zsh
andrewstuart: Agreed, I think "precedence" is clearer. Updated.

PR #3661, opened 2015-03-03
Fix invocation of gradlew

ericbottard:
Gradle wrapper is typically in the current directory, not on PATH.

Closes #4166

PR #4166, opened 2015-07-17
not in the path gradlew support

seblemaguer:
Hello,

as I do not put the "." in the path variable, my gradle wrapper (gradlew) cannot be run using the command "gradlew". As I assume it could affect other people, I propose this pull request to take into account the gradle wrapper of the current directory


mcornella:
UPDATE: sorry @seblemaguer, it seems #3661 was there first and (in theory) it should work the same way.

seblemaguer:
as long as the functionality is in the repository I'm ok :D

PR #4172, opened 2015-07-19
Fix little typo that makes the statement useless.

AlexTalker:
I'm sorry :smiley:


mcornella:
Oops :grin: 
/cc @robbyrussell

PR #4173, opened 2015-07-19
Devanagari Om fixed in dogenpunk theme

sharkySharks:
I removed the incorrect Devanagari conjunct symbol at the end of Om. The dash below the letter indicates that it should join the previous letter, which is not needed in spelling Om.

### DIFF
--- a/themes/dogenpunk.zsh-theme
+++ b/themes/dogenpunk.zsh-theme
@@ -3,7 +3,7 @@
 MODE_INDICATOR="%{$fg_bold[red]%}❮%{$reset_color%}%{$fg[red]%}❮❮%{$reset_color%}"
 local return_status="%{$fg[red]%}%(?..⏎)%{$reset_color%}"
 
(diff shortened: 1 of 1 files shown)


apjanke:
Nice. Works for me on OS X 10.9.

:+1:

sharkySharks:
Cool, I made the change locally since I started using the dogenpunk theme, which I really like! - thought others would benefit from change as well.

apjanke:
Cool. @dogenpunk, any objections to merging this for everyone?

dogenpunk:
None at all. In fact I'd made the same change locally and hadn't gotten around to creating the pull request. Thanks @sharkySharks!

sharkySharks:
Awesome! Not a problem, I like your theme!

sharkySharks:
Just checking in on this PR, anything else needed to get it pulled in? I'd love to share it with other folks once it is updated. Thx!

mcornella:
:+1: 
/cc @robbyrussell

PR #4178, opened 2015-07-22
virtualenvwrapper logic improved

KenMercusLai:
1. set ENV_NAME to current folder name instead of an empty string when .venv file is not found
2. optimized activate & deactivated logic


apjanke:
This seems like a substantial change to the existing behavior of `virtualenvwrapper`, which treats only git repo roots or `.venv`-customized dirs as project roots.

Conflicts with PR #4122, which fixes a regression related to treating git repo roots as project roots. Could you have a look at that discussion and comment there? At a minimum, this will probably need to be rebased after #4122 is merged. And the change in behavior should probably be put up for discussion with other `virtualenvwrapper` users.

apjanke:
Now that #4122 is merged, can you review to see if this is still needed, and if so, rebase it against the latest `master`? This will probably require some manual changes even if it is a clean rebase, since #4122 included a change to the search logic.

robbyrussell:
Closing as we never got this rebased. We appreciate the original intent, though!

PR #4181, opened 2015-07-23
Update z version to 5dc2a86

posva:
There is an old PR at #3248
That's why I'm creating a new one

Version: https://github.com/rupa/z/commit/5dc2a863ccdcefb28aaf87cc99c31c6619158ed4
Closes #3248.

### DIFF
--- a/plugins/z/README
+++ b/plugins/z/README
@@ -6,7 +6,7 @@ NAME
        z - jump around
 
 SYNOPSIS
-       z [-chlrt] [regex1 regex2 ... regexn]
+       z [-chlrtx] [regex1 regex2 ... regexn]
 
 AVAILABILITY
        bash, zsh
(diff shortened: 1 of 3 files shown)


apjanke:
The "Closes #3248" needs to be in the body of the PR description, not just the title, for it to work properly and auto-close the referenced issue.

apjanke:
This should close #4570, too.

apjanke:
What version of the upstream `z.sh` does this grab? Might be nice to note that in the plugin.

posva:
@apjanke There's unfortunately no version number :confused: 
I updated `z` again and fixed the commit

apjanke:
Even if it's not one of the [numbered releases](https://github.com/rupa/z/releases), could we identify it with the git commit ID?

PR #4186, opened 2015-07-27
Fix install.sh for non-GNU sed

Bob131:
Sorry if this seems needlessly picky, but I figure low hanging fruit is fruit all the same :)

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -23,14 +23,14 @@ fi
 
 echo "\033[0;34mUsing the Oh My Zsh template file and adding it to ~/.zshrc\033[0m"
 cp $ZSH/templates/zshrc.zsh-template ~/.zshrc
-sed -i -e "/^export ZSH=/ c\\
+cat ~/.zshrc | sed -e "/^export ZSH=/ c\\
 export ZSH=$ZSH
-" ~/.zshrc
+" > ~/.zshrc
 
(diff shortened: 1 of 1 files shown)


apjanke:
Duplicate of #3613.
Conflicts with #2589, which does basically the same thing as part of its change.
Fixes #3367
Fixes #3612

Sorry if you didn't see the #3613. That PR was closed because of the conflict with #2589. I just re-opened it now because I think #2589 is unlikely to get merged soon, and we should fix the portability issue you identified.

This implementation has an issue because it's of the form `cat file | foo > file`. Because the shell executes all parts of the pipeline in parallel, this creates a race condition because both `cat` and `foo` will attempt to open the file at the same time. If `cat` wins, this works. If `foo` wins, it'll truncate the file, and `cat` will read an open file, accidentally erasing the file's contents. This is system dependent: the failure will only be observed on some systems and configurations. To be correct, you need to output to a temporary file, and then replace the original input in a subsequent command; not in the same pipeline.

I've re-opened #3613. I think we should merge that one instead, because it doesn't have the race issue.

apjanke:
And I don't think it's "needlessly picky" at all: it's good to be portable. And this exact issue is breaking the installer on OpenBSD. Thanks for the submission!

apjanke:
P.S. Did you run in to another system that lacked `sed -i` support?

Bob131:
I can't say I did, but I'm running OpenBSD 5.7-release which is the latest release

PR #4189, opened 2015-07-27
agnoster: speed up git prompt

bmdhacks:
Problem

On detached-refs, the git prompt is very slow because it uses a command
to output all refs, which takes 32s on the twitter internal git repo.

Solution

Use git-rev-parse instead of git-show-ref to only output the head ref.
Runtime drops to 0.03 seconds.

Note

Interestingly, git-rev-parse isn't slow without the --head arguement so
I'll probably be cooking up a patch for the git team to speed this up
eventually, but it's still pointless to sigpipe git as it walks the
filesystem.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -78,7 +78,7 @@ prompt_git() {
 
   if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
     dirty=$(parse_git_dirty)
-    ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git show-ref --head -s --abbrev |head -n1 2> /dev/null)"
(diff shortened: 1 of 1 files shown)


apjanke:
Say, does this slowness on detached-refs also affect the [upstream agnoster theme](https://github.com/agnoster/agnoster-zsh-theme)? It's using `zsh`'s own `vcs_info` support instead of the external `git` command now.

bmdhacks:
Nope, upstream is fine.  I think vcs_info uses rev-parse correctly.  I'll close this in favor of that approach.

apjanke:
You might actually want to leave this open, as an interim fix. This a small change, so could be merged pretty quickly. The OMZ `agnoster` theme is not tightly tied to the upstream, and has diverged significantly. We're currently working on harmonizing them, but that's work-in-progress and will be a bigger change, so will take longer to get through the approval process, and there's no ETA on it.

mcornella:
:+1: /cc @robbyrussell

PR #4190, opened 2015-07-28
history-substring-search: add upstream updating and update to 2015_04_26

apjanke:
The `history-substring-search` plugin has an upstream repo at https://github.com/zsh-users/zsh-history-substring-search. The OMZ version is significantly behind.

This PR adds a script that can programmatically update the OMZ plugin from the upstream repo, by copying it in and adding OMZ-specific tweaks. This PR also updates the OMZ plugin with the latest upstream head version, c4a83561a12b898ff7145856d0e6ae50757e28ba (2015-04-26, current as of 2015-07-27).

Fixes #3913.

PR #4192, opened 2015-07-28
history-substring-search: bind arrows in both vi and emacs keymaps

### LINKED ISSUE #2735: history-substring-search no longer functions after latest update

apjanke:
This binds the arrow keys in both the `emacs` and `viins` keymaps so up arrow will "work" in both emacs and vi mode, and regardless of whether the `vi-mode` plugin is loaded before or after this.

Partially fixes #2735

Note that the arrow keys still won't work in vi mode if the `vi-mode` plugin is loaded until #4191 is also merged, because the widgets in `vi-mode` are preventing terminfo mappings from working.

### DIFF
--- a/plugins/history-substring-search/history-substring-search.plugin.zsh
+++ b/plugins/history-substring-search/history-substring-search.plugin.zsh
@@ -15,9 +15,13 @@ fi
 # Bind terminal-specific up and down keys
 
 if [[ -n "$terminfo[kcuu1]" ]]; then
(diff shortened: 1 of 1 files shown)


ikhthiandor:
The following bindkeys seem to work for me. As seen here. https://github.com/sorin-ionescu/prezto/issues/516

bindkey '^[[A' history-substring-search-up
bindkey '^[[B' history-substring-search-down

apjanke:
Closing in favor of #4193 since #4190 made it in.

PR #4193, opened 2015-07-28
history-substring-search: bind arrows in both vi and emacs keymaps (post-#4190 upstream update)

### LINKED ISSUE #2735: history-substring-search no longer functions after latest update

apjanke:
This is a version of #4192 that can be used if #4190 (harmonizing history-substring-search with upstream) gets merged in. It's a nontrivial merge, so I put in a separate PR, so people could see how it would work with the upstream-pulling mechanism.

Depends on #4190 (and includes its commits).
Closes #4192.
Partially fixes #2735.

### Description  ###

This binds the arrow keys in both the `emacs` and `viins` keymaps so up arrow will "work" in both emacs and vi mode, and regardless of whether the `vi-mode` plugin is loaded before or after this.

Note that the arrow keys still won't work in vi mode if the `vi-mode` plugin is loaded until #4191 is also merged, because the widgets in `vi-mode` are preventing terminfo mappings from working.

### DIFF
--- a/plugins/history-substring-search/README.markdown
+++ b/plugins/history-substring-search/README.markdown
@@ -1,7 +1,0 @@
-To activate this script, please include it the `plugins` variable within `~/.zshrc`
-
-  `plugins=(git history-substring-search)`
-
-See the "history-substring-search.zsh" file for more information:
-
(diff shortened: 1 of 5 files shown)


mcornella:
Merged. Now you can delete all those `histsubstr*` branches :joy:

PR #4199, opened 2015-07-31
fixing missing variable expansion for terminal title to be set

cherti:
changing some single quotes to double quotes, because with single quotes the variable does not get expanded and the terminal title ends up being '$LINE' instead of the actual command run. With double quotes the variable expansion takes place as desired.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -46,7 +46,7 @@ function omz_termsupport_preexec {
   local CMD=${1[(wr)^(*=*|sudo|ssh|rake|-*)]:gs/%/%%}
   local LINE="${2:gs/%/%%}"
 
-  title '$CMD' '%100>...>$LINE%<<'
+  title "$CMD" "%100>...>$LINE%<<"
 }
 
 precmd_functions+=(omz_termsupport_precmd)


mcornella:
This has never happened before, and the single quotes are there to solve escaping issues: #2341.
Can you provide a screenshot of your problem?

cherti:
Of course. Tested with urxvt and xterm, both affected. The screenshot is urxvt.

myerr

I'm currently only using the termsupport.zsh, there shouldn't be any cross dependencies from somewhere else in the oh-my-zsh-code, should there?
But still, the shell's qouting mechanism shouldn't be changed by that, I suppose.

details on software versions (all on Arch Linux):
rxvt-unicode (urxvt) v9.21
XTerm(318)
zsh 5.0.8

termsupport.zsh is used by simply sourcing it in my zshrc.

Is there further info I can provide to help shedding light on the matter?

mcornella:
Yes, please run `set -xv` to enable debug mode, then run `set +xv` to disable it immediately after. Then copy the generated output and paste it here or in a gist (remember to quote it like this:
\`\`\`
output
\`\`\`

cherti:
set -xv: https://gist.github.com/cherti/694969ffe5c55e80b721
set +xv: https://gist.github.com/cherti/97c6781a0145d5ae691c

mcornella:
This doesn't make any sense, are you sure the title was set to $LINE when you run those? 
The execution trail shows the title was set to `set+xv`. Maybe this only happens on certain commands or conditions...

mcornella:
Ok I see where your problem is:
> termsupport.zsh is used by simply sourcing it in my zshrc.

You should run the whole oh-my-zsh suite, that way it works as it should. I don't know why using only termsupport.zsh doesn't work, but that's not really the way it is intended to be run.

mcornella:
Forget what I said, I can reproduce on cygwin with `xterm-256color`. I don't know when exactly this started happening, let me dive into the code a little bit.

cherti:
can confirm the urxvt is the 256color-version here as well (not sure about the xterm). I'll just rerun `set -xv` and `set +xv` to make sure I didn't accidentally used a terminal still on my modified version.

cherti:
-xv: https://gist.github.com/cherti/a88172895fdaa052b2a9
+xv: https://gist.github.com/cherti/1608e8ee7eea18633df4

This time I can clearly confirm that the terminal title was definitely set to "$LINE" during the run of the command.

Just in case you want to take a look onto a second dataset apart from your machine. :)

apjanke:
I can reproduce this on both Cygwin and OS X, by starting up `zsh` with an empty `.zshrc` and sourcing only `lib/termsupport.zsh` and not the rest of OMZ. Running `sleep 3` results in the window title being set to `$TITLE` instead of 

> But still, the shell's qouting mechanism shouldn't be changed by that, I suppose.

It could. Zsh is very flexible and can be configured to emulate other shells and significantly change its behavior. There are some `setopt` options which affect behavior of both quoting and builtins like `echo`. If you're not pulling in all of `oh-my-zsh.zsh`, some of those options may not be set in ways the other code works with.

In this case, the relevant option is `prompt_subst`, which is off by default, but enabled [in lib/theme-and-appearance.zsh](https://github.com/robbyrussell/oh-my-zsh/blob/2193135ebc4ce0778ff6949b448e4526dbd34284/lib/theme-and-appearance.zsh#L44).

The OMZ `lib/termsupport.zsh` `title()` function uses [`print -P`](https://github.com/robbyrussell/oh-my-zsh/blob/2193135ebc4ce0778ff6949b448e4526dbd34284/lib/termsupport.zsh#L9-L12) to perform the variable expansion here. The `-P` option means "do prompt expansion on the input". Prompt expansion can include both `%`-escape sequences and `$foo` parameter expansion. But when `prompt_subst` is off, prompt expansion does not include `$foo` parameter expansion.

This prompt expansion is the point where `$LINE` should be expanded; thus the single quotes at the calling location are correct. If they're changed to double quotes, then in the normal OMZ case when `prompt_subst` is on, the string will be double-expanded, causing incorrect output when `$LINE` contains `$`, `%`, or certain other characters.

If you want to make the `title` functionality work under more cases and protect it against `prompt_subst` state, you can use `emulate -L zsh; setopt prompt_subst` at the beginning of the `title()` function to locally set the shell behavior to what its code is written to use. This will make it more portable, without introducing breakage to the normal OMZ setup like changing the quotes would. I've put in a PR to do this: #4203.

For future reference, if you run in to behavior like this, just report it as a bug instead of submitting a PR. The code needs to make sense in the context of the whole of OMZ, which is complicated. Like Marc says, individual files may not work outside it.

Also, it appears that you're running the [`/etc/zshrc` code from GRML](http://git.grml.org/?p=grml-etc-core.git;a=blob_plain;f=etc/zsh/zshrc;hb=HEAD), which contains extensive configuration to zsh. This may interact with OMZ and OMZ functions in unforeseen ways. That could be something to look in to if you have other unexpected `zsh` behavior.

cherti:
alright, resorting to that fix then, closing here.

PR #4200, opened 2015-07-31
Add alias for "systemctl cat" and "systemctl edit"

lislon:
I added few aliases, which I miss when managing systemd:
 * `sc-cat` -> `systemctl cat` 
 * `sc-edit` -> `sudo systemctl edit`


mcornella:
:+1: /cc @robbyrussell

PR #4202, opened 2015-07-31
Add JIRA setting to show dashboard instead of create issue

zackhsi:
My most common JIRA action is to view my dashboard, so that's what the
command `jira` should do.

### DIFF
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -3,11 +3,11 @@
 #         .jira-url in the current directory takes precedence
 #
 # If you prefer the jira command to show your dashboard, set:
-#JIRA_DASHBOARD="true"
+#   JIRA_DEFAULT_ACTION="dashboard"
 # in you .zshrc
 #
 # If you use Rapid Board, set:
-#JIRA_RAPID_BOARD="true"
+#   JIRA_RAPID_BOARD="true"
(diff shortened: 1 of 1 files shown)


zackhsi:
// @robbyrussell @mcornella

plugins/jira/jira.plugin.zsh
```diff
@@ -38,8 +42,11 @@ open_jira_issue () {
   fi
 
   if [ -z "$1" ]; then
-    echo "Opening new issue"
-    $open_cmd "${jira_url}/secure/CreateIssue!default.jspa"
+    if [[ "x$JIRA_DASHBOARD" = "xtrue" ]]; then
```
apjanke: You don't need the "x$FOO" trick in `zsh`. `[[ "$JIRA_DASHBOARD == true ]]` is sufficient.

apjanke:
How about adding a `jira dashboard` special argument (like `jira assigned` and `jira reported`), and a `$JIRA_DEFAULT_ACTION` control instead of `$JIRA_DASHBOARD`? That way people who don't want it as the default action could still use the new feature. And the customization could be used if other special actions get defined, instead of having to reconcile multiple boolean controls. (And people could even use the existing `assigned` and `reported` actions if they wanted.)

zackhsi:
Updated @apjanke

PR #4205, opened 2015-08-01
Fix git-prompt plugin

mcornella:
This PR pulls in the latest version from olivierverdier/zsh-git-prompt and adds the following modifications:

- Single call\* to `git status --porcelain -b`, thanks to @wkentaro (https://github.com/olivierverdier/zsh-git-prompt/pull/65, https://github.com/mcornella/oh-my-zsh/pull/2)
  This avoids using up to 4 calls to git, while also fixing the problem with locales other than English (fixes #3460):
  ```
$ export LC_ALL=fr_FR.UTF-8
$ git status
Sur la branche dev
Votre branche est en avance sur 'origin/dev' de 5 commits.
  (utilisez "git push" pour publier vos commits locaux)
Fichiers non suivis:
  (utilisez "git add <fichier>..." pour inclure dans ce qui sera validé)

	asdf

aucune modification ajoutée à la validation mais des fichiers non suivis sont présents (utilisez "git add" pour les suivre)
$ git status --porcelain
?? asdf
```

- Uses `*_functions+=()` syntax instead of `add-zsh-hook`.

- Removes unnecessary stuff from the zsh file which is only needed on the upstream project.

- Cleans up the syntax in gitstatus.py. 

_\* If HEAD is in detached mode, a second call to `git rev-parse` is made to get the short SHA1._

Closes #3222.

### DIFF
--- a/plugins/git-prompt/git-prompt.plugin.zsh
+++ b/plugins/git-prompt/git-prompt.plugin.zsh
@@ -1,32 +1,34 @@
 # ZSH Git Prompt Plugin from:
 # http://github.com/olivierverdier/zsh-git-prompt
-#
-export __GIT_PROMPT_DIR=$ZSH/plugins/git-prompt
+
+export __GIT_PROMPT_DIR=${0:A:h}
+
+export GIT_PROMPT_EXECUTABLE=${GIT_PROMPT_USE_PYTHON:-"python"}
+
+# Initialize colors.
+autoload -U colors
+colors
 
 # Allow for functions in the prompt.
 setopt PROMPT_SUBST
 
-## Enable auto-execution of functions.
-typeset -ga preexec_functions
-typeset -ga precmd_functions
-typeset -ga chpwd_functions
+autoload -U add-zsh-hook
 
-# Append git functions needed for prompt.
-preexec_functions+='preexec_update_git_vars'
-precmd_functions+='precmd_update_git_vars'
-chpwd_functions+='chpwd_update_git_vars'
+add-zsh-hook chpwd chpwd_update_git_vars
+add-zsh-hook preexec preexec_update_git_vars
+add-zsh-hook precmd precmd_update_git_vars
 
 ## Function definitions
 function preexec_update_git_vars() {
     case "$2" in
-        git*)
+        git*|hub*|gh*|stg*)
         __EXECUTED_GIT_COMMAND=1
         ;;
     esac
 }
 
 function precmd_update_git_vars() {
-    if [ -n "$__EXECUTED_GIT_COMMAND" ]; then
+    if [ -n "$__EXECUTED_GIT_COMMAND" ] || [ ! -n "$ZSH_THEME_GIT_PROMPT_CACHE" ]; then
         update_current_git_vars
         unset __EXECUTED_GIT_COMMAND
     fi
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mricherzhagen** [comment]
To make this work i had to change
```
status = [(line[0], line[1], line[2:]) for line in stdout.splitlines()]
```
to
```
status = [(chr(line[0]), chr(line[1]), line[2:].decode("utf-8")) for line in stdout.splitlines()]
```
in [gitstatus.py](https://github.com/mcornella/oh-my-zsh/blob/fix-git-prompt-plugin/plugins/git-prompt/gitstatus.py#L21)

Also i'm seeing a weird `git:` prefix that i cannot get rid of...

**Comment by mcornella** [comment]
The part about decode should be there, let me push a modification and tell me if it works ok.

The part about the `git:` prefix I presume is because you have it in `$ZSH_THEME_GIT_PROMPT_PREFIX` in some part of your code. Post `echo $ZSH_THEME_GIT_PROMPT_PREFIX` and your `zshrc` file.

**Comment by mcornella** [comment]
Updated

**Comment by mcornella** [comment]
@c4software @autra can you test this?

**Comment by c4software** [comment]
Yep i will try this week. I'm not able to do it right now.

**Comment by mricherzhagen** [comment]
Got rid of the `git:` prefix by removing the robbyrussel-theme. Works fine for me now with the latest change.

**Comment by mcornella** [comment]
I can see why it happens on an empty git directory. This is what `git status --porcelain -b` outputs:
```zsh
$ git status --porcelain -b
## Initial commit on master
```
which is then parsed incorrectly.

However, I couldn't check the git submodule one. Here's what I did:
```zsh
$ git submodule add https://github.com/rupa/z.git z # sample repository
$ cd z && git status --porcelain -b
## master...origin/master
```
Can you post a sample submodule repo so that I can reproduce?

wkentaro:
The second one may be because 
```
% git checkout HEAD~1
% git st --porcelain -b
## HEAD (no branch)
```

wkentaro:
I hope this fixes all of these problems.
https://github.com/mcornella/oh-my-zsh/pull/3

wkentaro:
@mricherzhagen @mcornella 
Could you plz review it, when u have time?

mricherzhagen:
Oh right. my submodule was in detached HAD, because i checked out a tag.

mcornella:
Pulled in latest fix, works correctly :+1:

wkentaro:
I think tag name should be displayed even if detached status.
So, what do you think about this?
https://github.com/wkentaro/oh-my-zsh/commit/56494821c4729cffb3c34e3357a182a2d8eacbdc

If you like this, please pull it.

mricherzhagen:
@mcornella your latest version does not work with Python3:

```
  File "[my home].oh-my-zsh/plugins/git-prompt/gitstatus.py", line 60, in <module>
    str(len(untracked)),
TypeError: sequence item 0: expected str instance, bytes found
```

Fix:
```diff
diff --git a/plugins/git-prompt/gitstatus.py b/plugins/git-prompt/gitstatus.py
index 3a7d214..1beae54 100644
--- a/plugins/git-prompt/gitstatus.py
+++ b/plugins/git-prompt/gitstatus.py
@@ -22,7 +22,7 @@ for st in status:
         if re.search('Initial commit on', st[2]):
             branch = st[2].split(' ')[-1]
         elif re.search('no branch', st[2]):
-            branch = check_output(['git', 'rev-parse', '--short', 'HEAD']).strip()
+            branch = check_output(['git', 'rev-parse', '--short', 'HEAD']).decode('utf-8').strip()
         elif len(st[2].strip().split('...')) == 1:
             branch = st[2].strip()
         else:
```

@wkentaro I like it. 
Your code has the same problem with Python3, though:
replace `tag_hash = output.split('-g')`  with `tag_hash = output.decode('utf-8').split('-g')`

Also at one point the plugin was supposed to prepend a git hash with a `:`, which got lost somewhere.

mcornella:
@mricherzhagen the `decode('utf-8')` is a recurring theme on this PR :joy: Thanks for the python 3 explanation, now I won't forget to check it. It's now solved.

wkentaro:
The implementation of showing tag name if possible is here. plz review it.
https://github.com/mcornella/oh-my-zsh/pull/4
The output will be like:
```
$ python gitstatus.py
fc48dba 0 0 0 0 0 1
$ python gitstatus.py
tags/1.0 0 0 0 0 0 1
```

mcornella:
It is now merged and working a-OK :+1:
@mricherzhagen please test the (hopefully) final release.

mricherzhagen:
Works fine for me. Thanks!

mcornella:
Cool, thanks for confirming. Good to go :+1: 
/cc @robbyrussell

PR #4206, opened 2015-08-03
adding optionnal zsh install for debian

aversag:
hi
I've installed ohmyzsh several times and always getting error because the server didn't have zsh installed
so I added a test in the install script
hope it'll help!
thank you

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -32,6 +32,12 @@ sed -i -e "/export PATH=/ c\\
 export PATH=\"$PATH\"
 " ~/.zshrc
 
+TEST_ZSH_INSTALLED=$(dpkg -l | grep -w 'zsh' | grep -v 'zsh-' | grep 'ii')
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @aversag, we won't install zsh from the install script. This project needs to support every platform and if we add automatic installation for one platform we'll end up adding it for all of them, and that's not a good situation to be in. We put installation instructions in [the wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Installing-ZSH) and that's all we can do about it.

Thank you for trying to help.

aversag:
ok thank you
so why not just checking if zsh is installed ? if not exit

mcornella:
That's definitely comming on the near future, as we move towards a fully-zsh install script.

apjanke:
There's a PR for checking whether zsh is installed now: #4241.

mcornella:
Please close this issue

PR #4208, opened 2015-08-03
Adds dnf plugin

### LINKED ISSUE #4184: update yum plugin or a new one

ah450:
Useful aliases for dnf
fixes #4184


mcornella:

PR #4209, opened 2015-08-03
Adds npmE alias

ah450:
Option to execute a command from closest ancestor node_modules folder

Example
```zsh
$ls
node_modules
other_folder
$npmE gulp
No gulpfile found
$cd other_folder
$npmE gulp
No gulpfile found
```


mcornella:
Here's the correct way to add quotes: 
```
 alias npmE='PATH="$(npm bin)":"$PATH"'
```
This way the alias uses the dynamic `$PATH` every time, and the npm bin folder will only be set once, not use the appropriate one depending on the folder you're in.

ah450:
@mcornella Thanks, fixed the quotes

mcornella:
:+1: 
/cc @robbyrussell

PR #4212, opened 2015-08-04
Fix chpwd clobbering in dirhistory and last-working-dir

NAR8789:
dirhistory and last-working-dir clobber each other's chpwd functions when used together. This PR changes them both to append to chpwd_functions instead of clobbering chpwd

### DIFF
--- a/plugins/dirhistory/dirhistory.plugin.zsh
+++ b/plugins/dirhistory/dirhistory.plugin.zsh
@@ -49,7 +49,9 @@ function push_future() {
 }
 
 # Called by zsh when directory changes
-function chpwd() {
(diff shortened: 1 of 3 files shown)


mcornella:
Thanks! One minor change to the code style: use `chpwd_functions+=(function)` instead of the other one (that's the preferred way to add an element to an array). Also, you don't need the `typeset -ga` stuff since those variables are already global and accessible when the plugins are sourced.

NAR8789:
@mcornella Thanks for the tips! I've made the style fixes as requested.

mcornella:
Cool! This is ready to merge
/cc @robbyrussell

PR #4214, opened 2015-08-04
chucknorris: add missing line/quote delimiters

queria:
few quotes were missing the '%' line delimiters,
so some of the quotes were appearing at once merged together


apjanke:
:+1: Tested on OS X 10.9.5.

To test: do `while true; do c=$(chuck); if echo $c | grep "Guinness Book" >/dev/null; then echo $c; break; fi;  done`.

Before:

```
$ while true; do c=`chuck`; if echo $c | grep "Guinness Book" >/dev/null; then echo $c; break; fi;  done
In fine print on the last page of the Guinness Book of World Records it notes that all world records are held by Chuck Norris, and those listed in the book are simply the closest anyone else has ever gotten.
There is no chin behind Chuck Norris' beard. There is only another fist.
```

After:

```
$ while true; do c=`chuck`; if echo $c | grep "Guinness Book" >/dev/null; then echo $c; break; fi;  done
In fine print on the last page of the Guinness Book of World Records it notes that all world records are held by Chuck Norris, and those listed in the book are simply the closest anyone else has ever gotten.
```

@mcornella or @robbyrussell, this is a trivial merge. Want to pull it in?

mcornella:
:+1:

PR #4217, opened 2015-08-04
Implement and use a standard, cross-platform open command

mcornella:
This PR implements an oh-my-zsh standard open function that is cross-platform and easy to use for all plugins and users as well to create their aliases.

The function is called `open_command` and it uses the `$OSTYPE` value to decide which command to run. 

This PR also changes those plugins that used old logic so that we have consistent, non-duplicate code throughout the project codebase. Plugins changed: `fasd`, `frontend-search`, `jira`, `lighthouse`, `node` and `web-search`.

These other plugins also have open commands but are OSX specific, so there's no need to use a cross-platform open command: `atom`, `bwana`, `marked2`, `osx`, `textmate` and `xcode`.

Fixes #1478.
Fixes #2988.
Fixes #3920.

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -15,6 +15,22 @@ function take() {
   cd $1
 }
 
+function open_command() {
+  local open_cmd
+
+  # define the open command
+  case "$OSTYPE" in
+    darwin*)  open_cmd="open" ;;
+    cygwin*)  open_cmd="cygstart" ;;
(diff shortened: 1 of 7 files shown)


mattkatz:
I'm pretty late to this party but bravo @mcornella thanks!

mcornella:
Thanks @mattkatz! You reminded me of [this comic](http://happymondaycomics.com/post/140178430035/how-people-are-on-the-internet) :smile:

PR #4222, opened 2015-08-06
Add aliases for different symfony envs

gharlan:


mcornella:
I like @delphiki alias notation better.

PR #4235, opened 2015-08-09
Fix xcode plugin

mcornella:
This PR does the following:

- Fix the `xc()` function by only looking for `.xcworskpace` and `.xcodeproj` files.
  It also cleans it up by using zsh arrays and localized variables.

- Change `xcsel` function into an alias, since it doesn't do anything extraordinary and `"$*"` might have side effects.

Closes #3259, #3950.

### DIFF
--- a/plugins/xcode/README.md
+++ b/plugins/xcode/README.md
@@ -4,7 +4,7 @@
 
 This plugin provides a few utilities that can help you on your daily use of Xcode and iOS development.
 
-To start using it, add the `xcode` plugin to your `plugins` array:
+To start using it, add the `xcode` plugin to your `plugins` array in `~/.zshrc`:
 
 ```zsh
 plugins=(... xcode)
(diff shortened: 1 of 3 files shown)


mcornella:
@DarthMike, @Palleas can you test this PR?

mcornella:
@apjanke can I get a pair of eyes more?
/cc @robbyrussell

DarthMike:
@mcornella Tested now:
'xc' and 'xcsel' work fine for me.

mcornella:
Thanks!

apjanke:
Changes look good to me. I briefly tested on OS X 10.9.

How about having `xc` return nonzero to indicate failure in the case where it [couldn't find any Xcode files to open](https://github.com/mcornella/oh-my-zsh/blob/fix-xcode-plugin/plugins/xcode/xcode.plugin.zsh#L6)?

mcornella:
@DarthMike, I have a doubt. `.xcworkspace` and `.xcodeproj` are just the extension of the file, or are they called exactly that? I'm saying this because of [this comment](https://gist.github.com/subdigital/5420709#gistcomment-1199139) that talks about a `myProject.xcodeproj` file...

@apjanke good point, added a return statement.

apjanke:
Looks like it's a file extension, and not a hidden directory, at least for new projects using Xcode 6.x. I fired up Xcode and made a new project called "Whatever", and it gave me a `Whatever.xcodeproj` file inside the project directory. Should probably be looking for `*.xc...` and not just `.xc...`. Googling this seems to agree, but I don't see anything authoritative.

I'm not a Mac developer; maybe someone with more Xcode experience can chime in.

DarthMike:
In OSX some directories, sometimes called bundles seem like files. XCode workspaces and projects are actually directories with extensions. Same as a .app. So yes you can treat it as an extension and it's fine. 

The error when opening might be because of XCode versions, not related to the finding of the file.

It's not safe to search only for ".xc*" because iOS and OSX projects can use bundles for images with extension ".xcasset". They generally should not be in the same level as the project, but I've seen it happening.

mcornella:
Ok, updated: https://github.com/robbyrussell/oh-my-zsh/pull/4235/files#diff-730a33c55f58e027157b17caad0a73fcR4

apjanke:
Updated code works for me. I created a couple projects in Xcode, including one with spaces and special symbols in the name, and `xc` opens them in Xcode for me. Failure to find a file returns an error status.

One possibly-surprising behavior that may be worth noting in the comments for this plugin: the `xcsel` command will change the version of Xcode that is used with the `xc*` command line tools. But it does not change which version of Xcode is used by the plugin's own `xc` command. I suspect that's because `xc` relies on file associations and the `open` command, instead of something like `xcrun`. (If you wanted to extend `xc` to use `xcsel`'s selection, it could be done with the `open -a <application>` option.)

apjanke:
Say, would you mind guarding that `simulator` definition code with a check for `xcode-select`'s existence? Otherwise, on non-Macs (and maybe on Macs/Darwin without Xcode installed?), I get a warning when starting zsh. (I sync my `~/.zshrc` across all my machines, so I use the same `$plugins` list everywhere. I'd rather not have to add OS-conditional logic to it.)

```
/home/janke/.oh-my-zsh/plugins/xcode/xcode.plugin.zsh:21: command not found: xcode-select
[~]
$
```

This should do it.

```
if which xcode-select &>/dev/null; then
  if [[ -d $(xcode-select -p)/Platforms/iPhoneSimulator.platform/Developer/Applications/iPhone\ Simulator.app ]]; then
    alias simulator='open $(xcode-select -p)/Platforms/iPhoneSimulator.platform/Developer/Applications/iPhone\ Simulator.app'
  else
    alias simulator='open $(xcode-select -p)/Applications/iOS\ Simulator.app'
  fi
fi
```

apjanke:
Oh, you know what... maybe that `simulator` alias should instead be a function that tests for the `Platforms` thing each time it's called. If that path inside Xcode is version-specific, then which path you need to use might change between `simulator` calls during a `zsh` session if `xcsel` has been called to switch Xcode versions.

mcornella:
I'm ok with that change, give me a minute.

mcornella:
Updated. I'm considering adding a README file as well. @DarthMike you use the plugin, could you tell me what's your typical workflow with it?

apjanke:
New `simulator()` looks good to me. I tested it by switching between Xcode 5.0.2 and Xcode 6.2, and it's now picking the simulator from the "selected" Xcode.

apjanke:
Hey, what do you folks think of this: [an `xcselv` that lets you select an Xcode using just the version number](https://github.com/apjanke/oh-my-zsh/tree/fix-xcode-plugin-janke)?

```
➜  ~  xcselv 5.0.2
selecting Xcode 5.0.2: /Applications/Xcode-5.0.2/Xcode.app
➜  ~  xcode-select -p
/Applications/Xcode-5.0.2/Xcode.app/Contents/Developer
➜  ~  xcselv 6.2
selecting Xcode 6.2: /Applications/Xcode-6.2/Xcode.app
➜  ~  xcode-select -p
/Applications/Xcode-6.2/Xcode.app/Contents/Developer
➜  ~  xcselv -
selecting Xcode -: /Applications/Xcode.app
➜  ~  xcselv 9.3.2
xcselv: Xcode version 9.3.2 not found
➜  ~
```

If you like it we could extend it to list the installed versions and do tab completion of the version numbers.

mcornella:
I've added a README for the plugin, let's see what you think of it. @apjanke I'd rather keep this PR to only fixing the plugin, not adding new functionality. But it is true that it would be more useful than the current `xcsel` alias, so if @DarthMike likes this let's pull it in.

DarthMike:
@mcornella Awesome work :smile: I generally use the plugin to open xcode from command line, as I use it for most 'out of IDE' work. So most times it's just using 'xc' and 'simulator'.

Regarding changing xcodes, I think it would be a good idea to add it; I don't switch xcodes so often but it can be really useful if you have to.

mcornella:
Ok, I finally made my mind as to add the `xcselv` function, I think it's harmless to add it and it looks good from my perspective, so I don't expect there to be much of a bug. This is ready to rock now!

Hopefully it will be merged soon. Cheers!
/cc @robbyrussell

apjanke:
Thanks!

I extended the `xcselv` command to list installed versions, do completion, have a usage listing, and describe the version naming convention in the README. Also changed "-" to "default" to make option parsing easier.

It's on [fix-xcode-plugin-apjanke-02](https://github.com/apjanke/oh-my-zsh/tree/fix-xcode-plugin-apjanke-02). Let me know if you'd like to fold it in here, or would rather have a separate PR for it.

```
[~]
$ xcselv -h
Usage:
  xcselv <version>
  xcselv [options]

Options:
  <version> set the active Xcode version
  -h        print this help message and exit
  -p        print the active Xcode version
  -l        list installed Xcode versions (long human-readable form)
  -L        list installed Xcode versions (short form, version names only)
[~]
$ xcselv -l
5.0.2   -> /Applications/Xcode-5.0.2/Xcode.app
5.1.1   -> /Users/janke/Applications/Xcode-5.1.1.app
6.2     -> /Applications/Xcode-6.2/Xcode.app
default -> /Applications/Xcode.app
[~]
$ xcselv -p
default (/Applications/Xcode.app)
[~]
$
```

Completion:

screen shot 2015-09-10 at 3 44 58 am

screen shot 2015-09-10 at 3 48 48 am

mcornella:
Wow Andrew that is `xcsel`lent! 😁 I'll apply it today when I get a moment.

mcornella:
Updated with an EOF newline fix in `_xcselv`. The extra information in the README file is really useful, thanks for taking the time! :+1:

apjanke:
You're welcome!

PR #4238, opened 2015-08-10
Add Baidu support to web-search plugin

wubaiqing:
This plugin adds baidu commands to launch the default
web browser to do web search:

e.g.

    baidu oh-my-zsh


mcornella:
It works correctly. Ok to merge :+1: 
/cc @robbyrussell

PR #4239, opened 2015-08-11
rvm plugin: Fix `gems` to work on Linux machine

yous:
Repeating the `-E` flag won't work with GNU sed.


yous:
This pull request fixes #1579, which is not merged but closed by #2809, due to #1594, which fixed rbenv plugin.

mcornella:
Huh, I could've sworn that #1579 and #1594 where the same thing. I see now that one fixed the `rbenv` plugin, and the other the `rvm` plugin.
Good to merge :+1: 
/cc @robbyrussell

yous:
Any updates?

yous:
Thanks!

PR #4240, opened 2015-08-11
Fix: Plugin knife: Add relative local path for commands which upload files

kierandoonan:
Currently any of the `knife * from file` commands autocomplete by the knife plugin only look in the current directory for the relevant files. This fix searches upwards for a .chef folder and then looks in the folder relative to that folder (this is similar to the way knife itself functions). This should affect the following subcommands: 
- `knife node from file`
- `knife environment from file`
- `knife role from file`
- `knife data bag from file`

It also adds an option (`KNIFE_RELATIVE_PATH`) that lets `knife cookbook upload` work in the same way. This is useful when you have multiple local Chef orgs and can't set a path to a single cookbooks folder.

### DIFF
--- a/plugins/knife/_knife
+++ b/plugins/knife/_knife
@@ -230,12 +230,11 @@ _cookbook_versions() {
 # Searches up from current directory to find the closest folder that has a .chef folder 
 # Useful for the knife upload/from file commands 
 _chef_root () {
-  slashes=${PWD//[^\/]/}
   directory="$PWD"
-  for (( n=${#slashes}; n>0; --n ))
+  while [ $directory != '/' ]
   do
     test -e "$directory/.chef" && echo "$directory" && return
-    directory="$directory/.."
+    directory="${directory:h}"
   done
 }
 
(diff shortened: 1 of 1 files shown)


plugins/knife/_knife
```diff
@@ -198,4 +227,16 @@ _cookbook_versions() {
...
 
+# Searches up from current directory to find the closest folder that has a .chef folder 
+# Useful for the knife upload/from file commands 
+_chef_root () {
+  slashes=${PWD//[^\/]/}
+  directory="$PWD"
+  for (( n=${#slashes}; n>0; --n ))
+  do
+    test -e "$directory/.chef" && echo "$directory" && return
+    directory="$directory/.."
```
mcornella: You can use `${directory:h}` to get the parent folder. Then you only need to check if `$directory = /` in your loop condition, and you don't need the slashes and the `for`.

kierandoonan:
Updated to use directory:h

mcornella:
Better :+1: I haven't tested the rest, it would be nice to have some more people test this.
/cc contributors @loxley @marcparadise @maplebed @mugenken @cwjohnston @moutten @franklouwers

nmyster:
Will this be merged? Very useful!

petrhecko:
Works fine for me, would be great to be merged!

mcornella:
Merged. It would be better those documenting comments where put in a README file.

maxmanders:
Thanks folks, and nice work @kierandoonan :-D

PR #4244, opened 2015-08-14
Emoji plugin - Fix wrong string comparison

Gnouc:
`==` in Perl if for numeric comparison, using it make the update failed to generate data.


mcornella:
[Source](http://stackoverflow.com/a/14046720) :+1: 
/cc @robbyrussell

apjanke:
:+1:

PR #4249, opened 2015-08-16
Use https instead of http

KrauseFx:
The Twitter link didn't use https


mcornella:
:+1: 
/cc @robbyrussell

apjanke:
Conflicts with #3664, requiring manual merge. I manually made the same changes over there too.

There's another twitter link [in the upgrade.sh script](https://github.com/robbyrussell/oh-my-zsh/blob/9f8d7d85c2cacce54a5839d90b93e90b22228624/tools/upgrade.sh#L12), too.

PR #4250, opened 2015-08-16
Fix divide by zero in emoji plugin

### LINKED ISSUE #4245: Divide by zero error for random_emoji in emoji plugin
I'm on master, HEAD is at 192de6b.

Enabled the emoji plugin, also running:

```
plugins=(git git-extras osx history history-substring-search emoji)
```

I can select an individual emoji ok using e.g. `echo $emoji[penguin]` but if I try the random function, I get a divide by zero error for any category (no category specified works ok).

```
$> random_emoji fruits
random_emoji:9: division by zero
```

mcornella:
- Uses the right variable name (`$group` instead of `$group_name`)
- Check that the names array has any element before attempting to pick an emoji out of it. This fixes the same problem on groups that don't exist (_e.g._ `random_emoji 404notfound`).

Fixes #4245.

/cc @robbyrussell @fallwith @apjanke

### DIFF
--- a/plugins/emoji/emoji.plugin.zsh
+++ b/plugins/emoji/emoji.plugin.zsh
@@ -255,7 +255,7 @@ function random_emoji() {
   if [[ -z "$group" || "$group" == "all" ]]; then
   	names=(${(k)emoji})
   else
-  	names=(${=emoji_groups[$group_name]})
+	names=(${=emoji_groups[$group]})
   fi
   local list_size=$#names
   local random_index=$(( ( RANDOM % $list_size ) + 1 ))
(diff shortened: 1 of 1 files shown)


apjanke:
Oops! Don't know how that variable name got through. Guess I didn't re-test after doing a rename after original testing.

:+1:  I can reproduce the issue, and this PR fixes it for me.

apjanke:
Oh – [this comment](https://github.com/mcornella/oh-my-zsh/blob/9a397254dacca687bf7d204ac9d31f02ca2af242/plugins/emoji/emoji.plugin.zsh#L79-L80) about "extra local arrays" can also be deleted; it refers to an older working version of the code that was using temporaries that no longer exist in the implementation I actually checked in. Forgot to remove the obsolete comment.

```
#
# These extra local arrays are used to allow more convenient formatting of the source code.
```

mcornella:
Done!

apjanke:
Looks good. Ready to merge IMHO. (And no reason to wait, since the recent merge of the emoji plugin is partially broken without it.)

PR #4251, opened 2015-08-16
Fix completion waiting dots function

mcornella:
### STATUS: READY

This PR solves the annoying problem of new line prompt being redrawn each when the terminal is small enough that the prompt + 5 chars being drawn don't fit the line. This is done by disabling automatic line wrapping using terminfo's `rmam` sequence.

It uses `%F{red}` as well, instead of `\e[31m` which might not be that bad for portability but it looks better and is clearer what the color is.

----

### Test

**Instructions:** you need a prompt that is sufficiently big that it occupies almost all the line space. There are 5 completion dots, so you'll need to have a prompt that is `$COLUMNS - command length - 4` minimum to trigger the new line prompt redraw. You also need to uncomment the line `COMPLETION_WAITING_DOTS=true` in your zshrc file to enable the feature.

Once that is in place, trigger the bad behavior with `command [TAB]` (it needs to be a command that hasn't been completed yet, I use `ls [TAB]`). The 5 dots will be printed and the prompt will be redrawn on the line below to accomodate for the extra 5 characters. 

After that, copy the new function definition and try to trigger the same behavior (you have to use a different command now or `exec zsh` to restart the shell). See that dots get printed but the prompt is not redrawn. 

Here's the new function definition:
```zsh
expand-or-complete-with-dots() {
    # toggle line-wrapping off and back on again
    [[ -n "$terminfo[rmam]" && -n "$terminfo[smam]" ]] && echoti rmam
    print -Pn "%F{red}......%f"
    [[ -n "$terminfo[rmam]" && -n "$terminfo[smam]" ]] && echoti smam

    zle expand-or-complete
    zle redisplay
}
```

----
### Results

**LEFT:** _without_ fix. **RIGHT:** _with_ fix.
<a href="https://asciinema.org/a/25388?autoplay=1" target="_blank"></a> <a href="https://asciinema.org/a/25389?autoplay=1" target="_blank"></a>

/cc @robbyrussell

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -60,7 +60,11 @@ zstyle '*' single-ignored show
 
 if [[ $COMPLETION_WAITING_DOTS = true ]]; then
   expand-or-complete-with-dots() {
+    # toggle line-wrapping off and back on again
+    [[ -n "$terminfo[rmam]" && -n "$terminfo[smam]" ]] && echoti rmam
     print -Pn "%{%F{red}......%f%}"
+    [[ -n "$terminfo[rmam]" && -n "$terminfo[smam]" ]] && echoti smam
+
     zle expand-or-complete
     zle redisplay
   }
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -58,7 +58,7 @@ zstyle ':completion:*:*:*:users' ignored-patterns \
 # ... unless we really want to.
 zstyle '*' single-ignored show
 
-if [ "x$COMPLETION_WAITING_DOTS" = "xtrue" ]; then
+if [[ $COMPLETION_WAITING_DOTS = true ]]; then
   expand-or-complete-with-dots() {
     echo -n "\e[31m......\e[0m"
     zle expand-or-complete
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -60,7 +60,7 @@ zstyle '*' single-ignored show
 
 if [[ $COMPLETION_WAITING_DOTS = true ]]; then
   expand-or-complete-with-dots() {
-    echo -n "\e[31m......\e[0m"
+    print -Pn "%{%F{red}......%f%}"
     zle expand-or-complete
     zle redisplay
   }


apjanke:
Hmmm. Doesn't seem to fix it for me, on OS X using iTerm2 or Terminal.app.

When using this change, the prompt is still redrawn on the next line.

screen shot 2015-08-18 at 2 10 52 pm

apjanke:
I'm seeing different behavior when running `xterm` on OS X. Its line wrapping behavior is different: it won't automatically wrap to the next line if you output characters beyond the window width. So it doesn't have the "prompt displayed on the next line" problem in the first place.

But there's another issue I found with testing it: `$COMPLETION_WAITING_DOTS` doesn't work under `xterm` in some cases. For me, even if you have the variable turned on, zsh still does normal tab completion without displaying the dots.

The problem seems to be that the key binding for the completion-with-dots isn't set. The `lib/*.zsh` files are loaded in alphabetical order. `completion.zsh` comes before `key-bindings.zsh`. There is a `bindkey -e` down in `key-bindings.zsh` which makes `emacs` the main keymap. This means that key bindings set in earlier `lib/*.zsh` files are applied to the default main keymap that was in effect when `zsh` started. In my iTerm2 and Terminal.app, the main keymap is `emacs` by default, so `completion.zsh`'s key bindings go in the `emacs` map, and `key-bindings.zsh`'s `bindkey -e` leaves things as they are. But when I run `xterm`, the default main keymap is `viins`, so `completion.zsh` adds a binding to the `viins` keymap, and then later, `key-bindings.zsh` switches to another keymap.

It *think* this turns out to be due to how I was launching xterm: from the command line in another `zsh` session. The [ZLE doco](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Keymaps) says this:

> either ‘emacs’ or ‘viins’ is also linked to the name ‘main’. If one of the VISUAL or EDITOR environment variables contain the string ‘vi’ when the shell starts up then it will be ‘viins’, otherwise it will be ‘emacs’. bindkey’s -e and -v options provide a convenient way to override this default choice. 

And I have `export EDITOR=vi` set in my `.zshrc`. So additional `zsh` sessions that inherit that environment will start out with the `viins` keymap. iTerm and Terminal are unaffected because they're launched from the GUI level, which doesn't have `$EDITOR` set. This sensitivity to the default keymap in OMZ could be fixed by moving `bindkey -e` way up in the startup sequence so it's before any key bindings which expect to go in to the default keymap. (The `vi-mode` plugin will similarly disable `$COMPLETION_WAITING_DOTS`, since it switches to `bindkey -v` after all the `lib/*.zsh` files have run.)

apjanke:
Maybe something else is going on here. The ZLE documentation doesn't explicitly say what the semantics of writing to `stdout` from a zle widget are. So I assume that it is just sent to the terminal, and zle is not aware of what was written, and it's not considered part of the buffer or prompt. (Widgets  modify the buffer variables to make zle aware of the line content.)

Maybe the original issue is that the printed dots are seen by the terminal, causing it to wrap to the next line (in terminals that do line wrapping), but zle wasn't aware of them in the first place. So zle doesn't know that the line has wrapped, and when it redisplays, it just backs up to the beginning of the line the cursor is on.

If that's the case, maybe the right thing to do is to check the line width, and make sure not to output so many dots that the line wraps. Or to use cursor control codes to move the cursor back where it started (and zle thinks it still is).

Here's one that uses the `sc` (save cursor) and `rc` (restore cursor) terminal capabilities to do that. (With a `sleep` thrown in to make debugging it easier.) This works for me, in that it still displays the red dots, and it prevents the prompt from being redrawn on a new line.

```
  expand-or-complete-with-dots() {
    tput sc
    print -Pn "%F{red}......%f"
    tput rc
    sleep 1
    zle expand-or-complete
    zle redisplay
  }
```

Hmm. Well, it almost works: if it's done on the bottom line of the terminal screen, the wrapped dots cause everything to scroll up one line, so the cursor position relative to the screen isn't the same place relative to the line of text, and you still get the prompt redisplay on the next line.

A production version should probably check whether sc/rc are supported by the current `$TERM`, and if not, maybe fall back to simpler cursor-movement commands, or skip the dots display, or just let it wrap, or whatever.

mcornella:
It seems I went all in too soon: 
bug screenshot

I "tested" this on babun running inside ConEmu and there was some wonkiness regarding column space and the scrollbar, so I probably misinterpreted my results. Let me reevaluate...

mcornella:
> But there's another issue I found with testing it: $COMPLETION_WAITING_DOTS doesn't work under xterm in some cases. For me, even if you have the variable turned on, zsh still does normal tab completion without displaying the dots.

This is probably because it's only displayed when there is enough computation time, otherwise it will be immediately refreshed—therefore it won't be seen (I haven't looked closely at what you say about bindkey definitions though).

For that reason I've made a test function to only test the prompt line redraw, at the press of `CTRL`+`X`,`CTRL`+`A`:

```zsh
PS1='$ '
print-to-prompt() {
        emulate -L zsh
        local N
        N=$[ $COLUMNS - $#BUFFER - 2 + 1] # 2 is the number of chars in prompt

        #### BEGIN testable code
        tput sc
        print -Pn %{%F{red}%}$(printf '.%.0s' {1..$N})%{%f%} >&2
        tput rc
        #### END testable code
        sleep 1

        zle redisplay
}
zle -N print-to-prompt
bindkey '^X^A' print-to-prompt
```

With that exact prompt configuration I can confirm everything you've said. Works everywhere except when on the bottom line of the terminal screen. I'll keep on testing stuff...

apjanke:
It wasn't the execution time... even when I stuck a `sleep 1` in `expand-or-complete-with-dots()`, it never showed up. It was the `bindkey`+`$EDITOR` thing. Sticking a `bindkey -e` in my `~/.zshrc` before loading OMZ cleared it up so the dots appeared. (Though that's an edge case that should probably be addressed in OMZ itself.)

apjanke:
I tried using `tput dch 6` (delete character) and `tput cub 6` to try to back up over the individual characters. They didn't work in either iTerm or Terminal.app: the cursor stops at the beginning of the wrapped line, instead of wrapping up into the previous line and deleting the original characters. (You can even check whether the terminal does this with `$terminfo[bw]`;

This is tricky. I think we might need to figure out the  space left on the line, and shorten the dots to avoid wrapping. I don't see an easy way to find out what column we're in: we don't know how long the prompt is in the general case. (Even if we did know the exact prompt string, we can't easily tell which characters are wide.) Maybe there's a terminal capability that would allow us to query for the current cursor position.

Maybe we're using `$TERM` definitions that don't match the actual behavior of the iTerm and Terminal.app terminal emulations. I have `TERM=xterm-256color`, and `$terminfo[sam]` is reporting `no`. Based on [this curses doco](http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html), that `sam` indicates whether "auto line wrapping" is on.

> semi_auto_right_margin 	sam 	YE 	Printing in last column causes cr

Based on the behavior we're seeing, I *think* that iTerm is doing that `sam` behavior.

mcornella:
~~I've found a workaround using `tput cuu1` when we're on the bottom line. This will move the cursor up 1 line after restoring the cursor (which I've noticed only restores its column, at least in my case). This way there will not be a new prompt displayed, but there will still be a new line below. Plus, I haven't figured out a way to know when we're on the bottom line of the screen...~~

I think I found a way—I don't know if it's very portable—using a [VT escape code to disable autowrap](http://www.vt100.net/docs/vt102-ug/chapter5.html#S5.5.2.8), thanks to [this guy's answer on superuser](http://superuser.com/a/600694). Here's the code:

```zsh
print-to-prompt() {
        emulate -L zsh
        local N
        N=$(( $COLUMNS - $#BUFFER - 2 + 1)) # 2 is the number of chars in prompt

        #### BEGIN testable code
        echo -n '\e[?7l' # disable autowrap
        print -Pn "%{%F{red}$(printf '.%.0s' {1..$N})%f%}"
        echo -n '\e[?7h' # enable back autowrap
        #### END testable code

        sleep 1
        zle reset-prompt
}
```

mcornella:
I've pushed the final solution and here are the results:

##### Behavior without fix
<a href="https://asciinema.org/a/25388" target="_blank"></a>

##### Behavior with fix
<a href="https://asciinema.org/a/25389" target="_blank"></a>

apjanke:
Ch-ching! That sounds like a great approach. And thanks for mentioning `echoti`. I like that this is a no-shellout implementation; that'll be nice for Cygwin.

Works for me in iTerm2, `xterm`, `urxvt`, and Terminal.app on OS X 10.9.5. (Terminal.app still is a little wonky for me with redisplaying prompt or wrapping text after some window resizes, but I think this is due to it mishandling wide characters in the prompt string, and not this `smam`/`rmam` dot printing.)

Two tweaks to consider: should probably test to see if the terminal supports `rmam` (e.g. with `if [[ -n $terminfo[rmam]`) before using it to avoid warnings.

screen shot 2015-08-22 at 5 06 59 pm

And is using the "no-movement" `%{...%}` guard sequence the right thing? I don't think it has any practical effect since it's not going through ZLE, and a redisplay is done afterwards. But the `.`s are printable characters and do cause cursor movement, so it seems incorrect to put them in `%{ %}` without adding glitch characters. And the `%F`/`%f` sequences don't need guards because they're internal `zsh` things so `zsh` already knows they're non-printing.

mcornella:
Updated with the latest suggestions. I didn't know if `%{...%}` was necessary but I didn't think it would make a difference for the worse. Just to be safe, I removed them as well. 

Regarding the check, maybe the one I used is overkill. Perhaps if the terminal supports `rmam` we can assume that it also supports `smam`. But I don't know where to find that information other than testing different terminals...

apjanke:
Cool.

Including a printable character inside `%{ %}` in a prompt (and maybe buffer?) is as much of an error as emitting a non-printable character without that guard: it results in a cursor offset error, but in the other direction. You have to include `%G`s ("glitches") to compensate. Doesn't matter here probably because the `print` is bypassing ZLE. But we don't want to set a bad example.

Did you push the removal of `%{ %}` to your branch? I still see them in there when I update.

That `rmam`+`smam` check might feel like overkill, but I think it's the most correct. Making that "rmam implies smam" assumption would always leave you open to an edge case. (E.g. some platforms might ship with a bad `terminfo` entry that has `rmam` but left out `smam`. Or some wacky emulator might only support a one-way switch. Or whatever.) I don't think there *is* any way to test that assumption, because we don't have access to all terminfo entries that are out in the wild, including custom ones made by users. Testing terminals themselves wouldn't be enough either: check out that terminfo doco. There are `-am` and `-nam` variants to terminfo entries that use the capability definitions to enable or disable auto-margins, even if the terminal supports them. Your test exactly represents what the code requires and leverages a standard portability mechanism; I say go with it.

Oh, hmmmm, there's another edge case: line wrapping might already be off; i.e. we're already in `rmam` mode in the terminal because that's its default, or because something at a higher level has turned it off. (Like, maybe `zsh` is running in an embedded shell window or line, like inside emacs or another editor?) In which case, we'd be turning it on as a side effect. Hmmmm. Is it possible to test whether `am` is currently enabled?

This is a rabbit hole. :)

###  Backspacing to the previous line   ###

BTW, I found a terminfo capability that indicates whether backspacing will wrap back to the previous line: `$terminfo[bw]`

mcornella:
> Did you push the removal of %{ %} to your branch? I still see them in there when I update.

Oops, forgot to rebase-edit. Now it's ok. 

> Oh, hmmmm, there's another edge case: line wrapping might already be off; i.e. we're already in rmam mode in the terminal because that's its default, or because something at a higher level has turned it off. (Like, maybe zsh is running in an embedded shell window or line, like inside emacs or another editor?) In which case, we'd be turning it on as a side effect. Hmmmm. Is it possible to test whether am is currently enabled?

Well, let's just hope that won't happen or else I'm gonna get a brain aneurysm :grin: 

I've tested what happens with the `$terminfo[am]` value, but that doesn't—it only indicates if the terminal has capability to autowrap lines, not what state it's in. I've also tested if a `emulate -L zsh` could fix that but as I suspected it doesn't make a difference.

But really, I'd expect that problem to happen only if the user presses tab after disabling `am` and not enabling it back. In that case, the function will enable it back for the user, and then he'd see a different behavior than expected. But this won't happen any other way, no code to have compatibility issues with. So l wouldn't worry much about that case. 

> BTW, I found a terminfo capability that indicates whether backspacing will wrap back to the previous line: `$terminfo[bw]`.

If you mean that to test it in conjunction with `echoti sc` / `echoti rc`, I'd guess it could work. But on `xterm-256color` it already says `no`, _i.e._ it doesn't support that.

apjanke:
Yeah, `$terminfo[am]` won't work: all the terminfo stuff is a static definition of the terminal capabilities (I think), not its state. It would require something like a terminfo capability that told you how to query for the `am` state, and then you'd have to send that to the tty and then capture the response. Or have some wrapper for that otherwise exposed in `zsh` or curses.

> I'd expect that problem to happen only if the user presses tab after disabling am ... no code to have compatibility issues with

I dunno about that. What if it's not a user action that issued the `rmam`, but a program configuring the terminal? E.g. `emacs`, while setting up the screen for editing, and a small window for an embedded shell to be displayed. Or some other `zsh` function which is doing stuff and has invoked `zle` for its own purposes? I don't know enough about terminal programming to know if this is something that would actually happen.

I'll see if I can figure out how to test this for a few scenarios, like `emacs` and `vim`. I don't currently use a feature like this, so don't know how to do it offhand.

> If you mean that to test [bw] in conjunction with 

Nah, I just mentioned it as a curiosity, since I had brought it up earlier. Like you say, dead-end, unless there's a way to turn it on, which I don't see.

PR #4253, opened 2015-08-17
Add Spotify CLI

rafaelstz:


apjanke:
This isn't structured as an Oh My Zsh plugin. Plugins that provide functionality besides completion definitions need to have a main `<plugin_name>.plugin.zsh` file that gets run when the plugin is loaded by OMZ, so it can define commands without any additional installation or configuration by the user. In this case, it should be a wrapper script that defines a `spotify()` function that calls the AppleScript file.

Looks like upstream repo is https://github.com/dronir/SpotifyControl.

Best way to do it would probably be to supply upstream with a PR that converted it to an OMZ plugin. That way it would be easy to keep in sync, and could (I think) be used with Prezto and Antigen as well.

rafaelstz:
Sorry, fix the structure.

apjanke:
I don't think that's the right fix. By doing a rename, `spotify.plugin.zsh` is still an AppleScript program, not a `zsh` program. What is needed is an *additional* `spotify.plugin.zsh` file in the `zsh` language that can provide a bridge between it and the `spotify` AppleScript program.

rafaelstz:
Sorry but I do not found the solution.

PR #4254, opened 2015-08-18
Cross-platform clipboard clipcopy() and clippaste()

### LINKED ISSUE #4232: copyfile and copydir are osx specific
They both use the pbcopy command which is only available on osx.

apjanke:
Fixes #4232

###  Summary   ###

Adds clipcopy() and clippaste() generic cross-platform CLI clipboard functions, so users and functions can work portably with the clipboard. Changes the following  plugins to use these new portable functions, instead of the Mac-only `pbcopy` command.

* copyfile
* copydir
* coffee

I put them in a core `lib/clipboard.zsh` file instead of a new plugin because this changes both multiple plugins to use these functions, and we don't have a way of specifying dependencies between plugins. Sticking them in another plugin would break `copydir`, `copyfile`, and `coffee` for their existing users.

CC: @mbauhardt (author of copyfile and copydir) and @coopermaruyama (author of coffee) since their plugins are changed by this.

The `cloudapp` plugin (cc: @matthewmccullough) may also be able to be changed to use this instead of `pbcopy`. But I didn't make that change myself, because it is using a Ruby script which then shells out, and I don't know if that shell-out will always get `zsh` and have OMZ loaded in it. If nothing else, it could adapt the logic used here to a Ruby function for picking the clipboard command portably. 

I've tested on OS X 10.9.5 and Windows 7. Will test on Linux once I have a chance to get on a Linux box too.

###  Interaction Risks   ###

Introduces new `clipcopy` and `clippaste` commands for all users. These command names aren't standard or part of widely installed programs from what I can tell, so it shouldn't mask any existing commands.

Users who have defined their own commands named `clipcopy` or `clippaste` outside OMZ will experience breakage due to the collision. We could change the names to `omz_clipcopy` and `omz_clippaste` to avoid the collision and potential breakage. But I think these might be useful interactive commands and the shorter names would be appropriate.


lib/clipboard.zsh
```diff
@@ -0,0 +1,70 @@
+# System clipboard integration
+#
+# This file has support for doing system clipboard copy and paste operations
+# from the command line in a generic cross-platform fashion.
+#
+# On OS X and Windows, the main system clipboard or "pasteboard" is used. On other
+# Unix-like OSes, this considers the X Windows CLIPBOARD selection to be the
+# "system clipboard", and the X Windows `xclip` command must be installed.
+#
+# When using X Windows, copied data will only be good for a single paste
```
jody-frankowski: Can you elaborate on this? This is not a behavior I can reproduce on my system, I can paste multiple times the same clipboard.

apjanke: Oops! This is just a mistake I made reading the `man xclip` documentation. I thought it defaulted to `-loops 1` and misread this section

> The default action is to silently wait in the background for X selection requests (pastes) until another X application places data in the clipboard, at which point xclip exits silently.

as saying it waits until another application *reads from* the clipboard.

I'll remove it.

apjanke:
Updated to remove bogus comments on how `xclip` works.

coopermaruyama:
+1

mcornella:
#4305 also tries `xsel` which might be there when `xclip` is not.

mcornella:
Also, I'm not a fan of those names. What about `copy_command` and `paste_command` like with the [`open_command`](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/functions.zsh#L18)?

apjanke:
I've added `xsel` support based on #4305's implementation. Someone who uses `xsel` want to test it out?

apjanke:
I kinda prefer the `clipcopy` and `clippaste` names. I like the shared prefix/leading substring, so you know that clipboard manipulation is `clip*`, and can do tab-completion on it to see what related commands are there. (Like how the X11-related commands all start with `x`, and OS X's clipboard commands start with `pb`.)

And I think having the word `clip` or `clipboard` in there is helpful, because it distinguishes a clipboard-related copy operation from a file copy operation, which is what some other common commands with `copy` or the shorthand `cp` in their name do. If I just saw the name `copy_command`, I would assume it was an abstraction for `cp`, not a clipboard operation.

I kind of also think the `_command` is redundant, since almost anything you do on the shell command line is a command. Going with a prefixed verb seems more concise.

Definitely open to hearing people's preferences on this though.

mcornella:
I agree on all your naming points, you really know how to be persuasive :smile:

Regarding `xsel`, I have tested it and it works well with both `xclip` and `xsel`. 

I had one small issue though: after I uninstalled `xclip`, the `which xclip` line was still returning true because I had an `xclip` alias. Once I unaliased it it run smoothly. Perhaps we should use `(( $+commands[xclip] ))` instead, and prefix calls with `command` to avoid running into redefined aliases or functions.

I would also expand the usage section of the `clippaste` function to make it more like the `clipcopy` function, with an example piping its output and another redirecting it to a file. Even though I know this should be known by any terminal user, explicitly putting it there makes it more consistent and could help some noobs.

That is all I found, it's definitely just little things and the PR is almost ready. Let me try this on cygwin and you'll get my shipit squirrel!

mcornella:
Oh yeah, one more thing: we need newlines on `lib/clipboard.zsh` and `plugins/copydir/copydir.plugin.zsh`:
```diff
diff --git a/lib/clipboard.zsh b/lib/clipboard.zsh
index 59a62fa..e044fd1 100644
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -72,4 +72,4 @@ function clippaste() {
       return 1
     fi
   fi
-}
\ No newline at end of file
+}
diff --git a/plugins/copydir/copydir.plugin.zsh b/plugins/copydir/copydir.plugin.zsh
index 4b918e8..c451062 100644
--- a/plugins/copydir/copydir.plugin.zsh
+++ b/plugins/copydir/copydir.plugin.zsh
@@ -2,4 +2,4 @@
 function copydir {
   emulate -L zsh
   print -n $PWD | clipcopy
-}
\ No newline at end of file
+}
```

apjanke:
Thanks! 

Fixed the newlines and expanded the `clippaste` usage section showing output to a pipe and a file. Added a couple examples because the `clippaste > <file>` looked a little confusing, what with the brackets being used for multiple purposes.

It could use a section in a README, I guess, but we don't have one of those for the `lib/` stuff. Eventually we should make a `doc/` dir with some API and command references, I guess.

> I had one small issue though: after I uninstalled xclip, the which xclip line was still returning true because I had an xclip alias. Once I unaliased it it run smoothly. Perhaps we should use (( $+commands[xclip] )) instead, and prefix calls with command to avoid running into redefined aliases or functions.

I dunno about being this defensive everywhere. In the installer and diagnostic tools, or places where it could be potentially destructive, yeah. But when a user redefines a command using an alias or function or an executable file in their `$PATH`, they're taking responsibility for making it work and be compatible with their other uses of that command. And that gives them power, too: you can use that redefinition to compose functionality, insert debugging hooks or safety features, and so on. I don't know if it's our job to start second-guessing them everywhere.

I mean, what if I had a weird experimental X setup (e.g. I built X with a prefix on all its commands) or alternate windowing system, but it had a clipboard and I wanted it to work here? I could define a function or alias that implemented `xclip` in terms of what I had installed.

mcornella:
I dunno about being this defensive everywhere. [...]

I guess my thoughts go more in the direction that there is a potential that people have grown their setup to use custom aliases for copy and paste operations, so there is the potential that those xclip and xsel commands might fail to accomodate that. But I understand this is a fringe case that, on the other hand, will direct people this way if it doesn't work correctly.

The `(( $+command[cmd] ))` form is idiomatic of zsh though, we should strive to use is over the `which &>/dev/null` form.

-----------

Works as well with cygwin, let's pull this in. :shipit:
:+1: /cc @robbyrussell

apjanke:
But I think `(( $+command[cmd] ))` will only detect external commands, not functions or aliases. I don't think it matters here – nobody's *really* going to do any of those examples I mentioned – but in the general case, I think we should allow aliases and functions as well. (Heck, depending on how some programs implement things, they might be functions in the main program distribution, or might even change between functions and external commands between versions. `nvm` and the family of `virtualenv` things seem to work this way.)

PR #4260, opened 2015-08-18
plugin symfony2 sf2.7 compatibility fix

fgibaux:
with symfony 2.7 command group titles are listed as commands. this commit prevents it.


LogansUA:
:+1: 
You also may change "sfroute" alias to "sf debug:router" because of: "The use of "router:debug" command is deprecated since version 2.7 and will be removed in 3.0."

fgibaux:
There is already another push request that fix the alias : https://github.com/robbyrussell/oh-my-zsh/pull/4218

PR #4270, opened 2015-08-21
Add update and auto-update for customization directory

mscalora:
Modified upgrade_oh_my_zsh() and check_for_upgrade.sh/_upgrade_zsh() to run tools/upgrade.sh in the custom directory if it exists and is executable.


apjanke:
For everything else in the `$ZSH_CUSTOM` directory, files at the same relative location as a file in the main `$ZSH` directory override the main file, and are run instead of, not in addition to, it. Should the same be done here?

PR #4271, opened 2015-08-21
added apt-get autoremove in ubuntu plugin

loctauxphilippe:
I use that command a lot of times, and there's no shortcut for that.
Now it does :)


loctauxphilippe:
@robbyrussell thanks for merging :)

PR #4273, opened 2015-08-22
Add a missing rbenv location

erbridge:
The [AUR package](https://aur.archlinux.org/packages/rbenv) for rbenv puts it in `/usr/bin/rbenv`, which is currently missing from the search paths.


abdelq:
Hi,
`/usr/bin/rbenv` is not a directory but an executable, therefore it's not valid to put it there.

erbridge:
You're right. I didn't really think before submitting...

The executable is actually linked to `/usr/lib/rbenv/libexec/rbenv`, and `/usr/lib/rbenv/` doesn't contain a `bin/`. I might take another look at this later.

PR #4276, opened 2015-08-24
GET API Elasticsearch plugin

fedelemantuano:
Plugin for Elasticsearch GET API.


robbyrussell:
Thanks, can you include a README?

fedelemantuano:
Yes, done.

mcornella:
You should also delete the comments from the `plugin.zsh` file and put it in the README. That way you can document the whole thing as well as include the author and possibility of using `json_reformat`.

sxamit:
can you add environment variable to change localhost to any host i want?

PR #4282, opened 2015-08-25
Added missing commands to Vagrant plugin

ferenczy:
I have added some missing Vagrant commands and `box` command's arguments and also updated descriptions of commands according to output of `vagrant list-commands` using the most recent version of Vagrant (1.7.4).

Also removed some trailing whitespace and fixed indentation.


mcornella:
:+1: 
/cc @robbyrussell

PR #4285, opened 2015-08-26
'ip' command compatibility and much quicker external IP detection

nmaggioni:


plugins/systemadmin/systemadmin.plugin.zsh
```diff
@@ -140,12 +140,16 @@ d0() {
...
 # gather external ip address
 geteip() {
-    curl http://ifconfig.me
+    echo "$(curl -s -S http://ipecho.net/plain)"
 }
 
 # determine local IP address
 getip() {
-    ifconfig  | grep 'inet addr:'| grep -v '127.0.0.1' | cut -d: -f2 | awk '{ print $1}'
+    if [ $(hash | grep '^ip=') ]; then
```
mcornella: What does hash output? It is a zsh builtin which shouldn't print anything by itself...

nmaggioni: ```hash``` is quite similar to ```which```, except is just lists all the executables known in the current $PATH, one per line. In this case I preferred it instead of ```which``` because some implementations of the latter produce output even if the program passed as a parameter is not found.

mcornella: Ok, I understand you're using that to test whether the `ip` command is available. Instead, you can use the much shorter and preferred method: `if (( ${+commands[ip]} )); then ...`. See [an example](https://github.com/robbyrussell/oh-my-zsh/blob/1605d8264c6cfb7858bac96427a99f8bd9224a21/plugins/fbterm/fbterm.plugin.zsh#L3).

For one, `hash` on my machine doesn't output anything (you have to call `hash -v`). Secondly, you avoid a call to `grep` which is an external process and therefore requires forking a new process.

plugins/systemadmin/systemadmin.plugin.zsh
```diff
@@ -140,12 +140,16 @@ d0() {
 
 # gather external ip address
 geteip() {
-    curl http://ifconfig.me
+    echo "$(curl -s -S http://ipecho.net/plain)"
```
mcornella: A better alternative could be `curl https://icanhazip.com`, which uses SSL and the output contains the necessary newline character. 

Also, what's with the `-sS` flags? Silent + show-errors; isn't it the same with no flags?

nmaggioni: I forgot about ```icanhazip``, that's a better alternative.
On my system if I call ```curl``` without parameters is outputs percentages, time elapsed and so on... Those flags hide them but still show errors so someone gets an hint on what's wrong.

plugins/systemadmin/systemadmin.plugin.zsh
nmaggioni: Thanks for the suggestion, I'm implementing it.

plugins/systemadmin/systemadmin.plugin.zsh
mcornella: Ok, I agree with you then.

mcornella:
Ok this is ready to merge :+1: 
/cc @robbyrussell

Issue #4246, opened 2015-08-14
Why font color overlaps and changed in iTerm2 on Mac


FinalTheory:
![qq20150814-1](https://cloud.githubusercontent.com/assets/2818951/9275566/79908882-42d0-11e5-85fe-c30b07986331.png)
![qq20150814-2](https://cloud.githubusercontent.com/assets/2818951/9275656/31fcd5ba-42d1-11e5-9eb0-d44ec36285f7.png)
Just as shown in pictures above, the small triangle looks so uncomfortable, and similar problem also appears in Terminal.app, with just same configuration.
iTerm version: Build 2.1.1
Oh my zsh and the themes are latest.

apjanke:
What version of OS X are you running? Is that the `agnoster` theme?

FinalTheory:
Mac OS 10.10.5
The agnoster theme

apjanke:
This looks like you're running in to the effects of "minimum contrast enforcement". This is a terminal emulator feature that ensures text is readable by adjusting foreground colors that are too similar to the background color of the window (in Terminal's case) or the character cell (in iTerm's case, I think). Powerline themes make that right-pointing segment shape by drawing a special right-pointing triangle character with a foreground color that is the same as the previous segment's background color.

You can see it in action by adjusting the background color. In Terminal, take a look at the "Window" tab under the Profile, select the Background Color & Effects, and adjust the color and opacity slider. You'll see the green triangle pop in and out of a brightened mode when you have a dark background that's like the green color.

In iTerm2, you can adjust this feature with the "Minimum contrast" slider in the Colors section of your profile. Terminal.app doesn't seem to have any way to turn it off or control it. You may just have to fiddle with your background color until you get one that doesn't trigger it.

FinalTheory:
Wow, that just work fine!
I spent a lot of time trying to find which option controls this behavior, but got nothing.
So I recommend to add your notification to the wiki page to remind others about this thing.
And thanks very much for your reply.

PR #4287, opened 2015-08-27
Better readability for the 'agnoster' theme

lfuelling:
I changed the current working dir color from black to white to improve readability.

This is what it looked like before:

be4

This is what it looks like now:

after

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -145,7 +145,7 @@ prompt_hg() {
 
 # Dir: current working directory
 prompt_dir() {
-  prompt_segment blue black '%~'
+  prompt_segment blue white '%~'
 }
 
 # Virtualenv: current working virtualenv


mcornella:
I believe this is too much dependent on the terminal color scheme... Your blue is not really blue, but a dark violet. It should look more like [in the example screenshot](https://cloud.githubusercontent.com/assets/2618447/6316862/70f58fb6-ba03-11e4-82c9-c083bf9a6574.png). 

In my setup, for instance, this change would make it look worse, and so would on a lot of people's setups as well. My suggestion is that you either change your color scheme or you can also override the `prompt_dir` definition in your zshrc file, by putting the new definition at the bottom of the file, like so:
```zsh
# all content at the top...
# the function definition at the bottom
prompt_dir() {
    prompt_segment blue white '%~'
}
```

apjanke:
That looks like the default ANSI blue color. The violet triangle is probably due to [minimum contrast enforcement](https://github.com/robbyrussell/oh-my-zsh/issues/4246) in the terminal lightening it.

The agnoster theme specifically targets the [Solarized color scheme](http://ethanschoonover.com/solarized), which is rather different from the default ANSI colors. If you're going to use agnoster, try setting up a profile that uses Solarized and see how that works for you.

lfuelling:
@apjanke I was already using the solarized dark color scheme. But I just reinstalled it and now I seem to have the Solarized colors. Thanks for the help with the contrast enforcement. 

@mcornella thanks for the tip, I will do it this way.

You can close the request, if you want.

mcornella:
You're the only one who can close this though. Right next to the comment button. 
Thanks!

PR #4291, opened 2015-08-28
Call correct ssh-add command

lespea:
Instead of hard-coding ssh-add you should use /usr/bin/env so it runs the correct one (for instance if you installed openssh with brew on mac the old version breaks)


mcornella:
You should remove that /usr/bin stuff entirely and let the `$PATH` do its job. It should work the same way in theory

lespea:
Yeah that makes way more sense; fixed

PR #4301, opened 2015-08-31
thefuck-alias is deprecated. Use thefuck --alias instead.

scoop:


robbyrussell:
Sorry, I missed this one and merged #4302 already

PR #4304, opened 2015-09-01
Fix gbda alias to support `color.ui = always` + exclude dev branches

cmalard:
The alias `gbda` actually doesn't work if `git config color.ui = always`.
Some invisible characters are included in branch names wich causes the following :
- master branch is not excluded
- on deletion, the output is : `error: branch 'x' not found.` And there is no deletion at all.

The `--no-color` option solves this problem.

Also, `command` seems unnecessary.

I excluded `develop` and `dev` branches to make it usable with Git Flow. Should I do this in another PR or is it ok to keep it here ?

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -53,7 +53,7 @@ alias gapa='git add --patch'
 
 alias gb='git branch'
 alias gba='git branch -a'
-alias gbda='git branch --merged | command grep -vE "^(\*|\s*master\s*$)" | command xargs -n 1 git branch -d'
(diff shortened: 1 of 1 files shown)


cmalard:
@mcornella : why `new` ? This is an update of an existing alias.

What should I do to accelerate the merge process ? :-)

mcornella:
I did some bulk-labeling so this one must have marked new by mistake. The `command` part is to ignore aliases or functions named `grep` and run the command directly, if you readd that I'll merge this right away.

mcornella:
The branch exclusion of `develop` and `dev` is fine.

cmalard:
`command` is back and I simplified the RegEx :-)

mcornella:
Perfect! :+1:

PR #4305, opened 2015-09-01
[plugin] adds clipboard plugin

pjvds:
The clipboard plugin adds a `copy` and `paste` alias that works in every environment.

## Copy and paste

``` zsh
$ echo "foo" | copy
$ paste
foo
```

## Copy from file

``` zsh
$ echo "foo" > document.txt
$ cat document.txt | copy
$ paste
foo
```

## Paste to file

``` zsh
$ echo "foo" | copy
$ paste > clipboard.txt
$ cat clipboard.txt
foo
```

## When tools are missing

``` zsh
No cliboard util found!
Please install one of the following utils:
   pbcopy
   xsel
   xclip
$
```

### DIFF
--- a/plugins/clipboard/clipboard.plugin.zsh
+++ b/plugins/clipboard/clipboard.plugin.zsh
@@ -1,0 +1,23 @@
+# pbcopy is expected to be available on Mac OSX
+if type pbcopy &> /dev/null; then
+  alias copy="pbcopy"
+  alias paste="pbpaste"
+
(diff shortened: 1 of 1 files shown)


mcornella:
We'll add those as core functions, see #4254. Please close

pjvds:
@mcornella, that is great. Looking forward to it.

mcornella:
Thanks, if you think of something else to be considered don't hesitate to post it there.

PR #4310, opened 2015-09-02
Add codeclimate autocomplete plugin

ablyler:


robbyrussell:
Thanks! can you provide a README, too?

PR #4315, opened 2015-09-02
Added Fancy-ctrl-z plugin

mbologna:
Use Ctrl-Z to switch back to Vim
I frequently need to execute random command in my shell. To achieve it I pause Vim by pressing Ctrl-z, type command and press fg<Enter> to switch back to Vim.

The fg part really hurt sme. I just wanted to hit Ctrl-z once again to get back to Vim. I could not find a solution, so I developed one on my own that works wonderfully with ZSH:

Source: http://sheerun.net/2014/03/21/how-to-boost-your-vim-productivity/

### DIFF
--- a/plugins/fancy-ctrl-z/README.md
+++ b/plugins/fancy-ctrl-z/README.md
@@ -1,0 +1,14 @@
+# Use Ctrl-Z to switch back to Vim
+
+I frequently need to execute random command in my shell. To achieve it I pause 
+Vim by pressing Ctrl-z, type command and press fg<Enter> to switch back to Vim.
+The fg part really hurt sme. I just wanted to hit Ctrl-z once again to get back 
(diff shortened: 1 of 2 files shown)


mbologna:
@mcornella - check this out, thanks!

mcornella:
This is already made into a plugin: http://sheerun.net/2014/03/21/how-to-boost-your-vim-productivity/#comment-2058519363

mbologna:
What about integrating into omz source code with this PR?

mcornella:
Ok, this works well. Please add a README file with a description instead of the comments in the `plugin.zsh` file, as well as crediting @sheerun.

mbologna:
@mcornella done, thanks for your suggestion :+1:

mcornella:
Ok, :shipit: 
/cc @robbyrussell

PR #4322, opened 2015-09-04
Fixed an edge case for grep - install breaks on Linux

ffxsam:
Users may have an option set for grep to use color.
This was inserting escape characters in "/usr/bin/zsh"

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -44,7 +44,7 @@ mv -f ~/.zshrc-omztemp ~/.zshrc
 TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
 if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
     echo "\033[0;34mTime to change your default shell to zsh!\033[0m"
-    chsh -s $(grep /zsh$ /etc/shells | tail -1)
+    chsh -s $(grep --color=never /zsh$ /etc/shells | tail -1)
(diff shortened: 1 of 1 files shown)


mcornella:
That's weird, how are you running the setup instructions (with which command)?

ffxsam:
The curl one. The setup is fine, I followed the instructions verbatim. It's just that the install script couldn't have predicted that I have an env var (`GREP_COLOR`) so that grep always uses color coding. And those escape characters were getting passed into `chsh` and causing an error.

apjanke:
That is strange. Setting `$GREP_COLOR` does not unconditionally enable coloring, at least where I'm familiar with it (e.g. on OS X and Debian Linux). It just specifies which color to use when coloring is on. Perhaps you have an alias or function set up to make `grep` do `--color=always`? (If it was the more common `--color=auto`, then it would only do colorization when outputting to a tty; subshell output capturing or piping it to `tail` would disable the colorization.)

Doing this change to use `grep --color` could break this for platforms where `grep` doesn't have a `--color` option. Would probably be better to defend against it by reverting whatever env var or alias etc is causing it to do the coloring.

This is something you might want to address in your system's setup. Other things that are calling out to `grep` are not going to be expecting it to add coloring, either.

ffxsam:
Ahh, you're right! It was an alias, not an env var. My mistake. :) Closing!

PR #4337, opened 2015-09-10
fix url-quote-magic not working issue

iCross:
Incompatibilites between 5.0.8 and 5.1:
bracketed-paste-magic may also be necessary in order to apply url-quote-magic.

Reference: 
https://github.com/zsh-users/zsh/blob/a9df6aaa702abf761b155cd842a7f6917be44139/Functions/Zle/url-quote-magic#L11


mcornella:
:+1: /cc @robbyrussell

PR #4338, opened 2015-09-10
nvm: alternate install location support, distinguish non-NVM versions, add README

### LINKED ISSUE #4296: nvm_prompt_info doesn't work with Homebrew installed nvm
The command `nvm_prompt_info` returns an empty string for me even though I have nvm installed (with Homebrew). It looks to me like there's a bug in lib/nvm.zsh in that it bails out if it can't find nvm at `$HOME/.nvm/nvm.sh` whereas my nvm is installed at `/usr/local/opt/nvm/nvm.sh`. On closer inspection it also looks like plugins/nvm/nvm.plugin.zsh and plugins/nvm/_nvm also make incorrect assumptions about the path where nvm is installed.

### LINKED ISSUE #4336: node version stucked at the terminal even after nvm deactivate
I'm using nvm at my system.

If I run `nvm deactivate` and `which node` for confirming I'm using the system node, the node version keep stucked in my terminal. For exemple:

```
ayrton@localhost ~/workspace/whatever/something-ng                      [14:27:12] 
> $                                                                ⬡ 0.12.7 [±develop]
```

Also, if I open a new terminal, this information keep stucked forever in my terminal:

```
ayrton@localhost ~/workspace/whatever/something-ng                      [14:30:07] 
> $                                                                ⬡ 0.12.7 
```

apjanke:
STATUS: Do not merge yet!

Fixes #4296
Fixes #4336
Closes #3659
Closes #3822 (addresses same install location issue, but only adds `$NVM_DIR` to locations it looks)
Closes #2312 (addresses the same problems, but with a substantially different implementation and behavior)
- Adds support for alternate NVM installation locations, including Mac Homebrew and user-defined `$NVM_DIR`. This is both in the plugin (which can load NVM from multiple locations now) and `nvm_prompt_info()` (which just defers to the `nvm` command to determine if it's installed and what environment is active)
- Uses `nvm current` to get version names that distinguish "system" and "none" Nodes from the specific versions installed in NVM.
### lib/nvm.zsh

Use `which nvm` and `nvm current` instead of a hardcoded path to the NVM install and `node -v`.

The `which rvm` makes it work regardless of where nvm is loaded from. 

Switching to `nvm current` instead of `node -v` nvm's version strings, which distinguish the "system" and "none" NVM environments, instead of reporting the specific version of the system node.js or erroring, respectively. This lets you differentiate from the system-wide version of node (called "system") and an NVM-managed Node instance (called "vX.Y.Z") even if they happen to have the same versions of Node.
### plugins/nvm

Searches for NVM in multiple well-known locations.

Adds support for optionally using NVM's own completion definitions, instead of the plugin's `_nvm`. I think we might want to switch over to that entirely in the future, or expand `_nvm` to match its functionality. Our `_nvm` is missing some options (which may only appear in some versions of NVM) and does not complete version names against the Node versions installed in NVM.

Adds a README.


apjanke:
Oops... looks like `$NVM_DIR` is the data directory (where NVM-controlled Node.js versions live), which defaults to the same directory that NVM itself is installed in, but isn't necessarily the same. Will need to break out the installation-locating env variable to a separate variable, like `$NVM_INSTALL_DIR` or `$NVM_HOME` or `$NVM_BIN_DIR`.

apjanke:
jeffcharles [confirms](https://github.com/robbyrussell/oh-my-zsh/issues/4296#issuecomment-139820378) this fixes #4296 for him.

I still need to do a bit more testing on the NVM_DIR vs NVM_INSTALL_DIR distinction.

mischah:
@apjanke Thanks for your efforts. Your PR works like a charm for me :sparkling_heart: 

My scenario:

- latest `oh my zsh`
- lates `nvm` via homebrew
- latest OS X
- Terminal.app 

Is there something missing before this gets merged?

apjanke:
Well... there's no outstanding bugs or features to be done; I'm just not confident in my understanding of the semantics of `nvm`'s environment variables, since I'm not an `nvm` user myself. I was hoping to do a bit more documentation reading or get review from people who know `nvm` before merging this. If you're familiar with `nvm` and this looks right to you, that helps.

mischah:
I’m pretty new to this nvm thing to be honest :relaxed: 

What about @robbyrussell?

mcornella:
We went with `NVM_DIR` on https://github.com/robbyrussell/oh-my-zsh/pull/5265 (it was the most suggested solution, if you think it's more appropriate we can change to `NVM_INSTALL_DIR`). This should be rebased.

apjanke:
Rebased, and modified to reflect what I think is correct post-#5265.

I think using `$NVM_DIR` that way is a mistake. It's already well-defined by NVM to mean your data directory, where your NVM-managed virtual environments are set up, not necessarily where the `nvm` or `nvm.sh` tools are installed. By default they're the same, but they can differ, and in the cases when they do differ, `$NVM_DIR` is not going to be where `nvm.sh` is. That's why I split them in to two variables in the first place.

The code in this PR will still look for `nvm.sh` under `$NVM_DIR` if `$NVM_DIR` is defined, so it will still support that use case. It just doesn't limit the search to there.

joaogbcravo:
👍 Please merge

ELLIOTTCABLE:
Any update on this? It looks good, and solves the problem. `=)`

jamesstonehill:
#5265 doesn't fix this issue.

My path finds the homebrew neovim before the other neoim, but `$NVM_DIR` still points to the neovim installed in ~/.nvm.
```
➜  ~ which nvim
/usr/local/bin/nvim
➜  ~ echo $NVM_DIR
/Users/James/.nvm
```

Not sure if this PR is the correct fix, but figured I'd flag this anyway.

larson-carter:
@apjanke Great work!, can you please rebase to master?

apjanke:
Uhh, actually, I'm not sure? It's been a while, and this repo has had a rename since I forked it, and I can't find my original forked repo on GitHub, and I'm not a Git or GitHub master, so... the relationship between my forked repo and this parent repo seems to have been lost. I can't find the branch for this PR, and `hub checkout <pr-url>`, which is what I usually use isn't working, so... maybe... no?

If I have the time, I'll just reconstruct this PR from the diffs available from GitHub on this PR page. Or if anybody else wants to do that work (and take committer credit), feel free!

larson-carter:
@apjanke Ok, I'm going to close this PR due to the inactive status of it. If you or someone else wants to implement these features you can just open a new PR.

PR #4340, opened 2015-09-10
Fixed broken git branch icon

pahnev:
This will fix the git branch icon not showing in the Amuse theme.

### DIFF
--- a/themes/amuse.zsh-theme
+++ b/themes/amuse.zsh-theme
@@ -12,7 +12,8 @@ PROMPT='
 %{$fg_bold[green]%}${PWD/#$HOME/~}%{$reset_color%}$(git_prompt_info) ⌚ %{$fg_bold[red]%}%*%{$reset_color%}
 $ '
 
-ZSH_THEME_GIT_PROMPT_PREFIX=" on %{$fg[magenta]%}⭠ "
+# Must use Powerline font, for \uE0A0 to render.
+ZSH_THEME_GIT_PROMPT_PREFIX=" on %{$fg[magenta]%}\uE0A0 "
 ZSH_THEME_GIT_PROMPT_SUFFIX="%{$reset_color%}"
 ZSH_THEME_GIT_PROMPT_DIRTY="%{$fg[red]%}!"
 ZSH_THEME_GIT_PROMPT_UNTRACKED="%{$fg[green]%}?"


apjanke:
This change looks like it switches from a standard Unicode ["leftward-triangle" character (U+2B60)](http://www.fileformat.info/info/unicode/char/2b60/index.htm) to the special private-use Powerline "branch" symbol. After it's applied, the theme will only work with special Powerline-patched fonts. Is that intentional?

(Not saying this is wrong; just trying to understand what's going on here. This sounds consistent with switching from the old Powerline code points to the new ones, as discussed in #4065.)

pahnev:
I would think so, because this theme seems to be designed for only Powerline fonts. But none of those fonts actually render correctly.

From https://github.com/robbyrussell/oh-my-zsh/wiki/Themes the description for this theme: 
`Shown in the screenshot with tmux and the powerline plugin (you might need to install one of the patched powerline fonts for it to look the same).`

apjanke:
Ah, yep, that makes sense. Would you mind using this PR to add a comment to the theme definition indicating that it uses Powerline fonts, to make the situation clearer to future maintainers?

pahnev:
Where would I add that definition? To the wiki page?

apjanke:
To the `amuze.zsh-theme` file, as a comment near the top. (But not at the very top, so as not to interfere with that editor mode line.) That way people looking at that private-use `\uE0A0` escape sequence will also be looking at the comment that explains how it should be interpreted.

pahnev:
Alright added comment above that line where `\ uE0A0` is used

apjanke:
Looks good to me.

Tested on OS X with a current Powerline-patched font, and the new code point shows up as the branch icon.

screen shot 2015-09-11 at 3 21 24 am

:+1:

@mcornella or @ncanceill, can we get a code review and merge on this?

mcornella:
Clear as cristal. As is now, I also get the left-arrow sign when [the original author intended a branch icon there](https://github.com/robbyrussell/oh-my-zsh/wiki/Themes#amuse). 

:+1: /cc @robbyrussell

apjanke:
@mcornella Cool. That's consistent: the old Powerline glyph mappings were using some non-private-use code points, which eventually got implemented by Unicode and systems supporting it, which is the whole reason for the change in the first place. This PR should fix that.

PR #4343, opened 2015-09-11
jira: add README.md

apjanke:
Adds a README.md based on existing comments.

Adds "emulate -L zsh" to a couple functions to defend against possible quoting issues.
Adds a "dumpconfig" action. (The "dumpconfig" action is for verifying the configuration behavior against what's documented in the README now, which is why I included this new behavior in a "add README" PR.)
Adds "error" to some of the error messages.


robbyrussell:
Thanks, @apjanke !

PR #4348, opened 2015-09-14
Display song currently playing in iTunes

andybateman:
Added function to display the song you're currently listening to in iTunes

 - itunes playing

osx.plugin.zsh

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -166,7 +166,7 @@ function itunes() {
 			if [ "$state" = "playing" ]; then
 				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
 				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
-				echo "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
+				print -P "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
 			fi
 			return 0
 			;;
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -161,6 +161,15 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [ "$state" = "playing" ]; then
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
+				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
+				echo "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
+			fi
+			return 0
+			;;
 		shuf|shuff|shuffle)
 			# The shuffle property of current playlist can't be changed in iTunes 12,
 			# so this workaround uses AppleScript to simulate user input instead.
@@ -199,6 +208,7 @@ EOF
 			echo "\tnext|previous\tplay next or previous track"
 			echo "\tshuf|shuffle [on|off|toggle]\tSet shuffled playback. Default: toggle. Note: toggle doesn't support the MiniPlayer."
 			echo "\tvol\tSet the volume, takes an argument from 0 to 100"
+			echo "\tplaying\tShow what song is currently playing in iTunes."
 			echo "\thelp\tshow this message and exit"
 			return 0
 			;;
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -161,9 +161,9 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
-		playing)
+		playing|status)
 			local state=`osascript -e 'tell application "iTunes" to player state as string'`
-			if [ "$state" = "playing" ]; then
+			if [[ "$state" = "playing" ]]; then
 				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
 				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
 				print -P "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -167,6 +167,8 @@ function itunes() {
 				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
 				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
 				print -P "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
+			else
+				print "iTunes is" $state;
 			fi
 			return 0
 			;;
@@ -208,7 +210,7 @@ EOF
 			echo "\tnext|previous\tplay next or previous track"
 			echo "\tshuf|shuffle [on|off|toggle]\tSet shuffled playback. Default: toggle. Note: toggle doesn't support the MiniPlayer."
 			echo "\tvol\tSet the volume, takes an argument from 0 to 100"
-			echo "\tplaying\tShow what song is currently playing in iTunes."
+			echo "\tplaying|status\tShow what song is currently playing in iTunes."
 			echo "\thelp\tshow this message and exit"
 			return 0
 			;;
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -166,9 +166,9 @@ function itunes() {
 			if [[ "$state" = "playing" ]]; then
 				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
 				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
-				print -P "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
+				echo -E "Listening to $fg[yellow]$currenttrack$reset_color by $fg[yellow]$currentartist$reset_color";
 			else
-				print "iTunes is" $state;
+				echo "iTunes is" $state;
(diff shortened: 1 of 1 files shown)


plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,15 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [ "$state" = "playing" ]; then
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
+				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
+				echo "You're currently litening to $fg[yellow]$currenttrack$reset_color by $fg[yellow]$currentartist$reset_color";
```
mcornella: There's a spelling error: litening → listening

I would also use [Prompt expansion colors](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Visual-effects) instead of `$fg`/`$reset_color`, as well as drop the `You're currently...` part, like so:
`print -P "Listening to '%F{yellow}Melancholy Hill%f' by %F{blue}Gorillaz%f"`

I cannot try if the osascript works because I'm on Linux, but @apjanke can confirm.
:+1:

andybateman: All good thoughts! Thanks!

apjanke: Confirmed: works for me on OS X 10.9.5.

screen shot 2015-09-16 at 3 28 24 pm

If you use the prompt expansion colors, you will need to escape the song and artist names in case they contain "%" or other prompt expansion escape characters. As it is, with `echo`, you should probably use `echo -E` in case they contain backslashes, too.

And if you want to be super hygienic, check to make sure the output is connected to a tty using `[ -t 1 ]` and only use the color control sequences if you're in fact outputting to a tty, to avoid contaminating output captured to files or other variables.

plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,15 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [ "$state" = "playing" ]; then
```
apjanke: Could use the zsh double-brackets here: `[[ $state == "playing" ]]`. It's preferred OMZ style; the single-bracket `[ ]` is more for portability.

plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,15 @@ function itunes() {
...
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
```
apjanke: `currenttrack` and `currentartist` should be `local`.

plugins/osx/osx.plugin.zsh
```diff
@@ -199,6 +208,7 @@ EOF
 			echo "\tnext|previous\tplay next or previous track"
 			echo "\tshuf|shuffle [on|off|toggle]\tSet shuffled playback. Default: toggle. Note: toggle doesn't support the MiniPlayer."
 			echo "\tvol\tSet the volume, takes an argument from 0 to 100"
+			echo "\tplaying\tShow what song is currently playing in iTunes."
```
apjanke: Could replace this whole series of `echo`s with ` <<EOF` here-document like the other usage sections. Would be more readable.

plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,15 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [ "$state" = "playing" ]; then
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
+				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
+				echo "Listening to $fg[yellow]$currenttrack$reset_color by $fg[yellow]$currentartist$reset_color";
```
apjanke: Should use `echo -E` just in case song or artist names contain backslashes.

plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,15 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [ "$state" = "playing" ]; then
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
+				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
+				echo "Listening to %F{yellow}$currenttrack%f by %F{yellow}$currentartist%f";
```
apjanke: The `%F` sequences only work with `print -P` not with regular `echo`.

screen shot 2015-09-16 at 3 41 04 pm

andybateman: Thanks @mcornella and @apjanke for your help, made some of the changes suggested.

andybateman:
Thanks @mcornella and @apjanke for your help, made some of the changes suggested.

apjanke:
Cool. Most of the changes look good.

The song and artist names still need to be quoted before passing to `print -P`, since they might contain escape sequences.

Consider an IDM or witch house fan with some funky song names in their collection.

screen shot 2015-09-17 at 3 56 26 pm

This doesn't interact well with the current code.

screen shot 2015-09-17 at 3 57 14 pm

Note that in addition to bad output, I got it to dump data from my private `~/.ssh` directory outside the prompt. The `$(...)` prompt expansion sequence can run arbitrary code; `print -P` is `eval`-equivalent. It could just as well have emailed or ftp'ed those keys to some random server.

This is a security problem. It wouldn't be hard for an attacker hosting audio data to craft metadata that would seriously compromise someone using this `itunes playing`. Or for weird/joke audio tracks to do something bad accidentally. And we all know that users get some of their audio files from sketchy sources.

Because audio metadata can come from the outside world (mp3s from elsewhere, streaming audio, etc), it should be considered unsanitized input, and needs to be strictly sanitized or escaped before passing to prompt expansion or other `eval` risks. I'm not confident we're capable of doing that; it's tricky. (`zsh` provides some quoting operators for parameter expansion, which would help, but you still have to be careful how you use them.)

I think it would be better to go back to using `echo -E` here, even if it means using the clumsier `$fg[...]` stuff instead of the `%F` sequences.

mcornella:
Hahaha I never thought about this before. I would guess `${(q)currenttrack}` oughta do the trick, but if `echo -E` is considered safe then let's definitely use that.

apjanke:
:)

I don't think `${(q)...}` works in this particular case, because you have to quote for the prompt expansion, not just shell interpretation. 

```
$ foo='%d%x$(danger)'
$ print -P "$foo"
zsh: command not found: danger
/Users/janke-zsh
$ print -P "${(q)foo}"
zsh: command not found: danger
/Users/janke-zsh
$ echo -E "$foo"
%d%x$(danger)
$
```

It doesn't escape the `%` characters (because they're special to `print -P`, but not special to the shell itself). And the `$(...)` could still get executed because its evaluation happens another step down inside the `print -P` call, not at the command level, so you'd need one more level of quoting. Or something. This multi-level `eval` and quoting stuff is hard to reason about.

I'm pretty sure that `echo -E` is known safe, and easier to reason about. Let's just stick with that.

andybateman:
@apjanke @mcornella I've moved it back to using 'echo', tried many variations of sanitation with 'print' but none were as tidy as just using 'echo -E'.

Tested with a variety of crazy track and artist names.

plugins/osx/osx.plugin.zsh
```diff
@@ -161,6 +161,17 @@ function itunes() {
 		vol)
 			opt="set sound volume to $1" #$1 Due to the shift
 			;;
+		playing|status)
+			local state=`osascript -e 'tell application "iTunes" to player state as string'`
+			if [[ "$state" = "playing" ]]; then
+				currenttrack=`osascript -e 'tell application "iTunes" to name of current track as string'`
+				currentartist=`osascript -e 'tell application "iTunes" to artist of current track as string'`
+				echo -E "Listening to $fg[yellow]$currenttrack$reset_color by $fg[yellow]$currentartist";
```
apjanke: Should close this with another `$reset_color` to avoid contaminating subsequent output that happens before a prompt or other colorizing output resets it.

screen shot 2015-09-20 at 6 03 01 pm

apjanke:
Looks good to me. (Aside from needing a final `$reset_color`.) I tried some weird track names too and they were all fine.

andybateman:
@apjanke Do you think it needs a $reset_color? I'd removed it as after reading through about using coloured text with echo, the $reset_color is implied at the end of the echo.  Do you disagree?

apjanke:
I disagree; there is no implicit color reset done by `echo`. It only works that way if you run `echo` as the only command in a sublist. It is the prompt that is issued before the next command which changes the color back. All `echo` does is output exactly the characters you give it (possibly subject to escape processing, and with an EOL).

Take a look at my screen shot: that `Hello world` from an entirely separate `echo` command gets colored yellow, too. Here's another example, where the coloring spills over to the output of `ls`, until `ls`'s own colorization changes it to something else.

screen shot 2015-09-20 at 7 50 57 pm


You may see other code doing this (leaving off a color reset at the end of output), but it's either just being sloppy, or intended for use in specific contexts where its caller will make sure a reset or subsequent color change is issued. IMHO, if you read documentation or examples that weren't doing a reset, that doco is sloppy too and should probably be changed.

andybateman:
Well @apjanke at this point I trust you more than other doco :)

Especially as you provided examples!

mcornella:
Great, this is ready to go then! One last thing: can you squash your commits to make it into one? If you don't know how to do it let me know.

apjanke:
:+1:

mcornella:
Better, :+1: 
/cc @robbyrussell

PR #4354, opened 2015-09-16
chruby: donʼt try to execute .sh files before loading

henrikhodne:
Iʼm not quite sure why this was here in the first place, but it was causing these errors for me:

```
_source_from_omz_settings:4: permission denied: /usr/share/chruby/chruby.sh
_source_from_omz_settings:8: permission denied: /usr/share/chruby/auto.sh
```

As far as I can tell, it should be enough to just check if the file exists and is readable?


mcornella:
:+1: There's no reason for the first part, that will run it, in fact.
/cc @robbyrussell

PR #4369, opened 2015-09-18
Rewrite of emoji-clock plugin based on emoji plugin

afh:


mcornella:
Hi @afh, thank you for your contribution. We better not add a dependency even if it means repeating the emojis. Otherwise, people who use this plugin will find that it doesn't work because they don't have the emoji plugin enabled, and we don't have an easy way to track or manage plugin dependencies yet.

afh:
I understand and had similar thoughts. Yet from a developer's perspective I find the revised implementation so much cleaner.

Is tracking or managing plugin dependencies on the oh-my-zsh roadmap?

mcornella:
> I understand and had similar thoughts. Yet from a developer's perspective I find the revised implementation so much cleaner.

I agree that one simple logic is better than a `case` switch. Back when I first encountered this plugin I thought of having an associative array of clocks but never got to implement it. Maybe you can go this way? I like that you're using prompt sequences instead of date which is an external command. 
So :+1: to this PR but lose the dependency on `emoji`.

> Is tracking or managing plugin dependencies on the oh-my-zsh roadmap?

There was some talk about it, mostly around having a better loading system, and I think we let the discussion dry out because there were already solutions out there. There's `antigen` which is a great tool and FWIW it could be used by oh-my-zsh like `node` uses `npm`, `ruby` uses `bundler`, etc. But no, there's nothing on the roadmap both in terms of discussion and of implementation. Maybe you can get the ball rolling? :wink:

afh:
I amended this pull request, so it uses a associative array for the emojis, copied from the emoji plugin
instead depending on the latter.

I'll have a look at `antigen`, thanks. Where would be the right place to get the :soccer: rolling?

mcornella:
Actually, I think the current code is cleaner and clearer.

PR #4373, opened 2015-09-20
add dotenv plugin

arteezy:
This plugin automatically loads project specific ENV variables from `.env` file when you `cd` into project root folder.

Storing project configuration in the environment becomes very popular lately. It's is one of the tenets of a [twelve-factor app](http://www.12factor.net). Anything that is likely to change between deployment environments, such as resource handles for databases or credentials for external services, should be extracted from the code into environment variables.

There already a lot of projects who embrace this type of configuration and `.env` file format:
[Ruby dotenv](https://github.com/bkeepers/dotenv)
[Node.js dotenv](https://github.com/motdotla/dotenv)
[PHP dotenv](https://github.com/vlucas/phpdotenv)

It's also [recommended by Heroku](https://devcenter.heroku.com/articles/heroku-local#set-up-your-local-environment-variables).

However, these projects are not suitable for shell-heavy work. They are bound to their language platform. If you often need to run small scripts, different language tools or daemons from terminal it's much more comfortable to preload ENV vars into current shell session. This plugin helps to do that automatically.

Although this plugin is very simple code-wise, I think it could be a part of oh-my-zsh core plugin list.

README file included.

### DIFF
--- a/plugins/dotenv/README.md
+++ b/plugins/dotenv/README.md
@@ -1,0 +1,34 @@
+# dotenv
+
+Automatically load your project ENV variables from `.env` file when you `cd` into project root directory.
+
+Storing configuration in the environment is one of the tenets of a [twelve-factor app](http://www.12factor.net). Anything that is likely to change between deployment environments–such as resource handles for databases or credentials for external services–should be extracted from the code into environment variables.
+
+## Installation
+
+Just add the plugin to your `.zshrc`:
+
+```sh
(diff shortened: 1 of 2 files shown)


nkostelnik:
Please merge, this is awesome 👍

joost:
+1 +1 +1 very elegant solution. Also useful to load variables used for a project by certain cli tools.

PR #4375, opened 2015-09-21
Fix capistrano problem when generating autocompletion

PrestanceDesign:
``shipit -v``, print only Capistrano version instead output all tasks when generating .cap_tasks\~ file.

### DIFF
--- a/plugins/capistrano/_capistrano
+++ b/plugins/capistrano/_capistrano
@@ -14,7 +14,7 @@ _arguments -C \
 _cap_tasks() {
   if [[ -f config/deploy.rb || -f Capfile ]]; then
     if [[ ! -f .cap_tasks~ ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
@phstc #2471

phstc:
@PrestanceDesign `-v` is for verbose mode, not for version.

```
cap --help
    -v, --verbose                    Log message to standard output.
    -V, --version                    Display the program version.

cap -V
Capistrano Version: 3.2.1 (Rake Version: 10.3.2)
```

Which version of Capistrano are you using?

PrestanceDesign:
@phstc my version is 3.4.0. This version seems to remove/replace verbose `-v`

phstc:
@PrestanceDesign alright, I tested it without `-v` and it also works for me. Thanks

@mcornella it's good to go :shipit:

mcornella:
:+1: /cc @robbyrussell

Issue #4376, opened 2015-09-21
install.sh fails if tput isn't available


fcrozat:
on some reduced system (Synology NAS), tput isn't available, causing install.sh script to fail to run. Fix incoming

mcornella:
@apjanke

apjanke:
Wow, I didn't even know there were systems skinny enough to not have `tput`.

The fix looks good to me. I don't have a `tput`-less system to test on right now, but I tested it on OS X to at least verify that it doesn't interfere with the normal case.

We should do the same thing to `tools/upgrade.sh` so you don't get surprised by the same problem a week from now when OMZ updates itself.

apjanke:
Can you close this if the PR worked?

WhyEee:
@fcrozat's original fix assumes which not to output anything to STDOUT in case the command is not found. That is not necessarily true on all systems. A better solution is to check the return value instead.

PR #4377, opened 2015-09-21
Do not fail if tput is not available

fcrozat:
Fixes issue #4376

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -2,7 +2,10 @@ set -e
 
 # Use colors, but only if connected to a terminal, and that terminal
 # supports them.
-ncolors=$(tput colors)
+tput=$(which tput)
+if [ -n "$tput" ]; then
+    ncolors=$($tput colors)
+fi
 if [ -t 1 ] && [ -n "$ncolors" ] && [ "$ncolors" -ge 8 ]; then
   RED="$(tput setaf 1)"
   GREEN="$(tput setaf 2)"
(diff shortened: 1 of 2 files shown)


apjanke:
Looks good to me, though I can only test it on the normal case where `tput` is present.

Want to hit [`tools/upgrade.sh`](https://github.com/robbyrussell/oh-my-zsh/blob/918d26e7dbcb907e673df24242c9688cd89019b9/tools/upgrade.sh#L4) with the same thing in this PR?

fcrozat:
PR updated to handle upgrade.sh

mcornella:
:+1: /cc @robbyrussell

apjanke:
BTW, if you phrase it as "Fixes #4376" instead of "Fixes issue #4376", GitHub will recognize it and add an auto-close link between the PR and issue. The extra "issue" in there means it doesn't recognize it.

apjanke:
Also BTW, does this actually work on the Synology? Because of the `set -e`, I think the script might still abort after that `tput=$(which tput)` line, since the `which` returns nonzero.

fcrozat:
I'll fix the commit entry to help github and will update you with status on Synology (didn't double check yet)

apjanke:
FYI, PR #4419 goes even further, ignoring all errors in the color setup. It's pretty blunt, but will get things working for now, even if it's still breaking on the Synology. Turns out, `which tput` isn't sufficient, because this code actually requires a `tput` that supports `terminfo` style capabilities.

PR #4381, opened 2015-09-22
Added mercurial library

mbologna:
The added library parses output of the hg (mercurial) command, and outputs
the same variables (which should be customized by theme) in your prompt.
I added an example of these variables to michelebologna theme.

This patch should uniform all the behaviour of all themes regarding mercurial.
In fact, some themes implement their own hg parsing _inside_ the theme.
This should be better implemented in a library, as happens with git library.

### DIFF
--- a/lib/mercurial.zsh
+++ b/lib/mercurial.zsh
@@ -1,0 +1,32 @@
+# Michele Bologna
+#
+# Implements a hg_prompt_info function that inspect the current mercurial repo
+# (if any) and then outputs the status of the repo, in a similar way with
+# git.zsh library.
+#
+# themes should customize:
+# * ZSH_THEME_HG_PROMPT_UNTRACKED - symbol to show in prompt if untracked 
+# files are present in the mercurial repo
+# * ZSH_THEME_HG_PROMPT_ADDED - symbol to show in prompt if added files are
+# present in the mercurial repo
+# * ZSH_THEME_HG_PROMPT_MODIFIED - symbol to show in prompt if modified files
+# are present in the mercurial repo
+
+hg_prompt_info() 
+{
+	local STATUS=""
+	if $(hg id >/dev/null 2>&1); then
+		local BRANCH=$(hg branch 2>/dev/null)
+		if `hg status | grep -q "^\?"`; then
+			STATUS="$ZSH_THEME_HG_PROMPT_UNTRACKED"
+		fi
+		if `hg status | grep -q "^[A]"`; then
+			STATUS="$ZSH_THEME_HG_PROMPT_ADDED$STATUS"
+		fi
+		if `hg status | grep -q "^[M]"`; then
+			STATUS="$ZSH_THEME_HG_PROMPT_MODIFIED$STATUS"
+		fi
+		echo "$BRANCH$STATUS"
+	fi
+}
+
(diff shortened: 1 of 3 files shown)


mbologna:
@robbyrussell @mcornella any news on this PR?

lib/mercurial.zsh
```diff
@@ -0,0 +1,32 @@
...
+# present in the mercurial repo
+# * ZSH_THEME_HG_PROMPT_MODIFIED - symbol to show in prompt if modified files
+# are present in the mercurial repo
+
+hg_prompt_info() 
+{
+	local STATUS=""
+	if $(hg id >/dev/null 2>&1); then
+		local BRANCH=$(hg branch 2>/dev/null)
+		if `hg status | grep -q "^\?"`; then
```
mcornella: I'd cache the output of `hg status` into a variable. Also, you don't need the backticks here, a simple `if echo $var | grep -q` will also work.

themes/michelebologna.zsh-theme
```diff
@@ -68,8 +68,14 @@ ZSH_THEME_GIT_PROMPT_AHEAD_REMOTE=">"
...
 ZSH_THEME_GIT_PROMPT_DIVERGED_REMOTE="$red<>"
 
+ZSH_THEME_HG_PROMPT_MODIFIED="$red*"
+ZSH_THEME_HG_PROMPT_ADDED="$green+"
+ZSH_THEME_HG_PROMPT_UNTRACKED="$blue%%"
+
 PROMPT='$username_output$hostname_output:$current_dir_output%1(j. [$jobs_bg].)'
 GIT_PROMPT='$(out=$(git_prompt_info)$(git_prompt_status)$(git_remote_status);if [[ -n $out ]]; then printf %s " $white($green$out$white)$reset";fi)'
 PROMPT+="$GIT_PROMPT"
+HG_PROMPT='$(out=$(hg_prompt_info);if [[ -n $out ]]; then printf %s " $white($green$out$white)$reset";fi)'
```
mcornella: I'm sure there's a cleaner way to go about this; and the same with `$GIT_PROMPT`. Have a look at [conditional substrings in prompts](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Conditional-Substrings-in-Prompts).

mcornella:
I've made a few suggestions. These are also implemented in the mercurial plugin in some way, if you want to have a look at it: https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/mercurial/mercurial.plugin.zsh.

mcornella:
Other older PRs that modify hg prompts are #2177 and #2373, and some newer ones like #4399 and #4591.

themes/michelebologna.zsh-theme
mbologna: Unfortunately, zsh offers no prompt conditionals to test for string length. 
Do you have any other ideas?

mbologna:
@mcornella I understand that other PR have been done. This PR address the proliferation of `mercurial` plugins. Why not handling mercurial just inside the native `lib` and standardize prompt variables, as we already do with git?

mbologna:
Closing as I use git more than mercurial nowadays.

PR #4386, opened 2015-09-23
Adds transfer.sh plugin

gabriel-kaam:


plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+function transfer() {
+  if [ $# -eq 0 ]; then
+    echo "No arguments specified. Usage:\necho transfer /tmp/test.md\ncat /tmp/test.md | transfer test.md"
```
apjanke: You can make this a multi-line "here doc" using `<<<EOF` instead of embedded `\n` for better readability. And should that extra `echo` be inside the string?

```
$ transfer
No arguments specified. Usage:
echo transfer /tmp/test.md
cat /tmp/test.md | transfer test.md
```

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+function transfer() {
+  if [ $# -eq 0 ]; then
```
apjanke: The `[[ ... ]]` double-brackets are more conventional zsh/OMZ style.

apjanke: The `[[ ... ]]` double-brackets are more conventional zsh/OMZ style.

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,20 @@
...
+    echo "No arguments specified. Usage:\necho transfer /tmp/test.md\ncat /tmp/test.md | transfer test.md"
+    return 1
+  fi
+
+  tmpfile=$( mktemp -t transferXXX )
+  if tty -s; then
+    basefile=$(basename "$1" | sed -e 's/[^a-zA-Z0-9._-]/-/g')
```
apjanke: You can replace `basename` and `sed` with native zsh parameter expansion (e.g. `${foo:t}` for basename). But since you're using this in a URL, you might want to replace the simple `sed` call with actual URL-escaping using the new `omz_urlencode()` function from #4129. That could get you nicer URLs and avoid collisions for files with special characters in their names.

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+function transfer() {
+  if [ $# -eq 0 ]; then
+    echo "No arguments specified. Usage:\necho transfer /tmp/test.md\ncat /tmp/test.md | transfer test.md"
+    return 1
+  fi
+
+  tmpfile=$( mktemp -t transferXXX )
```
apjanke: The temp file might not be necessary. `curl` writes the progress bar to the tty and not stdout (I think), so you could just capture the URL with `$(curl ... "https://transfer")`

gabriel-kaam:
@apjanke : fixed :)

janisozaur:
will this be merged?

gabriel-kaam:
up :) is there something missing?

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,25 @@
+# From https://transfer.sh/
+# Modified by Gabriel Kaam <gabriel@kaam.fr>
+
+function transfer() {
+  if [[ $# -eq 0 ]]; then
```
apjanke: Sorry I didn't see this earlier: the `$basefile` and `$url` variables (and any others meant for this functions internal use) should be made local, so they don't clobber the user's variables. This can be done by sticking `local basefile url` here at the top of the function. Should probably do a `emulate -L zsh` for portability, too.

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,25 @@
...
+    cat<<EOF
```
apjanke: This error message should probably go on `stderr` instead of `stdout`, using `cat >&2 <<EOF`.

plugins/transfer/transfer.plugin.zsh
```diff
@@ -0,0 +1,25 @@
...
+No arguments specified. Usage:
+transfer /tmp/test.md
+cat /tmp/test.md | transfer test.md
+EOF
+    return 1
+  fi
+
+  basefile=$(omz_urlencode "${1:t}")
+
+  if tty -s; then
```
apjanke: The only difference between the two branches of this `if` are the infile. Maybe it could be factored out, and construct the `curl` command in one place? Could help readability, and make it easier to add additional options later on. 

And `tty -s` seems like the wrong test to do here. You might be calling `transfer` inside a script or function which itself is hooked up to a pipeline, but still want to use the form where you give it a filename instead of using `stdin`. Alternately, if you allow the no-arg `stdin` form from the terminal as well, then a user could run transfer, and type or paste stuff from the keyboard to go in to the uploaded file.

How about something like:

```
# Default to reading stdin
local infile="$1"
basefile=$(omz_urlencode "${infile:t}")
if [[ -z $infile ]]; then
   infile="-"
fi

url=$(curl --progress-bar --upload-file "$infile" "https://transfer.sh/$basefile")
```

Or, if you wanted to be more friendly to beginning users, when you get no infile specified, you could then check to see if you're in a terminal, and if so give the user an error message, and require them to explicitly pass "-" as the infile if they really want to use `stdin` from the terminal. Otherwise, you can get the case where you're calling `curl --upload-file "$1"` and `$1` is empty, which will look like a normal web page hit and end up dumping out the `transfer.sh` usage instructions instead of an upload result URL, like this.

```
$ curl --progress-bar --upload-file "" "https://transfer.sh/"
transfer.sh: Easy file sharing from the command line
===
made with <3 by DutchCoders

Upload:
$ curl --upload-file ./hello.txt https://transfer.sh/hello.txt

Encrypt & upload:
```

gabriel-kaam:
Had to create a new PR , you can find here : https://github.com/robbyrussell/oh-my-zsh/pull/5944

PR #4387, opened 2015-09-23
Uniform git prompt equal remote

mbologna:
Uniform git prompt equal remote to all others prompts, setting the correct symbol
(if applicable) to $git_remote_status, instead of echoing it.

It also solves the problem pointed out in #3911


mcornella:
Please fix the indentation and it's good to go
:+1: /cc @robbyrussell

PR #4391, opened 2015-09-23
installer: use timestamped backups to preserve all old zshrcs

### LINKED ISSUE #2499: Some minor alteration concerning .zshrc.pre-oh-my-zsh
A suggestion, but would it be possible to copy the existing `.zshrc` file to `.zshrc.pre-oh-my-zsh-$(date +%s)` or the likes? I had a hanging install on password, had to kill it, then tried to install again and - of course - it overwrites my previous (correct) .zshrc file.

Obviously, I should have backed it up - and its not _that_ great of a loss, but still. Don't know if this is desirable or if people fear it might clutter up `~` a bit too much on the .zfiles potentially.

### LINKED ISSUE #4390: Please, oh please, ask, or at least warn, about overwriting ~/.zshrc on installation
Today is a little worse than it was before installing oh-my-zsh.

apjanke:
The installer backs up `~/.zshrc` to `~/.zshrc.pre-oh-my-zsh`. But it doesn't back up `~/.zshrc.pre-oh-my-zsh` if it exists, so if you run the installer multiple times, it can destroy your original `~/.zshrc`. We should probably avoid destroying any original user data where at all possible.

This PR add timestamped backups of the old `~/.zshrc.pre-oh-my-zsh` files, so no matter how many times you run the installer, you can still find your original `~/.zshrc` contents, unless you manually delete them yourself.

Fixes #4390 
Fixes #2499

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -58,6 +58,30 @@ main() {
     exit 1
   }
 
+  # Keep most recent old .zshrc at .zshrc.pre-oh-my-zsh, and older ones
+  # with datestamp of installation that moved them aside, so we never actually
+  # destroy a user's original zshrc
+  printf "${BLUE}Looking for an existing zsh config...${NORMAL}\n"
+  # Must use this exact name so uninstall.sh can find it
+  OLD_ZSHRC=~/.zshrc.pre-oh-my-zsh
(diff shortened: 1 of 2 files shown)


mcornella:
Closed by #5169.

PR #4392, opened 2015-09-23
Correct repo dir

ag4ve:
cvs's metadata is under CVS and not .cvs


mcornella:
Source: [https://web.archive.org/web/20130129231119/http://ximbiot.com/cvs/manual/cvs-1.12.13/cvs_2.html#IDX30]
:+1: /cc @robbyrussell

PR #4399, opened 2015-09-24
Speed up the mercurial plugin for large repositories

scottrice:
In very large repositories, certain hg commands can take a very long time. Two of those used in the mercurial plugin are particularly bad - `hg status` and `hg summary`. I removed the `hg summary` call by replacing it with `hg root`, which gives us similar information much faster.

`hg status` was a different beast, and since there is no "faster" replacement I ended up just mirroring the functionality of the git plugin. If you're unfamiliar, the git plugin simply shows the clean prompt if a certain config flag is set (`oh-my-zsh.hide-dirty`). This approach has multiple benefits, the most important being that you can configure it per-repository.

Finally, the multiple `in_hg` calls ended up calling `hg root` a lot. These were not super long calls on their own, but when we do it often we end up spending quite a bit of time calculating an answer we already know. To fix this I saved it in a local variable at the beginning of the `hg_prompt_info` call and remove it at the end.

### DIFF
--- a/plugins/mercurial/mercurial.plugin.zsh
+++ b/plugins/mercurial/mercurial.plugin.zsh
@@ -20,7 +20,7 @@ alias hgca='hg qimport -r tip ; hg qrefresh -e ; hg qfinish tip'
 alias hgun='hg resolve --list'
 
 function in_hg() {
-  if [[ -d .hg ]] || $(hg summary > /dev/null 2>&1); then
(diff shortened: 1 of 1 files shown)


scottrice:
Ping. Any chance of this going in? I've been using it locally for a while, but I'd like to bring my local checkout back in line with oh-my-zsh master.

PR #4408, opened 2015-09-25
Fix invalid test expression.

DoomHammer:
Homebrew installed chruby couldn't be found due to erroneous test expression. This PR fixes said issue.


robbyrussell:
Thanks @DoomHammer !

PR #4418, opened 2015-09-28
Encode64 Plugin - Fix wrong `echo` usage

Gnouc:
Currently, encode64 plugin using `echo -n` to print the content
of `$1` variable. This approach will not work with arbitrary data,
which contains sequence of escaped characters, since when many
`echo` implementation will expand them:
```
encode64 '\t\n'
```
is wrong since when it try to generate base64 of string contains a tab, a newline instead of literal `\t` and `\n`.

This commit change the usage to `printf`, which is builtin in all
POSIX shells and can print arbitrary data reliability:
```
printf '%s' '\t\n' | base64
```
will generate base64 of literal `\t` follow by `\n`


apjanke:
:+1: Tested on OS X 10.9.

mcornella:
:+1: /cc @robbyrussell

Issue #4382, opened 2015-09-22
New zsh pull, messed up my system :( please help. 


rjrobinson:
I updated to the latest version. And it messed me up a little... 

not sure what to do. I have oh-my-zsh in a .dotfile dir

here is my edited zshrc file 
```
echo "You is logged in.... "
export ZSH=$HOME/.dotfiles/oh-my-zsh

export ZSH_THEME="clean"

export CASE_SENSITIVE="false"
plugins=(brew cloudapp colorize compleat dirpersist gem git git-flow github osx mvn node npm nvm rv rails )

source $ZSH/oh-my-zsh.sh

export NVM_DIR=~/.nvm
source /usr/local/opt/nvm/nvm.sh

export PATH="$PATH:bin/subl"
export PATH="$PATH:/Users/rrobin008c/.nvm/versions/node/v0.12.7/bin/node"
export PATH="$PATH:/Applications/Postgres.app/Contents/Versions/9.4/bin"

#NVM
export NVM_DIR=~/.nvm
source $(brew --prefix nvm)/nvm.sh

LUNCHY_DIR=$(dirname `gem which lunchy`)/../extras
  if [ -f $LUNCHY_DIR/lunchy-completion.zsh ]; then
    . $LUNCHY_DIR/lunchy-completion.zsh
  fi
nvm use stable

# export PATH="/usr/local/sbin:$PATH"

export PATH="$PATH:$HOME/.rvm/bin" # Add RVM to PATH for scripting```


Any help would be great. None of my aliases work, In fact, it had over ridden the file. idk what happened....

mcornella:
Can you post the rest of your dotfiles directory?
And what do you mean that "it had overridden the file"? Which file?

apjanke:
> export PATH="$PATH:bin/subl"

That `bin/subl` is a relative path so it won't work in most situations. Maybe it should be `~/bin/subl` instead? You can also omit the `export`s from those `$PATH` setting lines: `$PATH` is already marked for export, and it stays exported through changes to its value.

Can you give us a specific example of what's not working? Like, an alias that was working before, and how it's behaving now? When you say "my aliases", do you mean aliases that you have defined yourself, or aliases defined by OMZ that you use?

The final "```" for quoting code in GitHub markdown needs to be on a line by itself. Right now, the fixed-width formatting is leaking to the rest of your comment.

rjrobinson:
[...]

PR #4419, opened 2015-09-28
installer: ignore errors in color setup, for FreeBSD portability

### LINKED ISSUE #4404: install.sh doesn't work on FreeBSD
It exits after this line: https://github.com/robbyrussell/oh-my-zsh/blob/master/tools/install.sh#L1

I'm not sure what `-e` does, it's hard to search for this command.  Removing this line and the installer works fine.  Wish I had more specific information about why this fails.

apjanke:
This is a hack to keep the installer working on systems without tput/terminfo, including FreeBSD (at least in the default installation for 10.2).

Fixes #4404.
Follows up on #3664.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1,5 +1,3 @@
-set -e
-
 # Use colors, but only if connected to a terminal, and that terminal
 # supports them.
 tput=$(which tput)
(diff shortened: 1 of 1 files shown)


mcornella:
What about the upgrade? #4382 might be due to `set -e` as well in the `upgrade.sh` script.

apjanke:
`upgrade.sh` doesn't do a `set -e`. Wouldn't have affected it.

mcornella:
Let's revisit the installer soon. I feel that recent changes have been a workaround on top of the other and maybe it's time to go full zsh. Thanks!

PR #4421, opened 2015-09-28
gulp plugin: missing opening double quote

jcpst:
This plugin works properly now.

### DIFF
--- a/plugins/gulp/gulp.plugin.zsh
+++ b/plugins/gulp/gulp.plugin.zsh
@@ -19,8 +19,8 @@
 # Grabs all available tasks from the `gulpfile.js`
 # in the current directory.
 #
-function $$gulp_completion() {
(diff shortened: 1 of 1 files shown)


nubs:
:+1: - this plugin is definitely broken without this fix.

mcornella:
:+1: /cc @robbyrussell

giordanovi:
There is a problem with the subtasks list. the plugin doesn't fetch the tasks with a : or . gulp.task("theme.assets", ....);
you can use a "gulp --tasks-simple" command to fetch the tasks list.

It's possible to implement this feature?

vassyz:
Does it actually need double quotes?

giordanovi:
No double quotes.
It should be something like this:
```
compls=$(gulp --tasks-simple >/dev/null | sort)
```

mcornella:
@giordanovi that command looks much better, if you know how submit a PR and we'll test it. Thanks!

giordanovi:
Hi @mcornella, 
sorry but i don't know how submit a PR. how can i do that?

giordanovi:
I was able to submit the pull request! Thanks

PR #4422, opened 2015-09-28
Honor ZSH_TMUX_ITERM2 for new-session and attach

shaxbee:
Option ZSH_TMUX_ITERM2 is not taking effect when attaching to or creating new session.
The fix is to scan arguments passed to tmux for new-session and attach and inject '-CC' option when eligible.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -48,7 +48,7 @@ if which tmux &> /dev/null
 	fi
 
 	# Set the correct local config file to use.
-    if [[ "$ZSH_TMUX_ITERM2" == "false" ]] && [[ -f $HOME/.tmux.conf || -h $HOME/.tmux.conf ]]
+	if [[ "$ZSH_TMUX_ITERM2" == "false" ]] && [[ -f $HOME/.tmux.conf || -h $HOME/.tmux.conf ]]
 	then
 		#use this when they have a ~/.tmux.conf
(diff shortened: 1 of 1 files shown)


plugins/tmux/tmux.plugin.zsh
```diff
@@ -60,8 +60,12 @@ if which tmux &> /dev/null
 	# Wrapper function for tmux.
 	function _zsh_tmux_plugin_run()
 	{
+		# We attach to existing or start new session
+		if [[ -n "${@[(r)attach]}" || -n "${@[(r)new-session]}" ]]
```
apjanke: Might also want to check for the `attach-session` and `new` command aliases.

Should this be picking up the `"ZSH_TMUX_FIXTERM` stuff too, so it matches the attach/new-session behavior of the other branches?

DelightRun:
Yes, this is a bug for all tmux command alias such as `ta`, `ts` etc, except `tmux` command generated by the wrapper function above.
So I think we'd better fix this bug asap since this PR has been here for almost 1 year.

larson-carter:
@shaxbee Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

plugins/tmux/tmux.plugin.zsh
mcornella: It's possible to make this into a one conditional with `|`:
```zsh
if [[ -n "${@[(r)attach|attach-session|new|new-session]}" ]]
```

praveendhawan:
Any updates on this?

robbyrussell:
Closing as we never got this rebased

PR #4427, opened 2015-09-29
Add alias for npm global install

xavhan:


mcornella:
It should go above line 4, the comments above don't apply to this alias. Other than that, :+1:

xavhan:
@mcornella better ?

mcornella:
Yes, very :smile: 
/cc @robbyrussell

PR #4429, opened 2015-09-29
Remove Atom plugin

RobLoach:
The Atom plugin is obsolete.

See #3193


mcornella:
Why was this merged? At least aliases could have been preserved, if the path logic was unnecessary...

RobLoach:
@mcornella Your comment at https://github.com/robbyrussell/oh-my-zsh/pull/3193#issuecomment-144097896 .

mcornella:
I was referring to the specific change of the `at` alias, not the whole plugin.

RobLoach:
> I was referring to the specific change of the at alias, not the whole plugin.

Ah, well the whole plugin is not needed as Atom by default will accomplish what it implemented. As for the alias, building off of what you recommended, one could just use `a='atom'` or `at='atom'`.

Tokynet:
Late to the party but yes, loosing the alias sucked

PR #4435, opened 2015-09-30
fix unmatched " in gulp plugin

victorhaggqvist:
fixes source error

    .oh-my-zsh/plugins/gulp/gulp.plugin.zsh:8: unmatched "


mcornella:
Solved in #4421. Please close this one and +1 the other

PR #4424, opened 2015-09-28
Update Simulator's Path to XCode 7.x

Gutem:

PR #4437, opened 2015-10-01
Update simulator shortcut for XCode 7

MathieuLoutre:
The location of the simulator has changed in XCode 7. It's now `"${devfolder}/Applications/Simulator.app"`. I've added an extra if to check for XCode 7 and change the path accordingly.


mcornella:
Duplicate of #4424

PR #4440, opened 2015-10-01
autoenv: support homebrew-installed autoenv

### LINKED ISSUE #4438: autoenv plugin looking in wrong folder
Hello,

I just installed [autoenv via homebrew](https://github.com/kennethreitz/autoenv#mac-os-x-using-homebrew):

```
$ brew install autoenv
$ echo 'source /usr/local/opt/autoenv/activate.sh' >> ~/.zshrc
```

Unfortunately, when I modify my plugins to include the autoenv plugin in my plugins list, I see the following error when I try to source .zshrc:

>     -------- AUTOENV ---------
>     Could not find ~/.autoenv/activate.sh.
>     Please check if autoenv is correctly installed.
>     In the meantime the autoenv plugin is DISABLED.
>     --------------------------

It looks like the plugin is looking for `~/.autoenv/activate.sh` instead of `/usr/local/opt/autoenv/activate.sh`.  What's the best approach for fixing this?

apjanke:
Homebrew installs autoenv into `/usr/local/opt`. This PR has the plugin look there, too.

Closes #4147 (because it depends on and includes its commits)
Fixes #4438 

Also fixes an issue with #4147 where if you don't already have `autoenv` loaded, you'll see a spurious `autoenv_init not found` error message.

### DIFF
--- a/plugins/autoenv/autoenv.plugin.zsh
+++ b/plugins/autoenv/autoenv.plugin.zsh
@@ -1,12 +1,25 @@
 # Activates autoenv or reports its failure
-if ! type autoenv_init &>/dev/null && ! source $HOME/.autoenv/activate.sh 2>/dev/null; then
-  echo '-------- AUTOENV ---------'
-  echo 'Could not find autoenv_init function or ~/.autoenv/activate.sh.'
-  echo 'Please check if autoenv is correctly installed.'
-  echo 'In the meantime the autoenv plugin is DISABLED.'
-  echo '--------------------------'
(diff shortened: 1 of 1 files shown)


apjanke:
Hey @robbyrussell: a bug made it through in the final version of this, which is going to break `autoenv` for users. Could you quick merge #4481 too, to fix it?

Sorry about that. Not sure how that got in; I probably screwed up some `git commit`/`git push --force` workflow.

PR #4442, opened 2015-10-02
modify robbyrussell.zsh-theme to distinguish root and ordinary user

YanhaoMo:
For root user zsh will have a blue color prompt, meanwhile for ordinary user, zsh display a green color prompt! I think it's very useful.


johnnagro:
:+1: the lack of distinction in the default theme causes a fair amount of confusion

PR #4443, opened 2015-10-02
Update github plugin to work with current hub versions

### LINKED ISSUE #4439: OS X El Capitan "command not found: ___main"
Hello - I just upgraded to El Capitan, and did the bidding of the brew doctor, but I am getting an error when I start my shell: 

~~~
zsh:12: command not found: ___main
(eval):1: no matches found: *:all-files
_tags:comptags:36: can only be called from completion function
_tags:comptry:55: can only be called from completion function
_tags:comptags:60: can only be called from completion function
_tags:comptags:67: can only be called from completion function
~~~

My diag dump is here: 
https://gist.github.com/RickCogley/937578f27a0ee21f7dbb

Could someone kindly take a look? 

Sincerely,
Rick

apjanke:
This is an update to a bunch of stuff in the `github` plugin, which seems to be broken when run using current versions of `git` and `hub`.

Fixes #4439 
Closes #4164 

###   Changes  ###

Removes old completion setup that breaks with current _git and _hub completions.
Removes apparently-unused `_github` completion definition.
Ruby is no longer required by hub; removes that test.
Does not define new completions for hub; they are now defined by hub itself.
Change the functions to use hub to create the repos on GitHub. The functions weren't working as is, because they didn't create the repos on the GitHub side, and you can't just push to a nonexistent repo to create it.
Add error checking to the functions.

Uses hub's alias definition instead of a custom `git()` wrapper function. This will let `hub`'s `_hub` completion definitions work out of the box (as long as you don't have `setopt complete_aliases` set).

Adds a README.

###   Notes   ###

The Homebrew-installed `hub` needs additional configuration for its `_hub` completions to be seen by the system `zsh`. Basically, you have to do this in your `~/.zshrc` before loading OMZ. Maybe this is something that OMZ should do itself as part of initialization?

```
if (( ! ${fpath[(I)/usr/local/share/zsh/site-functions]} )); then
  FPATH=/usr/local/share/zsh/site-functions:$FPATH
fi
```

Or, more robustly, this:

```
# Pick up additional site-functions that may not be on system zsh's
# $fpath by default
() {
  local site_dir site_dirs

  site_dirs=( /usr/local/share/zsh/site-functions )
  if type brew &>/dev/null; then
    site_dirs+=$(brew --prefix)/share/zsh/site-functions
  fi
  for site_dir ( $site_dirs ); do
    if [[ -d $site_dir  && ${fpath[(I)$site_dir]} == 0 ]]; then
      FPATH=$site_dir:$FPATH
    fi
  done
}
```


apjanke:
The [zsh-completions project](https://github.com/zsh-users/zsh-completions) also has a `_github` completion definition. [Their `_github`](https://github.com/zsh-users/zsh-completions/blob/3f55429b0e97ff2ba57781abd9b529ac89d186b6/src/_github) function looks newer and more comprehensive. Maybe we should remove the `_github` from this plugin and just recommend people install `zsh-completions` to get it instead?

ctrueden:
Thanks @apjanke! I confirm that this PR fixes issue #4439 on my system. With oh-my-zsh master, I see:
```shell
zsh:12: command not found: ___main
(eval):1: no matches found: *:all-files
_tags:comptags:36: can only be called from completion function
_tags:comptry:55: can only be called from completion function
_tags:comptags:60: can only be called from completion function
_tags:comptags:67: can only be called from completion function
```
Whereas with this PR merged, I see no such errors. And the github-specific features of git work as expected. :+1:

ctrueden:
@apjanke Regarding the zsh-completions project: I have that enabled as well (and it is great), but when I disabled the oh-my-zsh github plugin I did lose access to the hub commands.

Caveat emptor: I am not a dotfiles expert, so maybe I am doing something dumb. :wink:

You can view my zshrc (which uses oh-my-zsh via [zgen](https://github.com/tarjoilija/zgen)) [here](https://github.com/ctrueden/dotfiles/blob/master/zshrc).

apjanke:
@ctrueden: You still need the `github` plugin to alias `git` to `hub` and associate it with the right completion function. (Though it seems to use `_git` and not `_github`, so I'm not sure the `_github` completion definition inside the plugin is used at all.) I'm just saying maybe we should remove the completions part of it; not the entire plugin.

ctrueden:
> I'm just saying maybe we should remove the completions part of it; not the entire plugin.

OK cool, that makes sense. Thanks for the info!

apjanke:
I did some testing and it seems like that `_github` completion definition file is not used, so I've amended this PR to remove it.

mcornella:
Thanks for this very detailed PR!

apjanke:
You're welcome!

patcon:
Not sure why, but all my git completion broke a few weeks ago, and I traced it back to this merge commit when things stopped working.

This are confirmed working for me if I checkout @dfe10af

When working `which git` gives a function. After this merge, it simply gives the hub alias that I have elsewhere

anyhow, not sure I have time to dig deeper, as I'm happy to hold back updates, but just wanted to provide a data point

patcon:
Ah, and just realized that removing the `github` plugin brings back all my autocompletions, even on master -- no need to hold myself back on an old commit

apjanke:
See discussion under #4670 wrt completion.

PR #4444, opened 2015-10-02
git-extras: fix for compatibility with zsh _git

### LINKED ISSUE #4395: Git completion partly broken
Hi,

Firstly I don't know if the bug is specifically related to oh my zsh, but it started from my last update, so I'm posting here. Apologizes if it's not related.

So my issue is that since about a week, the completion for git checkout only works partially : for instance, if I use `gb <Tab>`, it shows the correct branches. But when I do `gco <Tab>`, it only shows a part of the branches names. Almost all of my branches are of the form `user/feature_or_fix/issue_number/name_of_the_branch`, and the completion only suggests the `name_of_the_branch` part of the branches. Here is an example : 
```
nyddogghr/feature/MEDIA-12/create_rails_5
```
`gb <Tab>` will show the full branch name, gut `gco <Tab>` will only show the `create_rails_5` part.
Of course, when selecting it it doesn't switch branch.

So is this an issue related to oh my zsh ? Does anyone else have this ?

Thanks for your help

apjanke:
This changes all the __git_* functions it was defining to __gitex_* to avoid collisions with the internal functions used inside _git from zsh. Those internal functions are subject to change, so redefining them can cause compatibility problems with different versions of zsh.

Adds a README with notes on compatibility.

Fixes #4395.

### DIFF
--- a/plugins/git-extras/README.md
+++ b/plugins/git-extras/README.md
@@ -1,0 +1,11 @@
+# git-extras
+
+This plugin provides completion definitions for some of the commands defined by [git-extras](http://github.com/tj/git-extras).
+
+##  Setup notes
+
+The completions work by augmenting the `_git` completion provided by `zsh`. This only works with the `zsh`-provided `_git`, not the `_git` provided by `git` itself. If you have both `zsh` and `git` installed, you need to make sure that the `zsh`-provided `_git` takes precedence.
+
+### OS X Homebrew Setup
+
+On OS X with Homebrew, you need to install `git` with `brew install git --without-completions`. Otherwise, `git`'s `_git` will take precedence, and you won't see the completions for `git-extras` commands.
--- a/plugins/git-extras/git-extras.plugin.zsh
+++ b/plugins/git-extras/git-extras.plugin.zsh
@@ -1,9 +1,12 @@
-#compdef git
 # ------------------------------------------------------------------------------
 # Description
 # -----------
 #
 #  Completion script for git-extras (http://github.com/tj/git-extras).
+#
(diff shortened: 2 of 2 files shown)


apjanke:
Two questions on this.

Should the `git-extras` plugin try to detect when it's running against the incompatible `_git` from `git` instead of `zsh`, and warn the user? When this happens, the `git-extras` completions simply won't work.

And, should we just contribute this whole plugin "upstream" to the `zsh-extras` project and remove it from Oh My Zsh? That way it could be maintained by the zsh-extras devs, would be available to people who don't use OMZ, and the version of completions you have loaded would track the version of `zsh-extras` you are running. The OMZ plugin then could just be a shell that detected and ran the completion initialization code from the `git-extras` installation. There's interest in it from `git-extras`: tj/git-extras#376.

benjaoming:
I don't have a qualified opinion about the questions raised. I found the initial commit that broken this, and it was because the git-extras plugin overwrote a shared git function. This solves the immediate problem and avoids future collisions.

Thanks @apjanke for fixing this so quickly!!

:+1:

apjanke:
Hey @spacewander and @agrimaldi, any concerns about this patch? I'm assuming that the interaction with `_git`'s own private internal functions was unintentional, and the `git-extras` plugin just wanted to re-use the code, not inject different completion behavior for the core `git` commands.

If no major concerns I'm going to get this pushed soon.

spacewander:
@apjanke 
Thank you for this fix. :+1:

mcornella:
Works as well on my end
:+1: /cc @robbyrussell

PR #4450, opened 2015-10-03
lib/git.zsh: Added git_commits_behind function

stevenspasbo:
When updating my git prompt, I noticed there was a function for git_commits_ahead, but not one for git_commits_behind. The whitespace edit was an unintentional side effect from emacs.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -86,7 +86,7 @@ function git_commits_ahead() {
 # Gets the number of commits behind remote
 function git_commits_behind() {
   if $(command git rev-parse --git-dir > /dev/null 2>&1); then
-    echo $(git rev-list --count $(current_branch)..$(git rev-parse --abbrev-ref --symbolic-full-name @{u}))
+    echo $(git rev-list --count HEAD..@{upstream})
   fi
 }
 
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -83,6 +83,13 @@ function git_commits_ahead() {
   fi
 }
 
+# Gets the number of commits behind remote
+function git_commits_behind() {
+  if git rev-parse --git-dir > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


dschaller:
:+1:

thuandt:
:+1:

sachin21:
:+1:

lib/git.zsh
```diff
@@ -70,6 +70,13 @@ function git_commits_ahead() {
   fi
 }
 
+# Gets the number of commits behind remote
+function git_commits_behind() {
+  if git rev-parse --git-dir > /dev/null 2>&1; then
```
sachin21: This function should use original git command.
plz change to this:
before:
```
  if git ...
```

after:
```
  if command git ...
```

lib/git.zsh
```diff
@@ -70,6 +70,13 @@ function git_commits_ahead() {
...
+    echo $(git rev-list --count $(current_branch)..$(git rev-parse --abbrev-ref --symbolic-full-name @{u}))
```
sachin21: too https://github.com/robbyrussell/oh-my-zsh/pull/4450/files#r42857898

warriorpostman:
Is this PR just in need of tester confirmation? I'd love to help out.

mcornella:
I'd expect this function to follow the same convention as the `*_ahead` function. That is, using appropriate variables and a similar algorithm, _i.e._ using `git log`, though the `*_ahead` function uses `grep`, `wc` and `tr` which could be dropped using the appropriate parameters to `git log`.

mcornella:
Ignore my last algorithm suggestion. It seems the fastest more concise way to get this is using `git rev-list` and the `HEAD..@{upstream}` range:
```
git rev-list HEAD..@{upstream}
```

No need for `current_branch` nor getting the revision hash of `@{u}` since `rev-list` already understands `@{u}`.

stevenspasbo:
Changes were made and are ready for testing.

mcornella:
LGTM

PR #4451, opened 2015-10-03
Add branch plugin

victor-torres:
This simply displays current Git or Mercurial branch.

The idea came when I first installed Mercurial plugin and it was a little slow on my Mac Air. 
Now it's simple and fast (it was annoying me).

I'm creating branch plugin because this one simply displays repository branch and **nothing else**.

### DIFF
--- a/plugins/branch/README.md
+++ b/plugins/branch/README.md
@@ -1,0 +1,33 @@
+# Branch
+
+Displays the current Git or Mercurial branch fast.
+
+## Speed test
+
+### Mercurial
+
+```shell
+$ time hg branch
+0.11s user 0.14s system 70% cpu 0.355 total
+```
+
+### Branch plugin
+
+```shell
+$ time sh /tmp/branch_prompt_info.sh
+0.01s user 0.01s system 81% cpu 0.018 total
+```
+
+## Usage
+
+Edit your theme file (eg.: `~/.oh-my-zsh/theme/robbyrussell.zsh-theme`) 
+adding `$(branch_prompt_info)` in your prompt like this:
+
+```diff
+- PROMPT='${ret_status}%{$fg_bold[green]%}%p %{$fg[cyan]%}%c %{$fg_bold[blue]%}$(git_prompt_info)%{$fg_bold[blue]%} % %{$reset_color%}'
++ PROMPT='${ret_status}%{$fg_bold[green]%}%p %{$fg[cyan]%}%c %{$fg_bold[blue]%}$(git_prompt_info)$(branch_prompt_info)%{$fg_bold[blue]%} % %{$reset_color%}'
+```
+
+## Maintainer
+
+Victor Torres (<vpaivatorres@gmail.com>)
--- a/plugins/branch/README.md
+++ b/plugins/branch/README.md
@@ -14,13 +14,13 @@ $ time hg branch
 ### Branch plugin
 
 ```shell
-$ time sh /tmp/branch_prompt_info.sh
-0.01s user 0.01s system 81% cpu 0.018 total
+$ time zsh /tmp/branch_prompt_info_test.zsh
+0.00s user 0.01s system 78% cpu 0.014 total
 ```
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


StefanYohansson:
:+1:

andCoffeeCode:
+1

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
+# Branch: displays the current Git or Mercurial branch fast.
+# Victor Torres <vpaivatorres@gmail.com>
+# Oct 2, 2015
+
+function branch_prompt_info() {
+  # Defines path as current directory
+  path=$(pwd)
```
mcornella: The `path` variable should be local, otherwise it leaks out and overrides the `$path` variable which is the env variable in zsh. Also, use special variables whenever you can instead of running a command: `$PWD` is a variable that already contains the cwd.

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
...
+  # While current path is not root path
+  while [ $path != '/' ];
```
mcornella: Use `[[` instead of single brackets. Also these ending semicolons aren't necessary

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
...
+  do
+    # Git repository
+    if [ -d ${path}/.git ];
```
mcornella: Wrap the whole thing inside double quotes in case there are spaces, otherwise there's a syntax error: `[[ -d "$var/.git" ]]`

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
...
+    then
+      echo '±' $(/bin/cat ${path}/.git/HEAD | /usr/bin/cut -d / -f 3-)
```
mcornella: You can use `command` before a command to force using the actual command (instead of, say, an alias or a function). So `command cat` and `command cut` will do. 

The `cut` part could also be optimized out with a bit of zsh magic: 
```zsh
echo ${"$(command cat "$var/.git/HEAD")"##*/}
```
and we can optimize out the `cat` part as well with `$(<"$var/.git/HEAD")`:
```zsh
${"$(<"$var/.git/HEAD")"##*/}
```

I know it looks really ugly, but it doesn't run any fork so it should be twice as fast.

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    fi
+    # Mercurial repository
+    if [ -d ${path}/.hg ];
+    then
+      echo '☿' $(/bin/cat ${path}/.hg/branch)
```
mcornella: The same optimization can be applied here

plugins/branch/branch.plugin.zsh
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    fi
+    # Defines path as parent directory and keeps looking for :)
+    path=$(/usr/bin/dirname $path)
```
mcornella: `dirname` can also be optimized with the [`:h` modifier](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers): `var="${var:h}"`

victor-torres:
Thank you @mcornella. I'll make these modifications and update the pull request ASAP.

mcornella:
That was fast haha. Ok, you have a little bit of homework so go ahead :wink:

victor-torres:
@mcornella anything else?

plugins/branch/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```shell
+$ time hg branch
+0.11s user 0.14s system 70% cpu 0.355 total
+```
+
+### Branch plugin
+
+```shell
+$ time sh /tmp/branch_prompt_info.sh
```
mcornella: This should be updated to use zsh

victor-torres:
My bad 😁
Just did it, @mcornella.

mcornella:
:+1:

victor-torres:
Thanks.

PR #4455, opened 2015-10-04
Fix 'ack-grep' bug for non-ubuntu users

### LINKED ISSUE #4454: /usr/bin/which: no ack-grep in (PATH)
I just updated and I'm getting this error when logging into my centOS box.

swvjeff:
Redirect STDERR along with STDOUT when looking for ack-grep

Fixes #4454
Closes #4456

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -22,7 +22,7 @@ alias _='sudo'
 alias please='sudo'
 
 ## more intelligent acking for ubuntu users
-if which ack-grep > /dev/null;
+if which ack-grep &> /dev/null;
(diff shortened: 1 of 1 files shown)


apjanke:
:+1:

If you put "Fixes #4454" and "Closes #4456" in the body of this PR description at the beginning of a line, it'll auto-close the bugs when this is merged.

@mcornella, this is a trivial change and tests fine; we should merge soon to clear up spurious errors users are seeing with this week's update.

swvjeff:
Will do, thanks for the tip!

PR #4463, opened 2015-10-05
Fetch Gulp tasks list in a better way

giordanovi:
Fetch Gulp tasks lists in a better way through 
```bash
gulp --tasks-simple
```

### DIFF
--- a/plugins/gulp/gulp.plugin.zsh
+++ b/plugins/gulp/gulp.plugin.zsh
@@ -2,12 +2,12 @@
 
 #
 # gulp-autocompletion-zsh
-# 
+#
 # Autocompletion for your gulp.js tasks
 #
 # Copyright(c) 2014 André König <andre.koenig@posteo.de>
 # MIT Licensed
-# 
+#
 
 #
 # André König
@@ -20,7 +20,7 @@
 # in the current directory.
 #
 function $$gulp_completion {
-    compls="$(grep -Eo "gulp.task\((['\"](([a-zA-Z0-9]|-)*)['\"],)" gulpfile.js 2>/dev/null | grep -Eo "['\"](([a-zA-Z0-9]|-)*)['\"]" | sed s/"['\"]"//g | sort)"
+    compls=$(gulp --tasks-simple >/dev/null | sort)
 
     completions=(${=compls})
     compadd -- $completions
--- a/plugins/gulp/gulp.plugin.zsh
+++ b/plugins/gulp/gulp.plugin.zsh
@@ -20,7 +20,7 @@
 # in the current directory.
 #
 function $$gulp_completion {
-    compls=$(gulp --tasks-simple >/dev/null | sort)
+    compls=$(gulp --tasks-simple 2>/dev/null)
 
     completions=(${=compls})
     compadd -- $completions
--- a/plugins/gulp/gulp.plugin.zsh
+++ b/plugins/gulp/gulp.plugin.zsh
@@ -20,9 +20,9 @@
 # in the current directory.
 #
 function $$gulp_completion {
-    compls=$(gulp --tasks-simple 2>/dev/null)
+    local compls=$(gulp --tasks-simple 2>/dev/null)
 
-    completions=(${=compls})
+    local completions=(${=compls})
     compadd -- $completions
 }
 


mcornella:
Good job! It works on my end, though I don't have a good `gulpfile.js` sample file to test the difference in speed. Let me prepare a test suite to try it out.

giordanovi:
Great! Good testing!

apjanke:
Does this obsolete  #4499 and #4487?

giordanovi:
I think Yes!
--simple-tasks get all task registered in current gulp enviroment

Il giorno ven 6 nov 2015 alle ore 13:26 Andrew Janke <
notifications@github.com> ha scritto:

adam-beck:
I don't htink this obsoletes #4499 because of folder structures set up with a `gulpfile.js` that references files in a `gulp/` directory. 

For instance:

```
gulpfile.js
gulp/
   scripts.js
   styles.js
   build.js
```

For me, `gulp --simple-tasks` just returns an empty string.

apjanke:
Any chance someone could make a little repo of reference test gulp setups like that, so we can play with the various PRs under different situations they're supposed to support?

giordanovi:
@adam-beck how do you include the files in gulp directory?
I've the same structures and it's working for me.

gulpfile.js
```javascript
var requiredir = require('require-dir');

var gulp = require('gulp');
var dirs = requiredir('./tasks/');
```

tasks/scripts.js
```javascript
var gulp   = require('gulp');

gulp.task('scripts', function() {
    /* some code ... */
});
```

adam-beck:
**gulpfile.js**
```
var fs =require('fs');

var options = {
  // some options
};

fs.readdirSync('./gulp/').filter(function(file) {
  return (/\.(js)$/i).test(file);
}).map(function(file) {
  require('./gulp/' + file)(options);
});
```

If you notice I have added an options to the individual required gulp files. This is for things like the build directory, temp directory, etc. This might be a use case specific to me so I wouldn't worry about that.


For clarity, here is a sample gulp file:
```
var gulp = require('gulp');
var someTool = //someTool

module.exports = function(options) {
  gulp.task('foo', function() {
   gulp.src(options.src)
   .pipe(someTool())
   .pipe(gulp.dest(options.tmp));
  });
};
```

giordanovi:
@adam-beck try to require gulp in main gulpfile.

```javascript
var fs =require('fs');
var gulp = require('gulp');

var options = {
  // some options
};

fs.readdirSync('./gulp/').filter(function(file) {
  return (/\.(js)$/i).test(file);
}).map(function(file) {
  require('./gulp/' + file)(options);
});
```

It works?

adam-beck:
Weird. Well then, I don't know what I was doing wrong. Nevermind.

giordanovi:
Your scenario works in my case, with and without require gulp in gulpfile.js

adam-beck:
That snippet I sent was not the full gulpfile. Sorry I should have made that more clear. But I was including it even though it was not necessary.

mcornella:
So this PR works with @adam-beck case? Can this be merged instead of #4499?

giordanovi:
On my end, it works perfectly.
I think it will work with babel (#4499). It's a gulp command.

plugins/gulp/gulp.plugin.zsh
```diff
@@ -20,7 +20,7 @@
 # in the current directory.
 #
 function $$gulp_completion {
-    compls="$(grep -Eo "gulp.task\((['\"](([a-zA-Z0-9]|-)*)['\"],)" gulpfile.js 2>/dev/null | grep -Eo "['\"](([a-zA-Z0-9]|-)*)['\"]" | sed s/"['\"]"//g | sort)"
+    compls=$(gulp --tasks-simple >/dev/null | sort)
```
mcornella: This should be `2>/dev/null`, and the `| sort` can be dropped because it's already done by `compadd`.

plugins/gulp/gulp.plugin.zsh
```diff
@@ -20,7 +20,7 @@
...
 
     completions=(${=compls})
```
mcornella: `$compls` and `$completions` should be `local`, otherwise they leak into the global namespace.

mcornella:
I've added a couple of things that need to be changed, once they're in place this is good to go!

giordanovi:
Done. Thank you

mcornella:
The `local` stuff isn't there yet? Did you commit it? It should be `local compls` and `local completions`. Thanks

giordanovi:
`local` stuff added

mcornella:
Thanks!

giordanovi:
Sorry @marcoccchan! 
The plugin doesn't work with the `local` stuff!
The shell report `638gulp_completion:local:1: not valid in this context:`
It' work well without it!

PR #4466, opened 2015-10-06
Added atom plugin for cygwin

rams3s:
Although the original atom plugin is now obsolete (as it adds no extra functionality over the default command line tool provided when installing Atom.io), this plugin helps using the tool on cygwin by translating Linux-like paths to Windows-like paths before calling the original script.

### DIFF
--- a/plugins/atom-cygwin/atom-cygwin.plugin.zsh
+++ b/plugins/atom-cygwin/atom-cygwin.plugin.zsh
@@ -1,0 +1,18 @@
+if [[ "$OSTYPE" == "cygwin" ]]; then
+    local _atom_path > /dev/null 2>&1
+
+    _atom_path=${LOCALAPPDATA}/atom/bin/atom
+
+    if [[ -a $_atom_path ]]; then
+        cyg_open_atom()
+        {
+            if [[ -n $1 ]]; then
+                ${_atom_path} `cygpath -w -a $1`
+            else
+                ${_atom_path}
+            fi
+        }
+
+        alias atom=cyg_open_atom
+    fi
+fi
--- a/plugins/atom-cygwin/README.md
+++ b/plugins/atom-cygwin/README.md
@@ -1,0 +1,17 @@
+## atom-cygwin
+
+This plugin allows invoking the Atom Editor from Cygwin using the `atom` command.
+For this purpose, it uses `cygpath` to convert between Unix and Windows paths.
+
+### Requirements
+
+ * [Atom](https://atom.io/)
+ * [Cygwin](https://www.cygwin.com/)
+
+### Usage
+
+Same as the original CLI script (pointed by `${LOCALAPPDATA}/atom/bin/atom`).
+
+ * If `atom` command is called without an argument, launch Atom
+ * If `atom` is passed a directory, open it in Atom
+ * If `atom` is passed a file, open it in Atom


plugins/atom-cygwin/atom-cygwin.plugin.zsh
```diff
@@ -0,0 +1,18 @@
+if [[ "$OSTYPE" == "cygwin" ]]; then
+    local _atom_path > /dev/null 2>&1
```
apjanke: This shouldn't be `local`. Under OMZ, it has no effect, because plugins are not loaded inside function calls. In contexts where it will have an effect (like Prezto), it will break `cyg_open_atom()`, which needs the `$_atom_path` value to persist.

If you want this to be local to avoid polluting the main namespace, stick it inside an anonymous function, and then have `cyg_open_atom()` use its own local `$_atom_path` variable and reconstruct the path each time it's called. `$LOCALAPPDATA` is unlikely to change during a session, so you don't need to capture the value once and hold it.

apjanke:
Does this actually work for you? When I try to run `atom` under Cygwin, I just get syntax errors, since Atom installs its shell scripts with DOS mode CRLF endings. (Current version of Atom, downloaded and installed today, on Windows 7.)

```
$ atom foo.txt
/cygdrive/c/Users/janke/AppData/Local/atom/bin/../app-1.0.19/resources/cli/atom.sh: line 2: $'\r': command not found
/cygdrive/c/Users/janke/AppData/Local/atom/bin/../app-1.0.19/resources/cli/atom.sh: line 4: syntax error near unexpected token `$'in\r''
/cygdrive/c/Users/janke/AppData/Local/atom/bin/../app-1.0.19/resources/cli/atom.'h: line 4: `  case "$opt" in
```

To get it to run, I need to call `atom.cmd` instead.

`atom` takes multiple file arguments and has options, too. Maybe this should go through all the arguments and fix paths, and ignoring options (starting with `-`).

plugins/atom-cygwin/atom-cygwin.plugin.zsh
rams3s: Thanks @apjanke for the info.

rams3s:
Indeed, I ran into the same issue but didn't find a good solution to address this line ending issue.
My 'fix' was to run:
```sed -i 's/\r//' ${LOCALAPPDATA}/atom/bin/atom```

Of course, this would need to be redone after next update.

I'll give a try at atom.cmd and will push an update when it's fixed.

Thank you for your help.

apjanke:
Cool.

Did it actually work for you even after that? I did that line-ending fix (but with `dos2unix` instead of `sed`) to `${LOCALAPPDATA}/atom/bin/atom` and the other `.sh` scripts it called, but it would still error out, complaining about bad paths. It looks like those `bash` files might be specific to MinGW, and not supported under Cygwin. I opened an upstream issue asking about it: atom/atom#9169.

Using `atom.cmd` might be the most reliable way for now.

plugins/atom-cygwin/atom-cygwin.plugin.zsh
apjanke: You're welcome.

I forgot to mention: as a general rule of thumb, whenever you end up redirecting the output of `local`, `typeset`, or something similar to `>/dev/null`, that's a sign it's being used incorrectly.

rvadai:
Guys, I added an Atom plugin but it only works on Mac OS X: https://github.com/robbyrussell/oh-my-zsh/pull/5308

Shall we merge our PRs to support both a Mac and cygwin?

apjanke:
Merging makes sense. OMZ typically has one plugin per tool and tries to make them portable.

Even though Atom installs an `atom` command in `/usr/local/bin` on OS X, it seems like there are some scenarios where it may not work. (Non-writable `/usr/local`, or multiple users with per-user Atom installs.)

rvadai:
Cool. @apjanke can you send me a PR for my repo: https://github.com/rvadai/oh-my-zsh
Then that will be included in my PR for this repo https://github.com/robbyrussell/oh-my-zsh/pull/5308

rvadai:
Guys, I closed my PR because I've found a way to install the Atom command line tools on a Mac (it's possible from the main menu).

robbyrussell:
Closing this as I believe the PR was closed, too

PR #4469, opened 2015-10-06
colored-man plugin: Quoted PATH variable

cknoblauch:
Some environments (in my case, cygwin) have PATH variables with directories with spaces in them, breaking colored-man. Adding quotes around the PATH environment variable makes the plugin run OK.


robbyrussell:
Hi, can you rebase? I've renamed this plugin

apjanke:
This quoting shouldn't be necessary under default `zsh` options. A more robust way to protect this might be to do `emulate -L zsh` at the beginning of the `man()` function, instead of worrying about quoting individual lines.

mcornella:
Not only that but quoting `PATH=...` won't work either.

PR #4471, opened 2015-10-08
Also check custom themes in `$ZSH_CUSTOM/themes`

mul14:
According to [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes) documentation,
the location of custom themes should be on `$ZSH_CUSTOM/themes`

```
zsh_custom
└── themes
    └── my_awesome_theme.zsh-theme
```


apjanke:
:+1:

This same change is also effectively included in #3743, which folds the `theme` plugin in to the main theme code, so they're both using the same search logic.

jamen:
This one has no conflicts currently.

PR #4476, opened 2015-10-09
sudo added to nmap commands which require sudo

meehow:


thuandt:
:+1:

sachin21:
Why do you add `sudo`?
I wanna konw

PR #4477, opened 2015-10-09
Update git.plugin.zsh

sadfuzzy:
Lightning fast 'git status' command


apjanke:
This masks the `gs` command from Ghostscript.

mcornella:
Besides that, you can define your own aliases in the zshrc file, at the bottom of it:
```zsh
# zshrc file contents
# more stuff
# ...
alias gs='git status'
```

Closing this one since it's not a good practice to have conflicting aliases. Thanks for your contribution though!

PR #4478, opened 2015-10-09
Allow more flexability in update prompt

dpelton:
Allows users to enter responses such as 'yes', 'yeah', 'yup', etc. for greater flexibility. This also helps the oh-my-zsh updater to conform with common bash conventions.


Swingline0:
This works as expected here. Essentially any word that begins with `Y` or `y` will evaluate as a confirmation for the upgrade. 

:+1:

mcornella:
Fixed in #4207

PR #4482, opened 2015-10-10
Add an alias for upstream

sachin21:
# WHAT
Added an alias for upstream.

# WHY
git plugin does not have that so far, so I'd like to use it.


sachin21:
@robbyrussell How about changes?

sachin21:
@robbyrussell Thanks for quickry merging :smile:

PR #4484, opened 2015-10-11
Also check for non-system-wide ZSH installation.

tonyxue:
Some users don't have root access on machines and the machines do not have zsh
installed. They can still have zsh installed in their own $HOME. The install
script should also check for that otherwise will report ZSH Not Installed.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -21,7 +21,7 @@ else
   BOLD=""
   NORMAL=""
 fi
-CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
+CHECK_ZSH_INSTALLED=$((grep /zsh$ /etc/shells || whereis zsh) | wc -l)
 if [ ! $CHECK_ZSH_INSTALLED -ge 1 ]; then
   printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
   exit


apjanke:
If you want to support non-system Zsh installs, you'll need to add an additional code path around `chsh`, since `chsh` will not work with a non-system `zsh`.

tools/install.sh
```diff
@@ -21,7 +21,7 @@ else
   BOLD=""
   NORMAL=""
 fi
-CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
+CHECK_ZSH_INSTALLED=$((grep /zsh$ /etc/shells || whereis zsh) | wc -l)
```
apjanke: Might want to use some `which` logic instead of `whereis` here. The Linux version of `whereis` still produces output when it doesn't find anything.

```
$ uname -a
Linux debbie 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt11-1 (2015-05-24) x86_64 GNU/Linux
$ whereis no-such-command
no-such-command:
$ whereis no-such-command | wc -l
1 
```

tonyxue:
You're right. I forgot about ```chsh```
But I realize that both ```which``` and  ```whereis``` produce output when the command doesn't exist. I'll try other method to fix that.

tonyxue:
@apjanke How about now?

tools/install.sh
```diff
@@ -86,10 +86,15 @@ mv -f ~/.zshrc-omztemp ~/.zshrc
...
-  if hash chsh >/dev/null 2>&1; then
+  # If this platform provides a "chsh" command (not Cygwin) and ZSH is installed
+  # globally, do it, man!
+  if hash chsh >/dev/null 2>&1 && [ "$CHECK_ZSH_INSTALLED" -ge 1 ]; then
     printf "${BLUE}Time to change your default shell to zsh!${NORMAL}\n"
     chsh -s $(grep /zsh$ /etc/shells | tail -1)
+  # If ZSH only installed non-globally.
+  elif [ "$CHECK_ZSH_INSTALLED_NON_GLOBALLY" -eq 0 ] && [ ! "$CHECK_ZSH_INSTALLED" -ge 1 ]; then
+    printf "I can't change your shell automatically because ZSH is not installed globally.\n"
+    printf "${BLUE}You need to start ZSH in your prefered way every time!${NORMAL}\n"
```
apjanke: "preferred" has two r's.

tools/install.sh
```diff
@@ -86,10 +86,15 @@ mv -f ~/.zshrc-omztemp ~/.zshrc
...
-  # If this platform provides a "chsh" command (not Cygwin), do it, man!
-  if hash chsh >/dev/null 2>&1; then
+  # If this platform provides a "chsh" command (not Cygwin) and ZSH is installed
+  # globally, do it, man!
+  if hash chsh >/dev/null 2>&1 && [ "$CHECK_ZSH_INSTALLED" -ge 1 ]; then
     printf "${BLUE}Time to change your default shell to zsh!${NORMAL}\n"
     chsh -s $(grep /zsh$ /etc/shells | tail -1)
+  # If ZSH only installed non-globally.
+  elif [ "$CHECK_ZSH_INSTALLED_NON_GLOBALLY" -eq 0 ] && [ ! "$CHECK_ZSH_INSTALLED" -ge 1 ]; then
+    printf "I can't change your shell automatically because ZSH is not installed globally.\n"
```
apjanke: this should maybe say "...or your system does not support `chsh`." (like cygwin, which doesn't have chsh, because... well, reasons.) May be easier to just leave out the "...because..." part? On Cygwin and similar systems, this would give a misleading diagnostic message.

apjanke:
Yeah, that looks about right.

Haven't tested this yet; would want to do so, because there's edge cases and stuff on different OSes.

PR #4486, opened 2015-10-12
Add option '--port' to 'rails server' alias.

gfvcastro:


KevinBongart:
Hey @gfvcastro & @robbyrussell, thanks for keeping this Rails plugin alive!

Just wanted to provide some quick feedback: I used to run my specs by typing `rsp` then pressing tab and it would be autocompleted into `rspec`. With this alias, I can't anymore :(

Sure, it's an extremely minor annoyance, perfectly comparable to this:

[](https://xkcd.com/1172/)

But I'm wondering whether such a specific command needs to be added as one of the default aliases for all Rails developers who use OMZ (I have a lot of aliases that I keep in my own `.zshrc`).

If other people start having the same issue, maybe we could reconsider?

gfvcastro:
@KevinBongart: Sorry. Oh... You are right...

KevinBongart:
@gfvcastro It's alright! Since then, I just adapted my workflow by typing one more letter :)

PR #4487, opened 2015-10-13
gulp plugin support for babel (and other compilers?)

aforty:
This simple change adds wildcard support to `gulpfile*.js`. Mine for example is named `gulpfile.babel.js` which enables the babel compiler to be used in the gulpfile.


apjanke:
Looks like this is obsoleted by the conflicting #4499, which does a broader search for gulpfiles. Does that one cover all your use cases for this one?

aforty:
Yeah looks like it does.

PR #4489, opened 2015-10-13
Include dotfiles in TAB completion

nmaggioni:
Includes hidden files and folders in ```TAB``` key completion.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -76,6 +76,8 @@ if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
 else
   compinit -i -d "${ZSH_COMPDUMP}"
 fi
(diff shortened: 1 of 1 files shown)


apjanke:
This seems more like a personal preference that should be left up to to the individual user to turn on if they want it.

nmaggioni:
In my opinion the vast majority of users would want this, and the documentation to enable this feature is quite hard to find on the web...

dragonxlwang:
+1, hope user can have a toggle function to enable showing hidden files

mcornella:
I agree with @apjanke on this one. This is more like a user preference, and enabling it is as easy as adding that same line to your zshrc file:
```zsh
# Enable hidden files completion
_comp_options+=(globdots)
```

PR #4490, opened 2015-10-14
Added drush compatibility

lucsrods:
Aliases and function for drush to help the drupal workflow.

### DIFF
--- a/plugins/drush/README.md
+++ b/plugins/drush/README.md
@@ -12,6 +12,7 @@ plugins=(... drush)
 ## Aliases
 | Alias | Description                                                           | Command                     |
 |-------|-----------------------------------------------------------------------|-----------------------------|
+| dr    | Display drush help                                                    | drush                       |
 | drca  | Clear all drupal caches.                                              | drush cc all                |
 | drcb  | Clear block cache.                                                    | drush cc block              |
 | drcg  | Clear registry cache.                                                 | drush cc registry           |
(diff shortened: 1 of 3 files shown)


plugins/drush/README.md
```diff
@@ -0,0 +1,31 @@
+# Drush plugin #
```
RobLoach: Don't need the suffixing #

RobLoach:
Could likely bring in the autocomplete stuff in https://github.com/drush-ops/drush/blob/8.0.1/drush.complete.sh .

lucasar:
@RobLoach hey man, thanks for the updates, and sorry for the delay. Regarding your suggestion i'll take a look and see what i can integrate in both scripts.

plugins/drush/drush.plugin.zsh
```diff
@@ -0,0 +1,47 @@
+# Drush support.
+
+function dren() {
+	drush en $@ -y
```
mcornella: This doesn't have the same indentation

plugins/drush/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Description ##
+This plugin offers aliases and functions to make the work with drush easier and more productive.
+
+## How to use ##
```
mcornella: This whole section should be restructured to use markdown tables. [See the `xcode` plugin for an example](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/xcode/README.md#aliases).

revagomes:
:+1:

lucasar:
Hey there @revagomes, thanks for the interest and suggestion, I'll take a look on their script and see what we can be using.

PR #4491, opened 2015-10-15
Add support in profiles plugin for directories

linkinpark342:
The profiles plugin currently has support to load a single "profile"
file for each component in the FQDN. This changeset makes it so the
plugin supports the filename actually being a directory with multiple
*.zsh files in it. Each of those files are then loaded in alphabetical
order


WPettersson:
Is there any reason this was closed? I find myself wanting this same feature, and I've implemented it myself as well (albeit in a slightly different manner).

PR #4492, opened 2015-10-15
Tweak pyenv plugin

arvindch:
I use pyenv a lot, but the inbuilt plugin is missing a few things, so I thought I'd update it.

1. 8a95ced - Initialize pyenv virtualenvs, to get autocompletion for them.
2. 8a362b2 - Re-enable pyenv rehashing. It's useful, so not sure why it's disabled.

Let me know if this is OK!


arvindch:
Great! I think this PR be closed now.

PR #4496, opened 2015-10-15
Fix install.sh for tput-less systems

WhyEee:
@fcrozat's original fix assumes `which` not to output anything to STDOUT
in case the command is not found. That is not necessarily true on all
systems. A better solution is to check the return value instead.

See also: #4376

### DIFF
--- a/plugins/gradle/gradle.plugin.zsh
+++ b/plugins/gradle/gradle.plugin.zsh
@@ -1,6 +1,6 @@
 #!zsh
 ##############################################################################
-# A descriptive listing of core Gradle commands 
+# A descriptive listing of core Gradle commands
 ############################################################################
 function _gradle_core_commands() {
     local ret=1 state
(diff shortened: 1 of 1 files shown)


apjanke:
That `if $(which tput); then` won't work right: that syntax captures the output and then tries to execute it. So doing that will run `/usr/bin/tput` with no arguments, producing an error message and giving you a false negative, so colors are not used.

```
$ sh -c "$(curl -fsSL https://raw.githubusercontent.com/WhyEee/oh-my-zsh/master/tools/install.sh)"
usage: tput [-V] [-S] [-T term] capname
Cloning Oh My Zsh...
Cloning into '/Users/janke/.oh-my-zsh'...
```

You want something like `if which tput >/dev/null 2>&1; then`.

WhyEee:
Thank you for pointing out that mistake.
I updated the pull request accordingly.

apjanke:
Looks good now, and tests out on my 10.9.5 machine.

BTW, you should make PRs from dedicated branches, not your `master` branch. Otherwise things will get messed up (in both the PR and clones of your fork) as the main `master` branch picks up other commits. Particularly important for Oh My Zsh, since PRs can sit a while due to the high incoming volume. Could you re-do this PR using a dedicated branch?

WhyEee:
This PR is now superseded by #4507.

PR #4499, opened 2015-10-15
Search for Gulp tasks in more locations

adam-beck:
Searches for tasks in the following locations: gulpfile.js, gulpfile.babel.js, and gulp directory of a project.

### DIFF
--- a/plugins/gulp/gulp.plugin.zsh
+++ b/plugins/gulp/gulp.plugin.zsh
@@ -20,7 +20,7 @@
 # in the current directory.
 #
 function $$gulp_completion {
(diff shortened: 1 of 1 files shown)


apjanke:
I think that `gulp/*.js` can end up producing a `zsh: no matches found: gulp/*.js` error message, even though stderr from `grep` is sent to /dev/null. Might want to use `gulp/*.js(N)` to allow it to silently produce an empty result set when there are no matches.

apjanke:
The `$compls` and `$completions` variables look like they should be `local`, too. Could that be fixed while you're in here?

apjanke:
This conflicts with #4487 and #4463. Can these be reconciled so we only have one outstanding PR for it?

adam-beck:
#4463 would make #4487 and it (#4463) does check for tasks inside of `gulpfile.babel.js` files. Therefore, this commit is obsolete as well. I would make your comments on #4463.

PR #4500, opened 2015-10-15
[FIX ERROR] Refactor for function in plugins/rake-fast/rake-fast.plugin.zsh

sachin21:
# WHAT
I refactored rake-fast plugin.

# WHY
If this plugin condition is `[[ "$OSTYPE" = darwin* ]];`,  the case of my stat An error will occur.
My stat doesn't have -c option because I'm using GNU core utilities that was installed in the Homebrew.

# ERROR LOGS
My environments:
OS: OS X Yosemite
stat: 8.24 (GNU core utilities)
```
$ rake [TAB] 
Try 'stat --help' for more information.
stat: invalid option -- '%'
Try 'stat --help' for more information.

_rake_does_task_list_need_generating:10: bad math expression: operand expected at `>='
stat: invalid option -- '%'
Try 'stat --help' for more information.
stat: invalid option -- '%'
Try 'stat --help' for more information.
_rake_does_task_list_need_generating:10: bad math expression: operand expected at `>='
stat: invalid option -- '%'
Try 'stat --help' for more information.
stat: invalid option -- '%'
Try 'stat --help' for more information.
_rake_does_task_list_need_generating:10: bad math expression: operand expected at `>='
```

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -8,17 +8,7 @@ _rake_refresh () {
 }
 
 _rake_does_task_list_need_generating () {
-  if [ ! -f .rake_tasks ]; then return 0;
-  else
-    if [[ "$OSTYPE" = darwin* ]]; then
-      accurate=$(stat -f%m .rake_tasks)
-      changed=$(stat -f%m Rakefile)
-    else
-      accurate=$(stat -c%Y .rake_tasks)
-      changed=$(stat -c%Y Rakefile)
-    fi
-    return $(expr $accurate '>=' $changed)
-  fi
+  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]]
 }
 
 _rake_generate () {


sachin21:
@robbyrussell How about this changes?

sachin21:
Oops. My title is very long :cry: 
Sorry.

sachin21:
Could anyone test this?
Cc: @mcornella @apjanke

mcornella:
This whole logic can be replaced with zsh specific operators. There are `-nt` (newer than) and `-ot` (older than) operators that can already do that without using `stat`. See https://github.com/robbyrussell/oh-my-zsh/pull/1655#issuecomment-67301041

sachin21:
@mcornella Awesome!! Thank you so much for telling me!
I fixed that and squashed commits :dancers:

mcornella:
The `return $?` line is not needed given that is, by default, what will happen once the function ends: the return code will not change. We can also join the *file-exists* check and the *file-is-newer-than* check:

```patch
_rake_does_task_list_need_generating () {
-   if [ ! -f .rake_tasks ]; then return 0;
-   else
-    if [[ "$OSTYPE" = darwin* ]]; then
-      accurate=$(stat -f%m .rake_tasks)
-      changed=$(stat -f%m Rakefile)
-    else
-      accurate=$(stat -c%Y .rake_tasks)
-      changed=$(stat -c%Y Rakefile)
-    fi
-    return $(expr $accurate '>=' $changed)
-   fi
+  [[ ! .rake_tasks -ot Rakefile ]]
}
```

Reasoning behind this from the docs above:
> `file1 -ot file2`
>   true if file1 exists *and* is older than file2.

That is, it will return false if either:

- `.rake_tasks` doesn't exist
- `.rake_tasks` is older than `Rakefile`,

_i.e._ it will return false when we need to generate the `.rake_tasks`. We need to return true though, so we put the bang `!` operator to negate it and that's all.

Note this means that if `.rake_tasks` and `Rakefile` are equally old we will regenerate the `.rake_tasks` file.

mcornella:
Oh wait, there's a pitfall to that succint version which is if there's no `Rakefile` the function will also return true. The less abbreviated version is more robust then:
```patch
_rake_does_task_list_need_generating () {
-   if [ ! -f .rake_tasks ]; then return 0;
-   else
-    if [[ "$OSTYPE" = darwin* ]]; then
-      accurate=$(stat -f%m .rake_tasks)
-      changed=$(stat -f%m Rakefile)
-    else
-      accurate=$(stat -c%Y .rake_tasks)
-      changed=$(stat -c%Y Rakefile)
-    fi
-    return $(expr $accurate '>=' $changed)
-   fi
+  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]]
}
```

sachin21:
> The return $? line is not needed given that is

yeah, I know. but I though `return $?` is easy to understand :dash: 
however you're right so i will fix that

sachin21:
I fixed that. Thank you for correcting my codes :smile:

mcornella:
Thanks for putting up with my OCD! :grin: Good to go :+1:

sachin21:
I'm happy :smiley:

KevinBongart:
sweet! thanks for fixing/refactoring @sachin21 :+1:

sachin21:
@KevinBongart Sure! :smile:

PR #4501, opened 2015-10-16
Replace rvm specific code to generic ruby_info for alanpeabody.zsh-theme

leomao10:
I replace these logic rvm specific code to generic `#ruby_info` provide in oh-my-zsh.

So that it show ruby-version as well once you add `chruby` plugin in .zshrc


mcornella:
Fixed in master.

PR #4505, opened 2015-10-16
Check for git before trying to upgrade OMZ

### LINKED ISSUE #4504: Upgrade should be avoided when git isn't available
I copy my .oh-my-zsh directory (and .zshrc) to multiple production servers without development tools, therefore no git installation.  The request to update is foolish, as it wouldn't work.

I propose a change to tools/check_for_upgrade.sh to look for git, and give up if it's not found.

arvindch:
Hi,

Noticed #4504 and thought I'd throw together a quick PR for it.

Fixes #4504
Closes #4506

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -26,6 +26,9 @@ fi
 # oh-my-zsh directory.
 [[ -w "$ZSH" ]] || return 0
 
+# Cancel upgrade if git is unavailable on the system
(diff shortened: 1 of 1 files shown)


sachin21:
I've tested it. Problem was not anything.
Test I did was remove the git command before running the check_for_upgrade.sh.
It works :+1:

mcornella:
I'm merging this but eventually a refactoring should be done.

PR #4507, opened 2015-10-16
Fix install.sh/upgrade.sh for tput-less systems

### LINKED ISSUE #4376: install.sh fails if tput isn't available

WhyEee:
@fcrozat's original fix assumes `which` not to output anything to STDOUT
in case the command is not found. That is not necessarily true on all
systems. A better solution is to check the return value instead.

Fixes #4376.
Supersedes #4496.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -2,9 +2,8 @@ set -e
 
 # Use colors, but only if connected to a terminal, and that terminal
 # supports them.
-tput=$(which tput)
-if [ -n "$tput" ]; then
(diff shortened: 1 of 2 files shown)


mcornella:
Good job @WhyEee, merging as per https://github.com/robbyrussell/oh-my-zsh/pull/4496#issuecomment-148749806. Works well :+1:

PR #4508, opened 2015-10-16
Fixd bug for pyenv plugin

Superbil:
Check pyenv-virtualenv can eval before use it

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -18,7 +18,10 @@ for pyenvdir in "${pyenvdirs[@]}" ; do
         export PYENV_ROOT=$pyenvdir
         export PATH=${pyenvdir}/bin:$PATH
         eval "$(pyenv init - zsh)"
-        eval "$(pyenv virtualenv-init - zsh)"
+
+        if pyenv commands | command grep -q virtualenv-init; then
+            eval "$(pyenv virtualenv-init - zsh)"
+        fi
 
         function pyenv_prompt_info() {
             echo "$(pyenv version-name)"


arvindch:
Oops, I totally forgot that the pyenv virtualenv plugin is not installed by default.

However, there is no pyenv-virtualenv-init command, so this won't work...

Instead, this will:

```zsh
if [[ "$(pyenv commands | grep virtualenv-init -c)" == "1" ]]
then
    eval "$(pyenv virtualenv-init - zsh)"
fi
```

Superbil:
@arvindch I found that command from `brew info pyenv-virtualenv`, so I use that package and test on it.

Look like check `pyenv commands` is a better way, I will fix my commit.

arvindch:
Oh, interesting. It must be an OSX thing, since it's not on my Arch system. Anyway, yeah, this works on both, so cool.

plugins/pyenv/pyenv.plugin.zsh
```diff
@@ -18,7 +18,10 @@ for pyenvdir in "${pyenvdirs[@]}" ; do
         export PYENV_ROOT=$pyenvdir
         export PATH=${pyenvdir}/bin:$PATH
         eval "$(pyenv init - zsh)"
-        eval "$(pyenv virtualenv-init - zsh)"
+
+        if [[ "$(pyenv commands | grep virtualenv-init -c)" == "1" ]] ; then
```
sachin21: I think this is better:
```sh
if [[ "$(pyenv commands | grep virtualenv-init -c  > /dev/null 2>&1)" == `false` ]]; then
*---* then *---*
```
If grep is succeeded, It'll see result, so I think that code is not good.

plugins/pyenv/pyenv.plugin.zsh
```diff
@@ -18,7 +18,10 @@ for pyenvdir in "${pyenvdirs[@]}" ; do
         export PYENV_ROOT=$pyenvdir
         export PATH=${pyenvdir}/bin:$PATH
         eval "$(pyenv init - zsh)"
-        eval "$(pyenv virtualenv-init - zsh)"
+
+        if [[ "$(pyenv commands | grep virtualenv-init -c  > /dev/null 2>&1)" == `false` ]]; then
```
mcornella: A better alternative and common practice is using `grep`'s `-q` (quiet) argument. The final version should be:
```zsh
if pyenv commands | grep -q virtualenv-init; then
```

Superbil: Cool ! This is more clear.

sachin21: I have not have that option.
I think the all version to be considered :)

sachin21: I propose this:
```
if pyenv commands | grep virtualenv-init > /dev/null 2>&1; then
```

mcornella:
FWIW, `grep -q` is [part of the POSIX spec](http://www.gnu.org/software/grep/manual/grep.html#index-g_t_002do-42), so it should be everywhere.

sachin21:
@mcornella I'm sorry :cry: 
I'm using grep alias is alias='ag'
I replaced to the_sliver_searcher lol lol lol

mcornella:
@sachin21 for that we should make sure we're using the `grep` command with the `command` builtin. So @Superbil please update to:

```zsh
if pyenv commands | command grep -q virtualenv-init; then
```

after that, I think everybody should be accounted for.

sachin21:
@mcornella That is certainly true.
That's good :D

Superbil:
@mcornella Ok, I update it.

sachin21:
:+1:

mcornella:
Good job! Thank you for your time :smile:

PR #4510, opened 2015-10-17
Allow overriding agnoster colours

lheckemann:


apjanke:
The `agnoster` theme has its own upstream repo now, [agnoster/agnoster-zsh-theme](https://github.com/agnoster/agnoster-zsh-theme). Want to submit this there instead, to prevent OMZ's agnoster from diverging further than it already has? We're looking at reunifying with upstream in the near future.

PR #4515, opened 2015-10-18
Make install script safer

brianloveswords:
This changeset wraps all of the commands in tools/install.sh in a
function and then calls that function as the last line of the
script.

The current install instructions ask the user to download the install
script using `curl` and pass the result to `sh`. This is totally
fine (as long as both the instructions and the script itself are served
using HTTPS), but the script should be written in a way such that it
doesn't start trying to actually *do* anything until the very last line.

The reason is due to the way `curl` works: if the socket drops before the
request is complete (server abruptly hangs up, client's internet flakes
out, etc.), `curl` will return the partial data that it received. Here
is an example of that:

partial file execution

A way this might cause issues for tools/install.sh is if the connection drops
after cloning but before the repository (L53-56). The .zshrc
configuration will not be copied and the shell will not be changed, but
if the user tries to run the install script again it will claim
oh-my-zsh is already installed (L31-39).

While this is not a particularly dangerous error condition (the user can
just delete .oh-my-zsh and re-run), it can certainly be confusing for
new users. This also helps future-proof the script for a time when it
might need to use a "dangerous" command, e.g. `rm`, and we want to make
sure it happens in the most transactional way possible.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1,3 +1,4 @@
+function main() {
 # Use colors, but only if connected to a terminal, and that terminal
 # supports them.
 tput=$(which tput)
@@ -117,3 +118,6 @@ echo 'p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.'
 echo ''
 printf "${NORMAL}"
 env zsh
+}
+
+main


brianloveswords:
The best way to view this diff is with "ignore whitespace" set to true: https://github.com/robbyrussell/oh-my-zsh/pull/4515/files?w=1

apjanke:
Oh, this sounds like a really good idea.

What do you think about just not indenting the top-level function contents? Since it's a single function that takes up the entire file, it doesn't seem necessary in terms of readability, and that way `git blame` won't assign every single line in the file to you and this commit.

brianloveswords:
Yep, I can do that!

brianloveswords:
Okay, this PR/commit has been `--amend`ed to not change indentation level.

tools/install.sh
```diff
@@ -1,3 +1,4 @@
+function main {
```
sachin21: For coding style, That has to add parenthesis to main.
before:
```
function main {
```

after:
```
function main() {
```
Check it out: https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide#functions

sachin21: And function's under 2 spaces needed.
ref: https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide#good-12

apjanke: Is your second comment about indenting the function body? I'm suggesting we not do that in this case, because it's just wrapping the whole file as a delayed-evaluation mechanism, and doing so would touch every line and have "git blame" point to this commit, which is not helpful.

brianloveswords: Added parens after function name. Waiting on 2 spaces per @apjanke's comment

sachin21: @apjanke I see, I’ll keep that in mind :smile: 
Thank you.

brianloveswords:
Rebased to fix conflict from 8c73eb4f446650adf8f82f72e65943a8834847f3 (which wasn't a *real* conflict per se). Let me know if there's anything else I should do to get this landed!

apjanke:
:+1: Looks good to me.

I tested on OS X 10.9.5 and Debian Linux 8, and looks good. 

Normal case:
`sh -c "$(curl -fsSL https://raw.githubusercontent.com/brianloveswords/oh-my-zsh/safer-install-script/tools/install.sh)"`
And the installation worked.

Testing a "partial download" case:
```
$ sh -c "$(curl -fsSL https://raw.githubusercontent.com/brianloveswords/oh-my-zsh/safer-install-script/tools/install.sh | head -119)"
sh: -c: line 119: syntax error: unexpected end of file
```
Which gives you an error, and no steps from the installer script had been performed.

@mcornella, can we get this queued up for a merge?

mcornella:
This is good thinking. Thank you for your contribution!

mcornella:
I redid this PR after reverting it because of a syntax error in dash (#4531). See [ca5632b](https://github.com/robbyrussell/oh-my-zsh/commit/ca5632bce10e1f5b0fe2a59151ae0fd776619d89?w=1). 

I added the indentation to match the style throughout the project. `git blame` can be used with `-w` to ignore whitespace changes and see the real author of the line.

brianloveswords:
Ahh good to know about http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_09_05, sorry about that!

mcornella:
I didn't know either so... all's well :smile:

apjanke:
Darn... sorry for not catching that. Is Ubuntu different enough from Debian that it needs to be tested separately? I thought it was Debian-based. Or is this a per-machine configuration thing?

mcornella:
It also happened on Debian, I didn't test it before merging. My mistake :grin:

apjanke:
But... *I* tested it on Debian, too, and it worked for me. What gives?

PR #4516, opened 2015-10-18
Fix `vagrant`'s completion

ELLIOTTCABLE:
When completing `vagrant` commands, it says “Completing _gem_ subcommand.” Oops. :P


apjanke:
:+1:

PR #4517, opened 2015-10-19
Add option to disable calling compinit. 

mfarrugi:
Add variable `ZSH_DISABLE_COMPINIT`. When set to true, stops omzsh from invoking `compinit`.

Particularly useful when Omzsh is being used as a plugin by zgen, antigen, etc., and the user wants to call `compinit` later.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -64,19 +64,21 @@ if [ -z "$ZSH_COMPDUMP" ]; then
   ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
-if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
-  # If completion insecurities exist, warn the user without enabling completions.
-  if ! compaudit &>/dev/null; then
-    # This function resides in the "lib/compfix.zsh" script sourced above.
-    handle_completion_insecurities
-  # Else, enable and cache completions to the desired file.
-  else
-    compinit -d "${ZSH_COMPDUMP}"
-  fi
-else
-  compinit -i -d "${ZSH_COMPDUMP}"
-fi
 
+if [[ $ZSH_DISABLE_COMPINIT != true ]]; then
+  if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
+      # If completion insecurities exist, warn the user without enabling completions.
+      if ! compaudit &>/dev/null; then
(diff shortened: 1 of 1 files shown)


oh-my-zsh.sh
```diff
@@ -64,19 +64,21 @@ if [ -z "$ZSH_COMPDUMP" ]; then
...
+      if ! compaudit &>/dev/null; then
+        # This function resides in the "lib/compfix.zsh" script sourced above.
+        handle_completion_insecurities
+      # Else, enable and cache completions to the desired file.
+      else
+        compinit -d "${ZSH_COMPDUMP}"
+      fi
   else
-    compinit -d "${ZSH_COMPDUMP}"
+      compinit -i -d "${ZSH_COMPDUMP}"
```
sachin21: I think 2 indents is better.

oh-my-zsh.sh
```diff
@@ -64,19 +64,21 @@ if [ -z "$ZSH_COMPDUMP" ]; then
...
-if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
-  # If completion insecurities exist, warn the user without enabling completions.
-  if ! compaudit &>/dev/null; then
-    # This function resides in the "lib/compfix.zsh" script sourced above.
-    handle_completion_insecurities
-  # Else, enable and cache completions to the desired file.
+
+if [[ $ZSH_DISABLE_COMPINIT != true ]]; then
+  if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
+      # If completion insecurities exist, warn the user without enabling completions.
```
sachin21: I think 2 indents is so better, too

apjanke: Yeah, 2-space indents are [standard for Oh My Zsh](https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide#indentation), and match the rest of this file.

mcornella:
`compinit` needs to be called because many plugins rely on it to call `compdef`.

mfarrugi:
@mcornella compinit  needs to be called once, at the end. Sometimes ohmyzsh is used as a library, instead of a package manager (eg. with zgen) and we would like for ohmyzsh to not call compinit a second time, as it can be perceptibly slow.

It's fine for this not to be upstreamed, as downstream package managers can be more explict about piecing out ohmyzsh.

For anyone else that stumbles across this, https://github.com/mfarrugi/.dots/blob/master/zsh/_zsh_plugins.zsh#L16 is my config for zplug which doesn't end up calling compinit twice.

PR #4520, opened 2015-10-19
Fix open_command inside tmux

eiel:
I want to use `open_command` inside tmux.

```
$ tmux
$ nohup open .
appending output to nohup.out
$ cat nohup.out
nohup: can't detach from console: No such file or directory
```

So, I modified `open_command` what not use nohup command inside tmux.

see also: http://stackoverflow.com/questions/23898623/nohup-cant-detach-from-console#comment36797854_23898623

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -17,6 +17,7 @@ function take() {
 
 function open_command() {
   local open_cmd
+  local nohup
 
   # define the open command
   case "$OSTYPE" in
@@ -28,7 +29,14 @@ function open_command() {
               ;;
   esac
 
-  nohup $open_cmd "$@" &>/dev/null
+  # Use nohup command
+  if [ ${TMUX:-NOT_USE_TMUX} = "NOT_USE_TMUX" ]; then
+    nohup="nohup"
+  else
+    nohup=""
+  fi
+
+  $nohup $open_cmd "$@" &>/dev/null
 }
 
 #


ada-u:
:+1:

mcornella:
I can't reproduce this. The output of `nohup` is silenced in the current version of the `open_command` function. Let me see some diagnostic info, please post the output of:
```zsh
which -a open
which -a open_command
nohup --version
```

eiel:
```
$ which -a open
/usr/bin/open
```
```
$ which -a open_command
open_command () {
	local open_cmd
	local nohup
	case "$OSTYPE" in
		(darwin*) open_cmd="open"  ;;
		(cygwin*) open_cmd="cygstart"  ;;
		(linux*) open_cmd="xdg-open"  ;;
		(*) echo "Platform $OSTYPE not supported"
			return 1 ;;
	esac

	nohup $open_cmd "$@" &> /dev/null
}
```
```
$ nohup --version
nohup: illegal option -- -
usage: nohup [--] utility [arguments]
```
I did not know how to determine the version of nohup command 

Sorry, I was wrong to write an example.
I fixed example

```diff
$ tmux
- $ open .
+ $ nohup open .
appending output to nohup.out
$ cat nohup.out
nohup: can't detach from console: No such file or directory
```

mcornella:
Oh, then you're not really using open_command. You should do `open_command .` which is the platform-agnostic function, or just `open .` since you're on OSX. The `nohup` in your case is not necessary.

lib/functions.zsh
```diff
@@ -28,7 +29,14 @@ function open_command() {
               ;;
   esac
 
-  nohup $open_cmd "$@" &>/dev/null
+  # Use nohup command
+  if [ ${TMUX:-NOT_USE_TMUX} = "NOT_USE_TMUX" ]; then
```
sachin21: @eiel I think better is:
```
  if [ ${TMUX:-NOT_USE_TMUX} = "NOT_USE_TMUX" ]; then
    nohup $open_cmd "$@"
  else
    $open_cmd "$@"
  fi  &>/dev/null
```

sachin21: @mcornella @apjanke What do you think?

mcornella: That is simply not needed. See https://github.com/robbyrussell/oh-my-zsh/pull/4520#issuecomment-149446951. The current version works correctly.

sachin21: Okay Thanks for the quicky reply :+1:

eiel:
This problem is that open_command is not working inside tmux .
open_command the web-search, jira, such as in and have been utilized and node plugin, these plugin uses open_command.
open_command and to open a URL in a web browser is being used.
However , it does not open a web browser and are inside tmux.
I am in the environment remains this open_command does not work , and would like to to move This is the purpose of this Pull Request.

apjanke:
I can reproduce, with OS X 10.9.5, zsh 5.0.2, tmux 2.1.

The problem isn't that @eiel is adding an unnecessary `nohup`: he's just demonstrating the breakage. The problem is that inside `tmux`, `nohup` fails, and the `open` doesn't happen. So the unconditional `nohup` that `open_command` is doing means it doesn't work properly when called inside a `tmux` session.

For example, here's a fresh `tmux` session I fired up locally inside iTerm2. The `open` works, opening the text file in BBEdit, my default text editor. The `open_command` call has no effect.

```
$ open omz_diagdump_20151008-223116.txt
[~]
$ open_command omz_diagdump_20151008-223116.txt
[✘ ~]
$
```

I don't know exactly what the semantics of `open_command` should be, so I'm not sure what the exact change should be, but it does seem that something like this PR is needed.

mcornella:
Ok, thanks for chiming in. I see now what he meant. 

Weirdly, when I tested this on my setup it worked correctly. I'll see if we need this or there is some other alternative that works everywhere though, I think I added the `nohup` part to be sure, but it may be replaced by [starting the process with `&| / &!`](http://zsh.sourceforge.net/Doc/Release/Jobs-_0026-Signals.html#Jobs):

> If a job is started with ‘&|’ or ‘&!’, then that job is immediately disowned. After startup, it does not have a place in the job table, and is not subject to the job control features described here.

Try this one-liner instead:
```zsh
$open_cmd "$@" &>/dev/null &|
```

mcornella:
I've submitted that change in the form of a commit at https://github.com/mcornella/oh-my-zsh/commit/454e89cd31b38e56a7c03621c4cf6f240b15fa5d.

eiel:
> $open_cmd "$@" &>/dev/null &|
 Awesome! :bow: :+1: 
It was confirmed that it is working well. :smile:

mcornella:
Ok, I've submitted https://github.com/robbyrussell/oh-my-zsh/pull/4534 to test separately. +1 there if it works for you

PR #4523, opened 2015-10-19
Mira theme: define `jenv_prompt_info` if not defined

defektive:
Should no longer see 'zsh: command not found: jenv_prompt_info'


apjanke:
May as well just provide this in `lib/themes-and-appearance.zsh` with the other dummy `*_prompt_info` functions. #4588 does so.

PR #4524, opened 2015-10-19
Fixed special chars error on plugins names and paths

jeanCarloMachado:
I was getting the following error:

```
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too many arguments
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too many arguments
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too many arguments
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too many arguments
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too many arguments
is_plugin:cd:4: too many arguments
is_plugin:cd:3: too man..
```

Due to special chars on plugins paths.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -38,12 +38,11 @@ done
 is_plugin() {
   local base_dir=$1
   local name=$2
-  test -f $base_dir/plugins/$name/$name.plugin.zsh \
-    || test -f $base_dir/plugins/$name/_$name
+   [[ -f $base_dir/plugins/$name/$name.plugin.zsh ]] || [[ -f $base_dir/plugins/$name/_$name ]]
 }
 # Add all defined plugins to fpath. This must be done
 # before running compinit.
-for plugin ($plugins); do
+for plugin ("$plugins"); do
   if is_plugin $ZSH_CUSTOM $plugin; then
     fpath=($ZSH_CUSTOM/plugins/$plugin $fpath)
   elif is_plugin $ZSH $plugin; then


apjanke:
Looks like a good idea.

Can you give me an example of some code to reproduce the original issue? (e.g. what were the plugin directory names that were causing problems?)

jeanCarloMachado:
I'm using oh-my-zsh on an Archlinux box installed via AUR. 
The plugins are the following: 

```plugins=(git archlinux common-aliases systemd vi-mode web-search)```

I'll try to reproduce the error on a docker machine.

apjanke:
Oh, hahaha: GitHub's emoji support is changing the string "`:cd:`" in your problem description into literal "💿" CD emoji icons. I totally misread what was going on here, I think. Could you wrap that code in triple-backquotes to get it to render verbatim?


screen shot 2015-10-31 at 4 41 31 pm


What special chars are on your plugins path? Where do you have OMZ installed on this machine?

I'm not sure this fix is fully correct: switching to `[[ ... ]]` will get rid of the "too many arguments" error, but if word splitting is happening in this context, then I think the test might just silently return false negatives for files with spaces or those special chars in them. I think you may need quotes there, too, like `[[ -f "$base_dir/plugins/$name/$name.plugin.zsh" ]]`.

Are there any options affecting word splitting turned on here? Normally `zsh` doesn't need quoting around much. Maybe there should be an `emulate -L zsh` added here for more robust protection.

I'd like to see a repro case before merging this, so we can get to the bottom of it.

mcornella:
If this cannot be reproduced anymore I'll consider just closing it. Thanks

mcornella:
Perhaps the problem is the `$base_dir` variable, which could happen if either `$ZSH` or `$ZSH_CUSTOM` have spaces in it. Let me know if that may be it

PR #4528, opened 2015-10-20
Use builtin cd and pwd in tmux plugin

thomdixon:
Use the builtin `cd` and `pwd`, as these could have been overridden by aliases or other plugins.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -37,7 +37,7 @@ if which tmux &> /dev/null
 
 
 	# Get the absolute path to the current directory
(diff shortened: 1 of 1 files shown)


apjanke:
You can remove the `cd`, `pwd`, and `dirname` dependencies entirely by just doing `zsh_tmux_plugin_path=${0:h}`. (The `local` doesn't do anything here, because this isn't run inside a function, at least in OMZ.)

If you want to enable `local`, then you can wrap most of this plugin in an anonymous or named function. You'll need to move the use of `$0` outside the function, though.

mcornella:
This same change is done in #4412

PR #4534, opened 2015-10-21
Drop the use of nohup on OSX for tmux compatibility

mcornella:
Apparently `nohup` is not needed in OSX, and using it has the side that it may break under tmux with the error

    nohup: can't detach from console: No such file or directory

For now, this commit only drops the use of `nohup` in OSX. But it *may* not be necessary under other similar platforms.

Fixes #4520

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -28,7 +28,10 @@ function open_command() {
               ;;
   esac
 
-  nohup $open_cmd "$@" &>/dev/null
+  # don't use nohup on OSX
+  [[ "$OSTYPE" != darwin* ]] && open_cmd="nohup $open_cmd"
+
+  $open_cmd "$@" &>/dev/null
 }
 
 #
@@ -67,7 +70,7 @@ function try_alias_value() {
 #
 # Arguments:
 #    1. name - The variable to set
-#    2. val  - The default value 
+#    2. val  - The default value
 # Return value:
 #    0 if the variable exists, 3 if it was set
 #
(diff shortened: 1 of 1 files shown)


apjanke:
Works more or less as expected for me on OS X 10.9.5, in and out of `tmux`. (That is, the file gets opened in native app per file associations, and command returns immediately.)

However, it always returns true, even when given a bad argument. I think the underlying commands all report a useful exit status; it would be nice to expose that.

```
[~]
$ open no-such-file.txt
The file /Users/janke/no-such-file.txt does not exist.
[✘ ~]
$ echo $?
1
[~]
$ open_command no-such-file.txt
[~]
$ echo $?
0
```

I don't think backgrounding or disowning the job is necessary on OS X; `open` uses IPC to invoke an application that's in the desktop session and is a child of `launchd`, and hands the file off to it. Similar with `cygstart` on Windows;  the Windows program that it hands off to is running in the desktop session, and isn't a child of the shell or terminal. Since all these "open" commands are IPC handoff things, backgrounding or disowning might not be needed at all. And in the case where there was an error doing the handoff or resolving a program, I think you'd want that status info synchronously.

mcornella:
Ok I've replaced the current solution with just droping `nohup` under OSX. I can't test that this works everywhere on OSX but you seem pretty confident that `open` doesn't need to be disowned. In fact, `xdg-open` in my debian system also behaves the same way, but I recall some system where calling that without `nohup` didn't work well. 

Anyway, I'm finally merging this one.

PR #4537, opened 2015-10-22
Add OSX support for command-not-found

agireud:
Added support for Brew on OSX.
Using command-not-found for Brew found at https://github.com/Homebrew/homebrew-command-not-found.

Close #4150

### DIFF
--- a/plugins/command-not-found/command-not-found.plugin.zsh
+++ b/plugins/command-not-found/command-not-found.plugin.zsh
@@ -23,3 +23,11 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if type brew &> /dev/null; then
+  if brew command command-not-found-init > /dev/null 2>&1; then
+    eval "$(brew command-not-found-init)";
+  fi
+fi


plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,11 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if [ -e /usr/local/bin/brew ]; then
+    if [ $(brew tap | grep homebrew/command-not-found) ]; then
```
sachin21: It should use 2 spaces.
Please change spaces from 4 spaces.

sachin21: ref: https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide

agireud:
Updated to conform to style guidelines.

thomdixon:
Works fantastically well. Thanks for this @agireud!

plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,13 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if [ -e /usr/local/bin/brew ]; then
+  if [ $(brew tap | grep homebrew/command-not-found) ]; then
```
sachin21: I suggest this:
```
  if brew tap | command grep -q homebrew/command-not-found; then
    ...something
```
`-q` is able to hide grep's result.

agireud:
Added -q to grep command to hide the result as per @sachin21's recommendation.

plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,13 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if [ -e /usr/local/bin/brew ]; then
```
dunn: FWIW: it's not encouraged, but it's possible to install Homebrew in locations other than `/usr/local`.

plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,13 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
...
+  if brew tap | command grep -q homebrew/command-not-found; then
```
xu-cheng: This is unnecessary as it's already covered by `brew command command-not-found-init`

plugins/command-not-found/command-not-found.plugin.zsh
bfontaine: You should probably check for `which brew` instead.

plugins/command-not-found/command-not-found.plugin.zsh
agireud: This check is to ensure that the homebrew/command-not-found tap has been installed, not to check the actual command being run which is done in the next line.

plugins/command-not-found/command-not-found.plugin.zsh
agireud: Noted, that is probably a better approach.

plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,13 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if which brew > /dev/null; then
+  if brew tap | command grep -q homebrew/command-not-found; then
+    if brew command command-not-found-init > /dev/null; then
```
bfontaine: Testing `brew command command-not-found-init` is sufficient, you don’t need to check that the repo is tapped because the command wouldn’t exist if it weren’t.

plugins/command-not-found/command-not-found.plugin.zsh
xu-cheng: `brew command` is used to check whether a command is installed. If you want to hide `Error: Unknown command: command-not-found-init`, all you need to do is using `brew command command-not-found-init > /dev/null 2>&1`

agireud: Understood.  Thanks!

plugins/command-not-found/command-not-found.plugin.zsh
agireud: Not a problem, I was a little bit confused but xu-cheng explained further to clarify.  Thanks!

plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -23,3 +23,11 @@ if [ -f /usr/libexec/pk-command-not-found ]; then
         return $retval
     }
 fi
+
+# OSX command-not-found support
+# https://github.com/Homebrew/homebrew-command-not-found
+if which brew > /dev/null 2>&1; then
```
sachin21:
```
if type brew &> /dev/null; then
```

Short than `2>&1`

apjanke:
Dupe of #4150 and #3278. Could these be reconciled? I think we should just use this one and close the others, because #4150 is hardcoded to only look in the default `brew` prefix, and #3278 is obsolete WRT current Homebrew.

plugins/command-not-found/command-not-found.plugin.zsh
agireud: Done.

tresni:
:+1: works for me!  I would say #4150 and #3278 should be closed in favor of this.

agireud:
So has a consensus been reached for this?  How many testers are normally required before approval?

mcornella:
Merging this one. Thanks guys!

PR #4539, opened 2015-10-23
Update uninstall.sh

Gnouc:
- Need dot `.` for POSIX compliant instead of `source`
 - Use modern syntax for command substitution
 - Fix missing double quotes

### DIFF
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -21,7 +21,7 @@ if [ -f ~/.zshrc.pre-oh-my-zsh ] || [ -h ~/.zshrc.pre-oh-my-zsh ]; then
 
   mv ~/.zshrc.pre-oh-my-zsh ~/.zshrc;
 
-  . ~/.zshrc;
+  echo "Your original zsh config was restored. Please restart your session."
 else
   if hash chsh >/dev/null 2>&1; then
     echo "Switching back to bash"
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -18,14 +18,14 @@ then
 
   if [ -f ~/.zshrc ] || [ -h ~/.zshrc ]
   then
-    ZSHRC_SAVE=".zshrc.omz-uninstalled-`date +%Y%m%d%H%M%S`";
+    ZSHRC_SAVE=".zshrc.omz-uninstalled-$(date +%Y%m%d%H%M%S)";
     echo "Found ~/.zshrc -- Renaming to ~/${ZSHRC_SAVE}";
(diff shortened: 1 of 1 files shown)


tools/uninstall.sh
```diff
@@ -18,14 +18,14 @@ then
 
   if [ -f ~/.zshrc ] || [ -h ~/.zshrc ]
   then
```
sachin21: Can you change this style?
```
if [ -f ~/.zshrc ] || [ -h ~/.zshrc ]; then
```

ref: https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide

Gnouc: @sachin21 well, updated it.

sachin21: :+1: :+1: :+1: :+1: :+1:

Gnouc:
@robbyrussell

ncanceill:
HI, and thanks for contributing.

I agree on everything. Keep in mind that `.` versus `source` is not an easy debate to settle.

On the one hand `.` is indeed the POSIX standard, and some shells do not support `source`. On the other hand, some other shells (mainly C-shells) do not support `.`.

Since the uninstall script should only be used through the `uninstall_oh_my_zsh` function which explicitly uses `sh`, it should work OK. However if someone tried to run it with `tcsh` it would fail. Maybe we should add a shebang?

@mcornella an opinion?

Gnouc:
@ncanceill: I think the shebang didn't have and the best thing we can do is add the instruction for user to use a POSIX shell.

In case of csh variants, this script won't work, it used many syntax that not supported by them, the if syntax, the command substitution `$(...)`, the dot `.` like you said. Even the `read -r -p` won't.

apjanke:
How about just removing the `source ~/.zshrc;` entirely? This script runs inside a noninteractive subshell, and under a shell which is not `zsh`. It doesn't seem to make a lot of sense to run a `zsh` startup file inside it.

I like the idea of adding a shebang line though. Not that it will do anything to the script's behavior as we use it, but it would serve as a nice bit of doco indicating that this file is to be executed as a script (instead of sourced) and indicating which shell it is to run under.

Gnouc:
@apjanke Removing the source seems to be good. We should also update documentation and print some message to notice the users that they must restart their session.

If we're ok with that, I will update the changes.

ncanceill:
I agree on removing the `source` statement altogether.

Thanks guys.

Gnouc:
@ncanceill

Gnouc:
What's state of this pull guys?

mcornella:
Done. Thank you @Gnouc!

PR #4542, opened 2015-10-23
Update autojump.plugin.zsh

alexparker:
I was adding autojump on a new machine and annoyingly forgot that autojump was to be installed on its own. This may be helpful.

### DIFF
--- a/plugins/autojump/autojump.plugin.zsh
+++ b/plugins/autojump/autojump.plugin.zsh
@@ -19,5 +19,5 @@ if [ $commands[autojump] ]; then # check if autojump is installed
     . `brew --prefix`/etc/autojump.sh
   fi
 else
-  echo 'Please install autojump to enable the plugin (https://github.com/wting/autojump)';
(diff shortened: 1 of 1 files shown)


plugins/autojump/autojump.plugin.zsh
```diff
@@ -18,4 +18,6 @@ if [ $commands[autojump] ]; then # check if autojump is installed
   elif [ $commands[brew] -a -f `brew --prefix`/etc/autojump.sh ]; then # mac os x with brew
     . `brew --prefix`/etc/autojump.sh
   fi
+else
+  echo 'Please install autojump to enable the plugin (https://github.com/wting/autojump)';
```
sachin21: Why do you have added a semicolon at the end of the line?
I think semicolon is not needed in this case echo.
because [oh-my-zsh's coding styles] has no semicolon with on line command.

alexparker:
Removed the semicolon. It's just a habit :) haha

PR #4544, opened 2015-10-24
customize segment separator of agnoster theme

allenlsy:
change SEGMENT_SEPARATOR of agnoster theme to a better choice

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -43,7 +43,7 @@ CURRENT_BG='NONE'
   # This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
   # what font the user is viewing this source code in. Do not replace the
   # escape sequence with a single literal character.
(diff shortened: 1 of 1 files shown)


mcornella:
Can you post a screenshot of before and after? This looks to me like the typical issue with non-updated Powerline fonts.

allenlsy:
So this is the terminal before:

screen shot 2015-10-24 at 11 56 46 pm

This is the terminal after:

screen shot 2015-10-24 at 11 55 02 pm

apjanke:
Yes, this looks like you have an old, incompatible Powerline font installed. The current code is correct, and you probably need to update your font. See #4091, #4083.

Have a look at [the comments above that character definition](https://github.com/robbyrussell/oh-my-zsh/blob/51677da85f98fb8003ea915edbef138df2f577b7/themes/agnoster.zsh-theme#L37-L46). This change is exactly what it is asking people not to do.

allenlsy:
@apjanke Thanks for you comments. You are correct. I will take a look at the comments

PR #4545, opened 2015-10-24
Added paulomagalhaes theme

magalhaespaulo:


sachin21:
> Do NOT send us themes

Check it out: https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes

@magalhaespaulo

mcornella:
As @sachin21 said, we're not bundling any new themes. You can make a repository for yours and add it to the [External themes section](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) of the wiki.

PR #4557, opened 2015-10-29
installer: fix ordering of Cygwin MSYS git check

### LINKED ISSUE #4551: Installer fails on windows if using Windows Git
Git clones the repo into C:/cygrive/c/... instead of /cygdrive/c/...

apjanke:
The test added in #3646 to prevent use of the incompatible MSYS `git` isn't working because the test ended up after the `git clone` call. This moves it to the right location.

Fixes #4551.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -53,11 +53,6 @@ main() {
     echo "Error: git is not installed"
     exit 1
   }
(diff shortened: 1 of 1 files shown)


norberto5757:
Now is created nao me pay

PR #4558, opened 2015-10-29
add more useful alias for macports

ibrother:
add these alias

* `prm pkg_name` to remove the pkg and its dependencies

* `pq pkg_name` to search the pkg

* `pqb path-to-filename` to find which pkg provides this file

* `pqf pkg_name` to list the files belong to that pkg

* `pvar pkg_name` to list the variants of the pkg


Marcholio:
This could be either closed or merged after resolving the conflicts as it has been open for almost three years.

Issue #4561, opened 2015-10-29
agnoster theme doesn't honor oh-my-zsh.hide-status


MikaelSmith:
For repositories where querying git for the prompt is very slow, it'd be nice to be able to disable it. The agnoster theme doesn't currently support a method to do this, but the standard appears to be
`git config --add oh-my-zsh.hide-status 1`.

drasill:
I love agnoster, and as my $HOME is on a git repo, I really need this too :)

jjdonov:
@MikaelSmith @drasill you can modify the them to honor the status, in agnoster.zsh-theme (or you custom copy) simply add an if statement:

```zsh
## Main prompt
build_prompt() {
  RETVAL=$?
  prompt_status
  prompt_virtualenv
  prompt_context
  prompt_dir
  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]];then
    prompt_git
    prompt_hg
  fi
    prompt_end
}
```

MikaelSmith:
My PR did something similar, but applying it for hg as well makes sense.

benatkin:
`hide-dirty` speeds it up for me!

``` bash
# current repo
git config --add oh-my-zsh.hide-dirty 1
# global (I use this on my VirtualBox guest where I have vagrant synced folders)
git config --global --add oh-my-zsh.hide-dirty 1
```

kjakapat:
In stead of completely disable prompt features. You can tweak your `agnoster.zsh-theme` to work specifically with your `git` repo by:

1. comment all lines that have `vcs_info`,
2. edit line in `prompt_git()` function from:
`ref="$vcs_info_msg_0_"` 
to 
`ref="$(git branch 2>/dev/null | grep -Po '(?<=\* ).*$')"` 

So, basically, you now disable all `vcs_info` actions and use a native `git` command instead to check a directory statuses. As a result, you still can see your useful git prompt and with a speed that as fast as you work in non-git directory.

MikaelSmith:
That sounds like a useful approach. My PR is pretty stale, I'm tempted to close it. Is there still value to it?

kjakapat:
Apart from suppress the feature, you can try [this](https://stackoverflow.com/questions/12765344/oh-my-zsh-slow-but-only-for-certain-git-repo/47606567#47606567).

ExpandingMan:
Almost a year later and this still seems not to be working for me, any update on what's the best approach here?

[...]

PR #4562, opened 2015-10-29
Disable prompt_git for oh-my-zsh.hide-status

MikaelSmith:
Fixes issue #4561

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -196,8 +196,10 @@ build_prompt() {
   prompt_virtualenv
   prompt_context
   prompt_dir
-  prompt_git
-  prompt_hg
+  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
+    prompt_git
+    prompt_hg
+  fi
   prompt_end
 }
 


themes/agnoster.zsh-theme
```diff
@@ -85,6 +85,9 @@ prompt_context() {
 
 # Git: branch/detached head, dirty status
 prompt_git() {
+  if [[ "$(git config --get oh-my-zsh.hide-status 2>/dev/null)" == "1" ]]; then
```
sachin21: I suggest this:
```
  if ! git config --get oh-my-zsh.hide-status &> /dev/null; then
    *--* something *--*
```

apjanke: I don't think that accomplishes the same thing. That's just testing whether the configuration setting is present, and treats its set values as equivalent.

```
[~/.oh-my-zsh on ⇄ master]
$ if ! git config --get oh-my-zsh.hide-status; then echo yep; fi
yep
[~/.oh-my-zsh on ⇄ master]
$ git config oh-my-zsh.hide-status 0
[~/.oh-my-zsh on ⇄ master]
$ if ! git config --get oh-my-zsh.hide-status; then echo yep; fi
0
[~/.oh-my-zsh on ⇄ master]
$ git config oh-my-zsh.hide-status 1
[~/.oh-my-zsh]
$ if ! git config --get oh-my-zsh.hide-status; then echo yep; fi
1
[~/.oh-my-zsh]
$
```

If you use that test, then it will take the early return for any repo where that config setting is not defined, and it will ignore its value and proceed for any repo where it is defined.

drasill:
Tested it, LGTM.

MikaelSmith:
Updated with a suggestion from the issue.

troglotit:
I've tested it, works for me. +1 on merging

MikaelSmith:
This seems to have been moved out of themes somewhere. Closing it as no longer relevant.

PR #4564, opened 2015-10-30
Handle systems where /etc/shells does not exist

fornwall:
The /etc/shells file may not exist on all system - one of them being when running on a non-rooted Android environment such as [Termux](http://termux.com/).

I got a couple of reports that Oh-My-Zsh does not install on Termux, so there is definitely interest in running it on Android phones & tablets. Once fixing the `install.sh` file it works fine!

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -24,10 +24,12 @@ main() {
   # which may fail on systems lacking tput or terminfo
   set -e
 
-  CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
-  if [ ! $CHECK_ZSH_INSTALLED -ge 1 ]; then
-    printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
-    exit
+  if [ -f /etc/shells ]; then
+    CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
+    if [ ! $CHECK_ZSH_INSTALLED -ge 1 ]; then
+      printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
+      exit
+    fi
   fi
   unset CHECK_ZSH_INSTALLED
 
@@ -90,12 +92,12 @@ main() {
   TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
   if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
     # If this platform provides a "chsh" command (not Cygwin), do it, man!
-    if hash chsh >/dev/null 2>&1; then
+    if [ -f /etc/shells ] && hash chsh >/dev/null 2>&1; then
       printf "${BLUE}Time to change your default shell to zsh!${NORMAL}\n"
       chsh -s $(grep /zsh$ /etc/shells | tail -1)
     # Else, suggest the user do so manually.
     else
-      printf "I can't change your shell automatically because this system does not have chsh.\n"
+      printf "I can't change your shell automatically because this system does not have chsh and /etc/shells.\n"
       printf "${BLUE}Please manually change your default shell to zsh!${NORMAL}\n"
     fi
   fi


tools/install.sh
```diff
@@ -90,12 +92,12 @@ main() {
   TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
   if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
     # If this platform provides a "chsh" command (not Cygwin), do it, man!
-    if hash chsh >/dev/null 2>&1; then
+    if [ -f /etc/shells ] && [ hash chsh >/dev/null 2>&1 ]; then
```
apjanke: I don't think you want the `[ ... ]` around the `hash` command in the second part of this line. It turns it into a broken textual condition test, and the `2>&1` inside the test is hiding the error message that results.

```
sh-3.2$ if [ hash chsh >/dev/null 2>&1 ]; then echo yup; fi
sh-3.2$ if [ hash chsh >/dev/null  ]; then echo yup; fi
sh: [: hash: unary operator expected
sh-3.2$
```

fornwall: Thanks a lot @apjanke! Fix and commits squashed.

fornwall:
For testers: Use the install.sh file from the branch to install oh-my-zsh with:

    sh -c "$(curl -fsSL https://raw.githubusercontent.com/fornwall/oh-my-zsh/etc-shells-may-not-exist/tools/install.sh)"

Test that the installer sets zsh as default shell if not already done, and that it handles the case when zsh is already the default shell as expected.

dar5hak:
Tested. Works on Termux 0.24.

deepakbndeepu:
Tested. Works on Termux 0.30.

fornwall:
@mcornella Is it possible to go forward here?

mcornella:
Sorry, I forgot about this one. It'll be merged shortly.

fornwall:
@mcornella Ping on this one :).

Let me know if there is anything I can help with to get this issue merged (or if you don't want to accept the change if it's outside scope)!

mcornella:
Sorry, I've been meaning to fix all issues related to the installer and I lost the perception of time. This should be fixed on #5169, which I'll merge soon even though it doesn't fix all issues.

fornwall:
@mcornella No problem, thanks for all your work! I'll await https://github.com/robbyrussell/oh-my-zsh/pull/5169 and close this one then.

fornwall:
@mcornella Note that comment at https://github.com/robbyrussell/oh-my-zsh/pull/5169/files#diff-55b046be8e081f97f37b7667e1c4bfa9R95 for https://github.com/robbyrussell/oh-my-zsh/pull/5169 to solve the problem experienced here.

apjanke:
Checking for `/etc/shells` is the way to go in the installer, since it defines the valid targets for a `chsh`. Doing a `which zsh` can return a `zsh` that is not in `/etc/shells`, and can't be used. So if `which zsh` is done, it still needs to be combined with a check for and in `/etc/shells`.

fornwall:
@apjanke Hm, right!

I guess the best solution is to check for both `command_exists chsh` and `test -r /etc/shells` before trying `chsh -s $(grep /zsh$ /etc/shells | tail -1)`, and hit the `I can't change your shell automatically` fallback otherwise. I'll update the comment.

fornwall:
@mcornella Should we await #5169, or could this PR be merged for now?

GabrielTK:
+1

sgeto:
Ping!

stach:
Tested. Works on Termux 0.56.

mcornella:
#5169 pulled this change and is now ready for testing (it introduces other fixes).

Thanks!

fornwall:
@mcornella Thanks! I verified the installation with the below command on a new Termux installation, and it seems to work fine:

    pkg install curl git zsh && sh -c "$(curl -fsSL https://raw.githubusercontent.com/mcornella/oh-my-zsh/refactor-installer/tools/install.sh)"

PR #4571, opened 2015-11-02
move ~/.zsh-update file to $ZSH_CACHE_DIR

### LINKED ISSUE #4550: Suggestion for better update checking
Hi!
I am using oh-my-zsh across multiple servers. I put oh-my-zsh folder in some distributed file system (MooseFS) so that the servers can share a same copy of oh-my-zsh. However I notice that oh-my-zsh check for update by reading a "LAST_EPOCH" variable in ~/.zsh-update. This makes it ask me for checking update when I login a new server even it is just updated when I login another server before.

Is it possible to move .zsh-update file into the git repository?  I think this may be more reasonable :)

Sincerely,
Chao

duchao0726:
Fixes #4550. The default path of file `.zsh-update` is changed to `$ZSH_CACHE_DIR`.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -7,7 +7,7 @@ function _current_epoch() {
 }
 
 function _update_zsh_update() {
-  echo "LAST_EPOCH=$(_current_epoch)" >! ~/.zsh-update
+  echo "LAST_EPOCH=$(_current_epoch)" >! ${ZSH_CACHE_DIR}/.zsh-update
 }
 
 function _upgrade_zsh() {
@@ -29,9 +29,9 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
+if [ -f ${ZSH_CACHE_DIR}/.zsh-update ]
 then
-  . ~/.zsh-update
+  . ${ZSH_CACHE_DIR}/.zsh-update
 
   if [[ -z "$LAST_EPOCH" ]]; then
     _update_zsh_update && return 0;
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -1,6 +1,12 @@
+# Set ZSH_CACHE_DIR to the path where cache files should be created
+# or else we will use the default cache/
+if [[ -z "$ZSH_CACHE_DIR" ]]; then
+  ZSH_CACHE_DIR="$ZSH/cache/"
+fi
+
 # Check for updates on initial load...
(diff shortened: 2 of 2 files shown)


tools/check_for_upgrade.sh
```diff
@@ -29,9 +29,9 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
+if [ -f ${ZSH_CACHE_DIR}.zsh-update ]
```
apjanke: Should probably have a `/` here, like `${ZSH_CACHE_DIR}/.zsh-update`. We probably shouldn't assume `$ZSH_CACHE_DIR` to have a trailing slash, since it may be set by the user.

ncanceill:
Can you please specifically include "Fixes #4550" in the body of the PR?

This way your issue will auto-close when this PR is merged.

tools/check_for_upgrade.sh
duchao0726: Hi, thank you for the suggestion. I mistakenly thought it would cause problem to use a `/` here since the variable `$ZSH_CACHE_DIR` in `oh-my-zsh.sh` also contains a `/`. It's fixed now.

ncanceill:
Excellent. Please be so kind as to squash your commits, in order to keep the history clean and atomic.

duchao0726:
Hi, the commits are squashed now.

ncanceill:
OK this looks good to me!

One last thing: how do we handle the switch for people who still have a `.zsh-update` in their home folder?

tools/check_for_upgrade.sh
apjanke: Nope, that's fine: in Unix, repeated `/` characters are (mostly) treated the same as a single one. Mostly to make this use case easier, I think.

duchao0726:
Hi ncanceill,

I think there should be something like `mv ~/.zsh-update ${ZSH_CACHE_DIR}/.zsh-update`. However I'm not sure about how I can compare `${ZSH_CACHE_DIR}` and `~/`.

How about this:
```
if [ -f ~/.zsh-update ] && [ ! -f ${ZSH_CACHE_DIR}/.zsh-update ]; then
    mv ~/.zsh-update ${ZSH_CACHE_DIR}/.zsh-update;
fi
```

JoshuaD84:
Do we have any idea how long this might take to be accepted? It's a great idea and I'd like to have it.

duchao0726:
I added some code to migrate `.zsh-update` file. This would keep home directory clean.

mcornella:
Pulled in e96a7b728a9c15f5615f4e41a79a5f0fe2b97712

PR #4572, opened 2015-11-02
Avoid using "PREFIX" that cause nvm to complain about.

aioutecism:
`nvm` complains about `PREFIX` when using sunrise theme:

```
nvm is not compatible with the "PREFIX" environment variable: currently set to "---"
Run `unset PREFIX` to unset it.
```

Renaming `PREFIX` to something else fixed this issue.

### DIFF
--- a/themes/sunrise.zsh-theme
+++ b/themes/sunrise.zsh-theme
@@ -11,9 +11,9 @@ B=$fg_no_bold[blue]
 RESET=$reset_color
 
 if [ "$USER" = "root" ]; then
-    PROMPTCOLOR="%{$R%}" PREFIX="-!-";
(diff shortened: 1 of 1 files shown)


apjanke:
It looks like that `$PREFIX` value never changes inside a session, so it can be used just once at prompt setup time. How about instead of changing the name, wrapping stuff in a function and using `local` to avoid leaking variables, like [this](https://github.com/apjanke/oh-my-zsh/tree/sunrise-no-PREFIX)?

florrain:
Pushing for this first fix to be merged in. This PR has been opened for 6 months and this is still impacting users (like me :no_mouth:)

mcornella:
Thanks for letting me know!

aioutecism:
Thx!

PR #4576, opened 2015-11-03
Add ggfr command to ask for rebase with master before push --force

aguspina:


mcornella:
This should be a practice internalized by the user.

PR #4577, opened 2015-11-03
Added plugin for desk

mblarsen:
Added completion of available desks for [desk](https://github.com/jamesob/desk).

### DIFF
--- a/plugins/desk/desk.plugin.zsh
+++ b/plugins/desk/desk.plugin.zsh
@@ -29,14 +29,17 @@ _desk() {
     case $state in
         (cmds)
             _values "desk commands" \
+                '.[Activate a desk]' \
                 'go[Activate a desk]' \
+                'ls[List all desks along with description]' \
+                'list[List all desks along with description]' \
                 'init[Initialize desk configuration]' \
-                'list[List all desks along with description]'
+                'version[Show version information]'
             ret=0
             ;;
         (desk)
             case $words[2] in
-                (go)
+                (go|.)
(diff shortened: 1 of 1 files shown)


plugins/desk/desk.plugin.zsh
```diff
@@ -0,0 +1,50 @@
...
+    done
+
+    _arguments \
+      '1:command:->cmds' \
+      '2:desk:->desk'
+
+    case $state in
+        (cmds)
+            _values "desk commands" \
+                'go[Activate a desk]' \
```
jamesob: `.` is also synonymous with `go`

plugins/desk/desk.plugin.zsh
```diff
@@ -0,0 +1,50 @@
...
+                'init[Initialize desk configuration]' \
+                'list[List all desks along with description]'
```
jamesob: `ls` as well

plugins/desk/desk.plugin.zsh
mblarsen: Added

plugins/desk/desk.plugin.zsh
mblarsen: Added

jamesob:
Nice! LGTM

mcornella:
I know this is old but now the project includes a [guide to set it up within OMZ](https://github.com/jamesob/desk#using-oh-my-zsh) so I don't think we need a plugin.

PR #4579, opened 2015-11-04
autocomplete list of maven modules

jprudent:
this can autocomplete the module name when typing

```
mvn -rf : 
```


plugins/mvn/mvn.plugin.zsh
```diff
@@ -180,6 +180,8 @@ function listMavenCompletions {
         archetype:generate generate-sources 
         cobertura:cobertura
         -Dtest= `if [ -d ./src/test/java ] ; then find ./src/test/java -type f -name '*.java' | grep -v svn | sed 's?.*/\([^/]*\)\..*?-Dtest=\1?' ; fi`
+        # list modules, useful combined with -rf
+        : `ls */pom.xml | sed -E 's;([^/]*)/pom.xml;:\1;'`
```
apjanke: This could probably be done in native `zsh` like `poms=(*/pom.xml(N)); echo ${poms:h}`. Might make this a bit more responsive by saving the extra shell-outs.

PR #4580, opened 2015-11-04
added path plugin

dbasinge:
Added path plugin which prints out path in nice format.


plugins/path/path.plugin.zsh
```diff
@@ -0,0 +1,4 @@
+# Prints out path nicely
+alias path='echo -e ${PATH//:/\\n}
```
apjanke: Is this missing a closing quote?

dbasinge:
Thanks, added missing quote

mcornella:
This is not suitable to occupy a whole plugin.

PR #4582, opened 2015-11-05
add plugin:percol

CodeFalling:
See [README](https://github.com/CodeFalling/oh-my-zsh/blob/percol/plugins/percol/README.md)

### DIFF
--- a/plugins/percol/percol.plugin.zsh
+++ b/plugins/percol/percol.plugin.zsh
@@ -1,0 +1,22 @@
+if which percol &> /dev/null; then
+    function percol_select_history() {
+        local tac
+        tac="tail -r"
+        BUFFER=$(fc -l -n 1 | eval $tac | percol --query "$LBUFFER")
+        CURSOR=$#BUFFER
+        zle -R -c
+    }
+
+    zle -N percol_select_history
+    bindkey '^R' percol_select_history
+fi
+
+if which marks &> /dev/null; then
+    function percol_select_marks() {
+        BUFFER=$(marks | percol --query "$LBUFFER" | awk '{print $3}')
+        CURSOR=$#BUFFER # move cursor
+        zle -R -c       # refresh
+    }
+    zle -N percol_select_marks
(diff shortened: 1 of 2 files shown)


plugins/percol/percol.plugin.zsh
```diff
@@ -0,0 +1,22 @@
...
+        BUFFER=$(fc -l -n 1 | eval $tac | percol --query "$LBUFFER")
+        CURSOR=$#BUFFER
+        zle -R -c
+    }
+
+    zle -N percol_select_history
+    bindkey '^R' percol_select_history
+fi
+
+if which marks &> /dev/null; then
```
apjanke: Should this test be nested inside the `if which percol` test above, since it also depends on `percol` being installed?

CodeFalling: You are right,I will fix it right now.

CodeFalling:
Anything else to do here?

popstas:
I've tested history search, I use mostly same code in my .zshrc for 2 months in Debian Squeeze, Ubuntu 14.04, Ubuntu 15.10 and all works fine.

Your plugin outputs empty history in gnome terminal, please use code from percol page, you need replace
```
tac="tail -r"
```
to
```
which gtac &> /dev/null && tac="gtac" || { which tac &> /dev/null && tac="tac" || { tac="tail -r" } }
```

And I don't know what is `marks` command and not tested it.

CodeFalling:
@popstas Thank you.I have update the code.By the way, `marks` is a command of [jump plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/jump)

popstas:
I've checked out marks functional too, it works, thank you, good plugin. I think that plugin ready to merge.
For my personal use I added one more line:
```
g() { percol --match-method regex --query="$*"; }
```
Usage:
```
find . | g
find . | g needle
```
But `g` conflicts with plugin `git`.

PR #4584, opened 2015-11-06
Added ndenv plugin

chiliburger:
# What

Added ndenv plugin for https://github.com/riywo/ndenv
# Why

So we can initialize ndenv in oh-my-zsh similar to how we initialize rbenv.

### DIFF
--- a/plugins/ndenv/ndenv.plugin.zsh
+++ b/plugins/ndenv/ndenv.plugin.zsh
@@ -1,0 +1,42 @@
+_homebrew-installed() {
+  type brew &> /dev/null
+}
+
+FOUND_NDENV=0
+ndenvdirs=("$HOME/.ndenv" "/usr/local/ndenv" "/opt/ndenv" "/usr/local/opt/ndenv")
+
+if _homebrew-installed && ndenv_homebrew_path=$(brew --prefix ndenv 2>/dev/null); then
+    ndenvdirs=($ndenv_homebrew_path "${ndenvdirs[@]}")
+    unset ndenv_homebrew_path
+fi
+
+for ndenvdir in "${ndenvdirs[@]}" ; do
+  if [ -d $ndenvdir/bin -a $FOUND_NDENV -eq 0 ] ; then
+    FOUND_NDENV=1
+
+    if [[ $NDENV_ROOT = '' ]]; then
+      NDENV_ROOT=$ndenvdir
+    fi
+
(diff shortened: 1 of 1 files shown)


apjanke:
Might want to stick all the code in a function and use `local` for the variables that aren't meant to stick around, instead of the cruder `unset`.

Could you also document it a bit with links to the `ndenv` program it's supporting and a description of what it does?

plugins/ndenv/ndenv.plugin.zsh
```diff
@@ -0,0 +1,42 @@
...
+if _homebrew-installed && ndenv_homebrew_path=$(brew --prefix ndenv 2>/dev/null); then
+    ndenvdirs=($ndenv_homebrew_path "${ndenvdirs[@]}")
+    unset ndenv_homebrew_path
+fi
+
+for ndenvdir in "${ndenvdirs[@]}" ; do
+  if [ -d $ndenvdir/bin -a $FOUND_NDENV -eq 0 ] ; then
+    FOUND_NDENV=1
+
+    if [[ $NDENV_ROOT = '' ]]; then
```
apjanke: If `$NDENV_ROOT` is already set by the caller, should that just be used directly, instead of searching for the installation location? Here, you get a situation where `$NDENV_ROOT` and `$PATH` may not match, if the caller supplied `$NDENV_ROOT` but it isn't the first extant location in your search order.

plugins/ndenv/ndenv.plugin.zsh
```diff
@@ -0,0 +1,42 @@
...
+
+    eval "$(ndenv init --no-rehash - zsh)"
+
+    alias nodes="ndenv versions"
+
+    function current_node() {
+      echo "$(ndenv version-name)"
+    }
+
+    function ndenv_prompt_info() {
```
apjanke: This PR should probably also add a dummy `ndenv_prompt_info` to [`lib/prompt_info_functions.zsh'](https://github.com/robbyrussell/oh-my-zsh/blob/b0b79e0abae5f1d228e16c36174bad2bf7ba5fee/lib/prompt_info_functions.zsh#L13-L17).

chiliburger:
@apjanke thanks for the feedback. Pretty much copied the rbenv plugin and changed references from rbenv to ndenv. Will get the suggested changes in shortly.

mcornella:
Not popular enough.

PR #4585, opened 2015-11-06
lib: add safe_alias function

ncanceill:
STATUS — **[[NOT READY]]**
### Summary

This PR adds the `safe_alias` function to `lib/functions` in order to prevent aliases defined in libs and plugins from overriding commands/functions/aliases already defined. It also introduces the `$ALLOW_OVERRIDE_ALIASES` config variable to allow overriding with a warning.
### Roadmap
- [ ] Discuss implementation
  - [ ] Fail when called with illegal arguments
  - [ ] Deal with flags of the `alias` built-in
  - [ ] Deal with the consequences of unset aliases, e.g. when used internally by a plugin
- [ ] Handle duplicate aliases
  - [ ] Intentional ones, e.g. reloading `~/.zshrc` with updated aliases
  - [ ] Unintentional ones, e.g. conflicting plugins
- [ ] Make all aliases safe
  - [ ] List the ones meant to be unsafe by design, e.g. `alias mv='nocorrect mv'`
### Contributions

The idea is from @ChrisJefferson, see https://github.com/robbyrussell/oh-my-zsh/issues/3682#issuecomment-78437079 and the discussion that followed

The implementation is from @blueyed, see https://github.com/robbyrussell/oh-my-zsh/pull/3990#issuecomment-111175759 (I only added parentheses to properly declare the `alias` variable as an array)

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -63,6 +63,32 @@ function try_alias_value() {
 }
 
 #
+# Alias only if not already a command/function/alias
+# Use option ALLOW_OVERRIDE_ALIASES to force overriding
+#
+# Arguments:
+#    1. my_alias='my alias value'
+#       Same argument you would pass to the "alias" built-in
+# Return value:
+#    0 if the alias was safely created, 73 otherwise
+#
+function safe_alias() {
+    emulate -L zsh
+    local cmd lhs rhs
+    cmd=(${(s:=:)1})
+    lhs=$cmd[1]
+    rhs=$cmd[2,-1]
+    if (( $+commands[$lhs] || $+functions[$lhs] || $+aliases[$lhs] )); then
+        if [[ $ALLOW_OVERRIDE_ALIASES = "true" ]]; then
+            >&2 echo "WARN: override \"$lhs\" with alias \"$rhs\""
+        else
+            return 73  # os.EX_CANTCREAT
+        fi
+    fi
+    alias $lhs="$rhs"
+}
+
+#
 # Set variable "$1" to default value "$2" if "$1" is not yet defined.
 #
 # Arguments:
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -37,6 +37,11 @@ ZSH_THEME="robbyrussell"
 # much, much faster.
 # DISABLE_UNTRACKED_FILES_DIRTY="true"
 
+# Uncomment the following line to allow aliases defined in libs and plugins to
+# override existing commands/functions/aliases. A warning will be printed for
+# every alias overriding an existing name.
+# ALLOW_OVERRIDE_ALIASES="true"
+
 # Uncomment the following line if you want to change the command execution time
 # stamp shown in the history command output.
 # The optional three formats: "mm/dd/yyyy"|"dd.mm.yyyy"|"yyyy-mm-dd"


lib/functions.zsh
```diff
@@ -63,6 +63,32 @@ function try_alias_value() {
...
+# Use option ALLOW_OVERRIDE_ALIASES to force overriding
+#
+# Arguments:
+#    1. my_alias='my alias value'
+#       Same argument you would pass to the "alias" built-in
+# Return value:
+#    0 if the alias was safely created, 73 otherwise
+#
+function safe_alias() {
+    local alias lhs rhs
```
apjanke: Maybe `emulate -L zsh` for safety?

lib/functions.zsh
```diff
@@ -63,6 +63,32 @@ function try_alias_value() {
...
+    alias=(${(s:=:)1})
+    lhs=$alias[1]
+    rhs=$alias[2,-1]
+    if (( $+commands[$lhs] || $+functions[$lhs] || $+aliases[$lhs] )); then
+        if [ "x$ALLOW_OVERRIDE_ALIASES" = "xtrue" ]; then
+            echo "WARN: override \"$lhs\" with alias \"$rhs\""
```
apjanke: Since this is a warning, should it go on STDERR instead?

lib/functions.zsh
```diff
@@ -63,6 +63,32 @@ function try_alias_value() {
...
+	    unset ALLOW_OVERRIDE_ALIASES
```
apjanke: Inconsistent indent size on this line.

ncanceill:
Want a fun shell command to run? Try this:
```
grep alias $ZSH/**/*.zsh | cut -f2 -d: | grep -vE "^[[:space:]]*#" | uniq | wc -l
     996
```
That's right, there are almost a thousand aliases defined in OMZ!

Should we brutally all make them safe with a quick `sed`?

***

I used `uniq` above because there are three duplicates. Two of them are in mutually exclusive `if` statements. The third one is the dual go/golang plugin.

Hey people, why do we have two Go plugins? (none of which is properly [documented](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins))

Actually, the "go" one is simply a symlink to the other one, but how do we keep that around @lukesteensen?

***

That leads us to: what if there are more duplicates to come? Should the function just silently fail? (which it will in its current form)

***

Ping @oryband @KhasMek @apjanke @blueyed @ChrisJefferson

lib/functions.zsh
ncanceill: Wao you're faster than your own shadow! Gimme a few minutes to take your input into account ;-)

apjanke:
This seems like a good idea.

That's a lot of aliases.

I'm not sure what the right thing to do is when it fails to create an alias due to collision. It sounds like the intention of `safe_alias` is for that to be an error which requires resolution by changing the OMZ or user config code, so it should probably emit an error message instead of just silently doing nothing. Maybe there should be a `-q` or a `polite_alias` variant which expects there to be conflicts sometimes and wants to silently defer to the existing definition.

`alias` takes options like `-g` and `-s`. Might want to change `safe_alias` to detect those in its arguments and either balk, or pass them through to `alias`. And some of the existing `alias` commands in OMZ use those switches, so either they shouldn't be auto-replaced, or `safe_alias` would need to support them.

There are some aliases which intentionally clobber existing commands. Would need to avoid auto-replacing them.

```
  alias mv='nocorrect mv'
  alias mysql='nocorrect mysql'
```

There's also the no-`=` calling form which queries current alias definitions and shouldn't be auto-replaced.

```
function alias_value() {
    alias "$1" | sed "s/^$1='\(.*\)'$/\1/"
    test $(alias "$1")
function try_alias_value() {
    alias_value "$1" || echo "$1"
```

Seems a little odd for this function to be unsetting `$ALLOW_OVERRIDE_ALIASES`, effectively making it a single-use variable. Normally scoping would be done by the caller. If they wanted it for just a single call, they could do `ALLOW_OVERRIDE_ALIASES=true safe_alias foo=bar`.

apjanke:
Turning in for the night; I'm on US time here.

One other gotcha: that search also catches the word `alias` when used in a non-command position.
```
    eval $(hub alias -s zsh)
      return POSIX path of (target of window 1 as alias)
      set output to output & ((item item_index of the_selection as alias)'s POSIX path) & the_delimiter
```

apjanke:
Oh: also, some people like to reload their `~/.zshrc` inside a running session, re-sourcing OMZ. There's even a `zsh_reload` plugin to do that. That's going to blow up if `safe_alias` is enabled, since everything will already be defined.

lib/functions.zsh
ncanceill: Would this not make the alias local to the function? Sorry but I do not know this well.

lib/functions.zsh
ncanceill: Granted.

lib/functions.zsh
apjanke: Nope. `emulate -L` localizes shell options, traps, and patterns, but not aliases or functions.

lib/functions.zsh
```diff
@@ -63,6 +63,32 @@ function try_alias_value() {
...
+# Return value:
+#    0 if the alias was safely created, 73 otherwise
+#
+function safe_alias() {
+    local alias lhs rhs
+    alias=(${(s:=:)1})
+    lhs=$alias[1]
+    rhs=$alias[2,-1]
+    if (( $+commands[$lhs] || $+functions[$lhs] || $+aliases[$lhs] )); then
+        if [ "x$ALLOW_OVERRIDE_ALIASES" = "xtrue" ]; then
```
apjanke: `[[ $ALLOW_OVERRIDE_ALIASES = "true" ]]` would be more idiomatic OMZ code style.

ncanceill: Granted. Were you not supposed to be sleeping by now?

lib/functions.zsh
ncanceill: OK then it cannot hurt!

ncanceill:
So valuable input, thanks @apjanke!

I understand that some plugins may be using aliases internally, but I am not sure this is good practice: better use functions instead, should they not? The 73 return code should be silent when the function is used inside a sourced script, if I am not mistaken.

As to global aliases, I think they are out of the scope of this feature. Maybe the "documentation" comment above the function definition should mention that more clearly? And where is this `-s` flag documented?

You are totally right about the `unset` statement, maybe I need some sleep myself. Fine by me anyway, since we want the function to also be callable from the shell session, do we not?

Finally, I totally agree that my search command has some false positives, which is why my sentence below the command does not specify an exact number. Anyhow all replacements will need to be done manually I fear — we cannot afford any mishap just by being sloppy.

apjanke:
You're welcome!

> I understand that some plugins may be using aliases internally, but I am not sure this is good practice: better use functions instead, should they not? 

Yeah, stuff intended for non-interactive or programmatic use should use functions instead of aliases. Zsh itself [recommends it be done that way](http://zsh.sourceforge.net/Doc/Release/Shell-Grammar.html#Aliasing).

Documentation for `alias` and its switches is [here ](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html#index-alias) under "Shell Builtin Commands" in the Zsh Reference Manual. The `-s` switch creates "suffix" aliases. Global aliases being out of scope seems fine. And documenting it would be good, but we should probably parse the inputs and be conservative about what we accept here, erroring out on any unrecognized options or argument forms.

Maybe now would be a good time to consider "debug" output for OMZ. It seems like the most usable behavior for `safe_alias` would be to silently do nothing when asked to redefine an existing command. But if a user was having trouble figuring out why their alias definition wasn't working, they could flip on an `$OMZ_DEBUG` setting, and in that case, `safe_alias` would print a warning when it failed to create an alias. (Or in a really verbose mode, print a log message on both success and failure, which could help the user figure out where the conflicting alias was defined.) Something that was detailed but not as verbose as the `zsh -x` tracing facility.

apjanke:
You know, this brings up a larger question: What should the precedence order for plugins be, in general?

The user specifies a list of plugins like `plugins=( osx themes nyan brew )`. Right now, that's the order they're loaded in. And it effectively defines a precedence order. But it's not like a function search path, where earlier items have higher precedence. Because shell programming is all procedural, later definitions – of variables, functions, shell options, and aliases – overwrite prior definitions and take precedence over them.

The `$fpath` construction order in `oh-my-zsh.sh` matches this, too. 
```
# Add all defined plugins to fpath. This must be done
# before running compinit.
for plugin ($plugins); do
  if is_plugin $ZSH_CUSTOM $plugin; then
    fpath=($ZSH_CUSTOM/plugins/$plugin $fpath)
  elif is_plugin $ZSH $plugin; then
    fpath=($ZSH/plugins/$plugin $fpath)
  fi
done
```
The plugin dirs are *prepended* to the path in load order, effectively reversing the load order list, so plugins later on the list take precedence over earlier ones for autoloaded function definitions (including completions), too. I don't know if that was intentional or not, but it does make things quite consistent.

There are some exceptions to this order. Things that change the context to which later definitions apply – like `vi-mode` changing the default keymap, affecting later `bindkey` calls – seem to go the other way. But it seems mostly consistent.

If we do this `safe_alias` thing, then we're changing the precedence order to have earlier entries in the lugins list take precedence, but only for aliases. Different classes of definitions then have different precedence order. If you have plugins which define functions and aliases that go together, you wouldn't be able to get a consistent definition set using plugin order to resolve conflicts, then.

Maybe this is just a philosophical concern and is less important than letting plugin aliases defer to system-defined commands. But it's something to consider.

robbyrussell:
⌛️ Going to close this out for now

PR #4591, opened 2015-11-07
speed up mercurial plugin

mduszyk:
Mercurial plug-in is slow when working with big repository. Modified it a bit so it is usable again. The changes are:
- changed functions not to call other functions which lead to multiple mercurial command executions
- replaced hg summary with hg branch which is faster
- replaced hg status with hg status -mar


robbyrussell:
⌛️  This has been in needing testers for 5 1/2 years. While I'm inclined to just close it, I'm going to go ahead and approve + merge this as the code looks straightforward enough.

PR #4592, opened 2015-11-07
Fixing strange git unstaged prompt behavior

jstnlef:
I noticed that for certain projects with unstaged changed I would get a 'U' character instead of the '●' character. This should fix that.

Original:
screen shot 2015-11-07 at 4 38 41 pm

With fix:
screen shot 2015-11-07 at 4 39 47 pm

(edit: I also have a custom coloring for virtualenvs active. You can ignore that)

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -36,7 +36,7 @@ CURRENT_BG='NONE'
   local LC_ALL="" LC_CTYPE="en_US.UTF-8"
   # NOTE: This segment separator character is correct.  In 2012, Powerline changed
   # the code points they use for their special characters. This is the new code point.
-  # If this is not working for you, you probably have an old version of the 
+  # If this is not working for you, you probably have an old version of the
   # Powerline-patched fonts installed. Download and install the new version.
   # Do not submit PRs to change this unless you have reviewed the Powerline code point
   # history and have new information.
@@ -118,7 +118,7 @@ prompt_git() {
     zstyle ':vcs_info:*' get-revision true
     zstyle ':vcs_info:*' check-for-changes true
     zstyle ':vcs_info:*' stagedstr '✚'
-    zstyle ':vcs_info:git:*' unstagedstr '●'
+    zstyle ':vcs_info:*' unstagedstr '●'
     zstyle ':vcs_info:*' formats ' %u%c'
     zstyle ':vcs_info:*' actionformats ' %u%c'
     vcs_info


apjanke:
I... *think* this is probably safe and effective?

[Here's](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Version-Control-Information) the reference doco for this feature. Looks like there could be a couple things going on here. Either this is a bridged/adapted repo and the `vcs-string` part is either `git-svn` or `git-p4` instead of plain `git`. Or there are user customization hooks active, in which case you're getting some form of `git+...` in there.

Do you know which of these is the case for you? Do you have any hooks defined? Or are these clones of Perforce or `svn` repos? I'm not sure that we should be setting the style for hook contexts as well. If it's the former case, maybe we should just add entries for the bridged repo types as well, like this. I don't really understand this part of `vcs_info` very well, but that seems more conservative.

```
zstyle ':vcs_info:git-svn:*' unstagedstr '●'
zstyle ':vcs_info:git-p4:*' unstagedstr '●'
```

On the other hand, I don't think we care about displaying different symbols for different VCS system types. I think the fact that Agnoster specifies `git` is just a historical quirk resulting from it originally being written to only support `git`. Maybe we should just style it the same for all VCSes instead, like this.

```
zstyle ':vcs_info:*' unstagedstr '●'
```

Agnoster now has [its own upstream repo](https://github.com/agnoster/agnoster-zsh-theme), if you want to discuss it there, too.

jstnlef:
Ah great. I'll open a PR there and set it to `zstyle ':vcs_info:*' unstagedstr '●'`

jstnlef:
@apjanke Interestingly enough, the upstream is significantly different that the version in oh-my-zsh. Is there going to be a merge to here shortly?

jstnlef:
@apjanke Actually, I didn't see this PR https://github.com/agnoster/agnoster-zsh-theme/pull/5. If that is merged I can make a PR in the upstream.

apjanke:
@jstnlef Yes, some folks are working on unifying upstream agnoster with the OMZ version. (Disclaimer: I am personally a big booster of this idea, which is why I'm pointing folks at upstream. If you're in agreement give me a +1 on that other PR.)

Who knows when it will actually happen though, so let's proceed with this PR for OMZ, and then back-port that change upstream too. :)

jstnlef:
Sounds good to me. I'll make the change here to `zstyle ':vcs_info:*' unstagedstr '●'`

jstnlef:
Change made.

apjanke:
Thanks!

Tested on OS X and looks good; I'm :+1: on this. @robbyrussell / @mcornella, can we merge this soon? It's a low-risk cosmetic change.

PR #4593, opened 2015-11-09
Enable using 'gls' on any OS for ls coloring

rtimmons:
New adopter to omz. I noticed it had aliased `ls` to `ls --color=tty` despite the check to ensure that this doesn't error-out. 

In debugging I found that I had already setup `ls` as an alias `alias ls="gls -F --color"`, so the check passes (since the check uses the aliased `ls`), and then re-aliases `ls` and wipes out the `gls`. Tricky.

At any rate it seems "more correct" to use `gls` if it's available no matter the OS. I moved the OpenBSD check first to keep the same behavior.

I don't have a NetBSD or OpenBSD to test, but I tried with and without `gls` installed on my mac and `ls` is properly aliased and colored.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -6,17 +6,18 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 if [ "$DISABLE_LS_COLORS" != "true" ]
 then
   # Find the option for using colors in ls, depending on the version: Linux or BSD
-  if [[ "$(uname -s)" == "NetBSD" ]]; then
-    # On NetBSD, test if "gls" (GNU ls) is installed (this one supports colors);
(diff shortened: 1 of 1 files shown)


mcornella:
Aliases should be defined after sourcing oh-my-zsh.sh. That's the more direct fix for your problem.

Now, the current logic is a mess. Your problem stems from the fact that the call to `ls` is done without the [`command` specifier](http://zsh.sourceforge.net/Doc/Release/Shell-Grammar.html#index-command) before it, so it uses the `ls` alias that you defined instead of the original ls command. That is the first problem to fix. The second is to use `gls` on OSX as well if the normal `ls` doesn't support coloring.

As is now, I cannot merge this PR. But if you want to apply my suggestions above I'll guide you if you need to. Sorry for the late reply!

robbyrussell:
⌛️ Closing due to how much time has lapses since feedback was provided

PR #4594, opened 2015-11-09
Dynamic generation of mix commands for zsh completion.

jinyeow:
I wrote my own zsh completion file for mix today (didn't realise that oh-my-zsh already has one). I noticed that depending on the project, the mix tasks can change because of dependencies like phoenix or ecto.
E.g. usually you wouldn't have 'mix phoenix.server' except in the project directory for a phoenix web app.
This is my solution for "dynamic" generation of the _1st_arguments:
I replaced most of the _1st_arguments with just 

```
"${(@f)$(mix help | egrep "^mix " | tail -n+2 | sed 's/^mix //;s/\s\+#\s/:/')}"
```

The above line will allow the zsh user to autocomplete all available mix tasks. I just finished this and thought I should submit it for other oh-my-zsh users.

### DIFF
--- a/plugins/mix/_mix
+++ b/plugins/mix/_mix
@@ -5,41 +5,7 @@
 
 local -a _1st_arguments
 _1st_arguments=(
-    'app.start:Start all registered apps'
-    'archive:List all archives'
-    'archive.build:Archive this project into a .ez file'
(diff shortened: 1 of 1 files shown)


mcornella:
Already in `mix-fast`.

PR #4597, opened 2015-11-09
Fix open_command for msys in the git-for-windows SDK

marking:
Added an option for the msys value of $OSTYPE which is set by the git-for-windows SDK
https://github.com/git-for-windows/build-extra/releases/tag/git-sdk-1.0.1

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -23,6 +23,7 @@ function open_command() {
     darwin*)  open_cmd="open" ;;
     cygwin*)  open_cmd="cygstart" ;;
     linux*)   open_cmd="xdg-open" ;;
+    msys*)    open_cmd="start" ;;
     *)        echo "Platform $OSTYPE not supported"
               return 1
               ;;


mcornella:
Can you tell me how to test this? How do you run oh-my-zsh from `Git for Windows`?

marking:
Here are the steps I used:

 - install the 'Git for Windows SDK' from https://github.com/git-for-windows/build-extra/releases/tag/git-sdk-1.0.1 (I believe anything built on MSys that has pacman included will do)
 - run {installDir}\mingw64_shell.bat
 - install zsh using pacman: 'pacman -S zsh'
 - from within bash, start zsh
 - install oh-my-zsh according to the instructions

eventually, I made a copy of mingw64_shell.bat named mingw64_zsh.bat that replaces launching mintty.exe with /bin/sh with /bin/zsh so I wouldn't have zsh running as a child process of bash, but I don't thin that is necessary for this.

mcornella:
Ok, I could try this, thank you for your clear instructions. Unfortunately, this doesn't work well with arguments with spaces in them. Let me investigate further...

mcornella:
Ok, I found the culprit. It seems the `start` command will interpret the first parameter as the title if this parameter contains spaces, so if you try `start "abc def"` it will open a new command prompt with the title being `abc def`. I solved it by adding `""`.

marking:
Oh, that’s right. I completely forgot about that. Thanks for the quick fix.



*From:* Marc Cornellà [mailto:notifications@github.com]
*Sent:* Monday, 30 November, 2015 01:43 PM
*To:* robbyrussell/oh-my-zsh
*Cc:* Mark Ingalls
*Subject:* Re: [oh-my-zsh] Fix open_command for msys in the git-for-windows
SDK (#4597)



Ok, I found the culprit. It seems the start command will interpret the
first parameter as the title if this parameter contains spaces, so if you
try start "abc def" it will open a new command prompt with the title being abc
def. I solved it by adding "".

—
Reply to this email directly or view it on GitHub
<https://github.com/robbyrussell/oh-my-zsh/pull/4597#issuecomment-160755614>
.

mcornella:
No worries, thanks for your contribution!

PR #4602, opened 2015-11-11
add --no-check-certificate option for wget

roramirez:


robbyrussell:
Hi there, I'm going to pass on this because we intentionally removed this back in Februrary after some discussions in #3607. It's not a safe default to instruct everyone to use

PR #4605, opened 2015-11-12
Makes the theme plugin sane

joehakimrahme:
A couple of improvements for the theme plugin
- Custom themes are looked for in `$ZSH_CUSTOM/themes`
- `lstheme` function doesn't cd to the directory anymore

### DIFF
--- a/plugins/themes/themes.plugin.zsh
+++ b/plugins/themes/themes.plugin.zsh
@@ -8,9 +8,9 @@ function theme
 	source "$RANDOM_THEME"
 	echo "[oh-my-zsh] Random theme '$RANDOM_THEME' loaded..."
     else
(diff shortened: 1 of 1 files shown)


apjanke:
Conflicts with #3337, #4471. See also #3743.

This would break the `theme` plugin for users who currently use it with custom themes because of the location change, and is inconsistent with the main `oh-my-zsh.sh`, which looks in both `$ZSH_CUSTOM` and `$ZSH_CUSTOM/themes`. The `themes` plugin should probably look in both places, too.

joehakimrahme:
Thanks for the review. I pushed a new commit taking into account both custom directories.
I haven't really looked into the issues you mention, but if there's any need, I can rebase my branch against their's.

PR #4608, opened 2015-11-14
aws: added an alias for changing the AWS access key set on a profile

cristim:


mcornella:
Shameless clone of #2949

PR #4617, opened 2015-11-16
Removed Similar Aliases

ahmadawais:
Two similar aliases which were causing issues.


ahmadawais:
@robbyrussell Kindly take a look.

mcornella:
:+1:

ahmadawais:
:)

PR #4619, opened 2015-11-16
Mercurial plugin: revise README

### LINKED ISSUE #4616: mercurial plugin readme file should be update
Sorry, I am not good at English.
In the readme file content

> e.g. plugins = (..., mercurial, ...) (that is pretty obvious).

should be update with
> e.g. plugins = (... mercurial ...) (that is pretty obvious).

en ,i means ..it should be replaced the `comma` into `backspace`.
because it will cause misunderstanding to some users

apjanke:
Closes #4616
Changes the `plugins=(... mercurial ...)` to be closer to correct zsh syntax, removing commas and extra spaces.
Revises discussion on configuring themes to avoid sounding like users should modify the theme definition file in the main OMZ folder. Prioritizes the suggestion of using a theme which already supports hg_prompt_info.

### DIFF
--- a/plugins/mercurial/README.md
+++ b/plugins/mercurial/README.md
@@ -2,23 +2,20 @@
 ### Usage
 Update .zshrc:
 
-1. Add name to the list of plugins, e.g. `plugins = (..., mercurial, ...)`
+1. Add name to the list of plugins, e.g. `plugins=(... mercurial ...)`
    (that is pretty obvious).
-2. Change PROMPT variable of current theme to contain current folder mercurial repo info:
+2. Switch to a theme which uses `hg_prompt_info`.
 
-   robbyrussel theme is used by default, so you need to modify PROMPT var
-   from [this file](https://github.com/robbyrussell/oh-my-zsh/blob/master/themes/robbyrussell.zsh-theme)
-   by adding `$(hg_prompt_info)` after `$(git_prompt_info)`, so currently it
-   looks next:
+   Or, customize the `$PROMPT` variable of your current theme to contain current folder mercurial repo info. This can be done by putting a custom version of the theme in `$ZSH_CUSTOM` or by changing `$PROMPT` in `.zshrc` after loading the theme.
 
-   ```diff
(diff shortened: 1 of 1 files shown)


mcornella:
> Default theme PROMPT changed

I submitted PR #4878.

> Also it's actually breaking ability to autoupdate oh-my-zsh

Please explain further what _it's_ means (i.e. why can't you autoupdate)

paskal:
```
➜  ~  upgrade_oh_my_zsh
Updating Oh My Zsh
error: Cannot pull with rebase: You have unstaged changes.
There was an error updating. Try again later?
➜  ~  cd .oh-my-zsh
➜  .oh-my-zsh git:(master) ✗  git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   themes/robbyrussell.zsh-theme

no changes added to commit (use "git add" and/or "git commit -a")
➜  .oh-my-zsh git:(master) ✗  git pull
Already up-to-date.
```

mcornella:
That happens because you made a change to the robbyrussell theme.

Say, do you want to keep the changes or do you not care about them?
—
Marc

paskal:
Readme told me to do so:) Today I found out `themes/robbyrussell.zsh-theme` is changed so I had to do some actions to update to latest version of this file and reapply these changes (adding hg prompt) again.
Problem is by default from time to time `oh-my-zsh` will ask you for upgrade and shouldn't perform it, it will be nice to have some information about that unexpected result of the change in readme, so people doing the change will know they have to make updates manually from now on.

mcornella:
Ok I understand now, the wording in the README is wrong. You don't have to
modify the actual theme file, you should instead:

a) Put the prompt modification at the end of your zshrc file.
b) Copy the theme file to $ZSH_CUSTOM/themes/ (create the folder if it
doesn't exist) and modify that one. Changes in the custom directory won't
be tracked or modified when updating.

I'll update the README to better explain this.
—
Marc

mcornella:
I just added the above changes. Can you comment on #4878?

paskal:
Marc, what information should I add there?

mcornella:
Just tell me if the README is clearer. You can see the [modified README directly here](https://github.com/mcornella/oh-my-zsh/blob/25850ed2509ae2b97c0e297417cac705033e6b89/plugins/mercurial/README.md).

PR #4621, opened 2015-11-17
Use _rake_command instead of bundled_rake when using bundler plugin

sachin21:
# WHAT
Use `_rake_command` instead of `bundled_rake` when using bundler plugin.

# WHY
An error occur if there is a Gemfile.
`bundled_rake` and `rake-fast` aren't compatible. so I'd like to use _rake_command.
However this changes doesn't become a solution to the root of the problem.
but itself (bundled_rake) have not any problem.

## erros logs

```
$ rake [tab]
Generating .rake_tasks...
rake aborted!
Don't know how to build task './bin/--silent'
```

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -12,7 +12,11 @@ _rake_does_task_list_need_generating () {
 }
 
 _rake_generate () {
-  rake --silent --tasks | cut -d " " -f 2 > .rake_tasks
+  if type bundled_rake &> /dev/null && type _rake_command &> /dev/null; then
+    _rake_command --silent --tasks | cut -d " " -f 2 > .rake_tasks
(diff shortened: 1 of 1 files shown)


sachin21:
@robbyrussell @KevinBongart How about this?

KevinBongart:
@sachin21 I'm not sure I understand the issue, could you try to give a little bit more information?

I've used this plugin along with gemfiles and never had that problem, so I'd love to know more about the context in which the error can be replicated.

sachin21:
Oops Okay. I'll write more information later. I'm busy today. Please wait :pray:

sachin21:
ah, Could you tell me your oh-my-zsh's plugins?

sachin21:
@KevinBongart I'm sorry. I hit upon a good idea.
I will create new pull-req for resolving the root of the problem.

PR #4624, opened 2015-11-18
Add alias ghc: Shorthand way to clone repos from GitHub

jamen:
Example:

``` shell
ghc jamen/sandbox
```

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -72,6 +72,7 @@ alias gcam='git commit -a -m'
 alias gcb='git checkout -b'
 alias gcf='git config --list'
 alias gcl='git clone --recursive'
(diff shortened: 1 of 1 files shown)


brookskindle:
[Have you heard about hub?](https://github.com/github/hub) It provides this same functionality with a much greater flexibility. With it you can `git clone jamen/sandbox` or even `git clone sandbox`.

Disclaimer: I'm not associated with oh-my-zsh -- just stumbled upon it. I'm also not associated with hub, either.

jamen:
I have heard of hub (that's actually partially what inspired me to make this), however I don't believe hub is included with oh-my-zsh by default, but it does support it through the git plugins.

I thought it would be nice to have a simple little function by the get-go.

PR #4626, opened 2015-11-19
Add support for git flow avh's `bugfix` branches

EspadaV8:
I'm not sure if I've done this completely correctly since I've just taken parts of the other functions and put them together to get the bugfix support in there. If there are any issues just let me know and I can clean it up.


EspadaV8:
I should say, this is for git-flow-avh 1.9.0 and above - https://github.com/petervanderdoes/gitflow-avh/blob/83d2a072ba3d0dace29b0fcbdfdaee2353810a6f/Changes.mdown#190 - when the bugfix feature as added.

jkaving:
I have just created a new PR (#6498) which replaces this one by using the latest official version of the completion script from https://github.com/petervanderdoes/git-flow-completion

PR #4628, opened 2015-11-19
Default to using terminfo to set the terminal title

moyamo:
Currently, the title is only set on supported terminals (i.e. xterm, urxvt, screen etc.). Using terminfo entries to set the terminal title adds support for many more terminals.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -32,10 +32,10 @@ function title {
         # Try to use terminfo to set the title
         # If the feature is available set title
         if [[ -n "$terminfo[fsl]" ]] && [[ -n "$terminfo[tsl]" ]]; then
-		echoti tsl
-		print -Pn "$1"
-		echoti fsl
-        fi
+	  echoti tsl
+	  print -Pn "$1"
+	  echoti fsl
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @moyamo, I didn't know about `tsl`/`fsl`, thank you for your contribution. We recently changed the `termsupport.zsh` code though, so this needs to be rebased. It should go [inside this section as an `else`](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/termsupport.zsh#L27-L32).

You should use the builtin `echoti` instead of `tput` as well, and test for those capabilities with
```zsh
[[ -n "$terminfo[fsl]" ]] && [[ -n "$terminfo[tsl]" ]]
```

Thanks!

moyamo:
Is this alright?

mcornella:
Almost perfect, there's an indentation mismatch. It should be 2 spaces wide. We're not yet following any sort of project-wide style convention but at least on a file-by-file basis it should be the same.
Thanks!

moyamo:
Fixed

mcornella:
Thanks! :smile:

PR #4630, opened 2015-11-21
Async prompt for the bira theme

yoavweiss:
I've added a new theme based on bira and [these async prompt instructions](http://www.anishathalye.com/2015/02/07/an-asynchronous-shell-prompt/), which adds the git status asynchronously, making it significantly easier to work on huge repos (e.g. Chromium, WebKit).


mcornella:
This has the potential to be a game changer. @apjanke was working on a non-blocking git prompt (#3725). Let me closely review this and hopefully we'll merge this with the current bira theme, no need for a second one. 
Thanks!

yoavweiss:
All credit should go to @anishathalye, that came up with that async method! I just banged on his example until it worked for bira :)

apjanke:
Very interesting! This could work better than the "give up after a timeout" mechanism I was working on, and be applicable to other things as well. (My way isn't non-blocking; it just puts a timeout on how long it'll block. #3914 is the current version.)

We should note that this effectively claims SIGUSR1 for its own use across the whole `zsh` session.

I've got a couple questions on how it works. What does it display while the asynchronous `git_prompt_info` call is still running? E.g. if you do a couple commands in a row in a repo, or `cd` from one repo to another. At the very beginning, the prompt is set to `PROMPT="${PROMPT_START} ${PROMPT_END}"`, then at the completion of the async `git_prompt_info` it's set to `PROMPT="${PROMPT_START} $(cat ${HOME}/.zsh_tmp_prompt) ${PROMPT_END}"` in the `TRAPUSR1` handler. But what about between then and the completion of the next call? It looks like the stale status indicators from the prior prompt's async `git_prompt_info` call is still displayed, so if things have changed between prompts (e.g. there was activity on the repo, or you cd'ed between different repos) then the prompt initially has wrong info, and the user can't visually distinguish stale from current info. Maybe at the beginning of each async call, the `$PROMPT` should be reset with an in-progress indicator replacing the `git_prompt_info` results? (It probably shouldn't be just blank, either, that's the same visually as "not in a repo" or "clean" for many setups.)

The temp file should probably go in `$TMPDIR` (or wherever `mktemp` goes to) and not `$HOME`, since `$HOME` may be shared across machines. And it should have `$$` incorporated in the filename, since you may have multiple `zsh` sessions running simultaneously on a single machine; need to keep their temp files separate.

themes/bira.zsh-theme
```diff
@@ -11,11 +11,18 @@ else
...
+
+PROMPT="${PROMPT_START} ${PROMPT_END}"
 RPS1="${return_code}"
 
 ZSH_THEME_GIT_PROMPT_PREFIX="%{$fg[yellow]%}‹"
 ZSH_THEME_GIT_PROMPT_SUFFIX="› %{$reset_color%}"
+ZSH_THEME_GIT_PROMPT_CLEAN="✔"
+ZSH_THEME_GIT_PROMPT_DIRTY="✗"
+
+source $ZSH/lib/async-git-prompt.zsh
```
apjanke: No need to `source` this. Everything in `$ZSH/lib` is always sourced by default. (And `$ZSH_CUSTOM` may override it.)

lib/async-git-prompt.zsh
```diff
@@ -0,0 +1,33 @@
+ASYNC_PROC=0
```
apjanke: Since this is a long-lived global variable, should have a more distinctive name to indicate its scope. Maybe `OMZ_GIT_ASYNC_PROC`?

lib/async-git-prompt.zsh
```diff
@@ -0,0 +1,33 @@
...
+function precmd() {
```
apjanke: Instead of using plain `precmd` (of which there can only be one), should be `_omz_git_async_precmd` and then added to the hooks list with `precmd_functions+=_omz_git_async_precmd`. That way it will play nicer with other things that have precmd hooks.

lib/async-git-prompt.zsh
```diff
@@ -0,0 +1,33 @@
...
+    function async() {
```
apjanke: The function name should probably also be prefixed, like `_omz_git_prompt_async`, since the function namespace is global and flat.

lib/async-git-prompt.zsh
```diff
@@ -0,0 +1,33 @@
...
+
+        # signal parent
+        kill -s USR1 $$
+    }
+
+    # do not clear RPROMPT, let it persist
+
+    # kill child if necessary
+    if [[ "${ASYNC_PROC}" != 0 ]]; then
+        kill -s HUP $ASYNC_PROC >/dev/null 2>&1 || :
```
apjanke: Can use `&>/dev/null` instead of `>/dev/null 2>&1` if you want to be more concise.

apjanke:
For readability and manageability, it might be nicer to just have a `$GIT_PROMPT_STATUS` variable inside `$PROMPT` (or `$RPROMPT`) and have the async code just update the value of the variable, instead of modifitying `$PROMPT` directly and baking the specific results in to it. That would make the `$PROMPT` string more readable (especially in the output of `set` or `echo $PROMPT`, which might show the state while the async call is still running). And it would make the technique easier to reuse for other prompts. Then you could just put `$GIT_PROMPT_STATUS` in the appropriate spot in your prompt, instead of having to write procedural code to reconstruct the prompt each time.

anishathalye:
Might be worth looking at this (https://github.com/mafredri/zsh-async) as a better way to do async stuff.

apjanke:
Nice find. Yeah, let's have a look at that zsh-async lib.

yoavweiss:
Thanks for the comments. I doubt I'll have the time to address them in the near future, so feel free to take my initial patch and run with it :D

robbyrussell:
⌛️ Closing due to how long this has been open. Happy to revisit if there is someone to help us with a refreshed PR. Sorry for not getting to it back in the day!

PR #4631, opened 2015-11-21
disable composer xdebug warning for autocomplete

koenpunt:
see composer/composer#4622

### DIFF
--- a/plugins/composer/composer.plugin.zsh
+++ b/plugins/composer/composer.plugin.zsh
@@ -7,11 +7,11 @@
 
 # Composer basic command completion
 _composer_get_command_list () {
-    $_comp_command1 --no-ansi | sed "1,/Available commands/d" | awk '/^[ \t]*[a-z]+/ { print $1 }'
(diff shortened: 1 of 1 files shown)


mcornella:
Sorry, I merged #4676 instead. Thanks!

koenpunt:
Oh well, `COMPOSER_DISABLE_XDEBUG_WARN` is probably the recommended way.. Because other errors that may arise are now suppressed as well

mcornella:
That's what I thought, but off the top of my head I only recall having seen errors about composer version being too old, which are not relevant when requesting for completion entries (it will show anyway after running whatever command it autocompleted).

I see [in this page](https://getcomposer.org/doc/articles/troubleshooting.md) that this is not exactly true ―there are other possible errors―, but let's go with the whole stderr approach and see what happens. In case any of those errors is triggered, the completion function should just return nothing, and a subsequent `composer diagnose` would show the whole trouble.

koenpunt:
If for any reason you consider to merge my changes, I've rebased my branch on current master :)

mcornella:
Ok thanks, if anything comes up I'll use this one. Thank *you* :smile:

PR #4635, opened 2015-11-24
support PROMPT_DIRTRIM in agnoster

colemickens:


colemickens:
Yeah, I think we can let this one go...

PR #4637, opened 2015-11-24
Update theme with rbenv support + cleanup

a-b:


mcornella:
Current theme version is better.

PR #4640, opened 2015-11-26
Theme "candy": display user@host in red if user == root

muffl0n:


robbyrussell:
Appreciate the contribution, but think we're going to pass on this particular submission. Thanks, though!

PR #4645, opened 2015-11-28
Created this new Bira based theme but with NVM instead of RVM.  

light24bulbs:
More and more folks are switching to node from rails. Bira is my favorite theme, so I thought I would make it appropriate for node development.  It now shows the node version where bira shows the ruby version.  Called it "nira" theme.  Preview is here: http://imgur.com/TnIzmkD


light24bulbs:
Any thoughts on this?  It is _just_ another theme, but it's a really great one.  And a common use case.

KamranMackey:
@light24bulbs I like it. :+1: from me. :)

mcornella:
[We can't accept more themes at the moment](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page.

If you want, you can modify the original bira theme to add the nvm prompt and I'll gladly review that.

PR #4646, opened 2015-11-29
Added plugin for Visual Studio Code

wyldphyre:
Added a plugin to enable triggering Visual Studio Code from the command line. Nothing fancy.

p.s. Not sure why the changes from Feb 2015 are included here. I thought they'd already been merged. I've rebased from the original, but I'm not that hot with Git, so not sure what's going on there.

### DIFF
--- a/plugins/vscode/README.md
+++ b/plugins/vscode/README.md
@@ -1,0 +1,17 @@
+## Visual Studio Code
+
+Plugin for Visual Studio Code, a text editor for Mac OS X, Windows, and Linux 
+
+### Requirements
+
+ * [Visual Studio Code](https://code.visualstudio.com)
+ 
+### Usage
+
+ * If the `code` command is called without an argument, launch VS Code
+
+ * If `code` is passed a directory, cd to it and open it in VS Code
+
+ * If `code` is passed a file, open it in VS Code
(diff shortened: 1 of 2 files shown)


Kriechi:
Can you please remove the Atom file and squash your commits?
As far as I can tell this only works for OSX at the moment, can you maybe add support for linux?
You can take a look at other plugins (e.g. Atom, Sublime Text, etc.)

wyldphyre:
I don't know how to remove files from the pull request. I wouldn' have included them in the first place if I could have seen a way to avoid it.

I'd rather not add support for stuff I can't test.

wyldphyre:
I'm going to close this pull request and start over with a fresh fork. Will try to support platforms other than OS X when I do.

Kriechi:
@wyldphyre: take a look at this function: 
https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/functions.zsh#L18

And for the file: just remove it, commit your changes and then squash all commits.

wyldphyre:
Repository is gone now. Will do it again at some point.

wyldphyre:
oh, forgot to say, thanks for pointing me to open_command.

PR #4649, opened 2015-11-30
znt: n-history supports multi-line cmds and starts with current buffer

psprint:
Update of ZNT to most recent version. Not much new code, but two useful features – multi-line entries are now supported by n-history, and it can start with what user currently typed into prompt. Also added some coloring to display of history


mcornella:
Can you also rename your README as in https://github.com/robbyrussell/oh-my-zsh/pull/4648?

psprint:
I renamed it after checking that it looks fine. Surprisingly it looks very fine. My original intention was to provide a text file for fine reading on terminal.

psprint:
Could someone quickly merge? Users who download now will have config files that do not include coloring of history entries. Lets minimize amount of people needing update of config files. The patch is well tested by me as I wrote it some time ago

PR #4651, opened 2015-11-30
Add support for "putty" $TERM in termsupport.zsh

ionelmc:


ionelmc:
For some reason I have to use that specific terminal type.

apjanke:
Probably needed because your Putty configuration is set up to report the terminal string as `putty`. (I think `xterm` is the default.) It's under Connection > Data > Terminal Details in the New Session dialog.

Should probably be `[[ "$TERM" == putty* ]]` here instead of just `[[ "$TERM" == putty ]]`, since there are variants like `putty-256color` and `putty-vt100` out there, and all should support the same window title control sequences.

ionelmc:
> Probably needed because your Putty configuration is set up to report the terminal string as putty. (I think xterm is the default.) It's under Connection > Data > Terminal Details in the New Session dialog.

Yes that's intentional, as previously stated. Numpad don't work with anything else.

mcornella:
Thank you @ionelmc

apjanke:
How's numpad misbehaving for you? This could be a variant of #2735/#3061, where `vi-mode` or certain other things were breaking terminfo-based key bindings due to bad keypad mode control.

One difference between the terminfo entries for `xterm` and `putty` (at least on my OS X box) is that the character sequences for the arrow keys match the sequences xterm produces in "keypad local" mode, where the `xterm` includes the sequences for "keypad application" mode, and the `putty` entry does not have an `[sr]mkx` entries for switching keypad modes. This is consistent with the #2735 breakage, which causes the terminal to be stuck in "keypad local" mode.

We recently merged #4191, which fixes `vi-mode` and restores the "keypad application" mode behavior, and you'll probably pick this up in the next update. After the update, you may find that using `xterm` works for you again.

(Even if that's what was breaking, there's still no reason not to add putty support to `termsupport.zsh` like this.)

ionelmc:
This seemed to be the solution that I followed: http://superuser.com/a/104016

I've updated the OMZ repo, still doesn't work with xterm ...

However, if I disable "application keypad mode" it seems to work with xterm.

Can anyone ELI5 this to me? I'm pretty noobish at this term stuff ...

apjanke:
> Can anyone ELI5 this to me? I'm pretty noobish at this term stuff ...

ELI5 might be hard. This terminal configuration stuff is pretty complicated, and the bug was kind of subtle. But basically:

* Different terminals and terminal emulators use different character sequences for different key presses or terminal control commands
 * And some, like Putty, can be configured to further change their behavior with respect to those control sequences
* Some terminals have multiple modes for their keypads, "application" (aka "transmit") and "local" modes, and they send different character sequences in the two modes
* The `terminfo` database has key/character sequence mappings for a bunch of terminals, so you can do terminal programming and key binding portably. The `$TERM` variable selects which entry is being used.

But the `terminfo` entries only have mappings for the keypad application mode sequences. So if you want to do portable terminfo-based bindings, you need to get the terminal in to keypad application mode while the command line editor is active. In zsh, this is done with `zle-line-init` and `zle-line-finish`, by making them emit the control sequences to switch between modes. (The terminfo names for these sequences are `smkx` to get in to application mode, and `rmkx` to go back in to local mode.) Base OMZ does this by default. And it defines key bindings for the arrows and keypad keys using terminfo.

But the `vi-mode` plugin broke our `zle-line-init`/`zle-line-finish` functions so the terminal got left in keypad local mode, so those terminfo-based key bindings didn't work.

So you've got this stack of a few layers that need to be configured to be in sync with each other. And if you Google for solutions to this terminal stuff, you'll find a lot of different ways to adjust the configuration, and some of them will work for your particular situation, but they won't necessarily be portable, or work with other code (like the stuff in OMZ) that's also working with terminal configuration.

Putty supports application/local keypad mode switching, and it's enabled by default. But the `putty` terminfo entry seems to be "lying" about it and saying that it doesn't. So OMZ doesn't attempt to switch in to application mode while editing the command line. And one way or another, you have your key bindings set up to use the local-mode key sequences, and everything lines up. Using the `xterm` terminfo entry but disabling application keypad mode is another way of accomplishing a similar alignment. (The downside of either of these might be that ncurses-based or visual programs like `vi` which do rely on switching to keypad application mode might have some issues.)

But the `xterm` terminfo entry's numpad entries use the application mode sequences. Which suggests that your zsh configuration, maybe in `~/.zshrc` after loading OMZ, or maybe in your OS's default `/etc` stuff, is setting up key bindings using hardcoded key sequences that match the keypad local sequences.

Hope that makes sense.

ionelmc:
> Hope that makes sense.

Thank you. This is very useful.

apjanke:
Great.

If you want, you could run `omz_diagnostic_dump` and post the results as a gist, and we could probably help you track down where those bindings or other terminal config changes are happening, and get the normal `xterm` entry working for you. Sounds like things are mostly working now, but it could be useful in case you want to use other terminals or other ncurses-based programs.

PR #4659, opened 2015-12-02
Update Docker Compose Plugin

RobLoach:
Also rename it to match the other oh-my-zsh plugin formats.


spencerhakim:
The original filename was correct. This isn't so much a plugin as just a completions definition file.

RobLoach:
Should we keep it as `_docker-compose` ?

ysoldak:
Feels like this is a duplicate of #4311 
I use docker-compose every day and can help test this new version of the plugin.
What are the expectations? Any formal checklist?

PR #4660, opened 2015-12-02
Avoid upgrade name conflict with the_silver_search

nvesely:
[`ag`](https://github.com/ggreer/the_silver_searcher) is a pretty popular `grep` alternative. To avoid name conflicts the upgrade alias here should be changed to `au`. I ran `apt-file search -x /bin/au$` and confirmed that Debian, in it's 40,000 or so packages, does not provide an `au` executable. Plus, `au` is better mnemonic for `apt-get`/`aptitude` `upgrade`/`safe-upgrade` than `ag`.

### DIFF
--- a/plugins/debian/debian.plugin.zsh
+++ b/plugins/debian/debian.plugin.zsh
@@ -50,7 +50,7 @@ if [[ $use_sudo -eq 1 ]]; then
     alias adg='sudo $apt_pref update && sudo $apt_pref $apt_upgr'
     alias adu='sudo $apt_pref update && sudo $apt_pref dist-upgrade'
     alias afu='sudo apt-file update'
-    alias ag='sudo $apt_pref $apt_upgr'
+    alias au='sudo $apt_pref $apt_upgr'
     alias ai='sudo $apt_pref install'
     # Install all packages given on the command line while using only the first word of each line:
     # acs ... | ail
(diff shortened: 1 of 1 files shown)


apjanke:
Conflicts with #3866, which takes a different approach to the same problem (noted in #3633).

fowlslegs:
Hmmm. On one hand my pull allows people to have both `the_silver_searcher` and (a more mnemonic) shortcut for `apt-get`/`aptitude` `upgrade`/`safe-upgrade`. On the other hand, a lot of people are probably used to typing `ag` to upgrade and #3866 will not confuse them on next their next update.

Perhaps a way to have the best of both would be to alias "upgrade" to `au` if `the_silver_searcher` is installed and otherwise keep the current `ag` alias. This way we can resolve this name conflict without breaking anyone's work-flow. Thoughts?

apjanke:
I'm not a big user of terse aliases, so I'll defer to heavier users on this. But switching between `ag`/`au` depending on what else is installed seems to be a not-great solution. The name of the command would vary between different systems, and even the same system over time, depending on what's installed. And if you had another plugin that wanted to use some of those names and did similar name-switching, then you'd have more conflicts and behavior that was dependent on plugin loading order. Complicated. Makes it harder for people to share & reproduce commands, write scripts, diagnose bug reports etc., which is already complicated when you have lots of aliasing going on. I'd think either switching to `au` or just not doing the alias if another `ag` exists would be preferable to the name-changing approach. And then let users manually resolve it if there's a conflict on their systems and they don't like how it works out.

Personally, I prefer this PR's approach of choosing a different name. But I'm not a stakeholder here; I just wanted to point out the conflict so more informed folks can resolve it.

This sort of collision seems to be an inherent risk when using these really short aliases. There's only so many 2-to-4 letter names to go around.

mcornella:
I've gone ahead and applied these changes in #7698: it's one simple change and one that users will surely will notice since there is a README documenting it.

Cheers

PR #4662, opened 2015-12-03
add alias plugin: cheatsheat your alias.

hqingyi:
with lots of 3rd-party amazing aliases installed, just need something to explore it quickly.

 **acs: alias cheatsheet**
1. `acs`: group all alias
2. `acs $keyword`quickly filter alias & highlight

### DIFF
--- a/plugins/aliases/aliases.plugin.zsh
+++ b/plugins/aliases/aliases.plugin.zsh
@@ -1,0 +1,13 @@
+# with lots of 3rd-party amazing aliases installed, just need something to get it quickly.
+# 
+# use alias as prefix:
+# - aliass: search alias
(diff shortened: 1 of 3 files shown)


HQingYi:
image

like this

ensonik:
I like this; I've been slowly learning the aliases with "alias-tips", but this would be a great addition.

robbyrussell (approved):
Did some testing today and it appears to be working nicely. Given how long this has been lingering, I'm going to allow this to go out!

juanitoddd:
How can I "group" my alias so when I run `acs` it shows:
`[custom_group]`
       `alias foo=bar`

PR #4663, opened 2015-12-03
znt: faster startup of tools using colorifying (e.g. n-history), optimized heap usage for older Zsh's

psprint:
Colorifying code has been reorganized and allows e.g. n-history to start even more instantly. Updated README.md

Also, Zsh's starting from 5.0.8 and before 5.2 have a specific heap handling that makes ZNT slower. Introduced are workarounds for this making ZNT much faster with large inputs


psprint:
Could someone merge? Tested this for 9 days in upstream ZNTools

psprint:
Thanks for the merge!

PR #4667, opened 2015-12-04
"fishy" theme: Shorten path .foo to .f, not .

dbohdan:
When the current path is `/home/user/.config/doublecmd` the prompt now reads `/h/u/.c/doublecmd`, not `/h/u/./doublecmd` as was the case. This matches what the Fish shell does.

Enclose the Perl snippet in single quotes instead of double quotes.

### DIFF
--- a/themes/fishy.zsh-theme
+++ b/themes/fishy.zsh-theme
@@ -5,7 +5,7 @@ _fishy_collapsed_wd() {
    BEGIN {
       binmode STDIN,  ":encoding(UTF-8)";
       binmode STDOUT, ":encoding(UTF-8)";
-   }; s|^$HOME|~|g; s|/([^/.])[^/]*(?=/)|/$1|g; s|/\.([^/])[^/]*(?=/)|/.$1|g;
+   }; s|^$ENV{HOME}|~|g; s|/([^/.])[^/]*(?=/)|/$1|g; s|/\.([^/])[^/]*(?=/)|/.$1|g
 ')
 }
 
--- a/themes/fishy.zsh-theme
+++ b/themes/fishy.zsh-theme
@@ -1,13 +1,13 @@
 # ZSH Theme emulating the Fish shell's default prompt.
 
 _fishy_collapsed_wd() {
-  echo $(pwd | perl -pe "
+  echo $(pwd | perl -pe '
    BEGIN {
-      binmode STDIN,  ':encoding(UTF-8)';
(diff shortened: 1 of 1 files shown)


themes/fishy.zsh-theme
```diff
@@ -1,13 +1,13 @@
...
+  echo $(pwd | perl -pe '
    BEGIN {
-      binmode STDIN,  ':encoding(UTF-8)';
-      binmode STDOUT, ':encoding(UTF-8)';
-   }; s|^$HOME|~|g; s|/([^/])[^/]*(?=/)|/\$1|g
-")
-} 
+      binmode STDIN,  ":encoding(UTF-8)";
+      binmode STDOUT, ":encoding(UTF-8)";
+   }; s|^$HOME|~|g; s|/([^/.])[^/]*(?=/)|/$1|g; s|/\.([^/])[^/]*(?=/)|/.$1|g;
```
mcornella: Have you tested that this works? Inside single quotes `$var` statements are not translated to its value, so `$HOME` and `$1` will be literal here. I think you can make this change without changing quotes, can you?

dbohdan: You're right: this is broken! Sorry, and thanks for catching it. I corrected this error in my local copy but failed to commit and push it. I'll do that in a moment.

The reason I changed the quotes is that interpolating shell variables inside a Perl snippet is usually considered a bad practice. Shell and Perl both use `$varname`, though each in a somewhat different way, so it can quickly get confusing. For example, here the `$1` is *not* supposed to be interpolated by the shell (notice the slash escaping the dollar sign before it in the double-quotes version) but `$HOME` *is.* For this reason I think a better solution is to access the value of the environment variable `HOME` through Perl's built in `%ENV` hash.

dbohdan:
I've pushed a fix for the `$HOME` problem. See my comment on the outdated diff above for details.

themes/fishy.zsh-theme
mcornella: > For example, here the $1 is not supposed to be interpolated by the shell (notice the slash escaping the dollar sign before it in the double-quotes version)

Oh right, I hadn't noticed that (you're referring to [this line](https://github.com/robbyrussell/oh-my-zsh/pull/4667#discussion-diff-47481372L8), of course). I agree with your solution then.

mcornella:
:+1:

PR #4672, opened 2015-12-05
modify Plugin:themes, better support for custom themes

SenZhangAI:
These feature was added:

### 1. search theme-files in `$ZSH_CUSTOM/themes`
The original source only search `$ZSH_CUSTOM`，but as the [oh-my-zsh-wiki-Customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization) said, the custom theme-files may better put into '$ZSH_CUSTOM/themes'.
so, '$ZSH_CUSTOM/themes' need to be searched.

### 2. lstheme for custom
The original source only list the default themes,
Now, the custom themes were added to be list behind the default themes

### DIFF
--- a/plugins/themes/themes.plugin.zsh
+++ b/plugins/themes/themes.plugin.zsh
@@ -1,24 +1,35 @@
 function theme
 {
     if [ -z "$1" ] || [ "$1" = "random" ]; then
-	themes=($ZSH/themes/*zsh-theme)
-	N=${#themes[@]}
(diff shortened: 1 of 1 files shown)


SenZhangAI:
as the file `oh-my-zsh.sh`, `$ZSH_CUSTOM`'s themes are sourced first, then `$ZSH_CUSTOM/themes`
so I need to change the priority to consider  `$ZSH_CUSTOM` first
I will close this pull request,and open the new one.

PR #4673, opened 2015-12-05
adding gchid, a useful shorthand to change user.name and user.email

dottorblaster:
I did this because I have this issue with multiple git identities on many projects between work, personal life and side projects so I needed this shorthand. I think this would be useful to someone else. So.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -224,3 +224,9 @@ alias gvt='git verify-tag'
 
 alias gwch='git whatchanged -p --abbrev-commit --pretty=medium'
 alias gwip='git add -A; git rm $(git ls-files --deleted) 2> /dev/null; git commit -m "--wip--"'
+
+gchid() {
+  [[ -n $1 ]] && {git config user.name $1}
+  [[ -n $2 ]] && {git config user.email $2}
+}
+


mcornella:
I think this is a good start, but not ready enough for widespread use as an identity management system. It's perfectly fine for personal use though, but I'd like a more automated process before considering merging this.

I'm talking about some sort of git profile management system where you specify which `user.name` and `user.email` corresponds to a given profile (like `work` or `personal` or whatever), and then call a function to set that profile (e.g. `gchid work`).

Another option would be to wrap the `git commit` command or alias such that you can easily use it with the [`--author` argument](https://git-scm.com/docs/git-commit) set to whatever you like:

> --author=<author>
Override the commit author. Specify an explicit author using the standard `A U Thor <author@example.com>` format. Otherwise `<author>` is assumed to be a pattern and is used to search for an existing commit by that author (i.e. `rev-list --all -i --author=<author>`); the commit author is then copied from the first such commit found.

Some other methods could be some sort of [git-hook system like in this post](https://orrsella.com/2013/08/10/git-using-different-user-emails-for-different-repositories/), or setting up different `.gitconfig` files and managing those. 

I think there is some hidden potential that we can tap into for a more straightforward way to switch identities, which possibly would become an entirely new plugin. I understand if you don't want to take on that much work, but as is now I'm not very inclined to pull this in. Thanks!

dottorblaster:
@mcornella this is how a PR should be commented, thanks :+1: 

I totally agree. To be honest, I was already thinking about expanding this with a complete identity management framework; I'm writing a bit of this in these days, let's see if I can put together some proper code :D

PR #4675, opened 2015-12-06
modify Plugin:themes, better support for custom themes

SenZhangAI:
### Feature

These feature was added:

#### 1. search theme-files in `$ZSH_CUSTOM/themes`
The original source only search `$ZSH_CUSTOM`，but as the [oh-my-zsh-wiki-Customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization) said, the custom theme-files may better put into '$ZSH_CUSTOM/themes'.
so, '$ZSH_CUSTOM/themes' need to be searched.

#### 2. lstheme for custom
The original source only list the default themes,
Now, the custom themes were added to be list behind the default themes

### Issue
if the custom theme has the same name with the default one, `lstheme` will list it again

### DIFF
--- a/plugins/themes/themes.plugin.zsh
+++ b/plugins/themes/themes.plugin.zsh
@@ -1,24 +1,35 @@
 function theme
 {
-    if [ -z "$1" ] || [ "$1" = "random" ]; then
-	themes=($ZSH/themes/*zsh-theme)
-	N=${#themes[@]}
-	((N=(RANDOM%N)+1))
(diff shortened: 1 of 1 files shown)


apjanke:
There's an existing PR #3743 which covers this functionality.

SenZhangAI:
Thanks @apjanke .
I see your code. That's great ：）

apjanke:
Thank you!

PR #4676, opened 2015-12-06
[Composer] redirect stderr to /dev/null for completion commands

bamarni:
Composer might display warning messages to stderr (xdebug enabled, version older than 30 day, etc). In those cases it can become user-unfriendly :)

screen shot 2015-12-06 at 13 54 43


mcornella:
:+1:

PR #4678, opened 2015-12-07
Add READNULLCMD command to template

davydovanton:
Hello!
I think `READNULLCMD` command is very useful and therefore I'll be happy to see this command in template file. What do you think?


apjanke:
The template is the wrong place for this; that's just for creating a basic user `.zshrc` and exposing OMZ configuration options. If this is going to be added, it should go somewhere in the `lib/*.zsh` files.

davydovanton:
@apjanke I think you're right, thanks!

PR #4686, opened 2015-12-10
Adds node version to prompt

EvanLovely:


EvanLovely:
2015-12-10 at 11 17 am

PR #4687, opened 2015-12-10
Add Ecosia to web-search plugin.

gmccue:
Adds the [Ecosia search engine](https://www.ecosia.org) to the web-search plugin.


mcornella:
love-tree

:heart_eyes:

Issue #4441, opened 2015-10-02
taskwarrior plugin causes segmentation fault


Shir0kamii:
Hi,

When I try to use autocompletion from the taskwarrior plugin, a segmentation fault happens.

I saw #3796 and #4237 but it doesn't change anything, it still crash with the new update.

PS: thanks for the hard work

apjanke:
Have you looked at #2044? Seems like there might be a version-dependent bug on some systems. What version of zsh are you running, and what operating system?

Shir0kamii:
I just took a look. I no longer have segmentation fault with the 5.1.1 version of zsh.

Now, when I type `task [tab]`, I have the following error message : `_arguments:450: _task_default: function definition file not found`

mcornella:
Please try deleting the cached completion files with `rm ~/.zcompdump*` and restarting the terminal.

Shir0kamii:
I did it and strangely, the sementation fault is back.

apjanke:
I can't reproduce on OS X 10.9 or 10.11.

Could you post the output of `omz_diagnostic_dump` as a gist so we can get some debugging info?

Shir0kamii:
See : https://gist.github.com/Shir0kamii/64b67c4e1594a6fd0be5

apjanke:
Thanks.

I've got an Ubuntu VM around here somewhere. I'll try to reproduce it there.

Shir0kamii:
Thanks for your work, it's a bit annoying to not have autocompletion with a verbose tool like this one.

Let me know if there's anything I can try to solve the problem !

Shir0kamii:
Hi !

Any news ?

apjanke:
Sorry, been busy with work and other projects; haven't had a chance to test this yet.

mcornella:
This was also reproduced by @xoyk at https://github.com/robbyrussell/oh-my-zsh/issues/3796#issuecomment-134982023. 
I could not reproduce that in my current system, I'm running task 2.4.4. Which version are you running (`task --version`)?

Shir0kamii:
My version was 2.5.0-beta3. I upgraded it to 2.5.1 and the autocompletion is working fine. Thank you and sorry for the bother.

Shir0kamii:
[...]

PR #4689, opened 2015-12-11
common-aliases: handle "dev" versions in version check

apjanke:
Uses `is-at-least` instead of a numeric comparison hack, so versions with non-numeric bits like "4.5.0-dev5" don't throw errors.

Fixes an issue noted in https://github.com/robbyrussell/oh-my-zsh/issues/4441#issuecomment-163790819.

### DIFF
--- a/plugins/common-aliases/common-aliases.plugin.zsh
+++ b/plugins/common-aliases/common-aliases.plugin.zsh
@@ -52,7 +52,7 @@ alias mv='mv -i'
 
 # zsh is able to auto-do some kungfoo
 # depends on the SUFFIX :)
-if [ ${ZSH_VERSION//\./} -ge 420 ]; then
+if is-at-least 4.2.0; then
   # open browser on urls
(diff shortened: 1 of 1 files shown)


Shir0kamii:
Hi,

This fix doesn't work for me. I now have this error : 

```
/home/shir0kamii/.oh-my-zsh/plugins/common-aliases/common-aliases.plugin.zsh:55: parse error: condition expected: is-at-least
```

For information, my version of zsh is `zsh 5.2-dev-0 (x86_64-unknown-linux-gnu)`

apjanke:
Oops. Fixed. Try again.

Shir0kamii:
It works, thank you.

mcornella:
For future reference, the `is-at-least` function [is loaded in `lib/misc.zsh`](https://github.com/robbyrussell/oh-my-zsh/blob/56345c26fbabd2ca01c67b258139153a1368ef32/lib/misc.zsh#L3). :+1:

PR #4690, opened 2015-12-11
Update link to git completion

kaelig:
Looks like the original link wasn't working, and this should be the right one.


mcornella:
Hi @kaelig, thanks for your contribution. Let's go one step further and use the github mirror though, I don't fully trust sourceforge at this point. Here's the link: https://raw.githubusercontent.com/zsh-users/zsh/master/Completion/Unix/Command/_git

kaelig:
No problem @mcornella, I've updated the contribution.

PR #4692, opened 2015-12-13
Added stackoverflow !bang

ludo237:
I think it's a nice addition to have a default `!stack ` bang


mcornella:
This feature is already present in the [`frontend-search` plugin](https://github.com/ludo237/oh-my-zsh/tree/master/plugins/frontend-search). 

Currently there is this division of “programming”-related stuff and *more general* stuff between the two plugins. This division may be a little blurry for the moment but we're keeping it that way for the moment.

You can still add that alias to your zshrc file if you don't want to load the whole `frontend-search` plugin though. Thanks

Issue #2750, opened 2014-04-25
Alt-Left-Arrow and Alt-Right-Arrow broken on urxvt in openSUSE


michaeltraxler:
since a recent "git pull" the always working bound-keys
```
^[^[[D (Alt-Cursor-Left)
^[^[[C (Alt-Cursor-Right)
```
are not resulting anymore in a backward-word (or forward-word), even though the correct bind keys are set.
```
"^[^[[D" backward-word
"^[B" backward-word
"^[[1;5D" backward-word
"^[b" backward-word
```
Alt-B and Alt-F are still working.
I'm puzzled.
Thanks for any help as these keys are really in my muscle-memory and I can't live without them...

apjanke:
Probably caused by #1355, which switched to using terminfo for key bindings.

Yeah, if that's absent, I think it might mess up some of your key bindings, because zsh might not be set up to switch in to the right terminal mode when it starts editing a line.

Is this still happening for you? If so, what platform/OS, terminal program, and `$TERM` value are you using?

michaeltraxler:
Hello,

yes, the problem is still happening on my systems (not all of them).

I use OpenSuSE13.2, urxvt and the TERM value is TERM=rxvt-unicode.
With xterm it works.
xterm sends "^[[1;3D" instead of ^[^[[D

On my notebook I have the same OS, same version of urxvt, same .oh-my-zsh
config and it is working here.

If I use ssh from the notebook (working ALT-Left) to the non working system,
it is still not working.
If I use the non working via vncviewer: still not working.

I don't see the difference....

Michael

apjanke:
Okay. I can reproduce this. when I run zsh in urxvt on OpenSuSE 13.2, normally, alt-left and alt-right do forward and backward word. When I have OMZ loaded, though, they insert "D" and "C" characters. If I comment out the `zle-line-init` function like you do, they start working again.

I think the problem might be that the `terminfo` entry for `rxvt-unicode` that ships with OpenSuSE 13.2 is incorrect. It has the wrong sequences defined for the left and right arrows, and for `smkx` and `rmkx`. That means the `echoti` commands in those `zle-line-init` and `zle-line-end` functions are outputting the wrong characters, which I think is messing up the interpretation of your arrow keystrokes. (To see this, compare the output of `infocmp -1 rxvt-unicode | grep '[rs]mkx\|kcu[bf]1` with the codes documented in `man 7 urxvt`.)

[...]

PR #4693, opened 2015-12-14
Bind also UP/DOWN arrow keys in history-substring-search

### LINKED ISSUE #2735: history-substring-search no longer functions after latest update

mcornella:
#### NOT READY

Without it only keypad arrow keys are working. See https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-160455066

Fixes #2735 

/cc @Konfekt

### DIFF
--- a/plugins/history-substring-search/history-substring-search.plugin.zsh
+++ b/plugins/history-substring-search/history-substring-search.plugin.zsh
@@ -19,8 +19,15 @@ if [[ -n "$terminfo[kcuu1]" ]]; then
   bindkey -M emacs "$terminfo[kcuu1]" history-substring-search-up
   bindkey -M viins "$terminfo[kcuu1]" history-substring-search-up
 fi
+if [[ -n "$terminfo[cuu1]" ]]; then
+  bindkey -M emacs "$terminfo[cuu1]" history-substring-search-up
+  bindkey -M viins "$terminfo[cuu1]" history-substring-search-up
+fi
 if [[ -n "$terminfo[kcud1]" ]]; then
   bindkey -M emacs "$terminfo[kcud1]" history-substring-search-down
   bindkey -M viins "$terminfo[kcud1]" history-substring-search-down
 fi
-
+if [[ -n "$terminfo[cud1]" ]]; then
+  bindkey -M emacs "$terminfo[cud1]" history-substring-search-down
+  bindkey -M viins "$terminfo[cud1]" history-substring-search-down
+fi


mcornella:
Updated with fix for copy-paste error

sunaku:
Are you sure about this? :confused: Does this really work? :cold_sweat:

I ask because [terminfo does not define](http://invisible-island.net/ncurses/man/terminfo.5.html#h3-Keypad-and-Function-Keys) any `cuu1` or `cud1` codes (without "k" prefix):

> The codes sent by the left arrow, right arrow,  up  arrow,
> down  arrow,  and  home keys can be given as kcub1, kcuf1,
> kcuu1, kcud1, and khome respectively.

Konfekt:
I am not sure, but over here the k versions had no effect.

mcornella:
`cuu1` and `cud1` *do* exist, but they are used to [move the cursor up or down respectively](http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html#tag_002_001_003_003):

> Variable         | Cap-name | Termcap | Description
> -------------------|---------------|--------------|-------------------------
> cursor_up       | **cuu1** | up | Upline (cursor up)
> cursor_down  | **cud1** | do | Down one line

For some reason your terminal only sends those when pressing `↑` or `↓`. What `$TERM` and terminal emulator do you have, @Konfekt?

Konfekt:
It is `rxvt-unicode`.

sunaku:
@mcornella Thanks for clarifying! :sweat_smile: And I agree that those really shouldn't be sent by arrow keys. :disappointed:

mcornella:
> It is `$TERM = rxvt-unicode` under `urxvt`.

Of course it is... :disappointed: There are [certain inconsistencies in terminfo files shipped with `urxvt`](https://github.com/robbyrussell/oh-my-zsh/issues/2750#issuecomment-116312650) because `ncurses` and `urxvt` maintainers don't get along...

Anyway, that should make it work for you and [the](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-41184614) [rest](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-41833244) [of](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-41833690) [the](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-42705323) [users](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-104939229). But I'd like to wait a bit before I read [this comment first](https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-119016010).

Thanks!

apjanke:
> >    It is $TERM = rxvt-unicode under urxvt.

> Of course it is... :disappointed: There are certain inconsistencies in terminfo files shipped with urxvt because ncurses and urxvt maintainers don't get along...

Is this on OpenSUSE? That's the only place I've seen where the `rxvt-unicode` entry is messed up. OS X and most other Linux distros seem to get it right.

See also https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-156856554 and https://github.com/robbyrussell/oh-my-zsh/issues/2735#issuecomment-164544003. In this particular case, @Konfekt might be getting that arrow key behavior because of a `zle-line-init` customization that clobbers OMZ's `smkx`/`rmkx` stuff.

Konfekt:
Bingo, it is OpenSUSE 42.1

mcornella:
Ok, let's see if we can come up with a better solution then, there's no rush to getting this out because everyone is already using custom key bindings. Thank you @apjanke

apjanke:
> Bingo, it is OpenSUSE 42.1

I bet you're getting hit by #2750, then. Check it out for more details.

apjanke:
I did some more reading about the OpenSUSE and `urxvt` stuff. It looks like their entry is internally consistent, but is different from entries on OS X and some other systems. The OpenSUSE one enables application mode switching for the cursor keypad with `smkx`; the OS X one only enables switching for the numeric keypad.

This looks like it's probably another `zle-line-init` issue: the terminfo-based key bindings are defined expecting application mode cursor keys, but if `smkx` isn't getting issued, they'll end up being wrong.

@Konfekt, what do you get for `which zle-line-init`?

Konfekt:
It is
```sh
zle-line-init () {
	echo -ne "\033]12;Black\007"
	echo -ne "\033[4 q"
}
```

Konfekt:
In the end, this sequence in `.zshrc` is the culprit:
```sh
# Modal cursor color for vi's insert/normal modes.
# http://stackoverflow.com/questions/30985436/
# https://bbs.archlinux.org/viewtopic.php?id=95078
# http://unix.stackexchange.com/questions/115009/
zle-line-init () {
  # zle -K viins
  echo -ne "\033]12;Black\007"
  echo -ne "\033[4 q"
}
# zle -N zle-line-init
zle-keymap-select () {
  if [[ $KEYMAP == vicmd ]]; then
    if [[ -z $TMUX ]]; then
      printf "\033]12;Black\007"
      printf "\033[2 q"
    else
      printf "\033Ptmux;\033\033]12;red\007\033\\"
      printf "\033Ptmux;\033\033[2 q\033\\"
    fi
  else
    if [[ -z $TMUX ]]; then
      printf "\033]12;Black\007"
      printf "\033[4 q"
    else
      printf "\033Ptmux;\033\033]12;grey\007\033\\"
      printf "\033Ptmux;\033\033[4 q\033\\"
    fi
  fi
}
zle -N zle-keymap-select
```
Commenting it out makes the arrow keys work. No need for
```sh
bindkey "\e$terminfo[kcub1]" backward-word
bindkey "\e$terminfo[kcuf1]" forward-word
```
as proposed in https://github.com/robbyrussell/oh-my-zsh/issues/2750#issuecomment-164678768

Konfekt:
Well, indeed commenting out only
```sh
zle-line-init () {
  # zle -K viins
  echo -ne "\033]12;Black\007"
  echo -ne "\033[4 q"
}
```
suffices (showing the wrong cursor vi-mode indicator until entering vim).

apjanke:
Yeah, that's clobbering the [`zle-line-init` that OMZ itself sets up in `lib/key-bindings.zsh`](https://github.com/robbyrussell/oh-my-zsh/blob/6ad6d114dc2b224edfad53177beafe7c4d344bb3/lib/key-bindings.zsh#L7-L16), changing the keypad mode that ZLE is in while editing the command line. Terminfo-based key bindings for function keys (including arrows) require the terminal to be in application keypad mode if `smkx` is defined in the terminfo entry.

If you want to customize `zle-line-init`, you need to incorporate whatever OMZ is doing with it, too. It's unfortunate; having the ability to add multiple hooks like you can with `$precmd_functions` and the like would be nicer.

mcornella:
Closing this as it seems that it's not necessary.

Konfekt:
Ok,  this means the issue of a customizable zle line init?

apjanke:
No, this was specifically about binding `cuu1` and `cud1` in addition to `kcuu1` and `kcud1`, which was a bit of a red herring. The `smkx` stuff is covered in other issues.

If you'd specifically like a nicer mechanism for customizing `zle-line-init`, please open a new issue on that specifically so we can discuss it without the noise of this `cuu1` vs `kcuu1` topic.

PR #4696, opened 2015-12-15
Fix LC_CTYPE

vincentbernat:
See #2012 for some context. Many unicode-related tasks need that
coaelescing to handle unicode correctly. If for some reason, the user is
using a not using a non-unicode locale, we want to ensure that at least
LC_CTYPE is set to an unicode locale.

It's difficult to know if a locale is an unicode one, so we assume that
we want UTF-8 as an unicode codec. Aliases are not handled, so LC_CTYPE
may be redefined even if it was already unicode aware.

The first available UTF-8 locale is used because there is no guarantee
that C.UTF-8 or en_US.UTF-8 are available. For LC_CTYPE, there is no
difference between those locales as coaelescing is not dependant on the
language.

LC_CTYPE is exported as oh-my-zsh may use some shell tools at some
point (sed, awk, ...).

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -34,9 +34,20 @@ else
 fi
 
 # only define LC_CTYPE if undefined
-if [[ -z "$LC_CTYPE" && -z "$LC_ALL" ]]; then
-	export LC_CTYPE=${LANG%%:*} # pick the first entry from LANG
-fi
+LC_CTYPE=${LC_CTYPE:-${LC_ALL:-${LANG:-C}}}
+case $LC_CTYPE in
+   *.utf8|*.UTF-8)
+    # All is correct
+    ;;
+   *)
+    # Need to select an UTF-8 locale
+    local -a available
+    available=("${(f)$(locale -a 2> /dev/null)}")
+    export LC_CTYPE=${${${${(@M)available:#*.UTF-8}[1]}:-${${(@M)available:#*.utf8}[1]}}:-C.UTF-8}
+    unset available
+   ;;
+esac
+export LC_CTYPE
 
 # recognize comments
 setopt interactivecomments
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -32,20 +32,17 @@ CURRENT_BG='NONE'
 
 # Special Powerline characters
 
-() {
-  local LC_ALL="" LC_CTYPE="en_US.UTF-8"
-  # NOTE: This segment separator character is correct.  In 2012, Powerline changed
-  # the code points they use for their special characters. This is the new code point.
-  # If this is not working for you, you probably have an old version of the
-  # Powerline-patched fonts installed. Download and install the new version.
-  # Do not submit PRs to change this unless you have reviewed the Powerline code point
-  # history and have new information.
-  # This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
-  # what font the user is viewing this source code in. Do not replace the
-  # escape sequence with a single literal character.
-  # Do not change this! Do not make it '\u2b80'; that is the old, wrong code point.
-  SEGMENT_SEPARATOR=$'\ue0b0'
-}
+# NOTE: This segment separator character is correct.  In 2012, Powerline changed
+# the code points they use for their special characters. This is the new code point.
+# If this is not working for you, you probably have an old version of the
+# Powerline-patched fonts installed. Download and install the new version.
+# Do not submit PRs to change this unless you have reviewed the Powerline code point
+# history and have new information.
+# This is defined using a Unicode escape sequence so it is unambiguously readable, regardless of
+# what font the user is viewing this source code in. Do not replace the
+# escape sequence with a single literal character.
+# Do not change this! Do not make it '\u2b80'; that is the old, wrong code point.
+SEGMENT_SEPARATOR=$'\ue0b0'
 
 # Begin a segment
 # Takes two arguments, background and foreground. Both can be omitted,


apjanke:
I can't quite follow the the logic in this PR's code or the discussion in the linked issue. But I'm leery of it. Sorry.

LANG and LC_* are control points for the user to set their locale or override parts of it, and downstream programs should respect that. Yeah, some users misconfigure them, but I don't think OMZ has enough information available to "fix" them if they're in a weird state.

OMZ has a large install base across a bunch of different systems, and some of them may really be on systems that really are not UTF-8, especially in Asia. We shouldn't assume it's user misconfiguration. And if it is, maybe we should help them fix their config, instead of hacking around it. Otherwise their stuff will only work inside an OMZ environment, and I don't think that's our goal. Plus, then OMZ really won't work in an unusual environment.

I don't think OMZ should be modifying `$LANG` or `$LC*` at all, and [where it is](https://github.com/robbyrussell/oh-my-zsh/blob/5adb06df0efa1ec251d8ef2e6cae2f1ed3e4107d/lib/misc.zsh#L36-L39), it should give a more detailed reason for doing so. I suspect that's an old system-specific hack. Maybe we should just remove that.

And IMHO, when discussing changing well-known values or interface points, we should include references/links to authoritative sources or standards, to ground the discussion.


#### Agnoster theme side note

The `agnosther.zsh-theme` file should be left alone. That locale setting is there so that this code will work when run under any locale. The `en_US` is chosen just because it's very widely deployed. The intent is for this code snippet to work anywhere, and to work toward harmonizing it with [the upstream repo](https://github.com/agnoster/agnoster-zsh-theme), not just within OMZ.

Regardless of the effective locale, to keep portability, you do need to keep this code block in a local function and set local `LC*`, because the `SEGMENT_SEPARATOR=$'\ue0b0'` is encoding-dependent.

mcornella:
The intent of [this configuration](https://github.com/robbyrussell/oh-my-zsh/blob/5adb06df0efa1ec251d8ef2e6cae2f1ed3e4107d/lib/misc.zsh#L36-L39) was always to enable UTF-8 symbols across all the codebase, so that any plugin or theme using OMZ can safely use them without caring about it. That's the reason behind taking it out of the agnoster theme: a single point of configuration and that's all.

> I don't think OMZ should be modifying $LANG or $LC* at all, and where it is, it should give a more detailed reason for doing so. I suspect that's an old system-specific hack. Maybe we should just remove that.

This only modifies `LC_CTYPE`, and the context of this is at https://github.com/robbyrussell/oh-my-zsh/pull/2012#issuecomment-117641650. It's a little hard to find but following the trail we go to #4091 and then #4065 which has a `character range` error. There's not a lot of those, so it feels more that we're expanding the original [`export LC_CTYPE=$LANG`](https://github.com/robbyrussell/oh-my-zsh/pull/2012/files#diff-d0d9b77e802868afae51086e4b3e7dd1L15) from the old days than we're actually fixing anything. But it is however an issue that I'm not comfortable just deleting those lines and call it a day.

But yes I agree with

> LANG and LC_* are control points for the user to set their locale or override parts of it, and downstream programs should respect that. Yeah, some users misconfigure them, but I don't think OMZ has enough information available to "fix" them if they're in a weird state.

> And IMHO, when discussing changing well-known values or interface points, we should include references/links to authoritative sources or standards, to ground the discussion.

This needs a lot of testing on very different locales and systems, but it's ok because there is no rush to it. Let's put this on the slow track and I'll work on devising a big enough test system to get a grip on the issue.

apjanke:
Okay. Yeah, it probably makes sense to require a UTF-8 environment for OMZ, since we're using it internally. And realistically, I'd guess pretty much 100% of OMZ's users are running on UTF-8 systems, even if they've changed their locale to something different. I just dunno about actually setting it ourselves, or setting up heuristics to do so. I really dunno.

Anyway, a few miscellaneous things.

I've never understood why `export LC_CTYPE=$LANG` is needed in the first place. The default effective value for `$LC_CTYPE` [is already `$LANG`](http://pubs.opengroup.org/onlinepubs/7908799/xbd/envvar.html) (unless `LC_ALL` is set, and then `LC_ALL` takes precedence anyway), so it doesn't seem like it would have any effect. Is it anticipating the user changing `$LANG` later on in the process? Are there noncomforming systems we know of that specifically require `LC_CTYPE` to be set, and don't use the normal locale fallback logic?

> LC_CTYPE=${LC_CTYPE:-${LC_ALL:-${LANG:-C}}}

This order doesn't seem right. `LC_ALL` has precedence over both `LC_CTYPE` and `LANG`, so should probably come first.

If we're going to test for the current locale's character encoding, how about checking it directly using `locale` queries, instead of writing our own `LC_*` resolution logic? Something like `encoding=$(locale charmap)` will give you the current effective character encoding. I think that's portable.

> available=("${(f)$(locale -a 2> /dev/null)}")

This ends up choosing `af_ZA.UTF-8` on my system. That should work fine, but it'll look weird. If someone's having locale issues, and isn't familiar with OMZ, it could be confusing.

If `LC_ALL` is set, setting `LC_CTYPE` won't work, because `LC_ALL` takes precedence.

The `lib/misc.zsh` file might be the wrong place for this. That's (currently) the thirteenth lib file loaded during OMZ startup. Locale is pretty basic, and if we want to enable/require UTF-8 "everywhere" in OMZ, this should probably happen way up in `oh-my-zsh.sh` before anything else is loaded.

The user could always change `LC_CTYPE` or `LC_ALL` after OMZ initialization runs, in which case calls to code with `\u` or other Unicode-dependent stuff would still break. Doing the protection locally inside functions would avoid this. (But then, who knows how that Unicode data will interact with the stuff the user is doing in a non-UTF-8 locale?)

### Maybe be stricter?

Do we have any examples of actual non-UTF-8 systems in the wild, as opposed to just users with misconfigured locale settings or terminal profiles? Maybe we'd be better off having OMZ throw an error or just not load if it's in a non-UTF-8 environment, and throw the ball back to the user to fix their locale. I just dunno what the right thing to do here is, because I'm not sure what scenarios we're actually fixing.

Have we seen a non-UTF-8 locale that's not `C` in the wild?

I know OMZ is supposed to be an easy on-ramp for `zsh` and the command line. But supporting a situation where someone requires OMZ to get their environment working (aside from features which are provided by OMZ) just doesn't feel right to me. 

###  Scenarios

Maybe this would be easier to think about if we have some concrete scenarios. Here's the ones I can think of.

* User set their terminal profile to not set locale variables, so their locale defaults to `C`.
* User set `LANG`/`LC_CTYPE`/`LC_ALL` to `C` or something else in their rc files because they read advice to do so somewhere.
* User set `LANG=C` deliberately to turn off localization and get asciibetical ordering or different regex matching behavior

###  Agnoster

That local locale setting (heh heh) in agnoster may be redundant once a change like this goes in, but I'd still like to leave it in for development workflow purposes. Unlike most of our themes, Agnoster was designed to run outside OMZ as well, so the mainline Agnoster will need to keep that defensive code. And if we keep in line with mainline Agnoster, then we could update it just by straight copying the files from [upstream](https://github.com/agnoster/agnoster-zsh-theme) into OMZ and slapping an extra README on it, instead of having to change any code.

vincentbernat:
For agnoster, it is in a separate commit to be able to discard it easily. For `LC_ALL`, you are right. I'll correct the order but I don't know what's the best to do if this is set. Maybe just error out like you said in this case since there is little reason to set this value to a non-UTF-8 locale, while there is a valid reason to set LANG in a such way.

As for selecting af_ZA.UTF-8, it may look odd but at least, we are sure it's installed, while selecting en_US.UTF-8, there is absolutely no guarantee that it is installed. On Ubuntu, I believe by default, it is installed on a server system as part of the few selected locales automatically installed and on desktop, it is installed because all locales get installed, but for example, on Debian, the user is asked which locales to compile and a French-speaking user may just select the appropriate fr_* one.

vincentbernat:
Well, as for the order, I don't think that `${LC_ALL:-${LC_CTYPE:-${LANG:-C}}}` makes more sense if we cannot override `LC_ALL`. I leave it is as is until there is some consensus on how to handle people with LC_ALL set to a non-UTF-8 locale.

apjanke:
> Well, as for the order, I don't think that `${LC_ALL:-${LC_CTYPE:-${LANG:-C}}}` makes more sense if we cannot override LC_ALL. 

At least you would be able to correctly detect what the encoding was, and whether you were able to successfully override it. If someone has set `LC_ALL` to a UTF-8 locale, then we don't need to set `LC_CTYPE`, and the attempt wouldn't be made. And you can check the encoding again after setting `LC_CTYPE`, to see if it took effect; if it is overridden by `LC_ALL`, we could issue a warning or error with a better message than the `character out of range` that people get when something requiring Unicode is encountered when in a non-UTF-8 locale.

`encoding=$(locale charmap)` may be easier anyway.

PR #4697, opened 2015-12-15
znt: more optimizing workarounds for 5.0.6 <= zsh < 5.2

psprint:
Optimizations towards heap management of set of Zsh versions. Search of three letter keyword is 6.7 sec vs. 15.3 sec in one of tests

### DIFF
--- a/plugins/zsh-navigation-tools/README.md
+++ b/plugins/zsh-navigation-tools/README.md
@@ -105,7 +105,7 @@ colorize output of the tools, via their config files (check out e.g. n-cd.conf,
 it uses this).
 
 ## Performance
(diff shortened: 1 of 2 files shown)


psprint:
Basically a two-line patch

mcornella:
We should have less frequent pull requests about this, maybe you want to adopt a stable/unstable branch or something?

psprint:
I will have this in mind. The project is finished and I had repeating ideas about polishing its speed. There should be no such pull requests at this point, I'm pretty sure. For eventual larger updates I will consider what you say. Thanks!

mcornella:
Cool, thank you!

PR #4700, opened 2015-12-15
Update gulp.plugin.zsh

giordanovi:
Remove `local` stuff that cause an error in shell


mcornella:
That's weird, it worked on my end...

PR #4702, opened 2015-12-15
colored-man-pages: Use Portable `less` Path

prikhi:
The `colored-man-pages` plugin breaks the `man` command on NixOS. This commit fixes the issue by using `/usr/bin/env` to determine the path to the `less` command.

### DIFF
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -25,7 +25,7 @@ man() {
 		LESS_TERMCAP_so=$(printf "\e[1;44;33m") \
 		LESS_TERMCAP_ue=$(printf "\e[0m") \
 		LESS_TERMCAP_us=$(printf "\e[1;32m") \
-		PAGER=/usr/bin/less \
+		PAGER="${commands[less]:-$PAGER}" \
 		_NROFF_U=1 \
 		PATH="$HOME/bin:$PATH" \
 			man "$@"


plugins/colored-man-pages/colored-man-pages.plugin.zsh
```diff
@@ -25,7 +25,7 @@ man() {
 		LESS_TERMCAP_so=$(printf "\e[1;44;33m") \
 		LESS_TERMCAP_ue=$(printf "\e[0m") \
 		LESS_TERMCAP_us=$(printf "\e[1;32m") \
-		PAGER=/usr/bin/less \
+		PAGER='/usr/bin/env less' \
```
mcornella: Thanks for spotting this! 

Let's use a builtin zsh feature though: `$commands[name]` will return the path of the command *name* if it exists (otherwise it returns blank). So here we could use that to even protect this in systems where `less` is not installed:
```zsh
PAGER="${commands[less]:-$PAGER}" \
```
This has the benefit of working also if `env` is not installed in `/usr/bin/env`.

prikhi:
Cool, fixed  - what does the `:-$PAGER` do?

mcornella:
It uses a default value ― in this case, `$PAGER`, in case `$commands[less]` is not defined or null. See [Parameter expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion).

That's perfect thanks :+1:

PR #4707, opened 2015-12-16
Add alias for branch delete

robr3rd:


shigemk2:
I think this is better because D option is a little dangerous.
```sh
alias gbd='git branch -d'
```

robr3rd:
I am 100% fine with that. Branch updated appropriately.

plugins/git/git.plugin.zsh
```diff
@@ -46,6 +46,7 @@ alias gapa='git add --patch'
 
 alias gb='git branch'
 alias gba='git branch -a'
+alias gbd='git branch -d'
```
sachin21: I think gdb is looks like a `gdb` command. so User will occurs gdb as a typo.

robr3rd: Though possible that it may be entered as a typo from time to time, the same could be said for most other aliases as well. If this is of concern, then honestly I'd be more worried about [`gsb`](https://github.com/robr3rd/oh-my-zsh/blob/patch-1/plugins/git/git.plugin.zsh#L193) (`alias gsb='git status -sb'`) if I were you--something that I *did actually* typo just now while testing for `gdb`.

The only time a typo matters, though, is when it causes immediate detrimental results, which in this case it should not because not only do most people not have `gdb` installed, but even if they do and they manage to typo this, the result would be:

1. Dropping them to a virtual terminal because the name of the branch is != to any file name in their current working directory, and since many branching models follow a `feature/*`, `feature-`, or `[ticket #]-description` convention, this is highly likely.
2. Complain that the file that *does* exist by the same name in their current working directory is not an executable type file.
3. Try and debug the file, which....is harmless.

jvz:
This alias has already been merged to the git plugin at some point.

PR #4712, opened 2015-12-18
when in a ssh session and in case of default user only display where …

ge045:
…instead of who@where:

This commit changes the prompt of the agnoster theme to only display the hostname instead of user@hostname in case the user is the default user and the session is ssh.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1,21 +1,21 @@
-![Oh My Zsh](https://s3.amazonaws.com/ohmyzsh/oh-my-zsh-logo.png)
-
+<p align="center">
+  <img src="https://s3.amazonaws.com/ohmyzsh/oh-my-zsh-logo.png" alt="Oh My Zsh">
+</p>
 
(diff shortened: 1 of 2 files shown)


themes/agnoster.zsh-theme
```diff
@@ -79,8 +79,10 @@ prompt_end() {
 
 # Context: user@hostname (who am I and where am I)
 prompt_context() {
-  if [[ "$USER" != "$DEFAULT_USER" || -n "$SSH_CLIENT" ]]; then
+	if [[ "$USER" != "$DEFAULT_USER" ]]; then
```
sachin21: Indent is broken.
Could you change indents from 1 tab to 2 spaces?
FYI: https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide#indentation

ge045:
@sachin21 yep, thanks. fixed it.

sachin21:
+1

ge045:
created pull request from master branch; will do a new request from feature branch

PR #4713, opened 2015-12-18
forklift: update homepage URL and tweak formatting

### LINKED ISSUE #4710: forklift plugin: broken link to app
The current link in Forklift's README goes to http://forkliftapp.com/forklift/, which is an Asian-language "nothing found" page for me.

Looks like the new site might be: http://www.binarynights.com/forklift/

apjanke:
Fixes #4710

### DIFF
--- a/plugins/forklift/README.md
+++ b/plugins/forklift/README.md
@@ -4,10 +4,12 @@ Plugin for ForkLift, an FTP application for OS X.
 
 ### Requirements
 
-* [ForkLift](http://forkliftapp.com/forklift/)
+* [ForkLift](http://www.binarynights.com/forklift/)
 
 ### Usage
 
-* If `fl` is called without arguments then the current folder is opened in ForkLift. Is equivalent to `fl .`
+<code>fl [*file_or_folder*]</code>
 
-* If `fl` is called with a directory as the argument, then that directory is opened in ForkLift
+* If `fl` is called without arguments then the current folder is opened in ForkLift. This is equivalent to `fl .`.
+
+* If `fl` is called with a directory as the argument, then that directory is opened in ForkLift. If called with a non-directory file as the argument, then the file's parent directory is opened.


plugins/forklift/README.md
```diff
@@ -4,10 +4,12 @@ Plugin for ForkLift, an FTP application for OS X.
 
 ### Requirements
 
-* [ForkLift](http://forkliftapp.com/forklift/)
+* [ForkLift](http://www.binarynights.com/forklift/)
 
 ### Usage
 
-* If `fl` is called without arguments then the current folder is opened in ForkLift. Is equivalent to `fl .`
+<code>fl *file*</code>
```
mcornella: This `file` is confusing, it should be folder or directory. The `<code>` should work with regular backticks. And perhaps putting the argument in brackets `[]` would underscore that it is an optional argument.

plugins/forklift/README.md
```diff
@@ -4,10 +4,12 @@ Plugin for ForkLift, an FTP application for OS X.
...
 
-* If `fl` is called with a directory as the argument, then that directory is opened in ForkLift
\ No newline at end of file
+If `fl` is called without arguments then the current folder is opened in ForkLift. This is equivalent to `fl .`.
```
mcornella: The rendered readme looks better if these are still list items, in my opinion.

mcornella:
I left a couple of suggestions but this is almost ready to merge

plugins/forklift/README.md
apjanke: I've switched it to `file_or_folder` and added some additional explanatory text.

The formatting doesn't work with regular backticks: then the nested '*...*' text gets rendered as literal asterisks, instead of italicizing it.

plugins/forklift/README.md
apjanke: Switched them back to list items.

PR #4718, opened 2015-12-21
Plugin rake-fast: add Rails compatibility

allenan:
This PR adds Rails compatibility to the rake-fast plugin. 

The problem I was running into is that currently, rake-fast is only watching the main Rakefile for changes in order invalidate the cache. However, in Rails, the main Rakefile rarely changes. Instead, it requires all the *.rake files defined in lib/tasks/.

This change checks to see if any .rake files have been changed, in addition to the existing checks.

I discussed this with @KevinBongart on the original rake-fast repo: https://github.com/KevinBongart/rake-fast/issues/4

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -8,7 +8,23 @@ _rake_refresh () {
 }
 
 _rake_does_task_list_need_generating () {
-  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]]
+  _rake_tasks_missing || _rakefile_has_changes || _sub_rake_files_have_changes
+}
+
+_rake_tasks_missing () {
+  [[ ! -f .rake_tasks ]]
+}
+
+_rakefile_has_changes () {
+  [[ Rakefile -nt .rake_tasks ]]
+}
+
+_sub_rake_files_have_changes () {
+  if find . -type f -name "*.rake" 2>/dev/null | grep -q .; then
+    [[ $(find ./**/*.rake(.om[1])) -nt .rake_tasks ]]
+  else
+    false
+  fi
 }
 
 _rake_generate () {


plugins/rake-fast/rake-fast.plugin.zsh
```diff
@@ -8,7 +8,19 @@ _rake_refresh () {
...
+_rake_tasks_missing () {
+  [[ ! -f .rake_tasks ]]
+}
+
+_rakefile_has_changes () {
+  [[ Rakefile -nt .rake_tasks ]]
+}
+
+_sub_rake_files_have_changes () {
+  [[ $(find ./**/*.rake(.om[1])) -nt .rake_tasks ]]
```
KevinBongart: Good to see you've split the logic into smaller methods, which will be easier for other maintainers who come across this file. Would you mind explaining this specific line a little more? In particular: the structure (do you want to limit this to rake files that are in the usual `lib/tasks` directory? or any rake file?) and the `(.om[1])` part.

allenan: Sure. I figured it should look at any *.rake file recursively in the current directory. Just in case there are nested folders within the lib/tasks directory or if rake files are defined elsewhere.

So the `./**/*.rake` part gets all files with the .rake extension anywhere in the current directory. Then in the parentheses, `.` specifies it must be a file (as opposed to a directory), `om` orders the results by modified at time, and `[1]` returns the first result (the most recently modified file). 

I used http://reasoniamhere.com/2014/01/11/outrageously-useful-tips-to-master-your-z-shell/ as a reference for the parts in the parentheses.

sachin21: My solution is only this:
```
function _rake_command () {
  if [ -e "bin/rake" ]; then
    bin/rake $@
  elif type bundle &> /dev/null && [ -e "Gemfile" ]; then
    bundle exec rake $@
  else
    command rake $@
  fi
}
```
this solution is all is well in my environment.
Why not?

sachin21: Oops my bad...sorry Please forget what I said now.

KevinBongart:
This works great on my Mac, thanks for the thorough explanation!

:+1: :shipit:

cc @mcornella @robbyrussell

sachin21:
@mcornella @robbyrussell @allenan How is it going? I really want to use this feature.

mcornella:
@sachin21 fixed in #5343.

sachin21:
@mcornella Thank you.

PR #2747, opened 2014-04-24
Add brew alias file

sachin21:
add to plugins/brew/brew.plugin.zsh

Please merge this pull request.

PR #4719, opened 2015-12-22
Add aliases for Homebrew

sachin21:
Added aliases for Homebrew.
This PR is retry for https://github.com/robbyrussell/oh-my-zsh/pull/2747.

### DIFF
--- a/plugins/brew/brew.plugin.zsh
+++ b/plugins/brew/brew.plugin.zsh
@@ -1,4 +1,20 @@
-alias brews='brew list -1'
 alias bubo='brew update && brew outdated'
 alias bubc='brew upgrade && brew cleanup'
 alias bubu='bubo && bubc'
+
+alias bb="brew bundle"
+alias bdr="brew doctor"
+alias bt="brew tap"
+alias bcl="brew cleanup"
+alias bill="brew install"
+alias brein="brew reinstall"
+alias bupd="brew update"
+alias bgrad="brew upgrade"
+alias brch="brew search"
+alias blink="brew link"
+alias bli="brew list"
+alias bsl="brew services list"
+alias bcat="brew cat"
+alias bcmds="brew commands"
+alias bcre="brew create"
+alias binfo="brew info"


KevinBongart:
Is this whole piece of logic worth it? While I'm sure it works, and it is indeed a clever way of avoiding `alias`, wouldn't repetition be easier for you, me, users and future maintainers to work with something more straightforward?

```zsh
alias bb='brew bundle'
alias bcat='brew cat'
alias bcl='brew cleanup'
alias bcmds='brew commands'
alias bcre='brew create'
alias bdr='brew doctor'
alias bgrad='brew upgrade'
alias bill='brew install'
alias binfo='brew info'
alias bli='brew list'
alias blink='brew link'
alias brch='brew search'
alias bsl='brew services list'
alias bt='brew tap'
alias bupd='brew reinstall'
```

sachin21:
I got it. It's done :smiley:

KevinBongart:
@sachin21 This looks good, thanks for simplifying it!

I see some cons against this PR though: not sure every Homebrew command is used so frequently that it would need an alias, which could potentially conflict with another command or program (for instance, I bet there are other programs named `bill`, `bt` and `blink` out there).

It's also unlikely someone would need an alias to reinstall brew, which is a rare and destructive operation, where you want to make sure you're typing the right command.

Maybe this PR could focus on bringing aliases for the most common Homebrew commands instead of finding shortcuts for every possible feature? What do you think?

I saw on #2747 that other users seem interested in your changes, so don't let me opinion get in the way of these aliases! :smile:

cc @ncanceill @nicolas-brousse @mcornella @robbyrussell

plugins/brew/brew.plugin.zsh
```diff
@@ -1,4 +1,19 @@
...
 alias bubo='brew update && brew outdated'
 alias bubc='brew upgrade && brew cleanup'
 alias bubu='bubo && bubc'
+
+alias bb="brew bundle"
+alias bdr="brew doctor"
+alias bt="brew tap"
+alias bcl="brew cleanup"
+alias bill="brew install"
+alias bupd="brew reinstall"
```
KevinBongart: This looks like a mistake, should it `brew update` or `brew reinstall`?

sachin21: my bad. fixed and added an alias for `brew update`.

sachin21:
@KevinBongart I think it should think by owner and collaborator, I want to know thinking by owner and collaborator.
but I'd say oh-my-zsh's user should use oh-my-zsh’s aliases than own aliases.

Cc: @ncanceill @nicolas-brousse @mcornella @robbyrussell

nicolas-brousse:
I'm agree with @KevinBongart that some commands are not frequently used. So they don't really need to have an alias (like `bcmds` and `bcre`).

Like there is [for `git`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L41) we could have this alias:
```sh
alias b='brew'
```
That's allow us to do `b commands` or `b create`.

sachin21:
I see. Thanks :smile:

PR #4722, opened 2015-12-23
Adding grails + sdkman scripts home

orkonano:
Add the home path of the sdkman current grails installation


mcornella:
I have no idea how to reproduce an installation so that there is a folder in that path, and this is already very old, so I'm closing it. Feel free to reopen and explain a bit more.

PR #4723, opened 2015-12-23
Update gitignore.plugin: comma separated arguments

low-ghost:
Seems like this is more like the intended effect, splits arguments with commas for proper gitignore.io api request. Thanks for teaching me about custom compdef completions!

### DIFF
--- a/plugins/gitignore/gitignore.plugin.zsh
+++ b/plugins/gitignore/gitignore.plugin.zsh
@@ -1,4 +1,4 @@
-function gi() { curl -sL https://www.gitignore.io/api/$@ ;}
+function gi() { curl -sL https://www.gitignore.io/api/`echo $@ | sed s/" "/,/g` ;}
 
 _gitignoreio_get_command_list() {
   curl -sL https://www.gitignore.io/api/list | tr "," "\n"


plugins/gitignore/gitignore.plugin.zsh
```diff
@@ -1,4 +1,4 @@
-function gi() { curl -sL https://www.gitignore.io/api/$@ ;}
+function gi() { curl -sL https://www.gitignore.io/api/`echo $@ | sed s/" "/,/g` ;}
```
mcornella: A better way to join arguments with a delimiter is with zsh builtin join syntax: `${(j:delimiter:)variable}`. In this case:
```zsh
curl -sL https://www.gitignore.io/api/${(j:,:)@}
```

low-ghost: Whoa killer stuff.Not too familiar with zsh specific syntax and wouldn't have even thought of something that awesome

apjanke: Check out the [14.3. Parameter Expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion) section in the ZSH reference manual if you want to learn more about it. It'll show you tons of stuff you can do like this, possibly replacing a bunch of external calls to `grep`, `sed`, `tr`, and the like.

mcornella:
I merged the short version

low-ghost:
thanks @mcornella

PR #4726, opened 2015-12-24
[battery plugin] Support multiple batteries

weirdan:
On a system with multiple batteries (like thinkpads) report percentage
and time remaining only for the active battery (the one being
discharged).

Ideally we should report all batteries, but acpi only shows time remaining
for the active battery. Also callers of these functions expect a single
return value. This is still better than reporting 596% remaining (like it
did on my laptop).

For the reference, the output of acpi command with multiple batteries looks
like this:

```
Battery 0: Unknown, 5%
Battery 1: Discharging, 86%, 03:14:04 remaining
```


dgrabla:
+1

PR #4733, opened 2015-12-28
Plugin pkgin

jefflance:
Propose a plugin providing autocomplete option for pkgin package manager from joyent (https://pkgsrc.joyent.com/).


mcornella:
Has no interest.

PR #4738, opened 2015-12-31
Vault basic autocompletion

valentinbud:
https://www.vaultproject.io

### DIFF
--- a/plugins/vault/README.md
+++ b/plugins/vault/README.md
@@ -1,0 +1,18 @@
+## Vault (https://www.vaultproject.io) autocomplete plugin
+
+- Adds autocomplete options for all vault commands.
+
+####Show help for all commands
+![General Help](http://i.imgur.com/yv5Db1r.png "Help for all commands")
+
+
+####Create new Vault token
+![Create token](http://i.imgur.com/xMegNgh.png "Create token")
+
+
+####Enable audit backends
+![Audit backends](http://i.imgur.com/fKLeiSF.png "Audit backends")
+
+
+
+Crafted with <3 by Valentin Bud ([@valentinbud](https://twitter.com/valentinbud))
--- a/plugins/vault/_vault
+++ b/plugins/vault/_vault
@@ -1,0 +1,400 @@
+#compdef vault
+
+typeset -a main_args
+main_args=(
+    '(-version)-version[Prints the Vault version]'
+    '(-help)-help[Prints Vault Help]'
+)
+
+typeset -a general_args
+general_args=(
+    '(-help)-help[Prints Help]'
+    '(-address)-address=-[The address of the Vault server. Overrides the VAULT_ADDR environment variable if set.]:address:'
+    '(-ca-cert)-ca-cert=-[Path to a PEM encoded CA cert file to use to verify the Vault server SSL certificate. Overrides the VAULT_CACERT environment variable if set.]:file:_files -g "*.pem"'
+    '(-ca-path)-ca-path=-[Path to a directory of PEM encoded CA cert files to verify the Vault server SSL certificate. If both -ca-cert and -ca-path are specified, -ca-path is used.Overrides the VAULT_CAPATH environment variable if set.]:directory:_directories'
+    '(-client-cert)-client-cert=-[Path to a PEM encoded client certificate for TLS authentication to the Vault server. Must also specify -client-key. Overrides the VAULT_CLIENT_CERT environment variable if set.]:file:_files -g "*.pem"'
+    '(-client-key)-client-key=-[Path to an unencrypted PEM encoded private key matching the client certificate from -client-cert. Overrides the VAULT_CLIENT_KEY environment variable if set.]:file:_files -g "*.pem"'
+    '(-tls-skip-verify)-tls-skip-verify[Do not verify TLS certificate. This is highly not recommended. Verification will also be skipped if VAULT_SKIP_VERIFY is set.]'
+)
+
+typeset -a audit_enable_args
+audit_enable_args=(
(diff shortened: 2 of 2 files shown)


valentinbud:
Hello,

Can someone tell me what do I need to do to get this merged in. I wouldn't like to keep
a separate oh-my-zsh repository just to have this in.

Thank you :).

plugins/vault/vault.plugin.zsh
```diff
@@ -0,0 +1,408 @@
...
+
+    if [[ $state == subcommands ]]; then
+        # (( CURRENT -- ))
+        curcontext="${curcontext%:*:*}:vault-$words[1]:"
+        _call_function ret _vault_$words[1]
+        # print $curcontext
+    fi
+}
+
+compdef _vault vault
```
mcornella: This is not needed since you have the `#compdef vault` statement at the beginning of the file. You should rename the file to `_vault` as well.

Also, you should call `_vault "$@"` at the end of the file so that the initial tab completion works. Otherwise, the first tab will only load the completion. You could also get rid of the `_vault` function and extract its content outside of it.

mcornella:
I have provided some changes you should make to the plugin. That said, it seems there's a completion option builtin into `vault` that would make the current hardcoded flags obsolete. See https://blog.jcoglan.com/2013/02/12/tab-completion-for-your-command-line-apps

plugins/vault/vault.plugin.zsh
valentinbud: @mcornella thanks for reviewing this :). I have renamed the file to `_vault` as you suggested. I have also removed the `_vault()` function and added `_vault "$@"` at the end of the file.

After I reload zsh if I write `vault` and hit `\t` I get the following error:

```bash
[...]
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep
_tags:comptags:36: nesting level too deep
[1]    29100 segmentation fault  zsh
```

I also post here the last lines from the `_vault` file.

```bash

local curcontext=$curcontext ret=1
_arguments : \
    ${main_args[@]} \
    '*:: :->subcommands' && ret=0
if ((CURRENT == 1 )); then
    _vault_commands && ret=0
fi
if [[ $state == subcommands ]]; then
    # (( CURRENT -- ))
    curcontext="${curcontext%:*:*}:vault-$words[1]:"
    _call_function ret _vault_$words[1]
fi

_vault "$@"
```

What can I do next?

mcornella: > I have also removed the _vault() function and added _vault "$@" at the end of the file.

Don't do both! :smile: Just remove the `_vault` at  the end of it and it should be fine.

valentinbud:
That's really cool :). It would remove all the boiler plate code from the plugin. But that's about another `vault` (https://getvau.lt/). The above plugin is for HashiCorp's Vault Project (https://www.vaultproject.io/). getvau.lt is a password manager while [vault](https://github.com/hashicorp/vault) is a 'tool for managing secrets', all kinds of them, API tokens, PKI certificates, etc.

plugins/vault/vault.plugin.zsh
valentinbud: Awesome! It works :). Should I make a new commit and push it on the branch this pull request was made from, `feature-branch` or should I use `git commit --amend` or should I squash the commits into a single commit and push them to `feature-branch`?

mcornella: Amend and push force is the right way to go :+1:

mcornella: Oh, making it into 1 single commit. That wasn't clear, if you have more than 2 commits you'll want to squash them.

mcornella:
> But that's about another vault (https://getvau.lt/). The above plugin is for HashiCorp's Vault Project (https://www.vaultproject.io/). getvau.lt is a password manager while vault is a 'tool for managing secrets', all kinds of them, API tokens, PKI certificates, etc.

Ok, I see that now. I didn't take a good look at it otherwise I would've caught that :sweat_smile:

mcornella:
Thanks!

plugins/vault/vault.plugin.zsh
valentinbud: Done :). It's only one commit, I apologise for not being clear enough. Could you please tell me, in a few words, what is the difference between using `_vault` or `_notation` and `vault.plugin.zsh`? I have seen oh-my-zsh plugins containing both.

mcornella: The `_name` files will be autoloaded (or lazy loaded) when the first completion is triggered, and used as if they are a whole function. Those `name.plugin.zsh` have to provide a `_name` function for completion, which will be loaded when the plugin file is sourced. 

The biggest reason to use the `_name` notation though is that it is patently clear that it is a completion file.

valentinbud: Thanks @mcornella.

PR #4739, opened 2016-01-03
Fix bug on robbyrussell theme face

leaxoy:
fix  robbyrussell theme PROMPT when in a git repo. When under a git repo, there were only one space left where other case two spaces.


apjanke:
While you're in there, it looks like there's an extra `%` in the prompt string.

```
PROMPT='${ret_status}%{$fg_bold[green]%}%p %{$fg[cyan]%}%c %{$fg_bold[blue]%}$(git_prompt_info)%{$fg_bold[blue]%} % %{$reset_color%}'
```

That lone `%` before the final `%{$reset_color%}` might be a mistake. `%`s in prompts are for escape sequences, so it's probably being interpreted as a ``% '` (%-space), which isn't a valid escape sequence, and is producing no output. If it's intended to show up as a literal `%` in the prompt, it needs to be escaped as `%%`. Otherwise, it and the following space could probably be removed. Its presence may be making it harder to figure out what the right spacing in the theme should be.

mcornella:
Fixed on 20b8b6c and 3d522ff.

PR #4742, opened 2016-01-03
pert calculator plugin

arzzen:
A  command line PERT calculator for quick estimates.


mcornella:
Has no interest. It also looks like this should be its own command-line utility.

PR #4747, opened 2016-01-05
missing rails rake task alias added

trkrameshkumar:


trkrameshkumar:
@sachin21 changed `rst` to `rsct` and tested other key words, pls check

sachin21:
@trkrameshkumar I think looks good. but I'm not collaborator. please wait owner or collaborator reply.
>> @robbyrussell @mcornella

PR #4751, opened 2016-01-06
git.plugin.zsh: Don't run Git hooks when making a WIP commit

allanlewis:
When making a WIP commit, we generally just want to save the state of the current branch temporarily, maybe because we want to push our work for backup purposes, or change branch to work on something else. Therefore, it's generally undesirable to run Git hooks, which might do things like run linters, because we probably don't care if our WIP has lint errors.


allanlewis:
Does anyone object to this? I think it's a good change, and it's a one-liner...

mcornella:
Great idea, sorry for the delay!

allanlewis:
Thanks, @mcornella!

PR #4752, opened 2016-01-07
added option to enable / disable ssh support with gpg-agent

### LINKED ISSUE #2121: gpg-agent plugin should check existence of ssh-agent before sourcing from GPG_ENV file

### LINKED ISSUE #2496: gpg-agent plugin starts ssh-agent when ssh-agent is already running
The gpg-agent plugin tests to see if any identities are loaded in an existing ssh agent. If no identities are loaded, it fails. This is an indirect way to determine if ssh-agent is running, at best. Submitted a PR with a better check.

decomposite:
Over the last week I've started using gpg-agent and was experiencing issues #2121 and #2496 without knowing what exactly is happening. Finally narrowed it down to this plugin so thought I might take a shot at it:

I originally wrote a check just to see if SSH_AUTH_SOCK or SSH_AGENT_PID ( `[[ -n $VAR ]]` ) is set and if it is, it'll start without ssh support. Reason being if those variables are set then either ssh-agent or gpg-agent is already running so you probably don't want that overwritten. This would work with OSX as well since the built in ssh-agent will still set the variable without actually showing itself as a "ssh-agent process" (I'm still annoyed by this being built in, but that's beside the point). Something like `[[ -n $SSH_AUTH_SOCK ]]` would work well for this.

However, I ended up changing this for what I think is a better solution.

I saw with the ssh-agent plugin that you could set arguments with zstyle. So I just set one up with this plugin as well which makes it so that it's off by default and that you should explicitly set it. This should solve issues involving gpg-agent overriding the variables needed for ssh-agent to do it's thing and is preferable in pretty much all situations i.e. if you really prefer gpg-agent then you absolutely wouldn't mind it overriding the variables set by ssh-agent, so why check for them in the first place? Just flip a switch.

I'd like to add that I really don't like the ssh support feature for gpg-agent. It doesn't bring anything new to the table as your SSH key is already symmetrically encrypted anyway and the only benefit I can think of is saving yourself a process. I honestly don't think it should be enabled unless you know you want it, even if there isn't ANY ssh-agent daemon running because it adds complexity to something that is pretty unrelated in most cases. It honestly seems like a case of feature creep to me.

Lastly, if this solution isn't okay I can PR my aforementioned one if you want. Just let me know.

FIxes #2121
Fixes #2496

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -1,5 +1,9 @@
 local GPG_ENV=$HOME/.gnupg/gpg-agent.env
 
+# Enable SSH support by adding this line to your .zshrc file:
+# zstyle :omz:plugins:gpg-agent ssh-support on
+local _plugin__ssh_support
+
 function start_agent_nossh {
     eval $(/usr/bin/env gpg-agent --quiet --daemon --write-env-file ${GPG_ENV} 2> /dev/null)
     chmod 600 ${GPG_ENV}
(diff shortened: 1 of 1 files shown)


fringerdarlinger:
Can I get an update on whether or not you can merge this? It currently has no conflicts :)

Issue #1984, opened 2013-07-19
unable to upgrade


brewster1134:
**[Go to solution](https://github.com/robbyrussell/oh-my-zsh/issues/1984#issuecomment-331808425)**

```
[Oh My Zsh] Would you like to check for updates?
Type Y to update oh-my-zsh: Y
Upgrading Oh My Zsh
Cannot pull with rebase: You have unstaged changes.
Please commit or stash them.
There was an error updating. Try again later?
```

RussellBradley:
Same problem... One liner fix.
$ .oh-my-zsh && git stash && upgrade_oh_my_zsh

kelthenoble:
Thanks russelBradley :+1:

OrcunOtacioglu:
The one liner solved my issue. Thank you RussellBradley!

dadocreate:
Thanks russelBradley :+1:

cviebrock:
And don't forget to `git stash pop` after you upgrade to re-apply your changes (if desired).

iamgoodbytes:
Thanks @RussellBradley - that works!

jinga50334:
@ODelibalta It works! thank you!

ylluminarious:
`$ sudo git reset --hard` fixed this for me.

zhanghanyuan:
Thanks russelBradley :+1:

QaziPython:
Thanks @RussellBradley :+1:

kayoub5:
if any of the above does not work for you try this
```sh
cd ~/.oh-my-zsh
git config core.filemode false
```

pksurecom:
Thanks russelBradley !

gerson:
Thanks @RussellBradley

Mark-Creeten:
Thanks russelBradley !

Karthicksree:
Thanks @RussellBradley.

eranhirs:
I happen to be working on a Windows 10 ubuntu bash, so the git changes I had were correlated to running `dos2unix` on the entire directory to fix the `^M` character problem.

After successfully updating, you might want to run
`cd $ZSH`
`find . -type f -print0 | xargs -0 dos2unix`
To fix errors correlated to the `^M` character.

Hope this helps someone.

shahzeb1:
On Mac: 
```shell
cd .oh-my-zsh
git stash
upgrade_oh_my_zsh
git stash pop
```

MrEmelianenko:
[...]

PR #4753, opened 2016-01-07
prompt user if repo has changes before upgrading

jamespullar:
If for any reason a file changes in your ~/.oh-my-zsh directory, auto upgrade fails. I added a user prompt in this scenario to offer to stash your changes and continue with the upgrade.

Fix for issue #1984.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -20,10 +20,7 @@ else
   NORMAL=""
 fi
 
-printf "${BLUE}%s${NORMAL}\n" "Updating Oh My Zsh"
-cd "$ZSH"
-if git pull --rebase --stat origin master
-then
+success_upgrading() {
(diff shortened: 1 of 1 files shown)


tools/upgrade.sh
```diff
@@ -34,6 +31,41 @@ then
...
+    success_upgrading
+  else
+    error_upgrading
+  fi
+}
+
+printf "${BLUE}%s${NORMAL}\n" "Updating Oh My Zsh"
+cd "$ZSH"
+
+if output=$(git status --porcelain); then
```
apjanke: Is this the right test? `git status --porcelain` has an exit status of zero even if there are changes, which is what this `if output=$(...); then` is testing. Maybe you want to capture its output and test if that is a nonempty string?

PR #4759, opened 2016-01-08
chruby plugin locals moved inside function

swrobel:
The locals `_chruby_auto` and `_chruby_path` were previously declared at the top of the plugin, but they only seem to be used inside of `_source_from_omz_settings()`. This caused the following odd behavior when using this plugin (locals & their values are printed when sourcing `.zshrc`:

```
$ source ~/.zshrc
_chruby_path=''
_chruby_auto=''
```

@webframp feel free to chime in if there was some rationale that I'm missing - I'm no shellscripting pro

### DIFF
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -16,9 +16,6 @@
 # rvm and rbenv plugins also provide this alias
 alias rubies='chruby'
 
-local _chruby_path
-local _chruby_auto
-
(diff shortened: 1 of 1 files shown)


webframp:
No specific need I know of to have them outside. LGTM :+1:

swrobel:
@mcornella can I get a merge here? Plugin author signed off on it.

mcornella:
:+1: Sorry!

PR #4765, opened 2016-01-11
Added list alias for npm plugin

mathiasvr:
Added the alias npmL for listing installed packages at root level (not showing sub-dependencies).
It also works with the global -g option.


mcornella:
Added in d29784e1aebab5e02db918b1ffbfa4d2d308799a

PR #4767, opened 2016-01-11
Uppercase

rafaelstz:


apjanke:
The case difference in the original is probably intentional. In prompts like `[y/N] `, the uppercase option is the default, so if you just hit Enter, or give it unrecognized input, it functions the same as entering `n`.

mcornella:
That's exactly right. By default we won't uninstall Oh My Zsh: this was introduced in #4464. Closing

PR #4774, opened 2016-01-14
Fixed calculation for battery percentage plugin

MichaelWolman:
Fixed the calculation for the bars to be filled in for the battery plugin. The old formula didn't work if `BATTERY_GAUGE_SLOTS` was changed to be anything other than `10`, the new formula generalizes to any size battery bar.

### DIFF
--- a/plugins/battery/battery.plugin.zsh
+++ b/plugins/battery/battery.plugin.zsh
@@ -91,7 +91,7 @@ elif [[ $(uname) == "Linux"  ]] ; then
   }
 
   function battery_pct_prompt() {
-    b=$(battery_pct_remaining) 
+    b=$(battery_pct_remaining)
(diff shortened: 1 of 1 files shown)


MichaelWolman:
This is the first time I submit a pull request, if I did anything wrong let me know and I'll fix it and try again. Thanks :)

plugins/battery/battery.plugin.zsh
```diff
@@ -91,7 +91,7 @@ elif [[ $(uname) == "Linux"  ]] ; then
   }
 
   function battery_pct_prompt() {
-    b=$(battery_pct_remaining) 
+    b=$(battery_pct_remaining)
```
apjanke: This isn't a problem introduced by you, but it looks like both `$b` and `$color` in this function should be `local`. This may be a good opportunity to fix that.

MichaelWolman: You're right, I went ahead and fixed those.

PR #4782, opened 2016-01-18
Add docker-machine plugin

jakubsacha:
Hey,

As described i add docker-machine plugin for zsh. You may want to merge it.


mdentremont:
There's also: https://github.com/zsh-users/zsh-completions/blob/master/src/_docker-machine

xjewer:
There is also official completion https://github.com/docker/machine/blob/master/contrib/completion/zsh/_docker-machine

PR #4788, opened 2016-01-19
[jump] ask user if existing mark should be overriden

berleon:
`mark` did not worked if the mark was already used.
But sometimes I want to change a previously set mark.
`mark` now simple asks, when an existing mark is detected and
overrides it, if the user agrees.


mcornella:
Fixed in #7197 to directly override if the user agreed to set the mark.

PR #4790, opened 2016-01-20
Avit theme fixes

lckarssen:
This pull request contains a fix for the use of `$CARETCOLOR` in the avit theme, so that a sudo change to root makes the caret red. 

The pull request also contains a small improvement to the theme: it makes the colours of the current directory, return status and two git symbols bold so that they stand out more on dark terminals (like in the Ubuntu default theme).


mcornella:
This added the `_current_dir` function but didn't use it in the prompt.

PR #4795, opened 2016-01-21
when in a ssh session and in case of default user only display where …

ge045:
…instead of who@where


robbyrussell:
We sat on this for over a decade. Guess we'll pass on this

PR #4799, opened 2016-01-25
[Symfony2 plugin] Use new "debug" namespace to avoid deprecation warning

mikaelrandy:
Some Symfony commands was moved to a new "debug" namespace, and fallback are removed in Symfony3.

Today, using "sfroute" or "sfcontainer" shortcut, this kind of warning is displayed : 
```shell
> sfroute
The use of "router:debug" command is deprecated since version 2.7 and will be removed in 3.0. Use the "debug:router" instead.
```

This PR update shortcuts to use new namespace.


felds:
:+1: 
I'm already using Symfony 3 and it works as expected.

gbprod:
:+1:

PR #4800, opened 2016-01-25
Added c_b alias for current branch

mauro-oto:
Allows you to use `git pull origin $(c_b)` or `git push origin $(c_b)` instead of `git pull origin $(current_branch)`, I think it's a convenient alias.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -13,6 +13,9 @@ zstyle -s ":vcs_info:git:*:-all-" "command" _omz_git_git_cmd
 function current_branch() {
   git_current_branch
 }
+
+alias c_b=current_branch
+
 # The list of remotes
 function current_repository() {
   if ! $_omz_git_git_cmd rev-parse --is-inside-work-tree &> /dev/null; then


felds:
I haven't tested it yet, but how does it work?
The function is called `git_current_branch`, shouldn't it be `alias g_c=git_current_branch` ?

mauro-oto:
@felds this is for the `git` plugin, the actual functionality of `current_branch` was pulled in to the core library at `lib/git.zsh`, so `current_branch` is sort of an alias for `git_current_branch`. I thought it made more sense for this shorter version to be in the plugin rather than on the core library.

kevin-verschaeve:
And what do you think of making this alias with the `-g` option ?

With this, we will be able to remove the `$()` and making the command looking: `git pull origin c_b`, c_b will be changed to the current branch before pulling.

(I'm testing this locally, it seems to work has expected, but need confirmation)

mcornella:
I'm not convinced that this will benefit anybody's workflow. What other uses of the `$(current_branch)` part are there that we would need to shorten this instead of, say, making an alias out of the whole thing?─the `git pull origin $(current_branch)` [already exists as `ggl`/`ggpull`](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugin:git#aliases).

felds:
This is not that useful on pull because it's done automatically when tracking the remote (`-u` on the first push).

I think this could only be useful when doing a `git push origin $(current_branch)`, but such an alias already exists: [`ggpush`](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugin:git#aliases)

kevin-verschaeve:
Ok, don't knew the `ggpush` alias. Indeed, this is useless.

mcornella:
Ok then, I'm closing this PR. @mauro-oto you can still add that alias in your zshrc file if you feel you still need it.

PR #4801, opened 2016-01-26
Fix autojump sourcing on OSX with Macports

### LINKED ISSUE #4625: autojump plugin needs update
The latest version of autojump does not have different files to source for the different shells but just one `autojump.sh` that takes care of sourcing the right file (that now is in `share_dir`) for the right shell.

rubendibattista:
The last version of autojump available on Macports does not have anymore different shell scripts (.sh, .zsh, .bash ...) to be sourced but just one autojump.sh that takes care of that located at `/opt/local/etc/profile.d/autojump.sh`

fix #4625


bashu:
@robbyrussell can you merge this one

PR #4811, opened 2016-01-30
Added shorthand for git rev-parse

adarnon:
I have added a command `grp` as a shorthand for `git rev-parse`. It's an annoying long command.


mcornella:
This isn't used as often as to require an alias. Completion is your friend ;)

PR #4813, opened 2016-01-31
Use value of xcode-select path instead of just using the finder default

regnerjr:
## Problem:

By default the `open` command uses the Finder's "Open With" value. The problem with this is that it seems to always be set to the Beta version of Xcode if Xcode is installed.
## Suggestion:

This PR proposes getting the version of Xcode to use from the `xcode-select` tool which is more commonly used to switch between versions of the Xcode command line tools. (Similar to rbenv but for Xcode version) This parameter is more likely to reflect a developers preference for which version of Xcode they would like to use for development. 
## Technical notes

By default the `xcode-select` tool return the full path to the Xcode binary, often "/Applications/Xcode.app/Contents/Developer", while `open -a` expects an Application path. Here we have used bash's parameter expansion to discard the `/Contents/Developer` part of the path.

### DIFF
--- a/plugins/xcode/xcode.plugin.zsh
+++ b/plugins/xcode/xcode.plugin.zsh
@@ -14,8 +14,9 @@ function xc {
     return 1
   else
     echo "Found ${xcode_proj[1]}"
-    XCODEPATH=`xcode-select -p`
-    open -a ${XCODEPATH%/*/*} "${xcode_proj[1]}"
+    active_path=$(xcode-select -p)
+    active_path=${active_path%%/Contents/Developer*}
+    open -a "${active_path}" "${xcode_proj[1]}"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


apjanke:
The `xcode-select` setting may also be pointing to an installation of the command line tools (without an Xcode) instead of an Xcode instance. You may also want to check for that and fall back to the old file-association technique if it isn't pointing to an Xcode.

regnerjr:
Thanks for the feedback @apjanke. I have made the suggested improvement. Thanks for taking a look.

apjanke:
Logic looks good to me now. :+1:

Could make that `active_path` local too, to avoid possible collisions.

mcornella:
Oh sorry I merged #6829 instead, I hadn't seen this one. Thanks for your contribution though!

PR #4814, opened 2016-01-31
Use env to find ssh-add

grahamc:
Some systems don't put `ssh-add` in `/usr/bin`.

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -45,7 +45,7 @@ function _plugin__start_agent()
   zstyle -a :omz:plugins:ssh-agent identities identities
   echo starting ssh-agent...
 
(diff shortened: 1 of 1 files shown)


plugins/ssh-agent/ssh-agent.plugin.zsh
```diff
@@ -45,7 +45,7 @@ function _plugin__start_agent()
   zstyle -a :omz:plugins:ssh-agent identities identities
   echo starting ssh-agent...
 
-  /usr/bin/ssh-add $HOME/.ssh/${^identities}
+  /usr/bin/env ssh-add $HOME/.ssh/${^identities}
```
apjanke: You probably don't need to use `env` here; that trick is mostly for shebang lines, which require full paths for the first argument. An unqualified `ssh-add $HOME/.ssh/${^identities}` should do the same thing.

PR #4822, opened 2016-02-03
Add invoker plugin

deepak:
Adds autocomplete for the invoker gem

more details about the gem at 
http://invoker.codemancers.com 
and https://github.com/code-mancers/invoker


mcornella:
Not popular enough.

PR #4823, opened 2016-02-03
bgnotify: execute notifu in background

ngreduce:
notifu 1.6 waits till notification was dissmised


ngreduce:
moved to standalone repo.
https://github.com/t413/zsh-background-notify/pull/7

PR #4830, opened 2016-02-06
Add git commit --no-edit aliases and fix git commit --signed alias name

### LINKED ISSUE #4827: Why `gcan!` has `-s` option?
It makes unnecessary text like:
```
Signed-off-by: Git Author <my@email.com>
```

I think that `-s` should be removed.

wanbok:
This PR fixes #4827 .

- `gcan!` is `gcan!` without `-s`.
- `gcans!` is `gcan! -s`.
- `gcn!`(`gc! --no-edit`) is added.


plugins/git/git.plugin.zsh
```diff
@@ -58,8 +58,10 @@ alias gbss='git bisect start'
 
 alias gc='git commit -v'
 alias gc!='git commit -v --amend'
+alias gcn!='git commit -v --no-edit --amend'
 alias gca='git commit -v -a'
 alias gca!='git commit -v -a --amend'
+alias gcane!='git commit -v -a --no-edit --amend'
 alias gcan!='git commit -v -a -s --no-edit --amend'
```
Stibbons: actually I would prefer

alias gcan!='git commit -v -a --no-edit --amend'
alias gcans!='git commit -v -a --amend -s --no-edit'

wanbok:
@Stibbons Check this commit, apply your prefer. :)

Stibbons:
i don't have rights to submit, what others think?

mcornella:
@Stibbons quoting the comment at https://github.com/robbyrussell/oh-my-zsh/issues/4827#issuecomment-180795971:

> I also agree, this would break user workflow, and adding this signature is a good practice to set up (signature might be different from the user that push the commit).

I agree with the naming suggested, but it will break the workflow if anyone is using the `gcan!` alias. As of right now I only know of you and @wanbok, but it's irresponsible to thing there's nobody else using it. It's particulary troublesome in this case because the `--no-edit` means the user won't notice that he's not signing off the commit. 

Also, Oh My Zsh doesn't have a good method to deprecate or warn users of alias changes, so **let's keep the `gcan!` alias as is and add `gcane!` for the time being**. _We can also add `gcans!` and instruct people to use that_, as the other one might be deprecated.

Stibbons:
You're right on the point omz doesn't have an alias change signaling feature.

Both solution are ok for me, I can adapt. For alias consistency sake, gcan! + gcans! is right. For user experience consistency sale, gcan! + gcane! is right....
Damned, it is an important choice, should be the author of the project to decide.

wanbok:
I agreed. So I've tried to warn users of alias changes. like this: Deprecated message

And also, `gcane!` is added again.

Stibbons:
hum... should be warned only for some days, or allow user to disable it after a while...

wanbok:
User can avoid this message by using `gcans!`.
I think If a command(or alias) is deprecated, it should be uncomfortably used.
And It should make users be noticed when users trying to use a deprecated command.

wanbok:
@mcornella Any plans for this PR? If it should not notice the deprecated message, i can modify it.

mcornella:
Sorry, I've been a little busy. I would remove the `gcan!` alias and turn it into a function. That way it's more readable.

wanbok:
Update this commit. please check it out. @mcornella

mcornella:
Ok let's pull this in, thanks.

PR #4832, opened 2016-02-07
pylint.plugin.zsh: Remove include-ids=y

eddie-dunn:
The flag `--include-ids` is soon to be deprecated for Pylint. This pull request
removes that option from pylint-quick


mcornella:
Background:

- From [1.5.0 version changelog](http://docs.pylint.org/changelog.html):

  > At the same time, –include-ids and –symbols will also be removed in Pylint 1.6.

- [Another reference to this deprecation](https://github.com/AtomLinter/linter-pylint/issues/60#issuecomment-134608167).

PR #4833, opened 2016-02-07
The Initial commit to the darks theme ;)

seif-allaya:
A new theme :)


mcornella:
Please notice that we don't accept any more themes. You can put it into a separate repository and add it to the [external themes wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes).

PR #4836, opened 2016-02-09
Update _brew to match latest

alias1:
Updated _brew to match latest in https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/brew_zsh_completion.zsh

### DIFF
--- a/plugins/brew/_brew
+++ b/plugins/brew/_brew
@@ -1,7 +1,7 @@
 #compdef brew
 #autoload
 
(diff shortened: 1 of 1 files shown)


mcornella:
Can you change the comment to use the commit sha instead of _master_? For instance, https://github.com/Homebrew/homebrew/blob/29f73d2212c2b202fe25f69dcbf440d8391fa4c9/Library/Contributions/brew_zsh_completion.zsh

alias1:
Done

PR #4837, opened 2016-02-09
Added git branch -vv to the alias list in git plugin

nagamalli9999:
git branch -vv is an invaluable tool to get information on the branches
and upstream


mcornella:
Closing due to lack of interest. You can also use `gb -vv` and people won't use it that often as to require an alias.

PR #4839, opened 2016-02-10
archlinux: fix the yasu alias

barraponto:
It seems like `--sucre` has always been something of an easter egg, it was deemed dangerous and dropped (see https://github.com/archlinuxfr/yaourt/pull/60). We can keep using `yasu` if we explicitly define the alias.

Not sure if this is a good idea, but the alternative would be to remove `yasu`: https://github.com/robbyrussell/oh-my-zsh/pull/4061


mcornella:
Fixed in 3b8d0ad9d0431bb76235f255458e43b2ab51b13b.

PR #4840, opened 2016-02-10
Add plugin for dynamic switch iTerm2 theme

idoo:
I want to dynamically change the theme in iterm2 depending on the time
of day


plugins/iterm2-dynamic-theme/iterm2-dynamic-theme.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+function set_iterm_profile() {
+  echo -e "\033]50;SetProfile=$1\a"
+}
+
+function start_dynamic_theme() {
+  HOUR=`date +%H`
```
apjanke: This variable should be `local`.

mcornella:
Has no interest.

PR #4841, opened 2016-02-10
git.plugin: remove the conflicting gvt alias

FiloSottile:
My Go vendoring tool https://github.com/FiloSottile/gvt is picking up some popularity, and many users have trouble initially because of the `git verify-tag` alias.

Now, this is a bit awkward.

I clearly am to blame here for not vetting the name, but the damage is done (I honestly didn't expect gvt to become popular...). Now, I honestly doubt `verify-tag` is that much used (or at all), so maybe you will consider scrapping the alias for the users convenience? :pray: :relaxed: 

See FiloSottile/gvt#19 and FiloSottile/gvt#4 for user reports.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -213,7 +213,5 @@ alias gup='git pull --rebase'
 alias gupv='git pull --rebase -v'
 alias glum='git pull upstream master'
 
-alias gvt='git verify-tag'
-
 alias gwch='git whatchanged -p --abbrev-commit --pretty=medium'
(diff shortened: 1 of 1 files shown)


mcornella:
I have no problem doing that, it's clear to me as well that the `verify-tag` alias is not that widely used. There's a non-zero chance that some power user is using it, but if that's the case they can figure out a `command not found` error.

Sooo your wish is granted :+1: 
Good luck on your OSS project!

FiloSottile:
:smiley:  Awesome :sparkles: Thank you!

PR #4842, opened 2016-02-11
Update autocompletion of homebrew cask.

pan-long:


mcornella:
Completion deleted in 84449fc8.

PR #4845, opened 2016-02-12
feat: add git difftool alias as mentioned in git cheatsheet alias list

patrickording:
feat: add git difftool alias as mentioned in git cheatsheet alias list
=> https://github.com/robbyrussell/oh-my-zsh/wiki/Cheatsheet#git

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -79,6 +79,7 @@ alias gd='git diff'
 alias gdca='git diff --cached'
 alias gdct='git describe --tags `git rev-list --tags --max-count=1`'
 alias gdt='git diff-tree --no-commit-id --name-only -r'
+alias gdtl='git difftool'
 gdv() { git diff -w "$@" | view - }
 compdef _git gdv=git-diff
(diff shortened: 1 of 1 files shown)


plugins/git/git.plugin.zsh
```diff
@@ -79,6 +79,7 @@ alias gd='git diff'
 alias gdca='git diff --cached'
 alias gdct='git describe --tags `git rev-list --tags --max-count=1`'
 alias gdt='git diff-tree --no-commit-id --name-only -r'
+alias gdtl='git difftool'
```
patrickording: Currently there are two identical aliases in the cheatsheet:

https://github.com/robbyrussell/oh-my-zsh/wiki/Cheatsheet#git

gdt => git diff-tree --no-commit-id --name-only -r
gdt => git difftool

mcornella:
> Currently there are two identical aliases in the cheatsheet:

This isn't true anymore, and `git difftool` isn't as widely used as the rest of git commands, so I'm closing this. Thanks for the interest though.

PR #4857, opened 2016-02-18
fix terminal color pollution

### LINKED ISSUE #4856: mvn plugins causes color pollution of terminal output
Enabling the `mvn` plugin causes color pollution for some commands, for example `env` or `gradle --debug` or any other command printing the exported environment variables of the plugin:

screen shot 2016-02-18 at 15 31 01

Removing the global scope of the variables fixes the problem as suggest by [dant3](https://gist.github.com/katta/1027800#gistcomment-949178).

depressiveRobot:
Fixes #4856

### DIFF
--- a/plugins/mvn/mvn.plugin.zsh
+++ b/plugins/mvn/mvn.plugin.zsh
@@ -1,24 +1,24 @@
 # mvn-color based on https://gist.github.com/1027800
-export BOLD=`tput bold`
-export UNDERLINE_ON=`tput smul`
-export UNDERLINE_OFF=`tput rmul`
-export TEXT_BLACK=`tput setaf 0`
-export TEXT_RED=`tput setaf 1`
-export TEXT_GREEN=`tput setaf 2`
-export TEXT_YELLOW=`tput setaf 3`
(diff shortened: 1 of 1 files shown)


apjanke:
Even though these are no longer exported, they're still global within the shell session, so you still have some risk of collision, and it can still mess up up your colors if you run `set` inside the shell session, like this.

screen shot 2016-02-21 at 1 49 22 am

The variables could be hidden using the `-H` attribute to `typeset`, which prevents their values from being included in the output of `set`, so you won't get the terminal codes output inappropriately.

This could be rewritten to use the standard [zsh `colors` variables](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#index-colors) to reduce the collision risk and not mess up `set` output. (Those variables are all hidden with `-H`.) Would save the cost of shelling out to those `tput`s on startup, too.

The zsh `colors` variables are available in OMZ by default, since they're [loaded in `lib/theme-and-appearance.zsh`](https://github.com/robbyrussell/oh-my-zsh/blob/31614f8c40787fda8d717f51bc7d97239782e82d/lib/theme-and-appearance.zsh#L2).

PR #4858, opened 2016-02-19
add tab command support for iTerm3

eiennohito:
iTerm3 has a bit different AppleScript syntax
https://www.iterm2.com/applescript.html

I have ported `tab` command for it.


mcornella:
@apjanke care to test this?

mcornella:
Wait, nevermind, seems #4907 is a better, more up-to-date solution. @eiennohito please +1 that other PR.

PR #4859, opened 2016-02-19
added caniuse to the web_search plugin

jensjns:


mcornella:
`caniuse` search can be found in [`frontend-search`](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/frontend-search) plugin.

PR #4863, opened 2016-02-20
Optimize the ys.zsh-theme make it more expressive

ysmood:
Remove most of the old hacks. Keep the theme clean and expressive.

Make the root user name highlight color more visible.

Add exit code indicator.

Add tty info.

ys-zsh-theme

### DIFF
--- a/themes/ys.zsh-theme
+++ b/themes/ys.zsh-theme
@@ -35,15 +35,16 @@ ys_hg_prompt_info() {
 	fi
 }
 
+local exit_code="%(?,%?,%{$fg[red]%}%?%{$reset_color%})"
 
 # Prompt format:
 #
-# PRIVILEGES USER @ MACHINE in DIRECTORY on git:BRANCH STATE [TIME] L:SHELL_LEVEL N:LINE_NUM
+# PRIVILEGES USER @ MACHINE in DIRECTORY on git:BRANCH STATE [TIME] tty:$TTY L:$SHELL_LEVEL N:LINE_NUM C:LAST_EXIT_CODE
 # $ COMMAND
 #
 # For example:
 #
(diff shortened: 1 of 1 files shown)


mhumeSF:
Anyone else feel this to be a bit noisy? I feel this kind of information is not needed for constant `ls` and `cd` commands. Maybe there could be a way to turn on in certain situations.

ysmood:
@mhumeSF Currently, I put the irrelevant information with grey color. The most important ones are red and yellow. I'm rethinking about the information density drawbacks.

P4ncake:
I agree with @mhumeSF. I like this theme, it's a fine job, but I feel this may be a bit too much

ysmood:
So how about remove the tty info? Something like below:

```
# user @ host in ~/Downloads on git:temp x [18:58:52] C:0
```

The `C: 0` is the last exit code.

elchinot7:
This is one of my favorites themes, please keep it simple. I agree with @mhumeSF .

ysmood:
I made a new PR #4940

zlalanne:
Could also add a check to only display return code when it's nonzero. So most of the time it's not on the screen

ysmood:
As I said:

> The C:127 is the last exit code, it appears only when the exit code is not zero.

I already implemented it.

ysmood:
@zlalanne As I said:

> The C:127 is the last exit code, it appears only when the exit code is not zero.

I already implemented it.

mhumeSF:
@ysmood Thank you! Favorite theme and I welcome the exit code status when not zero

elchinot7:
Thanks a lot @ysmood for this #4940 !

alexduros:
:+1: was feeling annoyed by this tty for few days :smile_cat:
thanks for the PR ! @ysmood

PR #4866, opened 2016-02-21
Fix bug in compass docs search

willmendesneto:
The plugin alias for `compass` is overriding default `compass` calls via CLI. Now `compassdoc` is used instead `compass`.

@robbyrussell Can you take a look in this PR, please?


mcornella:
:+1: Very exhaustive, thanks!

PR #4873, opened 2016-02-23
Aliases for git delete branch and git show

fmnoise:


plugins/git/git.plugin.zsh
```diff
@@ -46,6 +46,7 @@ alias gapa='git add --patch'
 
 alias gb='git branch'
 alias gba='git branch -a'
+alias gbd='git branch -D'
```
elazar: I'd prefer to see `gbd` map to `-d` and `gbdd` map to `-D`.

PR #4875, opened 2016-02-24
enable almostontop plugin

Valiev:
almostontop plugin description https://github.com/Valiev/almostontop/blob/master/README.md


mcornella:
Has no interest.

PR #4876, opened 2016-02-25
Update README.markdown

him229:
A sentence was wrong. Changed:

Most plugins (should! we're working on this) include a __README__, **which** documents how to use them.

to:

Most plugins (should! we're working on this) include a __README__, **with** documents how to use them.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -53,7 +53,7 @@ plugins=(git bundler osx rake ruby)
 
 #### Using Plugins
 
-Most plugins (should! we're working on this) include a __README__, which documents how to use them.
+Most plugins (should! we're working on this) include a __README__, with documents how to use them.
 
 ### Themes
 


mcornella:
The phrasing is right: a README documents how to use the plugins. What am I missing here?
cc @apjanke

apjanke:
Original phrasing is fine: there, "documents" is a transitive verb, "how to use them" is its object, and "README" is the antecedent of "which". The replacement looks like it's treating "documents" as a plural noun. Would need a preposition, too, like "with documents *about* how to use them". I think the original is clearer and scans better – it's clear that the README itself is supposed to document its plugin; the replacement could be read as asking for additional documents to accompany the README.

If anything, the single/plural agreement between plural "plugins"/'them" and singular "README" might be something to grumble about, but I think the meaning is clear as is.

English is hard.

mcornella:
Nice, I wanted a second opinion from a native speaker before rejecting this PR.

PR #4880, opened 2016-02-25
Improve git-prompt plugin

athimel:
3 improvements included :
- Display untracked files count (was « [some_branch|…] » ; now is : « [some_branch|…1] »)
- Display stash count (was « [some_branch|●2] » ; now is : « [some_branch|●2⚑1] »)
- Fix accentuated branch names support (was « <nothing> » ; now is : « [àccêntüé|✔] »)

"Stash" improvements comes from https://github.com/magicmonty/bash-git-prompt

### DIFF
--- a/plugins/git-prompt/gitstatus.py
+++ b/plugins/git-prompt/gitstatus.py
@@ -23,6 +23,18 @@ def get_tagname_or_hash():
         return hash_
     return None
 
+# Re-use method from https://github.com/magicmonty/bash-git-prompt to get stashs count
+def get_stash():
+    cmd = Popen(['git', 'rev-parse', '--git-dir'], stdout=PIPE, stderr=PIPE)
+    so, se = cmd.communicate()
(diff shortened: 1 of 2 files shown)


plugins/git-prompt/gitstatus.py
```diff
@@ -27,6 +28,18 @@ def get_tagname_or_hash():
         return hash_
     return None
 
+# Re-use method from https://github.com/magicmonty/bash-git-prompt to get stashs count
+def get_stash():
+    cmd = Popen(['git', 'rev-parse', '--git-dir'], stdout=PIPE, stderr=PIPE)
+    so, se = cmd.communicate()
+    stash_file = '%s%s' % (so.decode('utf-8').rstrip(), '/logs/refs/stash')
```
Obbi89: You cannot remove `decode('utf-8')` below and add it here
If decoding below breaks the output, maybe re-encoding it before printing is the right solution.

Marcholio:
This could maybe be closed due to inactivity?

53jk1 (approved):

PR #4881, opened 2016-02-26
Add autocomplete for vagga

webknjaz:
Hello,

I've written some autocomplete rules for https://github.com/tailhook/vagga
It would be nice to merge this to upstream

Have a nice day :)

### DIFF
--- a/plugins/vagga/_vagga
+++ b/plugins/vagga/_vagga
@@ -1,0 +1,26 @@
+#compdef vagga
+#autoload
+
+# This implements autocomplete for vagga daemonless containerization tool
+# src: https://github.com/tailhook/vagga
+# doc: https://vagga.readthedocs.org/
+
+_get_vagga_commands() {
+  vagga 2>&1 | sed '0,/Available commands:/d' | sed 's/^\s*//'
+}
+
+local -a _vagga_commands
+
+_vagga_commands=($(_get_vagga_commands))
+
+_arguments \
+  '(-h --help)'{-h,--help}'[show this help message and exit]' \
+  '(-V --version)'{-V,--version}'[Show vagga version and exit]' \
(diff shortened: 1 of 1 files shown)


webknjaz:
@tailhook
Paul this is PR with zsh completion rules for vagga. Plz take a look :)
If there's anything I missed — let me know.

rrader:
Nice. Maybe it's possible to utilize the vagga-completion.sh script that comes with vagga instead of parsing the vagga's output?

tailhook:
Yes, it's possible to utilize `vagga-completion.sh` (`bashcompinit`). But it's not the goal. The goal is probably to utilize the vagga's internal completion command that `vagga-completion.sh` uses.

The good thing about zsh-completion is that it allows to add descriptions to every entry in completion, so we would probably want to extend `vagga _compgen` command to have descriptions too.

/cc @anti-social

webknjaz:
Alright, I'll look into it as well. `vagga-completion.sh` doesn't work for me. Nevertheless, I'm going to check whether I could call internal `vagga`'s autocomplete API. Thanks @tailhook :)

tailhook:
Well, have you tried this?
```
autoload bashcompinit
bashcompinit
source /wherever/vagga/is/vagga-completion.sh
```

tailhook:
Okay, here is more clear completion for zsh. If you add the following to `.zshrc` completion work well (without option descriptions yet):
```
_vagga() {
    compadd -- $(vagga _compgen ${words[2,-2]} -- $PREFIX)
}
compdef _vagga vagga
```
It should be easy to convert it to a `#autoload` script, but I'm not sure exact syntax.

Also, it would be nice if someone check if I pass arguments to `vagga _compgen` correctly

robbyrussell:
⌛️ Am closing this for now as we'll need a README included. Happy to reconsider in the future.

PR #4882, opened 2016-02-26
Add "ghh" alias for command "git rev-parse HEAD"

mstred:
This means something like "git hash head", allowing the user to get the SHA-1 hash for the HEAD commit.


mstred:
I'll add it to the [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugin:git) as soon as it's merged.

PR #4884, opened 2016-02-26
add gc- alias for git checkout -

mb-dev:
git checkout - returns to the previous branch. I added an alias gc-. I was also considering gco-, but there are some checkout aliases without o, such as gcb.

Exampe:

```
> git checkout -b new-branch
Switched to new branch 'new-branch'
> git checkout -
Switched to branch 'master'
> gc-
Switched to branch 'new-branch'
```


mcornella:
Starting the alias with `gc-` would confuse it with `git commit` which is `gc`; it should be `gco-`. That said, it wouldn't save enough characters to be useful; `gco -` is right there on the keyboard.

PR #4886, opened 2016-02-27
Fix homebrew aws slow initialization time

hacfi:
Saves ~300ms initialization time

After activating the aws plugin my terminal opened noticeably slower and tracked it down to the two brew calls (`list` & `prefix`) which take ~150ms each. I don’t think a lot of people install homebrew in a different directory so this change doesn’t have a big disadvantage.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -1,11 +1,3 @@
-_homebrew-installed() {
-  type brew &> /dev/null
-}
-
-_awscli-homebrew-installed() {
-  brew list awscli &> /dev/null
-}
-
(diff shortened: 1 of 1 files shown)


kenden:
For me that change saves more than 2 seconds of execution.
It feels wrong to hardcode the path, but at least oh-my-zsh is usable.

hacfi:
We can also do one `$(brew --prefix awscli)` but personally I don’t want any unnecessary delay.

kenden:
I tried, but the difference in time was big

kenden:
This should be fixed by #6080

mcornella:
Indeed, #6080 fixed this.

PR #4887, opened 2016-02-27
Create npm completion on first run

hacfi:
I admit it looks a bit hacky but running `npm completion` takes ~280ms and really slows down the initialization time.

What do you think? Any nicer way of doing this?

By the way: Any reason `compinit` is run before all plugins get loaded? Would allow creating `_PLUGINNAME` files on the first run of a plugin.

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,4 +1,12 @@
-eval "$(npm completion 2>/dev/null)"
+__NPM_COMPLETION_DIR="${0:A:h}"
+__NPM_COMPLETION_FILE="${__NPM_COMPLETION_DIR}/npm_completion"
+
+if [[ ! -f $__NPM_COMPLETION_FILE ]]; then
+    npm completion > $__NPM_COMPLETION_FILE || rm -f $__NPM_COMPLETION_FILE
+    compinit -i -d "${ZSH_COMPDUMP}"
+fi
+
+source $__NPM_COMPLETION_FILE
 
 # Install dependencies globally
 alias npmg="npm i -g "
--- a/plugins/npm/.gitignore
+++ b/plugins/npm/.gitignore
@@ -1,0 +1,1 @@
+npm_completion


plugins/npm/npm.plugin.zsh
```diff
@@ -1,4 +1,12 @@
-eval "$(npm completion 2>/dev/null)"
+local __NPM_COMPLETION_DIR="${0:A:h}"
```
sachin21: Yuo have to use the `local` command in function.
This case, `local` command is not necessary. you can remove it.

hacfi:
Thanks @sachin21 - fixed!

wzrdtales:
Have stumbled over this and thought lets look if someone already did this before I do lay hands on it, thanks for sharing @hacfi ;)

wzrdtales:
@hacfi One addition, it would be beneficial to recreate the completion file, lets say every 30 days as it might change over time.

mcornella:
I like this idea but the current implementation doesn't convince me. Let me work on an alternative PR, I'll let you know when it's finished.

wzrdtales:
@mcornella Any update on this? I do mostly postpone updates of oh my zsh if it asks for them, just b/c I don't have the time to update the npm part again. The current implementation included in `oh my zsh` is actually horrible as it adds half a second boot up time, which is not acceptable. Thus this change is a must. To ask: What exactly do you plan to do different from this PR? And would it be an option to merge this PR and update oh my zsh later with your preferred option?

mcornella:
Sorry for the delay. This is my alternative:
```patch
diff --git a/plugins/npm/npm.plugin.zsh b/plugins/npm/npm.plugin.zsh
index 422d81c..ec595e7 100644
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,4 +1,10 @@
-eval "$(npm completion 2>/dev/null)"
+__NPM_COMPLETION_FILE="${ZSH_CACHE_DIR}/npm_completion"
+
+if [[ ! -f $__NPM_COMPLETION_FILE ]]; then
+    npm completion >! $__NPM_COMPLETION_FILE || rm -f $__NPM_COMPLETION_FILE
+fi
+
+source $__NPM_COMPLETION_FILE
 
 # Install dependencies globally
 alias npmg="npm i -g "
```

`$ZSH_CACHE_DIR` is `$ZSH/cache`. This is the preferred location to put stuff that is temporary and it is already git-ignored.

The `compinit` call wasn't necessary.

Another thing would be to add a way to regenerate that based on some predetermined time or when npm changes version, but for now the completion code is pretty much the same always so there isn't a hurry.

Let me know if you agree and I'll merge that.

wzrdtales:
Looks good for me and agreeing, replacing the file after a certain amount of time is not needed right now also it could be easily solved with just using `find`.

mcornella:
Fixed in 7c1ca0e. Thanks!

PR #4895, opened 2016-03-03
added plugin for Kirby CLI

mcbrwr:
Kirby is an easy to use file based CMS (https://github.com/getkirby). This plugin adds basic autocompletion and aliases for the Kirby CLI (https://github.com/getkirby/cli)


mcornella:
Has no interest.

PR #4904, opened 2016-03-07
New theme and new alias for nmap

Ciacho:
New nmap alias nmap_ping_scan to scanning subnets over ICMP
New theme - Ciacho


Ciacho:
Update theme Ciacho

robbyrussell:
Thanks, at this point in time... we're holding off on additional themes (unless it's a huge variation from existing ones).

Ciacho:
Yes, i see. but i also change nmap plugin. :)

PR #4906, opened 2016-03-07
Add my new theme

suissa:
My new beautiful theme ><


robbyrussell:
Thanks, at this point in time... we're holding off on additional themes (unless it's a huge variation from existing ones). 

https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes

PR #4907, opened 2016-03-08
Add iTerm2 v3 support

johankj:
[iTerm2 v3](https://www.iterm2.com/version3.html) is currently in beta and it doesn’t have a backward compatible [applescript syntax](https://www.iterm2.com/applescript.html).

This commit adds support for it in the `osx` plugin.


mcornella:
Merging, as there are 4 +1 already. Thanks!

PR #4908, opened 2016-03-09
Add ability to have multiple auto-connected sessions

AidenRHall:
As of now, if you turn on autoconnect and open a new terminal, it will connect you back to your an existing session, even if it's currently attached to by another terminal. This changeset changes the aforementioned behavior so that autoconnect sessions have special names. We use these like so: first we examine existing unattached sessions matching the autoconnect naming scheme, if we find one attach to it, else create a new one.

I don't see why anyone would want two terminal sessions with the same tmux session by default, so I didn't make it configurable. Maybe I'm horribly wrong via oversight, if so I would appreciate it if you let me know which use case I've overlooked. I want this badly and didn't see anything like it with the tmux tag in the issue tracker.


larson-carter:
@AidenRHall Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

robbyrussell:
Doing some repo cleaning. Not heard back so going to close this for now.

PR #4912, opened 2016-03-11
Hide git_prompt_status when hide-status is set

Neos3452:
On very large repos executing `git status` can take really long time. Setting hide-status had effect only on `git_prompt_info()` but not on `git_prompt_status()`. This makes `git_prompt_status()`check for hide-status as well.


mcornella:
Thanks!

PR #4913, opened 2016-03-12
Adding new Theme & missed correction in README

FKlama:
Based on robbytussell theme.
I mainly added the hostname so I can quickly tell on which host I am,
and use the symbols `#`/`%` depending if the user is root or not. The color of the hostname also changes from yellow to magenta for root.

Also added a missing change from `raw.github.com` to `raw.githubusercontent.com`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -104,7 +104,7 @@ The default location is `~/.oh-my-zsh` (hidden in your home directory)
 If you'd like to change the install directory with the `ZSH` environment variable, either by running `export ZSH=/your/path` before installing, or by setting it before the end of the install pipeline like this:
 
 ```shell
-export ZSH="$HOME/.dotfiles/oh-my-zsh"; sh -c "$(curl -fsSL https://raw.github.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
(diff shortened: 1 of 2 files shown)


README.markdown
```diff
@@ -104,7 +104,7 @@ The default location is `~/.oh-my-zsh` (hidden in your home directory)
 If you'd like to change the install directory with the `ZSH` environment variable, either by running `export ZSH=/your/path` before installing, or by setting it before the end of the install pipeline like this:
 
 ```shell
-export ZSH="$HOME/.dotfiles/oh-my-zsh"; sh -c "$(curl -fsSL https://raw.github.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
+export ZSH="$HOME/.dotfiles/oh-my-zsh"; sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
```
FKlama: Correction of README file missed this URL

mcornella:
[We can't accept more themes](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes). I'll merge this without the theme, though.

mcornella:
Nevermind, it was easier changing the README directly. Thanks for reporting that!

PR #4914, opened 2016-03-12
`gl` should do a `git log`

epegzz:
before `gl` was aliased to `git push`, which probably is not what one expects

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -145,7 +145,7 @@ compdef _git gk='gitk'
 alias gke='\gitk --all $(git log -g --pretty=format:%h)'
 compdef _git gke='gitk'
 
-alias gl='git pull'
+alias gl='git log'
 alias glg='git log --stat --color'
 alias glgp='git log --stat --color -p'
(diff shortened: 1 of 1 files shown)


plugins/git/git.plugin.zsh
```diff
@@ -145,7 +145,7 @@ compdef _git gk='gitk'
 alias gke='\gitk --all $(git log -g --pretty=format:%h)'
 compdef _git gke='gitk'
 
-alias gl='git pull'
```
sachin21: I am using `gl` command. plz add a new alias for `git pull` or please consider other solution.

thegeorgeous:
@epegzz, a lot of people may be using the shortcut. Changing it would very disconcerting

mcornella:
We're not considering changing that for the moment. OMZ has no method for safely removing or changing aliases as of yet, and we've had trouble in the past for doing so hastily, so I prefer waiting on any alias changes until something like #4585 is in place.

You should add your custom alias to your zshrc file, meanwhile.

abitrolly:
@thegeorgeous is there a way to gather usage stats on that?

PR #4918, opened 2016-03-14
Updates git tag sorting alias

FeliciousX:
Uses git built in sorting by version for tags


mcornella:
Git tag sorting is only [supported from 2.0 to newer versions](http://stackoverflow.com/a/22634649/5798232). The `sort -V` command is similarly flawed in that certain versions of `sort` don't have such a flag (OSX, for instance). 

So I don't know how to proceed here. Possibly `git 2.0` is more widely used than `sort -V` versions are, but I'd like to think about it.

FeliciousX:
have u thought about it? xD Else I'll just close this PR @mcornella

PR #4920, opened 2016-03-14
Add ASCII art theme

tcya:
Hi, I make an ASCII art theme. Hope you find it funny. Thank you.


mcornella:
We have lots of themes already, [please don't send us more themes](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes).

PR #4924, opened 2016-03-16
Remove signoff flag

zackhsi:
There should not be a signoff flag for this alias.

`gcan!`:
```
- g: git
- c: commit
- a: all
- n: no edit
- !: amend
```

No where in there is there a signoff flag.


mcornella:
Duplicate of #4830.

PR #4925, opened 2016-03-16
Remove undesirable hardcoding of PATH into zshrc

fvgs:
See b67961d for why hardcoding the PATH into zshrc is a bad idea and why it's better to allow the user to make those customizations as needed.

### DIFF
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -53,7 +53,7 @@ plugins=(git)
 
 # User configuration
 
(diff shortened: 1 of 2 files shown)


fvargas:
@robbyrussell @mcornella any update on this?

brthor:
👍

pjvds:
Any reason why this one is not pulled yet? Is there anything we can do to speed this up?

mcornella:
Sorry for the delay. I was working on expanding this to fix other installer issues (#5169).

PR #4938, opened 2016-03-23
add tab completions to readme

manu09:


mcornella:
It's cleaner if we add this as a second paragraph, not as a separate section. Something like:

```diff
WP-CLI is a set of command-line tools for managing WordPress installations. You can update plugins, set up multisite installs and much more, without using a web browser.

+This plugin adds [tab completion](http://wp-cli.org/#complete) for `wp-cli` as well as several aliases.
```

There's not enough content to warrant a separate section. What do you think?

PR #4943, opened 2016-03-25
Update 'tm' function to take multiple files on the command-line

MattLewin:
It seems counter-intuitive to me that `tm file1 file2` or `tm foo.*` only opens the first file. Thus, I took it upon myself to update `textmate.plugin.zsh` to open multiple files if specified. I kept all the existing functionality the same. In particular, if the first arg is a directory, the function ignores everything after it and opens the directory in TextMate.


mcornella:
:+1:

PR #4949, opened 2016-03-27
Add a very simple plugin to call issue Dash queries from the command-line

MattLewin:
Description pretty much says it all.


tresni:
Checkout #2557

mcornella:
Merged #2557 instead.

PR #4953, opened 2016-03-29
Adds support for virtualenv for the bira theme

bingorabbit:
It's a simple edit just to support the current running virtualenv based on the virtualenv plugin.


mcornella:
This removes the rvm prompt, please add the `virtualenv_prompt_info` function into the `$PROMPT` variable instead.

PR #4954, opened 2016-03-29
Update bira theme

jocelynthode:
I saw the latest update that changed the color of the name when you're in root. I thought it would be neat to have the symbol also change from '$' to '#' as a root prompt is always represented with '#'.

I also put  the return code in bold as I think it is not visible enough when not bold.

### DIFF
--- a/themes/bira.zsh-theme
+++ b/themes/bira.zsh-theme
@@ -23,7 +23,7 @@ local git_branch='$(git_prompt_info)%{$reset_color%}'
 
 PROMPT="╭─${user_host} ${current_dir} ${rvm_ruby} ${git_branch}
 ╰─%B${user_symbol}%b "
-RPS1="${return_code}"
+RPS1="%B${return_code}%b"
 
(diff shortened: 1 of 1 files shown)


themes/bira.zsh-theme
```diff
@@ -20,8 +22,9 @@ fi
 local git_branch='$(git_prompt_info)%{$reset_color%}'
 
 PROMPT="╭─${user_host} ${current_dir} ${rvm_ruby} ${git_branch}
-╰─%B$%b "
-RPS1="${return_code}"
+╰─%B${user_symbol}%b "
```
mcornella: It's shorter using `%(!.#.$)` (see [prompt expansion](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Shell-state)). 

The previous `user_host` stuff could be similarly abbreviated with a `%(!.$fg[red].$fg[green])`, but let's leave that for another time.

jocelynthode:
@mcornella Any ETA on merging this change ?

Issue #4942, opened 2016-03-24
asdf: Extendable version manager


RobLoach:
Add a plugin for https://github.com/asdf-vm/asdf . Allows handling installations of node, ruby, lua, and more.

mcornella:
First things first, we should have a basic need for a plugin. Right now the only thing I can see is a pretty slim completion file. Do you have some other functionality in mind?

Secondly, we should get a few users wanting that. If not, it's not really worth it dedicating another plugin in the core repo. You can always put the completion file in a custom plugin and use it yourself. If you later find there's something to add to the OMZ repository I'll gladly consider adding the plugin.

Thanks!

PR #4956, opened 2016-03-30
Add asdf version manager plugin

RobLoach:
[asdf](https://github.com/asdf-vm/asdf) is an extendable version manager with support for Ruby, Node.js, Elixir, [and more](https://github.com/asdf-vm/asdf#usage).

```
asdf plugin-add nodejs git@github.com:asdf-vm/asdf-nodejs.git
asdf install nodejs 5.9.1
```

This plugin simply loads asdf if it finds it. asdf provides the zsh completion for us. The corresponding issue for this is https://github.com/robbyrussell/oh-my-zsh/issues/4942 .

### DIFF
--- a/plugins/asdf/README.md
+++ b/plugins/asdf/README.md
@@ -1,0 +1,27 @@
+## asdf
+
+**Maintainer:** [@RobLoach](https://github.com/RobLoach)
+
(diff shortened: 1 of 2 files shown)


mcornella:
:clap:

PR #4957, opened 2016-03-30
localization support for clock in candy theme

TorrentKatten:
Now uses 12-hour clock or 24-hour clock depending
on user locale


mcornella:
For reference, [here's the documentation](http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Date-and-time) on the prompt sequence used for formatting Date and Time. 

The insides of `%D{}` use [`strftime`-type sequences](http://linux.die.net/man/3/strftime), so using `%X` will show:
> **%X**
> The preferred time representation for the current locale without the date.

And here's the difference between the old and new format strings:
```zsh
$ print -P '%D{[%I:%M:%S]}'
[12:49:46]
$ print -P '%D{[%X]}'      
[00:49:47] 
```

In my opinion, this is good to go! :+1:

PR #4961, opened 2016-03-31
Add git user profile functions for prompt display

aesopwolf:
Add functions for `git config user.name` and `git config user.email`.

Useful as visual indicator for people who need to switch between different work and home user accounts often.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -188,6 +188,18 @@ function git_compare_version() {
   echo 0
 }
 
+# Outputs the name of the current user
+# Usage example: $(git_current_user_name)
+function git_current_user_name() {
+  command git config user.name 2>/dev/null
+}
+
+# Outputs the email of the current user
+# Usage example: $(git_current_user_email)
+function git_current_user_email() {
+  command git config user.email 2>/dev/null
+}
+
 # This is unlikely to change so make it all statically assigned
 POST_1_7_2_GIT=$(git_compare_version "1.7.2")
 # Clean up the namespace slightly by removing the checker function


mcornella:
You could team up with @dottorblaster over at #4673 and make a dedicated plugin for git user profiles.

aesopwolf:
@mcornella good idea! It's nice to see someone else with a similar problem.

My original plan was to:

1. Get this pull request landed and get my team at work using it
1. Then later I can build a git user switcher via [git-extras](https://github.com/tj/git-extras)

Unless you think a git user switcher should be built as an oh-my-zsh plugin?

mcornella:
I like `git-extras` very much, I use it all the time, so I don't have a problem with building it over there. Having these prompt functions in the git plugin is a little bit weird though, but I guess it's the best option we have. Let me quickly review the functions and we'll get this merged.

lib/git.zsh
```diff
@@ -188,6 +188,24 @@ function git_compare_version() {
   echo 0
 }
 
+# Outputs the name of the current user
+# Usage example: $(git_current_user_name)
+function git_current_user_name() {
+  local ref
+  ref=$(command git config user.name 2> /dev/null)
+  local ret=$?
```
mcornella: This `$ret` variable is not used anywhere else, it can be safely deleted.

lib/git.zsh
```diff
@@ -188,6 +188,24 @@ function git_compare_version() {
...
+  echo ${ref}
```
mcornella: If you're gonna just `echo` the variable, you can make the whole function be:
```sh
function .. {
  command git config user.name 2>/dev/null
}
```

aesopwolf:
@mcornella thanks for your comments on the code. As you can tell I didn't know exactly what I was doing and had just copy/pasted from the `git_current_branch` function

Anyways, I just updated the two functions to use the one-liner you recommended. Should be good to go now

mcornella:
That's better. Now one last thing, squash the current 4 commits into a single one, and then edit the commit message to be more descriptive (you can use this same PR title). 

If you know how to use rebase you'll know how to do it; otherwise, ask and I'll guide you through the steps.

aesopwolf:
@mcornella oh yeah good call on the rebase. That history was pretty ugly. Would you mind taking another look?

mcornella:
Perfect!

aesopwolf:
Sweeeeeet! Thanks so much :)

PR #4966, opened 2016-04-01
Fixed an incorrect yaourt alias

zack-emmert:
alias yasu='yaourt --sucre' --> alias yasu='yaourt -Syua --noconfirm'


mcornella:
Obsolete

PR #4968, opened 2016-04-02
Add Sudo option to npm -g

trb143:
All sudo to be used with NPM -g option.
Follows pattern for other plugins


warriorpostman:
This would be a useful alias. I'm speculating @trb143 you are desiring this for working on Ubuntu?

trb143:
No fedora but yes it makes life easier.

mcornella:
Using `sudo` with npm is bad and we shouldn't encourage it by making it easier. The `sudo` alias already allows calling aliases by adding a trailing space (`alias sudo='sudo '`), so you can still use plugin aliases with `sudo <alias>`.

PR #4971, opened 2016-04-03
Peepcode theme: Don't try to use RVM if not installed

HaraldNordgren:
I just installed oh-my-zsh and I immediately fell for the Peepcode theme. The only problem is that it keeps trying to access the Ruby Version Manager (rvm-prompt) which I don't have installed, giving the same error after each command:

`zsh: no such file or directory: /home/harald/.rvm/bin/rvm-prompt`

This small piece of code checks for the existence of the ~/.rvm folder, and if it's not there it just disregards that part of the prompt.

### DIFF
--- a/themes/peepcode.zsh-theme
+++ b/themes/peepcode.zsh-theme
@@ -41,4 +41,10 @@ PROMPT='
 %~
 ${smiley}  %{$reset_color%}'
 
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
I created a feature branch, so closing this pull request. Changes are in #4971 instead.

PR #2752, opened 2014-04-25
Fix checkmark for clean repo in bureau theme

isqua:
The checkmark in clean repo wasn’t dispalyed. Now it fixed.

PR #4976, opened 2016-04-05
Bureau theme fixes

mcornella:
This PR fixes a bug introduced in #2752 where a repository being ahead or behind the remote won't be displayed unless there is some change in the staging area or the working directory. 

It also uses `grep -q` instead of the construct `grep ... &>/dev/null`.

cc @mtbtiago @mtnbikenc

### DIFF
--- a/themes/bureau.zsh-theme
+++ b/themes/bureau.zsh-theme
@@ -22,10 +22,12 @@ bureau_git_branch () {
   echo "${ref#refs/heads/}"
 }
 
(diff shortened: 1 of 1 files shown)


mtnbikenc:
Tested PR locally :+1:

mcornella:
@mtbtiago also confirms this works. The *diverged* status doesn't yet work, but it didn't work before so I'll fix it in a future PR.

PR #4985, opened 2016-04-07
Use default branch on recently created Mercurial repository

victor-torres:
After `hg init` command, sometimes Mercurial does not create `.hg/branch` file so we'll use 'default' as fallback, which is the master branch in Mercurial repositories.


victor-torres:
@mcornella can you take a look, pls?

mcornella:
Sorry for the delay, thanks for the patch! :smile:

victor-torres:
Thank you, @mcornella

PR #4988, opened 2016-04-10
Fix invalid "ls -G" alias.

reduardo7:


xuhdev:
Looks like you need to rebase now...

mcornella:
@reduardo7 what does this solve? What's your setup? (platform, operating system, installed ls's...)

reduardo7:
@mcornella this solve a problem with ZSH in Termux (https://play.google.com/store/apps/details?id=com.termux). You can try install it and configure with ZSH + oh-my-zsh. When you type `ll`, results in an error.

red15:
Causes problems for other users #5604

lib/theme-and-appearance.zsh
```diff
@@ -25,7 +25,7 @@ if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
       (( $+commands[dircolors] )) && eval "$(dircolors -b)"
     fi
 
-    ls --color -d . &>/dev/null && alias ls='ls --color=tty' || alias ls='ls -G'
+    ls --color -d . &>/dev/null && alias ls='ls --color=tty' || ls -G . &>/dev/null && alias ls='ls -G'
```
croachrose: This should be:

`ls --color -d . &>/dev/null && alias ls='ls --color=tty' || (ls -G . &>/dev/null && alias ls='ls -G')`

Note the parenthesis around the second test. This should fix the various ls colors issue that people are seeing.

PR #4989, opened 2016-04-11
Adding git function to determine repository name

evanchiu:
I find this useful to just have the name of the repository with the git status.  This allows me to see my current working directory and the git repo, without showing the full path.
Prompt sample:

prompt


renannprado:
I also would like to have this

aping1:
you could do less forking with inline replacement

`function git_repo_name() {
  local repo_path repo_name
  if git rev-parse --is-inside-work-tree &>/dev/null; then
    repo_path=$(git rev-parse --git-dir 2>/dev/null)
    print -l -- "${${${repo_path:a:h}%/.git}:t}"
  fi
}
`

mcornella:
Better use `--show-toplevel` and use the fact that the git command errors out when not in a git repository to avoid one call:
```zsh
function git_repo_name {
  local repo_path
  if repo_path="$(git rev-parse --show-toplevel 2>/dev/null)"; then
    print ${repo_path:t}
  fi
}
```

PR #4990, opened 2016-04-11
Add plenv plugin

zakame:
This adds support for [tokuhirom/plenv](https://github.com/tokuhirom/plenv). Based mostly from pyenv plugin.


mcornella:
It doesn't seem to be a used enough project for it to warrant a plugin, and the installation instructions are clear on the two steps needed to use it: add it to the $PATH, and add the eval sentence to zshrc file.

PR #4993, opened 2016-04-11
Remove default for git reset

matjack1:
The command will do the same as before, but now you can also specify
a path.

Example:
```
grh branch-name
grhh tag-name
```


matjack1:
@ncanceill I still see the point of this. I've recently set up a new machine and encountered the same problem :)

What do you think?

matjack1:
Thank you!

PR #5004, opened 2016-04-14
Add dockerhub to web-search plugin.

slimatic:


mcornella:
I don't think we need to duplicate docker search functionality when [the `docker` command itself has this feature](https://docs.docker.com/engine/reference/commandline/search/).

PR #5006, opened 2016-04-14
npm install --save --save-exact

brugnara:
Added npm save exact.


mcornella:
With all due respect, I think that command isn't run frequently enough to warrant a new, separate alias. You can use the `npm install --save` alias `npmS` and append it the `-E` argument, or define the alias in your own zshrc file.

PR #5012, opened 2016-04-16
brew: remove duplicated and outdated brew completion script

xu-cheng:
Homebrew has began to ship its own official zsh completion script.
So we could remove the duplicated and outdated version from OMZ.


joshka:
Hi,
I didn't notice this PR when I created a dupe at https://github.com/robbyrussell/oh-my-zsh/pull/5190. It has been merged, so this PR can be closed

PR #5021, opened 2016-04-19
add support for randomly selecting from a list of specific themes

sparr:
As a new oh-my-zsh user I set my theme to random to try out a bunch of themes. I found a few that I like, and now I want to just have it pick between those until I decide which one to stick with. This patch allows ZSH_THEME to optionally be an array instead of a string.


mcornella:
For future users who see this, now there's `$ZSH_THEME_RANDOM_CANDIDATES` for this.

PR #5025, opened 2016-04-20
Added --force-with-lease method to git plugin

jchannon:
This PR adds the ability to use `ggfl` to force push but with the option of `--force-with-lease` so you don't overwrite someone else's code.


joyarzun:
@ncanceill @mcornella Can you please review?. We will love to have this

PR #5033, opened 2016-04-22
znt: config option for instant selection when in search mode

psprint:
This is against:

https://github.com/psprint/zsh-navigation-tools/issues/14

For new users, following block will exist in n-history.conf:

```zsh
# When 1, pressing enter when in search mode will not
# only leave the mode, but also do selection and leave
# n-history
local NLIST_INSTANT_SELECT=0
```

Existing users need to add it to ~/.config/znt/n-history.conf. And change the variable to "1" if instant selection is desired.

The instant selection is: when in search mode, pressing enter leaves whole tool (e.g. n-history) selecting current element – instead of leaving only search mode. So one enter less is needed to do selection in typical Ctrl-R invocation of n-history.

### DIFF
--- a/plugins/zsh-navigation-tools/zsh-navigation-tools.plugin.zsh
+++ b/plugins/zsh-navigation-tools/zsh-navigation-tools.plugin.zsh
@@ -23,6 +23,9 @@ for i; do
     fi
 done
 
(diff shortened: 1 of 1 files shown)


doubleloop:
Tested, works for me

PR #5041, opened 2016-04-25
Add plugin: dasht (issue: 5040)

tmerse:
Gets list of installed _docsets_ via `dasht-docsets 2>/dev/null` and provides them as autocompletion for second, third, etc. arguments.

### DIFF
--- a/plugins/dasht/_dasht
+++ b/plugins/dasht/_dasht
@@ -1,0 +1,14 @@
+#compdef dasht
+_main() {
+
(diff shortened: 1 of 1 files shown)


sunaku:
More powerful ZSH completions are included in dasht version 2.1.0:
https://github.com/sunaku/dasht/tree/master/etc/zsh/completions

mcornella:
Not enough demand.

PR #5044, opened 2016-04-26
JIRA Plugin will open rapidboard

charpeni:
JIRA Plugin will open rapidboard instead of dashboard if JIRA_RAPID_BOARD is true.


robbyrussell:
Ooh, nice!

PR #5048, opened 2016-04-28
Some helpful aliases for git and vagrant

josefglatz:


mcornella:
zsh already corrects a typo:
```
$ gti
zsh: correct 'gti' to 'git' [nyae]?
```

If you want the vagrant aliases, you'll have to change this PR or submit a new one. You can have a look at other open PRs for vagrant aliases as well.

PR #5050, opened 2016-04-28
[symfony2 plugin] Fix deprecated usage

chalasr:
When using the command alias `sfcontainer` or `sfrouter`, the following warning occurs:

> ! [CAUTION] The use of "container:debug" command is deprecated since version 2.7 and will be removed in 3.0. Use the "debug:container" instead.


tjaari:
👍

PR #5051, opened 2016-04-29
Add `cda` alias for composer dump-autoload

JayBizzle:
This catches me out all the time. Surely the alias should be `cda`, not `cdu`

I've left `cdu` in there for people that are used to typing that now.


Anahkiasen:
While we're updating composer aliases, could `co` and `cod` be added for `composer outdated` and `composer outdated --direct`?

mcornella:
It's just a matter of training. If you're really bothered by it, I suggest adding `alias cda=cdu` in your zshrc file.

PR #5055, opened 2016-04-30
osx plugin: add partial support for iTerm2

krzysztof-magosa:
It looks that iTerm2 can be controlled using the same osascript commands.


jnozsc:
it may be duplicate with https://github.com/robbyrussell/oh-my-zsh/pull/4907

krzysztof-magosa:
You are right, closing then.

PR #5058, opened 2016-05-02
[Git plugin] Add/change aliases:

zheeeng:
- gbav: list all branches and their upstream branches
- gcam!, gcmsg!: feed amend option
- gclean!: clean operation should be used with caution
- grp: prune remote branches which don't exist anymore
- grao, grpo: add/prune remote repository—origin


mcornella:
> gbav: list all branches and their upstream branches

Same as `gba -vv`: short and not so much as a pain to warrant a new alias.

> gcam!, gcmsg!: feed amend option

I'd rather amend operations to show what is going to be committed, like in `gc!` and `gca!` that use `-v` (verbose).

> gclean!: clean operation should be used with caution

gpristine should be amended in the same way, but I think we can trust users to understand what clean and pristine do.

> grp: prune remote branches which don't exist anymore

Using `gfa` already prunes remotes.

> grao, grpo: add/prune remote repository—origin

Same as above. Also, you can use `gra` to add a remote.

For all these reasons, I'm closing this PR. Thanks for your work

PR #5060, opened 2016-05-02
Update docker completion to match docker upstream

cpuguy83:
See
https://github.com/docker/docker/blob/master/contrib/completion/zsh/_docker

### DIFF
--- a/plugins/docker/_docker
+++ b/plugins/docker/_docker
@@ -138,7 +138,7 @@ __docker_containers_names() {
     __docker_get_containers all names "$@"
 }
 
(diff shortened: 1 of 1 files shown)


cpuguy83:
I'll update this with the latest from upstream.
This is quite outdated now.

cpuguy83:
Excellent, looks like this was pulled in already.
PR will just update to docker 1.12.1 completion.

mcornella:
Thanks!

leifcr:
@mcornella Thanks!

PR #5061, opened 2016-05-02
Rewrite git-prompt plugin in ZSH instead of Python.

pawel-slowik:
Makes the plugin faster and eliminates dependency on Python.

Also fixes bugs (but won't add new features) referenced in pull requests:
#4991

_Improve the regex used by git-prompt to extract tagname_ - fixed, because tag extraction is done by splitting on ", " (comma + space), not with a regexp
#4880

_Fix accentuated branch names support_ - fixed, because git output no longer goes through python's decoding / encoding. Does not add new features from this PR:
- Display untracked files count
- Display stash count

Those are relatively easy to code. I could add it to this PR, but I think adding new features (vs fixing bugs) calls for a separate PR.


Obbi89:
Why did you close this?

pawel-slowik:
Because I no longer think it is worth merging.
The plugin is essentially abandoned at this point (unused and unmaintained).
It can just about be replaced with vcs_info.
I have stopped using it months ago (along with the rest of Oh My Zsh), so I wouldn't be able to properly test / rebase / fix bugs / answer questions etc.

PR #5070, opened 2016-05-07
Add a theme

andrewei1316:
I added a theme named andrewei.


mcornella:
[We can't accept more themes at the moment](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page.

PR #5072, opened 2016-05-07
Update vault plugin - add "list" command

prysmakou:
Add "list" command for vault.


mcornella:
The `vault` plugin is now deprecated since it uses old-version commands.

PR #5077, opened 2016-05-10
Added fossil plugin.

jgmdev:
This plugin adds completion support and prompt for fossil repositories.
The prompt will display the current branch and status been dirty or clean.

For those that don't know what fossil is: http://fossil-scm.org/


robbyrussell:
Thanks @jgmdev

PR #5080, opened 2016-05-11
Fix -n option argument error

### LINKED ISSUE #5079: gunwip throws "error: -n requires an argument"
image

The current alias is `gunwip='git log -n 1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'`

It should be `gunwip='git log -n1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'`

LM450N:
By removing the whitespace between the "-n" option and the argument
value "1", this commit fixes the issue of the "gunwip" alias.

fix #5079

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -208,7 +208,7 @@ alias gts='git tag -s'
 alias gtv='git tag | sort -V'
 
 alias gunignore='git update-index --no-assume-unchanged'
-alias gunwip='git log -n 1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'
+alias gunwip='git log -n1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'
 alias gup='git pull --rebase'
 alias gupv='git pull --rebase -v'
 alias glum='git pull upstream master'


sachin21:
I can't reproduce argument error with `git log -n 1`. What is your git version?(FYI: `git --version`)
btw my git version is `2.7.0`.

LM450N:
I was using git 2.7.4. This morning I upgraded to git 2.8.2 and the issue is still here. I also use [scm_breeze](https://github.com/ndbroadbent/scm_breeze) and a colleague with a similar configuration has the same error.

_OSX EC 10.11.4, zsh 5.2 (x86_64-apple-darwin15.0.0)_

mcornella:
Are you sure you're not using a wrapper for git, like [hub](https://github.com/github/hub) for instance? Run `which -a git` and post the output.

LM450N:
Yes, I'm also using hub. Do you think they are conflictual ? I'll try without it when I'll have more time. I keep you in touch. Thank you.

mcornella:
Please run the above command (`which -a git`) and also try `command git log -n 1`. I have no idea how `scm_breeze` wraps the git command but these commands above should tell the whole story.

mcornella:
Wait, nevermind, [here's how `scm_breeze` calls `git log`](https://github.com/ndbroadbent/scm_breeze/blob/b6531e8379122675168c053c7b7111746b4891c6/lib/git/aliases.sh#L26-L30).

So my guess is that `command git log -n 1` will work correctly, therefore the issue you're having is from `scm_breeze`. Please confirm.

LM450N:
`git log -n 1` : nothing

`git log -n1` and `git log -1`
image

`gl -n 1` : nothing

`gl -1` and `gl -n1`
image

mcornella:
Don't forget the `command` before it, it's important.

LM450N:
i don't understand how see the options could help but why not...

`git -a`

image

mcornella:
No, run "`command git log -n 1`". The whole thing, don't leave anything.

LM450N:
Sorry for my misunderstanding. It's working.

image

mcornella:
Ok, thanks. That confirms that there's an issue with scm_breeze, so you should report it there. In the meantime, you can put your custom `gunwip` alias at the end of your zshrc file so that you can keep using it with scm_breeze:
```
alias gunwip='command git log -n 1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'
```

LM450N:
Ok, thank you. Could you explain me what "command" does please ?

mcornella:
Sure. `command` is a [zsh built-in command](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html#index-command-1) that ignores aliases and functions and runs the actual executable that you pass it.

So `command git` will ignore any aliases and functions named "git" and run instead the git executable. Without `command`, the alias or function would be run instead.

A useful trick to know if there are aliases or functions is `which -a <name>`, so `which -a git` in my case returns:
```
$ which -a git
git: aliased to noglob hub
/usr/bin/git
```
because I'm using [hub](https://github.com/github/hub).

I hope that's useful.

LM450N:
Really useful! Thank you for your help.

PR #5081, opened 2016-05-11
Add alias for switching to previous branch ('git checkout -')

pvdb:
@robbyrussell / @mcornella - I've added a small convenience alias to the `git` plugin!

It is based on the alias suggested in [this article](https://marcgg.com/blog/2015/10/18/git-checkout-minus/):

> `alias g-='git checkout -'`

... but I decided to use the original syntax, to ensure compatibility with older versions of `git`:

> `alias g-='git checkout @{-1}'`

Hope you like!  :smile:

@pvdb

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -40,6 +40,8 @@ function work_in_progress() {
 
 alias g='git'
 
+alias g-='git checkout @{-1}' # ie. 'git checkout -'
+
 alias ga='git add'
 alias gaa='git add --all'
 alias gapa='git add --patch'


pvdb:
Poke @robbyrussell / @mcornella  :smile:

pvdb:
Poke @robbyrussell / @mcornella  :smile:

pvdb:
Poke @robbyrussell / @mcornella  :smile:

mcornella:
Not useful enough. You can just do `gco -`.

pvdb:
Hey @mcornella,

Thanks for reviewing my PR, and for your feedback...

> You can just do `gco -`

Agreed, but my understanding of the rationale and reasoning behind this particular plugin is to safe keystrokes when issuing `git` commands.

`g-` is just 2 keystrokes, whereas `gco -` is 5, more than double, hence me assumption it would be a nice addition to the plugin.

mcornella:
You can add it to your zshrc file. At some point adding more aliases isn't a good solution. Thanks

mcornella:
I stand by my first assessment.

> Not useful enough. You can just do `gco -`.

pvdb:
> Not useful enough. You can just do `gco -`.

Fair enough, but in that case...

```
alias glg='git log --stat'
alias glgp='git log --stat -p'
```

Shall I issue a PR to remove [the `glgp` alias](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh#L181), as you can just do `glg -p`, and/or add the alias to your `zshrc` file? :wink:

PR #5083, opened 2016-05-11
Add ability to specify tmux executable path in tmux plugin

strifey:
Have two versions of tmux installed. Don't want to rely on the version that PATH resolves to.


mcornella:
I think an alias in your zshrc file would do the same job:
```
alias tmux=/usr/bin/tmux
```

larson-carter:
@strifey Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

mcornella:
I don't think we'll merge this. For the looks of it this is a fringe use case. Make your case if you don't agree.

PR #5093, opened 2016-05-14
add event sink for custom folders

mscalora:
For those of us not lucky enough to get our (custom) theme included in the repo, keeping the theme up-to-date across machines/environments is a painful process. I've used several solutions over the years but would like to see some real support from oh-my-zsh. This PR is it pretty lightweight way to give a user's custom folder notification of when "upgrades" happen for the base oh-my-zsh.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -34,6 +34,12 @@ then
   printf "${BLUE}%s\n" "Hooray! Oh My Zsh has been updated and/or is at the current version."
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on twitter: https://twitter.com/ohmyzsh"
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "Get your Oh My Zsh swag at:  http://shop.planetargon.com/"
+
+  if [ -f $ZSH_CUSTOM/tools/custom_event.sh ]; then
+    # Notify user's custom code of upgrades
+    env ZSH=$ZSH ZSH_CUSTOM=$ZSH_CUSTOM /bin/sh $ZSH_CUSTOM/tools/custom_event.sh upgrade
+  fi
+
 else
   printf "${RED}%s${NORMAL}\n" 'There was an error updating. Try again later?'
 fi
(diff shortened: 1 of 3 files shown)


tools/upgrade.sh
```diff
@@ -34,6 +34,12 @@ then
   printf "${BLUE}%s\n" "Hooray! Oh My Zsh has been updated and/or is at the current version."
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on twitter: https://twitter.com/ohmyzsh"
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "Get your Oh My Zsh swag at:  http://shop.planetargon.com/"
+
+  if [ -f $ZSH_CUSTOM/tools/custom_event.sh ]; then
+    # Notify user's custom code of upgrades
+    env ZSH=$ZSH ZSH_CUSTOM=$ZSH_CUSTOM /bin/sh $ZSH_CUSTOM/tools/custom_event.sh upgrade
```
mcornella: Where is this `tools/custom_event.sh` file and what does it do?

mscalora: It is where the customizer provides a script to consume events. For the **upgrade** event with a github based customization they would have some code to pull the latest changes. For example, my customizations live at: https://github.com/mscalora/custom-zsh so I would have a script that would look somthing like (just an example):

    #!/bin/sh
    pushd $ZSH_CUSTOM
    if git pull --rebase --stat origin master ; then
        echo "Updated custom folder"
    else
        echo "Error: couldn't update custom folder"
    fi
    popd

mcornella:
Hi @mscalora, sorry for the late response. We're not ready to complicate the OMZ updating algorithm just yet, but we are ready to make it simpler and less hackish. Currently it's spread out across the init script, the `lib/functions.zsh` and 2 `tools` scripts, so there is room for improvement. Once we've simplified this perhaps it will be easier to add a hook there, so don't despair!

In the meantime, you can keep your custom modifications commited and when OMZ updates it should keep them at the top. I apologize again for the inconvenience.

PR #5094, opened 2016-05-14
Add rbenv support for ruby version display in the Avit theme

pzgull:
The function _ruby_version() which is called in prompt only checks for RVM. I just added rbenv support. It works as expected.


robbyrussell:
Thanks, pzgull!

PR #5095, opened 2016-05-15
linuxbrew plugin added

typekpb:
adds linuxbrew stuff to path


mcornella:
We're generally weary of plugins that mess with the `$PATH`. You should add that to your zshrc file instead.

PR #5099, opened 2016-05-18
Add list-timers to systemd user commands

wengole:
This is useful for listing "cron like" systemd jobs


robbyrussell:

Cheers,
Robby

PR #5101, opened 2016-05-18
Better cakephp3 completion

Draniu:
I've add subcommands and elements from database 
example

### DIFF
--- a/plugins/cakephp3/cakephp3.plugin.zsh
+++ b/plugins/cakephp3/cakephp3.plugin.zsh
@@ -1,19 +1,38 @@
 # CakePHP 3 basic command completion
 _cakephp3_get_command_list () {
-	cakephp3commands=($(bin/cake completion commands));printf "%s\n" "${cakephp3commands[@]}"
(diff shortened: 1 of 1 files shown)


plugins/cakephp3/cakephp3.plugin.zsh
```diff
@@ -1,19 +1,38 @@
...
+		fi
+		if (( CURRENT == 4 )); then
+			if [[ ${has3rdargument[(i)${words[3]}]} -le ${#has3rdargument} ]] ; then ;
+			 compadd `_cakephp3_get_3rd_argument`
+			fi
+		fi
   fi
 }
 
+has3rdargument=("all" "controller" "fixture" "model" "template")
```
mcornella: It's cleaner defining this variable inside the `_cakephp3` function. It should work as well.

Draniu: i move this varible into the funcction

mcornella:
Thanks!

PR #5103, opened 2016-05-19
Git add all and git commit

christoga:
Example:
```shell
$ gac "commit message"
```


ben4ever:
The alias `gcam` does almost the same thing already.
Also, if we are very precise, your `gac` alias isn't properly positioned since the list of aliases should be sorted alphabetically :wink:.

christoga:
Hm, seems that my PR command is actually have been created.
Since then, I will close this PR.

PR #5104, opened 2016-05-20
gitfast: update to the latest upstream Git and a couple of fixes

felipec:
Basically what the title says (gitfast plugin).


mcornella:
:+1:

PR #5106, opened 2016-05-20
Arcanist plugin

emzar:
This plugin adds many useful aliases for arcanist.


nayyarv:
This plugin is broken 

1. The `alias arc='arc cover'` makes `arc diff, arc land, arc lint` etc. impossible to use. This plugin is an all or nothing change for users since we can't ever revert to `arc foo` since it's now been re-aliased
2. None of the other commands now work, since `ard=arc diff= arc cover diff` which of course makes no sense.

You should probably make it `arcc = arc cover`.

emzar:
@nayyarv , yeah... and I created this [pull request](https://github.com/robbyrussell/oh-my-zsh/pull/5110) resolving issue, but it's still open.

PR #5111, opened 2016-05-23
refresh .rake_tasks when lib/tasks changed

ivanovaleksey:
Hi!

I have found that in Rails app file `.rake_tasks` doesn't refresh.
This happens because `Rakefile` always stays the same.
But all `rake` tasks and `.rake` files are placed in `lib/tasks` folder.

So I have prepared this PR in order to refresh `.rake_tasks` when something has changed in `lib/tasks` folder.

I'm new to shell scripting so any comments are welcome.

Thank you.

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -8,7 +8,19 @@ _rake_refresh () {
 }
 
 _rake_does_task_list_need_generating () {
-  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]]
+  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]] || (_is_rails_app && _tasks_changed)
+}
+
+_is_rails_app () {
+  [[ -e "bin/rails" ]] || [ -e "script/rails" ]
+}
+
+_tasks_changed () {
+  local is_changed=1
+  for file in lib/tasks/**/*.rake; do
+    if [[ $file -nt .rake_tasks ]]; then is_changed=0; fi
+  done
+  return is_changed
 }
 
 _rake_generate () {


plugins/rake-fast/rake-fast.plugin.zsh
```diff
@@ -8,7 +8,11 @@ _rake_refresh () {
 }
 
 _rake_does_task_list_need_generating () {
-  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]]
+  [[ ! -f .rake_tasks ]] || [[ Rakefile -nt .rake_tasks ]] || (_is_rails_app && [[ lib/tasks -nt .rake_tasks ]])
```
mcornella: `lib/tasks` is a folder right? Have you tested this? It's possible that the folder timestamp doesn't change when adding or modifying tasks inside of it.

ivanovaleksey: Yes, it seems rather strange for me too 😃  When I thought about how I could check whether new file has been created `lib/tasks` or existing file has been modified first thing I thought about was some kind of recursion check.

But this check seems to be working alright. I've tested it.
`stat -f%m lib/tasks` is updated both by adding new file and modifying existing one.

There is one thing I should clarify. When I do `git checkout some-branch` `stat -f%m lib/tasks` is changing. So than condition is resolved to `true`. I think it is how it should be. But I'd like you to confirm it.

mcornella: > There is one thing I should clarify. When I do git checkout some-branch stat -f%m lib/tasks is changing. So than condition is resolved to true. I think it is how it should be. But I'd like you to confirm it.

I guess that's how it should be. Do you know of some small project that I can use to test this?

ivanovaleksey: Any `Rails` project fit. For example you can take [this](https://github.com/ivanovaleksey/confucius) my simple draft project. As I understand all you need is ~~love~~ `lib/tasks` folder.

ivanovaleksey:
Hi @mcornella! Do you need any help in testing?

ivanovaleksey:
Hi @mcornella. What about this PR? 😇

mcornella:
Sorry I didn't find a way to test this well. @KevinBongart @leifcr @sachin21 can you guys take a look at this?

KevinBongart:
@mcornella @ivanovaleksey Confirming that with the current version of `rake-fast`, adding a new file to the `lib/tasks` directory of a Rails app won't trigger `rake` to automatically refresh the list of tasks, while with this PR, it does!

Note: adding a new task to an existing file did not for `rake` to refresh the list of tasks though. Not sure if we want to test this as well?

ivanovaleksey:
Sorry @KevinBongart it seems like I don't quite understand you. Do you mean that with this PR adding new task to an existing file will not make `rake` to refresh `.rake_tasks`?

KevinBongart:
It should refresh but didn't for me. Will try again today

leifcr:
It refreshes `.rake_tasks` new tasks when new files are created, but it does not refresh `.rake_tasks` when tasks are added to existing files. Merging the PR would benefit rails developers.

ivanovaleksey:
Thank you for testing. 
It is rather strange because for me `.rake_tasks` is updated for me when I add new task in `lib/tasks` folder. 
Seems like `lib/tasks -nt .rake_tasks` returns `false`.
I will check once again.

mcornella:
That's interesting. Please check if when a file is modified, its parent folder's modified timestamp is also updated.

sachin21:
I'm sorry for the late response.
I also tested this, it works. Looks good to me.
My test case:
```sh
$ rm .rake_tasks
$ rake [tab] # checking tasks
$ : > lib/tasks/task.rake # Erase tasks
$ rake [tab] # checking for the deleted tasks
$ git reset --hard HEAD # Restore tasks
$ rake [tab] # it was successfully restored
```

It works.

mcornella:
Thanks everyone! :heart:

ivanovaleksey:
Hi guys!
I was on a little vacation when it was merged. It was some test code version, but it indeed seems to work correctly.
Thanks for testing.
To be honest this is my first open-source PR. Hope it would be useful for others and would not produce any bugs.

ivanovaleksey:
As I remember there was some bug and now I have found it.
If I remove the whole `.rake` file i.e. `rm -f lib/tasks/foo.rake` `.rake_tasks` is not updated.
Should I fix it?

mcornella:
That's weird, in my system when a file is deleted the folder modified timestamp is also updated. See if you can figure out a reproducible test.

mcornella:
For instance:
```zsh
$ mkdir -p lib/tasks; zstat +mtime lib/tasks lib/tasks/*(N)
1471464519
$ touch lib/tasks/{example1,example2}; zstat +mtime lib/tasks lib/tasks/*(N)
lib/tasks 1471464527
lib/tasks/example1 1471464527
lib/tasks/example2 1471464527
$ rm lib/tasks/example1; zstat +mtime lib/tasks lib/tasks/*(N)
lib/tasks 1471464538             <-- this changed
lib/tasks/example2 1471464527
```

ivanovaleksey:
You are right, Marc. Thank you for the example.
But now in PR I only [check](https://github.com/robbyrussell/oh-my-zsh/pull/5111/files#diff-ef46552e72ba783013e8227056272bdbR20) for nested existing `.rake` files.
I was trying to combine current strategy with checking the whole `lib/tasks` directory timestamp (like it was in the beginning) but PR was already merged.

By the way where can I find `zstat` command?

mcornella:
Oh I hadn't seen that you changed it, I thought we were still using the `[[ lib/tasks -nt .rake_tasks ]]` form. Please submit a new PR changing it to that and I'll merge it right away.

zstat is a zsh module. If you don't have it you can do `zmodload zsh/zstat`.

ivanovaleksey:
Thanks Marc, I will submit new PR.
Running `zmodload` gives me an error.
```
$ zmodload zsh/zstat
zsh: failed to load module `zsh/zstat': dlopen(/usr/lib/zsh/5.0.8/zsh/zstat.so, 9): image not found
```
I didnot find any solution at Google.
Meanwhile `zmodload zsh/datetime` works fine.

ivanovaleksey:
Seems like it should be `zmodload zsh/stat`.

mcornella:
Yes, my bad 😁

PR #5112, opened 2016-05-23
Added grho: git reset --hard origin/$(current_branch)

autodidacticon:


vgaidarji:
+1

@autodidacticon You fixed typo in PR name but it remains in the commit message.

autodidacticon:
@vgaidarji Good catch.

vgaidarji:
I would vote for this feature +100 as probably I'm not the one who uses this command on daily basis.

autodidacticon:
@ncanceill

mcornella:
It's now `groh`.

PR #5113, opened 2016-05-24
Ditch smkx in favor of custom bindkey extension

### LINKED ISSUE #2369: Change in option+arrow keyboard movements in OS X?
Used to be that option+left arrow or option+right arrow would move the cursor in OS X's Terminal.app by word (mimicking, near as I can tell, the meta-f and meta-b keystrokes in *nix environments). I swear that was happening...at least until a week or so ago.

There was an update to OhMyZsh during this time and that's the only thing that I can think of in my Terminal environment that's changed so I thought I would ask. I have no idea where to start investigating this myself.

Thanks for all the hard work on OMZ, by the way. It's been a fantastic tool that I've very much enjoyed using. Props to everyone involved and happy holidays to all. :)

### LINKED ISSUE #2654: Numeric PAD is not usable anymore on MAC OS X
Since last update of yesterday 24th March my numerical pad is not usable anymore in my shell.
All our team developers are in the same situation.

### LINKED ISSUE #2698: Number pad home key fails to function with gnome-terminal
Gnome-terminal apparently uses different escape sequences for the normal home key and the home key located on the number pad. With the switch to using terminfo in 174c9177aa34b4c05bb5a1c6f637e6fa479a8e10 the number pad home key ceases to function.

### LINKED ISSUE #2735: history-substring-search no longer functions after latest update

### LINKED ISSUE #2750: Alt-Left-Arrow and Alt-Right-Arrow broken on urxvt in openSUSE

### LINKED ISSUE #3061: Cannot using home/end key after install oh-my-zsh
I using a pure Ubuntu:14.04 env.
After I install oh-my-zsh by curl.
I cannot using home/end key anymore.
I didn't change any settings in .oh-my-zsh folder.

### LINKED ISSUE #3495: Home and End keys aren't working
On Ubuntu 14.04 the Home|End keys work very well.
On Centos 5.10 these keys work on a bash but not in zsh.

Thus, sometimes the prompt lose its position:
selection_011

### LINKED ISSUE #3733: Home and End keys not working on CentOS
Avarus presented a solution that work great [here](https://www.linuxquestions.org/questions/linux-software-2/ins-end-delete-keys-are-mischevious-297024/)

```bash
#Rebind HOME and END to do the decent thing:
bindkey '\e[1~' beginning-of-line
bindkey '\e[4~' end-of-line
case $TERM in (xterm*)
bindkey '\eOH' beginning-of-line
bindkey '\eOF' end-of-line
esac

#To discover what keycode is being sent, hit ^v
#and then the key you want to test.

#And DEL too, as well as PGDN and insert:
bindkey '\e[3~' delete-char
bindkey '\e[6~' end-of-history
bindkey '\e[2~' redisplay

#Now bind pgup to paste the last word of the last command,
bindkey '\e[5~' insert-last-word
```

### LINKED ISSUE #3757: "end" and "home" keys don't work with putty
back to bash till this is fixed!!!

### LINKED ISSUE #4784: Home and End not working in PhpStorm terminal in Mac OS/X
If i go to bash, it works. In Oh My Zsh I only get system sound when pressing the home and/or end keys.

### LINKED ISSUE #4872: MobaXTerm "xterm" terminal - END, HOME, DEL keys not working
I don't know if putting this up on this project is the right way to go, but when I searched for tickets in this git, I saw similar ones, so here goes.

I have a CentOS 7.2.1511 machine, with ZSH 5.0.2 installed. I have the latest version of oh-my-zsh installed, along with the plugin ZSH-Syntax-Highlighting.

To connect to this machine, I use MobaXTerm (latest version). I connect to the CentOS machine through SSH and present the terminal as being of the "xterm" type (the default option for this SSH client).

The standard HOME, END, DEL, etc keys don't work out-of-the-box. I don't know how to make a Terminfo file for this case, and it seems like I shouldn't be needing to come up with a new Terminfo in the first place, seeing as MobaXTerm truly is an xterm terminal.

As such, I'd like to hear from you on this matter: is there some kind of known bug with this specific ZSH version that impedes MobaXTerm from working as intented?

Thanks for your time.

### LINKED ISSUE #4916: Num Pad can't input
I'm using Mac + ZSH for developer and Gentoo + ZSH for server.
Both of those my Keyboard's num-pad can't input any char(1,2,3,4,5,6,7,8,9,0,.,=,/,*,-,+,enter)
But when I change to bash, those keys are OK.
In any other Applicatons, those keys are OK too.

I'm using Razer's BlackWidow keyboard.

### LINKED ISSUE #5149: Plugin 'safe-paste' messes with 'history-substring-search'
`history-substring-search` doesn't work after enabling `safe-paste`, disabling it fixes the problem.

### LINKED ISSUE #5228: Help with application mode
Not sure how to ask for help on this. I'm using [VS Code](https://github.com/Microsoft/vscode) and there is an integrated terminal that uses [xterm.js](https://github.com/sourcelair/xterm.js). I use oh my zsh for my default shell on my mac. I'm using OS X 10.11.5. The issue I'm having (and not sure where the actual problem lies) is that xterm.js is being put into `Application Keypad` mode because of `echoti smkx` ([this line](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/key-bindings.zsh#L9)). Here is [the bug](https://github.com/sourcelair/xterm.js/issues/151) that we have filled with xterm.js.
Not sure where to go from here, any advice would be great.

### LINKED ISSUE #5237: Safe-paste plugin breaks Home/End keys
Enabling the safe-paste plugin breaks the Home/End keys for me. I can't move the the beginning or end of a line using Home/End. If I disable the plugin Home/End works again. I ran omz_diagnostic_dump and it shows that the bindkeys were the same value regardless of whether the plugin was enabled or disabled.
`"^A" beginning-of-line`
`"^[OH" beginning-of-line`

`"^E" end-of-line`
`"^[OF" end-of-line`

So I'm not sure what might be overriding this.

apjanke:
## Motivation

More than 2 years ago, Oh My Zsh merged the Pull Request #1355, which introduced a fix to be able to use terminfo sequences to specify key codes to `bindkey`, with the intention to make it standard for everyone and not having to change bindkey sequences all the time. 

**That experiment failed, and it's been obviously clear for quite a while.** Since then lots of issues have popped up where certain combinations of keys didn't work (see list of _Fixes_ below), and the culprit of that is that terminfo databases are not maintained all that well and there is not consistent support among Terminal Emulators.

This is why we're choosing to drop that experiment, and use a custom wrapper around bindkey named `omz_bindkey` were you'll be able to choose whichever modifier keys you want ─ <kbd>CTRL</kbd>, <kbd>ALT</kbd>, <kbd>SHIFT</kbd>, <kbd>META</kbd> ─ instead of having to use an obscure character sequence like `'^[[OH'`.

**Here's where _you_ come into play:** since this is a big change, we need a broad number of users to test in their respective configurations, with as many Terminal Emulators as we can, before we can safely merge this PR knowing that there won't be many users affected.

Thanks for taking the time to read this; in [this GitHub guide](https://help.github.com/articles/checking-out-pull-requests-locally/) you'll find how you can check out this PR locally. After that, restart your terminal or run `exec zsh` to reload Oh My Zsh.

Please report back if there are any keyboard shortcuts broken in your system with the following information:
- Zsh version: run `echo $ZSH_VERSION`.
- Terminal Emulator and its version: _e.g._ iTerm2, Terminal.app, Gnome Terminal, Guake, ...
- TERM variable: run `echo $TERM`.
- The broken keyboard shortcut and its character sequence which you can get [following this tutorial](https://asciinema.org/a/ejxdamcgxew2qlmyb8gdnrxeh) I've made.
- Your current `bindkey` configuration: run `bindkey` and post its output.
- Anything else that you think is relevant.

That is all. Thanks again for participating in this! ─ @mcornella

---
## Description

I think it's time to abandon `smkx`/`rmkx` and stick with "local" keypad mode, even if it requires us to hardcode a few character sequences instead of using terminfo for everything. This PR does so, adding an `omz_bindkey` function that allows us to still use terminfo capability codes in key bindings, and supports cursor keys with modifiers. This should fix a whole slew of these recurring cursor-key terminal issues.

Fixes #4916 – Num Pad can't input 
Fixes #4872 – Home and End not working in MobaXterm
Fixes #4784 – Home and End not working in PhpStorm terminal in Mac OS/X
Fixes #3757 – Home/End broken in PuTTY
Fixes #3733 – Home/End broken in CentOS
Fixes #3495 – Home/End broken in CentOS
Fixes #3061 – Home/End broken in Ubuntu
Probably fixes #2750 – Cursor key bindings broken in urxvt on openSUSE – or at least makes it moot
Fixes #2735 – history-substring-search key bindings broken after terminfo changeover
Fixes #2698 – Home/End broken in gnome-terminal
Fixes #2654 – Num keypad broken on OS X
Fixes #2369 – Change to option-arrow behavior on OS X
Fixes #5149 – Plugin 'safe-paste' messes with 'history-substring-search'
Fixes #5228 – Application mode not working on xterm.js
Fixes #5237 - Safe-paste plugin breaks Home/End
Avoids sourcelair/xterm.js#151 - zsh / oh-my-zsh - no scroll
...and all those other issues about Home/End or the numeric keypad.
## Discussion

Switching to use smkx/rmkx application mode and terminfo for portability was a good idea: you're _supposed_ to use portability libraries for stuff like this. But I think it's time to call it a failed experiment. Not because the implementation in OMZ was bad, but because terminfo and the various terminal programs out there don't have sufficient support for it.

I dug pretty deep in to this issue, and I think that the smkx/rmkx+terminfo solution _cannot_ be made to work for everyone. And you'll continue to see issues reported for it. But local mode with a few hardcoded sequences _can_ work.

The smkx/rmkx approach is defeated by a few factors:
- Terminal emulators [vary in their application mode behavior](https://docs.google.com/spreadsheets/d/1Ue5gb2hLjz058QEtEXwaw8YR7QoVZpSYzbXQKrRcGKk/edit#gid=0), and don't necessarily match `xterm`, which is what most people use for their `$TERM`, even if they're really running a different program
- terminfo conflates the mode setting for cursor keypads and the numeric keypads: [sr]mkx changes both, but we only want the cursor keypads changed
- the program-specific terminfo database entries for PuTTY, Terminal.app, and other programs may not be installed on a given system, or may be incomplete or out of date. We don't want to require custom installation of terminfo entries if we can possibly avoid it.

Using smkx/rmkx is also complicated because:
- Most terminal-related code snippets on forums, StackOverflow, and the rest of the internet are for local mode, so they won't work in OMZ if we use [sr]mkx
- `bash` only supports local mode, so bash cursor code can't be used under [sr]mkx
- Most people don't know about [sr]mkx, so educating people about it becomes part of every bug report for keypad problems
- Using [sr]mkx requires using the `zle-line-init` and `zle-line-finish` hooks, which conflicts with anything else that wants to use them (including some OMZ plugins).
#### zle-line-init/finish hooks

As part of this PR's changes, Oh My Zsh completely takes over the `zle-line-init` and `zle-line-finish` hooks, replacing them with functions that provide support for multiple hook functions to be registered. This is done for two independent reasons. Primarily, it's to clobber and remove any zle-line-init/finish hooks installed by zsh configuration files supplied by the system. Some Linux distributions define hooks which do the [sr]mkx thing, and we need to make sure that doesn't happen. Secondly, it allows for multiple OMZ plugins to install line-init/finish hooks without conflicting with each other.

As part of this, we're essentially having OMZ declare zle-line-init/finish to be completely its territory. I think this is okay.
## Implementation

The terminfo-based key binding code was a lot easier to read, because it used mnemonic terminfo capability codes instead of literal terminal control sequences. So this PR adds an `omz_bindkey` extension on top of `bindkey` that allows you to still use those codes. (It also adds logging support, which would make diagnosing key binding issues a lot easier, because you could see which plugins or user code is doing which binding.) I think this makes the key binding code a lot more readable, both as it is, and in the diffs for future PRs.

The `omz_bindkey` layer keeps a supplemental little database of character sequences based on xterm's local mode behavior (which is not in terminfo), and uses those for cursor keys. It also includes a couple hacks for specific known misbehaving or variant terminal programs. And it has support for binding modifier-key (alt/ctrl/shift) cursor keys, which terminfo doesn't itself provide.
## Future considerations

If you like this approach and merge it, I'll follow it up with a PR for multiple keymap support in the key bindings, which will take care of the issues caused by flipping between the `emacs` and `vi` keymaps during the OMZ startup sequence.

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -223,3 +223,13 @@ function omz_urldecode {
 
   echo -E "$decoded"
 }
+
+# Print input iff $_OMZ_DEBUG is enabled (>0)
+# All arguments are passed verbatim to print()
+function _omz_dbg_print() {
+  if ! (( $_OMZ_DEBUG )); then
+    return
+  fi
+  print "$@"
+}
+
--- a/lib/key-binding-support.zsh
+++ b/lib/key-binding-support.zsh
@@ -1,0 +1,200 @@
+# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html
+# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Zle-Builtins
+# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Standard-Widgets
+
+zmodload -i zsh/zutil
+
+# OMZ takes over zle-line-init/finish, to:
+#  a) provide hooks for multiple init/finish functions, like $precmd_functions does
+#  b) disable [sr]mkx on the few systems whose system zsh configuration enables it
+typeset -g -a zle_line_init_functions zle_line_finish_functions
+function zle-line-init() {
+  local fcn
+  for fcn in "${zle_line_init_functions[@]}"; do
+    $fcn
+  done
+}
+function zle-line-finish() {
+  local fcn
+  for fcn in "${zle_line_finish_functions[@]}"; do
+    $fcn
+  done
+}
+zle -N zle-line-init
+zle -N zle-line-finish
+
+# DEBUG: This is the old stuff we used to do to put the keypads in application mode
+# during command line editing. It's included here to make it easy to test the
+# new key binding logic under both normal and application modes. This should
+# go away once the modeless key binding stuff is tested.
+# Note: This breaks the zle-line-init/finish hooks from above.
+if (( $_OMZ_DEBUG_SMKX )); then
+  if (( ${+terminfo[smkx]} )) && (( ${+terminfo[rmkx]} )); then
+    function zle-line-init() {
+      echoti smkx
+    }
+    function zle-line-finish() {
+      echoti rmkx
+    }
+    zle -N zle-line-init
+    zle -N zle-line-finish
+  fi
+fi
+# end old keypad-mode stuff
+
+
+# Extra portability cursor key mappings for local-mode keypad and terminals
+# which deviate from typical terminfo entries. Supplements $terminfo.
+# Maps capability names to |-delimited lists of character sequences.
+# Reflects the capabilities of the $TERM defined at startup.
+typeset -AHg _omz_terminfo_extra
+_omz_terminfo_extra=(
+  # Start with the basic Xterm local-mode sequences. (Some people call 
+  # these "ANSI" key sequences.)
+  kcuu1   "\e[A"  # up
+  kcud1   "\e[B"  # down
+  kcuf1   "\e[C"  # right ("forward")
+  kcub1   "\e[D"  # left ("backward")
+  khome   "\e[H"  # home
+  kend    "\e[F"  # end
+  )
+# Portability entries for terminals we can't recognize from $TERM
+# Putty and others may send the VT220 editing keypad sequences for
+# Home/End. Bind those too, since there's no collision risk AFAIK.
+# TODO: Should we alias the whole 6-key VT220 editing keypad?
+_omz_terminfo_extra[khome]="$_omz_terminfo_extra[khome]|\e[1~"
+_omz_terminfo_extra[kend]="$_omz_terminfo_extra[kend]|\e[4~"
+# Do *not* bind '\e[Ow' here, which Putty may also send for End, because
+# it collides with application-mode numeric keypad sequences.
+
+# Terminal-specific portability entries for the current terminal, for
+# terminals whose local-mode sequences are known to differ from Xterm's
+case "$TERM" in
+  rxvt-unicode*)
+    # rxvt-unicde local-cursor sequences that differ from xterm
+    _omz_terminfo_extra[khome]="$_omz_terminfo_extra[khome]|\e[7~"
+    _omz_terminfo_extra[kend]="$_omz_terminfo_extra[kend]|\e[8~"
+esac
+
+# omz_bindkey - an extension to bindkey
+#
+# omz_bindkey [options] -t <capability> <command>
+# omz_bindkey -c <modifier> <cursor-key> <command>
+# omz_bindkey [...normal bindkey arguments...]
+#
+# Adds the following features on top of bindkey:
+# 1) Logging
+# 2) -t: binds a key, using terminfo and OMZ portability mappings
+# 3) -c: binds an Xterm-style modified cursor key sequence
+#
+# Returns:
+#   0 if a binding was made
+#   1 if no binding was made
+#   2 if an error occurred
+function omz_bindkey() {
+  emulate -L zsh
+  # Note: "seq" is short for "sequence" in these functions
+  # We must capture all of bindkey's options to get ordering right when calling it
+  local -a omz_opts bindkey_opts
+  omz_opts=()
+  bindkey_opts=()
+  local -A optvals
+  zparseopts -a bindkey_opts -A optvals -D -E t:=omz_opts c=omz_opts M:
+  if [[ -n ${omz_opts[(r)-t]} ]]; then
+    # OMZ terminfo-based binding
+    local cap=${optvals[-t]}
+    local retval=1 seq seqs
+    seqs=()
+    if [[ -n "${terminfo[$cap]}" ]]; then
+      seqs=("$seqs[@]" "${terminfo[$cap]}")
+    fi
+    if [[ -n "${_omz_terminfo_extra[$cap]}" ]]; then
+      seqs=("$seqs[@]" "${(s:|:)_omz_terminfo_extra[$cap]}")
+    fi
+    # Done with a while loop because for seems to break here
+    local ix=0
+    while (( ++ix <= $#seqs )); do
+      seq="${seqs[$ix]}"
+      _omz_dbg_print -r "omz_bindkey: terminfo: $cap '${(q)seq}' $@"
+      bindkey "${bindkey_opts[@]}" $seq "$@"
+      retval=0
+    done
+    return $retval
+  elif [[ -n ${omz_opts[(r)-c]} ]]; then
+    # "-c" for "composed modified-cursor" keys
+    _omz_dbg_print -r "omz_bindkey: modified-cursor: '$1' '$2' '$3'"
+    _omz_bindkey_modified_cursor_key_xterm "$1" "$2" "$3"
+  else
+    _omz_dbg_print -r "omz_bindkey: passthrough: ${(q)@}"
+    bindkey "${bindkey_opts[@]}" "$@"
+  fi
+}
+
+# Xterm style function key modifier codes
+typeset -Ag _omz_ti_modifiers
+_omz_ti_modifiers=(
+  'shift'           2
+  'alt'             3
+  'alt-shift'       4
+  'ctrl'            5
+  'ctrl-shift'      6
+  'ctrl-alt'        7
+  'ctrl-alt-shift'  8
+  'meta'            9
+  'meta-shift'      10
+  'meta-alt'        11
+  'meta-alt-shift'  12
+  'meta-ctrl'       13
+  'meta-ctrl-shift' 14
+  'meta-ctrl-alt'   15
+  'meta-ctrl-alt-shift' 16
+  )
+
+# Binds a modified cursor key sequence
+# Uses xterm modifier mask codes,
+# per http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-PC-Style-Function-Keys
+# When used in combination, modifier keys must be supplied in the following canonical order:
+#  meta ctrl alt shift
+function _omz_bindkey_modified_cursor_key_xterm() {
+  emulate -L zsh
+  local opts
+  opts=()
+  zparseopts -a opts -D -E M:
+  local modifier=$1 key=$2 widget=$3
+  local CSI='\e['  # Control Sequence Introducer
+  local seq modified_seq
+  local -A modifiers keys
+  set -A modifiers ${(kv)_omz_ti_modifiers}
+  keys=(
+    up    A
+    down  B
+    right C
+    left  D
+    home  H
+    end   F
+    )
+  if [[ -z "$modifiers[$modifier]" ]] || [[ -z "$keys[$key]" ]]; then
+    _omz_dbg_print -r "Invalid modified-cursor-key: modifier=$modifier key=$key"
+    return 1
+  fi
+  seq="${CSI}${keys[$key]}"
+  _omz_bindkey_modified_cursor_seq_xterm_step "$seq" "$modifier"
+  # Wokaround for iTerm2: it reports alt as meta, so bind alt-modified sequences
+  # to the meta-modified variant, too.
+  # See https://gitlab.com/gnachman/iterm2/issues/3753
+  if [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then
+    if [[ "$modifier" == "alt" ]]; then
+      _omz_bindkey_modified_cursor_seq_xterm_step "$seq" "meta"
+    fi
+  fi
+}
+
+function _omz_bindkey_modified_cursor_seq_xterm_step() {
+  local seq="$1" modifier="$2"
+  local modified_seq
+  local -A modifiers keys
+  set -A modifiers ${(kv)_omz_ti_modifiers}
+  modified_seq="${seq[1,-2]}1;$modifiers[$modifier]${seq[-1,-1]}"
+  omz_bindkey "${opts[@]}" "$modified_seq" "$widget"  
+}
+
--- a/lib/key-bindings.zsh
+++ b/lib/key-bindings.zsh
@@ -1,93 +1,39 @@
-# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html
-# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Zle-Builtins
-# http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Standard-Widgets
-
-# Make sure that the terminal is in application mode when zle is active, since
-# only then values from $terminfo are valid
-if (( ${+terminfo[smkx]} )) && (( ${+terminfo[rmkx]} )); then
-  function zle-line-init() {
-    echoti smkx
-  }
-  function zle-line-finish() {
-    echoti rmkx
-  }
-  zle -N zle-line-init
-  zle -N zle-line-finish
-fi
+# Specific key bindings
+# See also key-binding-support.zsh for helper functions used here
 
 bindkey -e                                            # Use emacs key bindings
 
-bindkey '\ew' kill-region                             # [Esc-w] - Kill from the cursor to the mark
-bindkey -s '\el' 'ls\n'                               # [Esc-l] - run command: ls
-bindkey '^r' history-incremental-search-backward      # [Ctrl-r] - Search backward incrementally for a specified string. The string may begin with ^ to anchor the search to the beginning of the line.
-if [[ "${terminfo[kpp]}" != "" ]]; then
-  bindkey "${terminfo[kpp]}" up-line-or-history       # [PageUp] - Up a line of history
-fi
-if [[ "${terminfo[knp]}" != "" ]]; then
-  bindkey "${terminfo[knp]}" down-line-or-history     # [PageDown] - Down a line of history
-fi
+omz_bindkey    '\ew' kill-region                      # [Esc-w] - Kill from the cursor to the mark
+omz_bindkey -s '\el' 'ls\n'                           # [Esc-l] - run command: ls
+omz_bindkey    '^r' history-incremental-search-backward  # [Ctrl-r] - Search backward incrementally for a specified string. 
+                                                      #    The string may begin with ^ to anchor the search to the beginning of the line.
 
-# start typing + [Up-Arrow] - fuzzy find history forward
-if [[ "${terminfo[kcuu1]}" != "" ]]; then
-  autoload -U up-line-or-beginning-search
-  zle -N up-line-or-beginning-search
-  bindkey "${terminfo[kcuu1]}" up-line-or-beginning-search
-fi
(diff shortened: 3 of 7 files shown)


mcornella:
Thanks for this Andrew, hopefully I'll have it checked out in less than a week. I agree on the *failed experiment* sentiment, I think this is the only way forward.

mcornella:
The `history-substring-search` plugin stops working after this PR. I think we have to also [the terminfo-style key bindings](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/history-substring-search/history-substring-search.plugin.zsh#L18-L25). 

Terminfo `kcuu1` and `kcud1` are `"^[OA"` and `"^[OB"` respectively, while using `CTRL`+`V` for UP / DOWN arrow I get `^[[A` and `^[[B` respectively.

What I don't get is why it works for you, perhaps you have overwritten `bindkey`s?

My `infocmp`:
```
#	Reconstructed via infocmp from file: /lib/terminfo/x/xterm-256color
xterm-256color|xterm with 256 colors,
	am, bce, ccc, km, mc5i, mir, msgr, npc, xenl,
	colors#256, cols#80, it#8, lines#24, pairs#32767,
	acsc=``aaffggiijjkkllmmnnooppqqrrssttuuvvwwxxyyzz{{||}}~~,
	bel=^G, blink=\E[5m, bold=\E[1m, cbt=\E[Z, civis=\E[?25l,
	clear=\E[H\E[2J, cnorm=\E[?12l\E[?25h, cr=^M,
	csr=\E[%i%p1%d;%p2%dr, cub=\E[%p1%dD, cub1=^H,
	cud=\E[%p1%dB, cud1=^J, cuf=\E[%p1%dC, cuf1=\E[C,
	cup=\E[%i%p1%d;%p2%dH, cuu=\E[%p1%dA, cuu1=\E[A,
	cvvis=\E[?12;25h, dch=\E[%p1%dP, dch1=\E[P, dim=\E[2m,
	dl=\E[%p1%dM, dl1=\E[M, ech=\E[%p1%dX, ed=\E[J, el=\E[K,
	el1=\E[1K, flash=\E[?5h$<100/>\E[?5l, home=\E[H,
	hpa=\E[%i%p1%dG, ht=^I, hts=\EH, ich=\E[%p1%d@,
	il=\E[%p1%dL, il1=\E[L, ind=^J, indn=\E[%p1%dS,
	initc=\E]4;%p1%d;rgb\:%p2%{255}%*%{1000}%/%2.2X/%p3%{255}%*%{1000}%/%2.2X/%p4%{255}%*%{1000}%/%2.2X\E\\,
	invis=\E[8m, is2=\E[!p\E[?3;4l\E[4l\E>, kDC=\E[3;2~,
	kEND=\E[1;2F, kHOM=\E[1;2H, kIC=\E[2;2~, kLFT=\E[1;2D,
	kNXT=\E[6;2~, kPRV=\E[5;2~, kRIT=\E[1;2C, kb2=\EOE,
	kbs=\177, kcbt=\E[Z, kcub1=\EOD, kcud1=\EOB, kcuf1=\EOC,
	kcuu1=\EOA, kdch1=\E[3~, kend=\EOF, kent=\EOM, kf1=\EOP,
	kf10=\E[21~, kf11=\E[23~, kf12=\E[24~, kf13=\E[1;2P,
	kf14=\E[1;2Q, kf15=\E[1;2R, kf16=\E[1;2S, kf17=\E[15;2~,
	kf18=\E[17;2~, kf19=\E[18;2~, kf2=\EOQ, kf20=\E[19;2~,
	kf21=\E[20;2~, kf22=\E[21;2~, kf23=\E[23;2~,
	kf24=\E[24;2~, kf25=\E[1;5P, kf26=\E[1;5Q, kf27=\E[1;5R,
	kf28=\E[1;5S, kf29=\E[15;5~, kf3=\EOR, kf30=\E[17;5~,
	kf31=\E[18;5~, kf32=\E[19;5~, kf33=\E[20;5~,
	kf34=\E[21;5~, kf35=\E[23;5~, kf36=\E[24;5~,
	kf37=\E[1;6P, kf38=\E[1;6Q, kf39=\E[1;6R, kf4=\EOS,
	kf40=\E[1;6S, kf41=\E[15;6~, kf42=\E[17;6~,
	kf43=\E[18;6~, kf44=\E[19;6~, kf45=\E[20;6~,
	kf46=\E[21;6~, kf47=\E[23;6~, kf48=\E[24;6~,
	kf49=\E[1;3P, kf5=\E[15~, kf50=\E[1;3Q, kf51=\E[1;3R,
	kf52=\E[1;3S, kf53=\E[15;3~, kf54=\E[17;3~,
	kf55=\E[18;3~, kf56=\E[19;3~, kf57=\E[20;3~,
	kf58=\E[21;3~, kf59=\E[23;3~, kf6=\E[17~, kf60=\E[24;3~,
	kf61=\E[1;4P, kf62=\E[1;4Q, kf63=\E[1;4R, kf7=\E[18~,
	kf8=\E[19~, kf9=\E[20~, khome=\EOH, kich1=\E[2~,
	kind=\E[1;2B, kmous=\E[M, knp=\E[6~, kpp=\E[5~,
	kri=\E[1;2A, mc0=\E[i, mc4=\E[4i, mc5=\E[5i, meml=\El,
	memu=\Em, op=\E[39;49m, rc=\E8, rev=\E[7m, ri=\EM,
	rin=\E[%p1%dT, ritm=\E[23m, rmacs=\E(B, rmam=\E[?7l,
	rmcup=\E[?1049l, rmir=\E[4l, rmkx=\E[?1l\E>, rmso=\E[27m,
	rmul=\E[24m, rs1=\Ec, rs2=\E[!p\E[?3;4l\E[4l\E>, sc=\E7,
	setab=\E[%?%p1%{8}%<%t4%p1%d%e%p1%{16}%<%t10%p1%{8}%-%d%e48;5;%p1%d%;m,
	setaf=\E[%?%p1%{8}%<%t3%p1%d%e%p1%{16}%<%t9%p1%{8}%-%d%e38;5;%p1%d%;m,
	sgr=%?%p9%t\E(0%e\E(B%;\E[0%?%p6%t;1%;%?%p5%t;2%;%?%p2%t;4%;%?%p1%p3%|%t;7%;%?%p4%t;5%;%?%p7%t;8%;m,
	sgr0=\E(B\E[m, sitm=\E[3m, smacs=\E(0, smam=\E[?7h,
	smcup=\E[?1049h, smir=\E[4h, smkx=\E[?1h\E=, smso=\E[7m,
	smul=\E[4m, tbc=\E[3g, u6=\E[%i%d;%dR, u7=\E[6n,
	u8=\E[?1;2c, u9=\E[c, vpa=\E[%i%p1%dd,
```

apjanke:
Ah, yes: that will need to be changed to use `omz_bindkey` instead of plain `bindkey`, and make sure `omz_bindkey` supports `-t` in conjunction with `-M`. I'll update this PR to do so.

I didn't notice it because I have some additional key bindings set in my `~/.zshrc` and I forgot to test it against a plain `~/.zshrc` too.

apjanke:
Okay, I've amended the PR to handle `-M` and have the `history-substring-search` and `dircycle` plugins use the new `omz_bindkey` mechanism.

mcornella:
The `dircycle` plugin stopped working because the `omz_bindkey` function uses `shift-ctrl` while the plugin uses `ctrl-shift`. 

If only one of these should work, I think `ctrl-shift` is a much more common occurrence. Another option is to support both permutations, but then all other key permutations should follow the same rule.

My suggestion is to use `ctrl-shift`, `ctrl-alt` and `ctrl-alt-shift`. The `meta-*` permutations should also change in the same fashion.

apjanke:
Good catch.

I think it'll be more readable if we stick to a canonical ordering. (Alternately, the `omz_bindkey` function itself could do string splitting and re-order the inputs to a canonical ordering, but I'd rather not spend the time to implement that now.)

I've switched it to using your suggested canonical ordering. The `dircycle` plugin now works in my testing (with Xterm).

mcornella:
Ok, this appeared to fix #5228, which I've added to the list of fixes. I'll see if we can accelerate the testing of this PR so we can merge it right away.

snussbaumer:
FYI I've switched to this PR for now, as this solves keybinding problems with the home/end/ctrl+left and ctrl+right keys in IntelliJ IDEA 2016.2.4 (on Ubuntu 16.04). I see no problem at this moment, but will log any issue here if I find one. Let me know if this is not the proper place for "testers".

mcornella:
@snussbaumer yes this is the proper place, thanks for commenting. Did you have to do anything extra after applying this PR, _i.e._ delete some `bindkey` commands from your zshrc file?

snussbaumer:
Another note : in IntelliJ terminal showkey gives me the following bindings : 

 - `^[[5D` for CTRL+LEFT
 - `^[[5C] for CTRL+RIGHT

Whereas when I'm in GNU terminal I get : 

 - `^[[1;5D` for CTRL+LEFT
 - `^[[1;5C` for CTRL+RIGHT

Luckily my changes in zshrc seem to be ok in both environments ... but still this is quite confusing

apjanke:
Good to hear.

That sounds like IntelliJ is behaving oddly for Ctrl-modified arrow keys. That `^[[5D` means "five spaces left" instead of "one space left, modified by Ctrl". The GNU terminal behavior is more normal for terminal emulators. That would explain why you still need bindings for `^[[5D` and `^[[5C`; they're non-standard.

Someone should report this as a bug upstream to IntelliJ.

mcornella:
For the `^[[5D` and `^[[5C` you could also use the wrapper `omz_bindkey` introduced in this PR. Here's how you would use it:
```
omz_bindkey -c ctrl left backward-word
omz_bindkey -c ctrl right forward-word
```

apjanke:
> For the `^[[5D` and `^[[5C` you could also use the wrapper `omz_bindkey` introduced in this PR. Here's how you would use it:

I don't think that will work. The `omz_bindkey` modified-cursor support is based on the standard `xterm` modifier-key control sequences. The IntelliJ terminal is deviating from that behavior, so the `ctrl left` and `ctrl right` bindings won't match it. `terminfo` doesn't have support for modified-cursor patterns, either. You'll need to continue to hardcode the sequences for IntelliJ to work around this.

I suspect this is a bug in the IntelliJ terminal plugin. It looks like it's naively inserting a `5` for "ctrl" into the `^[[D` etc sequences without realizing it has to add an explicit `1;` before it in the case of the cursor keys.

mcornella:
Oh right, my mistake, I meant the other two sequences.

apjanke:
Okay, yep, that'll totally work.

hjri:
Hello. I tried checking out this PR, however problem with up-arrow not doing partial search and non-working home/end keys still presists, same thing happens on master branch. 

Currenly my workaround is to execute `zle-line-init`, but that persists only for next command. doing `zle-line-init; zle-line-finish` however does not help. 

I am at loss, where should I look? I have two machines with more-or-less same setup, on one problem exists, on other problem is absent even on master branch. Currenly only important difference I see is that problem occurs on zsh version 5.2, while on zsh version 5.0.7 it does not. Both machines are running Debian, but different versions (testing/sid and stable respectively).

Daemoen:
Curious why this is still open and unmerged?  Is it because there are still more tests needed before feeling it's safe to merge?  to this date, even using master, vi-mode is still broken in numerous cases, as is safe-paste.

robbyrussell:
@mcornella Do you think this is close to merge-able?

HyP3r-:
Also from my perspective: why is this problem still not solved? For a terminal extension that is used by so many people (87,000 stars) it is embarrassing if simple things like crtl-left or the Numpad don't work. Please fix it! @robobenklein @mcornella @apjanke

robbyrussell:
We could really use some testers on this. Thanks in advance!

HyP3r-:
I [have merged this pull request into my fork](https://github.com/HyP3r-/oh-my-zsh/commit/a2aa4fa098d58d4fca206b8e4aa385490495a112) and have not encountered any problems with it. On these systems and platforms:
Debian Jessie, Stretch and Buster and Ubuntu 18.04.3 LTS and 19.10 on the Local terminals as well as via SSH on Windows with PuTTy 0.63 and 0.73 and Remote Desktop Manager 2019.2.21.0 and on Linux with ssh in the terminal. As well as on cygwin 3.1.2 with the local terminal.

larson-carter:
@apjanke 

Could you please upstream this, by getting rid of the conflicts? If so we would then be able to continue our reviews on it. 

Thank You!

apjanke:
Rebased and resolved conflicts; I _think_ I got it right.

larson-carter:
@apjanke Thank you.

@mcornella Do we want to proceed with a PR review?

pgimalac:
I've encountered a bug with this PR: when I type a command and go up to search in my history all commands starting with the first word appear (not necessarily the full command).
For example when I type "git pull" and up there are commands like "git push" or "git add" that can appear, while I would expect only commands like "git pull origin master".

apjanke:
Hmmmm... I can't reproduce your issue. When I check out my `~/.oh-my-zsh` against this PR and do "git pull" and hit up arrow, it only grabs the items from my history that specifically start with "git pull".

git-pull-and-do-up-arrow-under-ditch-smkx

Maybe you've got some other autocompletion configuration stuff going on? Are you all rebased on `master` and stuff?

pgimalac:
@apjanke I tried again and it didn't happen anymore !
I guess I did something wrong last time...
Sorry for the false alarm.

apjanke:
Okay, good to hear!

53jk1 (approved):

TheTechRobo:
What is the status on this PR?

PR #5116, opened 2016-05-25
Adding git clone, sudo, and rm -rf to lolcat aliased list

blackmjck:
Thought it needed a few more lols. :-)

### DIFF
--- a/plugins/lol/lol.plugin.zsh
+++ b/plugins/lol/lol.plugin.zsh
@@ -5,6 +5,9 @@ alias wtf='dmesg'
 alias onoz='cat /var/log/errors.log'
 alias rtfm='man'
 
+alias yolo='sudo'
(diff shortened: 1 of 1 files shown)


robjtede:
I personally don't see why this was overlooked.
The `yolo` command is totally worth it by itself.

mcornella:
`yolo` is now a git commit with a [whatthecommit.com](https://whatthecommit.com/) message:
https://github.com/robbyrussell/oh-my-zsh/blob/3d2542f41b8de36877b419f6e74e954d4db06a97/plugins/lol/lol.plugin.zsh#L48

`lulz` is too destructive to be deemed lol-worthy (more like loss-worthy amirite?)

and I'm just too lazy to jerry-rig a commit with just the takez alias.

PR #5117, opened 2016-05-25
Patch/hitchiker fortune

luizpericolo:
Hitchhiker's Guide to the Galaxy fortune in comemoration of Towel Day

### DIFF
--- a/plugins/hitchhiker/.gitignore
+++ b/plugins/hitchhiker/.gitignore
@@ -1,0 +1,1 @@
+fortunes/hitchhiker.dat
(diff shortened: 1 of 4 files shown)


luizpericolo:
Happy towel day 2017!

luizpericolo:
@mcornella Is there something missing for this to be merged?
Thanks!

rachael:
Resurrecting this because it's cute 

@mcornella

mcornella:
Can this download the fortune file on first run instead? It's a biiig file for everyone that won't be using the plugin.

TheTechRobo:
Hm, I'm wondering what place would be best for that download. `~/.oh-my-zsh/custom` ???

mcornella:
Noticed that we have several files above 24k including the one from the chucknorris theme, so it's fine we have one more.

Merged in d8b3e115e968225b5a0b999ea9292ac0263d5586 after several fixes.

PR #5119, opened 2016-05-26
Changed `alias gsta='git stash'` to `alias gsta='git stash save'`.

baig:
Right now doing `gsta "my commit message"` prints the git stash usage instead of stashing the changes with a commit message. This pull request fixes this.


mcornella:
Why did you close this? I was gonna merge it right away

baig:
I usually do `gsta pop ...` and I couldn't do it with this change. Just realized there is `gstp` for that. So reopening it.

mcornella:
Oh, right, I hadn't thought about it. Anyway, most `git stash` possible commands have an alias already, so I'm merging this. Thanks!

PR #5126, opened 2016-05-29
emacs plugin tweak to open buffer from stdin

emanresusername:
[inspired by this](https://github.com/davidshepherd7/emacs-read-stdin/blob/master/emacs-read-stdin.sh)


mcornella:
Can you provide a quick summary of usage info?

PR #5129, opened 2016-05-30
added alias in order to win time

fumseck:
I think that would be helpful

### DIFF
--- a/plugins/git-flow/git-flow.plugin.zsh
+++ b/plugins/git-flow/git-flow.plugin.zsh
@@ -25,6 +25,9 @@ alias gfl='git flow'
 alias gcd='git checkout develop'
 alias gch='git checkout hotfix'
 alias gcr='git checkout release'
+alias gff='git flow feature'
+alias gfh='git flow hotfix'
+alias gfr='git flow release'
 
 _git-flow ()
 {


plugins/git-flow/git-flow.plugin.zsh
```diff
@@ -25,6 +25,9 @@ alias gfl='git flow'
 alias gcd='git checkout develop'
 alias gch='git checkout hotfix'
 alias gcr='git checkout release'
+alias gff='git flow feature'
+alias gfh='git flow hotfix'
+alias gfr='git flow release'
```
mcornella: Can you rename them to `gfl*` so that they match the `gfl` alias (for `git flow`)?

fumseck: ok i do that

fumseck:
I change the alias

plugins/git-flow/git-flow.plugin.zsh
fumseck: I change the alias

robbyrussell:
Hi there, you have a lot of commits for this. Can you squash/resend as a single commit?

mcornella:
Squashed and merged.

PR #5130, opened 2016-05-31
add alias gh=git help

adispring:


mcornella:
I've renamed the alias to `ghh` because it conflicted with several tools, see 76821f50.

PR #5135, opened 2016-06-01
Print out current revision with grev

zackhsi:
```bash
➜  grev git:(master) grev
62efe2c72c40eede2cadbd08b5b4b58cf8b19443
```


zackhsi:
@mcornella ?

mcornella:
Can you comment on an instance when that command is necessary / helpful?

zackhsi:
Alice: "Bob, is your branch up to date?"
Bob: "One sec, let me see ... _checks the output of `grev`_ ... Yep, I'm running 1a2b3c, pulling now."

mcornella:
I'm not convinced. A simple `git show` or `git log` would do, and if you use it that much that you need this I don't think anyone else will benefit because you still have to remember another alias.
Thanks!

PR #5137, opened 2016-06-03
Respect git hide-status config on more functions

rakjin:
Doing `git status`, executed by git plugin, takes long time if the repo is heavy,
so I tried using `oh-my-zsh.hide-status` on my local git config.
However, it had no effect because some `git_prompt_xxx` functions call git commands
regardless of the `hide-status` configuration.
(In my case, it is `git_prompt_status` function, which called from sorin theme)

This patch fixes the case.

On the config name `hide-status`, which means "HIDE STATUS",
but in my humble opinion, users will take it as 'HIDE PROMPT', as I did.
Hiding all git-related prompts upon the config value is appropriate, I think.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -1,11 +1,19 @@
+# Checks hide-status config
+function git_prompt_hidden() {
+  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" == "1" ]]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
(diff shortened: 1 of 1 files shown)


mcornella:
The `git_prompt_status` function is really slow. I've tested the rest of the functions on a big repository and they don't seem to need this, so I've merged https://github.com/ohmyzsh/ohmyzsh/pull/4912. Open a new PR with a reproducible use case if any other function needs it.

PR #5140, opened 2016-06-04
Fix nvm command check

pcdevil:
Issue: When installing `nvm` via [`brew`](http://brew.sh) the `nvm.sh` script is located in the standard brew formula directory rather than in `~/.nvm`. This causes the nvm command check failing.

This merge request solves the problem with the simple `which nvm` command. No matter it's output, the exit code will be non-zero if there is no runnable `nvm` and the script will stop.


pcdevil:
Closing it in favor of #4338.

PR #5141, opened 2016-06-05
Docker Plugin: Display images for push/pull/build

jahands:
This will allow auto-completion of image names when pushing or pulling.
It is especially helpful when using custom registries.

This takes a long time to type out by hand every time:
`registry.example.com/user/project:tag`


mcornella:
Fixed on #5354.

PR #5152, opened 2016-06-11
github plugin: remove hub

blueyed:
It should not be included in the github plugin, because it is a tool by itself
and can cause some issues.

Apart from that it does (up to) 2 unnecessary calls to the `hub`
command.

After all, the integration boils down to:

```
alias git=hub
```

This is not enough for an own plugin, but users should just add this to
their zshrc.


mcornella:
I agree with your general sentiment but I think that removing the alias is at this point a backwards incompatible change, so I've opted for just removing the call to `hub` and having the alias still there. Use `unalias git` or override it with another alias if necessary.

PR #5153, opened 2016-06-12
add 'mvn clean install -DskipTests --offline' alias to speed up maven compile

intsilence:
It doesn't need to check the latest dependency for the most part.


mcornella:
Can you add this in the [mvn plugin README as well](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/mvn/README.md#aliases)?

intsilence:
OK, thanks for reminding!

PR #5155, opened 2016-06-13
Added spring-boot start/stop goals

adriens:
`spring-boot:start` and `spring-boot:stop` are missing see reference below :

http://docs.spring.io/spring-boot/docs/current/maven-plugin/index.html


robbyrussell:
⌛️ Given how old this is, going to assume it's been handled in recent updates to the plugin. If not, we'll consider a rebased PR

PR #5163, opened 2016-06-15
NixOS: Use autojump.sh to figure out location of autojump.zsh

mguentner:
`$HOME/.nix-profile/etc/profile.d/autojump.zsh` does not exist anymore, `autojump.sh` determines the shell and sources the appropriate file.


bashu:
@robbyrussell @mguentner why not merge this one?

PR #5164, opened 2016-06-15
Add ofd() to OSX plugin to open finder on current directory

mssalvatore:


plugins/osx/osx.plugin.zsh
```diff
@@ -151,6 +151,14 @@ function pfd() {
 EOF
 }
 
+function ofd() {
+  osascript 2>/dev/null <<EOF
+    tell application "Finder" 
+      open POSIX file "`pwd`"
```
mcornella: Use `$PWD` instead of running `pwd` here.

mcornella:
Oh wait, I have a feeling that `open_command .` should be enough to open Finder without having to write AppleScript. Can you check?

mssalvatore:
Yeah, looks like it covers it. I'll make a new commit.

mcornella:
Thanks! A simple alias should do the trick. I'll merge it right away.

mssalvatore:
Do you want me to add the alias down at the bottom? What's the convention for oh-my-zsh?

Also, what's the procedure for updating the osx section of https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins with the new alias?

mcornella:
Wherever you feel it looks better, there's not a solid styleguide for that. As to the wiki, you can add it right away. But that should really be a plugin README, so don't worry about it.

mssalvatore:
New pull request: https://github.com/robbyrussell/oh-my-zsh/pull/5165

PR #2023, opened 2013-08-05
[tmux] Added Force UTF-8 configuration option

d1823:
Hi! I've added an option to force enable utf8 usage in tmux, since quite a big number of people is putting utf8 enabling line into their tmux.conf file. That change might make it a little bit more convenient.

PR #5166, opened 2016-06-16
tmux plugin: Added support for forced unicode support (-u)

PeterMosmans:
When ZSH_TMUX_UNICODE=true then tmux is informed that UTF-8 is supported.

See also #5083 - it is probably cleaner to somehow merge these requests, so that the executable and optional parameters can be specified as parameter.


hjpotter92:
@mcornella bump! There are also #2023 #5776 and #6737 .

larson-carter:
@PeterMosmans Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

PeterMosmans:
Hi @larson-carter I rebased the PR onto the master branch, per your request.

plugins/tmux/tmux.plugin.zsh
```diff
@@ -12,6 +12,8 @@ alias tl='tmux list-sessions'
 alias tksv='tmux kill-server'
 alias tkss='tmux kill-session -t'
 
+
+=======
```
mcornella: This throws a not found error.

plugins/tmux/tmux.plugin.zsh
```diff
@@ -68,8 +71,9 @@ function _zsh_tmux_plugin_run() {
   # If failed, just run tmux, fixing the TERM variable if requested.
   if [[ $? -ne 0 ]]; then
     [[ "$ZSH_TMUX_FIXTERM" == "true" ]] && tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG") || \
-    [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")
-    $tmux_cmd new-session  
+    [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(-f "$ZSH_TMUX_CONFIG") || \
+    [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(" -u")
```
mcornella: Shouldn't this option be applied further up, like before the first tmux attach? I don't see it making sense here, after what is essentialy and else statement. Also, it checks `$ZSH_TMUX_CONFIG` instead of `$ZSH_TMUX_UNICODE`.

PeterMosmans: You're totally right, my bad. That will teach me using a different zsh plugin manager :)

plugins/tmux/tmux.plugin.zsh
PeterMosmans: Fixed

PR #5167, opened 2016-06-16
clone bureau, but without nvm info

khebbie:
I like the bureau theme, but don't use node, so I cloned the theme and removed nvm info.


khebbie:
I would close this one as #5143 is a better solution

PR #5169, opened 2016-06-16
installer: fix, refactor and add installer arguments

### LINKED ISSUE #2499: Some minor alteration concerning .zshrc.pre-oh-my-zsh

### LINKED ISSUE #2586: install.sh: `PATH=` with no `$PATH` inside

### LINKED ISSUE #3447: oh-my-zsh uninstall.sh issues - switch back to bash is unsatisfactory

### LINKED ISSUE #4131: Does not change default shell to zsh on NixOS
Shows following after installation of oh my zsh;

Time to change your default shell to zsh!\033[0m
Password: 
Cannot change ID to root.

### LINKED ISSUE #4317: Should export PATH contain $PATH in .zshrc?
To make [RVM](https://rvm.io/) work, I had to add **:$PATH** at the end of PATH in .zshrc:

```
export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:$PATH"
```

Shouldn't that be a default?

### LINKED ISSUE #4390: Please, oh please, ask, or at least warn, about overwriting ~/.zshrc on installation

### LINKED ISSUE #4393: [Discuss] What should installer behavior be, esp wrt launching new shells?
We've had some issues around the installer scripts and their interactive behavior for a while, not just with the implementation, but with the design and desired behavior. Let's hash out exactly what we want it to do, and why.

This issue is specifically _not_ about how `$PATH`  is altered by the constructor; I don't want that to distract from the other issues. `$PATH` is important, but it is out of scope for this issue. See #1359 for discussion on that.
### Questions to consider
- Should `install.sh` and `uninstall.sh` launch new interactive shell sessions as part of the installation?
  - If so, should they `exec`, or just run as sub-processes?
- How should the scripts be invoked, and what invocation methods should be supported?
  - Should they require OMZ to run?
- Should there be an non-interactive installation option?
- How about uniform diagnostics: always print errors on `stderr`, and exit non-zero upon error?
- Any other aspects of their behavior to consider or change?
### References
- #3447: oh-my-zsh uninstall.sh issues - switch back to bash is unsatisfactory
- #3664: printf/terminfo portability – has a long discussion about the subshell invocation
- #4355: Tracking: install script related issues
#### Invocation

It's not clear exactly how some of the `tools/*.sh` files can be run. E.g. for uninstallation, the README says to "run uninstall_oh_my_zsh from the command-line", which runs `tools/uninstall.sh` under `sh`. Can `uninstall.sh` also be run on its own? I think that would be nice, so you don't depend on a running OMZ instance to do the installation. But it has now shebang line and isn't executable.

Most of the files in `tools/` look like scripts to be run as external commands. I think it would be nice to support that where possible. And for scripts where it is supported, the files should be `+x` and have a `#!` at the top to indicate which shell they run under and support direct invocation.
#### On launching subshells and sourcing

If you run an interactive subshell (or any other interactive subprocess), then the behavior of the scripts is going to differ depending on whether you use the `curl ... | sh` or `sh -c "$(curl ...)"` form, because the subprocess takes over stdin.

If you launch a subshell from a script that's run directly, instead of fully pre-buffered using `$(cat ...)`, and the script is modified while the interactive subshell is still running, it can have unpredicatble behavior and/or errors after the subshell exits.

Sourcing `/etc/profile` and `~/.zshrc` files inside running sessions is problematic. They're not designed to be run multiple times, and you don't have a clean sesion because you still have state left over from the original startup file runs.

`uninstall.sh` sources `/etc/profile` or `.zshrc`, but it doesn't detect what shell it's running under.

I think the `source` calls should be removed from `uninstall.sh` completely; it doesn't know which is the right one to invoke, and can't end up with a clean shell environment anyway. It should just tell you to restart your shell session.
#### On uninstallation
- Should probably check for local changes in `~/.oh-my-zsh` before blowing it away, to avoid destroying user data. In fact, should maybe check to make sure it points to an OMZ repo.
#### Miscellaneous

As a sysadmin and a tester of OMZ, I always appreciate a fully noninteractive option for installation and management scripts, so they can easily be run repeatedly or scripted themselves.

We should consider the case of people who have their config files like `~/.zshrc` synced across multiple machines or otherwise under source control.

### LINKED ISSUE #4818: shell '/usr/bin/zsh' does not exist
Hi there, I'm setting a new machine and probably missed something.

I installed homebrew, git, zsh, then this.  During the installation I saw the warning:

```
chsh: WARNING: shell '/usr/bin/zsh' does not exist
```

and when I try to open a new session I see:

```
login: /usr/bin/zsh: No such file or directory
```

but it does look like I have zsh installed:

```sh
➜  ~  which zsh
/usr/local/bin/zsh
```

Why is the install script trying to `chsh` to `/usr/bin/zsh` as opposed to `/usr/local/bin/zsh`?

### LINKED ISSUE #5002: In the installer, add a way to "not" launch zsh after finishing
In the installer, add a way to "not" launch zsh after finishing.

It could be an environment variable that is being checked or whatever way you find appropriate.

I have my own "install_ubuntu.sh" that, in the middle, invokes OMZ installer, and I have to type exit everytime I want my installer to continue.

One way would be to leave it to the end.. but the installer should not just "nest" a shell call if you don't want to...

### LINKED ISSUE #5277: Install script set my shell to /usr/local/bin/zsh instead of /bin/zsh
The installer set my shell to `/usr/local/bin/zsh`, but I don't even have brew installed and that path doesn't exist.

I was able to fix this by manually settings my shell to /bin/bash in terminal preferences, then running `chsh -s /bin/zsh` and everything is working as expected now.

I should also mention that I am using macOS Sierra beta2

### LINKED ISSUE #5628: Zsh installer defaults to main repo even if installed from fork sync
I bought a new laptop and stumbled on this when setting it up. 
I made a Pr to pull from origin master when [-d .git] is true which should fix this.

### LINKED ISSUE #6119: uninstall zsh didn't revert shell back to bash on OSX
After running the uninstall command for zsh and restarting the terminal, I noted the zsh shell was still in use afterward. 

I had to use:

`chsh -s /bin/bash`

and restart the terminal to get the bash shell back.

I didn't think this was the expected result from the uninstall so I thought I would log an issue.

OSX Sierra 10.12.5

### LINKED ISSUE #6299: Uninstalling oh my zsh does NOT restore previous default shell
As title. Please fix your $#@!@#$ uninstall script, and excuse my language.

Running Xubuntu 14.10.

### LINKED ISSUE #7100: Suggestion: add a flag to bypass setting zsh as default shell
Currently oh-my-zsh install.sh script will try to set the default shell to zsh. But sometimes this will prompt the user to enter their password. This is somewhat annoying in auto-deploying scripts. I would suggest to add a flag to `install.sh` to bypass this section. Or is there already one? Thanks. 

```sh
  # If this user's login shell is not already "zsh", attempt to switch.
  TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
  if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
    # If this platform provides a "chsh" command (not Cygwin), do it, man!
    if hash chsh >/dev/null 2>&1; then
    # Else, suggest the user do so manually.
    else
      printf "I can't change your shell automatically because this system does not have chsh.\n"
      printf "${BLUE}Please manually change your default shell to zsh!${NORMAL}\n"
    fi
  fi

```

### LINKED ISSUE #7721: Installation script shouldn't alter default shell
## Summary
In the installation script there is a section that attempts to alter the default shell during installation. I don't think this should be the default behaviour as this is quite an intrusive change when you may just want to try out zsh + oh-my-zsh.

What do you think?

## Related
https://github.com/robbyrussell/oh-my-zsh/issues/4261

### LINKED ISSUE #7865: Make setting the default shell during installation optional
Setting the default shell to`/bin/zsh` has been causing me some problems. Particularly on linux systems, there is a problem that comes up where having `zsh` as the default shell causes snap packages to misbehave and not show in the application launcher. Changing the default shell back to `/bin/bash` fixes the problem.

https://forum.snapcraft.io/t/desktop-snaps-do-not-appear-in-the-dash-menu-on-gnome-in-17-10-final-beta/2340/13

Furthermore, there doesn't seem to be any way to stop this behavior from happening. Other than killing the install process when `chsh` is prompting for your password. Although, this may not even work if you've already granted `sudo` permissions on your current session (haven't tested that out).

Prompt user with a choice, if they would like to change their default shell.

Many users may not even know what this means, so an explanation of what changing this actually means would be a nice touch.

Also, during this prompt to change the default shell, there could be an explanation on configuring your terminal emulator to start `zsh` instead of `bash`. Since the main purpose of setting the default shell is to make `zsh` execute immediately when opening a terminal. Maybe that should be the approach instead of trying to change something on the system level?

**Additional context**
Some other related issues have been raised in the past about `oh-my-zsh`s setting of the default shell, both during the install and uninstall process. Where the uninstall script just assigns the default shell back to `/bin/bash` instead of the user's preferred shell (e.g. `fish`)
https://github.com/robbyrussell/oh-my-zsh/issues/7721

mcornella:
Ready for testing

This is an all-encompassing PR to fix once and for all all issues related to the OMZ installer, as well as adding features requested many times.

---
#### `zshrc` file mangling

- [x] #2499: Some minor alteration concerning `.zshrc.pre-oh-my-zsh`
- [x] #4390: Please, oh please, ask, or at least warn, about overwriting ~/.zshrc on installation PR
- [x] #4391: installer: use timestamped backups to preserve all old zshrcs

Fixes #2499. Fixes #4390. Fixes #4391.

#### Change of default shell issues

- [x] #4131: Does not change default shell to zsh on NixOS
- [x] #4564: Handle systems where /etc/shells does not exist
- [x] [#4818](https://github.com/robbyrussell/oh-my-zsh/issues/4818#issuecomment-178106951): installer doesn't check if the zsh obtained from `/etc/shells` exists
- [x] #5277: Install script set my shell to /usr/local/bin/zsh instead of /bin/zsh
- [x] #6569: Patch for detect shell file in Solus OS
- [x] #7254: Make sure the line begins with a (forward) slash ('/')

Fixes #4131. Fixes #4564. Fixes #4818. Fixes #5277. Fixes #6569. Fixes #7254.

#### Restore default shell

- [x] #3447: oh-my-zsh uninstall.sh issues - switch back to bash is unsatisfactory
- [x] #6119: uninstall zsh didn't revert shell back to bash on OSX
- [x] #6299: Uninstalling oh my zsh does NOT restore previous default shell
- [x] #6356: Fix for #6299: restore previous default shell with uninstall
- [x] #6608: Uninstall restores now the original shell

Fixes #3447. Fixes #6119. Fixes #6299. Fixes #6356. Fixes #6608.

#### `$PATH` management

- [x] #1359: Do not force PATH in the installer
- [x] #2586: install.sh: `PATH=` with no `$PATH` inside
- [x] #2589: Respect original PATH definitions on install
- [x] #4317: Should export PATH contain $PATH in .zshrc?
- [x] #4925: Remove undesirable hardcoding of PATH into zshrc

Fixes #1359. Fixes #2586. Fixes #4925. Closes #2589. Closes #4317.

#### Skip change of default shell

- [x] #4261: Prevent shell change during setup
- [x] #7100: Suggestion: add a flag to bypass setting zsh as default shell
- [x] #7266: Adds skip-chsh flag to switch login shell step
- [x] #7721: Installation script shouldn't alter default shell
- [x] #7865: Make setting the default shell during installation optional

Fixes #4621. Fixes #7100. Fixes #7266. Fixes #7721. Fixes #7865.

The installer respects the environment variable `CHSH`. Set to `no`, it will skip the chsh step. It also reads the flag `--skip-chsh` which has the same effect. For instance, these two are equivalent:
```sh
CHSH=no sh install.sh
sh install.sh --skip-chsh
```

#### Don't run zsh after install

- [x] #4393: [Discuss] What should installer behavior be, esp wrt launching new shells?
- [x] #5002: add a way to not launch zsh after finishing
- [x] #5853: Add a flag to prevent the "launch" of zsh
- [x] #5893: Allow batch mode installation
- [x] #6547: Add support for --silent option in install.sh script
- [x] #7790: Add non-interactive option to install script

Fixes #4393. Fixes #5002. Fixes #5853. Fixes #5893. Fixes #6547. Fixes #7790.

As in the skip-chsh feature above, the installer respects the variable `RUNZSH`, which when set to `no` will skip the zsh call when the install is done. The flag `--unattended` also sets `RUNZSH=no`, as well as setting `CHSH=no`. So these two are equivalent:
```sh
RUNZSH=no CHSH=no sh install.sh
sh install.sh --unattended
```

#### Forks

- [x] #3648: Installer: add ability to install from forked repos
- [x] #5628: Zsh installer defaults to main repo even if installed from fork sync
- [x] #7496: make installation script configurable for forks

Fixes #3648. Fixes #5628. Fixes #7496.

The installer respects the following environment variables, that can be set up when calling it with `VAR=value sh install.sh`.

- `REPO`: value in the form of `owner/project`. Useful for forks in GitHub (defaults to `robbyrussell/oh-my-zsh).
- `BRANCH`: default branch to check out once the clone has finished (defaults to master).
- `REMOTE`: full remote URL (supersedes `REPO` if set). You could for example choose an ssh remote URL, like `git@github.com:user/oh-my-zsh`, or another fork from GitLab or Bitbucket.

### Other

- [x] #5194: installer: git clone using ssh URL if https fails
- [x] #5320: Greatly improve the OMZ install.sh
- [x] #7007: Replace the current shell with the new ZSH instance

Fixes #5320. Fixes #7007.
Closes #5194.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1,114 +1,272 @@
-main() {
-  # Use colors, but only if connected to a terminal, and that terminal
-  # supports them.
-  if which tput >/dev/null 2>&1; then
-      ncolors=$(tput colors)
-  fi
-  if [ -t 1 ] && [ -n "$ncolors" ] && [ "$ncolors" -ge 8 ]; then
-    RED="$(tput setaf 1)"
-    GREEN="$(tput setaf 2)"
-    YELLOW="$(tput setaf 3)"
-    BLUE="$(tput setaf 4)"
-    BOLD="$(tput bold)"
-    NORMAL="$(tput sgr0)"
-  else
-    RED=""
-    GREEN=""
-    YELLOW=""
-    BLUE=""
-    BOLD=""
-    NORMAL=""
-  fi
+#!/bin/sh
+#
+# This script should be run via curl:
+#   sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
+# or wget:
+#   sh -c "$(wget -qO- https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
+#
+# As an alternative, you can first download the install script and run it afterwards:
+#   wget https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh
+#   sh install.sh
+#
+# You can tweak the install behavior by setting variables when running the script. For
+# example, to change the path to the Oh My Zsh repository:
+#   ZSH=~/.zsh sh install.sh
+#
+# Respects the following environment variables:
+#   ZSH     - path to the Oh My Zsh repository folder (default: $HOME/.oh-my-zsh)
+#   REPO    - name of the GitHub repo to install from (default: robbyrussell/oh-my-zsh)
+#   REMOTE  - full remote URL of the git repo to install (default: GitHub via HTTPS)
+#   BRANCH  - branch to check out immediately after install (default: master)
+#
+# Other options:
+#   CHSH    - 'no' means the installer will not change the default shell (default: yes)
+#   RUNZSH  - 'no' means the installer will not run zsh after the install (default: yes)
+#
+# You can also pass some arguments to the install script to set some these options:
+#   --skip-chsh: has the same behavior as setting CHSH to 'no'
+#   --unattended: sets both CHSH and RUNZSH to 'no'
+# For example:
+#   sh install.sh --unattended
+#
+set -e
 
-  # Only enable exit-on-error after the non-critical colorization stuff,
-  # which may fail on systems lacking tput or terminfo
-  set -e
+# Default settings
+ZSH=${ZSH:-~/.oh-my-zsh}
+REPO=${REPO:-robbyrussell/oh-my-zsh}
+REMOTE=${REMOTE:-https://github.com/${REPO}.git}
+BRANCH=${BRANCH:-master}
 
-  if ! command -v zsh >/dev/null 2>&1; then
-    printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
-    exit
-  fi
-
-  if [ ! -n "$ZSH" ]; then
-    ZSH=~/.oh-my-zsh
-  fi
-
-  if [ -d "$ZSH" ]; then
-    printf "${YELLOW}You already have Oh My Zsh installed.${NORMAL}\n"
-    printf "You'll need to remove $ZSH if you want to re-install.\n"
-    exit
-  fi
-
-  # Prevent the cloned repository from having insecure permissions. Failing to do
-  # so causes compinit() calls to fail with "command not found: compdef" errors
-  # for users with insecure umasks (e.g., "002", allowing group writability). Note
-  # that this will be ignored under Cygwin by default, as Windows ACLs take
-  # precedence over umasks except for filesystems mounted with option "noacl".
-  umask g-w,o-w
-
-  printf "${BLUE}Cloning Oh My Zsh...${NORMAL}\n"
-  command -v git >/dev/null 2>&1 || {
-    echo "Error: git is not installed"
-    exit 1
-  }
-  # The Windows (MSYS) Git is not compatible with normal use on cygwin
-  if [ "$OSTYPE" = cygwin ]; then
-    if git --version | grep msysgit > /dev/null; then
-      echo "Error: Windows/MSYS Git is not supported on Cygwin"
-      echo "Error: Make sure the Cygwin git package is installed and is first on the path"
-      exit 1
-    fi
-  fi
-  env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git "$ZSH" || {
-    printf "Error: git clone of oh-my-zsh repo failed\n"
-    exit 1
-  }
+# Other options
+CHSH=${CHSH:-yes}
+RUNZSH=${RUNZSH:-yes}
 
 
-  printf "${BLUE}Looking for an existing zsh config...${NORMAL}\n"
-  if [ -f ~/.zshrc ] || [ -h ~/.zshrc ]; then
-    printf "${YELLOW}Found ~/.zshrc.${NORMAL} ${GREEN}Backing up to ~/.zshrc.pre-oh-my-zsh${NORMAL}\n";
-    mv ~/.zshrc ~/.zshrc.pre-oh-my-zsh;
-  fi
-
-  printf "${BLUE}Using the Oh My Zsh template file and adding it to ~/.zshrc${NORMAL}\n"
-  cp "$ZSH"/templates/zshrc.zsh-template ~/.zshrc
-  sed "/^export ZSH=/ c\\
-  export ZSH=\"$ZSH\"
-  " ~/.zshrc > ~/.zshrc-omztemp
-  mv -f ~/.zshrc-omztemp ~/.zshrc
-
-  # If this user's login shell is not already "zsh", attempt to switch.
-  TEST_CURRENT_SHELL=$(basename "$SHELL")
-  if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
-    # If this platform provides a "chsh" command (not Cygwin), do it, man!
-    if hash chsh >/dev/null 2>&1; then
-      printf "${BLUE}Time to change your default shell to zsh!${NORMAL}\n"
-      chsh -s $(grep /zsh$ /etc/shells | tail -1)
-    # Else, suggest the user do so manually.
-    else
-      printf "I can't change your shell automatically because this system does not have chsh.\n"
-      printf "${BLUE}Please manually change your default shell to zsh!${NORMAL}\n"
-    fi
-  fi
-
-  printf "${GREEN}"
-  echo '         __                                     __   '
-  echo '  ____  / /_     ____ ___  __  __   ____  _____/ /_  '
-  echo ' / __ \/ __ \   / __ `__ \/ / / /  /_  / / ___/ __ \ '
-  echo '/ /_/ / / / /  / / / / / / /_/ /    / /_(__  ) / / / '
-  echo '\____/_/ /_/  /_/ /_/ /_/\__, /    /___/____/_/ /_/  '
-  echo '                        /____/                       ....is now installed!'
-  echo ''
-  echo ''
-  echo 'Please look over the ~/.zshrc file to select plugins, themes, and options.'
-  echo ''
-  echo 'p.s. Follow us at https://twitter.com/ohmyzsh'
-  echo ''
-  echo 'p.p.s. Get stickers, shirts, and coffee mugs at https://shop.planetargon.com/collections/oh-my-zsh'
-  echo ''
-  printf "${NORMAL}"
-  env zsh -l
+command_exists() {
+	command -v "$@" >/dev/null 2>&1
 }
 
-main
+error() {
+	echo ${RED}"Error: $@"${RESET} >&2
+}
+
+# Set up color sequences
+setup_color() {
+	ncolors=$(tput colors 2>/dev/null) || ncolors=0
+
+	# Only use colors if connected to a terminal that supports them
+	if [ -t 1 ] && [ $ncolors -ge 8 ]; then
+		RED="$(tput setaf 1)"
+		GREEN="$(tput setaf 2)"
+		YELLOW="$(tput setaf 3)"
+		BLUE="$(tput setaf 4)"
+		BOLD="$(tput bold)"
+		RESET="$(tput sgr0)"
+	else
+		RED=$(printf '\033[31m')
+		GREEN=$(printf '\033[32m')
+		YELLOW=$(printf '\033[33m')
+		BLUE=$(printf '\033[34m')
+		BOLD=$(printf '\033[1m')
+		RESET=$(printf '\033[m')
+	fi
+}
+
+setup_ohmyzsh() {
+	# Prevent the cloned repository from having insecure permissions. Failing to do
+	# so causes compinit() calls to fail with "command not found: compdef" errors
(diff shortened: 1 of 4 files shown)


tools/install.sh
```diff
@@ -1,116 +1,151 @@
...
+	export ZSH=$ZSH
+	" ~/.zshrc > ~/.zshrc-omztemp
+	mv -f ~/.zshrc-omztemp ~/.zshrc
+
+setup_shell() {
+	# If this user's login shell is not already "zsh", attempt to switch.
+	TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
+	if [ "$TEST_CURRENT_SHELL" != "zsh" ]; then
+		# If this platform provides a "chsh" command (not Cygwin), do it, man!
+		if command_exists chsh; then
```
fornwall: This assumes that `/etc/shells` exists if `chsh` exists, which is not the case on all systems (Android).

Could you either also check for `/etc/shells` here, or try `which zsh` to find the path before trying `$(grep /zsh$ /etc/shells | tail -1)` inside the if clause?

tools/install.sh
```diff
@@ -1,116 +1,151 @@
...
+	fi
+
+	echo "${BLUE}Using the Oh My Zsh template file and adding it to ~/.zshrc.${NORMAL}"
+
+	cp $ZSH/templates/zshrc.zsh-template ~/.zshrc
+	sed "/^export ZSH=/ c\\
+	export ZSH=$ZSH
+	" ~/.zshrc > ~/.zshrc-omztemp
+	mv -f ~/.zshrc-omztemp ~/.zshrc
+
```
mcornella: To do: fix missing `}`

tools/install.sh
v0idst4r: #6398 should fix this

lhorace:
Hello, 

I'm not sure if I should report an issue here or create a new issue? Mean time, testing the new installer fails here with Ansible.

```
tput: unknown terminal \"unknown\"
``` 

Is there a way around this?

mcornella:
Hey there, thanks for the bug report. Can you check now if it works without the workaround?

mrk-han:
For what it's worth: `cat /etc/shells` does not show `/usr/local/bin/zsh` but `which zsh` does show `/usr/local/bin/zsh`. Given this, will `chsh -s $(which zsh)` change the behavior for me?

PR #5176, opened 2016-06-19
Gb plugin, http://getgb.io 

dahankzter:
I have been doing a lot of Go coding lately and I found myself sorely missing Gb completion support.
A lot of copy pasting and adapting from other plugins but it seems to work now.

### DIFF
--- a/plugins/getgb/README.md
+++ b/plugins/getgb/README.md
@@ -1,0 +1,10 @@
+# *gb* plugin
+
+- aims to support completion for all *gb* commands
+- also supports completion for *gb* plugins such as *vendor*
+
+See http://getgb.io for the full *gb* documentation
+
+## Caveats
+
+- The *git* plugin installs an alias *gb* that usually conflicts with the *gb* program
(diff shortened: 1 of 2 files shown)


dahankzter:
Ping @robbyrussell this branch has not rotted yet.
Not sure who merges these days sorry if it's not you.

robbyrussell:
Can you provide a README for this, too?

dahankzter:
Absolutely, will do asap!

dahankzter:
README added. It is small but covers an important caveat that the Gb user probably knows already but still important.

dahankzter:
Does this README cover it @robbyrussell ? I can expand it but the GB site is really good.

mcornella:
I'll make a couple of adjustments and merge it right away. Thanks!

dahankzter:
Awesome!

mcornella:
Almost done. @dahankzter is there a reason behind naming the plugin `getgb` instead of `gb`? If not, I'll rename it to `gb` which is less confusing.

dahankzter:
The only reason was to not confuse with the git alias 'gb'. Gb is probably
better.

mcornella:
@dahankzter I've submitted #5207 which adds a few improvements. Please take a look at it and I'll merge it if you +1 that.

dahankzter:
+1 lets merge

PR #5177, opened 2016-06-20
Fix issue with `dash` as the default `/bin/sh` script executing program.

hjpotter92:
* `function` keyword removed
 * Fixup the `if` statement and operators

### DIFF
--- a/plugins/emacs/emacsclient.sh
+++ b/plugins/emacs/emacsclient.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-function _emacsfun
+_emacsfun()
 {
     # get list of available X windows.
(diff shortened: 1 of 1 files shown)


plugins/emacs/emacsclient.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
```
mcornella: I'd rather just change this one line here to `#!/usr/bin/env zsh`.

mcornella: Or not, it doesn't matter. Thanks for the PR

mcornella:
Thanks!

PR #5180, opened 2016-06-20
bundler plugin: Simplify retrieval of CPU count on OSX

clemensg:
Calling awk is not necessary here, sysctl has the `-n` flag to print the value.


mcornella:
That's great thanks! [Source](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man8/sysctl.8.html).

PR #5183, opened 2016-06-22
Add Alias To Discard All Unstaged Files

ramytamer:
Added alias `gda` => _git discard all_ to checkout all **unstaged** files & folders in the project.


mcornella:
`gco .` is not hard to remember.

PR #5188, opened 2016-06-27
Fix dnf plugin for Fedora users

awd123:
On Fedora, all dnf commands require root.


mcornella:
Can't reproduce on my end. Those commands all work under non-privileged users.

PR #5189, opened 2016-06-27
Agnoster: add current node version if .nvmrc is present

sheershoff:
Useful with nvm autoswitcher upon directory change (as described in nvm package).
A hexagon icong with node version. As less cluttering as possible.


robbyrussell:
Believe current convention is to lean on `.node-version` files. Going to close this for now

PR #5190, opened 2016-06-27
Remove _brew zsh completion

joshka:
Removing as this is now installed as part of homebrew automatically and this
version is out of date. See
https://github.com/Homebrew/brew/blob/master/share/zsh/site-functions/_brew


joshka:
Bump!

mcornella:
There are conflicts. Can you rebase?

mcornella:
Fixed in 5cf9263. Thanks for everything!

PR #5191, opened 2016-06-27
Deprecating brew cask plugin

joshka:
I'm suggesting that this be deprecated in favor of the modified script in the
zsh-completions project. See https://github.com/joshka/zsh-completions/commit/31b4e82887626eee236a9806ec9c4f5dee90056c

A copy of the relevant part of the commit message of the above is included here:

The original plugin suffers from the following issues (IMO):
1. it is part of oh-my-zsh which some have chosen not to use
2. it acts as a script overriding the existing brew completion rather than as a
standard autoloadable file, this means it must be manually installed rather than
automatically installed as part of compinit.

The changes from the plugin are to fix item 2, as well as update the commands to
those current in homebrew cask. This change relies on a change pending in the
homebrew completion to allow it to call subcommand completions (see
https://github.com/joshka/brew/commit/6b010c5700095685bad6797038e83ef558df5104)
and hence should not be merged until that has been merged.

Ideally, this completion would be installed as part of the homebrew cask
installation, but there doesn't seem to be an easy way to do so as it is
homebrew cask is installed as a homebrew tap. It also doesn't really fit in the
homebrew main repo for similar reasons.

See related PRs:
https://github.com/zsh-users/zsh-completions/pull/430
https://github.com/Homebrew/brew/pull/407


mcornella:
My understanding is that the completion will be shipped within homebrew or `brew cask` or whatever installs cask, so this won't be necessary anymore.

Blocked by https://github.com/Homebrew/brew/pull/407. Let me know when this is good to merge!

joshka:
Yep, still waiting on that and https://github.com/Homebrew/homebrew-completions/pull/135 to be merged. If you can suggest a better way to notify oh-my-zsh users of the deprecation than just removing the plugin, I'm all for it. Perhaps replace with a message on where to go or something similar?

joshka:
@mcornella Homebrew/brew#407 is now merged, so ready to remove this plugin.

mcornella:
Ok let me review the upgrade process here. Perhaps we could check if the official completion exists and display a warning advising to disable the plugin.

mcornella:
Would you say that `brew cask` gets updated frequently or is it a user's choice?

joshka:
It's a user's choice to update homebrew (core updates cask automatically). Perhaps replacing this with a warning message to install from elsewhere is correct for your users.

I'm mostly about trying not to cause a loop where `_brew` calls `_brew_cask` which calls `_brew`

joshka:
https://github.com/Homebrew/brew/pull/936 was merged in brew directly, which means every user that has Homebrew cask has more up to date completions than oh-my-zsh provides.

Hence I'd suggest that deprecating this directly is appropriate. I'm not quite sure how omz handles deleted plugins (I use prezto). Feel free to go ahead and replace this with some sort of message on omz startup instead perhaps if it breaks things.

mcornella:
Alright, merging this then.

PR #5194, opened 2016-06-30
installer: git clone using ssh URL if https fails

pierrickrouxel:
I added IBM i support. Just few changes.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -5,7 +5,7 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 # Enable ls colors
 if [ "$DISABLE_LS_COLORS" != "true" ]
 then
-  # Find the option for using colors in ls, depending on the version: Linux, BSD or IBM i
+  # Find the option for using colors in ls, depending on the version: Linux or BSD
   if [[ "$(uname -s)" == "NetBSD" ]]; then
     # On NetBSD, test if "gls" (GNU ls) is installed (this one supports colors);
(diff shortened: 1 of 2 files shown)


dabou:
oh yes, oh-my-zsh on IBM i !
Thank's :tada:

lib/theme-and-appearance.zsh
```diff
@@ -17,6 +17,8 @@ then
     # coreutils, so prefer it to "gls".
     gls --color -d . &>/dev/null 2>&1 && alias ls='gls --color=tty'
     colorls -G -d . &>/dev/null 2>&1 && alias ls='colorls -G'
+  elif [[ "$(uname -s)" == "OS400" ]]; then
```
apjanke: Could this be done with a generic test of whether `ls -G` is supported, instead of adding an OS-specific test?

pierrickrouxel:
Ok I will do that.

apjanke:
See also #4988, which looks like it's attempting the same thing.

pierrickrouxel:
I removed changes on ls appearance. The #4988 solves the problem more efficiently.

mcornella:
The HTTPS-to-SSH remote URL change won't be needed now that #5169 introduces the ability to change the remote URL via the `REMOTE` variable (so you could run `REMOTE=git@github.com:robbyrussell/oh-my-zsh.git sh install.sh`, for example). This PR will be automatically closed when #5169 is merged.

mcornella:
As I said above, #5169 adds the ability to configure the remote URL, so you can now use an SSH remote. 

Thanks!

mcornella:
Closed by #5169.

PR #5198, opened 2016-07-01
aws: Handle errors when aws_zsh_completer.sh not found

azmodude:
- Redirect `which` stderr output to /dev/null
- Check for empty `_aws_zsh_completer_path` variable

When awscli is not installed or the `aws_zsh_completer.sh` script is not found in the user's path, firing up zsh and having the plugin loaded results in ugly
- `/usr/bin/which: no aws_zsh_completer.sh in ...` and
- `.../plugins/aws/aws.plugin.zsh:source:36: not enough arguments` errors.

This patch does not solve the potential issue of not finding the script (Fedora puts it in `/usr/share/zsh/site-functions`, while e.g. Arch puts it in `/usr/bin`), but at least the plugin does no longer spit out errors on every load.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -30,8 +30,8 @@ compctl -K aws_profiles asp
 if _homebrew-installed && _awscli-homebrew-installed ; then
   _aws_zsh_completer_path=$(brew --prefix awscli)/libexec/bin/aws_zsh_completer.sh
 else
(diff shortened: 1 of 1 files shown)


jannotti:
I hadn't seen this, so I did something similar in PR #5264.

I also added a check for the common location of aws_zsh_completer.sh on AWS ec2 machines.  It'd be nice to have that work too.  (Maybe there should be a list of common locations, since "which" really isn't a good way to find such a file).

mcornella:
Fixed in #6745

PR #5201, opened 2016-07-02
Updated emacs plugin

### LINKED ISSUE #4221: emacs plugins with emacs >= 24.4
Hi,

Using emacs plugins with emacs >= 24.4 trigger an unexpected close whatever the command you are using (e, emacs, emacs -nw, ...).

soredake:
1. updated minimal version from 23 to 24
2. added new aliase that fix unexpected close with -nw option (fixes #4221)
3. removed some old aliases
4. fixed line that detects open frames, now it works

### DIFF
--- a/plugins/emacs/emacs.plugin.zsh
+++ b/plugins/emacs/emacs.plugin.zsh
@@ -10,7 +10,7 @@
 # - Configuration changes made at runtime are applied to all frames.
 
 
-if "$ZSH/tools/require_tool.sh" emacs 23 2>/dev/null ; then
+if "$ZSH/tools/require_tool.sh" emacs 24 2>/dev/null ; then
     export EMACS_PLUGIN_LAUNCHER="$ZSH/plugins/emacs/emacsclient.sh"
 
     # set EDITOR if not already defined.
@@ -18,15 +18,14 @@ if "$ZSH/tools/require_tool.sh" emacs 23 2>/dev/null ; then
 
     alias emacs="$EMACS_PLUGIN_LAUNCHER --no-wait"
     alias e=emacs
+    # open terminal emacsclient
+    alias te="$EMACS_PLUGIN_LAUNCHER -nw"
 
     # same than M-x eval but from outside Emacs.
     alias eeval="$EMACS_PLUGIN_LAUNCHER --eval"
     # create a new X frame
     alias eframe='emacsclient --alternate-editor "" --create-frame'
 
-    # to code all night long
-    alias emasc=emacs
-    alias emcas=emacs
 
     # Write to standard output the path to the file
     # opened in the current buffer.
--- a/plugins/emacs/emacsclient.sh
+++ b/plugins/emacs/emacsclient.sh
@@ -2,15 +2,15 @@
 
 _emacsfun()
 {
-    # get list of available X windows.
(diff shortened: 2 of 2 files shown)


mcornella:
Can you describe what this does in detail?

mcornella:
Thanks! @hjpotter92 @emanresusername can you test this and comment on it?

plugins/emacs/emacs.plugin.zsh
```diff
@@ -10,23 +10,22 @@
...
 
-if "$ZSH/tools/require_tool.sh" emacs 23 2>/dev/null ; then
+if "$ZSH/tools/require_tool.sh" emacs 24 2>/dev/null ; then
     export EMACS_PLUGIN_LAUNCHER="$ZSH/plugins/emacs/emacsclient.sh"
 
     # set EDITOR if not already defined.
     export EDITOR="${EDITOR:-${EMACS_PLUGIN_LAUNCHER}}"
 
     alias emacs="$EMACS_PLUGIN_LAUNCHER --no-wait"
+    alias temacs="$EMACS_PLUGIN_LAUNCHER"
```
hjpotter92: I think you meant to have a `-nw` option here as well.

emanresusername:
i can't claim to be an emacsclient expert, but seems to be working great for me.
running `e whatever.file` with no frames open now does the expected thing (i.e. open and switch to a new frame visiting `whatever.file`. instead of before where it would just do . . . well . . . nothing :/) so 👍 for that!
maybe some comments for the usecases for the new aliases? it seems like they're trying to do different things (correctly i'm sure), it's just not apparent to me what each one is meant for?

plugins/emacs/emacs.plugin.zsh
soredake: fixed

plugins/emacs/emacs.plugin.zsh
```diff
@@ -10,23 +10,22 @@
...
     alias e=emacs
+    # open terminal emacsclient
+    alias te="$EMACS_PLUGIN_LAUNCHER -nw"
 
     # same than M-x eval but from outside Emacs.
     alias eeval="$EMACS_PLUGIN_LAUNCHER --eval"
     # create a new X frame
     alias eframe='emacsclient --alternate-editor "" --create-frame'
 
-    # to code all night long
```
hjpotter92: Also, why remove these aliases? I understand that zsh would autocorrect the mistakes :stuck_out_tongue:

emanresusername: if you're trying to encourage usage of [thefuck](https://github.com/nvbn/thefuck) instead
i approve

emanresusername:
worked fine for me all day today 👍

emanresusername:
@mcornella @hjpotter92  anything gating merging in this one?

mcornella:
I was waiting for some input from @soredake, but your points don't seem to be that big of a deal. 

1. The deleted aliases thing should be solved with `setopt correct` which automatically guesses which command you meant if you mispell it.

2. The `te` alias runs emacs ignoring the window system, so it runs entirely on the terminal: [source](https://www.gnu.org/software/emacs/manual/html_node/emacs/Initial-Options.html#index-_002dnw).

mcornella:
So I am merging this because of your approval. Thanks! :+1:

PR #5202, opened 2016-07-02
Fix nvm plugin to use also check brew installs.

blainesch:
I installed nvm from brew, so this has the plugin check if that's the case

### DIFF
--- a/plugins/nvm/_nvm
+++ b/plugins/nvm/_nvm
@@ -1,7 +1,9 @@
 #compdef nvm
 #autoload
 
-[[ -s ~/.nvm/nvm.sh ]] || return 0
+if ![ -f $HOME/.nvm/nvm.sh ] && ![ $commands[brew] -a -f `brew --prefix`/opt/nvm/nvm.sh ]; then
+  return 0
+fi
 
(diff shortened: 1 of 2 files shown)


plugins/nvm/_nvm
```diff
@@ -1,7 +1,9 @@
 #compdef nvm
 #autoload
 
-[[ -s ~/.nvm/nvm.sh ]] || return 0
+if ![ -f $HOME/.nvm/nvm.sh ] && ![ $commands[brew] -a -f `brew --prefix`/etc/profile.d/autojump.sh ]; then
```
apjanke: `autojump.sh`? Should this be testing for `nvm.sh` instead?

OMZ generally prefers the double-bracket `[[ ... ]]` test syntax to the single-bracket `[ ... ]` syntax.

blainesch: Whoops. Fixed. I'll see if I can change to the double bracket syntax. As you can guess, I copied this from the [autojump](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/autojump/autojump.plugin.zsh) file.

PR #5205, opened 2016-07-04
Make Bash aliases available on zsh

shanavas786:
If there is a custom .bash_aliases file, load it also


mcornella:
This is too specific to users customization and if we add this we'll have to add checks for other kinds of customizations. Because of that, this should be left to the user to add to their zshrc files, respecting whatever kind of customization they've done.
Thanks for the contribution though.

PR #5218, opened 2016-07-10
Fix hg call when missing on archlinux with command-not-found

### LINKED ISSUE #5217: Invoking hg while it is not present in Agnoster theme
On Arch Linux with command-not-found plugin enabled and agnoster theme, and without mercurial installed, prompt looks weird
2016-07-10-193733_893x33_scrot

I believe we should check if hg is accessible, and if not, exit from prompt_hg function. It will fix this and also save time for each command for those who don't want hg to be installed.

atk91:
Fixes #5217: a bug when `you can install hg by typing ...` is shown like git branch. Also saves time for those who don't want hg to be installed and have command-not-found enabled as there is no need to search repo for hg each time prompt is being printed.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -127,9 +127,8 @@ prompt_git() {
   fi
 }
 
-prompt_hg() {
-  which hg > /dev/null 2>&1
-  if [[ $? -ne 0 ]]; then
+prompt_hg()
+  if ! type hg &> /dev/null ; then
     return
   fi
   local rev status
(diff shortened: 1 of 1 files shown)


atk91:
@stevenspasbo I agree. Updated pull request.

mcornella:
The idiomatic zsh way is using `$commands[hg]` in the following way:
```
if (( $+commands[zsh] )); then
...

In your case, this would be enough:
```zsh
(( $+commands[hg] )) || return
```

atk91:
@mcornella Okay, updated once again

PR #5219, opened 2016-07-11
better handling of loadable modules

rockdrilla:
prevent loading specific modules (oh-my-zsh/lib/*.zsh) via "config_exclude" variable instead of messing with .gitignore

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -27,12 +27,16 @@ fi
 
 
 # Load all of the config files in ~/oh-my-zsh that end in .zsh
-# TIP: Add files you don't want in git to .gitignore
-for config_file ($ZSH/lib/*.zsh); do
+# TIP: Add file names you don't want in array "config_exclude"
+for config_file (${ZSH}/lib/*.zsh); do
+  config_name=${config_file:r:t}
   custom_config_file="${ZSH_CUSTOM}/lib/${config_file:t}"
+  if [ -n "${config_exclude[(r)${config_name}]}" ]; then
+    continue
+  fi
   [ -f "${custom_config_file}" ] && config_file=${custom_config_file}
-  source $config_file
-done
+  source ${config_file}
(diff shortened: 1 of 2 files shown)


apjanke:
What's the motivation for this? OMZ expects all files from `lib` to always be loaded, and other OMZ code is written under that assumption.

rockdrilla:
I think it's better provide more elegant way to exclude some files than hard way to git-ignore them. Also, I've inspected OMZ/lib/ and not found any tight dependencies. May I ask you to point such cases?

apjanke:
> I think it's better provide more elegant way to exclude some files than hard way to git-ignore them.

But why do you want to do this in the first place?

> Also, I've inspected OMZ/lib/ and not found any tight dependencies. May I ask you to point such cases?

The dependencies are from plugins and themes, which may use stuff defined in `lib/` files. E.g. the `frisk` theme uses `bzr_prompt_info`, the `common-aliases` plugin uses `is-at-least`, which is autoloaded by `lib/misc.zsh` and `lib/diagnostics.zsh`, various plugins use `open_command` from `lib/functions.zsh`, and so on.

rockdrilla:
> But why do you want to do this in the first place?
> The dependencies are from plugins and themes, which may use stuff defined in lib/ files.
My motivation is to provide clean way to exclude modules from loading if end-user wants it to. Customization is the corner stone of OMZ, and why we force end-user to mess with git-ignore instead of modifying config parameter? End-user must have choice to enable/disable modules without carrying on OMZ code base integrity.
PS: as usual, i'm sorry about my poor english.

mcornella:
This is a feature that we won't implement anytime soon. If we did, the project would end up being a giant Frankenstein monster of things users want to ignore. Hopefully we can fix any issues you're having with lib files in a different way.

rockdrilla:
Closing PR. So long, and thanks for all the fish.

PR #5224, opened 2016-07-12
Add gcd for checking out default branch

craiglyons:
Some projects use alternate branches such as `development` for the main operating branch, and `master` is more protected (i.e. CI -> production, etc.).

Git repos have the option of a setting a `default` branch.  This alias is like a project-agnostic `gcm` and can help prevent accidental activity on the wrong branch.


jvz:
I kinda like this idea. I currently use `gcd` as an alias to checkout the `develop` branch as a parallel to `gcm`, but most repositories that use git flow style branch naming do indeed make `develop` or `development` the default branch, so this could be a better idea.

mcornella:
`gcd` is now `git checkout develop`.

PR #5225, opened 2016-07-12
require custom plugins recursively

audibleblink:
I had an issue where files deeper that one level in `$ZSH_CUSTOM` weren't being sourced.

As a workaround, I created my own file at the `$ZSH_CUSTOM` level that did the recursive sourcing. I made this change locally and benchmarked with `/usr/bin/time zsh -i -c exits` and found it was 100ms  faster.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -87,7 +87,7 @@ for plugin ($plugins); do
 done
 
 # Load all of your custom configurations from custom/
-for config_file ($ZSH_CUSTOM/*.zsh(N)); do
+for config_file ($ZSH_CUSTOM/**/*.zsh(N)); do
   source $config_file
 done
 unset config_file


apjanke:
```
for config_file ($ZSH_CUSTOM/**/*.zsh(N)); do
```

That is too broad. It will pick up files in `$ZSH_CUSTOM/plugins`, `$ZSH_CUSTOM/themes`, and other files which should only be sourced conditionally, or maybe not at all.

Restricting this to `$ZSH_CUSTOM/lib/**/*.zsh` might make sense. Though it's a significant change that could break things for existing users who might have something like a `$ZSH_CUSTOM/lib/disabled` directory, or their own custom solutions for conditionally sourcing files under a `lib/` hierarchy.

Probably better to stick with doing the recursive sourcing inside a file in your `$ZSH_CUSTOM`. I'm surprised that it's slower that way, though. How were you doing it from within the `$ZSH_CUSTOM` files?

audibleblink:
I had just one file, `sourcer.zsh` at the `$ZSH_CUSTOM` that had something that looked like:

```
for file ($(find . -name '*.zsh' | grep -v sourcer)); do
  source $file
done
```

I see now that my usage is not in accordance with the recommended, documented usage [here](https://github.com/robbyrussell/oh-my-zsh/blob/master/README.md#custom-plugins-and-themes)

So to solve my issue where plugin `$ZSH_CUSTOM/plugins/myplugin.zsh` wasn't getting loaded, I should just change `file.zsh` to `file.plugin.zsh` and the place `file` in my `plugins` array in my `.zshrc`?

apjanke:
Yes.

audibleblink:
For future searchers: 
I had two additional folders in `$ZSH_CUSTOM`, `aliases` and `functions`. I made a loader file again like I mentioned above with the following contents and it worked great.

```sh
fns="$HOME/.dotfiles/custom-omz/functions"
als="$HOME/.dotfiles/custom-omz/aliases"
source <(cat {"$fns","$als"}/**/*.zsh)
```

Issue #4758, opened 2016-01-07
gradle + gradlew autocompletion not working


kaushikgopal:
I'm almost positive this isn't something specific to my environment but the gradle + gradlew autocompletion never seems to works. Here's what i'm trying:

in my `.zshrc`:

```
 plugins=(brew git git-extras ruby rails gradle)
```

in my aliases.zsh:

```
alias gradlew="./gradlew"
alias gw="./gradlew"
alias gr="gradle"
```

on my console, as i type `gradlew <hit tab here>` nothing shows up. i've tried almost every combination with/without aliases etc.

I don't have this problem with git/brew/ruby or the other plugins shown above.

evilnoxx:
I have this problem too. Extracted to ~/bin/gradle-2.12, set GRADLE_HOME and added the bin dir to my path.
The gradle command seems to work just fine, but no autocomplete...

bes:
I have a similar issue, but instead of "nothing" I get this message (which prints out on the command line which is extra annoying)

`gradlew _gradle:80: command not found: gradlew`

Where the first gradlew is what I wrote and the rest is what is spewn out on my command line

bes:
I managed to partially solve it, for me at least.

In the directory `~/.oh-my-zsh/plugins/gradle` do the following:
```
ln -s gradle.plugin.zsh _gradle
ln -s gradle.plugin.zsh _gradlew
```

Now I can do 
```
./gradlew <TAB>
==== Gradlew Tasks ====
buildEnvironment   dependencies       help               model              properties         wrapper
components         dependencyInsight  init               projects           tasks
```

I would still like my *actual* tasks to show up, but it's a start...

bes:
I fixed so that I could list all tasks, and I created a pull request:

https://github.com/robbyrussell/oh-my-zsh/pull/5230

ryanstull:
I'm having a similar issue except I don't have gradlew, and I'm using sdkman to manage my gradle install.

$GRADLE_HOME = /home/ryan/.sdkman/candidates/gradle/current

Whenever I hit tab after typing gradle, I don't see any suggestions.

mcornella:
@ryanstull I assume you have the `gradle` plugin enabled. Run `functions -t _gradle_tasks` to debug the gradle completion function, trigger the gradle completion with `gradle [TAB]` then post the output that it spits.

MercurieVV:
[...]

PR #5230, opened 2016-07-14
Improve gradle plugin task parsing

bes:
Added _gradle and _gradlew as symbolic links to gradle.plugin.zsh,
otherwise the plugin was not properly loaded.

Output from `gradlew tasks --all` is now parsed in two levels,
first we find segments between `------...` and a newline.
Second, all those lines are parsed and cleaned using awk
and added to .gradletasknamecache.

Tested on gradle 2.13, and gradlew 2.14.

### DIFF
--- a/plugins/gradle/gradle.plugin.zsh
+++ b/plugins/gradle/gradle.plugin.zsh
@@ -66,7 +66,7 @@ _gradle_does_task_list_need_generating () {
 # are considered to be tasks. If and when gradle adds support for listing tasks
 # for programmatic parsing, this method can be deprecated.
 ##############
-_gradle_parse_tasks_to_gradletasknamecache () {
+_gradle_parse_tasks () {
   lines_might_be_tasks=false
   task_name_buffer=""
   while read -r line; do
@@ -78,7 +78,7 @@ _gradle_parse_tasks_to_gradletasknamecache () {
       if [[ "$lines_might_be_tasks" = true ]]; then
         # If a newline is found, send the buffer to .gradletasknamecache
         while read -r task; do
-          echo $task | awk '/[a-zA-Z0-9:-]+/ {print $1}' >> .gradletasknamecache
+          echo $task | awk '/[a-zA-Z0-9:-]+/ {print $1}'
         done <<< "$task_name_buffer"
         # Empty buffer, because we are done with the tasks
         task_name_buffer=""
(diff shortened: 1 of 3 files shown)


bes:
I think this issue is relevant for the pull request: https://github.com/robbyrussell/oh-my-zsh/issues/4758

froth:
Thanks a lot, works fine for me!

plugins/gradle/gradle.plugin.zsh
```diff
@@ -60,6 +60,35 @@ _gradle_does_task_list_need_generating () {
...
+  while read -r line; do
+    if [[ $line =~ ^-+$ ]]; then
+      lines_might_be_tasks=true
+      # Empty buffer, because it contains items that are not tasks
+      task_name_buffer=""
+    elif [[ $line =~ ^\s*$ ]]; then
+      if [[ "$lines_might_be_tasks" = true ]]; then
+        # If a newline is found, send the buffer to .gradletasknamecache
+        while read -r task; do
+          echo $task | awk '/[a-zA-Z0-9:-]+/ {print $1}' >> .gradletasknamecache
```
mcornella: I'm not 100% sure but this might cause multiple uncached writes to disk, and you need to delete the cache file beforehand. I'd rather make this a pure `echo` without adding it to the cache file just yet. Then you can replace the code on `_gradle_tasks` below for the following:

```diff
- gradle tasks --all | awk '/[a-zA-Z0-9:-]* - / {print $1}' > .gradletasknamecache
+ _gradle_parse_tasks $(gradle tasks --all)  > .gradletasknamecache
```

plugins/gradle/gradle.plugin.zsh
```diff
@@ -60,6 +60,35 @@ _gradle_does_task_list_need_generating () {
   [[ ! -f .gradletasknamecache ]] || [[ build.gradle -nt .gradletasknamecache ]]
 }
 
+##############
+# Parse the tasks from `gradle(w) tasks --all` into .gradletasknamecache
+# All lines in the output from gradle(w) that are between /^-+$/ and /^\s*$/
+# are considered to be tasks. If and when gradle adds support for listing tasks
+# for programmatic parsing, this method can be deprecated.
+##############
+_gradle_parse_tasks_to_gradletasknamecache () {
```
mcornella: I would also change this function name to `_gradle_parse_tasks`. See [my comment below](https://github.com/robbyrussell/oh-my-zsh/pull/5230/files#r78707652).

bes:
Thanks for the comments, new commit coming soon.

plugins/gradle/gradle.plugin.zsh
bes: Do you mean
```diff
-gradle tasks --all | awk '/[a-zA-Z0-9:-]* - / {print $1}' > .gradletasknamecache
+_gradle_parse_tasks "$(./gradlew tasks --all)"
```
Otherwise I didn't get it to work.

mcornella: Oh yeah quotes might be necessary. They're often a tricky thing within zsh.

mcornella:
That looks great, let me figure out a way to test this and I'll merge it right away.

mcornella:
@kaushikgopal @StanSvec @froth can you test this works?

froth:
@mcornella works fine for me with gradle and ./gradlew. Tasks are calculated correctly, cache is invalidated after changes in build.grade. I am happy :)

Definitely an improvement to the current master.

mcornella:
That's great to hear! Thanks @bes!

Issue #1, opened 2009-08-31
Git autocomplete doesn't work on filenames


robbyrussell:
It doesn't seem that filenames are auto-completing when using git. Might be an issue with the prompt or git functions.

Issue #3016, opened 2014-08-11
PAGER environment variable blindly overwritten


Hibou57:
In `$ZSH/lib/misc.zsh` at line 12, `export PAGER="less"`, blindly overwrites the `PAGER` environment variable. This should be changed into `env_default PAGER "less"`.

shreevatsa:
What is the resolution here? I saw #3359 and have that issue too -- I just installed OMZ and it just clobbered by SAVEHIST and HISTSIZE to the ridiculously low value of 10000. And it's also doing something strange with the pager, where it brings up the pager even for very short (or empty: e.g. "git diff" from a clean state) files.

shreevatsa:
Nonsensical. This has been closed without even a reply, when the issue is very much still present. The resolution to "your software does a broken thing for no good reason" is not "here's a way you can undo the damage after you realise it has already happened". I'm giving up.

mcornella:
Sorry @shreevatsa I wasn't clear enough. I was referring to this https://github.com/robbyrussell/oh-my-zsh/pull/3359#issuecomment-67022891, which says that these variables should be initialized *after* the `source oh-my-zsh...` line in `.zshrc`.

That way we avoid having bad defaults as it happened with the `SAVEHIST` and `HISTSIZE` variables on #3359. There could be a potentially better alternative, which is checking if those variables have really low variables (like zsh's default which puts it at 30), but the alternative of defining them after oh-my-zsh ensures we don't have to change it in the future.

And btw, sorry I didn't see your comment before. Do you have a better suggestion?

shreevatsa:
@mcornella Thanks for your gracious comment.

I don't have the patience to try oh-my-zsh again now after many sour experiences like this (lots of my settings were screwed up, and even tentative attempts to improve things, e.g. pull request #3869, were ignored), but perhaps for future users, you ought to write in bold and at the top of the installation instructions that it is recommended that `source oh-my-zsh...` be the first line in `.zshrc` and all your personal settings come later.

Most users advanced enough to use zsh and further use oh-my-zsh probably have their preferences too, and it's frustrating that the default behaviour of oh-my-zsh is to ignore all those preferences. I see in hindsight that a lot would be improved if I had put `source oh-my-zsh` at the top rather than at the bottom (but still not everything), maybe if the next time I try OMZ, if the documentation has been updated to say it prominently, I'll remember to do so.

Hibou57:
[...]

PR #5231, opened 2016-07-15
Do not override misc env set before sourcing oh-my-zsh

defrank:
Sourcing oh-my-zsh happens in zshrc, which will override settings of
profile and zshenv.  Treat misc values, `PAGER` and `LESS`, as default
settings without overriding existing values.

Fixes: #1, robbyrussell/oh-my-zsh#3016

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -19,8 +19,8 @@ fi
 setopt long_list_jobs
 
 ## pager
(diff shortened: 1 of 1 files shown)


dasJ:
:+1*

mcornella:
Thanks!

dasJ:
Thank you!

PR #5233, opened 2016-07-15
Removes .phing_target generation from phing plugin

donatj:
The phing plugin currently drops a `.phing_targets` cache file wherever you run it.  This isn't necessary, `phing -l` takes .03 seconds (on my system) so I can stand to run it each time.

I don't know of any other plugins that cache wherever you are like this, but I'm not a fan of things making messes on my filesystem.

Closes #3415.

### DIFF
--- a/plugins/phing/phing.plugin.zsh
+++ b/plugins/phing/phing.plugin.zsh
@@ -1,15 +1,6 @@
-_phing_does_target_list_need_generating () {
-  [ ! -f .phing_targets ] && return 0;
-  [ build.xml -nt .phing_targets ] && return 0;
-  return 1;
-}
-
 _phing () {
   if [ -f build.xml ]; then
-    if _phing_does_target_list_need_generating; then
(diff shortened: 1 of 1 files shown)


donatj:
I note that this is **very similar** to #3415. That appears to have

1. Stagnated
1. Contains a theme?

I'd be perfectly fine with either, I'd just like a solution, because it pooping out cache files like this is gross.

@mcornella noted on that:

> If you haven't noticed delays I'd get rid of it but I'm concerned that someone might have a bigger build.xml and does notice the performance gain.

Our build.xml is just under a thousand lines and it's nearly instant to list.  I mean all phing needs to do is parse the xml files for `<target>` names, even with a huge XML file that should be fairly reasonable.

mcornella:
Thanks for checking other open PRs. I'm willing to drop the cache generation then.

That other PR also adds `grep -v "Warning:"` to grep out Warnings from the output, do you think that's necessary?

donatj:
@mcornella I haven't personally ever seen the warnings in question, but did find them peaking around Phing's source code a little. 

I'm fine with adding the grep from the previous PR and I've gone ahead and cherry-picked the `grep -v "Warning:"` logic from #3415 into this to give @hrwath credit where credit's due.

mcornella:
Great job. Merging right away!

PR #5240, opened 2016-07-19
add MDN to web-search plugin

sappharx:


mcornella:
This is found in `frontend-search` search plugin. We made a conscious decision to separate both these search engines because, as you can see, there are a lot more in the other plugin. Have a look at the README: https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/frontend-search/README.md.

PR #5242, opened 2016-07-21
Add dynamic task completion for mix

jasonm23:
Improved completion for Mix, by dynamically building the task list from `mix --help`

Lists new mix tasks as they are installed via elixir dependencies


bitboxer:
@jasonm23 ping. Can you please rebase? I want this :wink:

PR #5247, opened 2016-07-22
Updated git_commits_ahead function

stevenspasbo:
* Function wasn't returning correct value
* Updated underlying git command
* Kept echo command from previous function for backwards compatibility
* Kept function consistent with git_commits_behind


mcornella:
Thanks! :+1:

PR #5251, opened 2016-07-24
Async npm

slavaGanzin:
```
time npm completion > /dev/null                                                                                                                                                                      (develop|✔)
npm completion > /dev/null  0,26s user 0,03s system 105% cpu 0,276 total
```

As you can see npm completion takes quarter of a second to run. And so it introduces quarter of second delay in every zsh initialization. My changes launch npm completion synchronously(blocking) only first time. All subsequent launches will not block zsh initialization.

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -2,11 +2,17 @@
 {
 
   COMPLETION="${ZSH_CACHE_DIR}/npm_completion"
-  [[ ! -f $COMPLETION ]] && npm completion 2>/dev/null >! $COMPLETION
-  source $COMPLETION
-  npm completion 2>/dev/null >! $COMPLETION &
-  
+
+  if [[ ! -f $COMPLETION ]]; then
+    npm completion 2>/dev/null >! $COMPLETION
+    source $COMPLETION
+  else
+    source $COMPLETION
+    npm completion 2>/dev/null >! $COMPLETION &
+  fi
+
   unset COMPLETION
+
   # Install dependencies globally
   alias npmg="npm i -g "
 
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,3 +1,4 @@
+<<<<<<< 38da9b9bb479d3abae6558ba738ec63403ffa2b0
 (( $+commands[npm] )) && {
     __NPM_COMPLETION_FILE="${ZSH_CACHE_DIR}/npm_completion"
 
@@ -10,6 +11,14 @@
 
     unset __NPM_COMPLETION_FILE
 }
+=======
+completion=~/.npm_completion
+if [ ! -f $completion ]; then
+  npm completion 2>/dev/null > $completion
+fi
+source $completion
+(rm $completion; npm completion 2>/dev/null > $completion &)
+>>>>>>> prevent error on first launch
 
 # Install dependencies globally
 alias npmg="npm i -g "
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -43,4 +43,3 @@ alias npmst="npm start"
 
 # Run npm test
 alias npmt="npm test"
-
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,25 +1,10 @@
-<<<<<<< 38da9b9bb479d3abae6558ba738ec63403ffa2b0
-(( $+commands[npm] )) && {
-    __NPM_COMPLETION_FILE="${ZSH_CACHE_DIR}/npm_completion"
-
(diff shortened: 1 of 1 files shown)


willmendesneto:
Can you elaborate about this change, please? What's the benefit in create and remove the `. npm_completion` file?

slavaGanzin:
@willmendesneto 

```
time npm completion > /dev/null                                                                                                                                                                      (develop|✔)
npm completion > /dev/null  0,26s user 0,03s system 105% cpu 0,276 total
```

As you can see npm completion takes quarter of a second to run. And so it introduces quarter of second delay in every zsh initialization. My changes launch npm completion synchronously(blocking) only first time. All subsequent launches will not block zsh initialization.



> What's the benefit in create and remove the . npm_completion file?

I use ```unsetopt clobber```, so I can't just ```npm completion > /.filename```

willmendesneto:
> As you can see npm completion takes quarter of a second to run. And so it introduces quarter of second delay in every zsh initialization

Cool! Can you add this information in your PR description, please? 

In that case, you can ignore your local clause using `>!`. This clause works well via command line without your local configuration too, as I tested locally.

```bash
time npm completion > ~/.npm_completion
npm completion > ~/.npm_completion  0.29s user 0.04s system 102% cpu 0.321 total
➜  ~ time npm completion >! ~/.npm_completion
npm completion >| ~/.npm_completion  0.29s user 0.04s system 102% cpu 0.325 total
➜  ~ unsetopt clobber && time npm completion >! ~/.npm_completion
npm completion >| ~/.npm_completion  0.30s user 0.04s system 102% cpu 0.339 total
```

More details on [ZSH Redirection documentation](http://zsh.sourceforge.net/Doc/Release/Redirection.html)

slavaGanzin:
@willmendesneto Thanks for ```>!```

willmendesneto:
You welcome! I tested and looks good. Could you rebase with master to test it again, please?

slavaGanzin:
@willmendesneto I rewrite a code a bit to simplify a bit

plugins/npm/npm.plugin.zsh
```diff
@@ -1,16 +1,10 @@
...
-    fi
-
-    [[ -f $__NPM_COMPLETION_FILE ]] && source $__NPM_COMPLETION_FILE
-
-    unset __NPM_COMPLETION_FILE
-}
-
+(( $+commands[npm] )) &&
+(
+  COMPLETION="${ZSH_CACHE_DIR}npm_completion"
```
mcornella: This breaks after changing `$ZSH_CACHE_DIR` to `$ZSH/cache`. Use the trailing slash.
Also, you should `unset` the variable at the end of the code.

plugins/npm/npm.plugin.zsh
```diff
@@ -1,16 +1,10 @@
...
+  [[ ! -f $COMPLETION ]] && npm completion 2>/dev/null >! $COMPLETION
+  source $COMPLETION
+  npm completion 2>/dev/null >! $COMPLETION &
```
mcornella: This will mean calling `npm completion` twice if there wasn't a `$COMPLETION` file.

plugins/npm/npm.plugin.zsh
```diff
@@ -1,16 +1,10 @@
...
+)
```
mcornella: You're using `()` which trigger a subshell. Use `{}` like it was before for no additional processes.

slavaGanzin: I was using subshell, because it eliminates need of ```unset COMPLETION```. 

Why use of subshell can be harmfull?

plugins/npm/npm.plugin.zsh
slavaGanzin: It will cache npm completion on first run. And will show you wrong completion if you will switch to older/newer version of npm (for example with [n](https://github.com/tj/n))

slavaGanzin:
https://github.com/robbyrussell/oh-my-zsh/pull/5251/commits/96f7c0762f4dd285e5f2f224659ccb3b21275185#diff-db3a476da51cf1961f9297776153ac10R6

I think this is "optimization": 
1. Harder to understand what's happening
2. Harder to read a code
3. It will save much less CPU time of all users around the world, then time we spent discussing it

plugins/npm/npm.plugin.zsh
willmendesneto: I think that in this case `COMPLETION` should use `COMPLETION="${ZSH_CACHE_DIR}/npm_completion"` because you're trying to add in the default folder.

willmendesneto:
@slavaGanzin @mcornella Any updates about this PR?

slavaGanzin:
Outdated. Already implemented in master

willmendesneto:
Sorry about that @slavaGanzin =/

slavaGanzin:
@willmendesneto It works better than I want it to. So I'm more than satisfied

PR #5253, opened 2016-07-25
Adds option for directory to Xcode xc function.

ashfurrow:
I've found myself doing this:

```sh
> xc
No xcworkspace/xcodeproj file found in the current directory.
> cd Project ; xc ; cd ..
```

To open a project in a subdirectory from the command line. And sometimes I forget to `cd ..` 😬 

So I thought I would add the option to the plugin. It's my first contribution, so let me know if there's anything I need to change in the PR.

### DIFF
--- a/plugins/xcode/README.md
+++ b/plugins/xcode/README.md
@@ -26,7 +26,7 @@ plugins=(... xcode)
 
 ###  `xc`
 
-Opens the current directory in Xcode as an Xcode project. This will open one of the `.xcworkspace` and `.xcodeproj` files that it can find in the current working directory.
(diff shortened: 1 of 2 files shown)


mauricerkelly:
@robbyrussell I gave this a test and it does the job perfectly for me!

@ashfurrow I often have exactly the same requirement - I like to sit at the root of our repo which consists of subdirectories that have workspace/project files so this is a great addition. Thanks also for drawing my attention to the plugin itself - I've long had my own shell function to do the same, and it'll be nice to retire it in favour of this one! 👍

ashfurrow:
Sounds like this has been tested, anything else I can do to help get it merged?

mcornella:
Thanks, that's perfect. Sorry for the delay!

PR #5255, opened 2016-07-26
README.md: Manual Installation: Clone the repository same as the tools/install.sh script.

SushiDude:


blueForestIcarus:
Are you alive sushidude?

mcornella:
That's up to the user to run it like that.

Issue #3766, opened 2015-04-10
Implement a lockfile to alleviate update prompt in multiple terminals


chmac:
I have iTerm2 set to auto open about a dozen terminals. When there's an update due, it asks me in every terminal if I want to upgrade. Would be awesome if it used some kind of mutex, maybe a file lock or something, to ensure that the prompt only appears once.

#feature-request

Thanks for an awesome library.

ncanceill:
Hi, and thanks for contributing.

What you are suggesting seems like a big change, so I would not count on it being added soon. Maybe you can start working on it yourself?

mcornella:
Here's a simple algorithm that will do what you want:

1. Touch `$ZSH/log/update.lock`
2. Update
3. Delete `$ZSH/log/update.lock`

It will be a little more complicated than that but it's a start. What do you think?

chmac:
The challenge with touch / update / delete is that you need to check if the file exists, then touch, and that's not an atomic operation. So two processes started at exactly the same time, might both decide that there is no lock file, and then both call touch at the same time, thereby creating a race condition.

However, google was helpful, tools exist to solve precisely this problem. [solo](http://timkay.com/solo/) could be useful. Apparently only 10 lines of code.

@ncanceill I'd guess it's not such a big change. I'm assuming there's some kind of occasional check to see if the update should be fired. It could be a case of just adding some kind of mutex call to that update.

Hmm, this whole thing seems much more accessible than I imagined. I guess a change in [check_for_upgrade.sh](../blob/master/tools/check_for_upgrade.sh) will probably do it. I might look into that myself and see about a pull request...

ncanceill:
@chmac I will be happy to test if you do so.

apjanke:
There's also basic `flock` support in the [zsh/system module](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fsystem-Module) that would allow you to do actual file locking. Something like this.

```
zmodload zsh/system
lockfile=$TMPDIR/omz-upgrade.lock
touch $lockfile
zsystem flock -t 1 $lockfile
```

robbyrussell:
Would someone be interested in taking a stab at a PR for this?

PR #5257, opened 2016-07-26
Update .zsh-update immediately

melvinsh:
Update the `LAST_EPOCH` in `~/.zsh-update` immediately before starting the upgrade. This prevents the "Would you like to check for updates?" prompt from showing up after starting the upgrade process and opening a new shell.

I'd appreciate it if this became the new behaviour. It's been bothering me for a while now 😄 

Current situation

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -11,9 +11,9 @@ function _update_zsh_update() {
 }
 
 function _upgrade_zsh() {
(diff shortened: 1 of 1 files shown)


robbyrussell:
Hi there -- thanks for the suggestion. This would be a short-term hack approach but we'd like to implement a proper lockfile for. Perhaps you'd be interested in helping us squash #3766?

melvinsh:
@robbyrussell I can give it a shot.

PR #5258, opened 2016-07-26
Create an alias to open up sublime project

hanpeter:
`stp` commands works in following way:

1. If the current directory is a part of a Git repo and if there is a `.sublime-project` file among its ancestor directories, the command opens up that sublime project.
1. If the current directory is a part of a Git repo but there isn't a `.sublime-project` file among its ancestor directories, the command opens up the root of the repo.
1. If the current directory is not a part of a Git repo, the command opens up the current directory.


robbyrussell:

Cheers,
Robby

PR #5260, opened 2016-07-27
Add stackoverflow search into web-search plugin

hiepph:


mcornella:
This is already present on the [`frontend-search` plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/frontend-search).

PR #5261, opened 2016-07-27
ssh-agent: Use /usr/bin/env to find ssh-add

mguentner:
This change is important when ssh-add is not inside /usr/bin e.g.
on NixOS.

Signed-off-by: Maximilian Güntner <code@maschinenpsychologe.de>

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -45,7 +45,7 @@ function _plugin__start_agent()
   zstyle -a :omz:plugins:ssh-agent identities identities
   echo starting ssh-agent...
 
-  /usr/bin/ssh-add $HOME/.ssh/${^identities}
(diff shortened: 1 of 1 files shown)


jannotti:
Why not just run ssh-add without the absolute path OR the use of env?  Doesn't this use env in order to search the PATH, when that's what normal command execution does anyway?

mguentner:
This should be identical. But since in line 40 [1] env is already used I thought it might be wise to keep the style.

[1] https://github.com/mguentner/oh-my-zsh/blob/395afa7276d23cb7d68829b09e8dbf5c3c3b105b/plugins/ssh-agent/ssh-agent.plugin.zsh#L40

apjanke:
I'd just remove the `/usr/bin/env` call in both places; it's not doing anything.

mcornella:
I've removed the `/usr/bin/env` and done some other work on the ssh-agent plugin in #5328.

PR #5262, opened 2016-07-28
Add shrink-path plugin to shorten directory paths (e.g. ~/f/b/quux)

shvenkat:
Shrink directory paths, e.g. `/home/me/foo/bar/quux` to `~/f/b/quux`.

For a fish-style working directory in your command prompt, add the following to your theme or zshrc:
```zsh
setopt prompt_subst
PS1='%n@%m $(shrink_path -f)>'
```

The following options are available:
```
-f, --fish       fish simulation, equivalent to -l -s -t.
-l, --last       Print the full name of the last directory.
-s, --short      Truncate directory names to the first character. Without
                 -s, names are truncated without making them ambiguous.
-t, --tilde      Substitute ~ for the home directory.
-T, --nameddirs  Substitute named directories as well.
```

The long options can also be set via zstyle, like:
```
zstyle :prompt:shrink_path fish yes
```

Note: Directory names containing two or more consecutive spaces are not yet supported.

Based on code by Daniel Friesel from http://www.zsh.org/mla/workers/2009/msg00415.html.

### DIFF
--- a/plugins/shrink-path/README.md
+++ b/plugins/shrink-path/README.md
@@ -1,0 +1,68 @@
+# A plugin to shrink directory paths for brevity and pretty-printing
+
+
+## Examples
+
+For this directory tree:
+```
+    /home/
+      me/
+        foo/
+          bar/
+            quux/
+          biz/     # The prefix b is ambiguous between bar and biz.
+```
+here are the results of calling `shrink_path <option> /home/me/foo/bar/quux`:
+```
+    Option        Result
+    <none>        /h/m/f/ba/q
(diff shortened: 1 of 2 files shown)


mcornella:
This looks very cool. Can you add a README file with instructions on how to use it? What you have in this PR description seems enough.

shvenkat:
@mcornella I added a README.

mcornella:
:tada:

huyle93:
For some reason when I run this `setopt prompt_subst
PS1='%n@%m $(shrink_path -f)>'` in my .zshrc file, nothing change in my directory.

PR #5264, opened 2016-07-29
Fix up search process for aws_zsh_completer.sh in the non-brew case.

jannotti:
When "which" returns nothing, lack of quotes around variable was
causing a syntax error.  Quotes are also needed in
the (admittedly unusual) case of a filename with spaces in it.
The test should be for existence, not executability, since the
file is sourced, and therefore need not be marked executable.

Which also means that the "which" command isn't really a great
way to find the file.  It might not be in a place that's on
normal PATHs.  So this commit adds a check for one common place
to look: the location of the completer file on Amazon Linux - seems
reasonable to make sure the aws command has nice completion on
aws machines!


mcornella:
Obsolete and "fixed" in #7364

PR #5265, opened 2016-07-29
nvm plugin improvements

### LINKED ISSUE #5270: nvm functionality doesn't work if nvm is installed in a custom directory
nvm supports being installed in a custom directory, by default it's installed in ~/.nvm but it can be specified anywhere by using the environment variable `$NVM_DIR`. Even if nvm is installed in `~/.nvm` the environment variable should still be set, otherwise nvm has to work out where it's installed each time you run it.

Currently all omz functionality regarding zsh (loader/completion/theme helper) hardcode the path as ~/.nvm and `$NVM_DIR` is never set.

lukechilds:
`nvm` isn't always installed in ~/.nvm. These changes allows nvm to be installed in any directory. It also now properly exports the `$NVM_DIR` env var if it hasn't been set and checks it in the completion and `nvm_prompt_info` rather than assuming `nvm` is in ~/.nvm.

Fixes #5270

### DIFF
--- a/lib/nvm.zsh
+++ b/lib/nvm.zsh
@@ -1,6 +1,6 @@
 # get the node.js version
 function nvm_prompt_info() {
-  [ -f "$HOME/.nvm/nvm.sh" ] || return
+  [[ -f "$NVM_DIR/nvm.sh" ]] || return
   local nvm_prompt
   nvm_prompt=$(node -v 2>/dev/null)
   [[ "${nvm_prompt}x" == "x" ]] && return
--- a/plugins/nvm/_nvm
+++ b/plugins/nvm/_nvm
@@ -1,7 +1,7 @@
 #compdef nvm
 #autoload
 
-[[ -s ~/.nvm/nvm.sh ]] || return 0
+[[ -f "$NVM_DIR/nvm.sh" ]] || return 0
 
 local -a _1st_arguments
 _1st_arguments=(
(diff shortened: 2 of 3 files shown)


mcornella:
:+1: but I would put the `$NVM_DIR` definition on `lib/nvm.zsh`, to have it as soon as possible.

lukechilds:
@mcornella Ah, I assumed `plugins/nvm/nvm.plugin.zsh` was loaded before `lib/nvm.zsh` (not sure why looking back 😃 🔫 ).

I'm not too familiar with omz, is there ever a scenario where `plugins/nvm/nvm.plugin.zsh`would be sourced but not `lib/nvm.zsh`? I'm  thinking it may be best to put the `$NVM_DIR` definition in both to make sure it's definitely always set.

lukechilds:
Wait, isn't `lib/nvm.zsh` setting up `nvm_prompt_info` as a helper function for a theme? So the theme helper function will always be called after the `nvm` plugin has been loaded so `$NVM_DIR` will exist?

lukechilds:
@mcornella Is this ok as is?

mcornella:
I wasn't sure what would happen if `$NVM_DIR` wasn't set, but after revising it further I see no possible problem with `nvm_prompt_info`. The exporting of `NVM_DIR` is [already done on `nvm.sh`](https://github.com/creationix/nvm/blob/364096a/nvm.sh#L113-L114) but it doesn't hurt anything doing it here.

TL;DR this is already ok.

hax:
Should lib/nvm.zsh move to /plugins/nvm ?

PR #5276, opened 2016-08-02
update alias of get a list of packages installed locally

daumie:
Use of alias  `allpkgs` =`aptitude search -F "%p" --disable-columns ~i`
 is deprecated and exits with an error <br/>
The script new script `dpkg --get-selections | grep -v deinstall` works perfectly ok


mcornella:
Thanks!

PR #5280, opened 2016-08-03
Add "system" to the completion list for plugin "chruby"

franklinyu:
Detect system Ruby in default `PATH`, and provide "system" completion if Ruby is found. Tested with [the Dockerfile](https://gist.github.com/franklinyu/b8deda6a5093a17b575679b5808b371f).

### DIFF
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -95,5 +95,11 @@ function chruby_prompt_info() {
 }
 
 # complete on installed rubies
-_chruby() { compadd $(chruby | tr -d '* ') }
+_chruby() {
+    compadd $(chruby | tr -d '* ')
+    local default_path='/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin'
+    if PATH=${default_path} type ruby &> /dev/null; then
+        compadd system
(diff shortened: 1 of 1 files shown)


franklinyu:
@webframp Any suggestion?

webframp:
Seems reasonable to me, I didn't have this particular need but it looks good.

franklinyu:
So no one want to test it out by now?

plugins/chruby/chruby.plugin.zsh
```diff
@@ -95,5 +95,11 @@ function chruby_prompt_info() {
 }
 
 # complete on installed rubies
-_chruby() { compadd $(chruby | tr -d '* ') }
+_chruby() {
+    compadd $(chruby | tr -d '* ')
+    local default_path='/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin'
+    if ( PATH=${default_path} type ruby > /dev/null); then
```
mcornella: This should be `&>/dev/null` to redirect also stderr. You can also get rid of the parentheses.

franklinyu:
@mcornella Any more suggestions?

mcornella:
That's great, thanks for your contribution!

PR #5282, opened 2016-08-04
Make tmux plugin refresh tmux global environments automatically.

delphij:
This is handy for tmux users who work in interactive mode.  By default, tmux would pass through e.g. X11, SSH agent, etc. via global environment.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -15,6 +15,8 @@ fi
 : ${ZSH_TMUX_AUTOQUIT:=$ZSH_TMUX_AUTOSTART}
 # Automatically name the new session based on the basename of PWD
 : ${ZSH_TMUX_AUTONAME_SESSION:=false}
+# Automatically pick up tmux environments
+: ${ZSH_TMUX_AUTOREFRESH:=true}
 # Set term to screen or screen-256color based on current terminal support
 : ${ZSH_TMUX_DETACHED:=false}
 # Set detached mode
@@ -158,6 +160,15 @@ function _zsh_tmux_plugin_run() {
   fi
 }
 
+# Refresh tmux environment variables.
+function _zsh_tmux_plugin_preexec()
+{
+  local -a tmux_cmd
+  tmux_cmd=(command tmux)
+
+  eval $($tmux_cmd show-environment -s)
+}
+
 # Use the completions for tmux for our function
 compdef _tmux _zsh_tmux_plugin_run
 # Alias tmux to our wrapper function.
@@ -183,4 +194,10 @@ if [[ -z "$TMUX" && "$ZSH_TMUX_AUTOSTART" == "true" && -z "$INSIDE_EMACS" && -z 
     export ZSH_TMUX_AUTOSTARTED=true
     _zsh_tmux_plugin_run
   fi
+fi
+
+# Automatically refresh tmux environments if tmux is running.
+if [[ -n "$TMUX" && "$ZSH_TMUX_AUTOREFRESH" == "true" ]] && tmux ls >/dev/null 2>/dev/null; then
+  autoload -U add-zsh-hook
+  add-zsh-hook preexec _zsh_tmux_plugin_preexec
 fi


delphij:
Hi, @robbyrussell are you looking for someone to do the testing, or would like some instructions?

To test:

- Enable the 'tmux' plugin in ~/.zshrc

ssh -A to the system, then run 'tmux'.

env | grep SSH_AUTH_SOCK

Detach the session (Ctrl+B, D), logout, then repeat ssh -A, tmux

Hit enter (which makes sure that zsh have a new prompt)

env | grep SSH_AUTH_SOCK

The result should be different.  ssh'ing to other hosts would be relayed to user's workstation for SSH agent authentication.

xals:
Hello @delphij,
You may use the preexec hook instead of precmd so you would not have to press enter just before running a command. Anyway, I don't know how it behaves on variable substitution, to be tested, I think.

delphij:
Good point!  I have updated my tree and it worked.

khsing:
I have tested this feature, truly helped me while I am connected through`tmux`, very good one. 

ps. second `tmux` is attach to the pervious session, not new one. 

```
➜  ~ env | grep SSH_AUTH # first ssh connection and tmux new session 
SSH_AUTH_SOCK=/tmp/ssh-dxCaf0fmj8/agent.22865
➜  ~ env | grep SSH_AUTH  # second ssh connection and tmux attach back
SSH_AUTH_SOCK=/tmp/ssh-xtau3V7aPt/agent.23157
➜  ~ 
```
Thank you very much.

MikeDacre:
Personally, I think this PR should be merged, the functionality is super helpful and can be easily toggled on/off by the user.

@xals I think you were asking about how changing variables that are also set in the `tmux update-environment` command, it is true that with this plugin those commands would be automatically overridden, and any variables set from them (e.g. `i=$DISPLAY`) will not be updated by the preexec hook.

@delphij I would suggest also adding `tmux ls >/dev/null 2>/dev/null` to your test on line 106, because otherwise if someone SSHs into another machine from within tmux, and that machine has this plugin installed, it will run anyway as both variables will be set, so then the user will get tmux errors on every prompt/command. I would suggest changing the test to:

```shell
[[ -n "$TMUX"  && "$ZSH_TMUX_AUTOREFRESH" == "true" ]] && tmux ls >/dev/null 2>/dev/null
```

plugins/tmux/tmux.plugin.zsh
```diff
@@ -93,6 +101,13 @@ if which tmux &> /dev/null
 			_zsh_tmux_plugin_run
 		fi
 	fi
+
+	# Automatically refresh tmux environments if tmux is running.
+	if [[ -n "$TMUX" && "$ZSH_TMUX_AUTOREFRESH" == "true" ]]
```
MikeDacre: I would suggest changing this to:

```shell
if [[ -n "$TMUX"  && "$ZSH_TMUX_AUTOREFRESH" == "true" ]] && tmux ls >/dev/null 2>/dev/null; then
```

This prevents the plugin from running in an ssh shell that is a child of tmux, and also complies with the oh-my-zsh coding style guidelines.

delphij: Done.

delphij:
I've rebased my changes over the latest version of the plugin.

carlosala (approved):
After, 8.5 years... LGTM! 🚀 
It seems like this PR got traction in the past, let's go for it!
Thanks for your contribution!

KRACK-BIT:
TL;DR - this feature silently breaks any setup if `tmux` is opened before `omz` is run, and silently breaks setups which attempts to initialize `ssh` + `xserver` + etc. variables after `omz` is run.

Thus, please (preferably) set this feature as disabled by default, or (otherwise) include warning prompts for the former issue, and a disclaimer in the plugin README about the latter issue.

---

After seeing how this broke several of my setups, I would like to propose this feature is set to opt-in, and not opt-out.

Whilst this feature can definitely be useful, it must be noted that `tmux show-environment -s` does not merely append or even overwrite from set variables in the global environment - in addition, it will *silently unset several key variables if they were not initialized in the global environment* i.e.

```sh
➜  / tmux show-environment -s
unset DISPLAY;
unset KRB5CCNAME;
unset SSH_AGENT_PID;
unset SSH_ASKPASS;
unset SSH_AUTH_SOCK;
unset SSH_CONNECTION;
unset WINDOWID;
unset XAUTHORITY;
```

Thus, since this feature runs `show-environment -s` after every command, not only will any later changes to variables in the global environment be silently undone, but if these particular variables were yet to be initialized before the `tmux` server was launched, *they now silently cannot be initialized by user configuration!*

---

To demonstrate, consider this dockerfile environment:

```dockerfile
## OMZ setup
FROM archlinux
RUN pacman -Syu --noconfirm zsh tmux openssh git
RUN chsh -s /usr/bin/zsh
RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"

## Demo setup
RUN sed -i -e 's/plugins=(git)/plugins=(git tmux ssh-agent)/' ~/.zshrc

## Demo env
CMD [ "/usr/bin/tmux", "new-session", "-s", "proof_of_concept", "zsh" ]
```

Here, you will find that despite no warnings being thrown at start-up, not only are `ssh-agent`'s variables getting silently dropped, but this issue cannot be patched post-hoc by the user, and w/o already knowing about this feature, drawing the connection to an omz preexec hook, let alone the `tmux` plugin is easy to miss!

```sh
Starting ssh-agent ...
➜  / echo "socket should be set... - socket=${SSH_AUTH_SOCK:-💀}"
socket should be set... - socket=💀
➜  / echo "let me try to fix this..." && SSH_AUTH_SOCK="value" && echo "socket=${SSH_AUTH_SOCK:-💀}"
let me try to fix this...
socket=value
➜  / echo "socket should be fixed now... - socket=${SSH_AUTH_SOCK:-💀}"
socket should be fixed now... - socket=💀
```

Thus, leaving this as a feature you must knowingly opt into seems sensible, as to avoid other users unaware of this feature from facing unexpected issues adopting this plugin.

carlosala:
Sounds good, already on master. Thanks for the idea!

PR #5283, opened 2016-08-04
Fix newline issue on Ys theme

Kirogo:
Hello !
I've started using Ys theme very recently, and I've noticed one small issue I'm proposing to fix here : the first line break of the prompt was not escaped. This created a newline before each prompt, which is not very elegant (especially when a new terminal is created).

Cheers,
Alexandre.


mcornella:
This is a feature of the [`ys` theme](https://github.com/robbyrussell/oh-my-zsh/wiki/Themes#ys), look at the screenshot.

You can use your modified theme as a custom theme though. See [customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes) for instructions.

PR #5287, opened 2016-08-07
(git): Add gpsup alias

SimenB:
Add alias to push new branch to origin, and set up local branch to track it.

I can update the Wiki if this change is accepted.

/cc @ncanceill as you're set up as maintainer.


mcornella:
:+1: I added that to the wiki already.

PR #5298, opened 2016-08-12
add kubectl plugin

pstadler:


robbyrussell:
Hi there, it looks like the kubectl provides the completion, itself, right? I'm not sure if we should pull that single line could be added to someone's .zshrc file. I'd need to see some +1s for this before accepting it because it feels _too_ thin to bundle into OMZ

pstadler:
I totally agree with you, but it would still be convenient for people not knowing about that feature.

michaelfavia:
FWIW I found this useful for the purposes of managing plugins and autocompletions in a uniform fashion. Thx for committing it.

PR #5299, opened 2016-08-12
Enhancement: add angular-cli (ng) completion for arguments and options

catull:
The angular-cli utility has a rough completion generator.

This plugin supports more arguments and options.
As soon as the angular-cli utility supports all options, this plugin will turn into a much simpler script, basically executing

```sh
eval `ng completion`
```


robbyrussell:
Thanks, can we also get a README file?

catull:
You asked for a README.md file, and you shall obtain it.

There aren't any aliases.

> He that badly requires aliases, let him first create them.

Skipping memorisation of cryptic short-cuts avoids pollution of users' precious brain cells...

Besides, ng already supports these short-cuts:
- `b` for `build`
- `d` for `destroy`
- `g` for `generate`
- `h` for `help`
- `i` for `install`
- `s` for `serve`
- `t` for `test`
- `v` for `version`

mcornella:
This has become polluted with update commits.

catull:
Thanks for merging it, really appreciated.

mcornella:
:+1:

mcornella:
Added info to the wiki as well: https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins#ng. Cheers!

catull:
My plan is to eventually reduce the script down to eval `ng completion`, we're not too distant from it.

mcornella:
Is this supposed to be a [mirror of the official completion](https://github.com/angular/angular-cli/blob/master/packages/angular-cli/utilities/completion.sh)? If so, we could use an algorithm similar to the one in the [npm plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/npm/npm.plugin.zsh).

catull:
The mirror as you aptly name it, still has some issues.

It does not properly work in ZSH if you have OPAM (OCaml package manager) installed.
In that case, `ng completion` produces BASH completion output.

The reason for this is that OPAM defines an shell function complete, which shadows the shell builtin function.
Unfortnately, `ng completion` checks if `complete` is defined and assumes a BASH.

In the past, `ng completion` received little attention: it was incomplete and misleading.
I track its progress and suggest changes.
In fact, I have a fix for the OPAM situation ready to be PR'ed.

At the same time I began on the plugin for OMZ.
Your suggestion along the lines of the npm plugin is exactly what I have in mind.
We could actually implement it that way today; but first I need the OPAM situation cleared.

In the end, I am all for maintaining as little code in the plugin and improving `ng complete`.

We're getting there eventually.

mcornella:
That's ok. If it helps, zsh sets the `$ZSH_VERSION` which is not there in bash. 

You can also look at [how `npm completion` is implemented](https://github.com/npm/npm/blob/master/lib/utils/completion.sh) but I see you'll have the same problem with `complete`. If that's the case you could swap the order, checking `compdef` and `compctl` first.

Just some ideas. Thanks!

PR #5304, opened 2016-08-14
Tag support for detached branche in git_prompt_info

jtyr:
If the HEAD is not a symbolic ref (e.g. refs/heads/master) but it points to a commit ID (branch is detached), the git_prompt_info shows commit ID even if the commit ID points has a tag. This patch allows to display the tag version if there is a tag for the HEAD commit ID.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -3,6 +3,7 @@ function git_prompt_info() {
   local ref
   if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
     ref=$(command git symbolic-ref HEAD 2> /dev/null) || \
+    ref=$(command git describe --tags --exact-match HEAD 2> /dev/null) || \
     ref=$(command git rev-parse --short HEAD 2> /dev/null) || return 0
     echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
   fi


jtyr:
The performance depends on the state of the branch. If the HEAD is a ref (e.g. `git checkout master`), there is only one command executed. If the branch is detached (e.g. `git checkout 8462d64`), there will be three commands executed. But if there is a tag pointing on the detached branch (e.g. `git checkout v1.0`), there will be only two commands executed. So the change affects performance of only detached branch which has no tag pointing on the commit. I believe this is an acceptable performance degradation.

Here is a simple performance assessment:

```
$ mkdir test
$ cd test
$ git init
$ touch README
$ git add -A
$ git commit -m "Initial commit"
[master (root-commit) 8462d64] Initial commit
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 README
$ git tag v1.0
$ time git checkout 8462d64
Note: checking out '8462d64'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 8462d64... Initial commit

real	0m0.004s
user	0m0.000s
sys	0m0.000s
$ time git checkout master
Switched to branch 'master'

real	0m0.004s
user	0m0.000s
sys	0m0.003s
$ time git checkout v1.1
HEAD is now at 8462d64... Initial commit

real	0m0.004s
user	0m0.000s
sys	0m0.003s
```

All the commands are pretty fast and their execution is almost negligible.

jtyr:
@robbyrussell Can I help with anything else to make this merged?

jtyr:
@robbyrussell Please could you review this?

bromeara:
This could probably be closed since gitfast plugin implements this feature.

robbyrussell:
Thanks for this, and I'm sorry — you answered my question about the performance cost the same day I asked it, in 2016, and then heard nothing back for nine years. That's on us.

The feature did eventually land, in #11318 in 2022. `lib/git.zsh` now falls back to `git describe --tags --exact-match HEAD` when HEAD is detached, which is the same behavior you proposed here first. Closing this one for that reason.

Two other people opened the same change while this sat here (#6487, #7677). You got there first.

PR #5308, opened 2016-08-15
Added Atom plugin for Mac OS X

rvadai:
This plugin adds the `atom` alias to ZSH to open the [Atom editor](https://atom.io).

**Only works on Mac OS X** at the moment

### DIFF
--- a/plugins/atom/atom.plugin.zsh
+++ b/plugins/atom/atom.plugin.zsh
@@ -1,0 +1,13 @@
+if  [[ "$OSTYPE" = darwin* ]]; then
+    local _atom_darwin_paths > /dev/null 2>&1
+    _atom_darwin_paths=(
+        "/Applications/Atom.app"
+        "$HOME/Applications/Atom.app"
+    )
+    for _atom_path in $_atom_darwin_paths; do
+        if [[ -a $_atom_path ]]; then
+            alias atom="open -a '$_atom_path'"
+            break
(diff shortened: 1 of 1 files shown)


kfkonrad:
No need to add this plugin. On OSX, Atom installs their commandline tool to /usr/local/bin/atom. In case /usr/bin/local isn't in your PATH already, just `export PATH=$PATH:/usr/local/bin/`. Not sure where it's installed on Linux, but it should be somewhere in the standard PATH.

rvadai:
I'm using Atom 1.9.8 on El Capitan and `/usr/local/bin/atom` doesn't exist. Maybe this was the case on the previous versions.

kfkonrad:
Well, the file does exist in recent versions: https://github.com/atom/atom/blob/1.9-releases/atom.sh
The file should be in `/Applications/Atom.app/Contents/Resources/app/atom.sh`
You might want to alias this file or create a plugin based on it.

rvadai:
`open` is an OS X command which will work even if they move this script to another location.

kfkonrad:
I found the reason you don't have `/usr/bin/local/atom`. You need to open Atom, click on `Atom` right next to the Apple Symbol and select `Install Shell Commands`. That should do the trick.

rvadai:
Yep it works!
I'm gonna close this PR.

Thanks!

PR #5314, opened 2016-08-17
Login shell after install

leoj3n:
Otherwise these files are not sourced:
1. /etc/zprofile
2. ~/.zprofile
3. /etc/zlogin
4. ~/.zlogin
5. ~/.zlogout
6. /etc/zlogout

Which can cause confusion.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -110,7 +110,7 @@ main() {
   echo 'p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.'
   echo ''
   printf "${NORMAL}"
-  env zsh
+  env zsh -l
 }
 
 main


mcornella:
Thanks, added to the list of problems at #5169.

leoj3n:
This is a bit tricky. If I were coming from Bash and had never used Zsh, then I would not have any user `~/.z*` files (probably) except the system might have some `/etc/z*` files like macOS:

```
bash$ cat /etc/zprofile
```
```sh
# system-wide environment settings for zsh(1)
if [ -x /usr/libexec/path_helper ]; then
       	eval `/usr/libexec/path_helper -s`
fi
```
Which may or may not make sense to tack on to a bash session's already existing `$PATH`.
```
bash$ cat /etc/zshrc
```
```sh
# Correctly display UTF-8 with combining characters.
if [ "$TERM_PROGRAM" = "Apple_Terminal" ]; then
       	setopt combiningchars
fi
```
Which probably would make sense to include in a new Zsh session.

So, really what should be done is if we're in a Zsh session, then `env zsh` but if we're in any other shell session (like Bash) then we should do `env zsh -l`.

leoj3n:
Or perhaps whether `chsh` succeeded or not could have some bearing on whether to use `-l` or not.

If this is going to be the user's new login shell (`chsh` succeeded), then we should probably emulate the login case from the start by using `zsh -l`.

Otherwise, zsh will never be opened as a login shell by the system, and most users will probably invoke it like `zsh`, not `zsh -l`.

leoj3n:
Another possibility if invoking a Zsh from within another Zsh is to do it like this `zsh -$-`.

This would preserve the settings of the user's current shell (would only make sense if current shell is Zsh)

leoj3n:
Inside of a script, `-$-` is the script shell's options, so not very useful unless the installation paradigm involves using `source`.

leoj3n:
Another reason not to use `-$-`: Since OMZ turns zsh options on and off, it would probably want a clean slate to work with.

AlJohri:
@leoj3n could you link me to some docs regarding `zsh -$-`? I'm having trouble googling it

leoj3n:
@AlJohri 

```
❯ echo zsh -$-
zsh -03459CDEJNQRTVWXZghilms
```

leoj3n:
@mcornella Since you merged this, you might also consider merging https://github.com/robbyrussell/oh-my-zsh/pull/5320

PR #5318, opened 2016-08-18
Set $SHELL to correct value after install

leoj3n:
This fixes a problem where `$SHELL` would equal f.ex `/bin/bash` after
running `env zsh`, breaking scripts that rely on `$SHELL` being set.

For example, GtiHub's hub [would think the login shell was set to bash, not zsh](https://github.com/github/hub/blob/713fd93b145e420e69884d7e12dd54426bd457bd/commands/alias.go#L44).

To illustrate:
```
bash $ env zsh
zsh  $ echo $SHELL
/bin/bash
zsh  $ exit
bash $ env SHELL=$(which zsh) zsh
zsh  $ echo $SHELL
/usr/local/bin/zsh
```

On further thought, this should only set `$SHELL` if `chsh` was successful.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -110,7 +110,7 @@ main() {
   echo 'p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.'
   echo ''
   printf "${NORMAL}"
-  env zsh
+  env SHELL=$(which zsh) zsh
 }
 
 main


leoj3n:
A better patch would be something like:

```
  TEST_LOGIN_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
  if [ "$TEST_LOGIN_SHELL" == "zsh" ]; then
    WHICH_ZSH="$SHELL"
  else
    WHICH_ZSH="$(which zsh)"
    if hash chsh >/dev/null 2>&1; then
      if ! grep "^${WHICH_ZSH}$" /etc/shells >/dev/null 2>&1; then
        WHICH_ZSH="$(grep /zsh$ /etc/shells | tail -1)"
        if hash "$WHICH_ZSH" >/dev/null 2>&1; then
          SHELL="$WHICH_ZSH"
        fi
      fi
    # Else, suggest the user do so manually.
    else
      printf "I can't change your shell automatically because this system does not have chsh.\n"
      printf "${BLUE}Please manually change your default shell to zsh!${NORMAL}\n"
    fi
  fi

  env "$WHICH_ZSH"
```

Note I haven't tried running this code.

leoj3n:
Done more properly in #5320

PR #5319, opened 2016-08-18
refresh .rake-tasks after changing anything inside lib/tasks

ivanovaleksey:
This PR introduces some fixes to the previous #5111.
Now it compares timestamp of anything inside of `lib/tasks` directory (and the directory itself) against `.rake_tasks`.
So removing anything would result in updating `.rake_tasks`.

I think test case might look like the following:
```
$ touch lib/tasks/baz.rake # update existing file
$ rake [tab] # should reload tasks

$ mkdir lib/tasks/foo
$ touch lib/tasks/foo/bar.rake  # create new file
$ rake [tab] # should reload tasks

$ rm -f lib/tasks/foo/bar.rake # remove .rake file in subfolder
$ rake [tab] # should reload tasks

$ rm -f lib/tasks/baz.rake # remove .rake file in root
$ rake [tab] # should reload tasks
```

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -12,13 +12,18 @@ _rake_does_task_list_need_generating () {
 }
 
 _is_rails_app () {
-  [[ -e "bin/rails" ]] || [ -e "script/rails" ]
+  [[ -e "bin/rails" ]] || [[ -e "script/rails" ]]
 }
 
 _tasks_changed () {
-  local is_changed=1
-  for file in lib/tasks/**/*.rake; do
-    if [[ $file -nt .rake_tasks ]]; then is_changed=0; fi
+  local is_changed paths
+  is_changed=1
+  paths=(lib/tasks lib/tasks/**/*(:f))
+  for path in $paths; do
+    if [[ $path -nt .rake_tasks ]]; then
+      is_changed=0
+      break
+    fi
   done
   return is_changed
 }


Gentleman-jun:
good

mcornella:
Doesn't it always stop at `[[ lib/tasks -nt .rake_tasks ]]`? My understanding on #5111 was that it was the only thing needed to check in all cases, even when a file was deleted. If you revert back to that behavior you can delete most of the `_tasks_changed` function.

ivanovaleksey:
As I understand folder`s _mtime_ is only changed when we create new file/directory or remove existing. The _mtime_ is not changed when we just edit existing file/directory.
For example
```
$ zstat +mtime lib/tasks # 1471521242

$ touch lib/tasks/new_tasks.rake
$ zstat +mtime lib/tasks # 1471765324 changed

$ echo 'some text' >> lib/tasks/new_tasks.rake
$ zstat +mtime lib/tasks # 1471765324 not changed

$ rm -f lib/tasks/new_tasks.rake
$ zstat +mtime lib/tasks # 1471765436 changed

$ mkdir lib/tasks/subdirectory
$ zstat +mtime lib/tasks # 1471765589 changed

$ touch lib/tasks/subdirectory/another_tasks.rake
$ zstat +mtime lib/tasks # 1471765589 not changed

$ echo 'some task' >> lib/tasks/subdirectory/another_tasks.rake
$ zstat +mtime lib/tasks # 1471765589

$ rm -f lib/tasks/subdirectory/another_tasks.rake
$ zstat +mtime lib/tasks # 1471765589

$ rm -rf lib/tasks/subdirectory
$ zstat +mtime lib/tasks # 1471765837
```
Notice that any changes in subdirectories (except of removing the the whole subdirectory) don't affect on parent directory _mtime_.

My initial logic in #5111 seemed to be wrong (though I tested it, perhaps I was missing something).

In this PR we check _mtime_ of all files and directories inside of `lib/tasks` so we can see if something was updated or removed somewhere within `lib/tasks`

plugins/rake-fast/rake-fast.plugin.zsh
```diff
@@ -12,13 +12,18 @@ _rake_does_task_list_need_generating () {
...
-  [[ -e "bin/rails" ]] || [ -e "script/rails" ]
+  [[ -e "bin/rails" ]] || [[ -e "script/rails" ]]
 }
 
 _tasks_changed () {
-  local is_changed=1
-  for file in lib/tasks/**/*.rake; do
-    if [[ $file -nt .rake_tasks ]]; then is_changed=0; fi
+  local is_changed paths
+  is_changed=1
```
mcornella: The `is_changed` variable can be deleted and do `return 1` or `0` instead. That way you also don't need the `break` in the for loop.

plugins/rake-fast/rake-fast.plugin.zsh
```diff
@@ -12,13 +12,18 @@ _rake_does_task_list_need_generating () {
...
+  paths=(lib/tasks lib/tasks/**/*(:f))
```
mcornella: What does `(:f)` do? I know `:` is to apply a modifier, and `f` is for filter based on file permission, but both of them together don't make sense.

mcornella: Ok found it on [Modifiers](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers):

> The following f, F, w and W modifiers work only with parameter expansion and filename generation. They are listed here to provide a single point of reference for all modifiers.

> f
>     Repeats the immediately (without a colon) following modifier until the resulting word doesn’t change any more.

Still, that doesn't seem to be doing anything.

mcornella:
Hi @ivanovaleksey, I've written a couple of suggestions. Here's the final function that I've come up with:
```zsh
_tasks_changed () {
  local -a paths
  paths=(lib/tasks lib/tasks/**/*(N))

  for path in $paths; do
    if [[ "$path" -nt .rake_tasks ]]; then
      return 0
    fi
  done

  return 1
}
```

The `(N)` is so that it doesn't show the `no matches found` error if there isn't any matching file.

mcornella:
I've fixed this on a branch in my repository, have a look if you want: [mcornella/oh-my-zsh@rake-fast-plugin-refactor](https://github.com/mcornella/oh-my-zsh/tree/rake-fast-plugin-refactor).

ivanovaleksey:
Hi @mcornella, thanks a lot for your advices and your code.
Sorry I didn't have enough time this week to answer you.
Should I fix this PR further or you will merge your branch?

plugins/rake-fast/rake-fast.plugin.zsh
ivanovaleksey: Yes, totally agree.

plugins/rake-fast/rake-fast.plugin.zsh
ivanovaleksey: I think `(:f)` should be placed here to get full paths i.e. based on file system root. But it doesn't work like this.
It seems like I was missing something.

mcornella:
I'll submit my branch as a PR so you can test it and give your approval. Give me a minute...

mcornella:
That's it, I've submitted #5343 for your review.

You'll notice I've kept your commit author information so it counts as a contribution as well.

Thanks!

Issue #5311, opened 2016-08-16
Oh My Zsh not enabled when doing `su` (root)


0x00hack3r:
I install this in my username and i when i do su : root (there is no themes)..

mcornella:
Use `sudo -E -s` to log as root, that way you don't have to install OMZ for root (which is not recommended). I have that aliased:
```zsh
alias suroot='sudo -E -s'
```

riteshpuj2013:
@mcornella your solution didn't work for me.

Can someone please help me use zsh as root?

sysRoot:
```
 jase@yxia  ~  sudo -E -s
[sudo] password for jase: 
[oh-my-zsh] Insecure completion-dependent directories detected:
drwxr-xr-x  11 jase users  4096 Jul 17 20:59 /home/jase/.oh-my-zsh
drwxr-xr-x 250 jase users 12288 Jul 17 20:59 /home/jase/.oh-my-zsh/plugins
drwxr-xr-x   2 jase users  4096 Jul 17 20:59 /home/jase/.oh-my-zsh/plugins/git

[oh-my-zsh] For safety, we will not load completions from these directories until
[oh-my-zsh] you fix their permissions and ownership and restart zsh.
[oh-my-zsh] See the above list for directories with group or other writability.

[oh-my-zsh] To fix your permissions you can do so by disabling
[oh-my-zsh] the write permission of "group" and "others" and making sure that the
[oh-my-zsh] owner of these directories is either root or your current user.
[oh-my-zsh] The following command may help:
[oh-my-zsh]     compaudit | xargs chmod g-w,o-w

[oh-my-zsh] If the above didn't help or you want to skip the verification of
[oh-my-zsh] insecure directories you can set the variable ZSH_DISABLE_COMPFIX to
[oh-my-zsh] "true" before oh-my-zsh is sourced in your zshrc file.
```
So... I am not clear on how to fix this without breaking anything... since you said it's not good to install Oh my ZSH as root.

intika:
Add this to `.zshrc`
```
su() {
	SUPARAM="su $*"
	bash -c "$SUPARAM"
}
```
If any issue run `reset` and retry... 
Note: this is for bash... zsh need to be run after su...

PR #5320, opened 2016-08-18
Greatly improve the OMZ install.sh

leoj3n:
Makes the installer more accomodating:
- Allow headless installs (on a fresh system). :arrow_forward: https://github.com/robbyrussell/oh-my-zsh/pull/5893#issuecomment-414151781
- Set $SHELL to the correct value on chsh success. :arrow_forward: https://github.com/robbyrussell/oh-my-zsh/pull/5318
- Handle edge cases in regards to locating and using zsh.
- Password entry prompt allows user to try again (after bad password).


leoj3n:
Headless Install:
---------------------
[asciicast](https://asciinema.org/a/cwkzhmos1pky2i4daxp6epe8y)

Password Retry:
---------------------
[asciicast](https://asciinema.org/a/8q6cokmle4hoayvuxs5fcdb0w)

Aborting Chsh:
-------------------
[asciicast](https://asciinema.org/a/6ru3evm0n8vc0t7299w2pmoo6)

Setting $SHELL on Chsh Success:
------------
[asciicast](https://asciinema.org/a/bebx6rx02xuo2cow4hylvm0t9)

leoj3n:
@robbyrussell Can you test?

leoj3n:
`0>/dev/null sh -c "$(<install.sh)"` to trigger "headless" for testing.

leoj3n:
:alarm_clock::smiling_imp:

leoj3n:
### Rebased and added two new commits:

- https://github.com/robbyrussell/oh-my-zsh/pull/5320/commits/b1205c87f1af1afd369c2ab12c463ddcdb326f3d Use a more portable test for no input TTY
- https://github.com/robbyrussell/oh-my-zsh/pull/5320/commits/7888d97cce838a91b396fd847875d92e6fea9fc8 Account for no zsh entry in /etc/shells
  > Because the installer no longer greps /etc/shells at the start.

### To try this out:

```sh
git clone https://github.com/robbyrussell/oh-my-zsh.git
cd oh-my-zsh
git fetch origin pull/5320/head:pr-5320
git checkout pr-5320
```

Before you run anything, move any existing configuration out of the way:

```sh
mv ~/.oh-my-zsh ~/.oh-my-zsh.bak
```

And change the shell to something other than Zsh:

```sh
chsh -s /bin/bash
# Enter your password
```

Now run the installer from the root of the just cloned git repo:

```sh
sh -c "$(<tools/install.sh)"
```

To see the added safety rails, try:

- Typing the wrong password.
- Entering an empty password.
- Cancelling with <kbd>CTRL-C</kbd>.

Reset as before, and run again and type the correct password.

### Testing No TTY

To try out the headless/unattended/batch/no-exec (or whatever you want to call it), reset the environment as before (move away ~/.oh-my-zsh; change shell is optional) and run:

```sh
0>/dev/null sh -c "$(<tools/install.sh)"
```

From the root of the just cloned git repo.

lhorace:
> `0>/dev/null sh -c "$(<tools/install.sh)"`

Unattended install using this method doesn't work here with Ansible. It would be nice if there was alternative method to doing this. E.g. `BATCH=true ./tools/install.sh`

leoj3n:
@lhorace I don't use Ansible, but the only way the changes in this PR wouldn't work is if the Ansible shell module doesn't support redirecting standard file descriptors.

lhorace:
I'm going to give https://docs.ansible.com/ansible/latest/modules/raw_module.html a shot. 

> Executes a low-down and dirty SSH command, not going through the module subsystem.

though

> There is no change handler support for this module.

I won't be able to manager failure cases.

> but the only way the changes in this PR wouldn't work is if the Ansible shell module doesn't support redirecting standard file descriptors.

I would doubt Ansible is a edge case, especially if this is the future preferred method for unattended install..

leoj3n:
@lhorace What is the code you're trying and what is the error you're seeing? Can't really help without that. I'd be surprised if there's not a simple solution to avoiding allocating a TTY for a command.

lhorace:
@leoj3n The problem here is the call to chsh, on some distribution, a password prompt is thrown which will hang the install waiting for input that will never arrive. Unfortunately, using your code as is, doesn't appear to work with Ansible shell module. Ansible will call `sh -c 0>/dev/null sh -c "$(<tools/install.sh)"`. Raw module passes it the remote machine as is but breaks if you want to install it on another user, as opposed to the current login user. E.g. `sudo -H -S -n -u root /bin/sh -c '"'"'echo BECOME-SUCCESS-kfujtfhyezlelpfwfduipvltweanupah; 0>/dev/null sh -c "$(</tmp/install.sh)"'"'"'`.

Looking over the script a few times, it appears to me that it would just be much easier to just git clone, and template `.zshrc`.

> I'd be surprised if there's not a simple solution to avoiding allocating a TTY for a command.

Besides the raw module? Not that I know over.

leoj3n:
@lhorace Thanks for testing.

I probably made a logic error in my code from commit https://github.com/robbyrussell/oh-my-zsh/commit/b1205c87f1af1afd369c2ab12c463ddcdb326f3d.

Can you try changing

```shell
if [ ! -t 0 ] && [ ! -p /dev/stdin ]; then
```

back to

```shell
if ! [ -t 0 ] || [ -p /dev/stdin ]; then
```

and let me know if that helps.

lhorace:
> if [ ! -t 0 ] && [ ! -p /dev/stdin ]; then

I found one instance here https://github.com/robbyrussell/oh-my-zsh/blob/1a861060de21fb7f3cc57e1535827772b0ac688e/tools/install.sh#L212, and change it accordingly. However it's failing to install OMZ properly. Git is not cloning.

See if I can look into it.

lhorace:
It breaks unfortunately with sudo, when trying to install as another user:

```
tput: unknown terminal \"unknown\"\nCloning into '/root/.oh-my-zsh'...
```

leoj3n:
@lhorace Probably fails because of `requiretty` in your distro. You could try removing the `requiretty` setting for your distro.

Although, I'm not sure why you need to run the OMZ installer with sudo? Think it's [not recommended](https://github.com/robbyrussell/oh-my-zsh/issues/5311#issuecomment-240106715).

lhorace:
> @lhorace Probably fails because of requiretty in your distro. You could try removing the requiretty setting for your distro.

Actually I disable requiretty earlier on in the playbook. Anyhow, you can safely ignore my earlie comment, it works just fine. I simply had to re-login/restart all `zsh` processes. 

>  Think it's [not recommended](https://github.com/robbyrussell/oh-my-zsh/issues/5311#issuecomment-240106715)

Thanks for pointing this out and I'm making amendments to my playbook. 

Though I need that fix as I'm pulling directly from this PR or you want me to submit a PR to your repository with the minor fix?

leoj3n:
@lhorace The logic should be correct now. Thanks again for testing. Please share the relevant part of your playbook with us when you've got it finalized!

lhorace:
Here's the relevant bits:

```
- name: Retrieving oh-my-zsh/install.sh from remote repository
  get_url:
    url: https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/5a770a94881cf735b9794f5aa8079f3f5cfd06df/tools/install.sh
    dest: /tmp
    sha256sum: 7f274cc43984588a6681055bd77f3df02706e1a275de20de1730af16aba0da24
    owner: rnduser
    mode: 0700

- name: Installing OMZ
  raw: 0>/dev/null sh -c /tmp/install.sh
```

mcornella:
Hi @leoj3n, thanks a lot for the many fixes you introduced here. I pulled some of these in #5169 which introduces other fixes and features. The no-TTY version can be used by calling the script with `--unattended` but it also works closing stdin as you did. I didn't add traps to the calls to chsh because I substituted that with checking if the chsh called succeded, and otherwise showing an error message *but* letting the install script continue. 

I'm sure you'll have a few more comments and suggestions which could improve the script further, so if you'd like to help speed up the testing and merge (which should happen early next week), you can wget or curl it from the URL `https://raw.githubusercontent.com/mcornella/oh-my-zsh/refactor-installer/tools/install.sh`. Read the comments at the top of the script to know how to call it.

Thanks!

mcornella:
Closed by #5169.

PR #5322, opened 2016-08-19
Clobber ssh environment regardless of setopt

DoomHammer:
To prevent

```
_plugin__start_agent:7: file exists: $HOME/.ssh/environment-$HOST
starting ssh-agent...
Could not open a connection to your authentication agent.
```


mcornella:
:+1:

PR #5326, opened 2016-08-20
Fix empty Git empty repo check

khenarghot:
In theme dogenpunk method for check Git repository empties  do not work in Linux and too slow for big repository, like kernel.

The default `grep` in Linux do not extend regular expressions so pattern for check not match. 

The `git log | grep "^fatal: ...."` is to slow for big repository. The limited log give the same result.


mcornella:
My `git log` message is `fatal: your current branch 'master' does not have any commits yet`. That, and the fact that `git log` will exit with an error if there isn't a commit, we should drop the `grep` of the check:

```diff
-if [[ $(git log 2>&1 > /dev/null | grep -c "^fatal: bad default revision") == 0 ]]; then
+if git log -n 1 >/dev/null 2>&1; then
```

khenarghot:
@mcornella  You are right. This solution is better. I will replace pull request.

PR #5328, opened 2016-08-21
Refactor of ssh-agent plugin

mcornella:
- Repost of #5261: deletes useless `/usr/bin/env` calls.
- Adds a README file.
- Formats to tabs, renames local variables and fixes needlessly complex variable syntax.
- Simplify PID check of last ssh-agent.
- Uses `ssh-agent -s` to force Bourne-style syntax on systems that have Csh-like default shells (for example, colleges or workplace).

Close #5261.

### DIFF
--- a/plugins/ssh-agent/README.md
+++ b/plugins/ssh-agent/README.md
@@ -1,0 +1,38 @@
+# ssh-agent plugin
+
+This plugin starts automatically `ssh-agent` to set up and load whichever
+credentials you want for ssh connections.
+
+To enable it, add `ssh-agent` to your plugins:
+
+```zsh
+plugins=(... ssh-agent)
+```
+
+## Instructions
+
+To enable **agent forwarding support** add the following to your zshrc file:
+
+```zsh
+zstyle :omz:plugins:ssh-agent agent-forwarding on
+```
+
+To **load multiple identities** use the `identities` style, For example:
+
+```zsh
+zstyle :omz:plugins:ssh-agent identities id_rsa id_rsa2 id_github
+```
+
+To **set the maximum lifetime of the identities**, use the `lifetime` style.
+The lifetime may be specified in seconds or as described in sshd_config(5)
+(see _TIME FORMATS_). If left unspecified, the default lifetime is forever.
+
+```zsh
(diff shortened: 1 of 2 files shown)


HuangShaoyan:
Hi @mcornella ,

The simplification of PID check causes ssh-agent to output some error on OS X.
```
ps: cmd: keyword not found
ps: no valid keywords; valid keywords:
starting ssh-agent...
```

Maybe we should replace 'cmd' with 'command' on OS X.

kavoori:
Hi @mcornella 

I just pulled in this commit and my iTerm and Terminal now prompts for passphrase every time I open a new tab with the following error messages as pointed out by @HuangShaoyan 
```
ps: cmd: keyword not found
ps: no valid keywords; valid keywords:
starting ssh-agent...
Enter passphrase for /Users/Rabi/.ssh/id_rsa:
```

I was able to get around the problem by reverting to older commit and I can confirm that the problem does not exist with this commit.

```
git checkout 3c8a333
```

Is anyone else reporting this?

HuangShaoyan:
Both 'ps -o cmd' and 'ps -o command' work on Ubuntu. According to man, 
```
cmd        CMD     see args. (alias args, command).
command    COMMAND see args. (alias args, cmd).
```

But 'ps -o cmd' is not support by OS X.
```
     args       command and arguments
     comm       command
     command    command and arguments
```

kavoori:
I can confirm that above workaround works. Thanks @HuangShaoyan for documenting it!

tonyseek:
The busybox `ps` will output error also:

```
ps: invalid option -- p
BusyBox v1.24.2 (2016-05-09 11:46:24 EDT) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

        -o COL1,COL2=HEADER     Select columns for display
        -T                      Show threads
starting ssh-agent...
```

I found this on a [Termux](https://termux.com) session.

mcornella:
Thanks for the fast response! I've changed the check command to `pgrep ssh-agent` which I've checked works in MacOS, busybox, OpenBSD, Solaris and all Linux systems. See 2c9d831.

Please update OMZ.

PR #5336, opened 2016-08-24
iTerm2 v3 tab commands syntax error fix

Multilul:
Fixes the "syntax error: expected line end ..." (-2741) errors that occur when using the tab, split_tab, and vsplit_tab commands in iTerm2 v3


mcornella:
:+1:

From [iterm2 Applescript docs](https://www.iterm2.com/documentation-scripting.html):

> #### Application
> The application exposes various properties and provides functions that are described in this section. For example:
> 
>     tell application "iTerm2"
>       create window with default profile
>     end tell

PR #5343, opened 2016-08-25
Repost of #5319 and refactor of rake-fast plugin

mcornella:
- Fix `_tasks_changed` function (repost of #5319).
- Fix formatting and syntax of plugin.
- Reword and reorganisation of README.

Close #5319

### DIFF
--- a/plugins/rake-fast/README.md
+++ b/plugins/rake-fast/README.md
@@ -2,22 +2,34 @@
 
 Fast rake autocompletion plugin.
 
-This script caches the output for later usage and significantly speeds it up. It generates a .rake_tasks cache file in parallel to the Rakefile. It also checks the file modification dates to see if it needs to regenerate the cache file.
+This plugin caches the output for later usage and significantly speeds it up.
+It generates a `.rake_tasks` cache file in parallel to the Rakefile. It also
+checks the file modification time to see if it needs to regenerate the cache
+file.
 
-This is entirely based on [this pull request by Ullrich Schäfer](https://github.com/robb/.dotfiles/pull/10/), which is inspired by [this Ruby on Rails trick from 2006](http://weblog.rubyonrails.org/2006/3/9/fast-rake-task-completion-for-zsh/).
+This is entirely based on [this pull request by Ullrich Schäfer](https://github.com/robb/.dotfiles/pull/10/),
+which is inspired by [this Ruby on Rails trick from 2006](http://weblog.rubyonrails.org/2006/3/9/fast-rake-task-completion-for-zsh/).
 
 Think about that. 2006.
+
+----------
+
+Since August of 2016, it also checks if it's in a Rails project and looks at
+rake files inside `lib/tasks` and their modification time to know if the
+cache file needs to be regenerated.
 
 ## Installation
 
 Just add the plugin to your `.zshrc`:
 
-```bash
-plugins=(foo bar rake-fast)
+```zsh
+plugins=(... rake-fast)
 ```
 
 You might consider adding `.rake_tasks` to your [global .gitignore](https://help.github.com/articles/ignoring-files#global-gitignore)
 
 ## Usage
 
-`rake`, then press tab
+Type `rake`, then press tab.
+
+If you want to force the regeneration of the `.rake_tasks` file, run `rake_refresh`.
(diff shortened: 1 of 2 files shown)


ivanovaleksey:
Hi Marc!
Thank you a lot for the help!
I believe you didn't do any mistakes so it works fine.
I try to test it at this weekend.

ivanovaleksey:
Marc when I am on your branch I don't have any basic commands, for example no `cat`, `rm`, etc.
Is something wrong with my configuration?

mcornella:
I'm not sure, check out this howto explaining how to fetch pull requests: https://help.github.com/articles/checking-out-pull-requests-locally/

ivanovaleksey:
Thanks for the link. 
I will try it in some hours.
But I am not sure if it can be the case.
I am on your `rake-fast-plugin-refactor` branch so I already have all your commits.

ivanovaleksey:
Marc I have noticed that commands are existing at the beginning but when I run `rake [tab]` it gives me 
```
rake _rake:6: command not found: cat
_rake:6: command not found: cat
_rake:6: command not found: cat
```
And now commands are gone.

It seems like the problem is in the curly braces [here](https://github.com/robbyrussell/oh-my-zsh/pull/5343/files#diff-ef46552e72ba783013e8227056272bdbR2).
I have replaced to `()` and it works. I don't know what `{}` means. 
Do you know why it behaves like this?

mcornella:
> It seems like the problem is in the curly brackets here.
> I have replaced to () and it works. I don't know what {} means. 
> Do you know why it can behave like this?

That's really weird, it shouldn't change anything. The only difference is that the `()` parenthesis runs the code in a subshell (separate from the rest), while the `{}` braces runs it in the same shell. It shouldn't affect the result...

Restore the code back to using braces, restart the shell and run `functions -t _rake` before trying to autocomplete rake. This will turn on tracing for the `_rake` function and will spit out the commands as they are run. Please post that.

mcornella:
Ok, I think I found the problem at using the name `$path` inside `_task_changed`, which then modifies the `path` variable for the rest of the shell. That's why if run inside a subshell it doesn't affect anything.

Will post a patch soon.

ivanovaleksey:
It seems like your are right Marc.
I have renamed `$path` and now it works with `{}`.
Sorry for this stupid mistake. 
When I was writing the code I thought about `$path` can conflict with one of reserved variables. 
But I thought it `$PATH` not `$path`.
At the moment everything worked so I thought it wouldn't be mistake to use `$path`.

mcornella:
Yeah it's a particularity of zsh, `$path` is the array version of `$PATH`. No worries!

ivanovaleksey:
Now `rake [tab]` prints `Generating .rake_tasks...` in the same line. 
Like `rake Generating .rake_tasks...`. Don't now is it worth to mention ...

Also [last commit](https://github.com/robbyrussell/oh-my-zsh/pull/5343/commits/22bcc585146f4a0d9336eb6be6835547fa8fbc14) after removing `lib/tasks` check this case doesn't work:
```
touch lib/tasks/some.rake
rake [tab] # regenerates .rake_tasks
rm -f lib/tasks/some.rake
rake [tab] # doesn't regenerate .rake_tasks because file doesn't no longer exist
```
After removing the file `lib/tasks` timestamp has changed so I think we should check it to.

mcornella:
> Also last commit after removing lib/tasks check this case doesn't work:

That's a good point, I force-pushed the patch.


> Now rake [tab] prints Generating .rake_tasks... in the same line.

I'm messing around with making this message better, I don't know how to properly show messages from completion files without messing with the prompt.

mcornella:
Ok, I've tried many things and nothing works so I'll leave that as is. The rest of the completion seems to work great, so if you agree I'll merge this right away.

ivanovaleksey:
Marc one thing with checking only `.rake` file timestamp:
```
mkdir lib/tasks/foo
touch lib/tasks/foo/bar.rake
rake [tab] # regenerates .rake_tasks
------------------------------------
rm -f lib/tasks/foo/bar.rake
rake [tab] # doesn't regenerate .rake_tasks because there was only one .rake file in directory and it was gone (lib/tasks/foo directory timestamp was updated)
```

mcornella:
You're right again, I changed it so that it's only a rename of the variable names. Force-pushed!

ivanovaleksey:
Now it seems to work as expected.
Thank you Marc!

mcornella:
Thank you for testing everything, you've been very patient!

PR #5346, opened 2016-08-26
put .zsh-update in $ZSH/cache/

akozlins:
Put .zsh-update file in ZDOTDIR or HOME (if ZDOTDIR is not defined).
May help keep HOME dir _clean_.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -7,7 +7,7 @@ function _current_epoch() {
 }
 
 function _update_zsh_update() {
-  echo "LAST_EPOCH=$(_current_epoch)" >! ~/.zsh-update
+  echo "LAST_EPOCH=$(_current_epoch)" >! ${ZDOTDIR:-${HOME}}/.zsh-update
 }
 
 function _upgrade_zsh() {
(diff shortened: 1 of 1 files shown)


mcornella:
Good idea!

A better location though is `$ZSH_CACHE_DIR` which is by default in `$ZSH/cache`. This directory is ignored by git, whereas `$ZDOTDIR` might be in someone's dotfiles and tracked by git.

akozlins:
Thanks for idea.
I introduced variable ZSH_UPDATE which is used as location for update file,
by default this location is set to $ZSH/cache/.zsh-update.
One thing that worries me is that using this location as default may
confuse some users as there will be .zsh-update in HOME directory
which is not used.

mcornella:
This was closed and I can't reopen because your master branch was force-pushed. I've recreated the changes with a minor correction on [mcornella/oh-my-zsh@refactor-ohmyzsh-update](https://github.com/mcornella/oh-my-zsh/tree/refactor-ohmyzsh-update). 

It's not yet ready to merge though because of what you said, the transition from ~/.zsh-update to $ZSH_CACHE_DIR/.zsh-update should be better handled.

akozlins:
I realised that i did not created separate branch for pull request, so I closed this one and forgot to reopen with new branch.
Here is reopened pull request robbyrussell/oh-my-zsh/pull/5415

PR #5347, opened 2016-08-26
add alias gcb

allenlsy:


mcornella:
[This is already in the git plugin](https://github.com/robbyrussell/oh-my-zsh/blob/f258083/plugins/git/git.plugin.zsh#L67).

PR #5348, opened 2016-08-26
Disable zsh file globbing when running pip

kevgathuku:
Certain pip install commands which involve square brackets such as `pip install requests[security]
` or `pip install -e .[testing]` fails with the following error: `zsh: no matches found:`

This patch disables globbing for the pip command, and makes the pip command works for such cases.

### DIFF
--- a/plugins/pip/pip.plugin.zsh
+++ b/plugins/pip/pip.plugin.zsh
@@ -76,3 +76,6 @@ zsh-pip-test-clean-packages() {
         echo "the djangopypi2 index is fine"
     fi
 }
+
+alias pip="noglob pip" # allows square brackets for pip command invocation
+


mcornella:
:+1: Great PR!

kevgathuku:
Thanks @mcornella
Happy to help ☺

j2kun:
@mcornella This PR broke the use of pip in virtual environments. This alias overrides the virtualenv's pip and uses the systemwide pip. It also prevents one from _seeing_ which pip is being used, because `which pip` spits out `pip is aliased to 'noglob pip'`.

While the intentions were good, this PR should be reverted and quotes should be used to avoid globbing.

mcornella:
> It also prevents one from seeing which pip is being used, because which pip spits out `pip is aliased to 'noglob pip'`.

You can bypass that by using `which -a pip` or `command -v pip`. While I agree that using quotes is enough, I'd like to keep the change if it's possible.

> This PR broke the use of pip in virtual environments.

I'm not very familiar with pip and virtualenv\*, how does this break it exactly?

j2kun:
Can't seem to reproduce the behavior. Probably was my mistake.

Still, if this user doesn't like the default shell behavior, they should put the alias in their own personal `.zshrc`, and not force counterintuitive command preferences on others. And now since I don't want zsh autoupdates messing with my own personal reversion of this commit, I'll have to turn off autoupdates :(

mcornella:
If you feel this is too confusing or it messes with your workflow I can consider reverting this. The thing is that we don't have usage statistics so I can't say if people use that more or less.

It's also a common pattern to alias commands with `noglob`, for instance the `git` alias has it. And people often complain more about the `no matches found` error than of not finding where a command is coming from (which is can be bypassed with the commands I wrote in the previous comment). 

So I'll wait to see if someone else finds this change annoying. In the meantime you can add `unalias pip` at the end of your `zshrc` file. There's not much else I can do I'm afraid.

daveman1010221:
Does this same behavior apply to pip3? I.e.:
`which -a pip`
returns:
`pip: aliased to noglob pip
/some_path/bin/pip`

whereas:
`which -a pip3`
returns:
`/some_path/bin/pip3
/usr/local/bin/pip3
/usr/local/bin/pip3`

In other words, it appears that pip3 does not get this same 'noglob' alias as does pip, which makes scripting inconsistent?

mcornella:
It doesn't apply that to pip3 no, would you be willing to submit a PR?

PR #5350, opened 2016-08-29
add stackoverflow

qneyrat:
add search with stackoverflow in web-search plugin :)


mcornella:
This is the second time someone asked for this (#5260), but it's already available in the [frontend-search](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/frontend-search) plugin.

PR #5358, opened 2016-08-31
Add composer folder recognizer

diogoazevedos:
Composer can be in other folder location, in my OS is in `$HOME/.config/composer` by default.


thomasruiz:
@robbyrussell fixed by https://github.com/robbyrussell/oh-my-zsh/commit/4fba92e04fa9b62b2259abc45eb92ca6a74f1639

PR #5359, opened 2016-08-31
Make ssh-agent plugin tolerate simultaneous login.

wca:
This works by creating a directory, and relying on a function-local trap to
ensure that it is cleaned up on exit/interrupt/etc.  This approach should be
portable to all OSs zsh runs on.


nunojpg:
@wca can you take a look at my PR #5585? Maybe if we manage to combine all the ssh-agent PRs, we can make it easier to be merged.

PR #5367, opened 2016-09-02
always source settings of old pgp-agent, if applicable

depressiveRobot:
Fixes the problem that the `GPG_AGENT_INFO` environment variable is only exported,
if no `gpg-agent` instance is already running.


mcornella:
Obsolete.

PR #5374, opened 2016-09-05
Adding nodenv plugin based on rbenv plugin.

janlindblom:
This plugin adds support for the node version manager nodenv (https://github.com/nodenv/nodenv).

The plugin is based on the rbenv plugin since nodenv itself is heavily inspired/based on rbenv.


Kriechi:
This should probably be based on the more recent pyenv plugin which does not call the very slow `brew --prefix` command everytime zsh is started.
I prepared a similar PR for rbenv: https://github.com/robbyrussell/oh-my-zsh/pull/7626

PR #5377, opened 2016-09-06
add more vim-like binds to vi-mode

DanielFGray:
adds vim motions for change/delete inside/around, and `cc` and `dd` from vi-mode

this may be out of scope for `vi-mode` since these are more `vim-mode` binds, please feel free to discuss

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -28,6 +28,96 @@ bindkey '^?' backward-delete-char
 bindkey '^h' backward-delete-char
 bindkey '^w' backward-kill-word
 
+bindkey -M vicmd 'cc' vi-change-whole-line
+bindkey -M vicmd 'dd' kill-whole-line
+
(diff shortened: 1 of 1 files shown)


plugins/vi-mode/vi-mode.plugin.zsh
```diff
@@ -28,6 +28,96 @@ bindkey '^?' backward-delete-char
...
+change-around() {
+  zle delete-in
+  zle vi-backward-char
+  zle vi-delete-char
+  zle vi-delete-char
+  zle vi-insert
+}
+zle -N change-around
+bindkey -M vicmd 'ca' change-around
+
```
okapia: Again, this doesn't seem to do anything that isn't already in core zsh unless you've got a slightly older version of zsh (before 5.0.8). What does bindkey -M viopp -L show you?
There should already be bindings for iw, aW etc that will work with vi-change (c) and vi-delete (d).
If you want i(, i" etc, look at Functions/Zle/select-bracketed and select-quoted.

PR #5383, opened 2016-09-07
No need to set PS1 by default.

12425:
Say I just want to add a simple line to `~/.zshrc` to inject my script.
Part of my `~/.zshrc` will be
```
# User configuration
source /path/to/my/config

source $ZSH/oh-my-zsh.sh
```

I set `PROMPT` in my script, but it will be overwritten by `PS1=`.
I don't like to add another line after `source $ZSH/oh-my-zsh.sh`, and I don't think `PS1` must be set defaultly. Therefore can we remove it, or add a condition checking for the setting?

Thank you!

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -32,9 +32,6 @@ else
     SCREEN_NO=""
 fi
 
-# Apply theming defaults
-PS1="%n@%m:%~%# "
-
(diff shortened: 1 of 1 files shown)


mcornella:
You seem to have a broken zshrc file, you should put all user configurations after the line `source ...`. That way you override any OMZ settings you want.

PR #5385, opened 2016-09-07
add docker commit option to list all containers and images

Wingless-Archangel:
add options in docker commit to be able to see all containers and images


mcornella:
Fixed on #5354.

PR #5386, opened 2016-09-07
Add FirewallD plugin

robjtede:
Could be too opinionated but this is what I use for firewalld.


robjtede:
do new plugins _require_ a readme?

mcornella:
Yes that's the goal. You can look at the [`tig` plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/tig/README.md).

robjtede:
Readme added

mcornella:
That's perfect thanks!

PR #5395, opened 2016-09-11
change git current branch old email with new email in batches

adispring:


plugins/git/git.plugin.zsh
```diff
@@ -32,6 +32,23 @@ function work_in_progress() {
     echo "WIP!!"
   fi
 }
+#change current branch old email with new eamil
+gmvemail() {
+  export OLD_GIT_EMAIL=$1
```
mcornella: Instead of using `export` use `local` here.

plugins/git/git.plugin.zsh
```diff
@@ -32,6 +32,23 @@ function work_in_progress() {
...
+  export NEW_GIT_EMAIL=$2
+  git filter-branch --commit-filter '
+    if [ "$GIT_AUTHOR_EMAIL" = "$OLD_GIT_EMAIL" ];
+    then
+            GIT_AUTHOR_EMAIL="$NEW_GIT_EMAIL";
+            git commit-tree "$@";
+    else
+            git commit-tree "$@";
+    fi' HEAD
+  unset OLD_GIT_EMAIL
```
mcornella: This isn't needed after you apply https://github.com/robbyrussell/oh-my-zsh/pull/5395/files#r78296768.

plugins/git/git.plugin.zsh
```diff
@@ -32,6 +32,23 @@ function work_in_progress() {
...
+  unset NEW_GIT_EMAIL
+  BRANCH_NAME=$(git symbolic-ref --short HEAD)
```
mcornella: You should also use `local` here, or delete the variable and put the `$()` directly in the command below.

adispring: I change this as you said

plugins/git/git.plugin.zsh
adispring: using local does not work here, only export works well, I changed a name to avoid modifying env var by accident

plugins/git/git.plugin.zsh
```diff
@@ -32,6 +32,22 @@ function work_in_progress() {
     echo "WIP!!"
   fi
 }
+#change current branch old email with new eamil
+gmvemail() {
+  export GMVEMAIL_GIT_EMAIL_OLD=$1
+  export GMVEMAIL_GIT_EMAIL_NEW=$2
```
adispring: Local var doesn't work here, I changed name to avoid modifying env var by accident

plugins/git/git.plugin.zsh
```diff
@@ -32,6 +32,22 @@ function work_in_progress() {
...
+    if [ "$GIT_AUTHOR_EMAIL" = "$GMVEMAIL_GIT_EMAIL_OLD" ];
+    then
+            GIT_AUTHOR_EMAIL="$GMVEMAIL_GIT_EMAIL_NEW";
+            git commit-tree "$@";
+    else
+            git commit-tree "$@";
+    fi' HEAD
+  unset GMVEMAIL_GIT_EMAIL_OLD
+  unset GMVEMAIL_GIT_EMAIL_NEW
+  git update-ref -d refs/original/refs/heads/$(git symbolic-ref --short HEAD)
```
adispring: changed here as you said

plugins/git/git.plugin.zsh
mcornella: Ok, then set up that inside a subshell:
```zsh
(
  export GMVEMAIL_GIT_EMAIL_OLD=$1
  export GMVEMAIL_GIT_EMAIL_NEW=$2
  git filter-branch ...
)
```

or just set the variables before the command, like so:
```zsh
OLD_EMAIL=$1 NEW_EMAIL=$2 git filter-branch --commit-filter '
    ...
' HEAD
```

I like the second one better.

plugins/git/git.plugin.zsh
```diff
@@ -34,18 +34,14 @@ function work_in_progress() {
 }
 #change current branch old email with new eamil
 gmvemail() {
-  export GMVEMAIL_GIT_EMAIL_OLD=$1
-  export GMVEMAIL_GIT_EMAIL_NEW=$2
-  git filter-branch --commit-filter '
-    if [ "$GIT_AUTHOR_EMAIL" = "$GMVEMAIL_GIT_EMAIL_OLD" ];
+  OLD_EMAIL=$1 NEW_EMAIL=$2 git filter-branch --commit-filter '
```
adispring: Just modify as you said.
Thanks for giving me tips to handle temp env var using subshell. Lots of shell knowledge have to learn.

mcornella: That's perfect 👌

You know, the shell world is full of obscure features so don't worry about it, I only learned most of them through serendipity 😊.

PR #5396, opened 2016-09-11
Create aliases for npm plugin

hjpotter92:
Add the following 3 aliases:
- npmvM
- npmvm
- mpmvp

To assist in versioning the project

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -35,3 +35,7 @@ alias npmst="npm start"
 # Run npm test
 alias npmt="npm test"
 
+# aliases for versioning commands
+alias npmvM='npm version major '
+alias npmvm='npm version minor '
+alias mpmvp='npm version patch '


slavaGanzin:
I suggest to separate npm completions and aliases. 
Generally I can't get why you need them if you have awesome zsh completion...

hjpotter92:
@slavaGanzin zsh completion requires me to type in `npm ve<tab>` which takes ~2s on my system. There is no completion yet for the `version` subcommand completion. So, I won't be able to complete them to `version major` etc.

slavaGanzin:
@hjpotter92 just curious. Why it takes 2s?

hjpotter92:
@slavaGanzin nearly 30 active plugins, a few of my own precmd etc. listeners, old laptop, some other factors I don't know about!

mcornella:
It's probably the node process starting up after starting the computer and caching files.

Unfortunately there is currently no way to separate aliases from completion. If you just want the completion, perhaps you can just add the zsh completion to your zshrc. From [npm/completion.sh](https://github.com/npm/npm/blob/master/lib/utils/completion.sh):

```zsh
  _npm_completion() {
    local si=$IFS
    compadd -- $(COMP_CWORD=$((CURRENT-1)) \
                 COMP_LINE=$BUFFER \
                 COMP_POINT=0 \
                 npm completion -- "${words[@]}" \
                 2>/dev/null)
    IFS=$si
  }
  compdef _npm_completion npm
```

PR #5402, opened 2016-09-11
New plugin weather information

wachambo:
This plugins let you get nice weather information of any location (your location if not specified).
It uses the wego web fronted [wttr.in](https://github.com/chubin/wttr.in) to obtain weather information.

Usage: $weather {location}


mcornella:
Has no interest.

PR #5413, opened 2016-09-14
Fix battery.plugin when acpi writes to stderr

reedriley:
I have a machine which outputs the following when `acpi` is run:
```
❯ acpi
No support for device type: power_supply
```

This ends up showing everywhere I include `battery_pct()` - including my prompt!  It's an easy fix - just make it consistent which arguments we're passing to `acpi` in the battery plugin.

### DIFF
--- a/plugins/battery/battery.plugin.zsh
+++ b/plugins/battery/battery.plugin.zsh
@@ -72,7 +72,7 @@ elif [[ $(uname) == "Linux"  ]] ; then
 
   function battery_pct() {
     if (( $+commands[acpi] )) ; then
-      echo "$(acpi | cut -f2 -d ',' | tr -cd '[:digit:]')"
+      echo "$(acpi 2&>/dev/null | cut -f2 -d ',' | tr -cd '[:digit:]')"
     fi
   }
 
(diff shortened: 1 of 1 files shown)


plugins/battery/battery.plugin.zsh
```diff
@@ -72,7 +72,7 @@ elif [[ $(uname) == "Linux"  ]] ; then
 
   function battery_pct() {
     if (( $+commands[acpi] )) ; then
-      echo "$(acpi | cut -f2 -d ',' | tr -cd '[:digit:]')"
+      echo "$(acpi 2&>/dev/null | cut -f2 -d ',' | tr -cd '[:digit:]')"
```
mcornella: The extra ampersand (`&`) isn't needed to redirect stderr and it's not common, though it doesn't hurt. The same applies to the other change.

reedriley:
No worries - and I'm happy to tweak it.

BUT - if I do, I'm fixing all of the references to `acpi 2&>/dev/null` in the file to be consistent.

mcornella:
Oh that's awesome, I didn't knew there were other instances of that. Thanks for the change!

PS. looks like Github's new review process is absolutely perfect :)

PR #5415, opened 2016-09-16
Move .zsh-update to $ZSH_CACHE_DIR

akozlins:
reopen pull request robbyrussell/oh-my-zsh/pull/5346


tools/check_for_upgrade.sh
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env zsh
 
+if [ -z "$ZSH_UPDATE" ]; then
```
mcornella: You can get rid of this check. The cache directory is a good enough location that it doesn't need to be customizable.

tools/check_for_upgrade.sh
```diff
@@ -1,13 +1,17 @@
...
+
 zmodload zsh/datetime
 
 function _current_epoch() {
   echo $(( $EPOCHSECONDS / 60 / 60 / 24 ))
 }
 
 function _update_zsh_update() {
-  echo "LAST_EPOCH=$(_current_epoch)" >! ~/.zsh-update
+  echo "LAST_EPOCH=$(_current_epoch)" >! $ZSH_UPDATE
```
mcornella: This variable should be quoted: `"$ZSH_UPDATE"`.

tools/check_for_upgrade.sh
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env zsh
 
+if [ -z "$ZSH_UPDATE" ]; then
+  ZSH_UPDATE="$ZSH/cache/.zsh-update"
```
mcornella: `$ZSH_CACHE_DIR` should be available here but I haven't checked. Use that instead of `$ZSH/cache` if it works.

tools/check_for_upgrade.sh
```diff
@@ -29,9 +33,9 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
+if [ -f $ZSH_UPDATE ]
```
mcornella: Same here, quoted. And the line below as well.

akozlins:
thanks for comments.
it seems that zsh_cache_dir is not set on first start,
it is only set in oh_my_zsh.sh after call to check_for_upgrade.sh,
so i set zsh_update directly

akozlins:
implemented in #6719

PR #5417, opened 2016-09-16
Added 'gbd' alias for 'git branch -d'

ferrerluis:
- Simple addition to the set of aliases available for `git`.
- Wondered why it was not available by default (let me know if there is a special reason).
- No other `git` alias has the same string (`gbd`) defined for it.


robbyrussell:

Cheers,
Robby

PR #5422, opened 2016-09-20
Add more docker compose aliases

leifcr:
Adds the following aliases:

```zsh
alias dco='docker-compose'
alias dcr='docker-compose run'
alias dce='docker-compose exec'
```

And sorts the aliases similar to `docker-compose help` order


willmendesneto:
+1 Tested and working. @robbyrussell Could you take a look on this PR, please?

mcornella:
Thanks!

PR #5426, opened 2016-09-20
Update docker completion from upstream

zhimsel:
Latest: https://github.com/docker/docker/blob/be9e3f59e625a7be05f21a23f6debfb3f6728573/contrib/completion/zsh/_docker


mcornella:
Pull in 9d91e92.

PR #5427, opened 2016-09-20
Fix cp plugin completion

### LINKED ISSUE #5420: cp plugin cpv has not path autocompletion
As title.
https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/cp/cp.plugin.zsh

Can anyone help?

mcornella:
Fixes #5420.

- Changed `-h` for `-hhh` which shows numbers in units of 1024 (k, M, G, T).
- Added `-r` to recurse directories.
- Removed `--`: use only if needed. That way we can do `cpv --help` and other combinations.


mtekman:
why the change from an alias to a function - alias would work fine, no?

PR #4263, opened 2015-08-18
Pacaur aliases in the Arch Linux plugin

DoctorJellyface:
[Pacaur](https://wiki.archlinux.org/index.php/Pacaur) is another AUR helper, though arguable one of the more popular ones. I (obv) use it too, and I missed aliases a lot, so I made them myself.

It's basically the same as the Yaourt/Pacman aliases. I tested every command I included. I also documented all my changes.

Hope you like it.

PR #5430, opened 2016-09-21
Implement a lockfile to alleviate update prompt in multiple terminals

### LINKED ISSUE #3766: Implement a lockfile to alleviate update prompt in multiple terminals

melvinsh:
This prevents the "Would you like to check for updates?" prompt from showing up after starting the upgrade process and opening a new shell.

Fixes #3766. I'm aware of the comments on that issue but I really don't think it's worth making this a lot more complex for an extremely rare race condition with no impact.

Current situation

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -10,10 +10,19 @@ function _update_zsh_update() {
   echo "LAST_EPOCH=$(_current_epoch)" >! ~/.zsh-update
 }
 
+function _create_lock() {
+  touch ~/.zsh-update.lock
+}
+function _remove_lock() {
+  rm ~/.zsh-update.lock
+}
+
 function _upgrade_zsh() {
+  _create_lock
   env ZSH=$ZSH sh $ZSH/tools/upgrade.sh
   # update the zsh file
   _update_zsh_update
+  _remove_lock
 }
 
 epoch_target=$UPDATE_ZSH_DAYS
@@ -21,6 +30,9 @@ if [[ -z "$epoch_target" ]]; then
   # Default to old behavior
   epoch_target=13
 fi
+
+# Cancel upgrade if upgrade is already in progress
+[[ ! -f ~/.zsh-update.lock ]] || return 0
 
 # Cancel upgrade if the current user doesn't have write permissions for the
 # oh-my-zsh directory.


schurig:
+1

FabianWolff:
> I'm aware of the comments on that issue but I really don't think it's worth making this a lot more complex for an extremely rare race condition with no impact.

It doesn't have to be a lot more complex. And while it is true that there would be no significant impact, why not implement it right in the first place? Atomic locking works easily with `mkdir`, `ln -s` and `rm`, although to use `rm`, one would have to use a "reverse lock", meaning a lock that is deleted upon acquisition and touched upon release. Have a look at #5456 to see what I mean (it uses `mkdir`, but the other methods would have worked, too).

melvinsh:
@robbyrussell @mcornella I spend a lot of time on two PR's that both work fine, and without consulting me you merge some other guy's PR. Not really what I was expecting from you and definitely not motivating me to continue working on this project.

mcornella:
That's totally fair. I understand you completely and I assume my responsibility. In this case I just went with the solution that was simplest and technically best. It's nothing personal but I understand you feeling that way.

The thing is that I'm not that experienced at being a maintainer and I don't have external help, so I will make mistakes like ignoring the work of others instead of replying that I don't quite like the implementation.This is the “I kinda like this but it’s just not right” situation from [this excellent post on tricks of being a maintainer by Jessie Frazelle](https://blog.jessfraz.com/post/the-art-of-closing/). I'm just winging it like everybody else, so I'll screw up and for that I apologize.

This happened as well in https://github.com/robbyrussell/oh-my-zsh/pull/4263#issuecomment-250975914, which I forgot about, so I made a symbolic gesture to the author in commit eaf1816. It wasn't the first time, and I'm sure it won't be the last (but I'll try). I say this so that you have some insight into my thought process, I'm not making excuses and it's totally understandable that you don't care.

The **TL;DR**: I'm sorry, I screwed up and I definitely will try not to make the same mistake again. Please don't feel discouraged from making another contribution, I'll make sure you have another chance at making a mark.

Sorry again! :frowning_face:

melvinsh:
Hey @mcornella, thanks for your reply and sharing your thoughts. These things happen and I don't blame you. Maybe in the future I can contribute with something else 😄

mcornella:
I'm glad you understand. Thanks! 😊

PR #5431, opened 2016-09-21
Add copybuffer function and keybinding

bradurani:
binds ctrl-o to copy the command line buffer to the system clipboard

### DIFF
--- a/plugins/copybuffer.zsh
+++ b/plugins/copybuffer.zsh
@@ -1,0 +1,14 @@
+# copy the active line from the command line buffer 
+# onto the system clipboard (requires clipcopy plugin)
+
+copybuffer () {
+  if which clipcopy &>/dev/null; then
+    echo $BUFFER | clipcopy
+  else
+    echo "you must install clipcopy to use this keybinding"
+  fi
+}
+
+zle -N copybuffer
+
+bindkey "^O" copybuffer
--- a/plugins/copybuffer/README.md
+++ b/plugins/copybuffer/README.md
@@ -1,0 +1,14 @@
+# `copybuffer` plugin
+
+This plugin adds a keybinding that copies the command line buffer contents to
+the system clipboard.
+
+```zsh
+plugins=(... copybuffer)
+```
+
+## Features
+
+| Keybinding | Command        | Description                                        |
+|------------|----------------|----------------------------------------------------|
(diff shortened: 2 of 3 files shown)


plugins/copybuffer.zsh
```diff
@@ -0,0 +1,14 @@
+# copy the active line from the command line buffer 
+# onto the system clipboard (requires clipcopy plugin)
+
+copybuffer () {
+  if which clipcopy &>/dev/null; then
```
mcornella: [`clipcopy` is a function provided by OMZ](https://github.com/robbyrussell/oh-my-zsh/blob/d51ee38/lib/clipboard.zsh#L18). This check shouldn't be needed but it's ok.

plugins/copybuffer.zsh
```diff
@@ -0,0 +1,14 @@
...
+    echo $BUFFER | clipcopy
+  else
+    echo "you must install clipcopy to use this keybinding"
```
mcornella: Since `clipcopy` is a function provided by OMZ, I would change this message to something more appropriate. For instance:
```
clipcopy function not found. Make sure you have Oh My Zsh correctly installed.
```

bradurani:
look ok?

plugins/copybuffer/README.md
```diff
@@ -0,0 +1,14 @@
...
+This plugin adds a keybinding that copies the command line buffer contents to
+the system clipboard.
+
+```zsh
+plugins=(... copybuffer)
+```
+
+## Features
+
+| Keybinding | Command        | Description                                        |
```
mcornella: No need for this table and features heading. Just a brief description of the CTRL+o keybinding will do.

mcornella: Also a brief example of why would you use this will help understand.

plugins/copybuffer/README.md
```diff
@@ -0,0 +1,14 @@
...
+|------------|----------------|----------------------------------------------------|
+| `ctrl-o`   | `copybuffer`   | Copies the command line buffer to system clipboard |
```
mcornella: "command line buffer" is too close to implementation, I'd prefer "currently written command line (`$BUFFER`)", which has both the abstract and the low-level information.

PR #5433, opened 2016-09-21
New git aliases 'grm' and 'grmc'

ferrerluis:
- `alias grm='git rm'`.
- `alias grmc='git rm --cached'`.
- No other `git` aliases will be overridden.
- `grm` and `grmc` aren't commands within `oh-my-zsh` yet.


alexDarcy:
I would be interesting in having these aliases in the main branch too. Thank you.

PR #5439, opened 2016-09-23
add task description to fabric completion

spacewander:
Before:

```
☁  scripts [git-deploy] ⚡ fab es_
es_clean_idx  es_health     es_store
```

Now we have task description appended:

```
☁  scripts [git-deploy] ⚡ fab es_
es_clean_idx  -- 删除elasticsearch的索引。
es_health     -- 检查elasticsearch健康情况。
es_store      -- 检查elasticsearch数据存储情况
```


mcornella:
Fixed logic to also show tasks without a docstring, and added quoting of colons (`:`) in docstrings. Also renamed the `_targets` function to the more namespace-specific `_fab_targets`.

Tested with Fabric 2 with this `fabfile.py`:
```py
from fabric import task

@task
def hello(c):
	"""
	This runs hello()
	in multiple lines
	"""
	print("Hello world!")

@task
def run_this(c):
	"""
	Run: this thing
	"""
	pass

@task
def no_docstring(c):
	pass

@task
def another_task(c):
	pass
```

PR #5440, opened 2016-09-23
Display command return code in the half-life theme

povesteam:
If the previous command returns an error code, it is displayed right-aligned, along with the big lambda character **Λ** which can have 2 meanings:
- Looks like an _up arrow_ to indicate which command returned the error code.
- Looks like The iconic Half Life logo, but in CAPS (Λ vs λ) to emphasise the error.

image


mcornella:
There doesn't seem to be interest in this. You can put the `RPS1` variable definition in your zshrc file.

PR #5443, opened 2016-09-24
Reference alias `gref` for unkown git aliases

ferrerluis:
- Added alias `gref` to `plugins/git/git.plugin.zsh`.
- Looks through aliases and greps those that include `git`.
- Then **greps again** for a custom string provided by the user.
- `gref` does **not** clash with any other alias or command within `oh-my-zsh`.
- **Increases productivity** and **speeds up references** to aliases directly related to git.
- Recently rebased (09/24/2016, 1:58 AM) with `robbyrussell/oh-my-zsh master`

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -200,6 +200,7 @@ alias grbc='git rebase --continue'
 alias grbi='git rebase -i'
 alias grbm='git rebase master'
 alias grbs='git rebase --skip'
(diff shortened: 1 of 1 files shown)


ferrerluis:
@robbyrussell Just friendly reminder this PR is still here 😄

mcornella:
Users should know how to look for aliases, including checking the [plugin README](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/README.md) (the ones from the git plugin will be there shortly).

PR #5451, opened 2016-09-26
added docker-compose run alias

ivoba:
added "docker-compose run" alias


ivoba:
i just saw there is a similar PR https://github.com/robbyrussell/oh-my-zsh/pull/5422
please use this.

PR #5454, opened 2016-09-26
nvm plugin only load nvm script if command doesn't already exist

### LINKED ISSUE #5453: nvm plugin reloads existing nvm
If nvm has already been loaded, the nvm plugin will load it again.

See: https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/nvm/nvm.plugin.zsh#L4-L5

Some people may want to load nvm on their own, but still utilize the completions from nvm plugin. For example, executing `nvm use default` on shell initializing can result in a significant slowdown.

See: https://github.com/creationix/nvm/issues/860

So if you want to load nvm with `--no-use` parameter, or implement other workarounds to this issue, you would lose all gains when loading nvm plugin through omz for the completions.

saschabratton:
Fixes #5453.

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -1,5 +1,8 @@
 # Set NVM_DIR if it isn't already defined
 [[ -z "$NVM_DIR" ]] && export NVM_DIR="$HOME/.nvm"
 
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

PR #5456, opened 2016-09-26
Implement a locking mechanism to avoid multiple update prompts

### LINKED ISSUE #3766: Implement a lockfile to alleviate update prompt in multiple terminals

FabianWolff:
This would fix #3766 and is a better solution than #5430 because it avoids even the rare race condition mentioned in the discussion there by using a proper, atomic locking mechanism (mkdir).

See http://mywiki.wooledge.org/BashFAQ/045 for reference.

Related: #5257.

Fixes #3766.
Closes #5430.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -29,31 +29,36 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
+if mkdir "$ZSH/log/update.lock" 2>/dev/null
 then
-  . ~/.zsh-update
-
-  if [[ -z "$LAST_EPOCH" ]]; then
-    _update_zsh_update && return 0;
-  fi
-
-  epoch_diff=$(($(_current_epoch) - $LAST_EPOCH))
-  if [ $epoch_diff -gt $epoch_target ]
+  if [ -f ~/.zsh-update ]
   then
-    if [ "$DISABLE_UPDATE_PROMPT" = "true" ]
+    . ~/.zsh-update
+
+    if [[ -z "$LAST_EPOCH" ]]; then
+      _update_zsh_update && return 0;
+    fi
+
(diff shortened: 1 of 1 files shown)


tools/check_for_upgrade.sh
```diff
@@ -29,31 +29,36 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
```
mcornella: This is perfect. No possibility of race condition here, only one of the processes will get the succesfull return code. :+1:

tools/check_for_upgrade.sh
```diff
@@ -29,31 +29,36 @@ fi
...
+if mkdir "$ZSH/log/update.lock" 2>/dev/null
 then
-  . ~/.zsh-update
+  if [ -f ~/.zsh-update ]
+  then
+    . ~/.zsh-update
 
-  if [[ -z "$LAST_EPOCH" ]]; then
-    _update_zsh_update && return 0;
-  fi
```
mcornella: Can you make this `rmdir` directly?

tools/check_for_upgrade.sh
```diff
@@ -29,31 +29,36 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-if [ -f ~/.zsh-update ]
+if mkdir "$ZSH/log/update.lock" 2>/dev/null
```
mcornella: I'd like to have these `if then` statements in the same line, an extra commit converting all these to `if; then` will do.

mcornella:
There seems to be a bug in the lines shown in the review. The lines referred to in the comments are the ones shown here: https://github.com/robbyrussell/oh-my-zsh/pull/5456/files?w=0.

PR #5457, opened 2016-09-26
up alias for cd ../

mancilladev:


mcornella:
You can just do `..` with OMZ so I see no need for this.

PR #5460, opened 2016-09-27
Added pacaur support to the arch linux plugin.

Majora320:
I added pacaur compatibility to the arch linux plugin, and refactored a bit of the logic.

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -64,7 +64,7 @@ elif (( $+commands[yaourt] )); then
   }
 else
   upgrade() {
-    pacman -Syu
+    sudo pacman -Syu
   }
 fi
 
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -1,12 +1,4 @@
-if ! (( $+commands[yaourt] )); then
-  upgrade() {
-    sudo pacman -Syu
-  }
-else
-  upgrade () {
-    yaourt -Syu
-  }
-
+if (( $+commands[yaourt] )); then
   alias yaconf='yaourt -C'
   alias yaupg='yaourt -Syua'
   alias yasu='yaourt -Syua --noconfirm'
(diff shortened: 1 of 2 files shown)


slavaGanzin:
@Majora320 Maybe you will be interesed to add this into plugin:

https://wiki.archlinux.org/index.php/Reflector

```sudo reflector --verbose -l 10 --sort rate --save /etc/pacman.d/mirrorlist &```

**Comment by Majora320** [comment]
Reflector has so many useful options (country code, etc.) that I think a custom alias created by the user would be more useful than a prescripted plugin @slavaGanzin, but good idea all the same. I should look into some more arch linux tools and perhaps add them here.

File: plugins/archlinux/archlinux.plugin.zsh [hunk 1]
```diff
@@ -35,6 +27,47 @@ else
...
+  upgrade() {
+    pacaur -Syu
+  }
+elif (( $+commands[yaourt] )); then
+  upgrade() {
+    yaourt -Syu
+  }
+else
+  upgrade() {
+    pacman -Syu
```
mcornella: You left the `sudo` here.

mcornella:
This is a duplicate of #4263, which fell out of my mind. That one has become corrupted after the big archlinux refactor, but I'd like to check if those are similar aliases and to add them to the README, as that other one does.

johnramsden:
I'd appreciate this being added. ```pacaur``` is far better than ```yaourt``` and should be considered a first class citizen.

mcornella:
Yes, as I said, I'm waiting for the README modification.

Majora320:
I'm editing to add the commands to the README and the `sudo` to the last else branch now.

mcornella:
💯 no need to hurry!

Majora320:
Sorry, I sounded kind of offensive, didn't mean to be hostile :)

Majora320:
Alright, done.

Majora320:
Wait never mind, I left `### pacaur` lowercase :P

Majora320:
Alright, now I can say that I'm done.

plugins/archlinux/archlinux.plugin.zsh
Majora320: Correct, fixed.

johnramsden:
Thanks!

mcornella:
Nice job thanks!

PR #5463, opened 2016-09-27
globalias fix #4834

### LINKED ISSUE #4834: Globalias
Can't find any mentions of global alias expand plugin 
http://blog.patshead.com/2012/11/automatically-expaning-zsh-global-aliases---simplified.html

May I add it to plugins?

slavaGanzin:
Expand your aliases 

alias

Explained:
http://blog.patshead.com/2012/11/automatically-expaning-zsh-global-aliases---simplified.html

Closes #4834

### DIFF
--- a/plugins/globalias/globalias.plugin.zsh
+++ b/plugins/globalias/globalias.plugin.zsh
@@ -5,7 +5,7 @@ globalias() {
 }
 zle -N globalias
 
-# space expands all global aliases
+# space expands all aliases, including global
 bindkey -M emacs " " globalias
 bindkey -M viins " " globalias
 
--- a/plugins/globalias/globalias.plugin.zsh
+++ b/plugins/globalias/globalias.plugin.zsh
@@ -12,3 +12,6 @@ bindkey -M viins " " globalias
 # control-space to make a normal space
 bindkey -M emacs "^ " magic-space
 bindkey -M viins "^ " magic-space
+
+# normal space during searches
+bindkey -M isearch " " magic-space
--- a/plugins/globalias/globalias.plugin.zsh
+++ b/plugins/globalias/globalias.plugin.zsh
@@ -1,0 +1,11 @@
+globalias() {
+   zle _expand_alias
+   zle expand-word
+   zle self-insert
+}
+zle -N globalias
+bindkey -e " " globalias
+bindkey -v " " globalias
+bindkey -e "^ " magic-space           # control-space to bypass completion
+bindkey -v "^ " magic-space
+bindkey -M isearch " " magic-space # normal space during searches
--- a/plugins/globalias/globalias.plugin.zsh
+++ b/plugins/globalias/globalias.plugin.zsh
@@ -4,8 +4,11 @@ globalias() {
    zle self-insert
 }
 zle -N globalias
-bindkey -e " " globalias
-bindkey -v " " globalias
-bindkey -e "^ " magic-space           # control-space to bypass completion
-bindkey -v "^ " magic-space
-bindkey -M isearch " " magic-space # normal space during searches
+
+# space expands all global aliases
+bindkey -M emacs " " globalias
+bindkey -M viins " " globalias
+
+# control-space to make a normal space
+bindkey -M emacs "^ " magic-space
+bindkey -M viins "^ " magic-space
--- a/plugins/globalias/README.md
+++ b/plugins/globalias/README.md
@@ -1,0 +1,37 @@
+#Globalias
+
+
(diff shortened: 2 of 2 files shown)


plugins/globalias/globalias.plugin.zsh
```diff
@@ -0,0 +1,11 @@
+globalias() {
+   zle _expand_alias
+   zle expand-word
+   zle self-insert
```
mcornella: I'd get rid of the self-insert here. That way you can do `....[space]` and continue making the path.

plugins/globalias/globalias.plugin.zsh
```diff
@@ -0,0 +1,11 @@
...
+}
+zle -N globalias
+bindkey -e " " globalias
+bindkey -v " " globalias
+bindkey -e "^ " magic-space           # control-space to bypass completion
```
mcornella: Separate these bindkeys by a newline and put the comments on top instead of on one side:

```zsh
# space expands all global aliases
bindkey -e " " globalias
bindkey -v " " globalias

# control-space to make a normal space
bindkey -e "^ " magic-space
bindkey -v "^ " magic-space

# use normal space during searches
bindkey -M isearch " " magic-space
```

plugins/globalias/globalias.plugin.zsh
mcornella: Oh right, if this isn't there then it doesn't add the extra space when there's nothing to expand. Forget about it.

plugins/globalias/globalias.plugin.zsh
mcornella: The fact that you do the `bindkey -v` makes it use the vicmd keymap. Use `bindkey -M` instead:
- `bindkey -M vicmd` instead of `bindkey -v`
- `bindkey -M emacs` instead of `bindkey -e`

slavaGanzin: @mcornella Thanks for this review and for showing me ```-M``` key

I switched to strict 'emacs' and 'viins' modes and so 'isearch' rebinding now obsolete

plugins/globalias/globalias.plugin.zsh
```diff
@@ -4,8 +4,11 @@ globalias() {
    zle self-insert
 }
 zle -N globalias
-bindkey -e " " globalias
-bindkey -v " " globalias
-bindkey -e "^ " magic-space           # control-space to bypass completion
-bindkey -v "^ " magic-space
-bindkey -M isearch " " magic-space # normal space during searches
```
mcornella: I do get that globalias is run instead of adding a space when I'm on `isearch` mode. Perhaps that's why I'm on emacs mode by default, and that messes it up. If I readd this back it works like I'd expect.

mcornella: It must be this from the [Zsh Line Editor docs](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#index-local-keymaps):

> While for normal editing a single keymap is used exclusively, in many modes a local keymap allows for some keys to be customised. For example, in an incremental search mode, a binding in the isearch keymap will override a binding in the main keymap but **all keys that are not overridden can still be used**.

plugins/globalias/globalias.plugin.zsh
```diff
@@ -4,8 +4,11 @@ globalias() {
...
+
+# space expands all global aliases
```
mcornella: Since `zle expand-word` is also used, this needs to be changed to `all aliases, including global`. I just tried it and every alias is expanded. That should also written in the README.

mcornella:
I've updated the README a bit. Have a look and I'll merge this if you're ok with the changes.

slavaGanzin:
Thanks for your notices during this PR

mcornella:
My pleasure 😊 Good to go then!

albertnetymk:
I find this plugin to be quite useful. However, I only want to expand my aliases sparingly. Therefore, I swapped the keymappings, `space` and `ctrl-space`. Surprisingly, it didn't work until I changed `self-insert` to `magic-space`. On second thoughts, should `self-insert` be `magic-space` from the first place? After all, regardless of what keybindings the user uses, there should be a space after the globalias expansion, isn't it?

slavaGanzin:
You can switch bindkeys:
```
bindkey -e " " magic-space
bindkey -e "^ " globalias
```

But I get strange '^@' symbol in the end of completion. Don't know why

albertnetymk:
That's what I meant by "Surprisingly, it didn't work". You can do the same, I suppose, "I changed `self-insert` to `magic-space`."

PR #5464, opened 2016-09-27
Add 'cdi', a function to edit current working path

jerome-pouiller:
Yet another way to manage current working directory. With 'cdi' (Change Directory Interactive), you can edit your current path as you edit a variable with 'vared'.

BR,

### DIFF
--- a/plugins/cdi/cdi.plugin.zsh
+++ b/plugins/cdi/cdi.plugin.zsh
@@ -1,0 +1,20 @@
+# vim: set sw=4 expandtab:
+#
+# Licence: GPL
+# Created: 2016-09-27 14:41:53+02:00
+# Main authors:
+#     - Jérôme Pouiller <jezz@sysmic.org>
+#
+# "Change Directory Interactive"
+#
+# Allow to change directory by editing current working directory path.
+#
+
+cdi() {
+    local _TMP=$PWD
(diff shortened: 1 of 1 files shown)


plugins/cdi/cdi.plugin.zsh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Licence: GPL
+# Created: 2016-09-27 14:41:53+02:00
+# Main authors:
+#     - Jérôme Pouiller <jezz@sysmic.org>
+#
+# "Change Directory Interactive"
+#
+# Allow to change directory by editing current working directory path.
+#
```
mcornella: Delete this whole comment and put the relevant information in a README (for example, [the `pj` plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/pj/README.md)).

plugins/cdi/cdi.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+# vim: set sw=4 expandtab:
+#
+# Licence: GPL
```
mcornella: If you want to add this plugin to OMZ you're accepting adding it with the MIT license. I'm afraid these are both incompatible.

PR #5474, opened 2016-09-29
add "svn_get_repo_url" function in the svn plugin

LanderlYoung:
The svn_get_repo_url function echoes the full path to the current repo URL.


mcornella:
Going through `svn` plugin issues and this is the last one open. Since this is really old I'll close it, but reopen if you still need this with an explanation of the reason behind it (I'd think it's more useful to get the 'Repository Root' though). Thanks

PR #5477, opened 2016-09-29
git auto fetch

slavaGanzin:
Automatically fetch changes from all remotes on cwd

Use case: I maintain a few projects as reviewer and often forget to update all repos before writing reviews.

Besides it awesomely works with git-prompt

### DIFF
--- a/plugins/git-auto-fetch/README.md
+++ b/plugins/git-auto-fetch/README.md
@@ -1,6 +1,6 @@
 # Git auto fetch
 
-Automatically fetches all changes from all remotes every time you cd into yout git-initialized project.
+Automatically fetches all changes from all remotes while you are working in git-initialized directory.
 
 ####Usage
 Add ```git-auto-fetch``` to the plugins array in your zshrc file:
@@ -8,7 +8,14 @@ Add ```git-auto-fetch``` to the plugins array in your zshrc file:
 plugins=(... git-auto-fetch)
 ```
 
-Every time you change directory to your git project all remotes will be fetched in background. Log of ```git fetch --all``` will be saved into .git/FETCH_LOG
+Every time you launch a command in your shell all remotes will be fetched in background.
+By default autofetch will be triggered only if last fetch was done at least 60 seconds ago.
+You can change fetch interval in your .zshrc:
+```
+GIT_AUTO_FETCH_INTERVAL=1200 #in seconds
+```
+Log of ```git fetch --all``` will be saved into .git/FETCH_LOG
+
 
 ####Toggle auto fetch per folder
 If you are using mobile connection or for any other reason you can disable git-auto-fetch for any folder:
--- a/plugins/git-auto-fetch/README.md
+++ b/plugins/git-auto-fetch/README.md
@@ -1,0 +1,22 @@
+# Git auto fetch
+
+Automatically fetches all changes from all remotes every time you cd into yout git-initialized project.
+
+####Usage
+Add ```git-auto-fetch``` to the plugins array in your zshrc file:
+```shell
+plugins=(... git-auto-fetch)
+```
+
+Every time you change directory to your git project all remotes will be fetched in background. Log of ```git fetch --all``` will be saved into .git/FETCH_LOG
+
+####Toggle auto fetch per folder
+If you are using mobile connection or for any other reason you can disable git-auto-fetch for any folder:
+
+```shell
+$ cd to/your/project
+$ git-auto-fetch
+disabled
+$ git-auto-fetch
+enabled
+```
--- a/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh
+++ b/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh
@@ -18,10 +18,10 @@ function git-auto-fetch {
     echo "${fg_bold[red]}disabled${reset_color}")
 }
 
-eval "original-$(declare -f zle-line-init)"
+eval "override-git-auto-fetch-$(declare -f zle-line-init)"
 
 function zle-line-init () {
   git-fetch-all
-  original-zle-line-init
+  override-git-auto-fetch-zle-line-init
 }
 zle -N zle-line-init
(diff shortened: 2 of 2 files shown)


mcornella:
This will be a problem when the user is on a mobile connection where the data consumed is very important. It also doesn't have a way to prevent or cancel the update.

If you can find a good solution to that go on and add a README to the plugin as well. Thanks!

slavaGanzin:
Maybe I can add global GIT_AUTO_FETCH which defaults to true and check it before autofetch. So you can negate it to prevent fetching?

And add GIT_AUTO_FETCH_MESSAGE
```
if [ $GIT_AUTO_FETCH_MESSAGE ]; then
echo "Fetching all remotes"
echo "To prevent autofetching(for this session): export GIT_AUTO_FETCH=false"
echo "To prevent showing this message in future 'echo GIT_AUTO_FETCH_MESSAGE=false >> ~/.zshrc"
fi
```

Or show that message if GIT_AUTO_FETCH is not set. So it can be added to your .zshrc.local file and works only on hosts where this behavior is acceptable.

slavaGanzin:
.git/AUTO_FETCH guard file can be added to mark repo as autofetched?

And two functions:
git_autofetch_enable
git_autofetch_disable

slavaGanzin:
@mcornella 

 https://github.com/slavaGanzin/oh-my-zsh/tree/25fcf0c265c682f092ce49a6849e5e09b38dffa9/plugins/git-auto-fetch

You always welcome to add some clarifications to my README.md

fazlearefin:
Hi @slavaGanzin ,

I've spent about 15-20 minutes testing your plugin. And here are my comments.

**`git-auto-fetch` function**
The function `git-auto-fetch` should be renamed to `git-no-auto-fetch`. On a fresh git repo, when I run this command for the first time it says **enabled** and creates the guard file and hence stops auto fetch. However this is contradictory to the **enabled** message shown when executing the function for the first time. So either change the function name or swap the messages in the `echo`.

**Functionality of the plugin itself**
Looking at the code it looks like a `git fetch` executes only when `cd`-ing into the base dir of a git repo. I understand this is the intent of your plugin but compared to `magicmonty/bash-git-prompt` your plugin is very much limited. The following actions also need to work:

- when `cd` into a sub-directory of the same git repo from the base dir of the repo
- when directly `cd` into the sub dir of a git repo
- when staying in the same git repo and the same dir but pressing return/enter

You might be asking why the additional actions should do a `git fetch`. The reason is I am not the only person working on the branch. While I am working on the branch there might be other developers who have `git push`ed and until I do a `git fetch` I won't get the latest code.

slavaGanzin:
@fazlearefin Many thanks for ideas and testing

I've implemented:
- when cd into a sub-directory of the same git repo from the base dir of the repo
- when directly cd into the sub dir of a git repo

And fixed wrong messaging, of course

But for:
- when staying in the same git repo and the same dir but pressing return/enter

I summon @mcornella for recommendation of suitable zle widget for that task. Mark can you help me?

fazlearefin:
@slavaGanzin 

Thanks for implementing the new requested changes so quickly. I have tested your latest commit and here's the status:

- `git-auto-fetch` function: **OK**
- when cd into a sub-directory of the same git repo from the base dir of the repo: **OK**
- when directly cd into the sub dir of a git repo: **OK**
- when staying in the same git repo and the same dir but pressing return/enter: **Pending**

slavaGanzin:
@fazlearefin Can you test it again, please?

slavaGanzin:
@mcornella @fazlearefin I'm using this on a daily basis for a month and has no issues with it. So I think this PR can be merged:
# it has documentation
# it works as described
# it is optional for oh-my-zsh users, so we not forcing something

Code has a bit of "magic", but I think nobody would contribute except me

fazlearefin:
@slavaGanzin 

Thanks for working on this feature. I tried your latest commit and it seems to have some issues. After enabling the `git-auto-fetch` plugin, when I launch a new zsh shell, I get this:

`git-fetch-all:wait:6: pid 0 is not a child of this shell`

And when I `cd` to a git repository, the shell just becomes very slow. I have to wait 4-5 seconds to type something to get any output.

slavaGanzin:
@fazlearefin I accidentally pushed my experimental version (in which I wanted to update RPROMPT to show "fetching" message)

Now it was reversed to stable one. Please retest. I'm sorry for that situation

slavaGanzin:
@fazlearefin @mcornella Any progress?

fazlearefin:
@slavaGanzin Apologies for the late reply. I've tested it and I can say it works! Good job!

slavaGanzin:
@mcornella can we merge this now?

kolemannix:
I would use this once merged

slavaGanzin:
@kolemannix can you test it before merge?

To avoid changing your .oh-my-zsh folder you can simply:
```
cd ~
wget https://raw.githubusercontent.com/slavaGanzin/oh-my-zsh/d2dfa69419845daebcfd20fed3253ae06faa2876/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh
echo "source ~/git-auto-fetch.plugin.zsh" >> ~/.zshrc
zsh
```

kolemannix:
Sure; I loaded it in just now. Will use all day today and see if I have any issues. Anything specifically to test for?

slavaGanzin:
@kolemannix Nothing specifically. Thanks

igauravsehrawat:
Still not merged?

fazlearefin:
i have been using the [zsh pure theme ](https://github.com/sindresorhus/pure) which has this functionality built-in.

slavaGanzin:
> Still not merged?
@fazlearefin Because of no active maintenance of this repo, I stepped a bit further and switched to fish shell (which I highly recommend). Same plugin: https://github.com/slavaGanzin/config/blob/master/fish/functions/fish_git_auto_fetch.fish

irvingpop:
This is a good PR and is working well for me (put in `$ZSH_CUSTOM/plugins/`) 😄  thank you @slavaGanzin for developing it!

mcornella:
Thanks for your contribution!

PR #5483, opened 2016-09-30
Change confusing 12h-time in xiong-chiamiov themes

firewavemtl:


mcornella:
This changes the date from `Fri Sep 30, 11:36` to `Fri Sep 30 23:36:54 2016`. That's too long, I'd rather change from `%I` to `%H` to use 24-hour-style time.

firewavemtl:
Here we go mcornella. It's too bad AM/PM cannot be added. But every one understand 24h.

mcornella:
> It's too bad AM/PM cannot be added. But every one understand 24h.

It's just a matter of space. Terminal space is very precious :wink: Thanks!

PR #5488, opened 2016-10-03
set better default colors for GNU ls instead of none.

xuhdev:
GNU coreutils ship a color setup command by default which can be used to
set a good default color theme for ls:

  https://www.gnu.org/software/coreutils/manual/html_node/dircolors-invocation.html

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -5,7 +5,7 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 # Enable ls colors
 if [ "$DISABLE_LS_COLORS" != "true" ]
 then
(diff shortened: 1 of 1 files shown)


mcornella:
This breaks the color scheme if you already have one. Let me work on a better implementation, but thanks for the idea!

xuhdev:
Thanks for taking care of this issue! Can you keep this open until you push your implementation? Thanks!

mcornella:
Working on it right now, will be merged soon don't worry.

mcornella:
Merged on 99dbf32. Cheers!

xuhdev:
:+1: Besides, the `uname -s` can be replaced by `OSTYPE` for faster loading

PR #5496, opened 2016-10-05
Use OSTYPE instead of uname whenever possible for better speed.

xuhdev:


lib/theme-and-appearance.zsh
```diff
@@ -6,11 +6,11 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 
 if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
   # Find the option for using colors in ls, depending on the version
-  if [[ "$(uname -s)" == "NetBSD" ]]; then
+  if [[ "$OSTYPE" == netbsd* ]]; then
```
mcornella: I'm not sure that's right, I don't have NetBSD machines to test with.

lib/theme-and-appearance.zsh
```diff
@@ -6,11 +6,11 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
...
     # On NetBSD, test if "gls" (GNU ls) is installed (this one supports colors);
     # otherwise, leave ls as is, because NetBSD's ls doesn't support -G
     gls --color -d . &>/dev/null && alias ls='gls --color=tty'
-  elif [[ "$(uname -s)" == "OpenBSD" ]]; then
+  elif [[ "$OSTYPE" == openbsd* ]]; then
```
mcornella: Same here, I don't have OpenBSD machines to test this.

plugins/battery/battery.plugin.zsh
```diff
@@ -64,7 +64,7 @@ if [[ "$OSTYPE" = darwin* ]] ; then
     [[ $(ioreg -rc "AppleSmartBattery"| grep '^.*"IsCharging"\ =\ ' | sed -e 's/^.*"IsCharging"\ =\ //') == "Yes" ]]
   }
 
-elif [[ $(uname) == "Linux"  ]] ; then
+elif [[ "$OSTYPE" = linux-gnu*  ]] ; then
```
mcornella: Just `linux*` will do.

plugins/sublime/sublime.plugin.zsh
```diff
@@ -1,4 +1,4 @@
-if [[ $('uname') == 'Linux' ]]; then
+if [[ "$OSTYPE" == linux-gnu* ]]; then
```
mcornella: Same, `linux*`

lib/theme-and-appearance.zsh
xuhdev: [Zsh `OSTYPE` comes from autoconf's `host_os`](https://github.com/zsh-users/zsh/blob/8f2ce89a0c80f6a0ecdc7c1189d6a94f620882f4/configure.ac#L46), and autoconf uses `host_os` to specify the bsds in [this way](http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=lib/autoconf/functions.m4;h=66abe2946d3617e5d91686ee9c164f004c5850a0;hb=2b4844a760d0046b98d93256232d4774c23120be#l914). The BSD OSTYPEs here should be correct.

plugins/sublime/sublime.plugin.zsh
xuhdev: I changed this to `*-gnu*`, because this plugin is really looking for a GNU env, not a Linux env. Setting this incorrectly would break kFreeBSD, which has a BSD kernel and GNU userland.

xuhdev: I changed it to `linux*` now. Seems I have misunderstood this plugin before.

PR #5497, opened 2016-10-05
Removing the user name and the Host name

rsm23:
As for me and many other developers, without the username and the
hostname is much much better and cleaner bash screen, any one can have
them back by just uncommenting them by removing the “#” from the
begining of the line.


mcornella:
You can just [add `DEFAULT_USER=$USER`](https://github.com/robbyrussell/oh-my-zsh/wiki/Themes#agnoster) in your zshrc file. Many others like having that.

PR #5500, opened 2016-10-07
Add non 0 exit code for missing jump targets

sirdiego:
This allows for the user to combine the jump command with
something else. In my example cd and jump are now combined
like this:

```bash
jumpcd() {
        jump $1 > /dev/null || cd $1
}
alias cd="jumpcd"
```

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -9,7 +9,7 @@
 export MARKPATH=$HOME/.marks
 
 jump() {
-	cd -P "$MARKPATH/$1" 2>/dev/null || echo "No such mark: $1"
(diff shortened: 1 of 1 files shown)


plugins/jump/jump.plugin.zsh
```diff
@@ -9,7 +9,7 @@
 export MARKPATH=$HOME/.marks
 
 jump() {
-	cd -P "$MARKPATH/$1" 2>/dev/null || echo "No such mark: $1"
+	cd -P "$MARKPATH/$1" 2>/dev/null || (echo "No such mark: $1"; return 1)
```
mcornella: A little style fix: use braces `{}` instead of spaces `()` which runs it in a subshell, with the added cost of creating a new process. The rest if perfect :+1:

PR #5504, opened 2016-10-08
Add OPAM

dannywillems:
This is the basic lines added automatically in the shell configuration file when we use `opam init`. It's mandatory.


mcornella:
Has no interest.

PR #5509, opened 2016-10-09
fixes file not found on MSYS2, it points to HOME directory

arlyoneel:
This error appears at start on MSYS2, this line fixes the problem.


mcornella:
I've pushed a better version of this in #5518. Please let me know if it works.

PR #5511, opened 2016-10-10
Git plugin - add alias for `git stash --all`

GantMan:
`gstall` = stash ignored, untracked, and tracked files.


mcornella:
Very useful thanks!

PR #5516, opened 2016-10-10
Fix ls colouring for Darwin.

rossmcf:
This change fixes the following error on OS X Sierra:

`ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]`

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -6,7 +6,7 @@ export LSCOLORS="Gxfxcxdxbxegedabagacad"
 
 if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
   # Find the option for using colors in ls, depending on the version
-  if [[ "$(uname -s)" == "NetBSD" ]]; then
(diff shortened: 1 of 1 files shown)


xuhdev:
Now there is confliction caused by my PR #5496 :(  uname should also be replaced by `OSTYPE`

mcornella:
Merged on 628d0bb10664453fe55c1e725ab89fa42dc4ab0a.

pfctgeorge:
this commit makes my ls no coloring in macOS Sierra.

drewhammond:
This broke ls on El Capitan 10.11.6

```
➜  ~ ls
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

jackie-qiu:
this commit makes ls no coloring in macOS Sierra

mcornella:
@pfctgeorge @drewhammond @jackie-qiu check out #5570 and please tell me there if it fixes it.

PR #5518, opened 2016-10-10
Replace symlinks with `source` for compatibility with MSYS2

mcornella:
Fixes #5509.


mcornella:
Closed for lack of interest.

PR #5529, opened 2016-10-13
Add autocompletion plugin for rustup subcommands

dwillie:
Added a new plugin for rustup which supports autocomplete across the top level commands and subcommands where they apply

### DIFF
--- a/plugins/rustup/_rustup
+++ b/plugins/rustup/_rustup
@@ -1,0 +1,128 @@
+#compdef rustup
+#autoload
+
+# Rustup zsh autocompletion
+
+local -a _commands
+_commands=(
+    'show: Show the active and installed toolchains'
+    'update:[NAME] Update Rust toolchains'
+    'default:<toolchain> Set the default toolchain'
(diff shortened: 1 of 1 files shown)


dgellow:
Any reason the PR hasn't been merged?

dwillie:
Not that I know of

robbyrussell:
Could we get a README file for this?

pkulev:
Oh, wow.
I've created separate plugin: https://github.com/pkulev/zsh-rustup-completion.
@robbyrussell, I can create PR here, but there is little quirk: rustup can generate autocomplete itself.
How can I integrate generation with standard oh-my-zsh update pipeline?

dwillie:
Sorry this sat here for so long - @pkulev's solution does sound superior to this, is there perhaps another example of a plugin using generated autocomplete?

mcornella:
The `npm` plugin uses cached completions, we could adapt that one.

ccqpein:
Is it still open?

mcornella:
ping

PR #5530, opened 2016-10-14
Fix last working directory plugin not working

sendyhalim:
This patch will fix last working directory plugin, 
I believe it's not working because of this 

```
[[ "$PWD" != "$HOME" ]] && return
```

I think we need to use equality operator to check if we're in home directory.

Tested manually

### DIFF
--- a/plugins/last-working-dir/last-working-dir.plugin.zsh
+++ b/plugins/last-working-dir/last-working-dir.plugin.zsh
@@ -16,8 +16,8 @@ lwd() {
 
 # Jump to last directory automatically unless:
 # - this isn't the first time the plugin is loaded
-# - it's not in $HOME directory
+# - it's in $HOME directory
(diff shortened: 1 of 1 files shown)


mcornella:
The logic right now says: if current directory is not `$HOME`, don't jump to last working directory. Which is the best heuristic we have to not jump when doing `Open Folder in Terminal`. See #5471.
With your change, this would mean that it would only jump to the last working directory if the current directory wasn't `$HOME`, so this is wrong.

What exactly isn't working in your setup?

sendyhalim:
Oh ok, I pull-ed yesterday and found out that it's jumping to last working dir in tmux. Steps to reproduce:

1. cd to dir A
2. Open tmux (we're now in dir A)
3. cd to dir B
4. split the pane, my expectation is that on the new pane, it would jump to dir B, instead it jumps to dir A, is it intended to behave like this?

PR #5536, opened 2016-10-14
Added Multiple "cd .." alias

sa02:
alias ..='cd ..  && ll'         # Go up 1 directory
alias ...='cd ../.. && ll'     # Go up 2 directories
alias ....='cd ../../.. && ll'     # Go up 3 directories


mcornella:
Already available [in core files of OMZ](https://github.com/robbyrussell/oh-my-zsh/blob/31a84e7/lib/directories.zsh#L6-L9).

sa02:
Nice. didn't know that

PR #5537, opened 2016-10-14
ssh-agent plugin: add check for .ssh folder

### LINKED ISSUE #5128: assumption that .ssh exist's
Hey good folks, 
awesome work with the project so far. I just encountered a slightly odd scenario.
If .ssh doesn't exists then the plugin https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/ssh-agent/ssh-agent.plugin.zsh results in giving out nasty error messages which can catch people off guard.

Can we add a check to ensure .ssh exists or if .ssh is created if it doesn't exists already.

iandrewt:
If `$HOME/.ssh` doesn't exist, echo `$HOME/.ssh not found, exiting..."
and return 1 instead of running the rest of the plugin.

This will fix #5128

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -1,6 +1,6 @@
 if [[ ! -d $HOME/.ssh ]]; then
-    echo $HOME/.ssh not found, exiting...
-    return 1
+    echo "$HOME/.ssh not found, creating now..."
+    mkdir ~/.ssh && chmod 700 ~/.ssh
 fi
 
 typeset _agent_forwarding _ssh_env_cache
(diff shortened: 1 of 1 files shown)


mcornella:
👍 but I'd prefer creating the directory instead.

```
mkdir ~/.ssh && chmod 600 ~/.ssh
```
should be enough.

iandrewt:
Should I still include a message to say this was done?

mcornella:
Yes please.

iandrewt:
Just noticed this PR still hasn't been merged, any updates?

nunojpg:
@iandrewt can you take a look at my PR #5585? Maybe if we manage to combine all the ssh-agent PRs, we can get @mcornella to take some time to look at them.

PR #5540, opened 2016-10-15
Detailed how to switch tabs

AOrobator:
Detailed how to switch tabs for OS X plugin

### DIFF
--- a/plugins/osx/README.md
+++ b/plugins/osx/README.md
@@ -15,22 +15,22 @@ Original author: [Sorin Ionescu](https://github.com/sorin-ionescu)
 
 ## Commands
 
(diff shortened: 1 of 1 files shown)


AOrobator:
Detailed how to switch tabs for OS X plugin

mcornella:
Wait how do you know which hotkeys work? I assume that depends on the terminal emulator you use...

AOrobator:
This works for iTerm2, doesn't work for the Terminal app because that app isn't supported. Haven't tried it out with other emulators, but iTerm2 has a lot of popularity.

mcornella:
Closing because this is Terminal-specific and can be found via other means.

PR #5545, opened 2016-10-17
OS X --> macOS

masoncodes:
It isn't OS X anymore, its macOS.


mcornella:
Can you check the rest of the project for more instances of OSX? A comment in the OSX plugin would be nice as well.

masoncodes:
sure

PR #5548, opened 2016-10-18
Modernize the OSX plugin

masoncodes:


apjanke:
The plugin filename and the name in the README documentation are now inconsistent ("osx" vs "macOS").

Might want to just leave the plugin name as is, despite Apple's branding change, because changing it would cause it to break silently for everyone who uses it (since OMZ does not produce warning messages when an undefined plugin is requested).

PR #5554, opened 2016-10-20
Add ability to show 24 hour format

mmilburn:
```
# Set ISO8601_HOURS to any string or value to get a 24-hour clock
# If ISO8601_HOURS is unset show 12-hour clock
# export before loading theme

export ISO8601_HOURS=1
ZSH_THEME="rkj-repos"
```


mmilburn:
Should have prefaced the commit message with "rkj-repos:"

mmilburn:
Please let me know if there is a better way to do this.

PR #5555, opened 2016-10-20
Add aliases for git remote

AlexVPopov:
Add aliases for:
- git remote prune
- git remote prune origin


mcornella:
You can achieve the same end result with `gfa` for `git fetch --all --prune`.

PR #5557, opened 2016-10-20
add docker-compose down alias

jotadrilo:


mcornella:
Renamed to `dcdn` for consistency with `dcup`. Thanks!

PR #5558, opened 2016-10-20
clarify on how to open ~/.zshrc file in README.md

landtanin:
I had to stckoverflow this issue out. So I think it would be better to specify the method here.


mcornella:
I've reworded your changes to make more obvious that you have to use your `favorite text editor`. Thanks for bringing up this issue.

PR #5562, opened 2016-10-21
[chruby] avoid calling brew --prefix 3 times

mikz:
once is enough, improves performance significantly

each call to brew is about 0.4 seconds on my machine

If you'd like to use this before merging, try https://github.com/robbyrussell/oh-my-zsh/pull/5562#issuecomment-321741502

### DIFF
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -30,7 +30,9 @@ _ruby-install-installed() {
 
 # Simple definition completer for ruby-build
 if _ruby-build_installed; then
-    _ruby-build() { compadd $(ruby-build --definitions) }
+    _ruby-build() {
+        compadd $(ruby-build --definitions)
+    }
     compdef _ruby-build ruby-build
 fi
 
@@ -41,21 +43,21 @@ _source_from_omz_settings() {
     zstyle -s :omz:plugins:chruby path _chruby_path
     zstyle -s :omz:plugins:chruby auto _chruby_auto
 
-    if [[ -r ${_chruby_path} ]]; then
-        source ${_chruby_path}
+    if [[ -r "${_chruby_path}" ]]; then
+        source "${_chruby_path}"
     fi
 
-    if [[ -r ${_chruby_auto} ]]; then
-        source ${_chruby_auto}
+    if [[ -r "${_chruby_auto}" ]]; then
+        source "${_chruby_auto}"
     fi
 }
 
 function () {
-    local _chruby_homebrew_prefix=$(brew --prefix chruby) &> /dev/null
+    local _chruby_homebrew_prefix="$(brew --prefix chruby) &> /dev/null"
 
-    if _homebrew-installed && [ -r ${_chruby_homebrew_prefix} ] ; then
-        source ${_chruby_homebrew_prefix}/share/chruby/chruby.sh
-        source ${_chruby_homebrew_prefix}/share/chruby/auto.sh
+    if _homebrew-installed && [ -r "${_chruby_homebrew_prefix}" ] ; then
+        source "${_chruby_homebrew_prefix}/share/chruby/chruby.sh"
+        source "${_chruby_homebrew_prefix}/share/chruby/auto.sh"
     elif [[ -r "/usr/local/share/chruby/chruby.sh" ]] ; then
         source /usr/local/share/chruby/chruby.sh
         source /usr/local/share/chruby/auto.sh
@@ -69,9 +71,9 @@ function ensure_chruby() {
 }
 
 function current_ruby() {
-    local _ruby="$(chruby |grep \* |tr -d '* ')"
-    if [[ $(chruby |grep -c \*) -eq 1 ]]; then
-        echo ${_ruby}
+    local _ruby="$(chruby | grep \* | tr -d '* ')"
+    if [[ -n "${_ruby}" ]]; then
+        echo "${_ruby}"
     else
         echo "system"
     fi
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -43,19 +43,19 @@ _source_from_omz_settings() {
     zstyle -s :omz:plugins:chruby path _chruby_path
     zstyle -s :omz:plugins:chruby auto _chruby_auto
 
-    if [[ -r "${_chruby_path}" ]]; then
-        source "${_chruby_path}"
+    if [[ -r "$_chruby_path" ]]; then
+        source "$_chruby_path"
     fi
 
-    if [[ -r "${_chruby_auto}" ]]; then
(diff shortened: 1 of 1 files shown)


plugins/chruby/chruby.plugin.zsh
```diff
@@ -20,8 +20,10 @@ _homebrew-installed() {
     whence brew &> /dev/null
 }
 
+local _homebrew_prefix_chruby=$(brew --prefix chruby) &> /dev/null
```
mcornella: This local won't work, better just `unset` the variable at the end.

mikz:
@mcornella fixed using anonymous function. Hope that is ok.
Also found another global variable, so changed it to local. And used one line definition and assign. Hope that is not against a style guide or something. Not sure why it was split into definition and assign.

mcornella:
> And used one line definition and assign. Hope that is not against a style guide or something. Not sure why it was split into definition and assign.

In certain ways that syntax fails, I don't remember exactly when, it might have something to do with arrays. Check that it works as well, I don't have `chruby` to test this, unfortunately.

The rest seems fine. The anonymous function shouldn't be a problem. Again, if you've tested it and checked that it works I'll merge it without a problem. Thanks!

mikz:
@mcornella I'm using it with homebrew installation and works just fine for me.

Just noticed some of the expansions are not properly quoted. Pushed that.

You were right about the array assignment, it is described in http://zshwiki.org/home/scripting/array.

However, chruby does that on its own so I deleted that whole function.
Also optimized `current_ruby` to run chruby just once.

I'm quite happy with it right now.

mikz:
@mcornella ping. I'm using this for quite some time without any issues.

plugins/chruby/chruby.plugin.zsh
```diff
@@ -45,46 +43,37 @@ _source_from_omz_settings() {
...
-_chruby_dirs() {
-    chrubydirs=($HOME/.rubies/ $PREFIX/opt/rubies)
-    for dir in chrubydirs; do
-        if [[ -d $dir ]]; then
-            RUBIES+=$dir
-        fi
-    done
-}
+function () {
+    local _chruby_homebrew_prefix="$(brew --prefix chruby) &> /dev/null"
```
mcornella: The `&> /dev/null` thing should go inside the parentheses.

plugins/chruby/chruby.plugin.zsh
```diff
@@ -45,46 +43,37 @@ _source_from_omz_settings() {
...
-    _chruby_dirs
-elif [[ -r "/usr/local/share/chruby/chruby.sh" ]] ; then
-    source /usr/local/share/chruby/chruby.sh
-    source /usr/local/share/chruby/auto.sh
-    _chruby_dirs
-else
-    _source_from_omz_settings
-    _chruby_dirs
-fi
+    if _homebrew-installed && [ -r "${_chruby_homebrew_prefix}" ] ; then
```
mcornella: Let's change all these parameter expressions to get rid of the curly braces, which is more idiomatic zsh. Braces should only be used if we have to append or prepend some other string. I'll make sure we have that written in our style guide.

mcornella: Also, use 2 brackets instead of 1. They can do more stuff and are backwards compatible, so by default we use double brackets. Same regarding the style guide.

plugins/chruby/chruby.plugin.zsh
```diff
@@ -45,46 +43,37 @@ _source_from_omz_settings() {
...
 function ensure_chruby() {
     $(whence chruby)
 }
 
 function current_ruby() {
-    local _ruby
-    _ruby="$(chruby |grep \* |tr -d '* ')"
-    if [[ $(chruby |grep -c \*) -eq 1 ]]; then
-        echo ${_ruby}
+    local _ruby="$(chruby | grep \* | tr -d '* ')"
```
mcornella: I think we can do this better. Let me work on a quick replacement.

mikz: Changed everything except this. Not really sure what to do here.

mikz:
@mcornella fixed! Not sure what to do about the last item though...

mikz:
@mcornella ping. There is just one last item, but I'm really not sure how to approach it.

chruby will print something like this:

```
   jruby-9.1.5.0
   ruby-2.3.0
 * ruby-2.3.1
```

and there are no options for different format other than creating a function that reads from `RUBY_ROOT`, verifies it is in `RUBIES` array and takes just basename.

I'd say now it is faster than it was before and using public chruby api.

mikz:
@mcornella I'd really like to merge this. Looks like other PRs are trying to fix it too https://github.com/robbyrussell/oh-my-zsh/pull/6080.

tilgovi:
On my machine, without customizing homebrew, `/usr/local/share/chruby` is where homebrew installed chruby for me. We could make this even fasting by only asking homebrew if we don't find it there, no? Why call `brew prefix` at all if we have a good starting guess?

mikz:
@tilgovi people can install homebrew to different prefix like `/opt` and this would break it for them. 

I imagine at some point all plugins could use some shared variable which would define the prefix and would be executed just once.

But there might be just very few plugins that need this, so might not be needed.

I'm using this for almost a year now without problems.

mikz:
If you'd like to try to use this before merging, then:

```shell
mkdir -p ~/.oh-my-zsh/custom/plugins/chruby/
curl https://raw.githubusercontent.com/mikz/oh-my-zsh/8f91babe95de5582eccb620014999e6da2075f90/plugins/chruby/chruby.plugin.zsh -o ~/.oh-my-zsh/custom/plugins/chruby/chruby.plugin.zsh
```

tilgovi:
I think maybe you misunderstand me.

The purpose of this PR is to make it faster.

*If* chruby exists at /usr/local/share why not use that before calling brew prefix?

The only way this would break would be if someone had it installed in /usr/local/share but for some reason didn't want to use that one, and wanted to use another copy of chruby installed in a custom brew prefix.

But why spend 0.3 seconds asking brew when we have a good first guess?

tilgovi:
And it's very strange to me that the sourcing from the settings is the last choice.

To me, it should be:
- explicit user setting
- default location
- ask homebrew

Brew should be the last resort because it's platform specific and only needed for users who do not use the default location.

mikz:
Sure, that is doable. Just change the order in the anonymous function that loads everything. Feel free to send a patch or open a PR to my branch.

tilgovi:
@mikz attempted here: https://github.com/mikz/oh-my-zsh/pull/1

tilgovi:
@mcornella anything else we can do here to make this mergeable?

mikz:
Looks like this was partially fixed by https://github.com/robbyrussell/oh-my-zsh/pull/6080. Closing.

PR #5564, opened 2016-10-23
Don't use sudo command on nowrite zypper commands

artur-shaik:
It is quite annoying to write root password when are doing simple read
request on zypper package manager. Like list repo, or print package
information.


artur-shaik:
@Atem18, @patrickelectric,  @r-darwish , hi, can you give a review to this PR?

Atem18:
@artur-shaik I don't have a SUSE to test but if you tested it, I have no problem.

artur-shaik:
Actually, I use it for long time without problems. Thank you for response.

PR #5565, opened 2016-10-23
Add a tip to ease agnoster first setup

GrmpCerber:


mcornella:
I've changed the link to point to [the agnoster section of the wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Themes#agnoster). Thanks!

PR #5566, opened 2016-10-23
Add alias for virtualenv activation

kronion:


grimm26 (changes requested):
Please add documentation for the alias you want to add.

carlosala:
@kronion, sorry for the late response (more than 6 years!). This alias is a bit "risky", as it's sourcing some file named `activate` without having control of what's doing. For example, some program might inject malicious code in that `activate` file and we'd be sourcing that...
It looks like it's better for a user-space alias, having that as default wouldn't be a great idea IMHO.
Thanks for your contribution!

Issue #5520, opened 2016-10-11
folder and files colour issue


MrKiven:
Hi:

I upgrade oh-my-zsh today, after update, folder and file do not display colour. It's ok before update.

MrKiven:
I've checkout to https://github.com/robbyrussell/oh-my-zsh/commit/d69f2850afc189310b40141c839480b42f71775c, is ok now.

MrKiven:
@mcornella

s7evinz:
I'm having the same issue. ls doesn't have colour highlight for directories, symlinks, executables etc.

MrKiven:
@s7evinz yeah, I've checkout to https://github.com/robbyrussell/oh-my-zsh/commit/d69f2850afc189310b40141c839480b42f71775c this commit, it's ok

jannylund:
Looks like the problem is that there is no test for `gls` before testing `$OSTYPE`. https://github.com/robbyrussell/oh-my-zsh/blob/628d0bb10664453fe55c1e725ab89fa42dc4ab0a/lib/theme-and-appearance.zsh#L9 makes os x skip the `else`  settings for colors if `gls` is missing.

mcornella:
Is it `ls -G` the one that works?

MrKiven:
`ls -G` works.

mcornella:
To update run `upgrade_oh_my_zsh`. Apologies for the mistake!

PR #5570, opened 2016-10-25
Check dircolors settings before using gls on darwin

### LINKED ISSUE #5552: Color ls broken after update
I just updated to the latest oh-my-zsh (commit f94d4e6a4019ed7fcc5470655b1154b3d1f67913) and color ls is now broken on my Mac running El Capitan.  Two recent log messages indicate fixes were checked in for just that (fixing 5520 and 5516), but it seems that has broken things for me:

```
% ls
/bin/ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
% which ls
ls: aliased to gls --color=tty
% which gls
gls () {
    /bin/ls -FC -lh $@
}
% /bin/ls
foo
% /bin/ls -FC -lh
total 0
-rw-r--r--  1 i831533  staff     0B Oct 19 12:06 foo
% gls --color=tty
/bin/ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
% /bin/ls -FC -lh --color=tty
/bin/ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
% /bin/ls -FC -lh -G
total 0
-rw-r--r--  1 i831533  staff     0B Oct 19 12:06 foo
```

It seems that the `--color=tty` option for Linux is being incorrectly passed to the Mac version of ls, which doesn't support it (instead supporting -G).

This all was working fine before this morning's update to oh-my-zsh.

mcornella:
`gls` seems to be installed by default or on most macOS systems, but its default color scheme sucks. This fix will make sure to only use it if it has been customised prior to running OMZ.

Fixes #5552

Related: #5516, #5520.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -4,6 +4,8 @@ autoload -U colors && colors
 # Enable ls colors
 export LSCOLORS="Gxfxcxdxbxegedabagacad"
 
+# TODO organise this chaotic logic
+
 if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
   # Find the option for using colors in ls, depending on the version
   if [[ "$OSTYPE" == netbsd* ]]; then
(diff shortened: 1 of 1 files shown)


tete1030:
Works good for me. The `ls` command comes back to original `ls -G`. I don't have dircolors settings and don't know how to write them so I didn't tested this commit with them. Thanks.

rossmcf:
Thanks for tidying up my mess here. I was clearly out of my depth with that one!

PR #5571, opened 2016-10-25
Add cwhois

frankiexyz:
Add new function cwhois for IP/ASN lookup 

`whois -h whois.cymru.com "$@"`


mcornella:
Has no interest.

PR #5584, opened 2016-10-28
Extend list of adb commands.

johnjohndoe:


johnjohndoe:
@robbyrussell Rebased onto current master.

johnjohndoe:
@robbyrussell Rebased on top of current master.

PR #5585, opened 2016-10-28
ssh-agent: improvements

nunojpg:
Comments are welcomed. In particular for 874f5af this should be considered to be provided directly by the plugin load framework.
I mean, it could be a option to mark plugins as non reentrant and oh-my-zsh would implement a mutex for those.

Btw, I found out about zsh and oh-my-zsh a few hours ago, so pardon my (coding) french if something is very off track.


nunojpg:
@mcornella Please take a look. Thanks!

PR #5593, opened 2016-11-01
wd: Make warp point name optional

Cretezy:


mcornella:
Please propose this change to the [official repository](https://github.com/mfaerevaag/wd).

Cretezy:
Thank you!

Issue #5007, opened 2016-04-15
Add plugin expand-aliases


jarmo:
I've created a plugin **expand-aliases**, which does what it says :) More information in the [README](https://github.com/jarmo/expand-aliases-oh-my-zsh/blob/master/README.md).

Since I'm not sure how should I do a pull request for a plugin, then I've made a separate repo here:
https://github.com/jarmo/expand-aliases-oh-my-zsh

sontags:
Yes Please!

mcornella:
Closing as per https://github.com/robbyrussell/oh-my-zsh/pull/5597#issuecomment-257819411.

PR #5597, opened 2016-11-01
added Jarmo Pretman's expand-aliases plugin

sontags:
As requested by the author Jarmo Pertman's in issue #5007 this PR includes his plugin expand_aliases into oh-my-zsh.

The LICENSE file from the original repo (https://github.com/jarmo/expand-aliases-oh-my-zsh) was removed due to redundancy (original repo and oh-my-zsh are both MIT).

All credits to @jarmo

### DIFF
--- a/plugins/expand_aliases/README.md
+++ b/plugins/expand_aliases/README.md
@@ -1,0 +1,25 @@
+# expand-aliases
+
+**Maintainer:** [Jarmo Pertman](https://github.com/jarmo)
+
+This plugin expands aliases while typing them in terminal.
+
+For example:
+```zsh
+alias foo="bar baz"
+
+$ foo
+# pressing space will expand that alias to... surprise, surprise
+$ bar baz
+```
+
+To **skip** expansion, use `control-space` instead of just `space`.
+
+## Installation
+
+Just add it to `oh-my-zsh` plugins list in `~/.zshrc`, but make sure it is added last (or not
+last if you prefer some aliases not to be expanded).
+
+`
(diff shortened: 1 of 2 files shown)


sontags:
@jarmo There you go... Sorry again...

jarmo:
I'm really thankful that you're making this pull request, but this also has a typo in it... but do no fret and let it be, because it's still correct in the README :)

Thanks again!

mcornella:
Hi all, thanks for this pull request. There is already a plugin that does this job though: [globalias plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/globalias).

That said, a few comments on the plugin:

1. The name has to be the same on both the directory and the `plugin.zsh` file. That means it should be `expand-aliases` on both.

2. zsh sets a `$aliases` associative array that stores all currently defined aliases like so:
  ```zsh
$ alias g=git
$ echo $aliases[g]
git
```

Thanks again.

jarmo:
@mcornella - that's cool, but that plugin didn't expand aliases at the time #5007 was created. Since it does now that, then my plugin is definitely not needed.

Thanks again!

mcornella:
Fair enough. These slip by very easily though.

Feel free to add your plugin information in the [external plugins](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins) wiki section.

jarmo:
Thank you for suggesting that, but there's no point now that globalias also does it. I will deprecate mine instead ;)

PR #5603, opened 2016-11-02
Changed dsa --> rsa in zshrc.zsh-template file

TimelyToga:
# Reasoning
OpenSSH 7.0 [disabled DSA keys by default](https://www.gentoo.org/support/news-items/2015-08-13-openssh-weak-keys.html) for a wide variety of reasons that are discussed [here](http://security.stackexchange.com/a/112818). This PR is merely updating the template for `.zshrc` file to use the now-standard `~/.ssh/rsa_id`.


mcornella:
Thanks!

PR #5605, opened 2016-11-03
Fix #5604: No DIR ( directory ) colors

### LINKED ISSUE #5604: No DIR ( directory )  colors => LS LL
I just install ubuntu 16.10
then install ZSH following this link https://gist.github.com/tsabat/1498393

commands LS, LL, etc 

screenshot from 2016-11-02 16-24-22

hmgle:
Fixes #5604


mcornella:
Thanks!

PR #5608, opened 2016-11-03
add docker aliases to docker plugin

fieteboerner:
use dk as base

it is tedious to write `docker` and its commands and parameters out while tinkering with docker containers.
These aliases are simplifying my workflows tremendously. Perhaps it is also useful for other people. Anybody is welcome to add more aliases so it fit also to own workflows and commands.


HaraldNordgren:
Looks good to me!

mcornella:
Reopen if you're still interested in these changes and I'll mix it up with #6527 in a common PR.

PR #5612, opened 2016-11-06
Docker-machine aliases

HaraldNordgren:
Useful aliases for `docker-machine`. Inspired by docker-compose/docker-compose.plugin.zsh.

### DIFF
--- a/plugins/docker-machine/docker-machine.plugin.zsh
+++ b/plugins/docker-machine/docker-machine.plugin.zsh
@@ -1,0 +1,19 @@
+# Authors:
+# https://github.com/HaraldNordgren
+#
+# Docker-machine related zsh aliases
+
+alias dm='docker-machine'
+
+alias dmc='docker-machine create'
+alias dmls='docker-machine ls'
+alias dmrestart='docker-machine restart'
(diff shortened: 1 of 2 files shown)


plugins/docker-machine/docker-machine.plugin.zsh
```diff
@@ -0,0 +1,20 @@
...
+alias dm='docker-machine'
+
+alias dmc='docker-machine create'
+alias dmls='docker-machine ls'
+alias dmrestart='docker-machine restart'
+alias dmrm='docker-machine rm'
+alias dmstart='docker-machine start'
+alias dmstop='docker-machine stop'
+alias dmst='docker-machine status'
+alias dmnfs='docker-machine-nfs'
```
twang2218: I like most aliases, I used very similar ones myself, however, `docker-machine-nfs` is not official docker-machine command, I don't think this one should be included.

HaraldNordgren: Updated.

HaraldNordgren:
Could someone merge or close this one?

PR #5613, opened 2016-11-06
[plugin][mercurial] Optimize in_hg and hg_get_branch_name

MightyCreak:
When not in the repo root dir `in_hg` used `hg summary` which is terribly slow on big repositories, and `hg_get_branch_name` called `hg branch` which is also pretty slow as well.

Based on the [branch](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/branch) plugin, I created the function `hg_get_dir` that fetch the `.hg` directory if it exists, going up in the directory tree. From there, we know if we're in an HG repository if `in_hg` returns a non empty string, and we can simply get the branch name from the content of the `.hg/branch` file.

### Benchmarks

Based on 10 calls to `hg_get_branch_name` in subdir (so that the old `hg summary` is called), here are the average times:
* Before: `/tmp/test.zsh  0.08s user 0.04s system 96% cpu 0.133 total`
* After: `/tmp/test.zsh  0.00s user 0.00s system 71% cpu 0.008 total`

It's faster and it uses less CPU!

The other changes after that are purely esthetic (fit in 80 columns and indentation fix).

### DIFF
--- a/plugins/mercurial/mercurial.plugin.zsh
+++ b/plugins/mercurial/mercurial.plugin.zsh
@@ -18,6 +18,12 @@ alias hgca='hg commit --amend'
 # list unresolved files (since hg does not list unmerged files in the status command)
 alias hgun='hg resolve --list'
 
+# Configuration
+
+# Comment this line if you don't want to show if the branch is dirty or not.
+# This can save a lot of time of very large repositories.
+SHOW_DIRTY_BRANCH=1
+
 function hg_get_dir() {
   # Defines path as current directory
   local current_dir=$PWD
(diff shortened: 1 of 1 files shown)


MightyCreak:
@ptrv: This PR might interest you!

plugins/mercurial/mercurial.plugin.zsh
```diff
@@ -18,25 +18,44 @@ alias hgca='hg commit --amend'
...
 
 function hg_prompt_info {
-  if [ $(in_hg) ]; then
-    _DISPLAY=$(hg_get_branch_name)
+  local branch=$(hg_get_branch_name)
+  if [[ $branch != "" ]]; then
     echo "$ZSH_PROMPT_BASE_COLOR$ZSH_THEME_HG_PROMPT_PREFIX\
-$ZSH_THEME_REPO_NAME_COLOR$_DISPLAY$ZSH_PROMPT_BASE_COLOR$ZSH_PROMPT_BASE_COLOR$(hg_dirty)$ZSH_THEME_HG_PROMPT_SUFFIX$ZSH_PROMPT_BASE_COLOR"
-    unset _DISPLAY
-  fi
```
berendkleinhaneveld: `fi` should not be removed here

plugins/mercurial/mercurial.plugin.zsh
```diff
@@ -18,25 +18,44 @@ alias hgca='hg commit --amend'
...
+$ZSH_THEME_REPO_NAME_COLOR${display}$ZSH_PROMPT_BASE_COLOR\
```
berendkleinhaneveld: This was probably meant to be `branch`?

MightyCreak:
You are right for both, I don't even have that in my local file, I don't know how these errors got there... noob mistake I guess ;)

I'll upload a new PR

MightyCreak:
Ok, there's been too much changes since 2016.
Let me do another proper PR.

MightyCreak:
Closed in favor of the new PR #6631

PR #5624, opened 2016-11-10
Fixed minimal.zsh-theme's check for in_hg.

Telling:
I experienced an issue with the minimal theme [a while ago](https://github.com/robbyrussell/oh-my-zsh/pull/3821). Turns out I've had a local version of the file checked out since.

The function `in_svn` uses exit codes, and the function `in_hg` echo's 1 if successful. I think the best solution would be to make `in_hg` behave the same way as `in_svn` - But I'm afraid such drastic measures will break too much.

This is only tested on my macOS 10.11.6 box.

### DIFF
--- a/themes/minimal.zsh-theme
+++ b/themes/minimal.zsh-theme
@@ -14,7 +14,7 @@ ZSH_THEME_HG_PROMPT_CLEAN=$ZSH_THEME_GIT_PROMPT_CLEAN
 vcs_status() {
     if [[ $(whence in_svn) != "" ]] && in_svn; then
         svn_prompt_info
-    elif [[ $(whence in_hg) != "" ]] && in_hg; then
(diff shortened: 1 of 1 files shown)


Telling:
I have a version where I've fixed the issue in the mercurial plugin, such that it behaves similar to the svn plugin (at least the `in_*` function). It seems that the only theme using the mercurial plugin in the oh-my-zsh repository is the minimal theme, i.e. quite easily fixed locally. However any external theme using the `in_hg` function will not work.

Should I create a new PR, or is the solution in this PR preferred? Thanks.

Telling:
Had to fix this locally once again.

mcornella:
Fixed in 4119f53004de987215e4ae2ce49d0f818ad58c63.

PR #5626, opened 2016-11-10
Corrected key/value split when value contains a `:`.

gbooker:
When the path in a wd contains a `:`, the wd plugin does not read the path past the `:` character.  This change makes the key all of the string up to the first colon (as before) but the value as all of the string after the first colon.


mcornella:
Corrected in the latest version.

PR #5629, opened 2016-11-11
Default installer to clone git origin master

DwordPtr:
#5628 I wrote this because many zsh users (like myself) have forks of oh-my-zsh that we use instead of the official version. This way someone with a fork can clone the repo they own and install their own version of oh-my-zsh.

Please let me know if you would like me to change anything as I am a total neophyte a contributing to open source.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -62,7 +62,7 @@ main() {
     fi
   fi
   # if running from the git repo use git's origin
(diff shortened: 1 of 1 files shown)


DwordPtr:
I'll probably add the space in on lunch.

DwordPtr:
@sgen This is done.

mcornella:
I don't understand, if `.git` directory exists, there's no need to clone because it has already been cloned, no?

PR #5633, opened 2016-11-13
Add `serve` completion plugin

SuperPaintman:
Hi! I added completion plugin for [serve](https://github.com/SuperPaintman/serve) HTTP static server


mcornella:
Not used enough. But good luck, it looks nice!

PR #5634, opened 2016-11-14
Make the upgrade script rebase from upstream remote if it exists

DwordPtr:
This should be helpful for people on a fork that want a happy path to upgrade.

It definitely comes with a complexity cost, but I totally think it's worth it.


mcornella:
Old and more difficult to implement than what it is.

PR #5635, opened 2016-11-14
[agnoster] add a short path option for the theme

theckman:
This change to the `agnoster` theme adds an option for using the name of
the current directory instead of a relative path. Assume you're in the
following directory: `~/.oh-my-zsh/custom/themes`. If you set
`AGNOSTER_SHORT_PATH=1` in your `.zshrc` file, the path will render only
as `themes` on the command line instead of `~/.oh-my-zsh/custom/themes`.

In the implementation of this option, another was added to give you even
more control if you want it (`AGNOSTER_CURRENT_PATH`). If
`AGNOSTER_SHORT_PATH` is unset and `AGNOSTER_CURRENT_PATH` is set, that
value will be used for the path value on the prompt.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -164,7 +164,11 @@ prompt_hg() {
 
 # Dir: current working directory
 prompt_dir() {
-  prompt_segment blue black '%~'
+  # if AGNOSTER_SHORT_PATH is set use %1~ for the path value
+  # that will turn '~/.oh-my-zsh/custom/themes' in to 'themes'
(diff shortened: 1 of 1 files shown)


theckman:
@mcornella @robbyrussell I see that you both have merged in the most recent changes to `agnoster` in the past. Might you have any interest in this change?

robbyrussell:
Sorry we never got back to you on this — nine years of silence is on us, not you.

Since we didn't get to it back then and nobody has raised it since, we're going to assume it hasn't been a deal-breaker for the community. Closing as part of some overdue housekeeping on the backlog. Thanks for taking the time to put it together.

PR #5638, opened 2016-11-14
Don't blindly overwrite history variables

dasJ:
OMZ should respect HISTSIZE and SAVEHIST if they are set before OMZ is run.


mcornella:
This is a duplicate of #3359 which was merged at some point, and later reverted in 1978a0923c864cacdc520adff83ed63f62c91073 because it broke people's workflow.

TL;DR you need to set them after sourcing OMZ so that you override our default settings.

PR #5639, opened 2016-11-14
autoupdate custom plugins

benwaffle:


slavaGanzin:
:+1: But what is $ZSH/custom?

benwaffle:
`$ZSH/custom` is where other plugins go, e.g. [zsh-syntax-highlighting](https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/INSTALL.md#oh-my-zsh)

tools/upgrade.sh
```diff
@@ -37,3 +37,11 @@ then
 else
   printf "${RED}%s${NORMAL}\n" 'There was an error updating. Try again later?'
 fi
+
+printf "${BLUE}%s${NORMAL}\n" "Updating Custom Plugins"
+cd "$ZSH/custom/plugins"
```
slavaGanzin: Let's use $ZSH_CUSTOM/plugins.

tools/upgrade.sh
```diff
@@ -37,3 +37,11 @@ then
...
+for dir in ./*; do
```
slavaGanzin: Let's substitute ```./*``` with ```$ZSH_CUSTOM/plugins/*```
And so ```cd``` would be obsolete

benwaffle: upgrade.sh does not have access to `$ZSH_CUSTOM`
It's run like this:
```zsh
function upgrade_oh_my_zsh() {
  env ZSH=$ZSH sh $ZSH/tools/upgrade.sh
}
```

benwaffle:
This PR now uses `$ZSH_CUSTOM` for the custom plugins dir

benwaffle:
@slavaGanzin do you have write access to the repo? or do we need @robbyrussell ?

What else is needed to merge this PR?

slavaGanzin:
@benwaffle Ben, I have no access. So this should be merged by maintainers

benwaffle:
superseded by #7539

PR #5641, opened 2016-11-15
Magento 2 plugin INIT

denisristic:
This plugin adds completion and aliases for the `magento` command. More information at https://github.com/denisristic/oh-my-zsh.


mcornella:
Has no interest.

PR #5643, opened 2016-11-16
feat: no CI on WIP commits

JaKXz:
I've been working with my team using ohmyzsh for a couple months now and it's going great! 

One problem is that since we pair program a lot, we find ourselves `gwip`ing and `gunwip`ing frequently to share WIP and switch drivers etc. This leads to a lot of CI resources being used unnecessarily. Since `[skip ci]` is used by the likes of TravisCI, CircleCI, GitLab CI and others I think it's fair to include here. Feedback is welcome! And if not mergeable, we'll probably just put this in our aliases ourselves, but I thought others could benefit. 

Thanks!


mcornella:
Great piece of information, thanks!

PR #5647, opened 2016-11-18
Fix: error messages while getting status of new (no commits) repository on non-english locales

psyriccio:
On non-english locales output of gitstatus are corrupted
before

Simple fix
after


mcornella:
Should be fixed with the call using `LANG=C`.

PR #5648, opened 2016-11-18
added alias for simple git stash command

ykl7:


mcornella:
There is already `gsta` for `git stash save` which is the default for `git stash` without arguments. The rest of possible, common `git stash` commands already have an alias as well.

PR #5650, opened 2016-11-18
'robbyrussell' theme file updated with changed PROMPT variable including the active virtualenv base name

darkdefender27:
To display the active python virtualenv base name inside the 'prompt' for theme='robbyrussell', added a script and injected it in the existing PROMPT variable.


mcornella:
This should be put into a custom theme, see [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes).

PR #5654, opened 2016-11-22
Adding new path to look for activate.sh

italomaia:
If autoenv was installed with pip and modifier --user, activate.sh will be at .local/bin


italomaia:
This is an easy PR. Are there any problems with it?

mcornella:
Thanks for bringing this to my attention!

PR #5655, opened 2016-11-22
Feature: mtab for osx plugin

ThisIsMissEm:
Hi!

This is a change I've ended up making to the osx plugin, and it's a source of near continuous merge conflict when I update oh-my-zsh. I find it super helpful to be able to open up multiple tabs in a give directory, so maybe others would find this useful too!

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -12,7 +12,7 @@ EOF
   echo "$the_app"
 }
 
-function tab() {
+function _tab() {
   # Must not have trailing semicolon, for iTerm compatibility
   local command="cd \\\"$PWD\\\"; clear"
   (( $# > 0 )) && command="${command}; $*"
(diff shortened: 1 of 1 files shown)


mcornella:
Hey, sorry for not seeing this. As I see it, this has two parts:
1. Changing the tab function to allow specifying a directory
2. Allowing to specify the number of tabs to open.

Number 2 can be achieved with the `repeat` builtin, like so:

```zsh
repeat N tab <directory>
```

Number 1 can't be achieved without breaking backward compatibility, which your solution does (all arguments were used as commands, and your `_tab` function only receives the directory argument). A possible patch is to detect whether `$1` is a directory, and then use that instead of `$PWD`:

```zsh
function tab() {
  local dir="$PWD"
  if [[ -d "$1" ]]; then
    dir="$1"
    shift
  fi

  # Must not have trailing semicolon, for iTerm compatibility
  local command="cd \\\"$dir\\\"; clear"
  (( $# > 0 )) && command="${command}; $*"
```

ThisIsMissEm:
Hey @mcornella I updated to the latest version of oh-my-zsh last night and noticed the ZSH_CUSTOM flag, so I'm now using that to run my tab command — I never have needed to run a given command in a new tab, instead opting to always open, then run; hence not encountering the backwards incompatibility.

I quite like the shortness of `mtab <N> <dir?>`, but I guess that could maybe be done with an alias or something.

Up to you what you do with it, now oh-my-zsh supports custom plugins, I don't have any more merge conflicts! I don't even have a fork anymore!

PR #5659, opened 2016-11-24
Add gsize command for git plugin

letientai299:
In case that you have not know about it, following are sample output for the new `gsize` alias: 
```sh
$ git count-objects -vH
count: 1414
size: 5.92 MiB
in-pack: 180
packs: 2
size-pack: 46.68 KiB
prune-packable: 0
garbage: 0
size-garbage: 0 bytes
```

@Stibbons, @ncanceill, please review.


mcornella:
This is more of a curiosity than a day-to-day feature. You can keep this alias in your zshrc file instead. Thanks for your contribution!

PR #5666, opened 2016-11-28
Added Themes

LinuxGogley:
I just used the theme for mortalscumbag and modified it to create two separate themes for user and root.  The color scheme is modified, and the hostname no longer is associated with the user to make the appearance clearer and simpler.

Thanks!
Liam


mcornella:

PR #5670, opened 2016-11-28
Add plugin for Swift Package Manager

thomaspaulmann:
This plugin provides a few utilities that make you faster on your daily work with the [Swift Package Manager](https://github.com/apple/swift-package-manager).

I will provide ZSH Completions for the Package Manager in a separate PR.


mcornella:
Great example of a good PR and a well-documented plugin! One last thing though: what do you think about calling it `swiftpm` or `swift-pm`? I've seen [SwiftPM being used a lot (more frequently than SPM)](https://github.com/apple/swift-package-manager/search?utf8=%E2%9C%93&q=%22swift+pm%22&type=Issues) and `swift-package-manger` is a bit of a mouthful :wink:

Whatever choice you go with, make sure you update both the README info and the name of the `plugin.zsh` file. Thanks!

thomaspaulmann:
@mcornella Thanks for the kind words and the fast response! I've renamed the plugin to `swift-pm` and it should be ready for merge ;-)

ddunbar:
FWIW, we (on the SwiftPM team) have tried to use "swiftpm" when we need an abbreviation for it.

mcornella:
Good to know thanks! 😊

thomaspaulmann:
@ddunbar Thanks for pointing out! To match your internal abbreviation, I will change the plugin name.

@mcornella Is that fine for you?

mcornella:
@thomaspaulmann yes it's fine. It's soon enough that we don't have to make it backwards compatible.

PR #5671, opened 2016-11-29
added alias for starting react native in iTerm

tristola:
A shortcut to be able to start bundler separately from run-ios or run-android


mcornella:
Going with #5711 which is more complete. Thank you though!

PR #5673, opened 2016-11-29
mix run takes file arguments

### LINKED ISSUE #5672: mix plugin prevents file completion with tab
Using the mix plugin prevents you from using tab completion to provide a file or a directory as an argument. `mix run`, for example, takes a path to an exs file to execute. If you type `mix run pr/rep/<tab>` it should complete to the priv/repo directory and allow you to continue typing a file name.

geofflane:
Fixes #5672

Allow file completion when using the `mix run` command

### DIFF
--- a/plugins/mix/_mix
+++ b/plugins/mix/_mix
@@ -86,6 +86,9 @@ case $state in
       (test)
          _files
          ;;
+      (run)
+         _files
+         ;;
     esac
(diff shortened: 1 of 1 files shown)


mcornella:
:+1:

Source: https://hexdocs.pm/mix/Mix.Tasks.Run.html

PR #5680, opened 2016-12-01
Removes appeal to purchase products

influenza:
It is tasteless and unnecessary to suggest to the user that they purchase "swag" when an upgrade occurs.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -33,7 +33,6 @@ then
   printf '%s\n' '                        /____/                       '
   printf "${BLUE}%s\n" "Hooray! Oh My Zsh has been updated and/or is at the current version."
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on twitter: https://twitter.com/ohmyzsh"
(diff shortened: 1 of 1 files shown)


robbyrussell:
Hi there --

I'd be open to a commit that toggles this on/off via the .zshrc configuration... with it displaying by default. We get about ~20 orders a month from people wanting stickers/shirts/mugs and I attribute much of that to having this in there. This helps pay for a p/t office admin to have a few extra hours each month to handle fulfillment and even more important to me that seeing people post photos of stickers/shirts on social media fills me with a lot of pride about what this community has built. 

We can disagree on taste. I'm okay with that. I hope that you are, too.

Optionally, you're more than welcome to modify this in your own fork of the project but I'm going to keep this in my branch.

mcornella:
I agree. Oh My Zsh doesn't make money from any other source of revenue, but definitely we don't make a profit. So a single line when you upgrade reminding you that you can support us with merch doesn't seem that big of a deal.

I can definitely see that you think that line is not serious, I would be open to change the wording but I'd like to keep it short and sweet. Thanks for understanding!

influenza:
Our disagreement on taste is acknowledged. 

Regards

PR #5681, opened 2016-12-01
Update README.md

GreenJoey:
Change in line 23.
Linux is not a UNIX distribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To learn more, visit [ohmyz.sh](http://ohmyz.sh) and follow [@ohmyzsh](https://t
 
 __Disclaimer:__ _Oh My Zsh works best on macOS and Linux._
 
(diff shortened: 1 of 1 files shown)


mcornella:
#wontfix, here's why: it says [Unix-based, meaning it has similar characteristics](http://unix.stackexchange.com/a/4097).

Linus Torvalds based his first Linux prototype on Minix which was a UNIX clone, [as you can read on his first emails describing Linux](https://www.cs.cmu.edu/~awb/linux.history.html). 

So I won't accept the current PR as it is because it's factually correct. I will accept though a change from 'Unix-based' to 'Unix-like' because neither macOS (which is based on BSD) nor Linux used exact Unix code, just its architecture and core philosophy.

mcornella:
Thanks for making me think about this, BTW.

GreenJoey:
No problem..

PR #5690, opened 2016-12-04
Add gpsuo git alias

jvz:
This alias is for pushing a new git branch to the origin remote. I find myself having to do this a lot in a pull request workflow, and it's easier to use this alias than doing a `gp` followed by a `fuck` to fix it.


mcornella:
`gpsup` does the same https://github.com/robbyrussell/oh-my-zsh/blob/3d2542f41b8de36877b419f6e74e954d4db06a97/plugins/git/git.plugin.zsh#L164

Issue #1675, opened 2013-03-28
Please help! Cannot uninstall zsh.


matheuscandido:
I installed zsh on my mac but I want to get back to bash. None of the usual commands work on it, sudo, ls, cd, nothing. When i try to uninstall zsh with the uninstall_oh_my_zsh I get the following message:
   Removing ~/.oh-my-zsh
   /Users/Matheus/.oh-my-zsh/tools/uninstall.sh: line 4: rm: command not found
   Looking for original zsh config...
   Switching back to bash
   /Users/Matheus/.oh-my-zsh/tools/uninstall.sh: line 24: chsh: command not found
   Thanks for trying out Oh My Zsh. It's been uninstalled.
But it doesn`t unisntall, when I restart Terminal there`s zsh. I tried to run the command by its complete path, with the command:
   Users/Matheus/.oh-my-zsh/tools/uninstall.sh
But I get a "zsh: permission denied". I also can`t run it with sudo, because sudo doesn`t work.
Please guys help me restore my old and good bash.

rbaindourov:
Removed zsh with apt-get.
Ran `apt-get --reinstall install bash`
Modified `/etc/passwd`
Ran `/usr/bin/chsh -s /bin/bash`

Help, how to get rid of this?

```
robert@superhost ~ $ sudo -s
[sudo] password for robert: 
sudo: /usr/bin/zsh: command not found
robert@superhost ~ $ 
```

mcornella:
You need to logout of your current user session and log back in again.

Mihara17:
Go into terminal 
Click on help
It explains everything

DominicTobias:
Restoring bash works thanks, however installing zsh with brew and then uninstalling doesn't seem to actually remove zsh i.e. it's still at bin/zsh (and don't know what other residual files are around)

mcornella:
`/bin/zsh` is provided by macOS according to [this StackExchange comment](https://apple.stackexchange.com/a/100476).

DominicTobias:
Ah makes sense thanks, also I read earlier it will be the default next OS version

bengrunfeld:
I am unable to uninstall Zsh. I don't mean "switch from zsh to bash". I mean get rid of it so that it no longer exists on my computer. Is there any way to do that?

**System:** macOS Sierra 10.12.6
**Version:** zsh 5.2 (x86_64-apple-darwin16.0)

Steps I tried:

```
[...]

PR #5693, opened 2016-12-06
Explain how to uninstall more fully

sshadmand:
Based on solution that saved me: https://github.com/robbyrussell/oh-my-zsh/issues/1675


mcornella:
This doesn't necessarily work for everybody.

PR #5694, opened 2016-12-07
Increase HISTSIZE and SAVEHIST by a factor of 10

HaraldNordgren:
My $HOME/.zsh_history file is full and around 6 Kb. So increasing that size to a maximum of 60 Kb shouldn't hurt.

Increasing it by a factor of 100 might be even better since the file would still be less than 1 Mb. Although that might make starting a new shell slow -- I'm not sure if there is a relation.

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -3,8 +3,8 @@ if [ -z "$HISTFILE" ]; then
     HISTFILE=$HOME/.zsh_history
 fi
 
-HISTSIZE=10000
-SAVEHIST=10000
+HISTSIZE=100000
+SAVEHIST=100000
 
 # Show history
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
@robbyrussell Merge this?

mcornella:
You can have these set up in your zshrc file. The defaults we ship in OMZ are sufficient for most people.

mcornella:
BTW, the numbers in HISTIZE and SAVEHIST are not file size, they are the number of commands saved. So increasing by 10 might mean more than going from 6kb to 60kb because the lines saved per command can be very long (minimum 20 characters for small commands, having in mind the timestamp saved).

PR #5699, opened 2016-12-09
The $HOME/.rbenv folder should have a bigger priority

ka8725:
This folder should get the biggest priority because it should be default accordingly to rbenv documentation.

This is the pr based on [this](https://github.com/robbyrussell/oh-my-zsh/commit/baf2ab1bf1096b954a759739cc41c587fb0cf5b2#commitcomment-20133062) conversation.


jshort:
I've run this this change for the last week and rbenv works as I expect it to.

robbyrussell:
Believe I merged in a similar request -- sorry for not catching this one first

PR #5705, opened 2016-12-11
add virtualenv support

tomwei7:
ys.zsh-theme add python virtualenv support

before
image
after
image

### DIFF
--- a/themes/ys.zsh-theme
+++ b/themes/ys.zsh-theme
@@ -35,6 +35,15 @@ ys_hg_prompt_info() {
 	fi
 }
 
+# Virtualenv info
+export VIRTUAL_ENV_DISABLE_PROMPT=1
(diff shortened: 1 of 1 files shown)


belak:
A few small comments:

* I believe [[ and ]] are usually preferred to [ and ]
* I'm not a huge fan of forcing VIRTUAL_ENV_DISABLE_PROMPT to on - this means if a user loads this theme then loads a different theme, there will be no virtualenv in the prompt at all. The agnoster theme (as an example) adds VIRTUAL_ENV_DISABLE_PROMPT in its check for displaying the virtualenv.

timfeirg:
checkout https://github.com/caiogondim/bullet-train.zsh#python-virtualenv-pyenv

tomwei7:
@timfeirg nice!

PR #5707, opened 2016-12-11
add custom tab title cmd ignore pattern support

hiberabyss:
Currently, commands like `ssh` or `sudo` could be ignored as term tab title. Sometimes, user may want to ignore some other commands as the title.

This pull request supports user to define their own command ignore pattern.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -68,8 +68,12 @@ function omz_termsupport_preexec {
     return
   fi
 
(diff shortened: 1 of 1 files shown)


mcornella:
This is not a feature we want to build into OMZ. You can maintain your own customization of the title function yourself though; look up how in [the customization wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization).

PR #5714, opened 2016-12-15
Fix emacs client terminal

### LINKED ISSUE #3305: 'e' command do not open emcas
I've installed emacs24 and enabled emacs plugin in or-my-zsh .zshrc file. Then when I tried open emacs using `e` command it is immedialty quiting (going back to promt). No error or any other messages are shown on command line.

My quick fix if to modify emacsclient.sh file:

``` diff
diff --git a/plugins/emacs/emacsclient.sh b/plugins/emacs/emacsclient.sh
index 625201a..636339d 100755
--- a/plugins/emacs/emacsclient.sh
+++ b/plugins/emacs/emacsclient.sh
@@ -1,7 +1,7 @@
 #!/bin/sh

 # get list of available X windows.
-x=`emacsclient --alternate-editor '' --eval '(x-display-list)' 2>/dev/null`
+x=`emacsclient -t --alternate-editor '' --eval '(x-display-list)' 2>/dev/null`

 if [ -z "$x" ] || [ "$x" = "nil" ] ;then
     # Create one if there is no X window yet.
```

Now, all emacs command will open emacs frame in current terminal. I'm not sure if this is the proper fix for this problem (eg. maybe one would prefer to have '-c' instead of '-t' there)

dluksza:
This is my simple solution to fix #3305 

With this small change 'e' command always open new emacsclient in terminal

### DIFF
--- a/plugins/emacs/emacs.plugin.zsh
+++ b/plugins/emacs/emacs.plugin.zsh
@@ -16,7 +16,7 @@ if "$ZSH/tools/require_tool.sh" emacs 24 2>/dev/null ; then
     # set EDITOR if not already defined.
     export EDITOR="${EDITOR:-${EMACS_PLUGIN_LAUNCHER}}"
 
-    alias emacs="$EMACS_PLUGIN_LAUNCHER --no-wait"
+    alias emacs="$EMACS_PLUGIN_LAUNCHER -t"
     alias e=emacs
     # open terminal emacsclient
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, @dluksza. A test of this worked for me

mcornella:
Reverted this in #7597 because it broke many people's workflow. Use `te` instead or redefine the emacs alias in your own zshrc file.

PR #5716, opened 2016-12-15
Fix plugin description

jmayday:
* fixed git-flow-avh.plugin.zsh description


jmayday:
closing due to incorrect fix

PR #5717, opened 2016-12-16
Adding $ or # signs after git indicator

aalireza:
The `prompt_char` function is a copy of the same one by @Bregor for [Gentoo theme](https://github.com/robbyrussell/oh-my-zsh/blob/master/themes/gentoo.zsh-theme).


robbyrussell:
My apologies for never getting around to this. Given how long this has lingered, I am closing to clean up our backlog a bit.

PR #5720, opened 2016-12-20
adding up plugin

pedrohenriquepires:
A simple plugin to go up in folders

usage: 
```
up NUMBER_OF_DIRECTORIES
```


mcornella:
I appreciate this is useful to you but there are other ways to do this.

One being the current global aliases which expand from `...` to `../..`, `....` to `../../..`, etcetera, having one extra dot be another level. This can be then used with or without a `cd`, given that we also set the option `auto_cd`.

Another solution is having a kind of `magic-dot` function which expands to an extra `/..` if it's appropriate. I have been working on [a prototype of that in my dotfiles](https://github.com/mcornella/dotfiles/blob/2547a21a/ohmyzsh-custom/magic-dot.zsh) and I will make a PR to the main repository when it's ready. [Here's a quick demo](https://asciinema.org/a/9jvejc2svmjo3rzu1tlrk9upx).

If you feel this solves your workflow and don't care, you can add that function to your zshrc file or [make it into a custom plugin](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization).

The way you solved it is kind of genius btw. Thanks!

pedrohenriquepires:
Oooh, I didn't know this feature... Living and learning, thanks! haha

Hope that we can use this `magic-dot` soon.

mcornella:
Yeah when I saw it I thought it was a killer feature as well. I'll let you know when it's ready for testing!  :wink:

Issue #5721, opened 2016-12-21
The git plugin has too many aliases


mcornella:
This is a recurring sentiment:

https://github.com/robbyrussell/oh-my-zsh/issues/3288#issuecomment-145481037
https://github.com/robbyrussell/oh-my-zsh/issues/3288#issuecomment-267463829
https://github.com/robbyrussell/oh-my-zsh/issues/3682#issuecomment-78437079

This has many harmful side-effects, mainly added commands (some of which are destructive and at one keystroke). Some aliases also might override installed commands, which then need to be selectively disabled or renamed. There is #4585 pending but no closer to being merged.

Meanwhile, more and more people want to add their own little alias, without thinking about the added benefits or costs to the common interest.

One possible fix for that is having bundles of aliases for specific workflows, such that if certain git commands are used together very frequently we alias them and group these aliases into specific functions or if branches. For example, if one does not use commit signing regularly, not having that be there by default.

This is a discussion issue and it is in a very early stage, so we're quite far of a PR fixing this.
What other solutions can you think of?

obcilion:
I'd be fine with removing all but the most common ones (add, status, branch etc.) and letting users define their own after that. Having bundles is also a good idea, but how would installing of alias bundles be managed?

mcornella:
My thinking is maybe adding some configuration variables, like `$GIT_PLUGIN_USE_SIGNING`, and having one for each bundle of aliases. The user would then set up these configuration variables in their zshrc file.

Another option would be to have multiple functions that define such aliases, and again, the user would call them in their zshrc files. 

I'm open to other suggestions and I'll think about other options we may have. But yes, many aliases could be purged from the main plugin and leave only the absolute necessary.

muesli:
I was getting rather annoyed by all kinds of "random" aliases defined on my system by various oh-my-zsh plugins, so I went ahead and separated the aliases from the plugins for the git and golang plugin at least.

See here:
https://github.com/muesli/oh-my-zsh/commit/9db77ac56f01458ea8aad30d75aa2a885cd5c345

This allows you to load aliases separately in your .zshrc:
aliasdefs=(golang)

shinenelson:
I'd prefer just segregating them into multiple plugins based on their usage. (Just my 2 ¢)

noraj1337:
[...]

PR #5723, opened 2016-12-22
Add 'git remote show origin' as grso

josephmilla:


RyanBrushett:
Check out issue #5721. Some good discussion about whether or not the git plugin already has too many aliases and whether or not they're useful.

josephmilla:
I think people would use this command a lot

mcornella:
`gr` is for `git remote`, and using autocompletion you can quickly get to `show origin`.

PR #5726, opened 2016-12-24
Improve UX for Spotify quit command in osx plugin

RyanBrushett:
Spotify command UX is a little weak for people who live and work almost entirely in shells.
- `spotify quit` should not open Spotify if it is not already running. Should confirm that Spotify is indeed not running.
- `spotify quit` should not blow away the user's shell once Spotify is quit. This can be a disruption to work flow.

This PR looks to add a few little checks which will help improve this experience.
This PR also adds a space to line 477 between `break` and `;;` for consistency. Doesn't seem like a big enough change to put in its own PR.

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -416,7 +416,7 @@ function spotify() {
         if [ "$(osascript -e 'application "Spotify" is running')" = "false" ]; then
           cecho "Spotify was not running."
         else
-          cecho "Quitting Spotify.";
+          cecho "Closing Spotify.";
           osascript -e 'tell application "Spotify" to quit';
         fi
         break ;;
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -323,7 +323,7 @@ function spotify() {
   if [ $# = 0 ]; then
     showHelp;
   else
-    if [ "$(osascript -e 'application "Spotify" is running')" = "false" ]; then
+    if [ "$1" != "quit" ] && [ "$(osascript -e 'application "Spotify" is running')" = "false" ]; then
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks! I assume you've verified it works but I'll pull someone to test this as well.

Looks good on paper though.
I specifically dislike the code style of the plugin but that's for another PR, I'll change it to something more idiomatic in the near future. 

Thanks for this!

RyanBrushett:
Yep I've been using it myself but definitely test it out. Sounds good to me!
And no worries! :) I love this plugin, happy to contribute!

xserrat:
Hi @mcornella & @RyanBrushett , I've just tested it and it works properly :) . However, I'd put the `Quitting Spotify.` at the else statement, because it'll more clear when spotify is quitting and when it 's already quitted:

```zsh

"quit"    )
   if [ "$(osascript -e 'application "Spotify" is running')" = "false" ]; then
        cecho "Spotify was not running."
   else
        cecho "Quitting Spotify.";
        osascript -e 'tell application "Spotify" to quit';
   fi
   break ;;
```

RyanBrushett:
Ahh good eye! Excellent point.
I'll touch that up.

Thanks for your review @xserrat !

RyanBrushett:
I've moved the `'Quitting Spotify.'` line to within the `else` block as per reviewer comments. I agree - the behaviour is clearer this way. Thanks for the review!

xserrat:
Thank you too for the PR, in my opinion, ready to merge!

mcornella:
Thanks to the both of you for making Oh My Zsh even more awesome than it is today! Oh My Zsh's great, but we can always improve it! ✊ 😂

I've changed `Quitting` in the message because `Closing` is more idiomatic English. Thanks again! ❤️

PR #5729, opened 2016-12-27
Change color for ref names to auto for shortcuts glol and glola

VannTen:
Allow the ref names to have differents colors if they are remote refs
or local refs, and another color for HEAD
(use the same coloring scheme as --decorate option)


gkaklas:
I agree: it is more beautiful and the color-coding of branches, HEAD etc is helpful. Also, I just noticed that almost every "git log" alias has "--decorate", except for `glol` and `glola`, so accepting this PR makes these aliases consistent with the other `--oneline`-like aliases

PR #5730, opened 2016-12-28
Change af-magic theme's branch color

haandol:
I love `af-magic` theme because of it's simplicity!
but its coloring is too simple to recognize which branch is now on.
so I changed branch color from 075, almost same color as path(prefix) color, to 078(same tone but green color)

And also I removed 'branch:' description for displaying branch information.
I think it is conceptually duplicated because the part is only for the branch status.

From

To

### DIFF
--- a/themes/af-magic.zsh-theme
+++ b/themes/af-magic.zsh-theme
@@ -27,7 +27,7 @@ else
 fi
 
 # git settings
-ZSH_THEME_GIT_PROMPT_PREFIX="$FG[078](branch:"
+ZSH_THEME_GIT_PROMPT_PREFIX="($FG[078]branch:"
 ZSH_THEME_GIT_PROMPT_CLEAN=""
 ZSH_THEME_GIT_PROMPT_DIRTY="$my_orange*%{$reset_color%}"
 ZSH_THEME_GIT_PROMPT_SUFFIX="$FG[078])%{$reset_color%}"
--- a/themes/af-magic.zsh-theme
+++ b/themes/af-magic.zsh-theme
@@ -27,7 +27,7 @@ else
 fi
 
 # git settings
-ZSH_THEME_GIT_PROMPT_PREFIX="$FG[075](branch:"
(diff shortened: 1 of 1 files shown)


mcornella:
LGTM but please change the color of the first parenthesis to be the same as the last.

mcornella:
Oh, I meant the same kind of blue. Right now they're all green as the branch name right?

haandol:
understood. 👍

themes/af-magic.zsh-theme
```diff
@@ -27,7 +27,7 @@ else
 fi
 
 # git settings
-ZSH_THEME_GIT_PROMPT_PREFIX="$FG[075](branch:"
+ZSH_THEME_GIT_PROMPT_PREFIX="($FG[078]branch:"
 ZSH_THEME_GIT_PROMPT_CLEAN=""
 ZSH_THEME_GIT_PROMPT_DIRTY="$my_orange*%{$reset_color%}"
-ZSH_THEME_GIT_PROMPT_SUFFIX="$FG[075])%{$reset_color%}"
+ZSH_THEME_GIT_PROMPT_SUFFIX="$FG[078])%{$reset_color%}"
```
RyanBrushett: The parenthesis before 'branch' is a different colour than the `SUFFIX` parenthesis.

The following example makes sure they're both the same colour:
```ZSH_THEME_GIT_PROMPT_SUFFIX=")$FG[078]%{$reset_color%}"```

haandol: I want to make parenthesis blue while only branch part is green. 
I checked my pull request works as I expected and you can see it works on the screenshot I uploaded.

As I know, color directives( ex. $FG[078]) affect to characters next to it, so I don't understand `$FG[078]%{$reset_color%}` does mean anything. I think the code set next characters green and reset the color to white immediately.

please check that and let me know if I'm wrong, thanks!

mcornella:
Thanks everybody!

hjpotter92:
Would help if you also update the image in [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Themes#af-magic)

PR #5738, opened 2016-12-31
Dev

aaron-goshine:


.gitignore
```diff
@@ -1,8 +1,43 @@
-# custom files
-custom/
-!custom/plugins/example
-!custom/example.zsh
-
-# temp files directories
-cache/
-log/
+.DS_Store
```
xserrat: Hi @aaron-goshine ! 

In order to follow best practices, it would be better without adding any reference about IDE or something like this at .gitignore file 😉

mcornella:
Please only keep these features in your own repository: you add too much customizations and in a single PR which is harder to review. 

Furthermore, you add your own theme which I can't merge because [we can't accept more themes](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page.

PR #5739, opened 2016-12-31
Fully ignore the custom directory in .gitignore

dstroot:
First of all these two lines force me to never delete any of the example files or I cannot take normal updates:

-!custom/plugins/example
-!custom/example.zsh

Second, custom/ means I cannot symlink the custom folder elsewhere (for example to include it in a backup or my .dotfiles).  By changing custom/ to custom we get full control of the custom directory while maintaining normal .oh-my-zsh updates.

Otherwise you get this:
```
~/.oh-my-zsh master
❯ upgrade_oh_my_zsh
Updating Oh My Zsh
error: cannot pull with rebase: You have unstaged changes.
error: please commit or stash them.
There was an error updating. Try again later?
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,5 @@
 custom
 
 # temp files directories
-cache
-log
+cache/
+log/
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 # custom files
-custom/
-!custom/plugins/example
-!custom/example.zsh
+custom
 
 # temp files directories
-cache/
-log/
+cache
+log


mcornella:
You shouldn't have to do either of these things.

The `example` plugin is reserved for only an example plugin and it only contains comments (no zsh code is run), so deleting it isn't needed.

Now for the symlink part: you don't need that. You can set up a custom `ZSH_CUSTOM` variable in your zshrc which is then the custom directory used by the init script.

TL;DR: wontfix because these are fixed using the custom features already present in Oh My Zsh.

dstroot:
"Now for the symlink part: you don't need that. You can set up a custom ZSH_CUSTOM variable in your zshrc which is then the custom directory used by the init script."

Awesome!  Did not know that (obviously).

mcornella:
There's a bit of boilerplate code in the default zshrc template with most of the available settings. We should probably explain that when installing Oh My Zsh 😅

El mié., 4 ene. 2017 20:49, Dan Stroot <notifications@github.com<mailto:notifications@github.com>> escribió:

"Now for the symlink part: you don't need that. You can set up a custom ZSH_CUSTOM variable in your zshrc which is then the custom directory used by the init script."

Awesome! Did not know that (obviously).

borland502:
Yeah, this bites me with the mackup project, but that's more a problem with how it does backup

PR #5744, opened 2017-01-02
Added racket plugin for auto-completion

fmind:
Hello,

The plugin includes the completion script created by the language maintainers:
https://github.com/racket/shell-completion

The completion works on my installation for the raco and racket commands.

Regards


mcornella:
Has no interest.

PR #5760, opened 2017-01-07
plugins/docker/_docker: update completion (upstream ee9bb3)

stepf:


Sparragus:
I was about to do the same.

The changes are a copy paste from https://github.com/docker/docker/blob/master/contrib/completion/zsh/_docker.

LGTM

mcornella:
Superseeded by newer PRs.

PR #5764, opened 2017-01-09
Add plugin for hexo blog framework

Misaghlb:
small auto completion plugin for hexo static site generator


mcornella:
Has no interest.

PR #4798, opened 2016-01-24
Restore ssh-agent variables if found alive gpg-agent

andrey-yantsen:
Those changes fix working with gpg-agent instead of ssh-agent when first one was run outside zsh.

PR #5771, opened 2017-01-11
Restore ssh-agent variables if found alive gpg-agent

andrey-yantsen:
Those changes fix working with gpg-agent instead of ssh-agent when first one was run outside zsh.

(instead of #4798)


mcornella:
Is this needed with the new version of the plugin?

klaper:
New version of plugin seems to fixing this problem

mcornella:
Thanks!

PR #5773, opened 2017-01-12
Ubuntu plugin improvements

XzAeRo:
* Update README for better documentation.
* Added `safe-upgrade-clean` and `agdu` aliases.
* Added Spotify control commands and it's respective documentation.


mcornella:
The README is better now, and the spotify function looks like it belongs to a different plugin.

PR #5776, opened 2017-01-13
Add `-u` parameter to the tmux command to support unicode in certain themes

hjpotter92:
Based on an answer provided here: <https://askubuntu.com/q/410048>

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -2,9 +2,9 @@
 # Aliases
 #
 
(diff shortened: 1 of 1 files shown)


larson-carter (approved):
Great work. This could help others out a lot.

robbyrussell:
Thanks for this, and I'm sorry it sat here for so long without a reply. We ended up covering this in #5166, which added an opt-in `ZSH_TMUX_UNICODE=true` for the same thing, so I'm closing this one out.

PR #5777, opened 2017-01-13
add '-s' attribute

jordan1997:
ADB changes.
Very ofter I want use '-s' for run command for the specific device.
But in the previous version when I input '-s' I can not show hint about command. (I fixed this)


mcornella:
Fixed in #6121

PR #5782, opened 2017-01-14
nvm: install and autoload correct node version if .nvmrc present

drGrove:


jsg2021:
would love to see this land 😊

PR #5785, opened 2017-01-15
git_prompt_info: show the tag name

robsonpeixoto:
Now the `git_prompt_info` will show the tag name instead of show the `sha1` commit.

before: `git:(35d6f37a3)`
now:`git:(v1.3.6)`


BastienAr:
I've added tag/tagname for greater clarity, but was in the same spirit. I will not create a PR to avoid flooding, but I just leave the code here in case of a merge of this PR.

```shell
ref="tag/$(command git describe --exact-match HEAD 2> /dev/null)" || \
```

robsonpeixoto:
@BastienAr open another one.

PR #5786, opened 2017-01-16
templates/zshrc.zsh-template: add default zstyle

HugoY:
The zstyle completion format is needed to correctly display
the _messages tag

Without zstyle ':completion:*' format 'Completing %d':

out

With zstyle ':completion:*' format 'Completing %d':
out2

You can see the 
```
Completing head
Completing commit object name
```
messages.


mcornella:
If we did this, it would be in `lib/completion.zsh` instead. Regardless, this is a personal preference.

PR #5787, opened 2017-01-17
Use 'apt' instead of 'apt-get' for Ubuntu >=16.04

HaraldNordgren:
`apt` replaces `apt-get` on Ubuntu 16.04.

So this code figures out which Ubuntu version we are running, and then uses the correct Apt version.

### DIFF
--- a/plugins/ubuntu/ubuntu.plugin.zsh
+++ b/plugins/ubuntu/ubuntu.plugin.zsh
@@ -5,8 +5,11 @@
 # https://github.com/trinaldi
 # Nicolas Jonas nextgenthemes.com
 # https://github.com/loctauxphilippe
+# https://github.com/HaraldNordgren
 #
 # Debian, Ubuntu and friends related zsh aliases and functions for zsh
+
+(( $+commands[apt] )) && APT=apt || APT=apt-get
 
 alias acs='apt-cache search'
 compdef _acs acs='apt-cache search'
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
Could someone merge or close this one?

HaraldNordgren:
Merge this?

HaraldNordgren:
@robbyrussell Merge this?

RlndVt:
Maybe use the same method of detecting apt/apt-get as proposed in the debian patch?

https://github.com/robbyrussell/oh-my-zsh/pull/6122/files

Seems like a neater method than relying on version numbering.

HaraldNordgren:
@RlndVt Update the code =)

plugins/ubuntu/ubuntu.plugin.zsh
```diff
@@ -8,15 +8,21 @@
 #
 # Debian, Ubuntu and friends related zsh aliases and functions for zsh
 
+if [[ -e $( which -p apt 2>&1 ) ]]; then
```
mcornella: A better check if `apt` exists, in one line:
```zsh
(( $+commands[apt] )) && APT=apt || APT=apt-get
```

plugins/ubuntu/ubuntu.plugin.zsh
```diff
@@ -91,7 +97,7 @@ aar() {
 		PACKAGE=${1##*/}
 	fi
 	
-	sudo apt-add-repository $1 && sudo apt-get update
+	sudo apt-add-repository $1 && sudo $APT update
 	sudo apt-get install $PACKAGE
```
mcornella: apt-get should also be changed here.

HaraldNordgren:
@mcornella Updated the code.

LIttleAncientForestKami:
Thank you @mcornella @HaraldNordgren @RlndVt  <bows>

PR #5802, opened 2017-01-23
Add lein aliases plugin that shows aliases from project.clj in autosuggestions

paasar:
Extends lein plugin if it is enabled.
"lein" should be before "lein-aliases" in `$plugins`

There is still a problem with one of my testers (macOS Yosemite). `lein <tab>` fails with:
`_lein_aliases:7: number expected`
And I can't figure out why. Updating to latest zsh wasn't helping.

But it works on El Capitan, Sierra and Xubuntu 16.10.

### DIFF
--- a/plugins/lein-aliases/lein-aliases.plugin.zsh
+++ b/plugins/lein-aliases/lein-aliases.plugin.zsh
@@ -1,0 +1,37 @@
+__LEIN_ALIASES_DIR="${0:A:h}"
+
+function _lein_aliases() {
+    unset _LEIN_ALIASES_STR
+    unset _LEIN_ALIASES
+
+    local ret=1 state
+    _arguments ':aliases:->aliases' && ret=0
+
+    case $state in
+      aliases)
+        _LEIN_ALIASES_STR=$(_read_lein_aliases)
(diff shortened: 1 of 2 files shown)


paasar:
Ok, I found an older Mac to debug this on.

It seemed that it doesn't like `local`.
After removing those I ran into another problem where it seems to try and read the `project.clj` file from a wrong place. I'll try to fix this whenever I have time.

paasar:
Ok, I was able to fix the issue with older macOS:ses. The problem was using `local` inside the case construct.
The other problem was just my own mistake with the project.clj file.

This PR is now ready for review and merge.

Marcholio:
Can this be merged?

paasar:
@Marcholio This is now ready for merge.

robbyrussell:
Hi @paasar, thanks for this.

Could you provide a README for this plugin, too? (can see other plugins for references)

Having said that, we aren't seeing much engagement on this to see that there is enough demand from the OMZ community to include this.

robbyrussell:
Am going to close this for now. Would be happy to reconsider with a README included

PR #5809, opened 2017-01-25
Aptible-CLI Plugin

Frederick888:
For `aptible-cli` from https://www.aptible.com/


Frederick888:
I'm no longer using this and considering it's a quite niche CLI it'd be better to move it to an independent repo imo. If anyone still needs this, feel free to open your repo and copy the files over.

PR #5810, opened 2017-01-25
Make all scripts in tools directory are executable

mul14:
We can upgrade oh-my-zsh by run this command `./upgrade.sh` instead of `sh ./upgrade.sh`


mcornella:
They can't be called directly as they are now, it's not as easy as flipping the x bit.

PR #5811, opened 2017-01-26
Fix check for VIRTUAL_ENV_DISABLE_PROMPT with the agnoster theme

chrw:
I had the problem using the agnoster theme and have Python virtualenv in place. I set the environment variable VIRTUAL_ENV_DISABLE_PROMPT and this should prevent the prompt to show the virtualenv Environment. The check was bad because with -n the variable is true if the variable is set. I fixed it to use -z and it works for me.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -193,7 +193,7 @@ prompt_dir() {
 # Virtualenv: current working virtualenv
 prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
-  if [[ -n $virtualenv_path && -n $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
+  if [[ -n $virtualenv_path && -z $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
     prompt_segment blue black "(`basename $virtualenv_path`)"
   fi
 }


belak:
I see what you're getting at here. It would be nice to disable parts of the prompt in agnoster, but this check appears to be here on purpose. With your changes, the prompt will show up multiple times (once provided by virtualenv, which is disabled by this variable and once by the theme).

chrw:
Ok. I can not reproduce the multiple prompts on my side. I don't think it's a feature of the theme, it's just the coloring, etc.? Is there a way to disable it in oh-my-zsh or the theme? I have a "default" virtualenv which will show up every time. ;)

belak:
Sorry for the confusion. I've just tested it to make sure I understood what was going on. With your change, the themed virtualenv part of the prompt never shows up at all (If you don't set VIRTUAL_ENV_DISABLE_PROMPT, the normal virtualenv display happens - you don't see the themed version... and if it is set, nothing shows up at all).

The idea behind using VIRTUAL_ENV_DISABLE_PROMPT in the theme was to only display the themed version when the normal virtualenv wouldn't display.

belak:
With your patch and VIRTUAL_ENV_DISABLE_PROMPT not set, it will display like this: https://www.dropbox.com/s/lcqc3jer598ien2/Screenshot%202017-03-03%2000.36.56.png?dl=0

chrw:
Maybe it's because I'm not using virtualenv outside of the "global environment". Do you use the zsh virtualenv plugin?

But I still don't get the logic of the check in the theme. If $VIRTUAL_ENV_DISABLE_PROMPT is set (without considering the value), go and show it? That was my motivation for the patch.

Maybe we need to add some more logic to avoid the prompt showing multiple times.

belak:
VIRTUAL_ENV_DISABLE_PROMPT is from virtualenv (https://github.com/pypa/virtualenv/blob/master/virtualenv_embedded/activate.sh#L56), and the usage here is to prevent the double display, as I showed in the above screenshot. I agree that logically it doesn't make sense, but that's why it's currently there. :)

chrw:
I close this issue for now, since I don't use virtualenv anymore and it seems to be a personal problem.

Issue #5816, opened 2017-01-26
(virtualenvwrapper) Changing from one virtualenv project folder to another doesn't automatically switch virtualenv


jayusa123:
If you have multiple projects with corresponding virtualenvs and use the oh-my-zsh virtualenvwrapper plugin, changing from one project folder to another will not automatically change the virtualenv for you.

jayusa123:
This appears to be fixed in the latest code.

Issue #4603, opened 2015-11-11
virtualenvwrapper plugin fail to deactivate virtualenv when parent directory is a git repository


gled-rs:
Consider the following directory layout:

/home/src
/home/src/gitrepo => git repository
/home/src/gitrepo/adirwithvirtualenv => a repository with a virtualenv ( either a .venv or else ).

When going from adirwithvirtualenv to gitrepo ( cd .. ), deactivate is not called because $ENV_NAME is filled with the name of the directory, as it is a git repository.

gled-rs:
Pull request: #4604

apjanke:
I think this is intended behavior. Git repos, even without venv-specific customizations, are considered to be project roots, so when this happens, it should switch you from the `adirwithvirtualenv` venv to the `gitrepo` venv. I think just calling `workon` without `deactivate` first will accomplish that. This change looks like it would end up fully deactivating the venvs in the example case. (Is that what it's intended to do? Or should `deactivate` just be called before activating the new environment?)

Have a look at the discussion in #4122, #4110, and #3918. There was a while when `virtualenvwrapper` did not consider plain git repos to be project roots, but this was a behavior accidentally introduced by #3918.

Maybe you'd want to add a setting that controls whether ordinary git repos are considered project roots? At any case, we should probably run this by the other `virtualenvwrapper` users before merging, since they've already had some unexpected behavior changes this year.

gled-rs:
In my case, gitrepo is not a virtualenv. Hence the annoying behaviour.

But yes, it may be a change for some people, thanks for the headsup !

Pull request has been updated to add an option DEACTIVATE_VENV_ASAP to control that behaviour.

Without setting that option, the behaviour is the same as previously.

apjanke:
Cool.

Maybe name it `VENV_DEACTIVATE_ASAP` so it's prefixed with the plugin name? We're starting to get a lot of config variables floating around, and that might help keep it organized.

jayusa123:
What's the status of this issue? It would appear that this is a problem not only when the parent directory is a git repo but any time the destination directory of the cd command is. (see referenced issue above)

jayusa123:
I just submitted a PR with my stab at this issue. I took a different approach which also resolves another issue (#5816). Any thoughts are much appreciated.

PR #5817, opened 2017-01-26
Fix a couple issues with virtualenvwrapper

jayusa123:
This PR contains fixes for the following issues:

1. If the destination directory of the `cd` command is also a git repository, the virtualenv will be automatically deactivated, as expected when navigating out of a project directory while using the virtualenvwrapper plugin. (Issue #4603)
1. If the destination directory of the `cd` command is also a virtualenv project directory, its corresponding virtualenv will be automatically activated. (Issue #5816)

Both issues were solved by splitting the activation and deactivation of virtualenvs into separate if blocks and reversing their order.

I've successfully tested this behavior on my setup and it looks good. Further testing by other users of the virtualenvwrapper plugin would be much appreciated!

### DIFF
--- a/plugins/react-native/README.md
+++ b/plugins/react-native/README.md
@@ -24,3 +24,6 @@ plugins=(... react-native)
 | **rnios5s** | `react-native run-ios --simulator "iPhone 5s"` |
 | **rnios6**  | `react-native run-ios --simulator "iPhone 6"`  |
 | **rnios6s** | `react-native run-ios --simulator "iPhone 6s"` |
+| _Logging_                                                    |
+| **rnland**  | `react-native log-android`                     |
+| **rnlios**  | `react-native log-ios`                         |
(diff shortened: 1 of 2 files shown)


sgtsquiggs:
This change causes the virtual env to deactivate/reactivate on every `cwd`. `CD_VIRTUAL_ENV` is set on line 74, and `VIRTUAL_ENV` is set by `workon`. When you change to a subdirectory, both pass the `-n` check and deactivate is called.

Is this desired?

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -61,6 +61,12 @@ if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
       else
         ENV_NAME=""
       fi
+      
+      if [[ -n $CD_VIRTUAL_ENV && -n $VIRTUAL_ENV ]]; then
```
sgtsquiggs: Suggest changing this to
```
if [[ -n $CD_VIRTUAL_ENV && "$ENV_NAME" != "$CD_VIRTUAL_ENV" ]]; then
```

jayusa123: Good catch. I just made that change.

jkassisbw:
closed without merging?

jayusa123:
This issue was accidentally closed after I force pushed a commit. I've added a new PR, #6687.

PR #5821, opened 2017-01-27
feat(awsplugin): use acp command to switch profile without the right …

ThamosIO:
…prompt

Will work well with:

https://github.com/caiogondim/bullet-train-oh-my-zsh-theme/pull/199


Marcholio:
This could either be merged or closed

mcornella:
We'll go with the solution at #7615.

PR #5824, opened 2017-01-28
Change how oh-my-zsh check if zsh is installed

### LINKED ISSUE #4955: Installer assumes zsh is not installed if it's not in /etc/shells
Hi,

I have zsh installed and used it already some time, but without oh-my-zsh. I would like to give it a try but ran into the error that the install.sh didn't recognize my already installed zsh?

```
[userX@hostX1]~% zsh --version
zsh 5.1.1 (x86_64-redhat-linux-gnu
[userX@hostX1]~% rpm -qa | grep zsh
zsh-5.1.1-4.fc23.x86_64
[userX@hostX1]~% echo $SHELL
/bin/zsh
[userX@hostX1]~% cat /etc/redhat-release
Fedora release 23 (Twenty Three)
[userX@hostX1]~% sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
**Zsh is not installed! Please install zsh first!**
[userX@hostX1]~% 
```

Could you give me a hint what's wrong? It's nothing special on my zsh or on my Fedora 23 setup.

Thanks,

regards,
ulf

yursan9:
In stateless OS like Solus or Clear Linux, they didn't have /etc/shells.
fix #4955

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -24,7 +24,7 @@ main() {
   # which may fail on systems lacking tput or terminfo
   set -e
 
-  CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
+  CHECK_ZSH_INSTALLED=$(grep -E ^"${USER}:" /etc/passwd|cut -d : -f 7)
   if [ ! $CHECK_ZSH_INSTALLED -ge 1 ]; then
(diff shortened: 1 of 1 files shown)


belak:
Unfortunately, this doesn't appear to work on OSX

v0idst4r:
I believe that my PR #6398 is a more elegant solution.

Issue #5560, opened 2016-10-21
Directory completion not working


rodfersou:
When I try to change directory:
```
$ cd /s/s.s.l <tab>
```

I expect that it completes as:
```
$ cd /srv/sc.social.like
```

but it is not working.. I  don't know where to check this bug.. it worked fine in my old ubuntu, but this week I start using arch linux, and this bug is driving me crazy..

omz-debug.log
https://gist.github.com/anonymous/b4329bd15bcb526658992031a0fd7269

mcornella:
This will be the issue to track this problem, for everybody that has it. So please subscribe for updates.

mcornella:
For now there is this workaround that seems to work for everybody: #1398 (comment). Have a look and report any issues.

rodfersou:
@mcornella the workaround didn't work for me.. is it possible to explain what commit start the mess? I didn't find it

rodfersou:
nevermind.. I could add the workaround now! thanks

PR #5825, opened 2017-01-28
Add old-omz-completion plugin

jasonm23:
Provides a simple way to use the old omz completion

https://github.com/robbyrussell/oh-my-zsh/issues/5560


plugins/old-omz-completion/old-omz-completion.plugin.zsh
```diff
@@ -0,0 +1 @@
+zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
```
rodfersou: This suggestion looks better (and more readable)
https://github.com/robbyrussell/oh-my-zsh/issues/5560#issuecomment-408190809

mcornella:
Has no interest.

PR #5830, opened 2017-01-30
Add 'grbom' alias -> 'git rebase origin/master'

theneva:
I often find myself doing `gfa` and then `grb origin/master` on a branch before pushing it to submit a PR.


mcornella:
If you have `master` up-to-date you won't need that since grbm will work as well. Otherwise typing it out and using autocompletion isn't that bad.
Thanks for your contribution

PR #5834, opened 2017-01-31
Add docker common functions

nicosommi:
Just a couple of useful aliases for docker commands.
- dserve will lift a little nginx (alpine based) serving static files from the current directory, it will output the port in which the container was bind
- dserve-php will lift an apache with php installed serving the current directory
- doc will run a command in a containerized zsh
- sdoc the same as doc but without bounding volumes. Useful for privacy / security reasons.


mcornella:
This has become polluted.

PR #5836, opened 2017-02-01
AppUp plugin, vagrant and docker-compose commands

mdeboer:
This plugins adds `start`, `stop`, `up` and `down` commands when it detects a docker-compose or Vagrant file in the current directory (e.g. your application). Just run `up` and get coding!

### Docker

Aside from simply running `up`, you can also extend your configuration by running `up <name>`, which will run `docker-compose` with both `docker-compose.yml` and extend it with `docker-compose.<name>.yml`. For more on extending please see the official docker documentation: https://docs.docker.com/compose/extends. Additional arguments will be directly supplied to the docker-compose.

### Compatibility

It does not interfere with any other docker/vagrant plugins as far as I know.

### DIFF
--- a/plugins/appup/README.md
+++ b/plugins/appup/README.md
@@ -1,0 +1,13 @@
+## AppUp
+
+**Maintainers:** [mdeboer](https://github.com/mdeboer)
+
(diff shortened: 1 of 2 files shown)


mdeboer:
An up-to-date and maintained version of this plugin is available here: https://github.com/Cloudstek/zsh-plugin-appup

PR #5838, opened 2017-02-01
gcob alias (checkout new branch)

jshort:
I prefer to use gco -b to create a new branch and check it out in one fell swoop.  This alias would enable that with `gcob`.


mcornella:
There is already `gcb`.

PR #5841, opened 2017-02-02
rails: create README

i5okie:
Just made readme file.
Could be used as a cheat sheet.


andrik:
Hi @i5okie,

What do you think about adding the description below? 

This plugin adds completion for [Ruby On Rails Framework](https://rubyonrails.org/) and [Rake](https://ruby.github.io/rake/) commands, as well as some aliases for logs and environment variables.

To use it, add `rails` to the plugins array in your zshrc file:

```zsh
plugins=(... rails)
```

PR #5842, opened 2017-02-02
Add profiling for ZSH startup

ddidier:
Hi, my installation was very slow (3 seconds to open a new terminal) so I needed to find what was causing it. I ended up with some little functions that I put together and polished. I'm really far from a shell expert so I hope my code isn't too bad ; and that it could be integrated to be useful for others. Feel free to change anything you want!

### DIFF
--- a/lib/profiler.sh
+++ b/lib/profiler.sh
@@ -1,0 +1,57 @@
+
+# Store the measurements indexed by the key (name)
+typeset -A _time_by_keys
+# Store the order the measurements were taken
+_ordered_keys=()
+
+# Return a timestamp in milliseconds
+function _time_in_ms() {
+    echo $(( $(date +%s%N) / 1000000 ))
+}
+
+# Start recording the elapsed time
+# @param key the name that defines the measurement (must be unique)
+function start_profiling() {
+    [[ $ENABLE_PROFILING = "true" ]] || return
+
+    local key="$1"
+    local time_ms=$(_time_in_ms)
+    _time_by_keys[$key]=$time_ms
+    _ordered_keys+=($key)
+}
+
+# Stop recording the elapsed time
+# @param key the name that defines the measurement (must be the same than the one used with 'start_profiling')
+function stop_profiling() {
+    [[ $ENABLE_PROFILING = "true" ]] || return
+
+    local key="$1"
+    local time_ms=$(_time_in_ms)
+
+    if [[ $_time_by_keys[$key] ]]; then
+        _time_by_keys[$key]=$(($time_ms - $_time_by_keys[$key]))
+    else
+        echo "WARNING: you must start the profiling with 'start_profiling \"$key\"')"
+    fi
+}
+
+# Print all the measurements
+function print_profiling() {
+    [[ $ENABLE_PROFILING = "true" ]] || return
+
+    local dots1='..................................................'
+    local dots2='......'
+
+    echo
+    for key in $_ordered_keys; do
+        local value=$_time_by_keys[$key]
+        local percent=$(($_time_by_keys[$key] * 100 / $_time_by_keys[TOTAL]))
+
+        local length1=$(( ${#key} + ${#value} ))
+        printf "%s %s %s ms " $key "${dots1:$length1}" $value
+
+        local length2=$(( ${#percent} ))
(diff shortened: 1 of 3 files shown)


aakoch:
Which system was this written in? I'm having issues running on macOS.

`date` using "%N" isn't supported on macOS: https://stackoverflow.com/questions/9804966/date-command-does-not-follow-linux-specifications-mac-os-x-lion

I'm also getting this error when trying to run profiler.sh (but not when I just type that line in the terminal).
`> ./profiler.sh
./profiler.sh: line 3: typeset: -A: invalid option
typeset: usage: typeset [-afFirtx] [-p] name[=value] ...`

ddidier:
Thanks for your input. I developed this on Ubuntu.

As I said I'm no shell expert, and as a matter of fact I know much more Bash than ZSH ;-)
And I never had a Mac which is a Unix and not a Linux.

For `date` not working with `%N` it seems that the only options are to either install `coreutils` to use `gdate` or to use some portable hack such as https://stackoverflow.com/questions/16548528/command-to-get-time-in-milliseconds/40067967#48765103. Can you tell me if this works on macOS?

For `_time_in_ms` AFAIK a function cannot return a value, so this is correct.

For `typeset -A` I don't know what to say since I cannot test on macOS and I didn't find anything with Google. Can you elaborate a little please?

aakoch:
I tried the hack you linked to, but that didn't work. Seems Mac's don't have `/proc/uptime`. (Not even `/proc`)
```
./time.sh: line 2: /proc/uptime: No such file or directory
./time.sh: line 4: /proc/uptime: No such file or directory
```

I tried `gdate` and that works; I just don't know how feasible it is since you would have to test if `date` or `gdate` is required.

For the `typeset -A`, I tried downloading `profiler.sh` and running it, but it error'd:
```
> ./profiler.sh                                                                                                                         
./profiler.sh: line 3: typeset: -A: invalid option
typeset: usage: typeset [-afFirtx] [-p] name[=value] ...
```
I'm not familiar with that command. What is it supposed to do?

ddidier:
So I fixed the date issue with a new commit.

But I don't know why `typeset -A` fails on macOS. I read that the Bash version is 3.x on macOS and that this type was added in 4.0, but since the script is using ZSH... What version of ZSH are you using? Moreover I see that this command is used 85 times in OhMyZsh. I don't know what to do...

mcornella:
For the date you can use zsh/datetime's [`$EPOCHREALTIME`](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#index-EPOCHREALTIME) which is a floating point value composed of `SECONDS.NANOSECONDS`. You can do `$(( EPOCHREALTIME * 1000 ))` to get the miliseconds and `print '%0.f' $(( EPOCHREALTIME * 1000 ))` to get the miliseconds without the decimals. This is built into zsh, the only thing is that you need to load the module zsh/datetime with `zmodload zsh/datetime` (you only need to load it once).

The `typeset -A` error is because @aakoch is running it as a script instead of sourcing it within OMZ like you did. @aakoch if you clone the pull request you should have the whole setup in place to test it properly.

mcornella:
> For the date you can use zsh/datetime's $EPOCHREALTIME which is a floating point value composed of SECONDS.NANOSECONDS.

This is what I mean:
```diff
# Return a timestamp in milliseconds
function _time_in_ms() {
-    echo $(( $(_gnu_date +%s%N) / 1000000 ))
+    printf %.0f $(( EPOCHREALTIME * 1000 ))
}
```

ddidier:
@mcornella thanks a lot!
@aakoch please try my latest commit

ddidier:
@mcornella Hi, any chance to have this merged in master?

KANGOD:
An oh-my-zsh startup profile timestamp is strongly needed, while this PR has been inactive for 2 years.
Hope the author would schedule this feature in future development.
And thanks ddidier, I might try your profiling.

maxbrunet:
I found sometime ago that Zsh has a built-in profiling modules: [zsh/zprof](https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html#The-zsh_002fzprof-Module), it has been enough to troubleshoot slowness for me

```zsh
# ~/.zshrc
zmodload zsh/zprof # top

# ...

zprof # bottom
```

PR #5843, opened 2017-02-02
Easier handling of user services

auipga:
prefix `uc-` is negotiable, maybe `usc-` or `scu-` ... what do you think?


auipga:
obsolete by #6661

PR #5847, opened 2017-02-04
fix if install nvm with homebrew can not load

zhenhappy:
fix if installed with homebrew can not load


mcornella:
Went with #8316 instead.

PR #5853, opened 2017-02-06
Add a flag to prevent the "launch" of zsh

Polynomdivision:
For me personally, it is very annoying when I install a fresh copy
of linux and I run my script to install everything but it gets stuck
at the installation of oh-my-zsh because it waits for zsh to exit.
Because of that I added the flag "--no-env-zsh", and yes I know that
it's a dumb name, to prevent the installer from "starting" zsh if the
user specifies the "--no-env-zsh" flag.
I didn't find a PR like this or a way to prevent the installer from
launching zsh in the script. I also don't know if it's useful for anyone
except me.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -110,7 +110,12 @@ main() {
   echo 'p.p.s. Get stickers and t-shirts at http://shop.planetargon.com.'
   echo ''
   printf "${NORMAL}"
-  env zsh
+
+  # Check if the user wants to switch into zsh right after the installation
+  # (! $1 = "--no-env-zsh")
+  if [ ! $1 = "--no-env-zsh" ]; then
+    env zsh
+  fi
 }
 
 main


slavaGanzin:
In your case I think you can launch oh-my-zsh installation in separate shell:

```
(sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)") &
```

Polynomdivision:
For me the problem with this would be (1) that then the installation and the "deployment" of the settings would run asynchronous and (2) the installation script would move an already existing zshrc to zsh.pre-oh-my-zsh. (2) would then interfere with the "deployment" and I personally prefer not using things like `sleep 2`.
For me these were the main inspirations for changing the installation script

belak:
I think #5893 (when it's merged) would also do what you want. This is an issue which keeps popping up across multiple PRs and issues, so I hope there's a resolution soon.

mcornella:
#5169 has added the option to disable that, either by setting the environment variable `RUNZSH=no` or passing the flag `--unattended`, which also skips the change of default shell. It has a bunch of other fixes, and it's now ready to test.

Thanks!

PR #5857, opened 2017-02-07
cursor color changing for vi-MODES

paranormal:
There is also \e instead of v for editing command line in EDITOR, like you would do it in postgress. Feel free to remove this change.

The bindkey -v were in original file, I'm not sure if it should be moved to user config like @slavaGanzin suggested. Feel free to decide.

@slavaGanzin I can't test what you mentioned about 'eval "override-vi-color-$(declare -f zle-line-init)"', but eval is evil anyway shell it or not. May be it can be fixed in other way?

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -44,6 +44,10 @@ zle-line-init () {
   override-vi-mode-zle-line-init
 }; zle -N zle-line-init
 
+zle-line-finish () {
+  zle -K viins
+}; zle -N zle-line-finish
+
 autoload -U select-quoted
 zle -N select-quoted
 for m in visual viopp; do
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -34,12 +34,14 @@ function zle-keymap-select() {
   fi
 }; zle -N zle-keymap-select
 
+eval "override-vi-mode-$(declare -f zle-line-init)"
 zle-line-init () {
   if [ $TERM = "screen" ]; then
     echo -ne $VIM_INSERT_COLOR
   elif [ $TERM != "linux" ]; then
     echo -ne $VIM_INSERT_COLOR
   fi
+  override-vi-mode-zle-line-init
 }; zle -N zle-line-init
 
 autoload -U select-quoted
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -54,7 +54,7 @@ zle -N edit-command-line
 
 bindkey -v
 
-# allow v to edit the command line (standard behaviour)
+# allow v to edit the command line (standard psql)
 autoload -Uz edit-command-line
 bindkey -M vicmd '\e' edit-command-line
 
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -54,7 +54,7 @@ zle -N edit-command-line
 
 bindkey -v
 
-# allow v to edit the command line (standard psql)
+# allow v to edit the command line (psql behaviour)
 autoload -Uz edit-command-line
 bindkey -M vicmd '\e' edit-command-line
 
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -1,23 +1,62 @@
 # Updates editor information when the keymap changes.
+
+if [ ! -n "${VIM_NORMAL_COLOR+x}" ]; then
+  if [ $TERM = "screen" ]; then
+    VIM_NORMAL_COLOR='\033P\033]12;#FF3333\007\033\\'
+  else
+    VIM_NORMAL_COLOR="\033]12;#FF3333\007"
+  fi
+fi
+
+if [ ! -n "${VIM_INSERT_COLOR+x}" ]; then
+  if [ $TERM = "screen" ]; then
+    VIM_INSERT_COLOR='\033P\033]12;#99FF33\007\033\\'
+  else
+    VIM_INSERT_COLOR="\033]12;#00FF00\007"
+  fi
+fi
+
 function zle-keymap-select() {
   zle reset-prompt
   zle -R
-}
+  if [ $TERM = "screen" ]; then
+    if [ $KEYMAP = vicmd ]; then
+      echo -ne $VIM_NORMAL_COLOR
+    else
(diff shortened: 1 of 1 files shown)


paranormal:
Few typo mistakes made me do 3 commits.
Everything is ok except override-vi-color use-case, I don't understand it anyway, @slavaGanzin could you fix by yourself?

japhir:
I've been using your gist code so far, and it works out great. But I still prefer the visual distinction between a bar cursor `|` for insert mode and a block cursor `█` for normal mode because it is identical to how it looks in my emacs with evil. I didn't get what you meant by needing to look to away from where you type, my buggy code substituted the cursor, not the beginning or end of the prompt.

Maybe doing both a colour and shape change would be even better?

slavaGanzin:
@paranormal 
> but eval is evil anyway shell it or not

It's just a way to compose functions in sh:
```
eval "override-vi-color-$(declare -f zle-line-init)" #Copy original zle-line-init 

zle-line-init () {                                                    #Redefine zle-line-init
  #Our custom logic
  override-vi-color-zle-line-init                            #Call original zle-line-init
}
```

> The bindkey -v were in original file

@japhir
> Maybe doing both a colour and shape change would be even better?

https://github.com/paranormal/oh-my-zsh/blob/c743766852fe4725ec82bac0ac659b43e675fa37/plugins/vi-mode/vi-mode.plugin.zsh#L3
You can set
```
VIM_NORMAL_COLOR='\e[2 q'
VIM_INSERT_COLOR='\e[6 q' 
#or
VIM_INSERT_COLOR='\e[6 q$VIM_INSERT_COLOR'  #and get color and bar/block simultaneously 
```
 and that should do the trick

paranormal:
Iit seems that the ideas are the same, we just could tweak realization to do right highlighting, colors or cursor change.
@japhir do you have any ideas apart from using VIM_NORMAL_COLOR/VIM_INSERT_COLOR to accomplish this?

paranormal:
> If you can find a better way, I'll appreciate.

Ok, I'll try to find a better way at the end of this day, could you describe me your use-case again. I don't really understand what is broken and why.

japhir:
@paranormal unfortunately I must say I'm a little over my head here. That is to say, I have no idea what I'm doing. The latest change has made my terminal open emacs (not the terminal version but the graphical version) every time I enter normal mode, so as far as I understand, I don't think the \e option is doing much good for me...

Also, all of these changes to the cursor indicator do not work once we enter a command-line editor such as vim or `emacsclient -t`.

paranormal:
@japhir hm.... try removing \e from VIM_NORMAL_COLOR='\e[2 q' or commend the misfortune bindkey binding otherwise.

> Also

I know, there's no way to anything about it, every time you use exec to something this feature is lost. Probably forcing it to NORMAL is an option. I'll take a look when this day is over.

japhir:
@paranormal Ah no, it even does it without the newly set `\e[2 q` stuff, just your latest commit. I've reverted to a working version of your original gist for now and will wait for the pull request to be merged.

slavaGanzin:
@paranormal 

zle-line-init is zsh internal function, you can't just override it: you will break something depending on it
In my case is my plugin for git auto fetching https://github.com/robbyrussell/oh-my-zsh/pull/5477 (which utilizes same trick to not overwrite original zle-line-init) and fzf

paranormal:
Ok, got it.
I see  we can test it with `whence` or just as you suggested. Could you apply your fix to my commit?

slavaGanzin:
@paranormal 
I have no access to your repo. Please copy from this commit:
https://github.com/slavaGanzin/oh-my-zsh/commit/ab0601c0e8333a4e56ccbad53c4b861825c863d6

Thanks for pointing to ```whence```:
```
whence -v zle-line-init
zle-line-init is a shell function from /home/vganzin/git-auto-fetch.plugin.zsh
```

@japhir 
```
#.zshrc
VIM_NORMAL_COLOR="\033]12;#FFCC00\007\033[6 q"
``` 
**\033[6 q** do the trick

japhir:
@slavaGanzin awesome! Thank you guys so much, you latest commit works as expected.

paranormal:
Done.

slavaGanzin:
@mcornella Marc, this PR is ready for merge. Can you review it, please?

paranormal:
Just figured out what to do with external shells, ssh/psql/mysql/irb/etc when we run some exec syscall.
This would reset state to INSERT.

paranormal:
@slavaGanzin btw. Are you sure you need zle-line-init in your module? Could your job be done in other way?
It seems to me that this functional shouldn't be triggered from modules not related to command line editing.

paranormal:
Actually no, don't commit that. I'll come with better solution. The change to INSERT should be only done if there was an exec to something. Otherwise it should remain in the same state as it was before.

paranormal:
Do you have any ideas guys how to proceed here? It's not a prompt2 and I can't see any way to detect if it's something we give shell for long term usage or just run an ordinary command.

slavaGanzin:
@paranormal 
> Are you sure you need zle-line-init in your module? 
There is no "yours" or "my" module. And you can't dictate anyone not to use ```zle-line-init``` because it will broke "your" module. It's ridiculous.

paranormal:
There is. It's your module which broke my module!!!

paranormal:
Just Kidding. Keep it simple, I have no hard feelings... No need to argue because of misleading my/your.
I don't code in zsh every day - this is why I asked.

paranormal:
What I need is to figure out is what to do with commands which have their shell. If you would run ssh from command line in COMMAND mode then you'll have misleading cursor for ssh session.

slavaGanzin:
> What I need is to figure out what to do with commands which have their shell. 

I tried a lot to change default readline mode to vi. Like this:
http://vim.wikia.com/wiki/Use_vi_shortcuts_in_terminal

But it doesn't work for me. So I just stick with emacs mode in it(a bit frustrating of course). 
So I think it would be a problem if I see normal cursor color in emacs-mode.

paranormal:
Or, we can merge this since we don't have other solution and this solution is better then non at all.

paranormal:
@slavaGanzin try mine https://gist.github.com/paranormal/fc294b27e80fcdc55ffef239be4be8d2

slavaGanzin:
I tried your inputrc and it doesn't work for node or ipython. (Archlinux, urxvt 9.22)

How readline is related to libinput?

paranormal:
Without readline... I had an issue with libinput recently and it shoot, nwm.

Node doesn't have readline because of GPL or some license issues, I'm too struggling using Node.js so I checked a long time ago...
Not sure about ipython since I hadn't worked in python for years.

paranormal:
Is it ready to be commited?

slavaGanzin:
@paranormal I'm not sure .inputrc can be part of oh-my-zsh distribution.

PR #5858, opened 2017-02-08
List runnable NPM scripts alias

lukebrowell:
Added this alias for short-hand way for getting list of run script targets, usage:
$ npmr                    # lists available scripts
$ npmr myScript    # runs myScript

GTM?


mcornella:
Already in the code.

PR #5864, opened 2017-02-10
Add several aliases for Yarn

Cryrivers:


plugins/yarn/yarn.plugin.zsh
```diff
@@ -1,4 +1,9 @@
 alias yi="yarn install"
+alias yg="yarn upgrade"
```
Cryrivers: i'm not sure why i put `yg` as an alias of `yarn upgrade`. 😂

PR #5866, opened 2017-02-10
Silent exit when virtualenvwrapper is not found

zhaostu:
I think this is a better behavior to just exit because virtualenvwrapper is not installed. It's kind of annoying when some machine does not have virtualenvwrapper and the error message keeps popping up.


belak:
If the plugin is enabled though, shouldn't it give a warning (as it currently does) letting the user know why that plugin isn't working? Otherwise, it would be confusing for people trying to debug this.

zhaostu:
@belak Other oh-my-zsh plugins, like rbenv, does not give a warning either when it cannot find the corresponding executables.

The current behavior is super annoying because for people who share configurations between machines, some of the machines may not have virtualenvwrapper installed, and this warning just keeps popping up when a new shell is launched.

belak:
That makes sense. The other plugins being that way are a good point.

zhaostu:
Probably nobody cared

PR #5868, opened 2017-02-11
Adds --show-signature to git log aliases

filidorwiese:
For GPG users, it's nice to have the `--show-signature` flag added to the git log aliases and it doesn't hurt anybody else :)


mcornella:
This breaks the display of users that have gpg installed but don't have the public key of the signer:

image

You can redefine them in your zshrc file.

PR #5880, opened 2017-02-14
Add the session restoration support for macOS

Kentzo:
Also known as Resume, this feature allows one to store
current environment and later restore it thanks to the fact
that Terminal.app (as well as iTerm) provdes the TERM_SESSION_ID
environment variable which is preserved between restarts.

### DIFF
--- a/plugins/session/README.md
+++ b/plugins/session/README.md
@@ -1,0 +1,29 @@
+Terminal.app on macOS provides the `TERM_SESSION_ID` environment variable into every shell process.  
+Out of the box, for bash, it allows one to define the `shell_session_save_user_state` function which  
+can write any shell command to the `SHELL_SESSION_FILE` file.
+
+If user then exit Terminal.app with open sessions, this function will be called for each of them.  
+On the next launch, Terminal.app will source session files into corresponding sessions.
+
+This plugin implements that logic for Zsh.
+
+E.g. to automatically restore Python's virtual environment, just add the following to you `.zshrc`:
+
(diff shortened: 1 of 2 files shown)


Kentzo:
Here is an example of how one could automatically restore Python's virtualenv:

```zsh
function shell_session_save_user_state() {
    # Resulting files is sourced at the very beginning,
    # which maybe to early.
    # It's better to have control over when session should be restored.
    echo "function restore_session() {" >> $SHELL_SESSION_FILE

    if [[ -n ${VIRTUAL_ENV} ]]; then
        echo source \"${VIRTUAL_ENV}\"/bin/activate >> $SHELL_SESSION_FILE
    fi

    echo "}" >> $SHELL_SESSION_FILE
}

if typeset -f restore_session >/dev/null; then
    restore_session
fi
```

53jk1 (approved):

robbyrussell:
@Kentzo I appreciate you sending this our way. I'm going to decline this for now as we haven't had any engagement from people who are looking to see this get included. If we were to see an uptick of interest, I would request that we rename the plugin to something more Termina.app-specific (as 'session' is a touch on the generic side).

If anyone is interested in using this, you can definitely toss it into your custom plugins/ folder!

Kentzo:
FWIW, this plugin is no longer needed: as of Big Sur (as of Catalina?) macOS provides native implementation of this functionality via `/etc/zshrc_Apple_Terminal`

PR #5883, opened 2017-02-15
Fix #5405 - emacsclient - don't open new window.

### LINKED ISSUE #5405: Opening files with emacsclient, creates a second frame...
When I open files with `e` from the command line emacs opens a second frame. Why not open the file in the main frame? 

https://github.com/jakobadam/oh-my-zsh/blob/master/plugins/emacs/emacsclient.sh#L8

kind regards,
Jakob

jakobadam:


mcornella:
We did that but it broke some people's machines; see https://github.com/robbyrussell/oh-my-zsh/issues/5405#issuecomment-483044531. We're looking for a better solution.

PR #5884, opened 2017-02-16
Fix pyenv slow issue

herblover:
Based on rbenv performance improvements commit.

@gabetax 's improvement - https://github.com/robbyrussell/oh-my-zsh/commit/daedfcab510b01d566a2dff6bc94a9e91a563a18#diff-3538c2cbfa22922156729920217acdee
@sheerun 's improvement - https://github.com/robbyrussell/oh-my-zsh/commit/0ed1adeaadfc03a5abb767df5ad76a1ddd2d7c69

This fix improve zsh startup time when pyenv plugin is enabled.

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -7,7 +7,7 @@ _pyenv-from-homebrew-installed() {
 }
 
 FOUND_PYENV=0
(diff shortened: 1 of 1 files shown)


mcornella:
The new version doesn't use homebrew so this PR is obsolete. Thanks for your effort though!

herblover:
@mcornella thank you! I'm happy with @cyphus 's work merged.

PR #5887, opened 2017-02-17
add context of ± 3 lines to grep

mralext20:
This is a useful addition to the grep file. 

i understand if you close this PR, as it is not perfect or usable by everyone, but many will probably like its addition.


belak:
I feel like this might be better as an alias, as it breaks the usage of grep when doing quick oneliners.

mcornella:
What @belak said. You can add that to the grep alias keeping the previous value by doing
```zsh
alias grep="$aliases[grep] -C 3"
```

PR #5891, opened 2017-02-18
Adds option to disable auto update prompt

RobFreiburger:
I noticed this option was missing from the template. When I followed the directions in README to enable it, I put it below line 59 without noticing it wouldn't have any effect by then. Prevent others from making a rookie mistake like mine.


mcornella:
Thanks!

Issue #5873, opened 2017-02-12
Install oh-my-zsh in a script


mkitby:
Hello,

I use this script to install oh-my-zsh
```
#!/bin/bash
sudo apt-get update
sudo apt-get install zsh -y
chsh -s $(which zsh)
sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
########################
mkdir ~/bin

stuff...
```

After install the oh-my-zsh, the script will exit and the reamining commands are not executed. What should I do if I want to execute it?

Thanks

loket:
Just ran into the same problem. The reason is the `env zsh` command in [install.sh](https://github.com/robbyrussell/oh-my-zsh/blob/551abfcbb48a0c001eadef80abc3276af4e9ad26/tools/install.sh#L113), which launches zsh at the very end of the installation.

So I was thinking the script could take in an additional argument (like `--no-env` or something more intuitive), which could be checked before running `env zsh` in the script.

That way you can install Oh My Zsh from a script, without breaking the existing install procedure.

I'll try to create a pull request when I have time.

mkitby:
Hi loket, Thanks for your input.

loket:
Obviously, while waiting for review of the pull request, you can use the script in my repo instead:

```shell
sh -c "$(curl -fsSL https://raw.githubusercontent.com/loket/oh-my-zsh/feature/batch-mode/tools/install.sh)" -s --batch || {
  echo "Could not install Oh My Zsh" >/dev/stderr
  exit 1
}
```

Comments are welcome! 😄

Ferrer-Jeremy:
You can also do:
```shell
wget https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh
sed -i.tmp 's:env zsh::g' install.sh
sh install.sh
```
The `sed` will remove the `env zsh` at the end

richardlock:
You can also run the following to do this in a one liner similar to the normal install:

```
sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh | sed 's:env zsh::g' | sed 's:chsh -s .*$::g')"
```

C0rn3j:
`env zsh` is now `env zsh -l` so the workaround would now be as follows:

```
[...]

PR #5892, opened 2017-02-18
Allow batch mode installation

loket:
Installing Oh My Zsh is currently not possible as part of a batch script (see #5873).

This pull request adds a `--batch` argument to allow installation without launching `zsh` at the end, allowing the installation script to be combined with other scripts.


loket:
Actually, executing with `sh -c` isn't working as expected. Closing issue for now.

PR #5893, opened 2017-02-18
Allow batch mode installation

loket:
Replacement for #5892. 

Installing Oh My Zsh is currently not possible as part of a batch script (see #5873).

This pull request adds a `--batch` argument to allow installation without launching `zsh` at the end, allowing the installation script to be combined with other scripts.

Also updated readme with instructions.

**Edit**: To test the pull request, replace `https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh` with `https://raw.githubusercontent.com/loket/oh-my-zsh/feature/batch-mode/tools/install.sh`. Something like this:

```bash
sh -c "$(curl -fsSL https://raw.githubusercontent.com/loket/oh-my-zsh/feature/batch-mode/tools/install.sh)" -s --batch || {
  echo "Could not install Oh My Zsh" >/dev/stderr
  exit 1
}
```


BurningSmile:
I would very much appreciate if this was accepted, this would fix my issue with scripting this as well!

I will go about testing this in some scripts in vm envs and let you know how it goes!

loket:
Should maybe consider renaming from `--batch` to something like `--non-interactive`?

AndrewSB:
+1 can we get someone a contributor to look at this and get it merged in?

YuMS:
@loket `--no-run-after`, ` will help people get the point at first glance.

loket:
@YuMS: Fine by me :smile: Any other suggestions? Otherwise I'll try to update the PR soon.

Jubijub:
Really looking forward to this

zwpaper:
@robbyrussell 
It have use it for a while, seems not problem for me, my use case:
1. create a docker with software installed
2. install `oh-my-zsh`
3. copy my `.zshrc`
4. start `sshd`

I use the docker as my dev environment, It's my daily used env, and recreated many times.
you can find my script here:
https://github.com/zwpaper/devdocker/blob/master/env.rc#L7

If you need any other help, I can try my best to help.

tools/install.sh
```diff
@@ -23,7 +23,27 @@ main() {
...
+check_optional_args() {
+  # Set default values
+  LAUNCH_ZSH_AFTER=1
+
+  # Evaluate arguments
+  for arg in $@; do
+    case "$arg" in
+      --batch)
+        LAUNCH_ZSH_AFTER=0
+        printf "${BLUE}--batch:${NORMAL} zsh will not be started after installation\n"
```
FloopCZ: This breaks when `${BLUE}` is an empty string (e.g., Docker environment). In such a case `--` is interpreted as argument separator and the command fails.

Please replace it by the following line to avoid the issue, or just don't write the `--batch ` at the beginning of the string:
```
        printf "%s\n" "${BLUE}--batch:${NORMAL} zsh will not be started after installation\n"
```

ameykusurkar:
Really looking forward to seeing this getting merged! Would finally allow me to call `install.sh` in my `Makefile` without blocking execution

loket:
Sorry for taking forever to update the PR! Thanks @FloopCZ, I merged your changes 😄

GeorgThomassen:
Any news regarding non-interactive installs?

GerkinDev:
Hey there! I need this feature too, pretty badly, because I reinstall oh-my-zsh almost every 2 weeks (thanks to WSL wiped out at every Windows update).

May I help you for anything to see this merged?

Cheers!

leoj3n:
Instead of a dubiously named flag, another solution might be to close stdin in the "batch" call to `install.sh`:

```zsh
0>/dev/null sh -c "$(<install.sh)"
```

OMZ could then be modified so that it runs "non-interactively" when the STDIN file descriptor is not bound to TTY:

https://github.com/robbyrussell/oh-my-zsh/blob/2da39ae8045efbfed80d2a3318aa9b3c9f637146/tools/install.sh#L210

I imagine this approach would also gracefully handle situations like when commands sent over ssh with no TTY bound.

mixn:
Any update on this? The current behaviour makes oh-my-zsh installs in dotfiles a real drag. 😢

GeorgThomassen:
@mixn It's it possible to make it install without userinteraction by removing the lines where it changes the environment 

`curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh -o $DOTFILES/omz.sh
sed -i ''  '/echo/d' $DOTFILES/omz.sh
sed -i ''  '/env zsh/d' $DOTFILES/omz.sh
sh $DOTFILES/omz.sh
rm -rf $DOTFILES/omz.sh`

this downloads the file removes all the `echo` lines, and the `env zsh` line. This is on OS X hence the empty quotes as the first argument for sed

mixn:
@GeorgThomassen I had something like that in mind, too, but first wanted to see what the current situation is on this. 🙂 Thanks a ton nevertheless, I’ll likely stick with that way of doing it for now. 🙏✋

lhorace:
I ran into one issue with this PR, either we `sudo chsh -s {$SHELL}` in cases where unprivileged user has sudo access with `NOPASSWD` enabled, or abandon using `chsh` in batch mode? Running this script, hangs Ansible on `chsh`, on RHEL, where a password maybe required to change own shell.

respencer:
I've read through this thread, but I'm not clear what the holdup is.

What is needed to get an accepted PR?

leoj3n:
@respencer I've got a PR (#5320) without conflicts that solves this issue by using the file descriptor close method [described above](https://github.com/robbyrussell/oh-my-zsh/pull/5893#issuecomment-414151781).

However, #5320 needs testers, and nobody has stepped up since it was opened in mid-2016.

respencer:
> @respencer I've got a PR (#5320) without conflicts that solves this issue by using the file descriptor close method [described above (comment)](https://github.com/robbyrussell/oh-my-zsh/pull/5893#issuecomment-414151781).
> 
> However, #5320 needs testers, and nobody has stepped up since it was opened in mid-2016.

Okay, I'm new to this project. I don't know how you work. Can I volunteer for that and what do you need me to do if I can?

No the testing part. The pre and post testing steps.

mcornella:
Hi everybody, #5169 pulled in these changes and renamed the flag to `--unattended`, which both skips the change of default shell and running zsh at the end of the installation. It has a bunch of other fixes, and it's now ready to test.

Thanks!

mcornella:
Closed by #5169.

PR #5898, opened 2017-02-21
Update Haskell Tool stack command

g1eny0ung:
Hi @robbyrussell , I start to use the stack plugin yesterday and notice the commands of current master branch's stack plugin is different to the current stack version 1.3.2.So I pull a request to update it.

And there has a **DEPRECATED** command **uninstall**, I also update it to the plugin.I think it can remind others this command is useless.Maybe it will remove at later stack versions.I will pay attention to the version change and remove it in time.😄


g1eny0ung:
Close this, move to #6230 .

PR #5899, opened 2017-02-21
added vagrant machine status to status prompt. show ▶ and ◾

roman-1983:
I have added the vagrant machine status to status prompt. IT will show ▶ for a running vagrant machine and ◾ for a machine that is not running.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -202,12 +202,21 @@ prompt_virtualenv() {
 # - was there an error
 # - am I root
 # - are there background jobs?
+# - is there a vagrant box - if yes, is it running?
 prompt_status() {
   local symbols
   symbols=()
   [[ $RETVAL -ne 0 ]] && symbols+="%{%F{red}%}✘"
   [[ $UID -eq 0 ]] && symbols+="%{%F{yellow}%}⚡"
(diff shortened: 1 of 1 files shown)


roman-1983:
It will look like this

1__roman_roman-mbp____development_house_local_www_ebird__zsh_

alfredbez:
Awesome, but is there a way to wrap it into a plugin?

alfredbez:
We could add this to the existing [vagrant plugin](https://github.com/robbyrussell/oh-my-zsh/blob/98d8d3429f8b9fc2c4c109fb199a31c8d1735699/plugins/vagrant/_vagrant) and create a prompt-function just [like the battery plugin](https://github.com/robbyrussell/oh-my-zsh/blob/98d8d3429f8b9fc2c4c109fb199a31c8d1735699/plugins/battery/battery.plugin.zsh#L47-L61) and then [use it in different themes](https://github.com/robbyrussell/oh-my-zsh/search?utf8=%E2%9C%93&q=battery_pct_prompt)

robbyrussell:
I'd prefer to keep this out of theme and approached via a prompt function

nrdsp:
This only works with the agnoster theme right? Powerlevel9k doesn't seem to handle it.

PR #5902, opened 2017-02-22
reorganized, added aliases for apt

nextgenthemes:


nextgenthemes:
I think the comdefs are actually not needed if they point to the exact same command the alias points to right? Just adopted it form the Debian plugin when i initially committed this.

mcornella:
Superseeded #5787. 

> I think the comdefs are actually not needed if they point to the exact same command the alias points to right? Just adopted it form the Debian plugin when i initially committed this.

You are right, compdefs aren't needed because zsh can see what the alias value is.

PR #5914, opened 2017-02-25
New agnoster-short theme

lamchakchan:
### Context

* Created a new theme file named `agnoster-short.zsh-theme`
  * This is a copy of the `agnoster.zsh-theme` file with a modification to the `prompt-dir` function to return only the current directory path.

### Motivation

* The `agnoster` theme is beautifully composed.  But the prompt grows too long when the directory path is deep.  I've seen various other attempts at fixing this adhoc and I thought having this available in a single place shared from `oh-my-zsh` is appropriate.  Also, this is a copy of the original with the fix to be backwards compatible for other users still wanting to keep the full $PWD path for their prompt printout.

### DIFF
--- a/agnoster-short.zsh-theme
+++ b/agnoster-short.zsh-theme
@@ -1,0 +1,228 @@
+# vim:ft=zsh ts=2 sw=2 sts=2
+#
+# agnoster's Theme - https://gist.github.com/3712874
+# A Powerline-inspired theme for ZSH
+#
+# # README
+#
+# In order for this theme to render correctly, you will need a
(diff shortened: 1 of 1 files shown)


belak:
I totally understand the need for this PR (I've opened a few against the agnoster theme - #5923 - to shrink the path when certain modules are loaded) but under "Contributing" in the README, they've requested that no more themes to be submitted. There are provided instructions in including an external theme if you want to go that route.

https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes

belak:
Also, just found #5635 when going through other open PRs.

mcornella:

PR #5923, opened 2017-02-28
Use shrink_path in agnoster theme if available

belak:
This makes the prompt shorter if the user tried to load the shrink_path module and will leave it alone if they didn't.

I left off `-s` because it makes any dir starting with a . turn into just a dot which can be really confusing.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -187,7 +187,11 @@ prompt_hg() {
 
 # Dir: current working directory
 prompt_dir() {
-  prompt_segment blue black '%~'
+  if type shrink_path &>/dev/null; then
+    prompt_segment blue black "$(shrink_path -t -l)"
+  else
+    prompt_segment blue black '%~'
+  fi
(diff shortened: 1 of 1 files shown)


MhhhxX:
This helped me so much. Thanks a lot :smile: 

But I use `shrink_path -f`. `shrink_path -t -l` doesn't work for me.

rmkpatchaa:
Hi,
When I add this option to my agnoster.zsh-theme file, all the screen in terminal turns blue :(

robbyrussell:
Sorry for the long silence on this one — nine years is on us, not you.

A few people did find it useful along the way, and thank you for that. But we never got to it, and `prompt_dir` has been rewritten since, so the change no longer applies to the current theme. Given that we didn't act on it back then, we're going to assume it hasn't been a deal-breaker for the community, and close it as part of some overdue housekeeping.

If shortening the path in agnoster is still something you want, an issue describing it as a config option is the way to restart the conversation.

PR #5931, opened 2017-03-02
Allow install to continue if zsh is in path

### LINKED ISSUE #5925: Install fails due to missing zsh
My current distro does have the /etc/shells file. can't we have another way of checking the existence of zsh other than that file's contents? maybe try checking if zsh exists in the current PATH.?

I had to manually create the file and populate it to complete the installation

In case ones stuck, here is what i did,
create a file ```/etc/shells```
add the line ```/zsh```

belak:
I still need to test this, but it should work. This allows zsh to be in /etc/shells or in the PATH for the install to continue.

Fixes #5925

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -24,9 +24,12 @@ main() {
   # which may fail on systems lacking tput or terminfo
   set -e
 
-  CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
-  if [ ! $CHECK_ZSH_INSTALLED -ge 1 ] && ! which zsh 2> /dev/null; then
-    printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
+  ZSH_LOCATION=$(grep /zsh$ /etc/shells | tail -1)
+  if [ -z "$ZSH_LOCATION" ]; then
+    ZSH_LOCATION=$(which zsh 2>/dev/null)
+  fi
+  if [ -z "$ZSH_LOCATION" ]; then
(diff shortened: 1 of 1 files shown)


belak:
Seems like this would be an alternate fix for #4955 as well.

paulmelnikow:
The reason for the `/etc/shells` check is that `chsh` will only accept a shell listed in that file.

Is there any workaround on systems where it's not?

belak:
That's why it tries to find zsh in /etc/shells before falling back to checking the path. It seems like a better situation to allow the rest of oh-my-zsh to install and fail when setting the shell rather than bailing out of the whole install.

paulmelnikow:
Or skip setting the shell. Makes sense to me.

belak:
The current logic already skips setting the shell if chsh doesn't exist... I think it makes sense to try even if /etc/shells is missing.

rigrassm:
@paulmelnikow `chsh` doesn't necessarily rely on /etc/shells to be present. I haven't looked into it too much yet but it works without that file being present on Solus. It may be that they have an altered version of it or have it compiled with different options, but it will definitely work without the file.

```
root@thanos /home/rigrassm # chsh --shell /bin/zsh
root@thanos /home/rigrassm # cat /etc/shells
cat: /etc/shells: No such file or directory
```

mati865:
Is something blocking this PR?

belak:
This repo has been fairly inactive (aside from a bit last month) for [almost a year](https://github.com/robbyrussell/oh-my-zsh/graphs/commit-activity), so I'm not sure when this will get in. I'm not aware of anything blocking this though.

mcornella:
I pushed a slightly different version of this (you being a co-author of the commit). Sorry for the delay.

PR #5937, opened 2017-03-06
Add git commit amend

techmexdev:
Added:
git commit --amend -m
as:
gcamd


robbyrussell:
Believe this is redundant with current `gcam` alias

PR #5940, opened 2017-03-07
Thefuck: homebrew install documentation and thefuck repo link

olets:
The `thefuck` plugin currently directly users who don't have `thefuck` to install it with `pip`. This pr adds a note about installing via `homebrew`, the route `thefuck` recommends for macOS (which oob doesn't have `pip`)


mcornella:
Thanks!

PR #5941, opened 2017-03-07
Add Mina to auto-bundled apps

Bounga:
This PR is intended to add [Mina](http://nadarei.co/mina/) deployment tool to automatically bundled apps.


mcornella:
There's a setting for that: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/bundler#add-additional-gems-to-be-wrapped.

PR #5942, opened 2017-03-07
Add Hanami to auto-bundled commands

Bounga:
Add [Hanami](http://hanamirb.org) to auto-bundled commands.


mcornella:
There's a setting for that: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/bundler#add-additional-gems-to-be-wrapped.

PR #5948, opened 2017-03-09
Add default hostname check to Agnoster theme

rwolfson:
This adds the ability to have a DEFAULT_HOSTNAME variable for folks that have a primary remote host.


robbyrussell:
Closing as we sat on this too long and given how popular that theme is, not seeing any community engagement on this.

PR #5951, opened 2017-03-10
Add `--force-with-lease` for git push

Mehonoshin:
Let's make force-push safe ❤️


havenchyk:
@mcornella could you please review it?

mcornella:
Why does this add force-with-lease to an alias that doesn't use force?

Currently we have two commands that use force, one of them uses force-with-lease, so the user can choose:
https://github.com/robbyrussell/oh-my-zsh/blob/1487a2ad844bbdff24e1db1cfb086138f477c7e9/plugins/git/git.plugin.zsh#L109-L116

PR #5959, opened 2017-03-13
feat(plugins): `octozen` shows an Octocat zen quote on startup

gtrefs:
Added a plugin which `curls` `https://api.github.com/octocat` on startup and displays the returned message.

### DIFF
--- a/plugins/octozen/README.md
+++ b/plugins/octozen/README.md
@@ -1,5 +1,11 @@
 # Octozen plugin
 
-Displays a zen quote of octocat on start up.
+Displays a zen quote from GitHub's octocat on start up.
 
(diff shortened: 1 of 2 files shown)


joyarzun:
It's nice but Sehll stays a little slow to start

gtrefs:
@joyarzun thanks for the feeback. The reason might be the HTTP call to the API. I am not sure how to mitigate this.

enderahmetyurt:
:octocat:

robbyrussell:
> The reason might be the HTTP call to the API. I am not sure how to mitigate this.

We'll opt to pass `--connect-timeout` to `curl` to try and keep it speedy for people.

PR #5962, opened 2017-03-15
Source custom files after theme files

xinhuang:
I use oh-my-zsh on a embedded board and would like to monitor the CPU temperature via r-prompt. Tried to do that by appending temperature to r-prompt in custom file, but because theme files are sourced last, this approach won't work.

So I moved source custom files to the end of .zshrc. This will allow user to do minor customization to prompts by simply appending instead of creating a new theme.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -86,12 +86,6 @@ for plugin ($plugins); do
   fi
 done
 
-# Load all of your custom configurations from custom/
-for config_file ($ZSH_CUSTOM/*.zsh(N)); do
-  source $config_file
-done
-unset config_file
-
 # Load the theme
 if [ "$ZSH_THEME" = "random" ]; then
   themes=($ZSH/themes/*zsh-theme)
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it sat unanswered for nine years.

What you wanted works today without changing the load order. A `precmd` hook registered from a file in `$ZSH_CUSTOM` runs after the theme is sourced, so you can build `RPROMPT` there:

```zsh
# $ZSH_CUSTOM/rprompt.zsh
typeset -g _base_rprompt= _base_captured=0
function _append_cpu_temp() {
  if (( ! _base_captured )); then
    _base_rprompt=$RPROMPT
    _base_captured=1
  fi
  RPROMPT="${_base_rprompt} $(your_temp_command)"
}
precmd_functions+=(_append_cpu_temp)
```

Sourcing the custom files after the theme would flip precedence for everyone who sets prompt variables in `custom/`, so I'd rather not take it that way. Closing for that reason, not because the need was wrong.

PR #5977, opened 2017-03-19
Add plugin fz - seamless fuzzy tab completion for z

changyuheng:
fz is the missing fuzzy tab completion feature of z. Check out its [homepage](https://github.com/changyuheng/fz) for more information.


mcornella:
This is a whole command-line utility of its own. Open a new repository and add it to the [External plugins page](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins) if you want.

PR #5979, opened 2017-03-20
Added publish alias for npm

milanito:
We use this command quite often at work, so I though it would be a nice thing to save this one second every time :)


dimensi0n:
Already done...

PR #5980, opened 2017-03-21
Update vi-mode.plugin.zsh

qzchenwl:


plugins/vi-mode/vi-mode.plugin.zsh
```diff
@@ -46,5 +46,5 @@ function vi_mode_prompt_info() {
 
 # define right prompt, if it wasn't defined by a theme
 if [[ "$RPS1" == "" && "$RPROMPT" == "" ]]; then
-  RPS1='$(vi_mode_prompt_info)'
+  RPS1="$(vi_mode_prompt_info)"
```
erydo: This would only expand the prompt once, when this module is loaded. I do not believe this is a correct change.

PR #5983, opened 2017-03-22
Update commands to match newer Yarn versions

Ionaru:
Added arguments in documentation of 'add', 'remove', 'upgrade' and 'why'
Added 'upgrade-interactive'
Removed 'self-update'


thearrow:
would love to see some of these updated yarn completions make it in!

https://github.com/robbyrussell/oh-my-zsh/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20yarn

PR #6006, opened 2017-03-31
Add alias for 'git rebase develop'

michaljuris:
In many companies the main development branch is often ```develop``` instead of ```master``` so this alias would help us in our workflow.

Also there already exists alias ```gcd``` for ```git checkout develop``` so this would complement it well.

@ncanceill


michaljuris:
@ncanceill @robbyrussell Any chance this could be reviewed?

PR #6011, opened 2017-04-06
Add nohup plugin

micrenda:
Adding a new pluging. When pressing Ctrl + H it will prefix the current command with nohup and will add &> `command_name`.out at the end (and vice versa).

It is very useful when launching a lot of background scripts

### DIFF
--- a/plugins/nohup/README.md
+++ b/plugins/nohup/README.md
@@ -4,7 +4,7 @@ Add `nohup` to the current command pressing the `Ctrl + H` shortcut
 
 ### Usage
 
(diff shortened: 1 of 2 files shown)


mcornella:
Has no interest.

micrenda:
My plugin was an amazing piece of genius. You do not deserve it :)

mcornella:
I'm not saying it wasn't ;) Feel free to make a repository out of it and add it to the [External plugins wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins).

PR #6015, opened 2017-04-07
themes/steeef: support conda env in prompt

amitbeka:
Add support for showing the conda env in the prompt and not only the
virtualenv.

### DIFF
--- a/themes/steeef.zsh-theme
+++ b/themes/steeef.zsh-theme
@@ -9,8 +9,12 @@
 
 export VIRTUAL_ENV_DISABLE_PROMPT=1
 
-function virtualenv_info {
-    [ $VIRTUAL_ENV ] && echo '('%F{blue}`basename $VIRTUAL_ENV`%f') '
+function virtualenv_or_conda_info {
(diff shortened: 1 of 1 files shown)


ckartchner:
Note for anyone else looking at this, VIRTUAL_ENV_DISABLE_PROMPT=1 is a virtualenv flag. 
To avoid having issues with conda, the conda prompt also needs to be disabled. 

This can be done by adding the following line to ~/.condarc
`changeps1: false`
or, it can be set from the CLI by entering:
`conda config --set changeps1 false`
I added the above line to my theme to ensure conda always behaves as I expect. 
[Conda changeps1 reference](https://conda.io/docs/user-guide/configuration/use-condarc.html#change-command-prompt-changeps1)

amitbeka:
Thanks for the info!

mcornella:
You can now use the [`conda-env` plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/conda-env), and use [the guide here](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes) to modify your theme to add support for this.

PR #6020, opened 2017-04-11
Fix handling spaces in $ZSH

miwilc:
the $ZSH variable has a problem with PATHs having spaces in them.

eg: automatic cloning to /home/Michael Wilcox/.oh-my-zsh
fails due to the variable being set to /home/Michael

This fixes the problem.
(Tested on [Cygwin](http://www.starlig.ht))

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -61,7 +61,7 @@ main() {
       exit 1
     fi
   fi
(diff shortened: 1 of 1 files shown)


some1else:
There are a few more occurances. Another PR covers the rest:
https://github.com/robbyrussell/oh-my-zsh/pull/6587

PR #6023, opened 2017-04-12
git plugin: new alias gdcw for git diff --color-words

jrmiller82:
Signed-off-by: James R. Miller <james.ryland.miller@gmail.com>

**One line pull request:** adds new git plugin alias `gdcw` for the built-in git command `git diff --color-words`.


mcornella:
Equivalent to `gdcw` for `--word-diff`, if git-config's color.ui is enabled.

PR #6032, opened 2017-04-17
added alias for listing all user's repo status 

ahmedfaragmostafa:
`gar` it's git alias which listing all status for all your git repositories on your local


mcornella:
This is both too specialized (not useful to most) and too infrequent to be worthy of an added alias.

PR #6034, opened 2017-04-18
Add function git_prompt_legend

gergelypolonkai:
I was tired not always knowing what each character in my Git prompt mean, so I wrote this little function to help me identify them.


mcornella:
You can use `print -P` to parse prompt expansion characters.

PR #6037, opened 2017-04-20
Create robbylambda.zsh-theme

CarlosRA97:
adding another more theme, this one is robbylambda a combination between the default and lambda themes


mcornella:

PR #6043, opened 2017-04-24
Use HTTPS for manual git clone to avoid  MITM

DonnchaC:
The git:// transport is completely unauthenticated. An attacker on the local or upstream network can easily man-in-the-middle an oh-my-zsh update and get remote code execution on your system. Only the https:// git transport should be used.


johnp:
Also note that the git:// transport uses a port (tcp/9418) that's often blocked in company or university firewalls, while https:// usually just works.

mcornella:
Good thinking. Thanks!

PR #6049, opened 2017-04-25
Add OpenShift autocompletion

kevinkirkup:
Add autocomplete for `oc` for OpenShift.


kevinkirkup:
Any updates? It looks like pull requests for new plugins have fallen a bit behind.

PR #6051, opened 2017-04-26
Fix for tmux plugin: Do not auto-load tmux inside of Emacs/Vim

15cm:
> [[Fix #541] of prezto](https://github.com/sorin-ionescu/prezto/commit/e5a0e33f1c6ee95159f9c765ef5dae842b07d389)

When Emacs and Vim are launched from outside of an interactive shell,
$TMUX and $STY are not set; check for Emacs and Vim environment
variables instead.


mcornella:
This PR has become polluted. Please submit another one.

mcornella:
Pulled in ff6b4c8.

PR #6058, opened 2017-05-01
platform independent pb{copy,paste} functions

emanresusername:


emanresusername:
don't remember closing this :/
but  moved [here](https://github.com/emanresusername/zsh-plugins/tree/40567777ca9a4bf2e6a0703da6049ccfba278c4d/pasteboard) in case y'all are already overwhelmed with plugins like with themese :sweat_smile:

emanresusername:
for posterity:
[oh-my-zsh already did this](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/clipboard.zsh#L1)  :scream:

PR #6064, opened 2017-05-04
Added alias for tmux

jasonfungsing:


larson-carter:
@jasonfungsing Hello there. Can you please give me a somewhat detailed description of what this is? What it an be used for and why it should be implemented?

robbyrussell:
Closing an old PR as I think we can let people add this to their personal prefs

PR #6068, opened 2017-05-05
List ZSH_CACHE_DIR in default .zshrc template and ensure the directory exists

MattSturgeon:
Two part PR inspired by the [patch](https://aur.archlinux.org/cgit/aur.git/tree/0001-zshrc.patch?h=oh-my-zsh-git) applied to the oh-my-zsh [AUR package](https://aur.archlinux.org/packages/oh-my-zsh-git/).

I kept each part as separate commits, but feel free to squash if you want to.

Firstly, list `ZSH_CACHE_DIR` in the `.zshrc` template file to make it easier for users to find and distributes to modify. I tried to position it close to similar variables like `ZSH_CUSTOM`.

Secondly, ensure that `$ZSH_CACHE_DIR` actually exists before loading plugins (etc) since AFAIK plugins do not check the directory exists before making use of it.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -25,6 +25,11 @@ if [[ -z "$ZSH_CACHE_DIR" ]]; then
   ZSH_CACHE_DIR="$ZSH/cache"
 fi
 
+# Create the ZSH_CACHE_DIR if it doesn't exist
+if [[ ! -d $ZSH_CACHE_DIR ]]; then
+  mkdir -p $ZSH_CACHE_DIR
+fi
+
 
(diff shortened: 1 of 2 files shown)


MattSturgeon:
This should be fairly simple to review - anyone got 5 mins to take a look?

MattSturgeon:
@mcornella @robbyrussell Really sorry to pester you guys, but I feel like a small change like this may get lost on such a busy project otherwise!

mcornella:
Fixed in d2269180b074263f4a0d3df3f9de15a1abae8110.

PR #6070, opened 2017-05-07
Add shoryuken as a bundled command

phstc:
Given that `shoryuken` besides the start command, now support AWS SQS commands, such as `ls`, `purge`, `mv` etc, would be cool to have it a bundled command.


mcornella:
You can [add it to `BUNDLED_COMMANDS`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/bundler/README.md#add-additional-gems-to-be-wrapped).

PR #6071, opened 2017-05-08
Plugin,Arcanist: Add completion for the Arcanist plugin

ZviRackover:
Copied from the Arcanist source tree:
https://github.com/phacility/arcanist/blob/master/resources/shell/bash-completion


mcornella:
Obsolete.

PR #6073, opened 2017-05-08
Clone repo silently

matthewfranglen:
Currently the git clone will write the following to stderr:

    Cloning into '/${HOME}/.oh-my-zsh'...

This silences that message.

References #3251


matthewfranglen:
@ditta95aR I am not quite sure what you are saying

PR #6077, opened 2017-05-10
uses the ./jq parser to show node specific info

CodisRedding:
Uses the ./jq parser to show node specific info

At the moment the only function written is the `pv|package_version` function that shows the current node projects package.json version.

While inside the root of a node project (containing a package.json file) type:

`> pv`

or 

`> package_version`

Displays the project version

`v8.0.1`


mcornella:
Has no interest.

PR #6079, opened 2017-05-11
[git plugin] Run command only if there is any input

grongor:
It doesn't make sense to run `git branch -d $BRANCH` if the `$BRANCH` is empty.

cc @ncanceill


grongor:
bump

PR #6080, opened 2017-05-11
speed up chruby plugin by eliminating(as much as possible) calls to brew

T0mK0:
We don't need to call 'brew --prefix' 3 times.
In worst case we need to call it once or in case of default brew installation - never.
Single call to brew is about 200-400 ms , so we save almost a second of shell startup time in worst case.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -1,9 +1,24 @@
 _homebrew-installed() {
   type brew &> /dev/null
+  _xit=$?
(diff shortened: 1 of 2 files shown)


T0mK0:
added aws plugin speedup

kenden:
This fix looks a bit less invasive https://github.com/robbyrussell/oh-my-zsh/pull/6113

mikz:
Chruby was fixed in https://github.com/robbyrussell/oh-my-zsh/pull/5562

PR #6083, opened 2017-05-12
Comment about export being optional in dotenv plugin is misleading.

Adaephon-GH:
@arteezy `dotenv` does not automatically set environment variables from the contents of `.env` but only sources it into the current shell sessio. Therefore it may very well be necessary to export variables explicitly in many cases, otherwise they will not actually be available in the environment of subsequently started programs. 

 Stating that `export` is optional may lead to confusion, for example this question on StackOverflow: https://stackoverflow.com/q/43878291/2992551

### DIFF
--- a/plugins/dotenv/README.md
+++ b/plugins/dotenv/README.md
@@ -14,7 +14,7 @@ plugins=(git man dotenv)
 
 ## Usage
 
-Create `.env` file inside your project directory and put your local ENV variables there.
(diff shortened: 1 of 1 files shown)


arteezy:
@Adaephon-GH I must apologize. It was indeed a mistake on my part. When I tested the plugin, `.env` files w/ and w/o `export` worked. I used `PORT` variable, because it was the shortest to type. Both file formats changed the port to different numbers. Now I think I had `PORT` pre-set globally somewhere.  It was already defined, that's why it was working.

I fixed this issue in attached PR. Now the plugin accepts both file formats as was stated. This time I tested it more thoughtfully and carefully, on different machines (macOS and Ubuntu) with respect to global config.

PR #6088, opened 2017-05-13
Added support for Bash on Windows

rinyre:
Automatically will convert commands to their `.exe` equivalent if one exists and the user is running Bash on Windows. This uses the information [provided by Microsoft](https://github.com/Microsoft/BashOnWindows/issues/423#issuecomment-221627364) to determine if the user is running Windows based on `/proc/sys/kernel/osrelease`'s contents.

Bash on Windows will remap any file paths under `/mnt/c` or  their appropriate drive letter to the Windows equivalent, allowing one to even open files using a Windows binary.

### DIFF
--- a/plugins/command-not-found/command-not-found.plugin.zsh
+++ b/plugins/command-not-found/command-not-found.plugin.zsh
@@ -31,3 +31,22 @@ if type brew &> /dev/null; then
     eval "$(brew command-not-found-init)";
   fi
 fi
+
+# Bash on Windows support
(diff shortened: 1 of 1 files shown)


mcornella:
I know this is old but currently zsh already checks if the command ending with .exe exists, when you have `setopt correct` active (which is a default in OMZ).

I'll close this since automatically running a different binary doesn't seem like a sound idea, and also every WSL distribution will have its own command-not-found package.

Issue #6075, opened 2017-05-10
dotenv plugin does not properly escape values


hackel:
Using the dotenv plugin, when a backtick (`) is used in the key value, it does not get escaped, and it actually tries to run the contents as a shell command.  Pretty huge security vulnerability here.
e.g. .env file:
```
KEY=asdf`ls /`fdsa
```
echo $KEY
```
asdfbackup/
bin/
boot/
...etc.
vmlinuz.old@fdsa
```

The plugin is just sourcing the file, but .env files are not shell scripts!

mcornella:
See https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/dotenv/README.md#disclaimer. The user is responsible for what content is in `.env` files.

PR #6093, opened 2017-05-17
Fix dotenv plugin accepted file format and clarify README

### LINKED ISSUE #6074: dotenv does not run when starting a terminal in the dir
It requires that you `cd` into it in order to run.

https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/dotenv/dotenv.plugin.zsh

This is a very common workflow in tmux: splitting windows all within the same dir. `chpwd` is not triggered by this.

arteezy:
FIxes #6083, closes #6074, #6075

This PR fixes a bug reported in #6083, when in most cases `dotenv` plugin ignored lines without `export`. Now it recognises them by temporarily enabling `set -a` option. If this option was enabled before, it stays unmodified:

```sh
if [[ -o a ]]; then
  source .env
else
  set -a
  source .env
  set +a
fi
```
Now the plugin accepts both file formats as was stated.

Also, clarified the README and added disclaimer about plugin's capabilites, while offering alternative solutions, which should close #6074 and #6075 issues.

### DIFF
--- a/plugins/dotenv/README.md
+++ b/plugins/dotenv/README.md
@@ -2,19 +2,19 @@
 
 Automatically load your project ENV variables from `.env` file when you `cd` into project root directory.
 
-Storing configuration in the environment is one of the tenets of a [twelve-factor app](http://www.12factor.net). Anything that is likely to change between deployment environments–such as resource handles for databases or credentials for external services–should be extracted from the code into environment variables.
+Storing configuration in the environment is one of the tenets of a [twelve-factor app](http://www.12factor.net). Anything that is likely to change between deployment environments, such as resource handles for databases or credentials for external services, should be extracted from the code into environment variables.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


Richard-Degenne:
Allow me to bump this PR up. I just spent a fair amount of time battling with my configuration before landing here.

cpitt:
Not sure if bumping works, but bump

vhquang:
I was about to submit the same PR. I am glad I do a check. Please also let me join the vote for this PR.

Richard-Degenne:
Yeah, for some reason, this PR has been open for over a year now, and we haven't had any feedback yet.

mcornella:
Thanks and sorry for the delay. Bumping helped!

vhquang:
@mcornella No problem, thank you for taking care of this. I notice there are over 700 open PR. That is for sure overwhelming.

PR #6101, opened 2017-05-23
Fix clean for tilde home directories

lukeb:
Ensure that wd clean! does not remove warp points to directories with tilde expansion home directories


mcornella:
Fixed in current version.

PR #6103, opened 2017-05-24
Fix autojump plugin for nix

aschmolck:
Under (at least recent versions of) nix the file is ending with .sh not .zsh.


mcornella:
Duplicate of #5163.

PR #6129, opened 2017-06-07
Add support for 'vaulted' shells in agnoster theme

torarvid:
If a user is currently in a [vaulted](https://github.com/miquella/vaulted) shell, they
will now have an indication of this in the agnoster prompt, not unlike the
virtualenv prompt. The foreground color will switch if the timer expires (which
usually signifies that the temporary Amazon keys generated by vaulted will be
invalid).

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -198,6 +198,23 @@ prompt_virtualenv() {
   fi
 }
 
+# vaulted: current vaulted shell
+prompt_vaulted() {
+  if [[ -z $VAULTED_ENV ]]; then
+    return
+  fi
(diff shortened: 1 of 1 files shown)


robbyrussell:

Improvements to agnoster are still welcome, but a permanent segment for a specific third-party tool isn't something we can carry in the default theme — agnoster loads for a very large number of people, and almost none of them run `vaulted`. The prompt also runs `uname`, `sed` and `date` on every redraw while a session is active, which everyone using the theme would pay for.

This works well as a personal customization. Overriding `build_prompt` in `$ZSH_CUSTOM` is a supported setup and it survives updates.

PR #6133, opened 2017-06-11
bgnotify issue - it is no more automatically closing. Adding a timeout

bwlt:
On OSX when using terminal-notifier from julienXX the notification does not automatically close (from about 1.5month).
This fix the problem with a timeout option


benc:
This seems an issue with terminal-notifier: https://github.com/julienXX/terminal-notifier/issues/223

julienXX:
terminal-notifier version 2.0.0 as been released with alerter features removed https://github.com/julienXX/terminal-notifier/releases/tag/2.0.0 so this issue is now fixed if you update.

PR #6136, opened 2017-06-14
add missing features in fox theme

fumseck:
Hi, 
I've added time,current day, git username in git prompt in fox theme, to be better,
and i changed  ☮ by @, because It was not very well displayed in the terminal.

Regards fumseck


fumseck:
capture du 2017-06-14 11-15-34

Issue #6106, opened 2017-05-25
gpg-agent plugin uses unsupported (deprecated) options


Czocher:
Since gpg 2.1 the `--write-env-file` option doesn't do anything:

```
$ gpg-agent --write-env-file
gpg-agent[6532]: WARNING: "--write-env-file" is an obsolete option - it has no effect
```

This causes the gpg-agent plugin to malfunction. It may be better to start the agent as it is defined in the gpg-agent documentation [here](https://www.gnupg.org/documentation/manuals/gnupg/Invoking-GPG_002dAGENT.html) and [here](https://www.gnupg.org/faq/whats-new-in-2.1.html#autostart), which means we'd have to:

1. Call `gpg-connect-agent /bye`
2. Set up the `GPG_TTY` variable: `export GPG_TTY=$(tty)`
3. If `enable-ssh-support` is in the `gpg-agent.conf` config file then set up the `SSH_AUTH_SOCK` variable: `export SSH_AUTH_SOCK="${XDG_RUNTIME_DIR}/gnupg/S.gpg-agent.ssh"`

This would probably simplify the plugin and allow it to work with new versions of gpg-agent, but it would also mean loosing backward compatibility.

I can create a patch for this issue and fix the plugin but I need to know if backward compatibility is a requirement or not?

What do you think @robbyrussell ?

felixsanz:
It doesn't malfunction, it just doesn't work :/

Czocher:
It works for the people who have older versions of gpg installed ;).

felixsanz:
Well, afaik he was talking about gpg 2.1, so i did.

simplycycling:
I'm not sure if my issue is related or not, but gpg-agent simply wasn't working at all for me, with the current gpg-agent plugin in oh-my-zsh, so I put the following into my .zshrc, below the plugins line, and now it works as I would expect (meaning ssh-add -L shows my ssh key, and if I need to swap out for a different Yubikey, I'm able to bounce gpg-agent without the "a gpg-agent is already running" message):

    # Set SSH to use gpg-agent
    unset SSH_AGENT_PID
    if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
      export SSH_AUTH_SOCK="/run/user/$UID/gnupg/S.gpg-agent.ssh"
    fi

Obviously, change the UID to whatever you need it to be.

[...]

PR #6140, opened 2017-06-19
Fixed the gpg-agent plugin for new gpg versions

Czocher:
The gpg-agent plugin did not work for gpg versions above or equal to
2.1 because of the `--write-env-file` option deprecation. This new version
works fine and also enables the ssh-agent support only if it is enabled in the
gpg-agent config file.

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -1,41 +1,14 @@
-local GPG_ENV=$HOME/.gnupg/gpg-agent.env
-
-function start_agent_nossh {
-    eval $(/usr/bin/env gpg-agent --quiet --daemon --write-env-file ${GPG_ENV} 2> /dev/null)
-    chmod 600 ${GPG_ENV}
-    export GPG_AGENT_INFO
-}
-
-function start_agent_withssh {
-    eval $(/usr/bin/env gpg-agent --quiet --daemon --enable-ssh-support --write-env-file ${GPG_ENV} 2> /dev/null)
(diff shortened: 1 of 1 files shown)


Czocher:
Fixes #6106

simplycycling:
Tested and working on Arch Linux, using GnuPG v2.1.21. Thanks @Czocher - great work!

mickaelperrin:
Warning if GNUPGCONFIG file doesn't exist it throws a warning on the grep command.

You should test the existence of the file before grepping:

    if [ -f "$GNUPGCONFIG" ] && grep -q enable-ssh-support "$GNUPGCONFIG"; then

Czocher:
Fixed the issue, thanks @mickaelperrin!

Shea690901:
Er...
Shouldn´t it better read:
`if [ -r "$GNUPGCONFIG" ] ...`
instead of
`if [ -f "$GNUPGCONFIG" ] ...`?

Czocher:
Thanks @Shea690901, changed accordingly to your suggestion.

mzanibelli:
There is a problem still.

`GNUPGCONFIG=${GNUPGHOME:-"$HOME/.gnupg/gpg-agent.conf"}`

Should be : 

`GNUPGCONFIG="${GNUPGHOME:-"$HOME/.gnupg"}/gpg-agent.conf"`

Otherwise we get an error saying that `$GNUPGCONFIG` is a directory when the environment variable is set.

Czocher:
Thanks @mzanibelli! I've created a new pull request with your fix, see #6403.

PR #6142, opened 2017-06-20
optimize load of pyenv with homebrew

LukeXuan:
This pull request attempts to optimize the load of pyenv with homebrew installed. As `homebrew --prefix pyenv` is a quite slow command(approx. 0.3s on my laptop). It would be better use the command less. Thus I added it to the last, and append the default homebrew pyenv installation path. As well as I eliminated the triple check into one check for the sake of speed.

I've tested for several cases. Looking forward to your response.


robbyrussell:

Cheers,
Robby

PR #6148, opened 2017-06-23
Support for the racket tools: raco and racket

catull:
Add support for Racket language tools `raco` and `racket`.


mcornella:
Has no interest.

catull:
@mcornella Too bad, can you at least state your reasoning ?
This PR is almost 2 years now.
What is your point of criticism ?
"Has no interest." is not helpful.

mcornella:
Meaning there are no user +1s or comments. It makes no sense to merge it if no users will use it.

catull:
How about now ?

PR #6165, opened 2017-06-29
pyenv plugin refactor (8x faster)

### LINKED ISSUE #4674: Conflict between the pyenv and oh-my-zsh's pyenv-plugin
Pyenv-plugin in oh-my-zsh did someting for pyenv, and make it has a very weird path. I add this problem in [pyenv 20151105 can't load "~/.pyenv"](https://github.com/yyuu/pyenv/issues/476).

### LINKED ISSUE #6017: pyenv plugin is slow on startup
I had issues with slow start of zsh. I looked into #5327. Specifically I found https://github.com/robbyrussell/oh-my-zsh/issues/5327#issuecomment-241226195 and tried `zsh -xv` and discovered that

    +_pyenv-from-homebrew-installed:1> brew --prefix pyenv

was very slow. Disabling `pyenv` plugin made starting my terminal much faster.

Expected: That `pyenv` is faster to start.

cyphus:
The current version of the pyenv plugin searches for pyenv using a predefined list of paths. If found, the plugin adds the pyenv bin directory to the user's `PATH` variable, and finally runs pyenv to load it into the environment. This is backwards. No matter which pyenv installation method you use, the `pyenv` utility will already be in your path.

I've refactored the plugin to simply load pyenv directly. It will also load pyenv-virtualenv if it's available, as in the original version. If pyenv is not in the user's path, it gracefully falls back to displaying the system-installed python version in the prompt.

#### Speed Increase

After refactoring, I noticed this made the plugin much faster. Let's profile it (with python of course):

profiler command: `$ python -m timeit "__import__('subprocess').call(['zsh', '/Users/cyphus/.oh-my-zsh/plugins/pyenv/pyenv.plugin.zsh'])"`

<dl>
  <dt>Original:</dt>
  <dd>10 loops, best of 3: <b>822 msec</b> per loop</dd>

  <dt>Refactored:</dt>
  <dd>10 loops, best of 3: <b>103 msec</b> per loop.</dd>
</dl>

This PR makes the pyenv plugin load 8x faster, while fixing underlying issues. Winner winner 🍗 dinner.

#### Issues closed by this PR

* Close #6142 
* Close #5884 
* Close #6139 
* Close #6017 
* Close #4674 (Note: Possibly. This looks to be an issue of stemming from initializing two separate pyenv installations into the same shell. The first one is initialized based on the user's `PATH`, and the second is initialized based on the plugin's search order. This PR would change the pyenv plugin to also initialize based on the user's `PATH`, it would initialize the same pyenv. I can't say for sure without recreating the user's issue.)

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -1,9 +1,9 @@
 # This plugin loads pyenv into the current shell and provides prompt info via
 # the 'pyenv_prompt_info' function. Also loads pyenv-virtualenv if available.
 
-if which pyenv > /dev/null; then
+if (( $+commands[pyenv] )); then
     eval "$(pyenv init - zsh)"
-    if which pyenv-virtualenv-init > /dev/null; then
+    if (( $+commands[pyenv-virtualenv-init] )); then
         eval "$(pyenv virtualenv-init - zsh)"
     fi
     function pyenv_prompt_info() {
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -1,50 +1,17 @@
-_homebrew-installed() {
-    type brew &> /dev/null
-}
+# This plugin loads pyenv into the current shell and provides prompt info via
+# the 'pyenv_prompt_info' function. Also loads pyenv-virtualenv if available.
 
-_pyenv-from-homebrew-installed() {
-    brew --prefix pyenv &> /dev/null
-}
-
-FOUND_PYENV=0
-pyenvdirs=("$HOME/.pyenv" "/usr/local/pyenv" "/opt/pyenv")
-
-for pyenvdir in "${pyenvdirs[@]}" ; do
-    if [ -d $pyenvdir/bin -a $FOUND_PYENV -eq 0 ] ; then
-        FOUND_PYENV=1
-        export PYENV_ROOT=$pyenvdir
-        export PATH=${pyenvdir}/bin:$PATH
-        eval "$(pyenv init - zsh)"
-
-        if pyenv commands | command grep -q virtualenv-init; then
-            eval "$(pyenv virtualenv-init - zsh)"
-        fi
-
-        function pyenv_prompt_info() {
-            echo "$(pyenv version-name)"
-        }
+if which pyenv > /dev/null; then
+    eval "$(pyenv init - zsh)"
+    if which pyenv-virtualenv-init > /dev/null; then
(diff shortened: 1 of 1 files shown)


brandonb927:
Here to add my $0.02: The default-provided pyenv plugin (for some reason or another) only insists on using the `/opt/...` path for looking for pyenv version installations. The PR here as-is works perfectly and I'll be using this one in place of the default one. Thanks @cyphus!

Victor-Savu:
Any progress on merging this? Anything that would make pyenv faster (while keeping functionality intact) is a big gain!

cyphus:
Thanks for the support on this.

I read about oh-my-zsh's plugin maintainers in the contributing docs. I'd like to offer to be the official maintainer for the pyenv plugin. As far as my qualifications, well, I read a ton of pyenv documentation and source code when developing these changes, and I may as well continue to put that knowledge to good use. Other than that, hopefully the code speaks for itself.

FrederickGeek8:
I really think this should be merged as soon as possible. I believe that its the only pyenv-related PR that really addresses the root problem with the default pyenv plugin. Not only does the current plugin use the Homebrew installation directory to install versions into (by setting `PYENV_ROOT` to this directory, which of course gets overwritten on an upgrade, destroying everything that you've done with pyenv), but there is, of course, the egregious performance issues that you outlined. The choice that it makes of setting `PYENV_ROOT` to be the Homebrew installation directory is not only a _terrible_ idea (I'm so devistated that I lost all my packages (custom and otherwise) when I upgraded pyenv), but it goes against the settings [recommended by the pyenv package itself](https://github.com/pyenv/pyenv#basic-github-checkout)!

Honestly, it sort of beyond me that this has been overlooked for this long. I now, of course, am using this pull request's version of the plugin, but this is something that I think should really be changed as soon as possible to avoid people losing their Python packages.

If this were to be merged, I think that it would be a great idea to include a note in the upgrade process to move the `versions` from the Homebrew install directory typically (`/usr/local/opt/pyenv/versions/`) to the pyenv user directory (typically `~/.pyenv/versions/`).

mcornella:
Thanks @cyphus!

Kriechi:
Thanks for merging this!
Out of curiosity, would this also work for the rbenv plugin? https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/rbenv/rbenv.plugin.zsh

ysolis:
which advantage have now the plugin if i need to define my $PATH BEFORE the plugin install ? the normal way is redefine your $PATH in the `# User configuration` AFTER the plugin activation in your `~/.zshrc` file

PR #6166, opened 2017-06-30
add grc plugin

akurach:
add grc plugin


mcornella:
Has no interest.

yeKcim:
@mcornella, I use this grc code to color `ps aux` (for example), why did you said that this plugin has no interest?

Issue #6118, opened 2017-05-31
OSX Plugin Spotify Search Broken


robertTechJocks:
**How to reproduce the problem:**
Search for any artist/album/song using the plugin

**What the correct behavior should be**
Spotify will find a match and start playing the match

**What the actual behavior is**
Spotify returns with "No results when searching for X"

xshotD:
This problem is most likely due to Spotify's breaking change of access. Any requests to their API now require a access token.

hnarayanan:
There is is an updated version of the project from where the original spotify plugin was derived from: https://github.com/hnarayanan/shpotify

This now has proper support for access tokens. It can be ported over to this project if there is interest.

PR #6170, opened 2017-07-01
Remove broken spotify features

nortti:
Spotify web api features require an access token now, so they don't work. See #6118.

Authorization should probably happen through the ohmyz.sh domain, so this this can't be fixed by just anyone. Until then, these features should probably just be removed.


hnarayanan:
Sorry, but FYI, this PR is obsolete. Please send patches to the parent project https://github.com/hnarayanan/shpotify in the future. The maintainers of the parent project take the responsibility of submitting stable versions of the script back into oh-my-zsh.

PR #6179, opened 2017-07-06
Create Fasd cache file in user home directory

EgidioCaprino:
If oh-my-zsh is installed globally, `$ZSH_CACHE_DIR`  may point to a directory which is not writable by the user or which is shared amog all the users. For example, in Arch Linux, it is pointing to `/usr/share/oh-my-zsh`.

This fix will create Fasd cache file as hidden in his home directory.


Cypher1:
Should this be implemented by changing the default for ZSH_CACHE_DIR to "~/.fasd-init-cache" so that users overwriting ZSH_CACHE_DIR aren't effected?

Thanks

EgidioCaprino:
I think that would be even better.

PR #6184, opened 2017-07-13
Update directories.zsh

yantaozhao:
Make `list` command looks sane.

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -24,10 +24,11 @@ alias rd=rmdir
 alias d='dirs -v | head -10'
 
 # List directory contents
-alias lsa='ls -lah'
-alias l='ls -lah'
+# alias lsa='ls -lah'
+alias l='ls'
 alias ll='ls -lh'
-alias la='ls -lAh'
+alias la='ls -Ah'
+alias lla='ls -lah'
 
 # Push and pop directories on directory stack
 alias pu='pushd'


lib/directories.zsh
```diff
@@ -24,10 +24,11 @@ alias rd=rmdir
...
 
 # List directory contents
-alias lsa='ls -lah'
-alias l='ls -lah'
+# alias lsa='ls -lah'
+alias l='ls'
 alias ll='ls -lh'
-alias la='ls -lAh'
+alias la='ls -Ah'
+alias lla='ls -lah'
```
Cypher1: Why change the alias for this from lsa?

lib/directories.zsh
```diff
@@ -24,10 +24,11 @@ alias rd=rmdir
 alias d='dirs -v | head -10'
 
 # List directory contents
-alias lsa='ls -lah'
-alias l='ls -lah'
+# alias lsa='ls -lah'
+alias l='ls'
```
Cypher1: This now doesn't show .dotfiles, I don't think it should be changed.

yantaozhao: `ls` is the abbreviation of *list*, and `l` is again the abbreviation of `ls`.
then, `la` indicates `ls -a`, `ll` indicates `ls -l`, and `lla` (or `lal`) indicates `ls -l -a`.

This derivation looks sane. 😄

Cypher1: Im not convinced that it's a big enough gain to justify breaking user's aliases. It is fairly sane though

lib/directories.zsh
yantaozhao: Please see previous comment.
`lla` and `lsa`, which is clear to know it's result is for `ls -l -a` ?
Meanwhile, `lsa` looks like `ls -a` but not `ls -l -a`. Isn't it?

lib/directories.zsh
yantaozhao: I don't think the persistence of old `ls` aliases (which looks a bit of messy) is reasonable, since the oh-my-zsh project doesn't reach the final version now.
And, some popular Linux distribution has already aliased the `ls` to `l` , `la` to `ls -a`, such as the Ubuntu.

Cypher1:
Okay, if there is a precedence I'm all for it.
Cheers

lib/directories.zsh
Cypher1: Okay, if there is a precedence I'm all for it.
Cheers

mcornella:
Breaking users' aliases is a no-no. Thanks though

PR #6185, opened 2017-07-13
shell prompt: display `$` if common user or `#` if root

yantaozhao:
shell prompt: display `$` if common user or `#` if root


mcornella:
This is a significant enough change that it warrants a new theme, please see the [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes) section for how you can add it as a custom theme.

PR #6187, opened 2017-07-13
Restore $IFS after using completion

danielbachhuber:
See https://github.com/wp-cli/wp-cli/issues/1919#issuecomment-120665900


danielbachhuber:
It doesn't look like there's much activity on this project now, but feel free to re-open if you'd ever like to merge this.

PR #5935, opened 2017-03-03
Fix Standard Error Redirection for composer plugin

jrisebor:
The standard error redirection was redirecting the output from export instead of the output from composer. This would print the message `Changed current directory to /home/username/.composer` for every new shell.
Moving the redirection inside the parentheses fixes this.

PR #6189, opened 2017-07-13
Silent composer output when setting PATH.

Fisiu:
Signed-off-by: Mariusz Fik <mariusz@fidano.pl>


SunMar:
This is a duplicate of #5935

PR #6196, opened 2017-07-18
[systemd] Added commands for system control

schtobia:
It would be very nice to have these systemctl command shortcuts, since they are now the desired way to manipulate the running state of the system. 

These commands are user verified by systemd itself, so we don't need to
`sudo` them.

@aelesbao, any comments on this?


schtobia:
So, how do we get this changeset into master?

PR #6203, opened 2017-07-25
Set color to red if root in gallifrey theme

awkwardbunny:


Marcholio:
I think that this should be either merged or closed as there hasn't been any activity for years.

PR #6204, opened 2017-07-26
Added some alias only for Laravel framework

defji:
for laravel magicians


Cypher1:
Can you update the name of this PR to reflect that the aliases are only for laravel?

Thanks

PR #6209, opened 2017-07-28
plugins/emacs: depend on emacs 25, not emacs 24

0xabu:
The emacsclient.sh script invokes emacsclient with --alternate-editor '', which is not supported by Emacs v24. (At least, not by Emacs 24.5 on Windows, but I doubt this is platform-specific.)

This may be the cause of issue #3305. The alternative fix, if you want to keep support for Emacs 24, is to remove the --alternate-editor '' from emacsclient.sh, but that looks like useful functionality.

### DIFF
--- a/plugins/emacs/emacs.plugin.zsh
+++ b/plugins/emacs/emacs.plugin.zsh
@@ -10,7 +10,7 @@
 # - Configuration changes made at runtime are applied to all frames.
 
 
-if "$ZSH/tools/require_tool.sh" emacs 24 2>/dev/null ; then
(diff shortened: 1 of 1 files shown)


mcornella:
`--alternate-editor` is supported in v24 (tested in 24.1 in macOS). The problem in #3305 is probably that no Emacs GUI is installed, it is unrelated to `--alternate-editor`.

In emacs 24.5 under Windows, does it automatically start the `emacs` daemon if not passed `--alternate-editor`? If not, I see no reason to support it under Windows.

PR #6210, opened 2017-07-28
agnoster: cut down on fork/execs improve performance when not in a repo

0xabu:
This PR improves performance, especially when we're not in a git repo, by cutting down the number of forks/execs needed just to render the prompt. It's a particularly noticeable speedup on Cygwin.

There are three sub-commits here for three distinct performance bugs; any of them could be applied independently if you don't like some of the others.

I've also submitted similar changes upstream as https://github.com/agnoster/agnoster-zsh-theme/pull/44

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -94,9 +94,9 @@ prompt_git() {
     PL_BRANCH_CHAR=$'\ue0a0'         # 
   }
   local ref dirty mode repo_path
-  repo_path=$(git rev-parse --git-dir 2>/dev/null)
 
   if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
+    repo_path=$(git rev-parse --git-dir 2>/dev/null)
     dirty=$(parse_git_dirty)
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git rev-parse --short HEAD 2> /dev/null)"
     if [[ -n $dirty ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
The `echo`s are not fork/exec's but `write()` (or `print()` which is basically the same). So using `PROMPT+=...` is not needed. It also creates a weird bug that makes my prompt be displayed one character to the right, leaving a space before it, until I restart the session with `exec zsh`. I haven't been able to find the culprit:

image

The downside of reverting that commit is that it makes it so that your `$jobstates` performance improvement doesn't return correct results. I suppose in PROMPT land we cannot access that variable.

themes/agnoster.zsh-theme
```diff
@@ -55,22 +56,22 @@ prompt_segment() {
...
   if [[ -n $CURRENT_BG ]]; then
-    echo -n " %{%k%F{$CURRENT_BG}%}$SEGMENT_SEPARATOR"
+    PROMPT+=" %{%k%F{$CURRENT_BG}%}$SEGMENT_SEPARATOR"
   else
-    echo -n "%{%k%}"
+    PROMPT+="%{%k%}"
   fi
-  echo -n "%{%f%}"
+  PROMPT+="%{%f%}"
   CURRENT_BG=''
```
mcornella: I found the bug: this should be `CURRENT_BG='NONE'` so that next prompt begins with the right background.

0xabu:
@mcornella re. "The echos are not fork/exec's but write()" -- I realise that, but you still do need a subshell (which is one fork/exec) to capture the output of the echo.

I will make the change you suggested -- thanks for catching this!

0xabu:
Gentle ping @mcornella -- I believe the last update addressed all of your PR comments, and I also tested with Mercurial.

mcornella:
Thanks!

PR #6213, opened 2017-07-29
Add wording consistency for update/upgrade controversy

lessthanzero:
Change main command for updating to “update_oh_my_zsh”. Seems to be
more conventional and human-readable. Some internal stuff still keeps
“upgrade”. Needs thorough testing. And maybe a transitional message for
existing users


mcornella:
Too big a change, I'll reserve this for an upgrade system overhaul.

PR #6220, opened 2017-08-02
Update: git.plugin.zsh --force => --force-with-lease.

aladdin-add:
https://developer.atlassian.com/blog/2015/04/force-with-lease/


mcornella:
Fixed in master

PR #6223, opened 2017-08-03
screen-ssh-agent plugin added; to enable the ssh-agent within screen

bartdorlandt:
A small script to enable the ssh-agent also within screen.


mcornella:
This should not be necessary, the ssh-agent plugin works everywhere. Reopen with an example of why it doesn't within screen, if there is one. Thanks!

PR #6230, opened 2017-08-07
Simple completions for stack exec and stack test

Cypher1:
Currently completion for haskell-stack is not aware of the local project files and so does not complete for commands like `stack exec <TAB>` or `stack time <TAB>`. I'm hoping to implement these completions.

This is my first attempt at a fix and a PR. I'd love constructive feedback on how to implement this.

I'm happy to wait for or merge in https://github.com/robbyrussell/oh-my-zsh/pull/5898 which updates descriptions and the sub-command completions to be up to date with the latest version of stack.

Alternatively we could stop creating our own stack completion and use stack's internal completion. This would come at the expense of not having neat command descriptions or being able to diverge from stack's own completions but would allow for better support in the future.

The following lines allow stack to do its own completion (which is pretty powerful as of Stack 1.5).
```
autoload -U +X compinit && compinit
eval "$(stack --bash-completion-script stack)"
```


Cypher1:
I've been using it recently and have found the native stack competitions to be more useful than our hard-coded versions (even with the helpful descriptions). As such I've updated the PR to use the internal stack completions.

I imagine that the backlog is quite huge at the moment. Thanks for considering this @robbyrussell when you have time.

g1eny0ung:
@Cypher1 Good job 👍 . But seems that our prs #5898 have already been flooded 😂 .

Cypher1:
Thanks, should have referenced yours here.

mcornella:
Sorry for the delay, folks.

g1eny0ung:
Never mind, glad this pr has finally been merged.

PR #6234, opened 2017-08-08
Fix git prompt in luke randall theme

aj-jaswanth:


robbyrussell:
We let this one sit for a long time, I'm going to nix this as nobody else has inquired about it.

Issue #6237, opened 2017-08-10
Some parse error?!


kaymio:
It doesn't matter which (virtual) TTY I'm opening I always get these as the first lines:

/home/X/.oh-my-zsh/plugins/colored-man-pages/colored-man-pages.plugin.zsh:19: defining function based on alias `man'
/home/X/.oh-my-zsh/plugins/colored-man-pages/colored-man-pages.plugin.zsh:19: parse error near `()'
/home/X/.oh-my-zsh/plugins/colored-man-pages/colored-man-pages.plugin.zsh:19: defining function based on alias `man'
/home/X/.oh-my-zsh/plugins/colored-man-pages/colored-man-pages.plugin.zsh:19: parse error near `()'

but everything is working.

terencode:
Having the same problem on Arch Linux after upgrading to zsh 5.4.1

toto-42:
Wouldn't it be better to change "man ()" to "function man()" in 'colored-man-pages.plugin.zsh'?

IHMO this is better than to force users to change back globally to an old, non default behavior considered error prone by the zsh development team.

marcelsiegert:
@toto-42 Yep, you're probably right. I updated my pull request.

kaymio:
When can we expect a fix for this? Or could you provide a quick fix?

terencode:
@kaymio: navigate to *usr/share/oh-my-zsh/plugins/colored-man-pages* then edit *colored-man-pages.plugin.zsh* by adding "function" before "man()" and you're done.
Credits to @marcelsiegert .

kaymio:
Thank you very much @marcelsiegert !

Zerophase:
Shouldn't the `()` be removed from the function. According to this it's only supported syntax from a bug. https://unix.stackexchange.com/questions/73750/difference-between-function-foo-and-foo

PR #6238, opened 2017-08-10
Prepend function keyword in colored-man-pages plugin

marcelsiegert:
Resolves: #6237


kiemlicz:
What is the status here?
This is veeery annoying...

PR #6240, opened 2017-08-11
composer: Fix bin directory when Composer is not available

RobLoach:
When [Composer](https://getcomposer.org/) isn't available, the `composer global` call results in an error. This pull request checks to see if Composer is available before making the call. When Composer isn't available, it will just manually set the directories.


plugins/composer/composer.plugin.zsh
```diff
@@ -51,5 +51,10 @@ alias cgrm='composer global remove'
 # install composer in the current directory
 alias cget='curl -s https://getcomposer.org/installer | php'
 
-# Add Composer's global binaries to PATH
-export PATH=$PATH:$(composer global config bin-dir --absolute) 2>/dev/null
+# Add Composer's global binaries to PATH, using Composer if available.
+if hash composer 2>/dev/null; then
+  export PATH=$PATH:$(composer global config bin-dir --absolute) 2>/dev/null
```
ricpelo: The `2>/dev/null` part should be *inside* the parenthesis to remove an unwanted info message, like this:

`export PATH=$PATH:$(composer global config bin-dir --absolute 2>/dev/null)`

RobLoach: This line is the same as it was. Should we be correcting that line in this PR?

RobLoach: Fixed it as part of this.

plugins/composer/composer.plugin.zsh
```diff
@@ -53,7 +53,7 @@ alias cget='curl -s https://getcomposer.org/installer | php'
 
 # Add Composer's global binaries to PATH, using Composer if available.
 if hash composer 2>/dev/null; then
-  export PATH=$PATH:$(composer global config bin-dir --absolute) 2>/dev/null
+  export PATH=$PATH:$(composer global config bin-dir --absolute 2>/dev/null)
 else
   export PATH=$PATH:~/.composer/vendor/bin
```
ricpelo: IMHO, it could be better if you do the following:

```
[ -d ~/.composer/vendor/bin ] && export PATH=$PATH:~/.composer/vendor/bin
[ -d ~/.config/composer/vendor/bin] && export PATH=$PATH:~/.config/composer/vendor/bin
```

With the above, a path is added to the PATH env variable only if that path exists.

Even better, we could do this:

```
if [ -d ~/.config/composer/vendor/bin ]; then
    D1=~/.config/composer/vendor/bin
    if [ -L $D1 ]; then
        L1=$D1
    fi
fi
if [ -d ~/.composer/vendor/bin ]; then
    D2=~/.composer/vendor/bin
    if [ -L $D2 ]; then
        L2=$D2
    fi
fi

if [ -n "$D1" ] || [ -n "$D2" ]; then
    if [ "$(readlink -f $D1)" != "$(readlink -f $D2)" ]; then
        if [ -n "$L1" ]; then
            export PATH=$PATH:$D2
        else
            export PATH=$PATH:$D1
        fi
    else
        if [ -n "$D1" ]; then
            export PATH=$PATH:$D1
        else
            export PATH=$PATH:$D2
        fi
    fi
fi
```

This way, we only add the a directory to the PATH if this directory exists and it's not a symlink to the other directory.

RobLoach:
@ricpelo Went with the first option. Seems like the simplest stop-gap. Thanks a lot :+1:

mcornella:
Thanks to both!

PR #6241, opened 2017-08-11
Add docker-compose start alias

VictorDenisov:


VictorDenisov:
@robbyrussell How long does it take to accept a pr? What can I do to accelerate it? This one looks like a very simple pr.

PR #6243, opened 2017-08-13
add termux commands to clipboard.zsh

emanresusername:
less-reinvent-the-wheel-y sequel to #6058

### DIFF
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -30,6 +30,17 @@ function clipcopy() {
     else
       cat $file > /dev/clipboard
     fi
+  elif [[ $OSTYPE == linux-android ]]; then
+    if (( $+commands[termux-clipboard-set] )); then
+      if [[ -z $file ]]; then
+        termux-clipboard-set
+      else
+        cat "$file" | termux-clipboard-set
+      fi
+    else
+      print "clipcopy: termux-api not installed: https://github.com/termux/termux-api-package" >&2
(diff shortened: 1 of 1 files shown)


lib/clipboard.zsh
```diff
@@ -30,6 +30,17 @@ function clipcopy() {
...
     fi
+  elif [[ $OSTYPE == linux-android ]]; then
+    if (( $+commands[termux-clipboard-set] )); then
+      if [[ -z $file ]]; then
+        termux-clipboard-set
+      else
+        cat "$file" | termux-clipboard-set
+      fi
+    else
+      print "clipcopy: termux-api not installed: https://github.com/termux/termux-api-package" >&2
```
mcornella: Link should be https://play.google.com/store/apps/details?id=com.termux.api

mcornella: Maybe https://wiki.termux.com/wiki/Termux:API , since the `termux-api` package is also needed and it's clearer reading it in that wiki.

lib/clipboard.zsh
```diff
@@ -30,6 +30,17 @@ function clipcopy() {
     else
       cat $file > /dev/clipboard
     fi
+  elif [[ $OSTYPE == linux-android ]]; then
```
mcornella: In my device, `$OSTYPE` shows `linux-androideabi`. Using `linux-android*` instead will also match this.

PR #6244, opened 2017-08-14
Add LXC autocompletion

woshilapin:
Hi, couldn't find any lxc related topic for completion in existing issues and pull requests.  I hope this can help!

### DIFF
--- a/plugins/lxc/README.md
+++ b/plugins/lxc/README.md
@@ -1,0 +1,11 @@
+# lxc
+
+This plugin adds completion for the Linux Containers CLI tool.
+
+To use it, add `lxc` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... lxc)
+```
+
+Updated on August 14th, 2017.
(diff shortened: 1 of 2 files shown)


okapia:
Looking at the upstream lxc sources on github, they include a bash completion in lxc/config/bash. If they're willing to include a bash completion, then I don't see why they couldn't also include a zsh one so you could send a pull request there instead. That has the advantage of potentially reaching all zsh users and not just oh-my-zsh users (e.g. someone such as myself - I have nothing todo with oh-my-zsh but was looking through the PRs).

As far as the function goes, I'd suggest declaring the arrays local and reset them every time. So instead of:
```
    _lxc_common_opts=( ... )
    _pkg() {
    ...
    }
    _pkg "$@"
```
Just do:
```
    local -a common
     common=( ... )
    ...
```
Also, the zsh convention is not to capitalize descriptions. Try to start descriptions with a verb so instead of [Name of the container], use [specify name of container]. Eliding "the" makes it shorter. Dont instead of Don't is ugly however, use double quotes outside: "[don't ...]"

For most commands, --help can't be combined with any other options so intead of:
```
    '(-h --help)'{-h,--help}'[Help list]'
```
use
```
    '(-)'{-h,--help}'[display usage information]'
```

woshilapin:
Just cleaning up some of my old PR. I won't pursue this (I don't really have the time and have no interest anymore). I'm closing it.

PR #6245, opened 2017-08-15
Update update_terminalapp_cwd to urlencode the hostname

sheuer:
Apple's Terminal doesn't open a new tab in your current directory if your hostname has UTF-8 characters in it. Percent encoding the host in addition to the path in update_terminalapp_cwd appears to solve this issue.


mcornella:
I believe this works correctly for non-UTF8 hostnames, and I trust your word that it fixes it for UTF-8 hostnames.

PR #6246, opened 2017-08-15
Update Yarn plugin alias

Willibaur:
With this PR we will have alias for major commands in Yarn plugin


roymiloh:
Hi @robbyrussell, `yo` is a popular command among [yeoman](http://yeoman.io/) users, and this alias overrides it. Any possibility to change it?

PR #6247, opened 2017-08-16
Add alias for force git push

ylogx:


contyk:
Eh, I'd rather not see this included -- force pushes should be a conscious action; having three letter aliases for it is just asking for trouble.

PR #6249, opened 2017-08-16
Add kops completion loading to kubectl plugin

MeanderingCode:


kenden:
Thanks for the PR.
Does it not make more sense to enable the kops plugin for this separately?
https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/kops/kops.plugin.zsh

PR #6251, opened 2017-08-16
Added percol plugin

robturtle:
It provides two functionalities that allow you search history and resume
background jobs with interactively incremental searching utility powered by
[Percol](https://github.com/mooz/percol).


robturtle:
created by mistake

PR #6252, opened 2017-08-16
Bundler plugin: add hanami to list of ruby commands

davydovanton:
Hello,
I added [hanami](http://hanamirb.org) command to list of bundler commands


7even:
@davydovanton sorry, an exact duplicate of this got merged in #6334 :)

PR #6253, opened 2017-08-17
systemadmin: correct sort order for psmem aliases

Cypher1:
Currently psmem[10] sorts in the opposite order to pscpu[10] and as such shows the 10 lowest memory users (and returns results in the opposite order).

This PR changes psmem10 and psmem to sort in the same order as pscpu10 and pscpu (which makes psmem10 show the top 10 rather than the bottom).

Should be a tiny bug fix that doesn't cause anyone problems


Cypher1:
@mcornella Any chance this could get merged?

mcornella:
LGTM. Can you also change the README?

Cypher1:
Thanks @mcornella when I wrote this there was no README. I've rebased and it seems that the README was technically incorrect (as the sort orders listed for the two commands is the same). In that case, I think this is no longer just an improvement to the consistency of the commands but a bug fix so that the behavior matches that stated in the README.

Please let me know if I've missed anything else that needs to be updated.

Cypher1:
Thanks @mcornella

PR #6259, opened 2017-08-21
Fixed typo in #5869

jarhat:


qwIvan:
The original already had a similar PR..
#6509

PR #6264, opened 2017-08-23
Fix 'wd show' for directories within $HOME

guywithnose:


mcornella:
Please post this PR to the [official repo](https://github.com/mfaerevaag/wd) and we'll update the plugin once a new version is released.

PR #6267, opened 2017-08-24
Add new git alias

greenchapter:
I added a new git alias for to diff changes from unstaged to HEAD


mcornella:
This is the git diff default.

PR #6269, opened 2017-08-27
Add org-mode-manager

mtekman:
This is an `org-mode` notes manager that simplifies creation/access to org-mode files.


mcornella:
Has no interest.

PR #6272, opened 2017-08-28
Fixed git ahead/behind lib functionality

stevenspasbo:


stevenspasbo:
The `git_commits_ahead` and `git_commits_behind` functions break when a user enters a repository without an upstream due to lines 80 and 90, both of which are like:
```
local commits="$(git rev-list --count @{upstream}..HEAD)"
```

My fix just redirects stderr and then otherwise assume that the output is correct.

stevenspasbo:
@robbyrussell do you think you could take a look at this?  Sorry to bug you but I'm not sure who else to tag.

Also, if you need some help going through old PR's and issues, I'd like to lend a hand.

AndreFlakowski:
I have the same Issue and made the same fix locally.
if used in themes this creates superfluous error outputs.
+1

mcornella:
Merged #6658 instead. Thanks for the PR though!

PR #6273, opened 2017-08-28
Add mkalias

mtekman:
mkalias

An alias management tool that creates/removes aliases and preserves them across shells/boots.


mcornella:
Has no interest.

PR #6275, opened 2017-08-30
recent versions of zsh flag functions declared w/o explicit 'function'…

shane1027:
… specifier as defined based on an alias, causing a parsing error and disabling the function in the shell upon launch.  Explicit 'function' specifier added, and problem is fixed


mcornella:
Fixed a long time ago.

PR #6276, opened 2017-08-31
Git plugin: add aliases for '--(no-)skip-worktree' & showing hidden files

henrebotha:
This adds some aliases for "hiding" files from git (in such a way that they won't be overwritten when you pull remote changes - subtly different to `assume-unchanged`):

* `ghide path/to/file` will hide the file from `git status`, `git add`, etc.
* `gunhide path/to/file` restores the file visibility.
* `gunhidea` restores the file visibility for all hidden files.
* `ghidden` lists all files hidden in this manner.

@ncanceill

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -156,7 +156,7 @@ alias ggsup='git branch --set-upstream-to=origin/$(git_current_branch)'
 alias gpsup='git push --set-upstream origin $(git_current_branch)'
 
 alias ghh='git help'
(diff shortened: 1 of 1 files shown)


henrebotha:
@mcornella @robbyrussell The listed maintainer for the Git plugin hasn't been active on GitHub in over a year, so I defer to you.

mcornella:
These are obscure enough that they won't be useful to most people. You can add them to your zshrc file or as a custom plugin.

PR #6286, opened 2017-09-03
Make prompt end in '$' and newline for clarity

dijonkitchen:
When nested deep in long directories or branches, the prompt is at the end of the terminal, so it's better on a newline by default and always in the same place.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -225,4 +225,4 @@ build_prompt() {
   prompt_end
 }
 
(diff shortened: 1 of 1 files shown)


mcornella:
This is a substantial change to the theme, so I suggest you add that line (`PROMPT=...`) to your zshrc file or copy the agnoster theme to the custom folder and modify it there. See 
You can set it up as a custom plugin, see [Customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization) for more info.

PR #6287, opened 2017-09-03
Add phx support to mix plugin

d6e:
Add support for the new phoenix project generator 'phx'. This project generator was released as part of phoenix 1.3.0.
http://phoenixframework.org/blog/phoenix-1-3-0-released


mcornella:
Already on master.

PR #6289, opened 2017-09-04
Adding whatthecommit to lol plugin

samber:
```
$ curl -s whatthecommit.com/index.txt
So my boss wanted this button ...

$ curl -s whatthecommit.com/index.txt
A long time ago, in a galaxy far far away...

$ curl -s whatthecommit.com/index.txt
Switched off unit test 15 because the build had to go out now and there was no time to fix it properly.
```


robbyrussell:
Hah, nice

PR #6291, opened 2017-09-06
Fix Spotify share url copying

tnwhitwell:
Piping cecho output to pbcopy caused the plugin to copy "[1m[32m-n(B[m" to the clipboard, not the url.

Changing to use 'echo' fixes that.

### DIFF
--- a/plugins/osx/osx.plugin.zsh
+++ b/plugins/osx/osx.plugin.zsh
@@ -533,7 +533,7 @@ function spotify() {
       url=${url#$remove}
       url="http://open.spotify.com/track/$url"
       cecho "Share URL: $url";
-      cecho -n "$url" | pbcopy
+      echo -n "$url" | pbcopy
       break;;
 
       -h|--help| *)


hnarayanan:
Thank you for this. Could you please contribute this as a patch to the upstream project for this script: https://github.com/hnarayanan/shpotify?

tnwhitwell:
@hnarayanan You're already sorted!

```python
            if [ "$2" = "" ]; then
                cecho "Spotify URL: $url"
                cecho "Spotify URI: $uri"
                echo "To copy the URL or URI to your clipboard, use:"
                echo "\`spotify share url\` or"
                echo "\`spotify share uri\` respectively."
            elif [ "$2" = "url" ]; then
                cecho "Spotify URL: $url";
                echo -n $url | pbcopy
            elif [ "$2" = "uri" ]; then
                cecho "Spotify URI: $uri";
                echo -n $uri | pbcopy
```

hnarayanan:
Thank you for verifying this.

tnwhitwell:
Made redundant by #6419

PR #6296, opened 2017-09-11
add k alias to kubectl

yagonobre:


kenden:
This alias is already present:
https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/kubectl/kubectl.plugin.zsh#L10

yagonobre:
nice

PR #6308, opened 2017-09-19
Update git.plugin.zsh

qkdreyer:
Added new alias 'gcae' for 'git commit --allow-empty'


mcornella:
Not used enough. You can just `gc --allow-empty` (autocompletion is very useful here).

PR #6311, opened 2017-09-20
Add robin theme

robinwolny:
Just slight modification of agnoster
- adding timestamp
- icon fixup


stevenspasbo:
New themes are currently [not](https://github.com/robbyrussell/oh-my-zsh/#do-not-send-us-themes) being accepted, but you can add a link to it on the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page.

PR #6325, opened 2017-10-01
fix: Useing 'type' instad of 'which', which is more robust

cLupus:
For reference, see: https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script.


cLupus:
The reason for this, is that when aliasing curl, to have a different default behavior (or use `noglob curl`), then `which` will report an error, thinking that curl does not exist.
Using `type` fixes this.

mcornella:
Fixed in master using `$commands[curl]` instead (idiomatic zsh).

PR #6329, opened 2017-10-04
git plugin: print uniformly https and ssh repositories.

iglesias:
Before this change, this repository info:

```
$ git remote -v
origin  git@github.com:iglesias/oh-my-zsh.git (fetch)
origin  git@github.com:iglesias/oh-my-zsh.git (push)
upstream        https://github.com/robbyrussell/oh-my-zsh.git (fetch)
upstream        https://github.com/robbyrussell/oh-my-zsh.git (push)
```

would be shown as

```
$ current_repository 
iglesias/oh-my-zsh.git (fetch) iglesias/oh-my-zsh.git (push) //github.com/robbyrussell/oh-my-zsh.git (fetch) //github.com/robbyrussell/oh-my-zsh.git (push)
```

This change makes it look like

```
$ current_repository
iglesias/oh-my-zsh.git (fetch) iglesias/oh-my-zsh.git (push) robbyrussell/oh-my-zsh.git (fetch) robbyrussell/oh-my-zsh.git (push)
```

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -18,7 +18,7 @@ function current_repository() {
   if ! $_omz_git_git_cmd rev-parse --is-inside-work-tree &> /dev/null; then
     return
   fi
-  echo $($_omz_git_git_cmd remote -v | cut -d':' -f 2)
+  echo $($_omz_git_git_cmd remote -v | awk 'BEGIN { FS = "github.com[\:\/]" } { print $NF }')
 }
 # Pretty log messages
 function _git_log_prettily(){


robbyrussell:
Hi @iglesias, I don't believe that the one line list of all the remotes is as simple to scan. What are you aiming to accomplish here?

iglesias:
Printing the https and ssh remotes in the same way. Both before and after the change produce a one-line list (this comes from the `echo $(...)` part). Compare the outputs of the first and second calls to current_repository above.

This was tested with zsh 5.1.1 in ubuntu 16.04.

iglesias:
@robbyrussell, any update here? Or maybe somebody else has time to have a look at it, @mcornella?

mcornella:
I really dislike this function, it's been there since 2012 (b1e4ef17543d4a236ad35417ac36560917e6cadc) and it isn't being used by anything else. If a user wants to use, `grv` (`git remove -v`) is a better command. I'll delete it instead unless I can think of a reason not to.

PR #6331, opened 2017-10-04
Add autocompletion for kops

mstuparu:


mcornella:
This is old, so I'll close this. Reopen and add a README if you're still interested.

PR #6332, opened 2017-10-05
Unbreak vi's visual mode

WhyNotHugo:
Vi (and zsh's vi mode) a few modes, one of them being visual mode, which is triggered by pressing <kbd>v</kbd>.

Entering it using <kbd>v</kbd> is the default on vi/vim/nvim, and also on zsh when using `bindkey -v`.

However, the `vi-mode` plugin overrides this with a custom action, effectively making this mode inaccessible and breaking vi mode.

This PR undoes that, since it seems to be a huge oversight when mapping that particular key, and makes vi's visual mode accessible when using the `vi-mode` plugin.

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -17,7 +17,6 @@ bindkey -v
 
 # allow v to edit the command line (standard behaviour)
 autoload -Uz edit-command-line
-bindkey -M vicmd 'v' edit-command-line
 
 # allow ctrl-p, ctrl-n for navigate history (standard behaviour)
 bindkey '^P' up-history


okapia:
I'm not a "tester", I'm a zsh developer (and not an omz user as such) but really please consider merging this PR. Breaking visual mode is really a pity. It was a new feature added three years ago, first appearing in 5.0.8. The plugin probably predates that.

I'd actually suggest that you bind edit-command-line to something else, ^X^E for instance. That doesn't do anything in vim so is unlikely to suffer the same fate as v did. Unfortunately some users may miss v being bound to edit-command-line so having a replacement key will make it easier to respond to any issue raised.

And for a vi plugin, consider enabling Functions/Zle/select-quoted, select-bracketed and surround from the zsh distribution. There's a comment at the top of each describing how.

If you need help, let me know.

WhyNotHugo:
> The plugin probably predates that.

Ah, that's interesting to know, and explains why this made it in in the first place, thanks for pointing it out. 🙂

WhyNotHugo:
Wow

mcornella:
#8004 implemented this same change. Update with `omz update` to see the changes.

PR #6333, opened 2017-10-05
Origin agnostic

simonvpe:
This PR makes the `tools/install.sh` script work for forks. Instead of using a hardcoded url for the origin it uses `git remote get-url` to figure out the remote url. It also checks which branch we're on and installs the same branch instead of master.

Also added support for submodules since they can be nice to have when installing custom themes in a fork.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -61,7 +61,14 @@ main() {
       exit 1
     fi
   fi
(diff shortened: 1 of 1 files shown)


mcornella:
This breaks normall installations. How does it know the remote when there is no git repository yet?

PR #6334, opened 2017-10-05
add hanami to the list of wrapped gems

7even:
This PR adds `hanami` to the list of gems wrapped by the `bundler` plugin. [Hanami](http://hanamirb.org/) is a Ruby web framework.


foxweb:
👍

PR #6340, opened 2017-10-07
Stop error "permission denied: /npm_completion"

### LINKED ISSUE #5874: npm plugin can attempt to write completion to root of filesytem
Where the npm plugin [set's the path to the completion file](https://github.com/robbyrussell/oh-my-zsh/blob/b70842bae4346c793e4932d467dd566baeee7d64/plugins/npm/npm.plugin.zsh#L2) the `ZSH_CACHE_DIR` env var is used as the parent folder:

```shell
__NPM_COMPLETION_FILE="${ZSH_CACHE_DIR}/npm_completion"
```

if `ZSH_CACHE_DIR` isn't set it will fall back to using `/npm_completion` as the full path for the completion file. This will most likely throw a permission error on every new session, e.g:

```
npm.plugin.zsh:5: permission denied: /npm_completion
```

It would be better to fall back to something else (possibly `/tmp`) if `ZSH_CACHE_DIR` isn't set.

AlexxNica:
Fixes #5874

Related:
https://github.com/lukechilds/zsh-nvm/issues/23
https://github.com/zsh-users/antigen/issues/586


henriquecof:
Is there any special setup to use this pluguin with ZSH? ZSH: COMMAND NOT FOUND: NPM

mcornella:
You just need to have npm in your `$PATH`.

henriquecof:
> You just need to have npm in your `$PATH`.

I'm sorry for the question, but should I only have it in the ZSH PATH or in my PATH operating system? I was doubtful about that.

PR #6353, opened 2017-10-13
Export LS_COLORS for linux

mparada:
Problem:
On Linux oh-my-zsh's default coloring of `ls` is not taken into account.

Solution:
The BSD version of `ls` uses a variable called `LSCOLORS` to set the coloring of output of `ls`. The Linux version uses `LS_COLORS` which is not set beautifully as `LSCOLORS` for BSD. So this PR sets `LS_COLORS` (valid for Linux) equal to `LSCOLORS` (valid for Mac), according to this conversion: https://geoff.greer.fm/lscolors/

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -3,6 +3,7 @@ autoload -U colors && colors
 
 # Enable ls colors
 export LSCOLORS="Gxfxcxdxbxegedabagacad"
+export LS_COLORS="di=1;36:ln=35:so=32:pi=33:ex=31:bd=34;46:cd=34;43:su=30;41:sg=30;46:tw=30;42:ow=30;43"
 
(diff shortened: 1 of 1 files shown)


mparada:
@robbyrussell, there is some confusion I believe.
Linux testers are needed, not BSD testers. On BSD the implementation of the `ls` command should be similar to the implementation in OSX (since OSX's implementation is based on BSD).

mcornella:
This will change on users who haven't set up `$LS_COLORS` themselves, since what they end up seeing after OMZ is sourced is the default coloring set by `dircolors -b`.
I agree the color style is nice, but I'm not sure we can merge this without significant pushback.

mparada:
I agree it would break a lot of users setups so I'm closing this PR. I think fix #11091 from @89iuv should be the way to go!

PR #6356, opened 2017-10-13
Fix for #6299: restore previous default shell with uninstall

zwhitten:


mcornella:
Hey @zwhitten thanks for this fix. I pulled the changes in #5169 which is now ready for testing. If you want to help speed up the testing and merge (which should happen early next week), you can wget or curl it from the URL `https://raw.githubusercontent.com/mcornella/oh-my-zsh/refactor-installer/tools/install.sh`. Read the comments at the top of the script to know how to call it.

Thanks!

mcornella:
Closed by #5169.

PR #6360, opened 2017-10-17
Renamed ggsup to gbsup

qkdreyer:
Since GPSUP acronym stands for "Git Push Set UpStream", I guess alias name for "Git Branch Set UpStream" should be named GBSUP, not GGSUP.


qkdreyer:
Ping @mcornella

mcornella:
Changing it at this stage will unnecessarily break people's workflow. Thanks for the ping though

mcornella:
See https://github.com/robbyrussell/oh-my-zsh/issues/5721. You can add your alias to your zshrc file.

PR #6365, opened 2017-10-21
Add nixos support for autojump plugin

JWZH:
As titled.


robbyrussell:
this file could use some DRYing up... in the meantime.. thanks!

PR #6366, opened 2017-10-22
Add shortcut for opening current branch in Jira

jradtilbrook:
@tresni I'd like to add a `br` shortcut/command to the jira plugin that will allow you to open the Jira ticket named after the git branch of the current directory. This works well for projects that name branches after ticket numbers and saves the user from having to type out the whole ticket reference.  
I doubt it would clash with any existing workflows in use because it is such a short command and it doesn't fit the regex for Jira tickets.  
Let me know what you think - I was already using an alias setup to achieve this but thought it might be useful to others.


tresni:
lgtm

PR #6367, opened 2017-10-22
Allow jira default action file based setting

jradtilbrook:
@tresni This ties in with my other PR #6366 to allow file based configuration of the default action. I like the file based config of the other settings and I can save a few lines from my zshrc and shave some environment variables.


tresni:
lgtm

PR #6370, opened 2017-10-25
add -backend-config flag to terraform init command

madpawel:
-backend-config flag was introduced in terraform 0.9.1 version


robbyrussell:
<3

PR #6372, opened 2017-10-25
Check first for bin/stubs directory in _rails_command and _rake_command

davidlibrera:
I've started to use stubbed bin's for rails application in order to avoid the use of bundle exec.
With this, it checks in bin/stubs directory before bin/


robbyrussell:

Cheers,
Robby

PR #6376, opened 2017-10-29
Enhanced Laravel 5 Plugin

sschlein:
This simple enhancement adds the colon if tab is pressed.

Previously, you could type `php artisan m` and hit tab. This resulted in `php artisan make`. To get to the end of the command, it was required to type ":mo" to get to `php artisan make:model`

The change in this PR automatically adds the colon and typing `php artisan ma` results in `php artisan make:`.  While looking like a very small change, it results in much faster autocompletition because you have to type 2 keys less (shift+:).


mdurao:
If I have a command like: `php artisan import:csv ../data/random.csv`

The plugin recognise my custom command but in the end I can't tab through the filesystem to search for a file... 

`art<tab> imp<tab> ..<tab>` <--- doesn't work

How can I add that functionality to the plugin?

PR #6391, opened 2017-11-03
fix parse_git_dirty() to work on Solaris

cy6erGn0m:
We need this as Solaris's tail doesn't support `-n` option. However it does support old `-[n]` format so we can simply use `tail -1` instead of `tail -n1`

Tested on Linux, Solaris 11.1 and FreeBSD 11.3. Haven't test it under MacOS as I don't have such machine


Wasapl:
Unfortunately on MacOS tail does not support `-[n]` format.

awasum:
@cy6erGn0m , Please fix conflicts.

rockagen:
bad, on manjaro, this update  will cause `tail: option used in invalid context -- 1`  on arch or manjaro system.

PR #6398, opened 2017-11-07
More compatible way of checking if zsh is installed

v0idst4r:
Instead of reading /etc/shells it simply uses the standard Unix command which. This should resolve various issues to do with the script not detecting zsh.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -24,12 +24,10 @@ main() {
   # which may fail on systems lacking tput or terminfo
   set -e
 
-  CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
-  if [ ! $CHECK_ZSH_INSTALLED -ge 1 ]; then
+  if [ -z `which zsh` ]; then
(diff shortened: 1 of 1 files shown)


belak:
There's also https://github.com/robbyrussell/oh-my-zsh/pull/5931 which tries to use /etc/shells, falling back to `which zsh` before giving up.

v0idst4r:
There's no point in checking /etc/shells at all imo

belak:
Maybe the best solution would be to check /etc/shells if it exists... possibly making sure `which zsh` matches what's in there. Otherwise, there's a chance that chsh will fail because the shell you want to switch to isn't in /etc/shells, as can be the case on OSX (probably the most common case if you install zsh from homebrew).

v0idst4r:
Lol I didn't notice that if tries to set zsh as your default shell. That should be something that the user does on their own.

mcornella:
I pushed a slightly more idiomatic version of this (with your authorship as well). Excuse the late reply!

PR #6403, opened 2017-11-07
Fix wrong $GNUPGHOME usage in gpg-agent plugin

Czocher:
$GNUPGHOME variable was used incorrectly and caused a grep error when
set. See #6140 for more details.


ekandler:
Thanks! It's working now.

PR #6416, opened 2017-11-14
Fixed Pacaur aliases

sunng87:
`pacaur -Syua` only updated AUR pakcages, wihch is incompatible with semantics of `pacupg` and `yaupg`. Removing `-a` here it would work for both main repos and aur.


robbyrussell:

Cheers,
Robby

PR #6428, opened 2017-11-17
Improve formatting of Terraform plugin output

chrisbelyea:
Improves the formatting of the Terraform plugin's `tf_prompt_info()`
function (see #6373 and #6418). It:
- adds a `tf:` prefix to indicate that what follows is the Terraform
  workspace
- adds colorization and formatting to make the appearance more similar
  to the `git-prompt` plugin
- adds a trailing space to separate the prompt from user input

### DIFF
--- a/plugins/terraform/terraform.plugin.zsh
+++ b/plugins/terraform/terraform.plugin.zsh
@@ -1,7 +1,10 @@
 function tf_prompt_info() {
+    RESET="%{${reset_color}%}"
+
     # check if in terraform dir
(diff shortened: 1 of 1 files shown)


chrisbelyea:
@docent-net @miry This builds on your work—thanks for writing and enhancing the plugin!

mcornella:
In my opinion this should be configurable by the user.

As is now, it's sufficiently barebones so as to be possible to wrap the `tf_prompt_info` function between a prefix and suffix that color the output.

PR #6429, opened 2017-11-18
Coinmarketcap - Plugin

danielmain:
Hello

I created a simple plugin that allows you to see the last USD Price of all crypto currencies listed in coinmarketcap.com

Since the cryptos are very popular this is the very simplest way to know the current price just from the command line.


mcornella:
Has no interest.

PR #6432, opened 2017-11-19
Nmap Plugin: add more aliases, and documentation

franccesco:
Add more alias for stealth, firewall bypass, find sql injection points, host discovery, smb check vulns, dns brute force, check badly configured devices and empty/default passwords :)


mcornella:
Not documented in the README, fork deleted.

PR #6434, opened 2017-11-19
Add plugin for bazel completion

israelfaria:
Hi, I have created a new plugin to add bazel zsh completion copied from their repo.

### DIFF
--- a/plugins/bazel/README.md
+++ b/plugins/bazel/README.md
@@ -1,0 +1,5 @@
+## Bazel autocomplete plugin
+
+A copy of the completion script from the
+[bazelbuild/bazel](https://github.com/bazelbuild/bazel/master/scripts/zsh_completion/_bazel)
+git repo.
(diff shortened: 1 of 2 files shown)


willdzeng:
Nice, this is very helpful but, no one is reviewing this?

israelfaria:
Rebased and updated from bazel upstream. Still hoping someday this will get merged!

jackwish:
FYI: you may try out a *custom* plugin https://github.com/jackwish/bazel

diwu1989:
can we have this merged please?
it's much easier to use this as an official plugin vs third party customization

larson-carter:
I'm going to leave this here for reference: https://github.com/bazelbuild/bazel/blob/master/scripts/zsh_completion/_bazel

We will have to find a way to automate the script from bazel, also we will have to actively maintain the zsh completion.

PR #6436, opened 2017-11-20
Add autocomplete plugin for minikube

kenden:
[Minikube](https://github.com/kubernetes/minikube) can generate zsh completion code since [version v0.24.0](https://github.com/kubernetes/minikube/blob/master/CHANGELOG.md#version-0240---11292017).

This PR allows to enable this easily with oh-my-zsh.

### DIFF
--- a/plugins/minikube/minikube.plugin.zsh
+++ b/plugins/minikube/minikube.plugin.zsh
@@ -1,0 +1,6 @@
+# Autocompletion for Minikube.
+#
+
+if [ $commands[minikube] ]; then
+  source <(minikube completion zsh)
+fi


AlexxNica:
@robbyrussell Is there a reason this hasn't been merged yet?

Frederick888:
Why does this command take nearly 0.5s to generate the completion code?

kenden:
@Frederick888 Could you add some details? Is it slower than the autocomplete of say [helm](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/helm/helm.plugin.zsh), [kops](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/kops/kops.plugin.zsh), [kubectl](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/kubectl/kubectl.plugin.zsh) ?

Frederick888:
Sorry I didn't make it clear. I wasn't talking about the completion, but the `minikube completion zsh` itself.

```
$ time minikube completion zsh
...balabala
...balabala
minikube completion zsh  0.17s user 0.04s system 43% cpu 0.488 total
```

I'm not using a particularly crappy machine (quad core i7) and I think half a second could be considered noticeable in terms of the launching time of a shell.

kenden:
@Frederick888 
I can't seem to reproduce (on a 3 years old macbookpro)
```sh
$ time minikube completion zsh
minikube completion zsh  0.06s user 0.01s system 106% cpu 0.063 total
```

Anyway, this is an issue for the minikube repo, not this one. 
Could you create one there, including the output of `uname -a`? Cheers!

AnthonyWC:
This works fine on my machine; I had similar #s as @Frederick888 and I don't notice any noticeable delay so it may just be local issue:

`minikube completion zsh  0.18s user 0.02s system 58% cpu 0.354 total`

Frederick888:
@AnthonyWC Hmmm... isn't ~0.400s still abnormally high? I'd expect a delay of ~0.01s... I don't do Go programming but apparently [cobra](https://github.com/spf13/cobra) takes quite a while to generate the auto-completion script compared with [clap](https://crates.io/crates/clap) which is used by Rustup in my example.

But since it's platform-specific, perhaps the issue goes deeper than I thought into the language itself, which makes it hard to address... (in fact, `minikube version` takes much longer than `rustup --version`.) Maybe compile-time evaluation or some build script tricks could be a better way to go. This can even be done by updating this PR, via caching the script somewhere and re-generate only in case of `minikube` updates. A bit ugly tho.

alexdor:
Is there a reason this hasn't been merged yet? 
ping @mcornella

michaelfavia:
Thank you!

PR #6440, opened 2017-11-21
Added ranger-cd plugin

b4tg5s:
@robbyrussell I have a suggestion.

#### ranger-cd
I propose a tiny zsh plugin for the `ranger` file browser: When ranger quits, the terminal's working directory is the same from which ranger was started before in the first place. This plugin changes the terminal's working directory to the last selected directory in ranger, after quitting ranger.


mcornella:
Has no interest.

PR #6448, opened 2017-11-23
Add twoLine theme

richardtbell:
screen shot 2017-11-23 at 13 47 51

Add's a theme which sets the prompt over two lines as shown in the attached image. It also has several symbols used to indicate git status.


folixg:
https://github.com/robbyrussell/oh-my-zsh/#do-not-send-us-themes

Issue #6404, opened 2017-11-08
[vi-mode] Mode indicator disappears when resizing window


b4tg5s:
@robbyrussell 
#### Problem
The vi-mode plugin indicator disappears when the window (of the terminal emulator) is resized. I was hoping to find a zsh widget hook which lets me reset the prompt when resizing, but I got nothing so far. This only happens when `$ZSH_THEME` does not touch the `$RPS1` or `$PROMPT` (in this case the theme handles these variables instead of `plugins/vi-mode`).

#### Reproduce
- activate `vi-mode` plugin
- use theme `robbyrussel` (or any other that does not touch `$RPS1` or `$PROMPT`)
- start zsh in gnome-terminal
- press <kbd>ESC</kbd> to enter normal mode (`vicmd`), red indicator appears on the right side
- resize the terminal window

#### Desired Behavior
- red indicator should stay until I leave normal mode

#### Actual Behavior
- red indicator disappears altough I  am still in normal mode

At this point I don't know in which direction I should be looking.

henrebotha:
Same issue here! Been driving me insane the last few days. I tried `zle-line-pre-redraw`, but no dice.

I suspect this has nothing to do with the vi mode plugin, because I'm not using it; I'm just using Zsh's built-in vi mode.

### Environment

I'm using a custom theme, and I can reproduce the issue with just this:

```zsh
# ~/.oh-my-zsh/themes/custom-theme.zsh-theme
function zle-line-init zle-keymap-select {
  local return_status="»"
  zle reset-prompt
}
zle -N zle-line-init
zle -N zle-keymap-select

PROMPT='${return_status}'

# ~/.zshrc
export ZSH=/Users/henrebotha/.oh-my-zsh
ZSH_THEME="custom-theme"
source $ZSH/oh-my-zsh.sh

bindkey -v
bindkey '^P' up-history
bindkey '^N' down-history
bindkey '^?' backward-delete-char
bindkey '^h' backward-delete-char
bindkey '^r' history-incremental-search-backward
bindkey '^s' history-incremental-search-forward

# ~/.zshenv
export SHELL="zsh"
```

I'm running inside tmux.

### Reproduction

1. Start the shell.
2. Resize the window.

### Expected behaviour

The prompt character is visible after resize.

### Observed behaviour

The prompt character is deleted.

b4tg5s:
@henrebotha 

[...]

PR #6449, opened 2017-11-23
fix bug-6404

b4tg5s:
This pr fixes two things:
* [bug-6404](https://github.com/robbyrussell/oh-my-zsh/issues/6404): Indicator is shown after window resize. In case the window is maximized, the indicator is shown but an error message is displayed: `widgets can only be called when ZLE is active`. Haven't found out yet how to get rid of it, but it works.
* Makes sure `MODE_INDICATOR` is drawn when the terminal is started. This is not the case with the actual plugin.


robbyrussell:

Cheers,
Robby

PR #6468, opened 2017-12-03
python.zsh to get the current python version

grantstephens:


mcornella:
This would only be useful with a tool like virtualenv, where the python version changes often. And we already have that in the `virtualenv` plugin.

PR #6469, opened 2017-12-04
Fix gpg-agent plugin checks

chrisfleming:
I discovered a few problems with the gpg-agent plugin changes on some machines. 

1. ${XDG_RUNTIME_DIR}  isn't always set
2. Older versions of gpg-agent do not use the "default location", therefore multiple agents are opened.

Therefore my solution is to use the gpgconf file to get the default location of the socket. We use this as the test to see if we need to start the agent.

Adding -use-standard-socket to starting the daemon means that older versions of gpg-agent behave in the same was as newer versions. 

I applied the same idea for the ssh agent. We can check if we have an ssh socket, if so then we set the SSH_AUTH_SOCKET - The reason for this is that newer versions of gpg-agent default to enabling ssh-agent therefore checking for the config variable wasn't reliable.

Hope this all makes sense.

Cheers
Chris

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -1,16 +1,17 @@
 # Enable gpg-agent if it is not running-
 # --use-standard-socket will work from version 2 upwards
-# If already running then new agent won't be started.
 
-gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
+AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`
+
+if [ ! -S ${AGENT_SOCK} ]; then
+  gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
+fi
 export GPG_TTY=$(tty)
 
-# Set SSH to use gpg-agent if it is configured to do so
-GNUPGCONFIG="${GNUPGHOME:-"$HOME/.gnupg"}/gpg-agent.conf"
-if [ -r "$GNUPGCONFIG" ] && grep -q enable-ssh-support "$GNUPGCONFIG"; then
-  if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
-    export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"
-    unset SSH_AGENT_PID
-  fi
+# Set SSH to use gpg-agent if it's enabled
+GPG_SSH_AUTH_SOCK="${AGENT_SOCK}.ssh"
+if [ -S ${GPG_SSH_AUTH_SOCK} ]; then
+  export SSH_AUTH_SOCK=$GPG_SSH_AUTH_SOCK
+  unset SSH_AGENT_PID
 fi
 
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -1,14 +1,16 @@
-# Enable gpg-agent if it is not running
-GPG_AGENT_SOCKET="${XDG_RUNTIME_DIR}/gnupg/S.gpg-agent.ssh"
-if [ ! -S $GPG_AGENT_SOCKET ]; then
-  gpg-agent --daemon >/dev/null 2>&1
-  export GPG_TTY=$(tty)
-fi
+# Enable gpg-agent if it is not running-
+# --use-standard-socket will work from version 2 upwards
+# If already running then new agent won't be started.
+
+gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
+export GPG_TTY=$(tty)
 
 # Set SSH to use gpg-agent if it is configured to do so
 GNUPGCONFIG="${GNUPGHOME:-"$HOME/.gnupg"}/gpg-agent.conf"
 if [ -r "$GNUPGCONFIG" ] && grep -q enable-ssh-support "$GNUPGCONFIG"; then
-  unset SSH_AGENT_PID
-  export SSH_AUTH_SOCK=$GPG_AGENT_SOCKET
+  if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
+    export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"
+    unset SSH_AGENT_PID
+  fi
 fi
 
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -9,9 +9,8 @@ fi
 export GPG_TTY=$(tty)
 
 # Set SSH to use gpg-agent if it's enabled
-GPG_SSH_AUTH_SOCK="${AGENT_SOCK}.ssh"
-if [ -S ${GPG_SSH_AUTH_SOCK} ]; then
-  export SSH_AUTH_SOCK=$GPG_SSH_AUTH_SOCK
+if [ -S "${AGENT_SOCK}.ssh" ]; then
+  export SSH_AUTH_SOCK="${AGENT_SOCK}.ssh"
   unset SSH_AGENT_PID
 fi
 


plugins/gpg-agent/gpg-agent.plugin.zsh
```diff
@@ -1,14 +1,17 @@
...
-if [ ! -S $GPG_AGENT_SOCKET ]; then
-  gpg-agent --daemon >/dev/null 2>&1
-  export GPG_TTY=$(tty)
+# Enable gpg-agent if it is not running-
+# --use-standard-socket will work from version 2 upwards
+
+AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`
+
+if [ ! -S ${AGENT_SOCK} ]; then
+  gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
```
piotron: --use-standard-socket since GnuPG 2.1 is default behaviour. And this option is not used.

plugins/gpg-agent/gpg-agent.plugin.zsh
```diff
@@ -1,14 +1,17 @@
-# Enable gpg-agent if it is not running
-GPG_AGENT_SOCKET="${XDG_RUNTIME_DIR}/gnupg/S.gpg-agent.ssh"
-if [ ! -S $GPG_AGENT_SOCKET ]; then
-  gpg-agent --daemon >/dev/null 2>&1
-  export GPG_TTY=$(tty)
+# Enable gpg-agent if it is not running-
+# --use-standard-socket will work from version 2 upwards
+
+AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`
```
piotron: GPG_AGENT_SOCKET=`gpgconf --list-dirs agent-ssh-socket`
would be enough like in my PR from two weeks ago (#6431 )

plugins/gpg-agent/gpg-agent.plugin.zsh
```diff
@@ -1,14 +1,17 @@
...
+if [ ! -S ${AGENT_SOCK} ]; then
+  gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
 fi
+export GPG_TTY=$(tty)
 
-# Set SSH to use gpg-agent if it is configured to do so
-GNUPGCONFIG="${GNUPGHOME:-"$HOME/.gnupg"}/gpg-agent.conf"
-if [ -r "$GNUPGCONFIG" ] && grep -q enable-ssh-support "$GNUPGCONFIG"; then
+# Set SSH to use gpg-agent if it's enabled
+GPG_SSH_AUTH_SOCK="${AGENT_SOCK}.ssh"
```
piotron: if you use gpg-conf with list dir, you will have full path for socket.

chrisfleming: Again, see above. Version 2.0 does not return the ssh path!

I would love if this was simpler, but I have found all these to be necessary. But I do think it is still necessary to support version 2.0 as CentOS 7 as well as Ubuntu 14.04 are both shipping with version 2.0.22 so I don't think this is an uncommon setup?

plugins/gpg-agent/gpg-agent.plugin.zsh
chrisfleming: I found that this didn't work in all cases, for example on an ubuntu 14.04 box gpg is on version  2.0.22 - On this box:  

gpgconf --list-dirs agent-ssh-socket

sysconfdir:/etc/gnupg2
bindir:/usr/bin
libexecdir:/usr/lib/gnupg2
libdir:/usr/lib/x86_64-linux-gnu/gnupg2
datadir:/usr/share/gnupg2
localedir:/usr/share/locale
dirmngr-socket:/var/run/dirmngr/socket
agent-socket:/home/chrisfl/.gnupg/S.gpg-agent
homedir:/home/chrisfl/.gnupg

Hence the longer version.

plugins/gpg-agent/gpg-agent.plugin.zsh
chrisfleming: Again as you said, it would be nice not to need this, but it is necessary on 2.0

plugins/gpg-agent/gpg-agent.plugin.zsh
piotron: @chrisfleming Okay, so why won't you use agent-socket and add it? Check if GPG>=2.1.13 then use `agent-ssh-socket`, if not then use `agent-socket` and eventually add .ssh to it?

plugins/gpg-agent/gpg-agent.plugin.zsh
piotron: Okay, but on 2.1 it can throw information about deprecation on each new session, so again we need to add version checking.

plugins/gpg-agent/gpg-agent.plugin.zsh
piotron: This check is not only related to GPG_SSH_AUTH_SOCK,  you've removed check which is leaving gnupg agent intact if there is no ssh support enabled.

plugins/gpg-agent/gpg-agent.plugin.zsh
brammittendorff: Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial

```
gpg --version
gpg (GnuPG) 2.1.11
libgcrypt 1.6.5
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: ~/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
```

```
gpgconf --list-dirs agent-ssh-socket
sysconfdir:/etc/gnupg2
bindir:/usr/bin
libexecdir:/usr/lib/gnupg2
libdir:/usr/lib/x86_64-linux-gnu/gnupg2
datadir:/usr/share/gnupg2
localedir:/usr/share/locale
dirmngr-socket:/home/bram/.gnupg/S.dirmngr
dirmngr-sys-socket:/var/run/gnupg2/S.dirmngr
agent-socket:/home/bram/.gnupg/S.gpg-agent
homedir:/home/bram/.gnupg
```

plugins/gpg-agent/gpg-agent.plugin.zsh
chrisfleming: I did wonder about this - but checking/parsing for the version felt error prone and complicated versus dumping the output to /dev/null which the plugin was alreading doing and  which will hide the  deprecation  warning?

plugins/gpg-agent/gpg-agent.plugin.zsh
chrisfleming: The check itself is:

````zsh
    if [ -S ${GPG_SSH_AUTH_SOCK} ]; then
````
So we are checking that the ssh socket exists, which feels to me like a good robust way of checking for support

Version 2.2 of gpg-agent says"The OpenSSH Agent protocol is always enabled,"; so it is possible to use gpg-agent without adding the option to the config file. 

But it is a good question, if someone doesn't want to use gpg-agent as an ssh-agent then we probably shouldn't do the setup. This could either be  by going back to the previous technique of checking the config or possibly by checking that SSH_AUTH_SOCK isn't  already set?

File: plugins/gpg-agent/gpg-agent.plugin.zsh (hunk 1 above)
Comment (piotron) [reply plugins/gpg-agent/gpg-agent.plugin.zsh]: Version checking would give us also option to turn on support for gpg v1 again. Question is, if we want to support old GPG?
I think we shouldn't suppress messages for thing which is not generating any output if behaves correctly.
About cuts, I've checked behaviour of gpgconf on v2.0 and before 2.1.13 and now I understand why you used it. But we could use sth like ``` AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`'.ssh' ``` to do only one allocation.

I'll sit today/tomorrow and try to create some clean solution without hacks.

Comment (chrisfleming) [reply plugins/gpg-agent/gpg-agent.plugin.zsh]: > Version checking would give us also option to turn on support for gpg v1 again. Question is, if we > want to support old GPG?

The oldest box I use regularly is on CentIOS 6 - which was released in 2011 and that is using version 2.0.14 of gpg-agent 

The  --use-standard-socket option was introduced in: GnuPG 1.9.14 released (2004-12-22)

> I think we shouldn't suppress messages for thing which is not generating any output if behaves correctly.
Agree Completly. 

>About cuts, I've checked behaviour of gpgconf on v2.0 and before 2.1.13 and now I understand why you used it. But we could use sth like AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`'.ssh' to do only one allocation.

For the moment I've removed the allocation of GPG_SSH_AUTH_SOCK

> I'll sit today/tomorrow and try to create some clean solution without hacks.

Sounds good to me.

**Comment by frenck** [comment]
Confirming this works on Mac with GPGTools as well.

File: plugins/gpg-agent/gpg-agent.plugin.zsh [hunk 4]
```diff
@@ -1,14 +1,16 @@
...
-  export GPG_TTY=$(tty)
+# Enable gpg-agent if it is not running-
+# --use-standard-socket will work from version 2 upwards
+
+AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`
+
+if [ ! -S ${AGENT_SOCK} ]; then
+  gpg-agent --daemon --use-standard-socket >/dev/null 2>&1
 fi
+export GPG_TTY=$(tty)
```
mcornella: This can be `$TTY` directly, no need for an extra fork.

mcornella:
What's the status of this PR? Is it working in all platforms and versions?

Should we add version checking in case `--use-standard-socket` is killed in future versions? For that there is the [`is-at-least` zsh function](https://github.com/zsh-users/zsh/blob/master/Functions/Misc/is-at-least): `is-at-least min-version $VERSION`.

Let me know.

Phill93:
Working on macOS 10.13.4

plugins/gpg-agent/gpg-agent.plugin.zsh
chrisfleming: No problem

chrisfleming:
@mcornella  Thanks for the review, I didn't know about the is-at-least min-version function - At the moment I mask errors for versions that don't support  --use-standard-socked by redirecting the output to /dev/null But if I do  a version check then it's possible to remove this and prevent masking fucture errors. So I will do this.

Testing wise, I have this running on MacOS, Ubuntu  12.04, 14.04 and 16.04 as well as CentOS 5,6 and 7

Basically any release of GNU PG from the last ten years....

mcornella:
Ok I'll pull this one and we can make any necessary improvements later on, since you've confirmed that this works everywhere. Thanks for your dedication!

tnwhitwell:
This has broken functionality on my macOS 10.13.4. Unsure if that's a me problem as-yet, but my gpg agent seems to create `S.gpg-agent.ssh` without being configured to do so. The ssh-agent plugin then does not create the `SSH_AUTH_SOCK` environment variable, breaking the agent.

PR #6482, opened 2017-12-10
add theme cobalt2

salapao2136:
now, I add theme cobalt2 in oh-my-zsh. I hope helpful because It is used frequently.


folixg:
Quoting the [README](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes):

>### Do NOT send us themes
>
>We have (more than) enough themes for the time being. Please add your theme to the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page.

PR #6483, opened 2017-12-10
Add Plugin for SDKMAN

rahulsom:
[SDKMAN!](http://sdkman.io/) is a tool for managing parallel versions of multiple Software Development Kits on most Unix based systems.

It is commonly used in the JVM community for downloading tools like ant, maven, gradle, groovy and of late, java too.


ryanstull:
I find myself needing this often, it would be great if we could get this merged!

mcornella:
Can you add it to the wiki? https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins

PR #6485, opened 2017-12-11
execute gradlew with "gradle" when gradlew-file is present

YoshiMan:
With this pull request you will be able to execute ./gradlew with a simple gradle comand. You dont have to distinguish between gradle and ./gradlew, just use gradle.


vanniktech:
This is the opposite of what I want. I always want to run with my pre installed Gradle version since I'm often jumping between projects. Can we make this an option?

mcornella:
@vanniktech you can put `unalias gradle` in your zshrc file.

PR #6487, opened 2017-12-12
git_prompt_info: display tags when available

papanikge:
As of now, the `git_prompt_info` method just prints the branch, and if there isn't any, it prints the SHA. I think printing a tag, when there is one, is pretty helpful.

This PR implements just that.


tonyluj:
+1

Especially in some repos which have many tags like Linux kernel, and we need to switch tag A to tag B frequently.

I wonder if we have a release schedule for this feature?

robbyrussell:
Thanks for this, and apologies for the very long silence.

This landed in #11318 back in 2022 — `lib/git.zsh` now falls back to `git describe --tags --exact-match HEAD` when HEAD is detached, which is exactly the behavior you were after here. So I'm closing this one out.

Sorry you didn't hear back at the time. You were right about it.

PR #6490, opened 2017-12-13
Add support for ForkLift 3

### LINKED ISSUE #6565: Plugin: Forklift gets an error while executing, because of new version of Forklift?
@bitboxer I use Forklift in the Version 3. Maybe that is the reason, why I get this error, when I try to use Forklift in the command shell with this Plugin (https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/forklift):

`227:231: execution error: „Finder“ hat einen Fehler erhalten: Parameterfehler. (-50)`

jkaving:
This adds support for ForkLift 3, which uses a different
application id and also uses a popover instead of a sheet for
entering the directory to go to.

This also improves the handling of different versions of ForkLift,
by first choosing any currently running instance, and if none is
running starting the newest available version.

Fixes #6565.


jcalado:
👍 can this please be merged? :) thanks

PR #6497, opened 2017-12-19
Add official git prompt

qiu-noname:
$ __git_ps1


robbyrussell:

Oh My Zsh already ships git's own `git-prompt.sh` — it's the `gitfast` plugin, which has been in the repo since 2012 and is kept in sync with upstream:

```zsh
plugins=(... gitfast)
```

That gives you `__git_ps1` without adding a 500-line file to `lib/`, which loads in every shell whether or not someone wants it. So I'm closing this one.

PR #6503, opened 2017-12-21
Added trizen to upgrade alias (archlinux plugin)

cravacuore:


mcornella:
Closed in favor of #6650.

PR #6506, opened 2017-12-24
Uppercases the issue in open issue command in the JIRA plugin

mattmanuel90:
Uppercases the a-z characters of the jira issue for convenience.
before: abc-123 
after: ABC-123

### DIFF
--- a/plugins/jira/jira.plugin.zsh
+++ b/plugins/jira/jira.plugin.zsh
@@ -64,7 +64,7 @@ function jira() {
       local issue_arg=$(git rev-parse --abbrev-ref HEAD)
       local issue="${jira_prefix}${issue_arg}"
     else
-      local issue_arg=$action
+      local issue_arg=$( tr '[a-z]' '[A-Z]' <<< $action)
       local issue="${jira_prefix}${issue_arg}"
     fi
     local url_fragment=''


mcornella:
I know this is old, but why is this necessary? I'm not familiar with JIRA so I don't immediately know. Reopen with an explanation if there's still interest in this.

mattmanuel90:
Hey I just got this email. I apologise for the non explanation for this change.

This plugin allows CLI support to interface with JIRA which is an issue tracking system used by many companies.

JIRA allows an issue to be opened via a URL with the browser by a link. Typing the command ```jira ABC-100``` will link to ```https://example.atlassian.com/browse/ABC-100``` which is a valid link, where ABC-100 is the ticket 100 belong to project ABC.

The current plugin requires the user to input the issue in capital case as JIRA by default doesnt allow tickets to be linked by in lowercase. The command ```jira abc-123```, will link to a broken url i.e ```https://example.atlassian.com/browse/abc-100```

This change allows the user to input the ticket in either case, its just a minor improvement in convenience so the user doesn't have to reach for the shift key.

i.e typing ```jira abc-100``` will link to ```https://example.atlassian.com/browse/ABC-100```

plugins/jira/jira.plugin.zsh
```diff
@@ -64,7 +64,7 @@ function jira() {
       local issue_arg=$(git rev-parse --abbrev-ref HEAD)
       local issue="${jira_prefix}${issue_arg}"
     else
-      local issue_arg=$action
+      local issue_arg=$( tr '[a-z]' '[A-Z]' <<< $action)
```
mcornella: You can uppercase a variable with this zsh syntax: `${(U)var}`, so you can get rid of the `tr` call and just use `${(U)action}`.

PR #6507, opened 2017-12-24
kube-ps1 zsh prompt helper

jonmosco:
Prompt helper to use with kubectl.  Assists when using multiple clusters and namespaces.

### DIFF
--- a/plugins/kube-ps1/README.md
+++ b/plugins/kube-ps1/README.md
@@ -1,0 +1,76 @@
+Kubernetes prompt for zsh
+=========================
+
+A Kubernetes (k8s) zsh prompt that displays the current cluster cluster
+and the namespace.
+
+Inspired by several tools used to simplify usage of kubectl
+
+NOTE: If you are not using zsh, check out [kube-ps1](https://github.com/jonmosco/kube-ps1) designed for bash
+as well as zsh.
+
+## Requirements
+
+The default prompt assumes you have the kubectl command line utility installed.  It
+can be obtained here:
+
+[Install and Set up kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+
+If using this with OpenShift, the oc tool needs installed.  It can be obtained from here:
+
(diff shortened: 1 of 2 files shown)


plugins/kube-ps1/README.md
```diff
@@ -0,0 +1,76 @@
...
+* ocp - OpenShift
+
+## Install
+
+1. Clone this repository
+2. Source the kube-ps1.zsh in your ~./.zshrc
+
+ZSH:
+```
+source path/kube-ps1.sh
```
serdardalgic: A suggestion, in order to be sourced at the startup of a new terminal by [oh-my-zsh.sh](https://github.com/robbyrussell/oh-my-zsh/blob/master/oh-my-zsh.sh) script, plugin names are added to a variable in zsh.rc called `plugins`, as explained in the [wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins#example). For this, you need to rename the file to `kube-ps1.plugin.zsh`. It would be more practical for using the plugin IMHO.

serdardalgic: I'd like to use this plugin, but it would mess up my PROMPT, which I'm quite content with it. However, I'd love to use this plugin as an addition to my current PROMPT, that would be awesome! @jonmosco would you consider that? I'd write a PR when I find time.

jonmosco: I can rename that, no problem.

This can be used in addition to your current PROMPT and should not mess your current one up.  I will add the new version of this along with the rename as well.  Also, check https://github.com/jonmosco/kube-ps1 for other instructions related to this prompt.

JacobAMason:
Shouldn't enabling this pluggin just enable the kube-ps1 function? I'm not entirely sure how the pluggin system works, but this is what happens for pluggins like gitfast

mcornella:
Fixed in 9711d8f. Run `upgrade_oh_my_zsh` to get the latest changes.

PR #6508, opened 2017-12-24
slurm command autocompletion

vmiheer:
1/n salloc: complete possible hostnames

https://slurm.schedmd.com/


mcornella:
Has no interest.

PR #6510, opened 2017-12-25
archlinux: upgrade function should be add $@

qwIvan:


mcornella:
The purpose of the `upgrade` function is to just upgrade all the installed packages. We could add that, it doesn't affect people who just run `upgrade`, but that is just not the point of the function.

PR #6512, opened 2017-12-26
Add kubernetes context manager

drym3r:
I couldn't find a convention to edit the wiki file, and I understand that this plugin should be added there. I may do it if you point me on the right direction.

What do you think, this plugin should be different than the [kubectl](https://github.com/drym3r/oh-my-zsh/tree/master/plugins/kubectl) one? I have no problem on add the one I made to that one.

PD: Thanks for all your work!

### DIFF
--- a/plugins/kube/README.md
+++ b/plugins/kube/README.md
@@ -1,0 +1,14 @@
+# Kubectl context manager
+Kubectl, the kubernetes client, needs context to work correctly. By default, it uses the file on `~/.kube/config`, which has an horrible design. With this plugin, the better approach is to not have a `~/.kube/config` file and `n` `~/.kube/something-config`. If you do so, you'll be able to manage multiple contexts easily.
+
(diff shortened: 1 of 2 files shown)


hoshsadiq:
why not just use `kubectl config set-context`?

drym3r:
@hoshsadiq This provides more than that. First of all, by default you don't have set a kubectl context, which is a good idea for example when you have 10 different clusters and haven't used k8s in a few days. The first order you'll run will be run against the last k8s cluster you used. Of course the logical thing would be to first see in which context you are, but we're human and sometimes we will forget to do that.
On the other hand, some tools like kops may overwrite your ~/.kube/config when creating a cluster. And the last part, if you want to share your cluster's config, it's very uncomfortable to disassembly the k8s configuration file.

jonmosco:
check out the kube-ps1 plugin:
https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/kube-ps1/README.md

drym3r:
I see that that feature was created two days before mine, and of course wasn't yet merged. They both work assuming different kube config files structure. I guess you can close this, if you doesn't see value on having two different approaches.

robbyrussell:
Closing per @drym3r's suggestion

PR #6513, opened 2017-12-26
Add autocompletion to docker-machine plugin

poma:
I was surprised that the plugin's doesn't do its supposed main function - autocompletion, so I've added it from the official repo


deeptechs:
I was also surprised, still not released in oh-my-zsh. @poma thanks for adding it.

If still not released here, follow the instructions visitors, 
https://docs.docker.com/machine/completion/#zsh

mcornella:
Fixed in master. Thanks though!

PR #6522, opened 2017-12-31
Update license year

m1guelpf:
TO DO:

- [x] Wait to the next year (100% done, depends on timezone)
- [ ] Merge!

:octocat:


NieDzejkob:
I'd add a point "make some other change" between these two, because that's how copyright years work.

PR #6524, opened 2018-01-02
fix(screen): respect `DISABLE_AUTO_TITLE`

cono:
I would like to be able to set title (permanent) for one my tab but can't do this, there is a variable in CORE of OMZ which disable auto titling, I would like same behavior to be adopted by the screen.plugin:
https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/termsupport.zsh#L55

### DIFF
--- a/plugins/screen/screen.plugin.zsh
+++ b/plugins/screen/screen.plugin.zsh
@@ -30,11 +30,14 @@ if [[ "$TERM" == screen* ]]; then
   # tell GNU screen what the tab window title ($1) and the hardstatus($2) should be
   function screen_set()
   {
-    # set the tab window title (%t) for screen
-    print -nR $'\033k'$1$'\033'\\\
-
     # set hardstatus of tab window (%h) for screen
     print -nR $'\033]0;'$2$'\a'
+
+    if [[ "$DISABLE_AUTO_TITLE" == true ]]; then
+      return
+    fi
+    # set the tab window title (%t) for screen
+    print -nR $'\033k'$1$'\033'\\\
   }
   # called by zsh before executing a command
   function preexec()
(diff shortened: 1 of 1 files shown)


robbyrussell (commented):
Thanks for this... and sorry, eight years is a ridiculous amount of time to wait for a reply.

The bug's real, I checked. The screen plugin defines its own `precmd`/`preexec` and blanks out core's `title()`, so `DISABLE_AUTO_TITLE=true` does nothing at all inside screen right now. Your fix is the right shape.

Couple of things though:

1. **The hardstatus still goes out.** `screen_set` sends two sequences... `ESC k` for the window title (`%t`) and `OSC 0` for the hardstatus (`%h`). The early `return` catches the first one but not the second, so with `DISABLE_AUTO_TITLE=true` screen still gets a fresh hardstatus on every prompt and every command. Core treats the variable as "don't set any titles" (`lib/termsupport.zsh:50,56`), so I think both should go. Nice side effect: you wouldn't need to reorder the two `print` calls anymore.

2. **The guard.** `[[ "$DISABLE_AUTO_TITLE" == true ]]` blows up under `setopt nounset` if the variable isn't set. Core does it like this:

   ```zsh
   [[ "${DISABLE_AUTO_TITLE:-}" != true ]] || return 0
   ```

Optional, if you feel like it: a line in `plugins/screen/README.md` mentioning that the plugin honors `DISABLE_AUTO_TITLE`.

Oh, and the trailing-newline bit of your diff is already fixed upstream, so don't worry about rebasing for that.

Either way's fine by me... push an update whenever you get to it, or just say the word and I'll make these two changes myself and merge it with you credited. After making you wait this long I'm not going to hand you homework.

cono:
> Either way's fine by me... push an update whenever you get to it, or just say the word and I'll make these two changes myself and merge it with you credited. After making you wait this long I'm not going to hand you homework.

Not sure who Claude was talking to, but I don't have access to push :D

robbyrussell (approved):
Thanks for the patience on this one, and sorry it took eight years. I pushed the two changes we talked about since you couldn't: same guard form as core, and it now suppresses the hardstatus too.

carlosala (approved):

carlosala:
Thanks!

PR #6526, opened 2018-01-04
Added aliases for docker.

abhinavmanchanda:
Ensured there's no duplication with docker-compose. cc: @AeonAxan


NieDzejkob:
Keep in mind that when you push to a branch, the PR updates. You don't have to close it and create a new one.

PR #6531, opened 2018-01-04
Auto load nvm node version when .nvmrc is exist in directory

rafaellucio:
Hi, this is a version to load Node.js when .nvmrc exists and the version is different.

Is there a PR #5782 by @drGrove with this implementation, but does not check if the user is interested in using this feature, as I include a conditional in `.zshrc` to check this feature for the version of the nvm load node automatically.

I added a README.md to explain how to add a nvm plugin in `.zshrc`

Thanks.


rafaellucio:
Closed this PR because merged a solution, thanks!

PR #6533, opened 2018-01-06
Add support for Mac to dirhistory plugin

peppelan:
Add shortcuts for mac keyboards as an alternative to alt+left
abd alt+right: mac users can now use opt+left and opt+right.

On acceptance of this PR, I will update the wiki page accordingly.

Review of @jeffwilliams would be appreciated, thanks.

### DIFF
--- a/plugins/dirhistory/dirhistory.plugin.zsh
+++ b/plugins/dirhistory/dirhistory.plugin.zsh
@@ -119,6 +119,8 @@ zle -N dirhistory_zle_dirhistory_back
 # xterm in normal mode
 bindkey "\e[3D" dirhistory_zle_dirhistory_back
 bindkey "\e[1;3D" dirhistory_zle_dirhistory_back
+# Mac teminal (alt+left/right)
+bindkey "^[b" dirhistory_zle_dirhistory_back
 # Putty:
 bindkey "\e\e[D" dirhistory_zle_dirhistory_back
 # GNU screen:
@@ -127,6 +129,7 @@ bindkey "\eO3D" dirhistory_zle_dirhistory_back
 zle -N dirhistory_zle_dirhistory_future
 bindkey "\e[3C" dirhistory_zle_dirhistory_future
 bindkey "\e[1;3C" dirhistory_zle_dirhistory_future
+bindkey "^[f" dirhistory_zle_dirhistory_future
 bindkey "\e\e[C" dirhistory_zle_dirhistory_future
 bindkey "\eO3C" dirhistory_zle_dirhistory_future
 


jeffwilliams:
This looks good to me. Thanks for improving it!

NAR8789:
@peppelan thanks! But... this merge maps `opt+b` and `opt+f`, not `opt+left` and `opt+right` as stated. Unfortunately this conflicts with emacs-style line navigation--`opt+b` should navigate back one word in the line, and `opt+f` should navigate forward one word.

I'm a little confused actually... previously I could already use `opt+left` and `opt+right` on mac. These were already specified in the file as

```
 bindkey "\e\e[D" dirhistory_zle_dirhistory_back
 bindkey "\e\e[C" dirhistory_zle_dirhistory_future
```

Would it be alright to revert this merge?

peppelan:
Hi, sure that's alright to revert the merge if it causes issues.

Anyways, the mapping I committed did work for me, and it also solved the issue of the existing mapping opt+left and opt+right not working on my mac.

Which is pretty much an experience which is opposite as yours... 
trying to figure out how this could be possible, and ideally *I would love* to revert the commit and have the existing mapping working for me as well, which OSX and terminal are you using?

Myself, I am on High Sierra 10.13.2, and use standard Mac OS Terminal with default settings.

NAR8789:
Interesting! I am about to go to sleep, but I will try this in the morning... I haven't upgraded to High Sierra yet. I am on Sierra 10.12.6, and I am using iTerm2 3.1.5.

In the morning I'll try on standard Mac OS Terminal, and let you know what I see there. Thanks for the info!

peppelan:
Thanks to you, let me know & goodnight.

NAR8789:
@peppelan I tried it in standard Mac OS Terminal this morning, and I am indeed seeing the behavior you described! Thank you for the details to reproduce! It looks like Terminal maps `opt+left` and `opt+right` to `opt+b` and `opt+f`, respectively, to make line navigation more closely resemble overall OS X line navigation. I have a screenshot at the end of my comment showing where it's mapped if you're curious.

Given that it's mapped by default however, I think your remapping still makes sense, since it'll be the right out-of-box experience for Terminal users. Rather than rolling this back then, @peppelan what do you think of adding a check to only enable these keybindings if the terminal is Terminal? I'm actually not familiar with zsh scripting best practices, so I'll defer to @jeffwilliams on this, but it looks like you can check the value of `$TERM_PROGRAM`.

```
> echo $TERM_PROGRAM
Apple_Terminal
```

And it looks like there's precedent in the oh-my-zsh codebase for checking for Apple_Terminal: https://github.com/robbyrussell/oh-my-zsh/search?utf8=%E2%9C%93&q=Apple_Terminal&type=

Some additional background: here are the settings where it was mapped in Terminal for me (Terminal > Preferences > Keyboard):
screen shot 2018-01-26 at 10 31 51

xiao-maven:
Fantastic! Thank you so much @peppelan!

NAR8789:
Fantastic! Thank you so much @peppelan!

PR #6537, opened 2018-01-08
add install silent mode

PanarinM:
to prevent breaking of environemnt setup scripts i added silent mode which basically just prevents the `env zsh` comands, so the setup script can continue


NieDzejkob:
IMO, this should be a flag that the main install script accepts. Otherwise, you are creating a maintenance burden.

mcornella:
Closing in favor of #6547.

PR #6539, opened 2018-01-09
Add error message if no binary found, and more common aliases.

michaelfavia:
This just adds on to my previous pull request with a couple more useful aliases we're tired of typing around here and adds error checking for a binary. 

Interestingly, because most people add their PATH exports at the bottom of their .zshrc files the plugin loaded the aliases but not the autocomplete (which is behind a sanity check) for some of our developers. This change prevents that and lets them know about the problem simply.


mcornella:
This is obsolete.

PR #6540, opened 2018-01-11
Added plugin and completion for sfdx

### LINKED ISSUE #7645: Support list of commands available for sfdx
Like SFDX is a lot of commands if you display next commands for parent display our life is easy.

Reference: https://github.com/wadewegner/salesforce-cli-zsh-completion

wadewegner:
This is a completion for the Salesforce CLI. I'm on the prod engineering team and it's quite useful.

Fixes #7645

### DIFF
--- a/plugins/sfdx/README.md
+++ b/plugins/sfdx/README.md
@@ -1,0 +1,11 @@
+# sfdx plugin
+
+This plugin provides autocompletion for the [Salesforce DX](https://developer.salesforce.com/tools/sfdxcli) CLI.
+
+To use it, add `sfdx` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... sfdx)
+```
+
+Original repository: https://github.com/wadewegner/salesforce-cli-zsh-completion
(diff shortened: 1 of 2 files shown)


robbyrussell:
@wadewegner This looks great -- could you we do two things? 1. rename this to salesforcedx (as I think it'll help increase adoption from OMZ users as it'll be a recognizable name) and 2) add a README.md file with some usage examples?

Thanks in advance!

chandra2ravi:
@wadewegner You completed below? and it's available?
could you we do two things? 1. rename this to salesforcedx (as I think it'll help increase adoption from OMZ users as it'll be a recognizable name) and 2) add a README.md file with some usage examples?

mcornella:
I think `sfdx` is a good name for the plugin as long as it's popularly called that (the command is named the same way). Having a README is important though, see https://github.com/robbyrussell/oh-my-zsh/issues/7175#issue-365263634 for instructions on how to write a proper README for Oh My Zsh.

rsoesemann:
@wadewegner looks like you are just a mini step away from getting merged ;-)

PR #6541, opened 2018-01-11
Fix MacOS compatibility of the gpg-agent plugin

Czocher:
MacOS holds the GPG agent socket in a different directory than
Linux-based systems. Asking gpgconf for the correct path is the
most straightforward approach to fix this problem.

See #6106 for further details and discussion.


mcornella:
Is it verified to work also on Linux systems? I don't have gpg-agent installed so I can't verify what the call to gpgconf returns.

mcornella:
Nevermind, I'm closing this one as it is a duplicate of #6469.

PR #6546, opened 2018-01-12
Update ssh-agent plugin README.md

tbaschak:
The instructions listed adding 

```zsh
zstyle :omz:plugins:ssh-agent agent-forwarding on
```

However the script is looking for a `yes` value later on, Docs should say to use `yes` not `on`.

### DIFF
--- a/plugins/ssh-agent/README.md
+++ b/plugins/ssh-agent/README.md
@@ -14,7 +14,7 @@ plugins=(... ssh-agent)
 To enable **agent forwarding support** add the following to your zshrc file:
 
 ```zsh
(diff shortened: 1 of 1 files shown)


mcornella:
The value `on` is still valid. The [documentation](http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Standard-Styles) says: 
> Note that the values of several of these styles represent boolean values. Any of the strings ‘true’, ‘on’, ‘yes’, and ‘1’ can be used for the value ‘true’

`on` seems like a more semantic word than `yes`

PR #6547, opened 2018-01-12
Add support for --silent option in install.sh script

Liquidsoul:
This is an implementation of #6537 using a flag in the existing script.
This allows to skip the execution of the zsh shell at the end of the installation process.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -108,7 +108,11 @@ main() {
   echo 'p.p.s. Get stickers, shirts, and coffee mugs at https://shop.planetargon.com/collections/oh-my-zsh.'
   echo ''
   printf "${NORMAL}"
-  env zsh -l
+  if [ "$1" == "--silent" ]; then
+    printf "Skipping launch of zsh shell. You can launch it yourself or restart your terminal."
+  else
+    env zsh -l
+  fi
 }
 
-main
+main "$@"


karanrajpal14:
Just what I need, actually. Hope @robbyrussell merges this.
Makes setting up my environment on other machines a lot quicker.

seognil:
I use this simple dirty way for my auto deployment

```shell
sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh | sed -e 's/^\s*chsh -s/sudo chsh -s/g' -e 's/^\s*env\szsh.*$/#/g')"
```

It will do two things before running the install script,
remove `env zsh` line,
replace `chsh -s` line with `'sudo chsh -s'`
(failed somehow, but won't pause my script, I can do this later manually anyway)

Liquidsoul:
@mcornella 👋 why the conflicts status here? What can I do to fix that?

BenCoffeed:
👍 This would be fantastic as currently, I havea . similar "simple dirty way" to @songnil as I actually run this in an ansible playbook on all of my Vagrant sandboxes for devs.

mcornella:
#5169 pulled in these changes and renamed the flag to `--unattended`, which also skips the change of default shell. It has a bunch of other fixes, and it's now ready to test.

Thanks!

mcornella:
Closed by #5169.

PR #6550, opened 2018-01-14
plugins/git/git.plugin.zsh: added 'git am' alias

x4m3:
added the alias 'gam' for the command 'git am'


mcornella:
`git am` is not used commonly enough nor by enough people to warrant a new alias.

PR #6562, opened 2018-01-19
Fix the env variable to set when it does not exist

maxhelias:
When the env_default function is used on a variable that does not yet exist, it creates an error message when executing the command: "source ~ / .zshrc"

_Example :_
With this feature #5231, I **never** define PAGER or LESS

oh-my-zsh

_Solution :_
Add a check on the variable before executing the command

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -95,8 +95,8 @@ function default() {
 #    0 if the env variable exists, 3 if it was set
 #
 function env_default() {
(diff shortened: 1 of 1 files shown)


maxhelias:
It seems to be fixed by https://github.com/robbyrussell/oh-my-zsh/commit/55575b88f9e47d88fab02c26fc69e4af48af1cd9#diff-395fc7e22b4a6ef8aacc81bee65def01

PR #6569, opened 2018-01-22
Patch for detect shell file in Solus OS

frbuceta:
Patch for install Oh-My-Zsh in Solus

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -23,8 +23,13 @@ main() {
   # Only enable exit-on-error after the non-critical colorization stuff,
   # which may fail on systems lacking tput or terminfo
   set -e
-
-  CHECK_ZSH_INSTALLED=$(grep /zsh$ /etc/shells | wc -l)
+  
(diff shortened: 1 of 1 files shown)


chikamichi:
Cool patch, but actually you may workaround this by running `sudo chsh -s /bin/zsh [username]` where the username is mandatory and expected to be yours of course. Which means it's just a matter of editing install instructions over at Oh-My-Zsh's wiki I guess :)

chikamichi:
At the same it's true that this missing /etc/shells binding does affect script-install, not just manual install, so OMZ might be better off fixing the issue in the install script.

frbuceta:
Perfect :) I would love to see it working perfectly Oh My Zsh by Solus Os

mcornella:
Hey there, I pulled this change in #5169 and it's now ready to test (I pulled other fixes and features).

Thanks!

mcornella:
Closed by #5169.

PR #6570, opened 2018-01-24
Added robbyrussell theme but without ➜ and ✗

muhammadmuzzammil1998:
Updated ➜ and ✗ with -> and x because they are not supported in Windows Subsystem for Linux.

Shows a '?' in box instead of ➜ and ✗.


muhammadmuzzammil1998:
Cool. :)

PR #6571, opened 2018-01-24
add some aliases

yeKcim:
what is a ohmyzsh plugin without some powerfull aliases?


patrick-motard:
Nice work! Would it be useful, as part of this PR, to add a section to the taskwarrior plugin's readme that includes all of the aliases for this plugin?

yeKcim:
@patrick-motard:  I'll probably add more alias and make them better in near future, i'll update readme then

PR #6578, opened 2018-01-26
Adds support for trizen

asm0dey:
Since pacaur is marked unmaintained in AUR and author himeself tells it's not maintained anymore I've added support for trizen


x4m3:
thank you! can't wait for it to be merged to start using it !

asm0dey:
@x4m3 well, looking at number of unmerged pull requests I think that it will happen somewhen around never. To be honest I do not understand policy which @robbyrussell use to merge or not to merge PRs. It may be easier for you just to commit this file locally.

x4m3:
that's true, but i saw that @robbyrussell merged changes from december a
couple days ago, i think he has just a lot of work todo, maybe in a couple
of months, who knows?

but it would be nice to have a second maintainer or something to get more
stuff going , but it's his project, he does whatever he wants to do

zlalanne:
Would love to get this in too. Looking at the PR, what do you think of adding it to the top of the if/elif check to look for aur helpers. That way trizen takes precedence over pacaur if they are both available?

x4m3:
> 30 янв. 2018 г., в 16:12, Zack Lalanne <notifications@github.com> написал(а):

> Would love to get this in too. Looking at the PR, what do you think of adding it to the top of the if/elif check to look for aur helpers. That way trizen takes precedence over pacaur if they are both available?

i don't think so, if you have both pacaur and trizen on your machine, maybe you're doing  something wrong.

> That way trizen takes precedence over pacaur if they are both available?

what if you want pacaur by default and not trizen? i think it's too complicated to decide for the user, the use *should* just use one aur helper and stick to it. hope that made sense.

-x4m3

asm0dey:
@zlalanne they have different aliases, so they won't interfere, you can have all the tools if you want :)

PR #6587, opened 2018-01-31
Quote $ZSH where necessary in install script

### LINKED ISSUE #4470: spaces in $HOME
Hy!

My $HOME env variable contains spaces... it's an osx installation with an ssd and a hdd. I call the hdd: "Data HD" and it's the host for my home folder..

Both install.sh and the default .zshrc fails with this setup. I fixed it by replacing ZSH=~/.oh-my-zsh lines in both to ZSH="$HOME/.oh-my-zsh"

Now it seems to work :)

fredden:
The install script fails if the user's home directory contains a space or other white-space character. This fixes that error. See issue #4470 for some discussion around this. This patch doesn't solve any issues in other software, only in OMZ.

Fixes #4470

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -61,7 +61,7 @@ main() {
       exit 1
     fi
   fi
-  env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git $ZSH || {
+  env git clone --depth=1 https://github.com/robbyrussell/oh-my-zsh.git "$ZSH" || {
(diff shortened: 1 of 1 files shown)


some1else:
Can we merge this? I made the same patch, not knowing a PR exists.
This situation is common for users of Apple Migration Assistant, as it adds numbers to avoid conflict. For instance, `Username` will be renamed to `Username 1`, if `/Users/Username` already exists.

Issue #6109, opened 2017-05-26
How do you display seconds in time?


ancon:
please help me 
How do you display seconds in time?
like the bash cmd
histtimeformat="%F %T"

eremit:
%*
Current time of day in 24-hour format, with seconds.

ancon:
@eremit 
# Uncomment the following line if you want to change the command execution time
# stamp shown in the history command output.
# The optional three formats: "mm/dd/yyyy"|"dd.mm.yyyy"|"yyyy-mm-dd"
# HIST_STAMPS="mm/dd/yyyy"
how to set it? please help me , 3q~

eremit:
@ancon okay, probably I was not really paying attention to your report..

see the following issue as reference:

https://github.com/robbyrussell/oh-my-zsh/issues/3466

In a nutsell:
* add one line HIST_STAMPS="yyyy-mm-dd"
* add one alias: alias history='fc -il 1'

ancon:
@eremit thanks~
but it didn't work...
it can display the minute, but it display the seconds~
![2017-06-01 16 16 48](https://cloud.githubusercontent.com/assets/6762451/26670832/ce87cb6e-46e5-11e7-973a-411b21315a23.png)

eremit:
@ancon use this one:

```
alias history="history -t'%F %T'
```

eremit:
Hiya @ancon 
is your issue now resolved by the alias that I provided?

ancon:
@eremit thanks a lot.
it work, but did not perfect.
I just think that display the second is unnecessary.

ancon:
@CristianCantoro it work perfect! thanks a lot!

tjharman:
I just had no idea why I wasn't seeing timestamps in my history either after setting this.

CristianCantoro:
@tjharman said:
> I just had no idea why I wasn't seeing timestamps in my history either after setting this.

If you are interested in this feature you can support this pull request: PR https://github.com/robbyrussell/oh-my-zsh/pull/6770

mcornella:
I've merged #6770 to include this feature. Now setting `HIST_STAMPS` to a custom format will automatically update the history alias to show the custom timestamp, no need to re-alias `history`.

tjharman:
I find that it works if it's set to mm/dd/yyyy
But in New Zealand we use dd/mm/yyyy and if I set that, my output looks as follows:

```
[...]

PR #6589, opened 2018-01-31
Clarify how to enable timestamps in history (#6109)

CristianCantoro:
(see also https://github.com/robbyrussell/oh-my-zsh/issues/6109 for further info)

The comment in the `.zshrc` is not very clear about the fact that you also need to change the alias of `history` to display timestamps.


mcornella:
This is not necessary because the alias is automatically set up in `lib/history.zsh`.

CristianCantoro:
For some reason it was not working for me, see also https://github.com/robbyrussell/oh-my-zsh/issues/6109. Would you be interested in a pull-request to enable showing a custom format for timestamps?

PR #6593, opened 2018-02-02
add option to disable the system beep

bassclarinetl2:
Disables the beep upon tab completion.  Only tested on Ubuntu wsl.


mcornella:
This is a setting that you can just add to the zshrc file, there is no need to document it in the template zshrc.

PR #6594, opened 2018-02-02
Support new naming convention in Bundler

7even:
As [Bundler](https://bundler.io) now allows using `gems.rb` and `gems.locked` filenames instead of `Gemfile` and `Gemfile.lock`, the bundler plugin should try searching not only for `Gemfile` but also for `gems.rb` when it is checking if we are inside a bundled project.

Currently the bundler plugin doesn't prepend bundled commands with `bundle exec` if the project uses `gems.rb` instead of a `Gemfile`.


foxweb:
👍

exviva:
👍 any way we could get this one merged?

PR #6595, opened 2018-02-02
Add theme "simplycool"

Tom1380:


kfkonrad:
Please keep in mind to add a description in the Themes wiki pages, in case the request is accepted.

PR #6315, opened 2017-09-25
Update _docker-compose to latest version

qrevel:

PR #6607, opened 2018-02-06
Update docker-compose plugin completion script

ajmath:
Pulls in the latest docker-compose completion script from the [docker-compose](https://github.com/docker/compose/blob/master/contrib/completion/zsh/_docker-compose) git repo.


ajmath:
Replaces #6315

PR #6608, opened 2018-02-06
Uninstall restores now the original shell

tonyqui:
Hi there, 
after having uninstalled oh-my-zsh I've noticed that the environment was not restored correctly. 
The aim of this PR is to:
1 - Store the current shell into ~/.shell.pre-oh-my-zsh, when running install.sh
2 - Restore the default shell (previous to oh-my-zsh), when running the uninstall script.


fredden:
See also pull request #2626.

tonyqui:
Looks similar ... shall we merge the other then?

mcornella:
Hi @tonyqui I pulled the fix in #5169 which introduces other fixes and is now ready for testing.

Thanks!

mcornella:
Closed by #5169.

Issue #5877, opened 2017-02-13
How to add `-X` flag to LESS


StyxOfDynamite:
I love the colorized output in the `less` pager but need to also keep the output on the screen after quitting.

How can I add the `-X` arg to the LESS environment variable?

StyxOfDynamite:
Solved it. Needed to add the flag in the following file.

oh-my-zsh/lib/misc.zsh

PR #5363, opened 2016-09-01
misc: please don't overwrite my PAGER/LESS options

dlamotte:
I have these specified in `~/.zshenv`, please don't overwrite them :)

Great project btw!  Thanks!

PR #6611, opened 2018-02-08
add -F and -X to default $LESS environment variable

larsxschneider:
The option '-F' causes 'less' to automatically quit if the contents fit 
the screen and the option '-X' causes 'less' to not clear the screen after 
quit. I think both options are generally useful for terminal applications.

They are in particular useful for Git as it runs all output through a
pager. Git will run 'less' with '-FRX' by default if the environment 
variable $LESS is not defined [1]. Since oh-my-zsh used to set $LESS to 
'-R', Git would not override this setting. Consequently, Git would 
display even a single line of output in a pager and the user would need
to explicitly quit that pager (see mailing list discussion [2]).

Therefore, lets change the oh-my-zsh default value for $LESS to '-FRX'.
This would be useful for oh-my-zsh Git users and likely for users of
other applications that use 'less' too.

[1] https://github.com/git/git/blob/36438dc19dd2a305dddebd44bf7a65f1a220075b/Documentation/config.txt#L819-L821
[2] https://public-inbox.org/git/2412A603-4382-4AF5-97D0-D16D5FAAFE28@eluvio.com/

---

#5363 #5877 are LESS related issues but they don't fix the problem described here.
/cc @peff @jasonracey

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -20,7 +20,7 @@ setopt long_list_jobs
 
 ## pager
 env_default PAGER 'less'
-env_default LESS '-R'
+env_default LESS '-FRX'
 
 ## super user alias
 alias _='sudo'


Ikke:
Either setting it to `-FRX` or not setting it at all would help a lot of users (from time to time in the #git irc channel, we have have to tell people to unset LESS for git to work as expected).

BadSingleton:
+1 to set to `-FRX`

larsxschneider:
@mcornella gentle ping 😊

mcornella:
lgtm

larsxschneider:
Thank you 😊

Chris911:
The `-X` flag breaks scrolling in iTerm2 and Terminal on macOS.

Other people are reporting this as well: https://apple.stackexchange.com/questions/332465/scrolling-doesnt-work-in-terminal-for-commands-like-less?noredirect=1#comment427136_332465 and #7025.

roberthawdon:
I really don't like this setting, and Less doesn't seem to have the option to turn it off on individual runs. Shouldn't this have been made a default commented out option in the .zshrc?

larsxschneider:
Oh, I am sorry for the regression @Chris911 and @roberthawdon 😢  ! I am using iTerm2 too and I currently cannot replicate the issue. Can you tell me if you use other versions?

*Here are my versions:*
macOS 10.13.6 
iTerm2 Build 3.1.7
```
$ less --version
less 487 (POSIX regular expressions)
Copyright (C) 1984-2016  Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Homepage: http://www.greenwoodsoftware.com/less
```

Chris911:
macOS 10.13.6
iTerm2 Build 3.1.7

Tried with less 487 and 530. I tried a bunch of things yesterday and the only fix that works is overwriting `LESS` to remove the `-X` option.

@larsxschneider the issue is with scrolling with a mouse/trackpad in case that wasn't clear.

larsxschneider:
Hmm. Did you install the iTerm2 shell integration? I did not -- maybe that is the difference? But even if you did, I wonder how that could affect the Apple Terminal app on your end 🤔

Chris911:
Looks to me like with `-X` less is no longer recognized as an alternate screen program. Both Terminal and iTerm have specific scroll settings for what they call "alternate screen mode" (eg. editors like vim, nano). Because `-X` is used to not clear the screen less doesn't open in an "alternate screen" which in turn disables the scrolling features which send page up/down signals to the shell instead of scrolling the app itself. 

**iTerm**: 
screen shot 2018-07-31 at 10 34 43 am

**Terminal**: 
screen shot 2018-07-31 at 10 35 01 am

mcornella:
Reverted on 106f826075979ef1a6875cedd2d098e601f2e3f2. Thanks for reporting this use case, I wasn't aware of this because I never used mouse scrolling on less.

Unfortunately, both `-F` and `-X` have to be removed because, even if `-F` is set, less won't automatically cat the input, so it ends up not showing anything.

Thanks again!

P-Seebauer:
also does not work nice when you're using some window manager and are about to change layout. Not to happy about this, am also unsetting it in my .zshrc for now.

PR #6615, opened 2018-02-10
Add alias glr for git log --reverse.

DevotionGeo:
I wanted to add another one the `gl` for `git log`, but that's already taken for `git pull`, so I added the one I use often when I want to go to the starting commits of some repository which is `git log --reverse` for which `glr` is a nice alias, which isn't taken already.


mcornella:
This is not so commonly used that it needs an alias in OMZ. If you use it often, you can add it to your zshrc file. Thanks

PR #6631, opened 2018-02-19
Improve hg prompt

MightyCreak:
I made sure that each commit is for a very specific feature and each of them have a full description of what they do.

The main changes are:
* Support mercurial in bira theme
* Improve mercurial prompt function speed


MightyCreak:
@berendkleinhaneveld this is the PR that replaced the old one you reviewed (#5613).

It is now up to date with the latest upstream master, has an appropriate name, and has much more interesting features!

PR #6637, opened 2018-02-21
Respect RBENV_ROOT and PYENV_ROOT if set (quick merge for rbenv and pyenv plugins)

razor-x:
If `RBENV_ROOT` or `PYENV_ROOT` have been set, assume this is intentional by the user and put that path at the front of `rbenvdirs` or `pyenvdirs`. :smile_cat: 

The current (hostile) behaviour ignores these  officially documented env vars.

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -9,6 +9,10 @@ _pyenv-from-homebrew-installed() {
 FOUND_PYENV=0
 pyenvdirs=("$HOME/.pyenv" "/usr/local/pyenv" "/opt/pyenv")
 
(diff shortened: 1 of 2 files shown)


mcornella:
No longer needed in the current version. Thanks!

razor-x:
@mcornella I see the pyenv update you made. I will can test if this works correctly with `PYENV_ROOT`, but I don't any update for rbenv.

mcornella:
You're right, for `rbenv` that might still be needed.

razor-x:
@mcornella I can now confirm that using the new pyenv plugin **completely breaks pyenv** for me. It does not find pyenv with `PYENV_ROOT` set and it does not find it when installed to the "default" location `~/.pyenv`.

What is the intended behaviour and use for this plugin now?

PR #6640, opened 2018-02-23
add alias grmi, rm files from ignore file

cos120:
add alias `grmi`
rm files but still tracked. it's useful when update ignore file.


mcornella:
Too niche and too infrequent. You can put it in your zshrc file though.

PR #6646, opened 2018-02-27
Better read command syntax

dschwartz783:
This syntax is more concise. The read command can be placed into question mode, which causes it to behave like a normal Y/n prompt. The read command, in question mode, will also return 0 if the user type y (or Y), which allows it to be used directly in an if statement.


mcornella:
`read -q` has the behavior of defaulting to `n` if no `y` or `Y` is provided, which is the opposite of what the prompt says.

dschwartz783:
This was a long time ago. Don't even think it's a good idea anymore, so...

PR #6649, opened 2018-02-28
Add 'gs' As Additional Shortcut For 'git status'

sfawaz:


raybotha:
+1 I always add this alias manually.

hjpotter92:
`gs` is also ghostscript cli.

```
➜ gs
GPL Ghostscript 9.20 (2016-09-26)
Copyright (C) 2016 Artifex Software, Inc.  All rights reserved.
This software comes with NO WARRANTY: see the file PUBLIC for details.
GS>
```

mcornella:
As @hjpotter92 said, we don't use `gs` so as not to override ghostscript. You can add custom aliases to your zshrc file or in a custom plugin.

PR #6652, opened 2018-03-01
Add different cursors for normal and insert mode

Hvalfisk:


robbyrussell:
We'll close this given how long it's been pending more details for quite some time.

Issue #6036, opened 2017-04-20
Mercurial issue with rkj-repos theme


AleksNovak:
Hi,
Since I installed mercurial package on my system I had a issue with hg.

### Issue
```
┌─[me@clearbox] - [~] - [2017-04-20 12:37:20]
└─[0] <Mercurial Distributed SCM

basic commands:

 add           add the specified files on the next commit
 annotate      show changeset information by line for each file
 clone         make a copy of an existing repository
 commit        commit the specified files or all outstanding changes
 diff          diff repository (or selected files)
 export        dump the header and diffs for one or more changesets
 forget        forget the specified files on the next commit
 init          create a new repository in the given directory
 log           show revision history of entire repository or files
 merge         merge another revision into working directory
 pull          pull changes from the specified source
 push          push changes to the specified destination
 remove        remove the specified files on the next commit
 serve         start stand-alone webserver
 status        show changed files in the working directory
 summary       summarize working directory state
 update        update working directory (or switch revisions)

(use 'hg help' for the full list of commands or 'hg -v' for details)>
```

### Code
```
function hg_prompt_info {
hg prompt --angle-brackets "\
<hg:%{$fg[magenta]%}<branch>%{$reset_color%}><:%{$fg[magenta]%}<bookmark>%{$reset_color%}>\
</%{$fg[yellow]%}<tags|%{$reset_color%}, %{$fg[yellow]%}>%{$reset_color%}>\
%{$fg[red]%}<status|modified|unknown><update>%{$reset_color%}<
patches: <patches|join( → )|pre_applied(%{$fg[yellow]%})|post_applied(%{$reset_color%})|pre_unapplied(%{$fg_bold[black]%})|post_unapplied(%{$reset_   ↪color%})>>" 2>/dev/null
}
PROMPT=$'%{\e[0;34m%}%B┌─[%b%{\e[0m%}%{\e[1;32m%}%n%{\e[1;30m%}@%{\e[0m%}%{\e[0;36m%}%m%{\e[0;34m%}%B]%b%{\e[0m%} - %b%{\e[0;34m%}%B[%b%{\e[1;37m%}   ↪%~%{\e[0;34m%}%B]%b%{\e[0m%} - %{\e[0;34m%}%B[%b%{\e[0;33m%}'%D{"%Y-%m-%d %I:%M:%S"}%b$'%{\e[0;34m%}%B]%b%{\e[0m%}
[...]

PR #2068, opened 2013-08-29
Merge rkj & rkj-repos themes and add some comments to them. Add git-note completion plugin.

rkj:

PR #6655, opened 2018-03-03
[rkj-repos] Check for 'hg prompt' and exit if not found

cori:
Doesn't "fix" it per se, but tells the user how to fix.

Might be impacted by PR #2068

Solves #6036.


Lavoaster:
Could this instead return an empty string in the `hg_prompt_info` function? Alternatively, should it only exit if mercurial is installed? 

I don't have mercurial installed and this update has broken the theme for me.

cori:
That's a good point that I hadn't taken into consideration. Can you post an issue and I'll send another PR in that patches this, since this PR's already merged and closed.

PR #6658, opened 2018-03-04
Fix git_commits_{ahead,before} when no upstream branch is defined

### LINKED ISSUE #6656: git_commits_ahead echoes an error message if no upstream has been set for the current branch
I find myself using git_commits_ahead n my prompt; when creating a local branch without upstream the prompt is constantly filled with errors such as

fatal: no upstream configured for branch 'branch-without-upstream'

git-version 2.16.1

wilderjds:
If @{u} is not defined, git rev-list will give an error; redirect
to stderr the error and deal with this case in what follows.

This fixes #6656

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -77,8 +77,8 @@ function git_current_branch() {
 # Gets the number of commits ahead from remote
 function git_commits_ahead() {
   if command git rev-parse --git-dir &>/dev/null; then
-    local commits="$(git rev-list --count @{upstream}..HEAD)"
-    if [[ "$commits" != 0 ]]; then
+    local commits="$(git rev-list --count @{upstream}..HEAD 2>/dev/null)"
+    if [[ -n "$commits" && "$commits" != 0 ]]; then
       echo "$ZSH_THEME_GIT_COMMITS_AHEAD_PREFIX$commits$ZSH_THEME_GIT_COMMITS_AHEAD_SUFFIX"
     fi
   fi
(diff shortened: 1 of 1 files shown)


lib/git.zsh
```diff
@@ -77,8 +77,8 @@ function git_current_branch() {
 # Gets the number of commits ahead from remote
 function git_commits_ahead() {
   if command git rev-parse --git-dir &>/dev/null; then
-    local commits="$(git rev-list --count @{upstream}..HEAD)"
-    if [[ "$commits" != 0 ]]; then
+    local commits="$(git rev-list --count @{upstream}..HEAD 2>/dev/null)"
+    if [[ -n $commits ]] && [[ "$commits" != 0 ]]; then
```
Shea690901: Any special reason you don't write:
``` if [[ -n "$commits" -a "$commits" != 0 ]]``` ?

IMHO putting the '"' around any tested environment variable is **always** a very good idea and using '-a' instead of terminating one test and running after that another one should speed up the thing a little...

Shea690901: Two issues:
1. You should **always** soround any environment variable with '"' when testing, otherwise you might accidently execute:
```[[ -n ]]``` which is an syntax error...
1. Why are you using two tests and combing there results instead of using just a single test testing both conditions?
```[[ -n "$commits" -a "$commits" != "0" ]]``` or
```[[ -n "$commits" -a $commits -ne 0 ]]```
Don't know which one is faster ;)

Shea690901: Btw.:
```[[ "$commits" != "0" ]]``` alone just works fine for this case ;)

wilderjds:
> Btw.:
> ```[[ "$commits" != "0" ]]``` alone just works fine for this case ;)
For some reason it does not work for me; the if clause is executed if $commits 
is empty

wilderjds:
On Monday, March 5, 2018 8:47:34 PM EST Shea690901 wrote:
> Any special reason you don't write:
> ``` if [[ -n "$commits" -a "$commits" != 0 ]]``` ?

possibly my ignorance; however the above does not work for me. I changed it to
```if [[ -n "$commits" && "$commits" != 0 ]]``` 
which seems to do the trick

Shea690901:
Concerning
> Any special reason you don't write: ``` if [[ -n "$commits" -a "$commits" != 0 ]]``` ?

~~Strange...
While, for easier reference, I just looked into ```man test```, I also tested this (just to be sure, I didn't make any mistakes).
I'm using zsh 5.3 ...~~

Ouch, just found my mistake: ```[[``` uses a little different syntax, during my tests I used ```if [ ...``` instead of ```if [[ ...```...

Concerning
> ```[[ "$commits" != "0" ]]``` alone just works fine for this case ;)

I typed a little too fast and send before completely thinking through **what** I had typed, as far as I was concerned from my side I had deleted this comment because of it's error, at least it vanished from my view...

mcornella:
Thanks for the PR!

PR #6664, opened 2018-03-06
composer: Fix slowness during startup

### LINKED ISSUE #6647: composer plugin is slow on startup
I found that it takes more than 300ms on my mac to load composer plugin.

As a result of further examination, the composer command which is calling for path setting consumed most of the time.

```zsh
# Add Composer's global binaries to PATH
export PATH=$PATH:$(composer global config bin-dir --absolute 2>/dev/null)
```

```console
% time composer global config bin-dir --absolute >/dev/null 2>&1
composer global config bin-dir --absolute > /dev/null 2>&1  0.18s user 0.09s system 88% cpu 0.301 total
```

Since the output of `composer global config bin-dir --absolute` command does not change so frequently, I think that loading time can be shortened by caching.

stucki:
With the merge of #5886, oh-my-zsh sometimes takes up to several seconds to load (see #6647).

This is caused by the `composer global config bin-dir --absolute 2>/dev/null` command which takes a lot of time to load on some systems.

The most simple solution would be to revert #5886. If that won't happen, here is a patch which caches the output of that command. The patch will not have any effect unless caching is enabled in `.zshrc`:

    zstyle ':completion:*' use-cache yes

Fixes #6647


Frederick888:
On top of the cache, how about adding something like
```
alias composer='php -d xdebug.remote_enable=0 -d xdebug.profiler_enable=0 composer'
```
...to disable xdebug when running composers? Currently every time I open a new terminal it connects to my IDE which blocks the startup process. (No idea why but if I run the command later manually it never loads xdebug, which is the correct behaviour as described in https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer.)

I have the alias in .zshrc before oh-my-zsh is loaded and it seems to work perfectly.

RobLoach:
Using a cache makes sense to me. The xdebug addition sounds nice, would do well in a separate PR.

CyrilMazur:
In my case the culprit was svn 🤷‍♂️ Turns out composer calls the `svn info --xml` command at startup, and svn is not installed on Catalina by default (and I don't use svn myself) and it took 1-2s for the OS to return. I installed svn via `brew install svn` and now composer is fast again.

PR #6671, opened 2018-03-12
add nvmng plugin

wzrdtales:
This adds the [nvmng](https://github.com/wzrdtales/nvm-ng). plugin, which loads nvm with the flag --fast-reuse, which is exposed
by nvm-ng. This is a hard fork of nvm which contains patches that accelerate load time
for nvm, to get back to a more smooth behavior of your shell.

nvm-ng can be found here

https://github.com/wzrdtales/nvm-ng

Signed-off-by: Tobias Gurtzick <magic@wizardtales.com>

### DIFF
--- a/plugins/nvmng/README.md
+++ b/plugins/nvmng/README.md
@@ -1,0 +1,10 @@
+# Description
+
+This plugin loads the nvm-ng hard fork of nvm, with the --fast-reuse patch
+applied. This improves loading times for your shell significantly.
+
+# Installation
+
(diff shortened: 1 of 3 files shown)


wzrdtales:
ping

mcornella:
Has no interest. It also seems to just do the same as the `nvm` plugin, with the difference that you just need to install nvm-ng instead of nvm.

wzrdtales:
@mcornella Yes that is obviously the case. But the difference is, that this initializes it different than the nvm plugin (it adds an options flag). otherwise the nvm plugin could have been used.

PR #6678, opened 2018-03-18
Change the "br" action to "branch" in the jira plugin to be more consistent with other actions

ehudon:
# Context

The `br` action is the only action that is used as an abbreviation. 

# Suggestion Solution

This PR aims at:

1. Change the `br` action to `branch` to be aligned with other actions which don't use appreviations and also to be more consistent with the name of other actions.

2. Add the `branch` action in the list of possible actions in `README.md` which was not previously listed.


robbyrussell:

Cheers,
Robby

PR #6696, opened 2018-03-27
Added a myissues option

zopanix:
This will allow the user to directly open the my open issues page. This
will make it easier  for the user to open his open issues in jira.


mcornella:
Thanks!

PR #6698, opened 2018-03-28
Plugin for pipenv

### LINKED ISSUE #6679: plugin for pipenv
[pipenv][1] has command completion for zsh, having a plugin would make it easy and clean to include this in a zshrc file. I've made an attempt to create the plugin at https://github.com/gangleri/oh-my-zsh/tree/pipenv_plugin/plugins/pipenv would this be useful to anyone else?

[1]: https://github.com/pypa/pipenv

gangleri:
Add completion for pipenv
Auto activate and deactivate for pipenv shell

fixes #6679


cassianoleal:
Would it be useful if it tried the parent directories if a `Pipfile` is not found in `$CWD`?

To me this is one of the main features lacking in Pipenv itself. Presently what that means is if I `cd` into a project's subdirectory I can't easily get into my Pipenv shell.

metakermit:
Tried this branch out and seems to work ok in my opinion. You can navigate around children directories and Pipenv stays open, while navigating to a different project closes the Pipenv shell. It is a bit confusing that the outer shell will be on a different path (the root of the project), but that's got more to do with how Pipenv works – opening a new shell instead of sourcing.

austinmcconnell:
Any chance this might get merged in soon? I'd love to have this functionality.

owenstranathan:
I just wrote my own plugin for this and was going to submit a PR before I found this one. 👍 for a merge.
If you're curious mine is [here](https://github.com/owenstranathan/pipenv.zsh)

AlexGascon:
Hi @gangleri! I've also developed a pipenv plugin (https://github.com/AlexGascon/pipenv-oh-my-zsh). Basically, it adds aliases for the most common pipenv commands.

I was thinking of opening a PR, but I don't think that it would make sense to have two different pipenv-plugin PRs. Would you mind if I submit a PR with my changes to your repo, so everything gets included here?

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+_pipenv() {
+  eval $(env COMMANDLINE="${words[1,$CURRENT]}" _PIPENV_COMPLETE=complete-zsh  pipenv)
+}
+
+_togglePipenvShell() {
+  # deactivate shell if Pipfile doesn't exit and not in a sub dir
+  if [[ ! -a "$PWD/Pipfile" ]]; then
+      if [[ "$PIPENV_ACTIVE" == 1 ]]; then
+          if [[ "$PWD" != "$pipfile_dir"* ]]; then
+            exit
```
eliwind: Perhaps `cd -` here, so that the parent shell changes back to where it was before cd-ing into the directory with a Pipfile.  Otherwise, cd-ing out of a pipenv project directory exits the `pipenv shell` but the parent shell is still inside the pipenv project directory.

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+  autoload -U compinit && compinit
+  compdef _pipenv pipenv
+fi
+
+export PROMPT_COMMAND=_togglePipenvShell
+if [ -n "$ZSH_VERSION" ]; then
+  function chpwd() {
+    _togglePipenvShell
+  }
+fi
```
eliwind: Is `add-zsh-hook chpwd _togglePipenvShell` preferable to redefining `chpwd`?

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+alias pl="pipenv lock"
+alias po="pipenv open"
+alias pr="pipenv run"
+alias psh="pipenv shell"
+alias psy="pipenv sync"
+alias pu="pipenv uninstall"
+
+alias pwh="pipenv --where"
+alias pv="pipenv --venv"
+alias ppy="pipenv --py"
```
eliwind: Maybe make these aliases configurable somehow, or at least add something to disable them?  Otherwise, they're adding to the global namespace of aliases, and people may well have these commands already meaning something else.

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+            exit
+          fi
+    fi
+  fi
+
+  # activate the shell if Pipfile exists
+  if [[ "$PIPENV_ACTIVE" != 1 ]]; then
+    if [[ -a "$PWD/Pipfile" ]]; then
+      export pipfile_dir="$PWD"
+      pipenv shell
```
eliwind: Perhaps cd - here, so that the parent shell changes back to where it was before cd-ing into the directory with a Pipfile. Otherwise, cd-ing out of a pipenv project directory exits the pipenv shell but the parent shell is still inside the pipenv project directory.

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+  fi
+}
+
+if [[ "$(basename ${(%):-%x})" != "_pipenv" ]]; then
+  autoload -U compinit && compinit
+  compdef _pipenv pipenv
+fi
+
+export PROMPT_COMMAND=_togglePipenvShell
+if [ -n "$ZSH_VERSION" ]; then
```
mcornella: You don't need to do this, the plugin will be executed by zsh so `$ZSH_VERSION` will always be set.

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+    fi
+  fi
+}
+
+if [[ "$(basename ${(%):-%x})" != "_pipenv" ]]; then
+  autoload -U compinit && compinit
+  compdef _pipenv pipenv
+fi
+
+export PROMPT_COMMAND=_togglePipenvShell
```
mcornella: `PROMPT_COMMAND` is useless in zsh. What did you intended this to do? If you wanted this to run before every prompt, you should put it in `precmd_functions`, like so: `precmd_functions+=(_togglePipenvShell)`. I'm not sure you need it though, if you already have it as a `chpwd` hook.

plugins/pipenv/pipenv.plugin.zsh
mcornella: `chpwd_functions+=(_togglePipenvShell)` is preferred.

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -0,0 +1,51 @@
...
+  if [[ "$PIPENV_ACTIVE" != 1 ]]; then
+    if [[ -a "$PWD/Pipfile" ]]; then
+      export pipfile_dir="$PWD"
+      pipenv shell
+    fi
+  fi
+}
+
+if [[ "$(basename ${(%):-%x})" != "_pipenv" ]]; then
+  autoload -U compinit && compinit
```
mcornella: This is already done by OMZ, you don't need to call compinit again. Also, the check for file name above (`if [[ $(basename ...`) is superfluous because it will always return `pipenv.plugin.zsh`, hence it will always execute.

I would just leave the `compdef _pipenv pipenv` and put it right below the `_pipenv` function.

plugins/pipenv/pipenv.plugin.zsh
mcornella: I'm not worried about aliases being defined, but make sure that they don't clash with common commands. For example, [`pv` is a popular command](https://linux.die.net/man/1/pv) to monitor pipes.

AlexGascon:
@mcornella  I've submitted a PR to @gangleri's repo to fix the aliases that I think that can be more problematic (`pr` --> `prun` and `pv` --> `pvenv`) and to add a README.md, in case you want to take a look there and suggest further changes before they get merged. 

About the automatic activation/deactivation part, I'll leave those changes to @gangleri as he's who coded the original feature.

owenstranathan:
@rholmboe I’m not sure what’s going on with compinit but the deactivate needing you to `cd` twice is a consequence of pipenv’s implementation and the way shells work. Under the hood pipenv spawns an interactive shell inside of a running python interpreter. And when you stop the shell you actually fall back into a parent shell that spawned it. So although this plugin’s deactivation on `cd` functionality does change your directory It’s futile because on killing the sub shell that it changes directory into it puts you the user back in the shell that spawned the subshell which is currently in the directory that spawned the `pipenv shell `

For example do the following
``` 
$ me@computer:~ $ zsh # this spawns a subshell.  
$ me@computer:~ $ cd /etc # in subshell go to another dir 
$ me@computer:/etc $ ^D # press ctrl + d to kill the sub shell
$ me@computer:~ $ # I’m back at home

```

So with that said I think the auto deactivate is ill fated and  should probably be removed in favor of users explicitly deactivating sub shells


<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

lanceofwhichwedream:
@gangleri has there been any movement on this? I'd love to see this plugin accepted into the master branch. Is there anything I could assist you with?

mcornella:
I've applied the requested changes. Now this is much cleaner and it should work equally well.

Regarding the automatic activation/deactivation, you can for now add a hack in your zshrc file to redefine the `_togglePipenvShell` function with an empty body, and it will not execute:
```
_togglePipenvShell() {}
```

Let me know if this is ready to be merged.

ajilty:
@mcornella tested and working nicely. I had to use `--allow-unrelated-histories --strategy-option ours` to test the merge. Good to go?

tschaub:
> You can navigate around children directories and Pipenv stays open, while navigating to a different project closes the Pipenv shell. It is a bit confusing that the outer shell will be on a different path (the root of the project)

#9206 proposes a way to exit the pipenv shell and change to the target directory when `cd`ing out of a `Pipfile` containing directory.

PR #6699, opened 2018-03-29
add support for per-theme subfolder

TamCore:
This will allow it to git-clone third-party themes into $ZSH_CUSTOM/themes/$ZSH_THEME and therefore make it somewhat easier to maintain


mcornella:
Wontfix. Current behavior can mimic this with a symlink.

PR #6702, opened 2018-03-29
Add Google Flutter auto-completion plugin

sashko:


febs:
How do I install this manually?
Thanks

sashko:
@febs you just copy file to `~/.oh-my-zsh/plugins/`

sashko:
Updated plugin for Flutter 0.8.2.

mcornella:
Wait, I was going to merge it, but there's a richer completion file over at [zsh-completions](https://github.com/zsh-users/zsh-completions/blob/master/src/_flutter), I highly recommend using that instead. Thanks for your contribution!

PR #6703, opened 2018-03-30
Allow specifying a group name for the tmux session

bsiegel:
Allow specifying a `ZSH_TMUX_GROUP` which, if set, will be used as the session group name when invoking `attach` or `new-session`.


larson-carter:
@bsiegel Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

bsiegel:
@larson-carter Done, looks like things have changed quite a bit since I wrote the initial PR, so I reimplemented it.

mcornella:
Can we make it into a single setting, default being empty string? Then, to check if it's not an empty string use `[[ -n "$var" ]]`. Also there was a recent push as well so it needs a rebase. Thanks!

qjzTreeZ:
This PR will cause an error because it puts `-t "$ZSH_TMUX_SESSIONGROUP_NAME"` before `new-session`. (#8397)

PR #6716, opened 2018-04-09
Update archlinux.plugin.zsh

dalu:
-alias pacfiles='pacman tFs'
+alias pacfiles='pacman -Fs'

pacman tFs makes no sense. I assume who ever wrote this meant `-Fs`


mcornella:
I see #5909 was opened long ago. Thanks for bringing this to our attention though!

PR #6719, opened 2018-04-10
move ~/.zsh-update file to $ZSH_CACHE_DIR

### LINKED ISSUE #4550: Suggestion for better update checking

pcr910303:
Fixes #4550. The default path of file .zsh-update is changed to $ZSH_CACHE_DIR.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -7,10 +7,13 @@ function _current_epoch() {
 }
 
 function _update_zsh_update() {
-  echo "LAST_EPOCH=$(_current_epoch)" >! ~/.zsh-update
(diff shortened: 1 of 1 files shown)


mcornella:
Duplicate of #4571

PR #6726, opened 2018-04-13
Add aliases for helm plugin

alzubaidi:
Add helm auto complete support based on helm [documentation](https://docs.helm.sh/helm/)


plugins/helm/helm.plugin.zsh
```diff
@@ -1,7 +1,85 @@
 # Autocompletion for helm.
 #
-# Copy from kubectl : https://github.com/pstadler
+# Author: https://github.com/alzubaidi
```
kenden: https://github.com/fleeto is also an author of this file

alzubaidi:
I used it quit often when I worked with K8 to get some app deployed but once done, it will not be used this much. I thought it would be nice addition for DevOps people maybe.

mcornella:
I'll reject this PR because it only adds aliases and of commands that won't be used frequently. Thanks for your interest

PR #6728, opened 2018-04-14
add a year progress bar plugin

eightpigs:
Looks like this: 

28% ▓▓▓▓▓▓▓▓▓░░░░░░░░░░░░░░░░░░░░░░░ 104/365

### DIFF
--- a/plugins/year-progress/year-progress.plugin.zsh
+++ b/plugins/year-progress/year-progress.plugin.zsh
@@ -14,15 +14,11 @@ year_progress() {
     # progress=current/year
     value=`echo "$current $len" | awk '{printf ("%.2f\n",$1/$2)}'`
 
-    # Get the number of columns in the terminal
-    tput init 
-    cols=`tput cols`
-
     val2=$(($current*100/$len))
 
     info=$val2%'  '$current/$len
 
-    cols=$(($cols-${#info}))
+    cols=$(($COLUMNS-${#info}))
 
     # Fill the proportion of the screen
     scale=`echo "$cols" | awk '{printf ("%.2f\n",$1/100)}'`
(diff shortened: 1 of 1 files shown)


plugins/year-progress/year-progress.plugin.zsh
```diff
@@ -0,0 +1,50 @@
...
+
+    # leap year: 366
+    if [[ ($(($year%4)) == 0 && $(($year%100)) != 0) || $(($year%400)) == 0 ]] ;then
+        len=366
+    fi
+
+    # progress=current/year
+    value=`echo "$current $len" | awk '{printf ("%.2f\n",$1/$2)}'`
+
+    val2=$(($current*100/$len))
```
ehudon: I think the code would read better if `val2` would be renamed to `percent`.

plugins/year-progress/year-progress.plugin.zsh
```diff
@@ -0,0 +1,50 @@
...
+    echo -n $val2%' '
+
+    # Previous days
+    for ((i=0; i<$val; i++))  
+    do
+        echo -n "▓"
+    done
+
+    # The remaining days
+    for ((i=0; i<$((cols-$val)); i ++))
```
ehudon: Replace `i ++` by `i++` to be consistent with the loop just above.

mcornella:
I'm sorry but at some point we have to start rejecting plugins given the vast amount of them that already come prebundled. You can use your plugin as a custom plugin looking at the [Customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization) wiki or even make a dedicated github repository and adding it to the list of [external plugins](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins). Thanks for your understanding.

PR #6737, opened 2018-04-18
Add ZSH_TMUX_UTF8 option to tmux plugin

samogot:
@jmatth


samogot:
oops. looks like it is quite popular PR. It is partial duplication of: #5776, #5166, #2023

larson-carter:
@samogot Can you please rebase this PR onto our latest master branch. Then update this PR with a Commit please? Thank You!

robbyrussell:
Thanks for this, and sorry it sat for so long without a reply.

Oh My Zsh already has this as `ZSH_TMUX_UNICODE`, which passes `-u` to tmux the same way. It landed in #5166 in 2016, but it wasn't documented in the README until 2019, which is probably why it was easy to miss.

```zsh
ZSH_TMUX_UNICODE=true
```

Closing this one for that reason. Thanks again.

PR #6739, opened 2018-04-18
Create cordova.plugin.zsh

navelpluisje:
Created a start for and added some default aliases for the use with cordove


mcornella:
Has no interest.

PR #6745, opened 2018-04-19
Update AWS completion

hoshsadiq:
Using `which` to determine the path `aws_zsh_completer.sh` is vastly faster than using brew to determine the base path first. Often the completer is already in your PATH, so it makes more sense to check which first.

Also added additional check for in case the file isn't found.

I can't give you comparison benchmarks, as which is a shell builtin, and time doesn't show the times:
```
$ time brew --prefix awscli
/usr/local/Cellar/awscli/1.15.0
brew --prefix awscli  0.33s user 0.15s system 96% cpu 0.499 total

$ time which aws_zsh_completer.sh
/usr/local/bin/aws_zsh_completer.sh

$ type -f which
which is a shell builtin
```

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -42,11 +42,11 @@ function aws_profiles {
 
 compctl -K aws_profiles asp
 
-if _homebrew-installed && _awscli-homebrew-installed ; then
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks @hoshsadiq!

PR #6746, opened 2018-04-19
[rkj-repos] Make `hg prompt` check less strict

### LINKED ISSUE #6743: rkj-repos theme now depends on mercurial / hg prompt to be installed to work
As per #6655, the theme was updated to show information on how to enable hg prompt support. However, this has now meant that users that don't have mercurial installed (👋Hii) can no longer use this theme without installing both mercurial and hg prompt.

There are few possible solutions to this:
- Exit, only if mercurial is installed and hg prompt is not found.
- Return an empty string in the `hg_prompt_info` if `hg prompt` is not found.

mcornella:
Move the `hg prompt` check inside the hg_prompt_info function so that
it returns an empty string if hg-prompt isn't installed.

Fixes #6743.

cc @Lavoaster @cori

### DIFF
--- a/themes/rkj-repos.zsh-theme
+++ b/themes/rkj-repos.zsh-theme
@@ -1,7 +1,7 @@
 # user, host, full path, and time/date on two lines for easier vgrepping
 
 function hg_prompt_info {
-  if grep -q "prompt" ~/.hgrc; then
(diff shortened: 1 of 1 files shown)


Lavoaster:
@mcornella Looks like @cori gave it a 👍 reaction. (This is just so you get a notification)

PR #6046, opened 2017-04-25
Add completions to asdf plugin

sphinxc0re:

PR #6749, opened 2018-04-20
Adds Homebrew and completion support to asdf plugin

jburwell:
* Modifies the search logic for asdf to include Homebrew when it is
installed.  The implementation is adapted from the pyenv plugin.

* Includes command completions for both regular and Homebrew installs.
This enhancement was inspired by PR #6046.

### DIFF
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -1,33 +1,17 @@
-_homebrew-installed() {
-    type brew &> /dev/null
-}
+# Find where asdf should be installed
+ASDF_DIR="${ASDF_DIR:-$HOME/.asdf}"
 
-FOUND_ASDF=0
-DEFAULT_ASDF_DIR="$HOME/.asdf"
-
-if [ -d $DEFAULT_ASDF_DIR/bin ] ; then
-    FOUND_ASDF=1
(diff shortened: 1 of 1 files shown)


mcornella:
For what I understand, the only places that asdf can be in is either `$ASDF_DIR` or in whatever path Homebrew installs it, so there's no need to check all these directories: just checking the Homebrew path if the other wasn't found should be enough.

Also, if you installed it with homebrew you can probably get what directory is installed in using `$commands[asdf.sh]` or something similar. But that depends on how Homebrew makes the install. What does `brew --prefix asdf` return?

jburwell:
@mcornella my original thinking around the search path was to follow pyenv's approach.  Based on your feedback, I realized that these additional search paths were present to pyenv when it may be installed as part of distribution packages.  Since asdf does not appear to be packaged by any distros, this search logic is unnecessary.  Therefore, I have updated the patch to search ``$HOME/.asdf`` and the Homebrew location (if Homebrew is installed).

Also, as part of this update, I extracted the common bits of asdf startup into a common code block.

Finally, ``brew --prefix asdf`` provides the root directory of the asdf installation under Homebrew.  This approach avoid hard coding the root path or assuming the directory layout used by the asdf Homebrew package.

mcornella:
I've made a number of changes that simplify the logic, and also fixes the paths that you were looking for in the case of Homebrew installations.

PR #6580, opened 2018-01-27
Fine-tune dirhistory plugin mappings for Mac

peppelan:
This change addresses side-effects of PR #6533 experienced on Mac terminals other than the default provided with OSX.
See the exchange with @NAR8789 attached to PR #6533 for more details.

PR #6756, opened 2018-04-23
Add directory hierarchy navigation to dirhistory

fzuviria:
Changes to be committed:
	modified:   dirhistory/dirhistory.plugin.zsh

New Feature:
  Navigate directory hierarchy using ALT-UP and ALT-DOWN. (mac keybindings not yet implemented)
  ALT-UP moves to higher hierarchy (cd ..)
  ALT-DOWN moves into the first directory found in alphabetical order

### DIFF
--- a/plugins/dirhistory/dirhistory.plugin.zsh
+++ b/plugins/dirhistory/dirhistory.plugin.zsh
@@ -2,6 +2,10 @@
 #   Navigate directory history using ALT-LEFT and ALT-RIGHT. ALT-LEFT moves back to directories 
 #   that the user has changed to in the past, and ALT-RIGHT undoes ALT-LEFT.
 # 
+#   Navigate directory hierarchy using ALT-UP and ALT-DOWN. (mac keybindings not yet implemented)
+#   ALT-UP moves to higher hierarchy (cd ..)
+#   ALT-DOWN moves into the first directory found in alphabetical order
+#
 
 dirhistory_past=($PWD)
 dirhistory_future=()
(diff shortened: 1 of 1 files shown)


fzuviria:
@jeffwilliams Hey jeff, awesome plugin
Im adding two more keybingings, check it out, tell me what you think

jeffwilliams:
Thanks, fzuviria, this is a nice improvement. 

I don't have a mac, but I wonder if @peppelan would be able to add the mac keybindings for it after commit?

peppelan:
I'll do that with great pleasure. +1 and thanks for the feature :-)

fzuviria:
hmm, maybe I should update the doc in the plugin's site as well. Is that versioned? Where do I look for? First time contributor >_<

peppelan:
For the Mac mappings, I couldn't manage to find a working binding.

While the original bindings committed by @fzuviria works like a charm on iTerm2 (which I use most of the time), I could not find a specific mapping that works for Apple Terminal (like I did in PR #6580).

In Apple Terminal, typing `opt+up` in `cat > /dev/null` outputs the `^[^[[A` code, which doesn't do the magic when passed to `bindkey`.

On the net, I also found other alternatives, like `\e\e[A`, `^[^[0A`, but none of them is working 😞

peppelan:
(PS: thanks for the change, I was missing a lot the alt+up especially)

jeffwilliams:
@fzuviria To update the documentation I think you just need to go to the wiki and edit the page when you're logged in (click the edit button on the top right):

https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins

It gets versioned automatically.

jeffwilliams:
@peppelan Thanks for looking into it. Too bad about the Apple Terminal bindings, but having iTerm2 is good.

fzuviria:
So, what should be done now?
I checked the activity of last week and P.R.s newer than this are getting merged. Am I missing something?
BTW, small changes to adhere to coding style

mcornella:
Thanks!

PR #6763, opened 2018-04-24
Chuck Norris depends on nobody

povesteam:


mcornella:
Much as I appreciate the humor, I predict some users would find that message confusing.

PR #6775, opened 2018-04-26
gpg-agent Don't set SSH_AUTH_SOCK if ssh-agent plugin is enabled

### LINKED ISSUE #6772: PR #6469 appears to have broken gpg-agent plugin checks
#6469 appears to have broken the check on my machine:

OS: macOS 10.13.4  
GPG Version: MacGPG2 2.2.3

My gpg agent seems to create `S.gpg-agent.ssh` without being configured to do so. The ssh-agent plugin then does not create the `SSH_AUTH_SOCK` environment variable, breaking the agent.

Rolling back to 37c2d0d appears to have resolved it for me locally.

Starting the agent:
```
gpg-agent --daemon --use-standard-socket
gpg-agent[54927]: WARNING: "--use-standard-socket" is an obsolete option - it has no effect
gpg-agent[54928]: gpg-agent (GnuPG/MacGPG2) 2.2.3 started
```
```
~/.gnupg  find . -type s
./S.gpg-agent.extra
./S.gpg-agent.ssh
./S.dirmngr
./S.gpg-agent
./S.gpg-agent.browser
```
```
~/.gnupg  cat gpg-agent.conf
default-cache-ttl 600
max-cache-ttl 7200
```

chrisfleming:
fixes #6772 that was introduced in #6469 

Don't setup  SSH_AUTH_SOCK if the ssh-agent plugin is enabled.

### DIFF
--- a/plugins/gpg-agent/gpg-agent.plugin.zsh
+++ b/plugins/gpg-agent/gpg-agent.plugin.zsh
@@ -1,7 +1,5 @@
 # Enable gpg-agent if it is not running-
 # --use-standard-socket will work from version 2 upwards
-typeset _gpg_ssh_socket
-
 AGENT_SOCK=`gpgconf --list-dirs | grep agent-socket | cut -d : -f 2`
 
 if [ ! -S ${AGENT_SOCK} ]; then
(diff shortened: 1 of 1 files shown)


mcornella:
I was ready to merge this one but upon reading and researching, I think we should go back to checking if enable-ssh-support is set in the config file. I've yet to understand why in #6772 the `gpg-agent` was creating the ssh socket though.

PR #6776, opened 2018-04-27
[lib/git.zsh] Add git_commit_status

shtrom:
Add a simple function, `git_commit_status` that summarise the status of the current branch with respect to upstream, e.g.,

    u+2-3

For a branch two commits ahead and three behind.

Signed-off-by: Olivier Mehani <shtrom@ssji.net>

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -112,6 +112,14 @@ function git_commits_behind() {
   fi
 }
 
+function git_commits_status() {
+  local ahead=$(git_commits_ahead)
+  local behind=$(git_commits_behind)
+  if [[ -n "$ahead$behind" ]]; then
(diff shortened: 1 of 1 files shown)


shtrom:
@mcornella conflict fixed by removing the `no_optional_lock` commit. It's now handled by `__git_prompt_git` from `master`.

robbyrussell:
Thanks for this, and sorry... you rebased this yourself back in 2020 and still nobody answered you. That's not great.

I'm going to close it, but not because it's wrong. It's that nothing in the repo calls it, and `git_remote_status` already covers ahead/behind in one git call where this one needs four. Adding a helper to `lib/` that no shipped theme uses is a maintenance cost we'd carry forever for something you can drop in your `.zshrc` in a few lines.

If you're using it in a custom theme, keep it there... that's the supported spot for it, and it'll survive updates.

Thanks again, and sorry for the wait.

PR #6784, opened 2018-04-30
Fix branch not updating on checkout in steeef

shekibobo:
#6359 broke branch name updates in the steeef theme prompt. `$2` apparently is an insufficient replacement for `$(whence -c $(fc -l $HISTCMD $HISTCMD))`.


mcornella:
What command did you use to do the checkout? As it is now, as long as the command contains `git` in it it will update, after expanding all aliases in it.

shekibobo:
I'm using `g co master`, where my bash alias `g = git`, and `git = hub`. So it's probably because I'm using `hub` instead of `git` that's causing the problem, but it's still a problem that makes steeef a problematic theme if I'm using a tool that wraps `git` itself.

mcornella:
So using whence will have the same impact as `$2`. What was there previously was to always update the prompt regardless of the command typed, but we "fixed" it to do what it intended. I guess we could check for hub as well as git? Care to fix the PR?

shekibobo:
Seems to work when adding `hub` in the checks with `$2`.

PR #6791, opened 2018-05-03
git: add pull rebase --autostash aliases

Poyoman39:
This is quite a nice feature from git. I'm not sure if it's better adding it to gup / gupv, or if it should be added to a new alias :)


vrcca:
You read my mind! I was thinking about sending a PR to add `gupa` and `gupav`. The reason is that sometimes we want to make sure that we have a clean code tree, so making `gup` default to `--autostash` could cause some confusion. What do you think? 😄

Poyoman39:
I totally agree with you 👍 After doing some gup --autostash i think a gupa would be really cleaner ! :P

Poyoman39:
I just updated my pull request to do so :)

emmanueltouzery:
this looks great, i hope this gets merged!

PR #6797, opened 2018-05-05
"Zsh not installed" fix #6796

### LINKED ISSUE #6796: Zsh is not installed! Please install zsh first!
hello robbyrusell !

after install like your description on debian 9.4 i got this error:

Zsh is not installed! Please install zsh first!

zsh is install

```
# /etc/shells: valid login shells
/bin/sh
/bin/dash
/bin/bash
/bin/rbash
/bin/zsh
/usr/bin/zsh
```

two hours ago i install it on a another debian 9.4 computer and it works and i see you make changes one hour ago

please fix

mriedmann:
Quick fix for #6796

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -24,10 +24,10 @@ main() {
   # which may fail on systems lacking tput or terminfo
   set -e
 
-  if command -v zsh >/dev/null 2>&1; then
+  command -v zsh >/dev/null 2>&1 || {
     printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
     exit
-  fi
+  }
 
   if [ ! -n "$ZSH" ]; then
     ZSH=~/.oh-my-zsh


CryDeTaan:
How about this?
`if ! hash zsh 2>/dev/null; then
    printf "${YELLOW}Zsh is not installed!${NORMAL} Please install zsh first!\n"
    exit
  fi`

mriedmann:
@CryDeTaan also completely valid.

Just to clarify: The `hash` solution was present in line 50 and got replaced with this (see https://github.com/robbyrussell/oh-my-zsh/commit/4fec0a46e73479cf3b9e7f953750af5ed5df87ac#diff-55b046be8e081f97f37b7667e1c4bfa9):
```
printf "${BLUE}Cloning Oh My Zsh...${NORMAL}\n"
  command -v git >/dev/null 2>&1 || {
    echo "Error: git is not installed"
    exit 1
  }
```
so I used the "or side-effect" approach. There is already a (closed) PR for this solution: #6798

adambullmer:
Any reason you don't just add as `!` to the existing check? Just looks like a simple case of inverted logic to me.

```bash
if ! command -v zsh >/dev/null 2>&1; then
# ...
```

mriedmann:
As I sad, I just wanted to provide a concise solution (because this solution was replaced at another line earlier). A simpler fix (the one you suggested) can be found in PR #6798

PR #6801, opened 2018-05-06
Minor fix: "stdin" instead of "stding"

mauroporrasp:


juanibiapina:
This should be fixed in the official completion instead: https://github.com/docker/cli/blob/master/contrib/completion/zsh/_docker

PR #6803, opened 2018-05-07
Allow completion of dot directories

### LINKED ISSUE #3775: Insert slash (/) as a result of tab-completion after two dots (..)
When using other terminals, if I want to write a path like e.g. `../somefolder` I'm used to doing:
```
$ ..(tab!)somef(tab!)
```
This works perfectly like that in CYGWIN, MINGW32 or the GNOME Terminal with bash, but that `..(tab!)` doesn't work on oh-my-zsh on iterm2 on Mac OS X Yosemite because the tab key does not automatically complete the slash `/` symbol after `..`

I know it might seem a very small detail, but I'm so used to that it really bugs me having to reach the real `/` key, especially in non-American keyboards where it does not have a dedicated key (for me, on a Spanish keyboard, `/` is `Shift + 7`).

### LINKED ISSUE #6543: cd .. completion not working
I just installed oh my zsh on a new machine and the 'cd ..' completion is not working. All the other directory and filename completions work properly, but when I type 'cd ..' no proposals for directories in the upper directory are provided. Before installin ohmyzsh, on bash the completion worked properly. 

I'm on a Macbook Pro 13'' touchbar with High Sierra 10.13.2 .

mcornella:
Fixes #3775, fixes #6543.

This makes zsh treat `.` and `..` as directories to be completed. That means it will insert a slash (`/`) when these directories are provided in the command line.

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -25,6 +25,9 @@ else
 fi
 unset CASE_SENSITIVE HYPHEN_INSENSITIVE
 
+# Complete . and .. special directories
+zstyle ':completion:*' special-dirs true
+
 zstyle ':completion:*' list-colors ''
 zstyle ':completion:*:*:kill:*:processes' list-colors '=(#b) #([0-9]#) ([0-9a-z-]#)*=01;34=0=01'
 


asqui:
Should this be off by default, since it's a change in completion behaviour?

(I was certainly surprised about the change and had to go digging to find that this was the cause.)

mcornella:
There were users surprised by the lack of this feature, and enabling it just means you have to type a letter after the dot to complete to hidden files, which I presume is your use case.

PR #6804, opened 2018-05-08
Verify core.autocrlf config before upgrade

anuarneto:
If you let core.autocrlf true during upgrade it breaks oh-my-zsh!
This solves the problem by verifing and keeping the state


mcornella:
We can bypass the global git configuration directly by setting the option only in the repository. So, presumably, we would only need `git config core.autocrlf false` before the pull. I'm not sure if `false` is the right option for that though.

anuarneto:
Just to config the repository is really a better solution then mine one.
But I do think to set it to `false` at this repo is better.
With it `false` it works always, with it `true` it breaks!

PR #6807, opened 2018-05-08
kube-ps1 rename main function

david972:
Plugin kube-ps1

To use it as a prompt element


mcornella:
This is a modification to make it compatible with Powerlevel9k, but it isn't dependent on it. To make it compatible without having to change the plugin, just define a `prompt_kube_ps1` function which calls `kube_ps1` and there won't be a need to change the whole thing for other users.

PR #6815, opened 2018-05-10
[plugins/runtime] Add simple plugin counting a command's execution time

shtrom:
Signed-off-by: Olivier Mehani <shtrom@ssji.net>


mcornella:
Has no interest.

PR #6819, opened 2018-05-12
user default login shell with username(LOGIN).

avsmips:
**SYNOPSIS**  
> **chsh** [options] [LOGIN]

change shell for current user thru sudo mode.


mcornella:
When the user is not specified `chsh` changes the shell of the current user.

PR #6820, opened 2018-05-14
Display tag instead of commit revision

b4nst:
When a tag is checkout, display `tag/[tag-name]` instead of `[commit-sha]`


carlosala:
This looks quite good! Could you rebase on top of master to be able to merge? Thanks!

b4nst:
Closing in favor of #11318. I had to recreate a fork since I removed the original one.

PR #6827, opened 2018-05-15
Disable menu-select completion on buggy zsh versions

### LINKED ISSUE #6818: Ctrl+C during tab completion hides prompt
Using oh-my-zsh with absolutely default settings, the following happens:

https://asciinema.org/a/w5gDwZS2xjZpxoKGLAz8JP5tV

The way to reproduce is:

1. Enter the start of a command that's ambiguous.
2. Press TAB to show options.
3. Press TAB again to move selection to first option.
4. Press Ctrl+C to cancel selection.

Now your prompt is hidden.

mcornella:
Also abbreviate zstyle syntax.

From https://github.com/robbyrussell/oh-my-zsh/issues/6818#issuecomment-389110400:
> I've narrowed it down to versions between https://github.com/zsh-users/zsh/commit/32f5d3d8c16b4f3a11fa39c0ee378d72336ba853 (which introduces the bug) and https://github.com/zsh-users/zsh/commit/1d8b5285226afba4f1ef9030cf862c14b975c284 (which fixes it). 
> That means that versions 5.1.1 and 5.2 have this bug (first reported [here](http://www.zsh.org/mla/workers/2015/msg03349.html)).

Fixes #6818.

cc @orlp

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -11,7 +11,11 @@ setopt always_to_end
 
 # should this be in keybindings?
 bindkey -M menuselect '^o' accept-and-infer-next-history
-zstyle ':completion:*:*:*:*:*' menu select
+
+# These versions delete the prompt on CTRL+C while on menu selection
+if ! [[ $ZSH_VERSION == 5.1.1 || $ZSH_VERSION == 5.2 ]]; then
+  zstyle ':completion:*' menu select
+fi
 
 # case insensitive (all), partial-word and substring completion
 if [[ "$CASE_SENSITIVE" = true ]]; then


orlp:
I'm not exactly sure if this workaround is a net positive. The hidden prompt is a rare event (only after cancelling tab completion) and is instantly fixed by pressing enter once.

At the same time you lose functionality from not seeing the selection, which now affects everyone, even if they didn't Ctrl-C.

Finally, this clutters zsh with a workaround for an already out of date zsh version. How long will this workaround be maintained?

mcornella:
That is also what I thought, but the completion menu is still displayed even if it's only selectable with tabs. Regardless, I agree that this is a solution to a temporary bug that causes more trouble than it solves.

On another note, I've found that if you CTRL+G instead of CTRL+C you will cancel the menu without activating the bug.

Thanks for your attention.

PR #6830, opened 2018-05-16
[shrink-path plugin] unset chpwd_functions before running `cd`

dt-rush:
This avoids side-effects if the user has set any `chpwd_functions` which cause output, such as is the case if the user is using `auto-ls`.

I was using `auto-ls` plugin alongside shrink-path in `agnoster` them. I had, in my `.zshrc` file:

```
ZSH_THEME="agnoster"
plugins=(git autojump lol colored-man-pages copyzshell shrink-path)
AUTO_LS_COMMANDS=(ls)
AUTO_LS_NEWLINE=false
```

Without this fix, my prompt was filled with the contents of my home directory. It took up the whole terminal, hahaha.

### DIFF
--- a/plugins/shrink-path/shrink-path.plugin.zsh
+++ b/plugins/shrink-path/shrink-path.plugin.zsh
@@ -94,6 +94,11 @@ shrink_path () {
         (( tilde )) && dir=${dir/$HOME/\~}
         tree=(${(s:/:)dir})
         (
+                # unset chpwd_functions since we'll be calling `cd` and don't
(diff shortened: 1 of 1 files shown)


dt-rush:
@mcornella why the revert?

mcornella:
Using `cd -q` is a better fix which doesn't have side-effects. See https://github.com/robbyrussell/oh-my-zsh/commit/66cb4005ab6cfcd6b092c1b482e767af214e83bb.

dt-rush:
Ah I see now, thanks for the explanation. Didn't see that commit. Cheers!

PR #6834, opened 2018-05-17
[plugins/terraform]Update terraform plugin with v0.11.7 commands and help

MattLewin:
Terraform has been rapidly developed, so the plugin was way out of date. I've updated it to reflect all the new and removed functionality.


mcornella:
Thanks!

PR #6841, opened 2018-05-20
Fix 6840 Check emacsclient version instead of emacs's

lbolla:
This is much faster.


robbyrussell:

Cheers,
Robby

Issue #2355, opened 2013-12-17
Virtualenvwrapper plug-in crashes Iterm2 tab on tab complete


wcbeard:
When trying to type `workon lytic`, where `lytic` is a virtualenv, an iTerm tab closes when I try to tab-complete the virtualenv name.

Hitting tab at `➜  ~  work|` completes to `workon`, but hitting it at `➜  ~  workon l|` often quickly prints out 

```
➜  ~  workon l......-zsh(60772,0x7fff7a987310) malloc: *** error for object 0x7fbefa828638: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

before simply closing the tab. (If I do it quickly enough after opening the tab, iterm gives a warning message before closing the tab.)

I have no idea whether this is an oh-my-zsh/zsh problem, a virtualenv problem, or a Mac problem, nor how to debug malloc_error_break.

I'm using Mac OS 10.9, 
virtualenv==1.10.1
virtualenv-clone==0.2.4
virtualenvwrapper==4.1.1
zsh 5.0.2 (x86_64-apple-darwin13.0.0)
oh-my-zsh 357ab49

PR #6842, opened 2018-05-20
Prefer virtualenvwrapper_lazy

### LINKED ISSUE #6839: Prefer virtualenvwrapper_lazy to speed up loading
`virtualenvwrapper` plugin is quite slow (it takes 200ms to load on my machine).
Would it be possible to prefer using `virtualenvwrapper_lazy.sh` when available, which lazily sources what's needed?

lbolla:
This gives much faster start up times and only loads virtualenvwrapper
when needed.

Fix #6839

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -1,6 +1,14 @@
 virtualenvwrapper='virtualenvwrapper.sh'
+virtualenvwrapper_lazy='virtualenvwrapper_lazy.sh'
 
-if (( $+commands[$virtualenvwrapper] )); then
+if (( $+commands[$virtualenvwrapper_lazy] )); then
+  function {
+    setopt local_options
+    unsetopt equals
+    virtualenvwrapper=${${virtualenvwrapper_lazy}:c}
+    source ${${virtualenvwrapper_lazy}:c}
+  }
+elif (( $+commands[$virtualenvwrapper] )); then
   function {
     setopt local_options
     unsetopt equals


mcornella:
Can you try to reproduce https://github.com/robbyrussell/oh-my-zsh/issues/2355#issue-24397582? I think the bug should still be there, that's why we removed support for it in #2615.

lbolla:
https://github.com/robbyrussell/oh-my-zsh/issues/2355#issue-24397582 seems like an issue with MacOS's iTerm: I don't have a Mac to test.
Can't we just enable lazy virtualenvwrapper on Linux and Windows, if it's broken on Mac?

mcornella:
I'm not sure that's really a Mac, the [bug report to zsh's mailing list](http://www.zsh.org/mla/workers/2014/msg00848.html) doesn't mention it, though it is true that many in the issue thread do mention Mac and iTerm.

Can you just try if completing a virtualenv name using `workon <virtualenv><TAB>` closes the tab or prints an error like the one in the issue? If not, we can just pull this and reenable virtualenvwrapper_lazy.

lbolla:
`workon <virtualenv><TAB>` works fine for me, using urxvt and xterm on Linux and zsh 5.5.1.

mcornella:
Thanks for looking into it!

clokep:
This seems to have broken automatically activating virtualenvs when moving to directories in git projects. I was able to revert 90a5bd06ca76aff04f76a5d43a432ed9340dd78d and it started working again.

lbolla:
@clokep I guess the "auto venv activation plugin" should run "workon" upon init, which will trigger the lazy loading of virtualenvwrapper.

mcornella:
I pushed an update to fix the logic. Run `upgrade_oh_my_zsh` to update.

clokep:
@mcornella Thank you! I confirmed it is working for me now!

PR #6844, opened 2018-05-20
Cache kubectl completion script to file to speed up sourcing

### LINKED ISSUE #6843: Loading kubectl plugin is slow
The main reason is that it regenerates the completion script over and over, with `kubectl completion zsh`.

lbolla:
Fixes #6843.


lbolla:
Any comments on this PR?

mcornella:
Thanks!

PR #6846, opened 2018-05-21
add plugin 'universalarchive()' to conveniently compress files

Konfekt:
Lets you compress files by a catchy command `ua <format> <files>` into various compression formats, for example `<format>` could be `zip`, `xz`, ..

For example,
```sh
ua xz *.html
```
will compress all `html` files in a folder `folder` into `folder.xz`.


gaving:
Pretty handy plugin, is there any appetite out there to get this merged?

Konfekt:
Thank you for the support. This also became part of [zimfw](https://github.com/zimfw/archive) (with slight enhancements since then).

mcornella:
I have a few review comments, but what's the status? Do you still want this in? I'd go ahead with it but I need to know someone will maintain the plugin.

Konfekt:
Well, I could maintain it.

mcornella:
I guess what I'm asking is whether you'll be maintaining this version or should it be the same code.

Konfekt:
It must not be the same code.

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
+alias ua='universalarchive'
```
mcornella, line 1: This won't get loaded, the appropriate plugin path nomenclature is `plugins/<name>/<name>.plugin.zsh`. You should reconsider whether you want it to be named `ua` or `universalarchive`. If you want the latter, it should be `plugins/universalarchive/universalarchive.plugin.zsh`; otherwise it should be `plugins/ua/ua.plugin.zsh`.

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
...
+    echo "Usage:   ua <archive type> <files>"
+    echo "Error: Please give a valid file or directory name!"
+  fi
+
+  if [ -f "${name}.${type}" ]; then
+    name=$(mktemp --dry-run "./${name}_XXX.${type}")
+  fi
+
+  case "$type" in
+    tar)          shift; tar      cvf            "${name}.tar"       "${@%%}" ;;
```
mcornella, line 35: If you're going to run `shift` in all cases you can put it outside the `case` switch.

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
...
+    tbz|tar.bz)   shift; tar      cvjf           "${name}.tar.bz2"   "${@%}"  ;;
```
mcornella, line 36: What does `${@%%}` do, and why is it `${@%}` (1 %) here?

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
...
+    tbz2|tar.bz2) shift; tar      cvjf           "${name}.tar.bz2"   "${@%%}" ;;
+    txz|tar.xz)   shift; env XZ_OPT=-T0 tar cvJf "${name}.tar.xz"    "${@%%}" ;;
```
mcornella, line 38: You don't need `env`, you can set the variable directly:
```suggestion
    txz|tar.xz)   shift; XZ_OPT=-T0 tar cvJf "${name}.tar.xz"    "${@%%}" ;;
```

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
...
+    Z|compress)   shift; compress -vcf           "${@%%}"          > "${name}.Z" ;;
+    zip)          shift; zip      -rull          "${name}.zip"       "${@%%}" ;;
+    7z|7zip)      shift; 7z       u              "${name}.7z"        "${@%%}" ;;
+    rar)          shift; rar      a              "${name}.rar"       "${@%%}" ;;
+    lzo)          shift; lzop     -vc            "${@%%}"          > "${name}.lzo" ;;
+    xz)           shift; xz       -vc -T0        "${@%%}"          > "${name}.xz" ;;
+    lzma)         shift; lzma     -vc            "${@%%}"          > "${name}.lzma" ;;
+    *)  echo "universalarchive(): archive a file or directory."
+        echo "Usage:   ua <archive type> <files>"
+        echo "Example: ua tbz2 PKGBUILD"
```
mcornella, line 53: This mixes both names. I think what you really want is to name it `ua`, so name the function that instead. Also, use `$0` instead of a hardcoded name.

plugins/ua/universalarchive.plugin.zsh
```diff
@@ -0,0 +1,59 @@
...
+  elif [ -d "$input" ]; then
+    name="${input%/*}"
+  else
+    echo "universalarchive(): archive a file or directory."
+    echo "Usage:   ua <archive type> <files>"
+    echo "Error: Please give a valid file or directory name!"
+  fi
+
+  if [ -f "${name}.${type}" ]; then
+    name=$(mktemp --dry-run "./${name}_XXX.${type}")
```
mcornella, line 31: `--dry-run` is exclusively a GNU mktemp feature:

[1] http://man.openbsd.org/mktemp.1
[2] https://www.freebsd.org/cgi/man.cgi?query=mktemp
[3] https://docs.oracle.com/cd/E23823_01/html/816-5165/mktemp-1.html#scrolltoc

[See this post](url) for the rest of mktemp docs across platforms.

plugins/ua/_universalarchive
```diff
@@ -0,0 +1,7 @@
+#compdef universalarchive
+#autoload
+
+_arguments \
+  '(-r --remove)'{-r,--remove}'[Remove archive.]' \
```
mcornella, line 5: You specify a `-r`/`--remove` option but there's none implemented. I think this is a copy-paste artifact from the `extract` plugin.

mcornella (changes requested):
Alright, I'm submitting everything I saw that needs fixing. The README should be updated as well due to some of the changes required.

plugins/ua/universalarchive.plugin.zsh
Konfekt: Thank you. In case `*` shift is not run, but I put it outside anyway.

mcornella (approved):

mcornella:
Made several improvements.

The completion file didn't work at all, here's what I changed:
1. The `#compdef` line needed to have the name of the command to be completed, in this case, `ua`.
2. The `#autoload` line wasn't needed.
3. The `_arguments` line didn't work like you expected. Instead, it was suggesting files ending in the archive format extension.

In the `ua` function, I added a few checks to make sure we got a valid input, and fixed the output filename logic to be consistent:
1. The output uses the name of the parent directory whenever there is more than 1 input file or directory.
2. Kept the syntax POSIX-shell compatible, though this will be probably be used only in Zsh.
3. Fixed filenames using double extension when the `mktemp` call was triggered.
4. Added a safety check to make sure wrong logic doesn't override any files.

I also reordered the case switch logic and removed all duplicate archive formats. Now only the file extension is used, which makes it easier to maintain.

Finally, I reworded some stuff in the README and cleaned up the style.

If you agree with the changes, I'll merge this and add you to the `CODEOWNERS` file as the maintainer of the plugin (this means you'll get pinged when there are PRs modifying the plugin).

Konfekt:
Thank you very much for the improvements that I all agree with.

Maverick1872:
Something to consider: this plugin would likely be a prime candidate for adding "universal" extract functionality as well. Would be a very handy little plugin if it was a "one stop shop".

Konfekt:
This is what https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/extract achieves.

gaving:
# 🎉 

cheers guys!

PR #6861, opened 2018-05-26
Export random theme so we know what it was

millimeep:
For those times when you're really enjoying a random theme and want to know which one it was, but you can't find the zsh start up message!

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -99,7 +99,7 @@ if [[ "$ZSH_THEME" == "random" ]]; then
   fi
   N=${#themes[@]}
   ((N=(RANDOM%N)+1))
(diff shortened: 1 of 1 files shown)


mcornella:
You don't really need to export the variable, `$RANDOM_THEME` will be in the same namespace so you can just echo that directly.

millimeep:
The point is to be able to retrieve it easily days later when the startup message may have scrolled out out of sight...

mcornella:
Exported variables don't survive until the next shell session. They are only visible to child processes, but when the parent process (the one that exported the variable) dies, the exported variable dies with it.

So what you want is to access that variable while it is accessible, before the zsh session closes.

PR #6866, opened 2018-05-27
[plugins/heroku] add container options

djetelina:


jdxcode:
this should be closed as it's using the official autocomplete now

PR #6867, opened 2018-05-28
[plugins/archlinux] add alias for yay

specter119:
[Yay](https://github.com/Jguer/yay) another Yogurt - An AUR Helper written in Go

Considering the  [Arch wiki: AUR helpers](https://wiki.archlinux.org/index.php/AUR_helpers), I think `yay` is more robust and powerful than `yaourt`.

I alias `yay` with the same commands as `yaourt`, I think most people will not use them together.

### DIFF
--- a/plugins/archlinux/README.md
+++ b/plugins/archlinux/README.md
@@ -2,6 +2,26 @@
 
 ## Features
 
+#### YAY
+
+| Alias   | Command                            | Description                                                         |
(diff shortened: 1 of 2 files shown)


mati865:
Definitely `yay` should be above outdated `pacaur` and `yaourt`.

juliuswz:
Up +

mcornella:
Can you add these to the README?

mcornella:
👍

PR #6868, opened 2018-05-28
Update autojump.plugin.zsh

rl1k:
if like me when you have an error  like `$HOME/plugins/autojump/autojump.plugin.zsh:cd:3: string not in pwd: ..`

you can replace the autojump.plugin.zsh by mine, the problem was you must escape the . in the script


iNoSec:
i have no conflict??

mcornella:
I merged your changes using `source` instead. Just for the record though, you shouldn't alias `.` to anything else, you're going to run into other problems.

PR #6870, opened 2018-05-29
remove unnecessary warning from virtualenvwrapper.plugin.zsh

timfeirg:
I think this warning message is meaningless since 90a5bd06ca76aff04f76a5d43a432ed9340dd78d (from https://github.com/robbyrussell/oh-my-zsh/pull/6842), the reason being:

* if virtualenvwrapper.sh is being used, $WORKON_HOME will be
immediately (to the default value), no need for this warning
* if virtualenvwrapper_lazy.sh is being used, $WORKON_HOME will be set
(to the default value) later, automatically:

image

* if none of the initialization scripts are available, for example
if virtualenvwrapper was uninstalled, the installation command will be
printed during zsh start:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ custom/
 # temp files directories
 cache/
 log/
+
+.DS_Store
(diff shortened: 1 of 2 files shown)


timfeirg:
cc @lbolla, RFC @mcornella

mcornella:
Hi @timfeirg, see #6882 for discussion.

PR #6871, opened 2018-05-29
Add git add intent-to-add commands

lapanti:
For use with a workflow such as https://medium.com/@nodunayo/tell-git-your-intentions-1ed062294320


lucsrods:
This alias to patch include new files will also be a nice to have:
`alias gapan='git add --intent-to-add . && git add --patch'`

Based on [this question](https://stackoverflow.com/questions/14491727/git-add-patch-to-include-new-files).

karlhorky:
@lucsrods I opened #6901 for this last year (aliased to `ganpa`)

PR #6873, opened 2018-05-29
Update cloud.zsh-theme

miqianmimi:
By changing the $PS1(prompt), the path and host can be displayed now. This change makes it much easier without losing the beauty.


mcornella:
This changes the prompt line significantly. See the [customization wiki](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization) for instructions on how to make it into a custom theme.

PR #6874, opened 2018-05-29
[plugins/archlinux] add alias for pikaur

isac322:
According to [AUR helper](https://wiki.archlinux.org/index.php/AUR_helpers), [pikaur](https://github.com/actionless/pikaur) is a substitute for *yaourt* with many functions.

So I alias pikaur with prefix `pi`

Example:
`piin` is same as `pikaur -S`


shanff:
I missed this one when I just committed my add for pikaur.  I chose to use the prefix of pik- though.  Either or works for me.  Mine is #7128

PR #6886, opened 2018-06-02
Add integration and autocompletion to jabba

2m:
This PR adds a plugin which contains needed shell integration script (`jabba` function) for jabba to work correctly as well as autocomplete code for some of the jabba commands (instal, uninstall and use).


2m:
I have packaged this plugin to a separate repository so it is possible to use it from oh-my-zsh as an external plugins or using a different zsh manager.

https://github.com/2m/zsh-jabba

PR #6890, opened 2018-06-05
Add merge continue

alzubaidi:
Add alias `gmc` to `git merge --continue`


mcornella:
It's just really syntactic sugar for `git commit`.

PR #6892, opened 2018-06-05
[Plugins] sprunge alike as sprunge is out of order

ahanjrah:
Sprunge is not working per https://github.com/rupa/sprunge/issues/43 and https://github.com/rupa/sprunge/issues/45

ix.io is a similar service, can be found at: http://ix.io

### DIFF
--- a/plugins/ixio/ixio.plugin.zsh
+++ b/plugins/ixio/ixio.plugin.zsh
@@ -1,0 +1,66 @@
+# Updated for ix.io as sprunge is not working anymore [https://github.com/rupa/sprunge/issues/43]
+# Modded by ahanjrah <aman dot hanjrah -at- gmail>
+# Kudos to the original creator/maintainer
+#
+# Contributed and SLIGHTLY modded by Matt Parnell/ilikenwf <parwok -at- gmail>
+# Created by the blogger at the URL below...I don't know where to find his/her name
+# Original found at http://www.shellperson.net/sprunge-pastebin-script/
+ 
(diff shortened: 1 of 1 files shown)


etu:
https://ptpb.pw/ is also nice (source here <https://github.com/ptpb/pb>). One benefit over ix.io is https :-)

mcornella:
sprunge seems to be working as of today, but ix.io could be useful as well. Seems though that [their client](http://ix.io/client) is better because it uses all of the features of the service. Can you use that instead and add a basic README?

> https://ptpb.pw/ is also nice (source here https://github.com/ptpb/pb). One benefit over ix.io is https :-)

[This one's dead now](https://github.com/ptpb/pb/issues/246).

etu:
@mcornella Yeah... Well aware of that, and it's sad :(

https://github.com/etu/webpaste.el/commit/7345c5f62d5cff4d84379eaf5dc8b2bb8bc4f99c

Thanks anyways though :)

mcornella:
> sprunge seems to be working as of today, but ix.io could be useful as well. Seems though that [their client](http://ix.io/client) is better because it uses all of the features of the service. Can you use that instead and add a basic README?
> 
> > [ptpb.pw](https://ptpb.pw/) is also nice (source here [ptpb/pb](https://github.com/ptpb/pb)). One benefit over ix.io is https :-)
> 
> [This one's dead now](https://github.com/ptpb/pb/issues/246).

Closing due to inactivity.

PR #6893, opened 2018-06-05
[plugins/git] Updated git clone alias

kkalavantavanich:
`git clone --recursive` has been deprecated in favor of `--recurse-submodules`.  
Also added `-j8` to speed up clone.
See: https://stackoverflow.com/questions/3796927


steelrooter:
👍

kkalavantavanich:
@mcornella So what do you suggest then: `--recurse-submodules` or `--recursive`, or both?

mcornella:
I think `--recurse-submodules` is ok, just lose the `-j8`.

kkalavantavanich:
Done

mcornella:
Thanks!

PR #6896, opened 2018-06-06
Add simple plugin to launch visual studio code

mariusbreivik:
Plugin for visual studio code heavily inspired by the sublime plugin.
Supports macos and linux.

### DIFF
--- a/plugins/visualstudiocode/README.md
+++ b/plugins/visualstudiocode/README.md
@@ -1,0 +1,16 @@
+## visual studio code
+
+Plugin for Microsoft Visual Studio Code. A brilliant text and code editor.
+This plugin is heavily inspired by the Sublime Text Plugin
+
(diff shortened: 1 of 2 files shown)


plugins/visualstudiocode/README.md
```diff
@@ -0,0 +1,16 @@
+## visual studio code
+
+Plugin for Microsoft Visual Studio Code. A brillian text and code editor.
```
jolkdarr: Typo: brilliant

mariusbreivik: Fixed

mcornella:
VS Code already puts the code binary in its PATH, except for MacOS where the setup is [built into the app](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line).

PR #6901, opened 2018-06-08
Add ganpa alias for untracked files

karlhorky:
Building off @lapanti's #6871 and https://stackoverflow.com/a/45640990/1268612, this adds an alias that adds all untracked files via `--intent-to-add` before entering patch mode.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -43,6 +43,7 @@ alias g='git'
 alias ga='git add'
 alias gaa='git add --all'
 alias gapa='git add --patch'
+alias ganpa='git add --intent-to-add . && git add --patch'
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it sat here for years without an answer.

Someone opened the same alias more recently in #13609, with README docs, so I'm closing this one and letting that one carry it.

Worth saying though: your version is the correct one. You wrote `git add --intent-to-add .` with the pathspec, and #13609 leaves the `.` off... which makes that half of the alias a no-op. I'll flag it over there.

Thanks again.

karlhorky:
@robbyrussell no problem, thanks for the update. If there's a way to attribute me in the final commit for #13609 then happy to receive that.

PR #6903, opened 2018-06-08
convenient vscode plugin for oh-my-zsh

MarsiBarsi:


mcornella:
I'm for it, but can we cut back on the number of aliases? I'm sure some of them will be used next to nothing...

PR #6910, opened 2018-06-12
Add fzf plugin

### LINKED ISSUE #3003: Add junegunn's fzf to plugins.
It allows for `ctrl+r` to use fuzzy search and `ctrl+t` does fuzzy autocompletion: https://github.com/junegunn/fzf

maxbrunet:
Add [junegunn's fzf](https://github.com/junegunn/fzf) fuzzy auto-completion and key bindings

* Try to find `fzf` installation directory in [official and suggested locations](https://github.com/junegunn/fzf#installation)
  (Homebrew, Git, Arch Linux, Fedora)
* Add `bin` directory to `PATH`, enable auto-completion and key bindings
  (extracted from the [official install script](https://github.com/junegunn/fzf/blob/0.17.4/install#L250-L282))
* Support for disabling auto-completion or/and key bindings

Closes #3003


briantully:
Any update on this? Currently it seems fzw keybindings are incompatible with ohmyzsh, so this plugin would be a huge win.

mcornella:
Thanks @maxbrunet!

lalo:
Is this working? I did a quick test and it seems that the key bindings don't last. If I manually add the `source "$fzf_base/shell/key-bindings.zsh"` at the end of my .zshrc it will work. Can somebody repro?

plugins/fzf/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Uncomment the following line to disable fuzzy completion
+# export DISABLE_FZF_AUTO_COMPLETION="true"
+
+# Uncomment the following line to disable key bindings (CTRL-T, CTRL-R, ALT-C)
+# export DISABLE_FZF_KEY_BINDINGS="true"
+
+plugins=(
+  ...
+  fzf
```
doubleloop: note: if fzf plugin is placed before vi-mode plugin, ctrl-r shortcut will not work by default

captainju:
Arch user here, I installed the default fzf package https://www.archlinux.org/packages/community/x86_64/fzf/

There is an error when loading the plugin:
`/usr/share/oh-my-zsh//plugins/fzf/fzf.plugin.zsh:source:42: no such file or directory: /usr/share/fzf/shell/key-bindings.zsh`

Indeed, the file `key-bindings.zsh` is located in `/usr/share/fzf/`
Should we add this path into the plugin, or modify the arch package to include the shell folder?

PR #6911, opened 2018-06-12
virtualenvwrapper: Support pyenv

maxbrunet:
Add support to `pyenv` using [`pyenv-virtualenvwrapper`](https://github.com/pyenv/pyenv-virtualenvwrapper).

* Determine that `pyenv` is in use if `PYENV_SHELL` is exported
* Run `pyenv virtualenvwrapper_lazy` (else `pyenv virtualenvwrapper`)

Closes #2927


cyphus:
Looks good! I don't use virtualenvwrapper so I can't test it but provided it works I say LGTM.

Marcholio:
Can this be merged if it works well?

maxbrunet:
I recommend to use [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv) instead which is supported by the [pyenv plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/pyenv).

PR #6916, opened 2018-06-14
Update git plugin with fast dirty check

JackDrogon:
Update git plugin with fast dirty check.


robbyrussell:
Thanks for this, and sorry it sat so long.

Prompt speed in big repos is a real problem and I'm glad you took a run at it, but I'm going to close this one.

Two reasons. `git diff --quiet` doesn't see untracked or staged-only changes, so the dirty marker would quietly stop showing up in the cases people notice most. And the speedup is smaller than it looks... I measured about 17.6ms vs 13.0ms per call on a 6k-file repo.

The bigger fix landed in the meantime: #12257 made the git prompt async, so the dirty check doesn't block your prompt at all anymore.

Thanks again for digging into it.

PR #6918, opened 2018-06-14
Fixed curl-output used for auto-completion

domrim:
Removed the status-output with the `-s` argument.


mcornella:
Merged #6898 instead (with your co-authorship). Thanks for bringing this to my attention.

PR #6919, opened 2018-06-14
use official heroku autocomplete

jdxcode:
https://blog.heroku.com/announcing-cli-autocomplete


jdxcode:
The CLI expects everything to be in this location though

mcornella:
Fair enough. Sorry for the delay.

jdxcode:
thank you so much!

PR #6922, opened 2018-06-16
Added some help and example to lib/directories so nOObs like me can start being more awesomer!

skydivekevin:
Added some help and a sweet example to the lib/directories to help newcomers with using alias'. As a new zsh user myself, I was a bit unsure of how to use an alias to automate a task, so it took some fiddling around. This would have helped me a bit I think, and hopefully it helps others as well!

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -1,4 +1,15 @@
 # Changing/making/removing directory
+# Example of using an alias to automate the creation of a basic
+# environment as follows:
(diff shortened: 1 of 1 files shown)


mcornella:
User facing documentation should be written somewhere where it is likely to be found by users. Alias documentation in particular is zsh-specific, so it better be in a wiki page dedicated to first steps with zsh, or something of that sort.

PR #6925, opened 2018-06-17
change color blue+black -> blue+white

HyundongHwang:
I am an avid fan of the agnoster theme.
However, the blue + black combination of prompts is often visually inadequate to make mistakes.
So I modified it as follows.

before:

after:

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -187,14 +187,14 @@ prompt_hg() {
 
 # Dir: current working directory
 prompt_dir() {
-  prompt_segment blue black '%~'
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry for the very long silence.

You were right that blue-on-black is hard to read on a lot of terminals... someone else hit the same thing in #12497. But the theme changed since you opened this: the colors are configurable now, so you can set exactly what you wanted from your `.zshrc`:

```zsh
AGNOSTER_DIR_FG=white
```

The specific lines your patch touched don't exist anymore after #12505, so there's nothing left to merge here. Closing for that reason.

Thanks for flagging it.

PR #6929, opened 2018-06-18
fix path completion issue with go run subcommand

### LINKED ISSUE #6928: golang plugin: path completion doesn't work with the `run` subcommand
## The Problem:
Path completion doesn't work as expected when using the `go run` subcommand. I believe this issue is related to issue #4323.

## How to reproduce:
1. `cd` into a subdirectory on your `GOPATH` (one not containing any `.go` files, such as a project's parent directory.)
2. type `go run <TAB>`

For example, say you're the directory `$GOPATH/src/github.com`:

```
github.com
├── app1
│   └── main.go
└── app2
    └── main.go
```

Typing `$ go run <TAB>` on the command line gives no path completion.

---

### Expected Behavior
Tab completion allows cycling through directories until we locate a `*.go` file in a subdirectory.

### Actual Behavior
Tab completion does not allow cycling through directories.

banjeremy:
Fixes #6928

### DIFF
--- a/plugins/golang/golang.plugin.zsh
+++ b/plugins/golang/golang.plugin.zsh
@@ -135,7 +135,7 @@ __go_tool_complete() {
   run)
       _arguments -s -w : \
           ${build_flags[@]} \
-          '*:file:_path_files -g "*.go"'
+          '*:file:_files -g "*.go"'
       ;;
   tool)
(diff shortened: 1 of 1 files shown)


mcornella:
Great write-up!

PR #6941, opened 2018-06-21
Update git.plugin.zsh

partounian:
Correct gcl command to use `--recurse-submodules` instead of deprecated `--recursive`, add some very useful commands.


partounian:
Closed but if someone wants to see some more awesome git aliasing check this out https://github.com/partounian/dotfiles/blob/master/git/aliases.zsh

PR #6942, opened 2018-06-23
Make tools/* executeable for easier use

smeagolthellama:
just a simple `chmod u+x *`.


mcornella:
These tools shouldn't be used directly. Instead, use `upgrade_oh_my_zsh` and `uninstall_oh_my_zsh`.

PR #6953, opened 2018-06-27
Create super useful gac and gacp aliases

partounian:
`gac` will add all files and will open your editor to commit everything
`gac 'Initial commit'` will add all files and commit with the message 'Initial commit'
`gac  plugins/git/git.plugin.zsh 'Create super useful gac and gacp aliases'` will add the plugin file with the commit message. You can list as many files/folders that you'd like to commit

gacp has the same functionality but will also push your commit


partounian:
Closed but if someone wants to see some more awesome git aliasing check this out https://github.com/partounian/dotfiles/blob/master/git/aliases.zsh

PR #6955, opened 2018-06-28
Add clean options to archlinux plugin

naortega:
I added some clean options for pacaur and pacman. If you'd like to add those for trizen and yaourt then that'd be nice.


mcornella:
Superseded by #10091.

PR #6959, opened 2018-06-29
Add argument to set644 and set755 perms alias.

ardeidae:
This can avoid big mistakes as this I just made.
I thought I could give it an argument. So, now, it is possible.

### DIFF
--- a/plugins/perms/README.md
+++ b/plugins/perms/README.md
@@ -4,6 +4,6 @@ Plugin to handle some unix filesystem permissions quickly
 
 ### Usage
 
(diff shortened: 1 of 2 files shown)


mcornella:
No need to rename the functions, just use `set644` and `set755` and you can get rid of the aliases.

ardeidae:
@mcornella, this is fixed. I didn't know how to modify my pull request...

ardeidae:
I've changed to accept any number of arguments. Readme updated too.
Is it ok for you?

mcornella:
Thanks!

PR #6960, opened 2018-06-30
removed alias irb='bundled_irb'

europ:
### bundled irb can lead to LoadError

Executing `irb` and performing some `require` commands inside of the *interactive ruby shell* in a directory, which contains `Gemfile` there is a possibility that it will result in a `LoadError`. This behavior is unexpected because irb is just a plain Ruby console which does not depends on Gemfile.

User who wish like to use only `irb` not `bundle exec irb` in a directory including Gemfile will have to face this potential `LoadError (cannot load such file -- XYZ)`. The fact that the user is not aware of the bundled irb may be very nuisance.

It was added via https://github.com/robbyrussell/oh-my-zsh/pull/2638

If the alias `irb` for bundled irb will be removed, it will be required to execute `be irb` for the bundled irb.

### DIFF
--- a/plugins/bundler/bundler.plugin.zsh
+++ b/plugins/bundler/bundler.plugin.zsh
@@ -15,7 +15,6 @@ bundled_commands=(
   foodcritic
   guard
   hanami
-  irb
   jekyll
   kitchen
   knife


mcornella:
I haven't found feedback either for or against bundling `irb`. There is thing https://github.com/rubygems/bundler/issues/6929 but as a total ruby noob I can't decide what to do, so I'm going to close this because I don't see a big interest in removing that.

In the meantime, you can use the [`UNBUNDLED_COMMANDS` variable](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/bundler#exclude-gems-from-being-wrapped) to exclude `irb` from being bundled.

Skipants:
Sorry for the necro, but I was pulling my hair out wondering what was changing how irb functioned until a Google search led me here. I saw that irb was aliased but assumed that it was coming from either asdf or rubygems/bundler.

The issue here is that `bundle exec irb`, which the ohmyzsh/bundler plugin is aliasing `irb` to, will only load gems that are specified by a Gemfile. This causes two issues that are unconventional, in my opinion:

1. It will not load global gems. eg. If I install a gem globally with `gem install 'rvm'` I cannot require it in a `irb` session. 
2. Running irb in any project that uses bundler will not work unless that project specifies `gem irb` in its Gemfile. Maybe our conventions are a bit different, but if developers are using tools that are not necessarily specific to a project I expect them to install it as a global gem locally and not put it in the project's Gemfile. This keeps it clean from a bunch of different dev's workflows.

Before using ohmyzsh I could simply `irb` anywhere and, if I needed the local Bundler context, run `require 'bundler/setup'` in my irb session. My current workaround is to `unalias irb`. I agree with the OP's PR and I personally think this is a bit of overstepping by ohmyzsh bundler's plugin.

PR #6970, opened 2018-07-06
agnoster: allow disabling VCS checks for performance

phiresky:
For me, this significantly decreases the prompt delay:

Disabling hg: 0.8s saved for every prompt (!)
Disabling bzr: 0.24s saved for every prompt

An alternative might be to check the existence of a repo some other way,
but since this has to be done recursively it's not trivial.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -24,6 +24,16 @@
 # hostname to whether the last call exited with an error to whether background
 # jobs are running in this shell will all be displayed automatically when
 # appropriate.
+#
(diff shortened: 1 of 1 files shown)


mcornella:
I won't pull this change. Agnoster is a very modifiable theme as it is and, as it happened with `prompt_context`, you can just disable specific functions with `func() {}`:
```
prompt_git() {}
prompt_hg() {}
prompt_bzr() {}
```

phiresky:
thanks for the info. did not realise i could directly override those functions after setting the theme

PR #6980, opened 2018-07-10
* Adding alias git branch description

yacine:
* Add alias gbdsc to git branch description
* Add alias gbedsc to git branch --edit-description


mcornella:
Not common enough as to warrant dedicated aliases.

PR #6984, opened 2018-07-11
add idris plugin (a functional language)

doofin:


mcornella:
The language is not popular enough to justify having a dedicated plugin within OMZ. You can add it to the [external plugins list](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins) though.

doofin:
That's not true . http://reddit.com/r/idris

PR #6986, opened 2018-07-12
pip: fix indexes URL

### LINKED ISSUE #6981: pip plugin does not create cache
#### Problem 
Pip plugin with default pip index url "https://pypi.python.org/simple/" produces an empty cache.

#### Reproduce
```zsh
$ zsh-pip-clear-cache
$ zsh-pip-cache-packages
$ cat ~/.pip/zsh-cache # nothing here
```
#### Reason
Manually calling
```zsh
$ curl https://pypi.python.org/simple/
```
produces a 301 response "Moved permanently". This curl command does not follow redirects, so the *pip.plugin.zsh* script does not receive the index of the packages.

#### Suggested fixes
- Make command: "curl -L https://pypi.python.org/simple/" in the *pip.plugin.zsh* script, to follow redirects.
- Change the pip index url to "https://pypi.org/simple/", which seems to be the final url.

Both fixes work in my setup. Kindly let me know which one is preferred and if you 'd like me to submit a PR.

biazzotto:
Fixes #6981.

### DIFF
--- a/plugins/pip/pip.plugin.zsh
+++ b/plugins/pip/pip.plugin.zsh
@@ -31,7 +31,7 @@ zsh-pip-cache-packages() {
       tmp_cache=/tmp/zsh_tmp_cache
       for index in $ZSH_PIP_INDEXES ; do
           # well... I've already got two problems
-          curl $index 2>/dev/null | \
+          curl -L $index 2>/dev/null | \
               zsh-pip-clean-packages \
                >> $tmp_cache
       done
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks! Can you also change the curl command to follow redirects? (`curl -L`)

drdavella:
Any chance this is going to be merged? Can someone push the change to use `curl -L`?

PR #6990, opened 2018-07-13
Add alias to prune remotely deleted branches from the local space

ricokareem:
Adding this command `gbp` _(initials stand for "git branch prune")_ to remove local, clean copies of merged and pruned branches.
Best run after:
`git remote prune <remote_name>`
**or** 
`git fetch --all --prune` _(alias: gfa)_


ricokareem:
Ah!  I should change the description.   `gbp` doesn't just check if the branch was merged to master/dev/develop, but rather if the remote was deleted altogether.  That would include spikes and other temporary branches that may not ever get merged, but are deleted remotely nonetheless.  It also will work if your integrating branch is not named "dev" or "develop".

mcornella:
Hmm okay after a bit of testing I see they don't do the same. I'm not sure that I'd want this one though because it has many requirements:

1. Branches need to be tracking their upstream counterparts, which is not always the case. Normally you do `git push --all` which updates the remote branches but doesn't automatically set them to track them.
2. A branch has to be deleted in the upstream repository (while not being merged into master). Normally that would mean discarding a pull request from GitHub, which doesn't happen that often to require an automated command.

Because of 1 and 2, I can't see the usefulness in having a dedicated alias for something that only someone with a very specific workflow would find useful. You can of course add it to your zshrc file or turn it into a dedicated gitconfig alias.

Thanks

ricokareem:
It's totally your prerogative.  The use case is for the day-to-day work on a company's code repository, and not for personal or open-source projects (it's probably not of much use in the latter cases).

For the day-to-day "company code" worker, the problem it's attempting to solve is getting rid of local copies of branches (branches usually not created by you) that litter your local workspace and may no longer need.  
Let's say that over 2-3 sprints, you have checked out a bunch of feature branches because you have been doing peer code reviews.  Those branches will remain in your local space forever until you manually delete them one by one.  `git pull -p` and `git fetch -p` only prunes the branches from the remotes/* directory.
The catch is that those features got rejected --- or for any myriad of reasons just recreated under a new name and the old one deleted (this happens **regularly** in companies where you have a lot of developers, some new-hires, some coding cowboys, etc. who are still learning that company's dev team conventions).

Whenever this happens, it becomes your problem and your problem alone.  The team member has deleted the offending branch, but you're still left with a local copy, and `gbp` will not remove it because it was not merged.  You're left with useless orphaned branches you didn't create, that over time, add up (imagine doing a `git branch` and having it scroll for pages).

_Note: for case 1, you mentioned above:  you usually **don't** want to prune local un-pushed branches that are of your own, since those are probably just WIP, still working on tests, refactoring, or any other reason you don't wish to share your code with the team yet._

PR #6993, opened 2018-07-15
kube-ps1: upgrade script from upstream

robsonpeixoto:
Get the new version from https://github.com/jonmosco/kube-ps1


jonmosco:
Some of the updates in thie PR are not necessary as they deal with bash settings. I just added in the updates from upstream and this PR can be closed once #6995 is merged. Thanks for the interest!

PR #6995, opened 2018-07-15
updating kube-ps1 to align with upstream changes

jonmosco:
Add changes from upstream to plugin.  README updated.
Fixes #6993 and #6512 
Also addresses issue from upstream project: https://github.com/jonmosco/kube-ps1/issues/57


jonmosco:
Some of the updates in thie PR are not necessary as they deal with bash settings.  I just added in the updates from upstream and this PR can be closed once #6995 is merged.  Thanks for the interest!

PR #7003, opened 2018-07-19
use mercurial plugin for hg info in ys theme

### LINKED ISSUE #7002: ys theme should use hg_prompt_info instead of custom code
As well as being unnecessary work, the logic is slightly wrong and it fails to identify mercurial subdirectories. I am preparing a PR to fix this.

jshholland:
Fixes #7002


mcornella:
Blocking: pull in hg prompt function to lib files.

PR #7007, opened 2018-07-19
Replace the current shell with the new ZSH instance

Luca0208:
This Replaces the currently running ZSH process with the new one using `exec` instead of creating a new Process.
This way, when the user `exit`s out of the the new shell it will not pop them back into the shell from which oh-my-zsh was installed from.


mcornella:
This has been pulled in #5169, which is now ready to test.

Thanks!

mcornella:
Closed by #5169.

PR #7012, opened 2018-07-20
Insert alias to checkout changed files

guiihlopes:
`git checkout -- .` undo changes for all files that is not added or commited. I added an alias `gcall` (git checkout (all)) (files).

Example:
```
> gcall .
> gcall path/to/file
```


mcornella:
Same as gco

mcornella:
You don't need the `--`.

guiihlopes:
If you have a branch name with the same `path/to/file` you will need it unless you want to `checkout` to your branch, but I do understand your point

PR #7014, opened 2018-07-23
Gradle Plugin: Also support settings files

Xerus2000:
You can change the location of the buildscript in the settings file, thus it should also be checked for.


Xerus2000:
Could you kindly merge this or state objections?

Xerus2000:
?

PR #7026, opened 2018-07-30
Get  list of conflicted files.

vrkansagara:
alias added for the list of conflicted files.


mcornella:
Just use one of the aliases to `git status`.

PR #7028, opened 2018-08-01
Add plugin README template

D-sha:
Add a default plugin readme template

I thought it may be helpful to have a template for plugin README's
then plugins can be required to have a readme before being merged.

Does anyone else think is a good idea?
Comments on format or content?

### DIFF
--- a/plugins/plugin.README.template.md
+++ b/plugins/plugin.README.template.md
@@ -1,0 +1,35 @@
+# README TEMPLATE
+The following is a template to be used when submitting plugins.
+Please fill out any relevant sections below and include with your plugin files:
+
+# Title
+
+* Description
+* What it does, benifits etc
+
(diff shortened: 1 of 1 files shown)


grikomsn:
As commented on https://github.com/robbyrussell/oh-my-zsh/issues/7175#issuecomment-425757996, it might be a good idea to make several kinds of readme templates for different kinds of plugins. It's a great start, though. 👍

mcornella:
I'm hesitant to merge this because it would mean having another file in the OMZ directory, I'd rather have this in the wiki. But I like the template! I think it covers most plugin types. Thanks!

dimensi0n:
Is this on the wiki ?

D-sha:
No its not i kinda forgot about it sorry, What section should it go in?

PR #7030, opened 2018-08-02
Fix: remove clashing of ubuntu plugin with 'ag'

mbologna:
Ubuntu plugin defines an alias named 'ag'.
The improved-grep called 'the_silver_searcher' (https://github.com/ggreer/the_silver_searcher)
is installed as `ag` in `/usr/bin`.
This patch changed the Ubuntu plugin aliases in order not to clash with 'ag' and
changes other aliases to keep consistency.


MenkeTechnologies:
i notice this conflict as well with silver searcher.  
However with other plugins, like the suse `zin` is alias for sudo zypper install and sc-stop is alias for sudo systemctl stop so the sudo is usually not taken into account.  However about just changing the ag to age

mbologna:
@MenkeTechnologies I changed the other aliases as well to keep coherence among all the aliases defined in the ubuntu plugin.

mbologna:
@Marcholio conflicts resolved. Ready to be merged.

mcornella:
Fixed in #7700.

PR #7034, opened 2018-08-03
Feat: Adding untacked file changes before upgrade

Gcaufy:
I modified some files in `.oh-my-zsh`.
Every times when I do a `upgrade_oh_my_zsh`, it throws an error, because my git repo is not clean.
So I have to go to the directory and do `git bash && git pull && git stash pop` manually.

I add this code to do it automatically, if you have untracked files, then stash for you.

This is the testing result:
uploadfile

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -20,8 +20,18 @@ else
   NORMAL=""
 fi
 
-printf "${BLUE}%s${NORMAL}\n" "Updating Oh My Zsh"
+STASH_NAME="STASH BEFORE UPDATING"
+
 cd "$ZSH"
+
+if [ -n "$(git status --untracked-files=no --porcelain)" ]; then
+  # Working directory clean excluding untracked files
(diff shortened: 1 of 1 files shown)


mcornella:
Using the stash is not the proper way to handle this. You should commit your changes instead. Then your upgrades will work correctly, unless there's a conflict (which you'll have to resolve yourself). You can also override files using our [Customization process](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization).

Gcaufy:
`Customization process` looks good for me. I can use that way.

but I still think we can apply `git stash`. once user changed something irrelevant. it's not my repo, and I don't want to commit the changes. why it stop me from upgrading. apparently we can pull down the updates. Even when the user changes have a conflict with the updates, no matter you do `git pull` or `git stash && git pull && git stash pop`, they both fail. so if no conflicts, then user do not need to be aware the changes they did.

So the reason I think it should be applied: 
1. Advantage for upgrading.
2. No risks.

Issue #7036, opened 2018-08-05
check_for_upgrade does not work on shell init


ccqpein:
Hi guys, I just noticed my shell do not notify me `[Oh My Zsh] Would you like to update? [Y/n]:` for a long while.

I check shell code of `check_for_upgrade.sh` and search where this script running. I find in `oh-my-zsh.sh` line 14. There is not any information show in my terminal when I run manually `env ZSH=$ZSH ZSH_CACHE_DIR=$ZSH_CACHE_DIR DISABLE_UPDATE_PROMPT=$DISABLE_UPDATE_PROMPT zsh -f $ZSH/tools/check_for_upgrade.sh`

I have changed `${ZSH_CACHE_DIR}/.zsh-update` to `LAST_EPOCH=0`. Also, when I run `check_for_upgrade.sh` line 41 to 53 in shell (I have set $epoch_diff and $epoch_target), it works fine.

At endding, I figure out the problem. For some reason, `$ZSH/log/update.lock` already exist then `if` in 32 line of `check_for_upgrade.sh` return directly without running body of code. 

I thought it might be nice if we rmdir once before we enter if expression.

ccqpein:
I add two lines to check `update.lock ` status and I sent pull request. 

[pull request 7037](
https://github.com/robbyrussell/oh-my-zsh/pull/7037)

mcornella:
Remove the lock manually and `upgrade_oh_my_zsh`, there is now a fix in place.

PR #7037, opened 2018-08-05
delete update.lock first incase it already exist

ccqpein:
Fix check_for_upgrade does not work on shell init

delete update.lock first incase it already exist

[issue 7036](https://github.com/robbyrussell/oh-my-zsh/issues/7036)

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -29,7 +29,7 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
-# delete update.lock if it alreday exist
+# Delete update.lock if it alreday exists
 rmdir $ZSH/log/update.lock 2>/dev/null
 
 if mkdir "$ZSH/log/update.lock" 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


D-sha:
I'm wondering if this would be better fixed by removing the final `rmdir $ZSH/log/update.lock` outside the final 'fi' thus ensuring it is always deleted?

ccqpein:
The probable	cause of this problem happened on my side might be something wrong in final if-fi make script exit before final `rmdir $ZSH/log/update.lock` runs. If we move `rmdir $ZSH/log/update.lock` outside and exit accidentally in if-fi, we still do not delete `$ZSH/log/update.lock`. And this problem will happen next time. However, `$ZSH/log/update.lock` will be deleted next time, and at next-next time, everything will be fine. But we lose second update chance.

D-sha:
This would defeat the purpose of the lock, would it not? If you opened two shells at the same time the second would delete the lock and both would try to update?
On looking at it I think that massive if block needs rewriting....

mcornella:
I recently pushed a fix for the lock not being deleted. This is no longer necessary.

PR #7051, opened 2018-08-09
Fix aws profiles list function

pkleanthous:


mcornella:
Can you post what results the completion gives before and after this PR is applied?

pkleanthous:
Running  aws_profiles before the PR
```
~ » aws_profiles
~ »
```

Running  aws_profiles after the PR
```
~ » aws_profiles
ww1-London
ww1
ww1-Ireland
ww2
~ »
```

mcornella:
See, the `aws_profiles` function is meant to be called only when using completion (in this case, of the `asp` function), that's why it uses the weird `reply=()` construct. But if you run the command inside it should give the proper results:
```
grep profile "${AWS_CONFIG_FILE:-$HOME/.aws/config}"|sed -e 's/.*profile \([a-zA-Z0-9_\.-]*\).*/\1/')
```

If that's the case this PR is not needed.

mcornella:
Fixed in master.

PR #7052, opened 2018-08-10
Improve common-aliases suffix alias rules (and respect `ZDOTDIR`)

alexander255:
Improvement for suffix alias rules:

  * Editable text files:
     * Greatly expanded list of matched text/source file formats (taken from gtksourceview)
     * Respect both `$EDITOR` and `$VISUAL` for text/source files
     * Do not launch text editor for source files with shebang and executable bit present
  * Image files:
     * Expanded list of matched image file formats
  * Media files:
     * Expanded list of matched audio/video file formats
     * Use `$XMVIEWER` environment variable instead of hard-coding `mplayer` as the one true media player
  * Document files:
     * Expanded list to also include important office formats
     * Launch these files using `(xdg-)open` instead of hard-coding random software packages (including *Adobe Acrobat Reader*)

Also look for `.zshrc` in `$ZDOTDIR` (if set) instead of hard-coding `~/`


alexander255:
How exactly do I get testers, if you don't mind me asking?

ntninja:
Still a thing ^

TheTechRobo:
you get testers by 

1) people interested in your pr
2) you can get 1) by telling people about it on the discord server or something

mcornella (changes requested):
This adds a lot of suffix aliases; I'd keep these to a minimum of commonly-used ones. I'd also separate this in 2: `text` documents that open with either `$VISUAL` or `$EDITOR`, and the rest that open in the default system program, which can be handled by the `open_command` function, e.g.:

```zsh
alias -s pdf='open_command'
```

It's fine to have image, media and document formats in separate arrays, but all should be suffix-aliased to `open_command`. Of course, do a few tests to see that the `open_command` function works correctly with them.

Sorry for the long wait to review this; I didn't yet have a working idea of how to handle the PR.

plugins/common-aliases/common-aliases.plugin.zsh
```diff
@@ -54,27 +54,110 @@ autoload -Uz is-at-least
...
+  if [[ -n "$VISUAL" ]] || [[ -n "$EDITOR" ]]; then
+    _aliases_launch_editor() {
+      # directly launch matched files with executable bit and shebang present
+      if [[ -x "${1}" ]]; then
+        read -r <"${1}"
+        if [[ ${#REPLY} -ge 3 ]] && [[ ${REPLY:0:2} = '#!' ]]; then
+          "$@"
+          return $?
+        fi
+      fi
```
mcornella, line 71: I think we can ignore whether the file has a shebang and execute it directly if it has an execute-bit, which is the behavior we'd get if no suffix-alias was present.

PR #7059, opened 2018-08-13
fix(gpg-agent): correctly overwrite `$SSH_AUTH_SOCK` and other improvements

### LINKED ISSUE #7082: gpg-agent plugin overwrites $SSH_AUTH_SOCK
https://github.com/robbyrussell/oh-my-zsh/blob/abca62add1a6d0ac34e697beac8682076d5c1dd7/plugins/gpg-agent/gpg-agent.plugin.zsh#L13
The above line doesn't check if `enable-ssh-agent` is actually enabled. It merely checks for the existence of the phrase.

The problem is if the phrase is present and commented out, it will trigger, e.g. `# enable-ssh-support allows gpg to act as a ssh agent`

Changing the grep command to `grep -qe '^enable-ssh-support' "$GNUPGCONFIG"` should eliminate this.

EDIT: I'd fork and open a PR, but the rate at which PRs are merged is a bit of a worry.

dpgao:
This commit improves several aspects of the plugin:

*   Get `SSH_AUTH_SOCK` directly from `gpgconf` and only when the current shell is not a child of `gpg-agent`.

    From [`man gpg-agent`](https://www.gnupg.org/documentation/manuals/gnupg/Agent-Examples.html#Agent-Examples):
    > If you enabled the Ssh Agent Support, you also need to tell ssh about it by adding this to your init script:
    > 
    >     unset SSH_AGENT_PID
    >     if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
    >       export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"
    >     fi
    The use of the `if` test here is explained [here](https://wiki.archlinux.org/index.php/GnuPG#Set_SSH_AUTH_SOCK):
    > **Note**: The test involving the `gnupg_SSH_AUTH_SOCK_by` variable is for the case where the agent is started as `gpg-agent --daemon /bin/sh`, in which case the shell inherits the `SSH_AUTH_SOCK` variable from the parent, `gpg-agent`.

*   Update the agent's tty before each command execution to ensure that passphrase prompts are displayed in the correct tty.

    This fixes a bug where if the plugin is used together with the `tmux` plugin and ssh support is enabled, `ssh` passphrase prompts will be displayed in the incorrect `tty` in some cases, causing the tmux window to close with a `server lost` message.

    The fix is implemented according to [`man gpg-agent`](https://www.gnupg.org/documentation/manuals/gnupg/Agent-Options.html#Agent-Options):
    > Note: in case the gpg-agent receives a signature request, the user might need to be prompted for a passphrase, which is necessary for decrypting the stored key. Since the ssh-agent protocol does not contain a mechanism for telling the agent on which display/terminal it is running, gpg-agent’s ssh-support will use the TTY or X display where gpg-agent has been started. To switch this display to the current one, the following command may be used:
    > 
    >     gpg-connect-agent updatestartuptty /bye

*   In addition, according to [`man gpg-agent`](https://www.gnupg.org/documentation/manuals/gnupg/Invoking-GPG_002dAGENT.html#Invoking-GPG_002dAGENT), `gpg-connect-agent` can also be used to start `gpg-agent`:
    > The agent is automatically started on demand by gpg, gpgsm, gpgconf, or gpg-connect-agent. Thus there is no reason to start it manually. In case you want to use the included Secure Shell Agent you may start the agent using:
    > 
    >     gpg-connect-agent /bye
    So we can start `gpg-agent` using this command without the need to check if it is already running.

@robbyrussell @mcornella Please review and provide feedback. Thank you!

Fixes #7082


dpgao:
@mcornella The latest commit incorporates the suggestion from #7082 by reading the `ssh-support-enabled` configuration option from `gpgconf`, so that we don't need to parse the configuration file ourselves. This technique was taken from [this mailing list message](https://lists.gnupg.org/pipermail/gnupg-users/2016-April/055859.html).

AdrienLemaire:
Not sure why there hasn't been any follow-up on this PR, but I'm afraid this will not work anymore.

```console
$  gpg-agent --daemon --use-standard-socket
gpg-agent[166139]: /home/dori/.gnupg/gpg-agent.conf:7: argument not expected
gpg-agent[166139]: WARNING: "--use-standard-socket" is an obsolete option - it has no effect
$ gpg --version
git version 2.24.1
```

$GNUPGCONFIG
```
pinentry-program /usr/bin/pinentry-curses
# cache once per session
max-cache-ttl 60480000
default-cache-ttl 60480000

# used by ohmyzsh gpg-agent plugin
enable-ssh-support 1
```
In my case, the only way to have this plugin work properly with ssh support is to comment the line ^^

dpgao:
@AdrienLemaire Does my patched version work for you?

AdrienLemaire:
@dpgao my bad! I thought the error meant that gpg-agent didn't support enable-ssh-support anymore, but after checking [the configuration doc](https://www.gnupg.org/documentation/manuals/gnupg/Agent-Configuration.html) and `gpg-agent --help`, I now understand that the config line should be `enable-ssh-support`  (without the following ` 1`). 
Will try your plugin now

AdrienLemaire:
@dpgao actually, something is wrong with this patched version for me. The gpg-agent doesn't seem to remember my passphrase for the whole session, and I'm continuously asked to type the gpg passphrase (also note that the pinentry title is ` Please enter the passphrase for the ssh key`, which is confusing because the ssh passphrase is different from the gpg passphrase.

dpgao:
@AdrienLemaire Is your secret key stored on your machine (as opposed to a smart card)? If so, have your added its keygrip to the `sshcontrol` file?

AdrienLemaire:
@dpgao indeed, no smart card.
In ~/.gnupg/sshcontrol, I have a 40 hex digits string (I believe this is the keygrip you're talking about) followed by 0 (no TTL caching). 
Actually, as I type these lines, I do not have further pinentry prompts asking for the gpg passphrase. Thanks for giving me a direction, I'll try debugging this further next time I get a prompt. :handshake:

AdrienLemaire:
@dpgao getting the prompt again, and could confirm that the keygrip was still in ~/.gnupg/sshcontrol before answering the passphrase.

chiselwright:
I arrived here via [this tweet](https://twitter.com/ohmyzsh/status/1363378865876295682). I have some interest in replacing my groady gpg-agent setup with a simple plugin, but I've never tested any oh-my-zsh changes before. Is there a recommended process to follow? (nothing leaps out from a quick internet search)

plugins/gpg-agent/gpg-agent.plugin.zsh
```diff
@@ -1,16 +1,13 @@
...
   unset SSH_AGENT_PID
+  if [[ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]]; then
+    export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"
+  fi
+  autoload -U add-zsh-hook
+  add-zsh-hook preexec _gpg-agent-update-tty
+  function _gpg-agent-update-tty {
+    gpg-connect-agent updatestartuptty /bye &>/dev/null
+  }
 fi
```
mcornella, line 13: Is this supposed to only be set when `enable-ssh-support` is set?

PR #7064, opened 2018-08-14
Add Git alias for `git diff --staged`

bittner:
An alternative alias for `gdca = git diff --cached`:

> `gds` = `git diff --staged`

Equivalent option according to [git-diff man page](https://linux.die.net/man/1/git-diff).

Mentioned in #3167.


bittner:
Thanks!

PR #7065, opened 2018-08-16
Remove the white space

DenniJensen:
The white space is causing an error and bundler cannot find any commands!

```sh
bundler: command not found: _rails_command
Install missing gem executables with `bundle install`
```


mcornella:
Nevermind, you must be running `be rails` with the [`rails` plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/rails/rails.plugin.zsh#L27-L31) loaded. I see no other choice than to revert that change then.

Issue #5703, opened 2016-12-10
Python anaconda


beatorizu:
This packages and environment manager is powerful and useful python tool and its could be a great idea if oh-my-zsh framework had a plugin for that

mcornella:
Reopen if there is still interest in this. FWIW, there exists [this one](https://github.com/dougalsutherland/anaconda.zsh) which you could clone in your custom folder, see [customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#adding-a-new-plugin).

PR #2508, opened 2014-02-07
ENH: add plugin for Anaconda Python environments

andreas-h:
Building on the virtualenvwrapper plugin, I wrote a plugin with the same functionality for users of the Anaconda Python distribution (which uses the conda packaging system).

PR #7067, opened 2018-08-17
Added current working anaconda env to agnoster prompt

John-P:
I've attempted to add the current anaconda env to the prompt for the agnoster theme. If someone else could test it or give some feedback it would be much appreciated. This is very similar to the current virualenv part of the prompt. This may also address issue #5703. Possibly related to PR #2508 although the agnoster virtualenv prompt component doesn't appear to use the virtualenv plugin in any way.

The addition only shows the env if not in the base env. This was in an attempt to follow the agnoster goal, "only show you *relevant* information", as if a user has anaconda installed then thier .zshrc or .bashrc will always start in the base env unless an additional conda activate command has been added.

The 'changeps1' option in the anaconda configuration is checked to avoid adding the env to the prompt twice (once unformatted and once formatted for agnoster).

I also noticed that there is an [agnoster repo](https://github.com/agnoster/agnoster-zsh-theme) and was not sure if i should send my pull request over there or here in oh-my-`zsh`.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -205,7 +205,7 @@ prompt_conda() {
   local conda_env="$CONDA_DEFAULT_ENV"
   if [[ -n $conda_env ]]; then
     if [[ -z $CONDA_PROMPT_MODIFIER ]]; then
-      prompt_segment blue black "($CONDA_DEFAULT_ENV)"
+      prompt_segment blue black "conda:$conda_env"
     fi
   fi
 }
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -205,7 +205,7 @@ prompt_conda() {
   local conda_env="$CONDA_DEFAULT_ENV"
   if [[ -n $conda_env ]]; then
     if [[ -z $CONDA_PROMPT_MODIFIER ]]; then
-      prompt_segment blue black "conda:$conda_env"
+      prompt_segment blue black "conda:$(basename $conda_env)"
     fi
   fi
 }
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -204,8 +204,7 @@ prompt_virtualenv() {
 prompt_conda() {
   local conda_env="$CONDA_DEFAULT_ENV"
   if [[ -n $conda_env ]]; then
-    local conda_changeps1="`conda config --show changeps1`" 
-    if [[ $conda_changeps1 == "changeps1: False" && $conda_env != "base" ]]; then
+    if [[ -z $CONDA_PROMPT_MODIFIER ]]; then
       prompt_segment blue black "($CONDA_DEFAULT_ENV)"
     fi
   fi
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -198,6 +198,19 @@ prompt_virtualenv() {
   fi
 }
 
+# Anaconda: current working conda env
+# To replace the conda modification of PS1 do `conda config --set changeps1 False`
+# or add 'changeps1: False' to your .condarc file.
+prompt_conda() {
+  local conda_env="$CONDA_DEFAULT_ENV"
+  if [[ -n $conda_env ]]; then
+    local conda_changeps1="`conda config --show changeps1`" 
+    if [[ $conda_changeps1 == "changeps1: False" && $conda_env != "base" ]]; then
+      prompt_segment blue black "($CONDA_DEFAULT_ENV)"
+    fi
+  fi
+}
+
 # Status:
 # - was there an error
 # - am I root
@@ -216,6 +229,7 @@ prompt_status() {
 build_prompt() {
   RETVAL=$?
   prompt_status
+  prompt_conda
   prompt_virtualenv
   prompt_context
   prompt_dir


themes/agnoster.zsh-theme
```diff
@@ -198,6 +198,19 @@ prompt_virtualenv() {
...
 }
 
+# Anaconda: current working conda env
+# To replace the conda modification of PS1 do `conda config --set changeps1 False`
+# or add 'changeps1: False' to your .condarc file.
+prompt_conda() {
+  local conda_env="$CONDA_DEFAULT_ENV"
+  if [[ -n $conda_env ]]; then
+    local conda_changeps1="`conda config --show changeps1`" 
+    if [[ $conda_changeps1 == "changeps1: False" && $conda_env != "base" ]]; then
```
avrahamshukron: You basically invoke `conda` every time `PS1` is evaluated. This will incur a heavy performance penalty on the user.
Instead, you could check the `CONDA_PROMPT_MODIFIER` env variable.

When `changeps1` is set to `true`, this var will be `(<env name>)`.
When it is `false`, the var will be empty.
In addition, I don't see a reason to hide the `base` env. I would really like to know if I'm activated or not.

```suggestion 
    if [[ -z $CONDA_PROMPT_MODIFIER ]]; then
```

themes/agnoster.zsh-theme
```diff
@@ -198,6 +198,19 @@ prompt_virtualenv() {
...
+      prompt_segment blue black "($CONDA_DEFAULT_ENV)"
```
avrahamshukron: Well, I know that the convention for `venv` is to wrap the env name in `()`, but - why? it doesn't add anything to the user.
May I suggest:
```suggestion
      prompt_segment blue black "conda:$conda_env"
```

This will let the user know that they are in a conda env name `$conda_env`.

Attached is an example
image

avrahamshukron (changes requested):
Some performance-affecting changes are required, and some cosmetic changes suggested

themes/agnoster.zsh-theme
John-P: Thank you for your suggestions on performance and formatting. The `()` was to be consistent with conda but I agree that your formatting is much more clear.

avrahamshukron: Sure thing. I hope your PR will be accepted soon. I actually thought that this PR is abandoned because it was opened since Aug 2018...

The format works for me, because I use both venv and conda for different tasks. I never understood those () around the env name anyway...

avrahamshukron:
@mcornella could you take a look at this PR? I think it is ready for merge.

themes/agnoster.zsh-theme
```diff
@@ -198,6 +198,18 @@ prompt_virtualenv() {
...
 }
 
+# Anaconda: current working conda env
+# To replace the conda modification of PS1 do `conda config --set changeps1 False`
+# or add 'changeps1: False' to your .condarc file.
+prompt_conda() {
+  local conda_env="$CONDA_DEFAULT_ENV"
+  if [[ -n $conda_env ]]; then
+    if [[ -z $CONDA_PROMPT_MODIFIER ]]; then
+      prompt_segment blue black "conda:$conda_env"
```
avrahamshukron: I just learned that `$CONDA_DEFAULT_ENV` might contain the full env path (like `$CONDA_PREFIX`) and not just the env name.

This happens when you create a dedicated env outside of conda's default env location, e.g with command like:
```sh
conda create --prefix /arbitrary/path/to/myenv
```

So in order to prevent the promp to be too long it is better to `basename` it.
```suggestion
      prompt_segment blue black "conda:$(basename $conda_env)"
```

avrahamshukron (changes requested):

themes/agnoster.zsh-theme
John-P: An excellent suggestion. I can see how this could easily make the prompt far too long. I think it would be nice to somehow indicate that it is an environment in a non-default location if possible. I cannot think of a nice way to do it though. Perhaps by changing the segment colour to a different colour such as yellow?

kbakdev (approved):

avrahamshukron (approved):

michcio1234:
@avrahamshukron @53jk1 will this be merged? :)

avrahamshukron:
@michcio1234 I wish, but I don't have permissions to merge...

What I did in the meantime is to override the function in my own zshrc, because I really need this feature.

53jk1:
I'm not a collaborator.

avrahamshukron:
@mcornella Is there a reason this feature should not be merged?

robbyrussell:
Thanks for this, and I'm sorry... this got two approving reviews back in 2020 and then sat here anyway. That shouldn't have happened.

Conda support did land in agnoster eventually, in #13160 last June, so I'm closing this one out.

But your version had something the one that shipped didn't: you skipped the segment when conda's own `changeps1` was already printing the env name, so it wouldn't show up twice. You also ran it through `basename` so a path-activated env didn't dump the whole path into the prompt.

I've just opened #14042 to bring both of those over, with you credited on the commit. Turned up a third one while I was in there... the shipped version doesn't escape `%`, so an env name containing one breaks the prompt.

Sorry again for the wait, and thanks... the idea was right, it just took us far too long.

PR #7073, opened 2018-08-17
Show which git remote is set for upstream along with branch

brunoborges:
This change shows in the prompt which remote is set as upstream for the current branch.


MattLewin:
I really like this change, as git frequently yells at me when I try to push or pull and haven't set the upstream branch. The one issue I have with this change is that it doesn't display the branch name if there *is* no upstream branch. When I'm working on something that doesn't warrant a remote branch, I still want to know what branch I'm on. This is a fairly frequent situation in my experience, especially when working on a small team and using git flow.

Is there a way to modify the prompt to display the local branch if there is no remote branch?

Thanks!

brunoborges:
Thanks @MattLewin. The goal of this PR is to show which remote is set as upstream of the branch user is on. 

I don't understand your ask though. Mind to provide an example?

MattLewin:
@brunoborges, here is a screenshot of a terminal window that demonstrates what happens when I update `git_prompt_info()` with your changes.

screen shot 2018-08-30 at 8 11 57 pm

1. Using `lib/git.zsh` from `master`, you can see on the right side that I'm on `develop`. I then switch to the local-only branch, `foobar`.
2. The right side shows I am on `foobar`. I switch back to `develop`.
3. I have your changes to `git_prompt_info` in `~/.oh-my-zsh/custom/lib/git.zsh` and use `src` to reload all my zsh goodness. 
4. As you can see, the right side shows `origin/develop`. I then check out `foobar`.
5. Now the right side is blank, presumably because `foobar` has no upstream branch. Checking out `develop` once again...
6. We see `origin/develop` displays as you intended.

Does this make sense? Could this be resulting from something specific to my theme or some (mis)configuration I've got?

brunoborges:
Are you using the change I made on this PR?

MattLewin:
For reasons I don’t understand, your question below is not showing up on the web.

I assume you are asking this question in response to my recent post with the screenshot.

At step #3, where I type ‘src’, I load the changes you made in your PR. My hope was to demonstrate to you my perceived different between what is in the master oh my zsh branch and your PR for lib/git.zsh.

If you did not see my post with the screenshot, let me know, and I will stick it in an email.

brunoborges:
@MattLewin see if change in 8077616 solves it

MattLewin:
I've been @brunoborges's patch for nearly two months now, and I'm uncertain how I lived with out it.

Any reason this can't be merged into `master` so everyone can benefit?

MattLewin:
@brunoborges, I tripped over another area for improvement: the case where you are operating in "detached" mode.  I could not, for the life of me, figure out how to fork from your `patch-1` branch and make the necessary changes to handle for this situation.  Instead, I am attaching the patch to the file for you to incorporate into the PR. The `git_prompt_info` function now looks like this:

```shell
function git_prompt_info() {
  local ref
  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
    if branch=$(command git symbolic-ref HEAD 2> /dev/null); then
        # Branch name in the form "<upstream branch name>/<brach name>"
        ref=$(command git for-each-ref --format='%(upstream:short)' ${branch} 2> /dev/null) || return 0
        info=$(command echo "${ref#refs/heads/}")
        if [[ "${info}" == "" ]]; then
            # Handle for no remote branch
            info=$(command echo "<no-remote>/$(git rev-parse --abbrev-ref HEAD)" 2> /dev/null) || return 0
        fi
    elif [[ "$(command git tag --points-at 2>/dev/null)" != "" ]]; then
        # Detached at a tag
        info=$(command git tag --points-at 2>/dev/null)
    else
        # Detached at unnamed revision
        info=$(command git rev-parse --short HEAD 2> /dev/null) || return 0
    fi
    echo "$ZSH_THEME_GIT_PROMPT_PREFIX${info}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
  fi
}
```
[patch-ml.txt](https://github.com/robbyrussell/oh-my-zsh/files/2539507/patch-ml.txt)

brunoborges:
I will give it a try @MattLewin , thanks!

brunoborges:
Indeed, the detached mode is a problem.

brunoborges:
@MattLewin just pushed a better way to check for init-only folders. On 6dc67e3.

brunoborges:
@mcornella any chance this can be merged?

brunoborges:
Closed in favor of #9188

megbxk79 (approved):

megbxk79 (approved):

PR #7078, opened 2018-08-20
uVenv Plugin init version

vincentto13:
This is initial version of plugin, which helps to manage python
projects and their virtualenvs.


mcornella:
While I appreciate the effort, we already have plugins to manage virtualenv. I'd suggest working on it more and creating your own repository. Since you've made it into a plugin, you can also add it to our [External plugins list](https://github.com/robbyrussell/oh-my-zsh/wiki/External-plugins).

Cheers

vincentto13:
I found virtualenv plugin within plugins directory which seems to be less feature rich version of the plugin I pushed. Do you think it is reasonable to replace/update that plugin?

Issue #6989, opened 2018-07-12
Propose to show `git log -p` before the upgrade


techtonik:
It is similar to #2971 but doesn't require any manual log housekeeping. Basically, modify:
```
[Oh My Zsh] Would you like to update? [Y/n]: 
```
to
```
[Oh My Zsh] Would you like to update? [Yng?]: ?
Yy - yes
Nn - no
Gg - show git log with patches
```

robbyrussell:
Given that we've since replaced the updating process with a new formatting to show what changes have been made, I'm going to close this.

Would welcome some PR ideas for how we could, potentially, show what they _would_ be getting when they respond with Yes

PR #7083, opened 2018-08-22
Added options to update menu. Included git log command (#6989)

charliecjung:


mcornella:
This will only show the current commits of Oh My Zsh, that is, the commits of the current version. If you wanted to show the update patches you would have to `git fetch` before.

charliecjung:
added "git fetch" before showing git log.

mcornella:
This doesn't respect the user decision to either update or not (a `git fetch` already fetches all new commits).

PR #7086, opened 2018-08-23
.gitingore prevents adding itself, so breaks the function.

QwertyJack:
Change `git add .gitignore` to `git add -f .gitignore`.


mcornella:
Thanks!

PR #7088, opened 2018-08-25
DietPi Plugin

ichbestimmtnicht:
Hello,
I made a Plugin for the DietPi https://github.com/Fourdee/DietPi System. Installing zsh is a bit tricky (sudo without path doesn't work and most of them need sudo) and so other users don't have to create it themself.

Greetings
ichbestimmtnicht


mcornella:
This shouldn't be necessary, see https://github.com/Fourdee/DietPi/issues/424

PR #7090, opened 2018-08-26
Clarify random theme setting

slel:
Clarify random theme setting, in particular how to know
which theme was loaded if picked at random.


robbyrussell:
Hey there -- I really appreciate the tidying up!

Cheers,
Robby

PR #7093, opened 2018-08-27
Add aliases for kubectl nodes

InTheCloudDan:
I saw there was a couple older PRs opened for various parts of node management but I don't think they all follow the established format and cover the entire get/edit/describe/delete convention.


robbyrussell:
Thanks, @InTheCloudDan !

PR #7095, opened 2018-08-28
Updating 'sbcl' to 'sbcln'

ryanstull:
Changing 'sbcl' to 'sbcln' so it doesn't collide with 'Steel Bank Common Lisp', a popular lisp implementation.

@mcaserta please review


mcaserta:
@ryanstull ok, no problem. Thanks.

PR #7098, opened 2018-08-28
Adding the Kafka plugin

Dabz:
It's a very simple plugin that do a bunch of _describe on the most convenient CLI tools of Apache Kafka. 
This file is actually automatically generated from the help menus, the whole source code is available on https://github.com/Dabz/kafka-zsh-completions.

It's probably not perfect at all, so if you have any feedback, I take!

BTW: Code generation is pure evil, unfortunately I didn't find a better way to avoid starting a JVM at startup...


ciudilo:
@mcornella would be awesome to have this plugin in oh-my-szh.

robbyrussell:
While this looks interesting, we aren't seeing a lot of interest from the community (in +1/votes) so going to close this. Sorry it took so long! 

If this is something the community is interested in, please let us know!

PR #7105, opened 2018-08-31
Fixes jump plugin printf path output

rlafferty:
Using the `jump` plugin, using the `marks` command will yield this output:
```
$ marks
desktop marks:printf:5: bad option: ->
dotfiles        marks:printf:5: bad option: ->
home    marks:printf:5: bad option: ->
```

the `marks` function uses `printf` with `->` and I believe `-` is used by `printf` for left-justification.  changing this to `-- "->"` seems to render the appropriate output.

```
desktop -> /Users/uname/Desktop
dotfiles        -> /Users/uname/.dotfiles
home    -> /Users/uname
```

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -32,7 +32,7 @@ marks() {
 		local markname="$fg[cyan]${link:t}$reset_color"
 		local markpath="$fg[blue]$(readlink $link)$reset_color"
 		printf "%s\t" $markname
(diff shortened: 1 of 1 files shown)


mcornella:
We can use `--` to stop the parsing of hyphen options, like so: `printf -- ->...`.

This way we keep the same format while fixing the errors.

rlafferty:
@mcornella - good idea!  Updated.

mcornella:
Thanks!

PR #7119, opened 2018-09-06
fzf: Fix shell directory for archlinux package

### LINKED ISSUE #7110: [fzf plugin] Wrong key-bindings.zsh folder
I installed fzf using Arch Linux package manager. 
I just activated the fzf plugin.
Now, when I open a new shell, I get the following error message:

```
/home/XXXX/.oh-my-zsh/plugins/fzf/fzf.plugin.zsh:source:42: no such file or directory: /usr/share/fzf/shell/key-bindings.zsh
```

Indeed, my default `fzf_base` is `/usr/share/fzf`. But [here](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/fzf/fzf.plugin.zsh#L42), you try to `source "$fzf_base/shell/key-bindings.zsh"`. But my `fzf_base` folder contains: `completion.bash  completion.zsh  key-bindings.bash  key-bindings.zsh`

Ping @maxbrunet

maxbrunet:
A small change to fix #7110, I keep it minimal to stay as close possible to the [upstream install script](https://github.com/junegunn/fzf/blob/8540902a35361dba217cc3ad86ff2215ae51918e/install#L266-L278).

I also made a change to not clutter PATH if `fzf` is already available.

I tested these changes on the following `fzf` installations:
* MacOS + brew
* Archlinux + Pacman
* Fedora + RPM
* Linux + Git

### DIFF
--- a/plugins/fzf/fzf.plugin.zsh
+++ b/plugins/fzf/fzf.plugin.zsh
@@ -24,6 +24,13 @@ fi
 
 if [[ -n "${fzf_base}" ]]; then
 
+  # Fix fzf shell directory for Archlinux package
+  if [[ ! -d "${fzf_base}/shell" ]] && [[ "$(uname -n)" == 'archlinux' ]]; then
+    fzf_shell="${fzf_base}"
+  else
+    fzf_shell="${fzf_base}/shell"
+  fi
+
   # Setup fzf
   # ---------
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks Maxime!

PR #7127, opened 2018-09-10
dash: update dash bundle identifier

voldemortensen:
At some point the bundle identifier for Dash updated.

image


mcornella:
Thanks!

PR #7128, opened 2018-09-10
Added aliases for pikaur aur helper

shanff:


homocomputeris:
An alias for 

> **How to upgrade all the dev (-git) packages at once?**
> `pikaur -Syu --devel --needed`

would be great

now-im:
I just did this myself. To be sure, I searched the pull and here it is!
@shanff One minor suggestion. If the aliases are `pk` rather than `pik`, I think it would also work great.

robbyrussell:

I'm closing it, but not because we don't want pikaur... the plugin does still take AUR helpers, and pikaur is about as widely used as trizen, which is already in there.

The problem is this particular branch can't be salvaged: your fork is gone, so there's nothing to rebase. The diff also brings back an `abs`/`aur` block that was removed in #9803, and it defines `trupd`, which trizen already uses.

If you still use pikaur and want to send a fresh one against current master, it'd be welcome... a `pik*` alias block plus a README table row, matching how yay and trizen are done. No pressure either way after this long.

PR #7132, opened 2018-09-12
Added autocomplete support for poetry 

declared-lambda:
Added autocomplete for [poetry](https://github.com/sdispater/poetry).

### DIFF
--- a/plugins/poetry/_poetry
+++ b/plugins/poetry/_poetry
@@ -1,0 +1,143 @@
+#compdef poetry
+
+_poetry_d9a6c1dd06bd2d26_complete()
+{
+    local state com cur
+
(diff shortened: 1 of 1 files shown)


mcornella:
You can just add this to your zshrc file:
```
source <(poetry completions zsh)
```

declared-lambda:
Ah my bad.

declared-lambda:
Using that in zshrc produces the following:

```
source <(poetry completions)                             1
_tags:comptags:36: can only be called from completion function
_tags:comptry:55: can only be called from completion function
_tags:comptags:60: can only be called from completion function
_tags:comptags:67: can only be called from completion function
```

PR #7134, opened 2018-09-12
add aliases for newer iPhone models

esthor:
Added aliases for iphone 7, 7 Plus, 8, 8 Plus, SE, and X


mcornella:
Thanks. Would be great having a README documenting them, if you're up for the task.

esthor:
Haha...sure, I can do that.

esthor:
Added the documentation for them. I will try to remember to add new aliases later this month when new ios simulators are made available.

PR #7135, opened 2018-09-13
update react-native readme to document new aliases

esthor:
Added documentation of the new aliases added to the 'plugins/react-native'


mcornella:
Thanks!

PR #7140, opened 2018-09-17
svn-fast-info: fix svn-fast-info printfs

Melebius:
The `printf` command in the `svn-fast-info` plugin was not formed well, empty variables would mess things up. Moreover, there were superfluous spaces, resulting in a bloated info.

This corrects the PR #5813 which, apart from the correction, introduced an unintentional folder rename. The commits in this pull request should be squashed.


Melebius:
@mcornella Thanks for viewing this. May I ask you why you added `Status: conflicts`? Where are the conflicts? Both GitHub web and Git on my machine don’t report this branch as conflicting with `master`.

mcornella:
These labels were applied automatically, I'll see if there was an error in the script. As per the change, seeing a screenshot of the differences between the old and the new version would really help move this PR forward.

mcornella:
Nevermind I see it's just quoting and an extra `%s` in printf to make it coincide with the number of arguments. :+1:

PR #7146, opened 2018-09-20
added new iphone model simulators to readme

esthor:
Added new iPhone XR, XS and XS Max simulator aliases to the README documentation

--Sorry, should have made it all one PR


mcornella:
Thanks!

Issue #4413, opened 2015-09-28
gitfast is GPL


apjanke:
Some of the code in the `gitfast` plugin is GPL (GNU Public License). 

[git-prompt.sh](https://github.com/robbyrussell/oh-my-zsh/blob/b05ef1034fffd7b0c29f1558a8a357ccce7ae74b/plugins/gitfast/git-prompt.sh#L3-L4)

```
# Copyright (C) 2006,2007 Shawn O. Pearce <spearce@spearce.org>
# Distributed under the GNU General Public License, version 2.0.
```

[git-completion.bash](https://github.com/robbyrussell/oh-my-zsh/blob/b05ef1034fffd7b0c29f1558a8a357ccce7ae74b/plugins/gitfast/git-completion.bash#L3-L5)

```
# Copyright (C) 2006,2007 Shawn O. Pearce <spearce@spearce.org>
# Conceptually based on gitcompletion (http://gitweb.hawaga.org.uk/).
# Distributed under the GNU General Public License, version 2.0.
```

[_git](https://github.com/robbyrussell/oh-my-zsh/blob/b05ef1034fffd7b0c29f1558a8a357ccce7ae74b/plugins/gitfast/_git#L5)

```
# Copyright (c) 2012-2013 Felipe Contreras <felipe.contreras@gmail.com>
```

This is incompatible with the MIT License that Oh My Zsh is licensed under. Looks like this code should be pulled, or we should add a mechanism and doco about including code using other licenses.

Most of the stuff in `gitfast` appears to simply be a copy of the completion code shipped with the `git` distribution itself. (E.g. in my Homebrew-managed `git`, similar files are installed to `/usr/local/share/zsh/site-functions`.) You could probably change the plugin to load it from the `git` installation on the local machine instead of carrying the code around in OMZ, which would fix the licensing issue.

felipec:
I'm the author of the `_git` script from Git anyway, and I don't have any problem with using an MIT license.

The other two are a problem.

The `git-completion.bash` script isn't actually needed, we could use whatever is installed in `/usr/share/bash-completion`. I'm fine with deleting it. I only put it in gitfast so we could have the latest that is actually tested.

But `git-prompt.sh` is trickier, because there is no standardized way of distributing such scripts. If this is a big deal we could ask the main authors if the file could be relicensed. I'm the main author, but there's at least one other.

robbyrussell:
[...]

PR #7152, opened 2018-09-24
Update gitfast plugin

### LINKED ISSUE #6137: gitfast: support git's worktree command in tab completion
The `compdef` files for git and gitfast plugins can be extended to include `worktree` definitions.

### LINKED ISSUE #6206: gitfast doesn't have to load the git plugin
The gitfast plugin does not depend on the [git plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/git), which defines a set of shell aliases. I don't like that many aliases, but I need gitfast. Could you help to prevent gitfast from loading the git plugin? One can always enable it by adding it in `.zshrc`.

mcornella:
- Updates git completion and git-prompt to v2.10 (partially fixes #6137)
- Stops loading the git plugin (fixes #6206)

@felipec can you check out if I didn't break anything, and also if we could update the completions to a newer version?

#### TODO:
- [ ] fix GPL incompatibility (#4413)

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -67,6 +67,15 @@ __gitcomp ()
 	esac
 }
 
+__gitcomp_direct ()
+{
+	emulate -L zsh
+
+	local IFS=$'\n'
+	compset -P '*[=:]'
+	compadd -Q -- ${=1} && _ret=0
+}
+
 __gitcomp_nl ()
 {
 	emulate -L zsh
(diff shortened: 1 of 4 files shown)


felipec:
The only thing you missed is a change for $OSTYPE. I added a script to automatically update the version of upstream (87b435b7128dfa6e070bff2669d039c143d409ad) to avoid those issues. Not that it matters much.

PR #7154, opened 2018-09-24
Add docker and git aliases + minor syntax correction

EVODelavega:
Added a couple of aliases for docker, docker-compose, and git plugins, also changed the incorrect hybrid function definitions (`function name() {` is a cross between ksh's `function foo {` and POSIX `foo() {`), and changed them for the POSIX compliant version.

Aliases added:

docker:

- `drmi`: removes dangling images
- `dwipe`: kills all containers, removes them, and removes all images (blank slate)

docker-compose:

- `dcdv`: docker-compose down -v

git:

- gpcb: pushes current branch, if it's not tracking a remote branch (ie new branch), it runs `git push --set-upstream origin $(git_current_branch)`
- gupff: git pull -ff (for those who don't want all the aliases to be rebase-oriented)

The git-functions rely on a syntax that only works by accident. The function keyword is supported to be compatible with ksh, but it's essentially ignored if the function name is followed by `()`. The POSIX (and correct bourne-based shell) style is `<fname>() {}`.


plugins/docker/_docker
```diff
@@ -3023,6 +3023,12 @@ _dockerd() {
 
 _docker "$@"
 
+# aliases 
+# remove docker intermediate images
+alias drmi='docker rmi -f  $(docker images -aq --filter dangling=true) 2> /dev/null'
```
slmg, line 3028: What is the benefit of this command over `docker image prune` out of interest? As far as I know only dangling images are removed (in version 19.03 at least)

mcornella:
This does too many things at once and some of them are wrong (e.g. aliases in `_docker` file). Reopen if you're still interested in the changes and I'll update with a more detailed review.

PR #7165, opened 2018-09-27
golang: basic module completions

icholy:


mcornella:
Oh my bad, I saw #7665 first.

PR #3215, opened 2014-10-12
add a space.

overtrue:
before:
image

after: 

image

PR #7169, opened 2018-09-28
Remove spaces in the cloud theme

lePig:
on macOS Mojave and macOS High sierra

qq20180928-232821 2x


jounathaen:
I don't know how this theme works, but it seems, that besides removing the space, you also changed the unicode symbol from `⚡` to `✗`. I'm just asking if this was on purpose.

jounathaen:
Besides this, the space was added recently in #3215 . So I think you should mention @overtrue and @ncanceill in this pull request, and argue why it is necessary to re-add the space.

lePig:
@jounathaen sorry. unicode symbol from ⚡ to ✗ is a mistake。so i'll close this issue first. Thx

PR #7172, opened 2018-09-29
updater: add --autostash to git pull

bitbonsai:
If I have custom configs (like theme customizations) I have to stash my changes and get them back after the update. 

By adding the --autostash on upgrade.sh, if I have any changes not commited they'll be reapplied after the upgrade, allowing me to have temporary customizations without any harm to the upgrade process.

Fixes #4753


stephwestgarth:
I've tested this locally and it works great. Super helpful when I have a few small changes in my configs that aren't worth creating a whole fork for.

jgranzow86:
This change has broken the upgrade for me as my version of git (1.8.3.1) that comes with macOS (10.15.3) doesn't have the '--autostash' option.

mcornella:
Does this work (`cd $ZSH` first)? `git -c rebase.autoStash=true pull --rebase --stat origin master`

mcornella:
I've committed this change. You're gonna have to manually do the `pull --rebase` first, then it will function normally on subsequent automatic upgrades. You won't get the autostash behavior, but it won't error out anymore.

jgranzow86:
Thank you that worked perfectly, I was hoping there was a way I didn't know of, to keep this useful change.

PR #7173, opened 2018-09-29
Add option to color any help command

revolter:
This allows you to use `colored git log --help` for example, to get
colored output.


revolter:
Could I get a review on this one?

mcornella:
:+1:, but we should probably change the plugin name in the future to something like `colored`. Can you add a basic README? See https://github.com/robbyrussell/oh-my-zsh/issues/7175#issue-365263634 for instructions.

revolter:
@mcornella, Done. Let me know if the wording is ok.

PR #7174, opened 2018-09-30
ssh-agent: autoload identities not already loaded

### LINKED ISSUE #3019: ssh-agent Plugin on Mac OS X doesn’t work because launchd set the $SSH_AUTH_SOCK
As I see, [ssh-agent plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/ssh-agent/ssh-agent.plugin.zsh#L61) launches `ssh-agent` only if variable `$SSH_AUTH_SOCK` is empty. But on Mac OS X it is not empty:

```
$ echo $SSH_AUTH_SOCK
/tmp/launch-aY7Yo1/Listeners

$ lsof $SSH_AUTH_SOCK
COMMAND PID  USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
launchd 244 isqua   11u  unix 0x2e8ad5a2c9398243      0t0      /tmp/launch-aY7Yo1/Listeners
```

When I launch `ssh-agent` it (as i understand) doesn’t connect to this socket:

```
$ ssh-agent
SSH_AUTH_SOCK=/var/folders/56/bq22bg3d4xz6w81jxj8c3b90msfw89/T//ssh-jd61LOgtorda/agent.2619; export SSH_AUTH_SOCK;
SSH_AGENT_PID=2620; export SSH_AGENT_PID;
echo Agent pid 2620;

$ lsof -c ssh-agent | grep unix
ssh-agent 2620 isqua    3u    unix 0x2e8ad5a2cd0617bb       0t0          /var/folders/56/bq22bg3d4xz6w81jxj8c3b90msfw89/T//ssh-jd61LOgtorda/agent.2619
```

Why does the launchd create the socket and set the variable? How ssh-agent should use it? How should I modify the plugin to make it work?

### LINKED ISSUE #6979: ssh-agent plugin doesn't load multiple identities
Hi,
I'm using `ssh-add` plugin https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/ssh-agent. In the README.md, it says this:
```
To load multiple identities use the identities style, For example:
zstyle :omz:plugins:ssh-agent identities id_rsa id_rsa2 id_github
```

I already added this in my `~/.zshrc` before the line that sources oh-my-zsh like this:
```
zstyle :omz:plugins:ssh-agent identities id_rsa prd-key stg-key
source $ZSH/oh-my-zsh.sh
```
My `.ssh/` folder is having these keys:
```
 ll .ssh
-rw-------@ 1 122462    1.6K Feb 27 11:20 id_rsa
-rw-------@ 1 122462    1.7K Mar 28 14:54 prd-key
-rw-------@ 1 122462    1.6K Feb 27 11:26 stg-key
```

But when I executed `ssh-add -l`, only `id-rsa` was shown instead of all 3 keys:
```
ssh-add -l
2048 .ssh/id_rsa (RSA)
```

I'm expecting something like this:
```
ssh-add -l
2048 .ssh/id_rsa (RSA)
2048 .ssh/prd-key (RSA)
2048 .ssh/stg-key (RSA)
```
How can I solve this ? Did I miss anything ?

mcornella:
With this PR the ssh-agent plugin checks the `ssh-add -l` output for the identities added, and adds all those specified by the user that haven't been added yet.

We also decouple the logic of starting ssh-agent from the logic of adding identities, meaning that even if ssh-agent has been started by some other means (like [launchd](https://github.com/robbyrussell/oh-my-zsh/issues/6979#issuecomment-411088832)) we can still ssh-add the user's identities.

Fixes #3019
Fixes #6979

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -2,20 +2,27 @@ typeset _agent_forwarding _ssh_env_cache
 
 function _start_agent() {
 	local lifetime
-	local -a identities
-
-	# start ssh-agent and setup environment
 	zstyle -s :omz:plugins:ssh-agent lifetime lifetime
 
+	# start ssh-agent and setup environment
+	echo starting ssh-agent...
 	ssh-agent -s ${lifetime:+-t} ${lifetime} | sed 's/^echo/#echo/' >! $_ssh_env_cache
 	chmod 600 $_ssh_env_cache
 	. $_ssh_env_cache > /dev/null
+}
 
-	# load identies
+function _add_identities() {
+	local id line
+	local -a identities ids
 	zstyle -a :omz:plugins:ssh-agent identities identities
 
-	echo starting ssh-agent...
-	ssh-add $HOME/.ssh/${^identities}
+	# get list of loaded identities
+	for line in ${(f)"$(ssh-add -l)"}; do ids+=${${(z)line}[3]}; done
+
+	# add identities if not already loaded
+	for id in ${^identities}; do
+		[[ ${ids[(I)$HOME/.ssh/$id]} -le 0 ]] && ssh-add $HOME/.ssh/$id
+	done
 }
 
 # Get the filename to store/lookup the environment from
@@ -42,6 +49,8 @@ else
 	_start_agent
 fi
 
+_add_identities
+
 # tidy up after ourselves
 unset _agent_forwarding _ssh_env_cache
-unfunction _start_agent
+unfunction _start_agent _add_identities


trepanning-io:
I can confirm this fix worked in my setup: macOS 10.14.2, ZSH 5.6.2, OMZ 2614b7e.

It fixes the problem described.

bcollard:
it works, please accept PR 👍

bcollard:
it solves #6979 and might fix this one as well: #7352

mcornella:
Thanks all!

arcenik:
With this MR, one of my ssh id is re-loaded each time because the *ssh-add -l* command show the comment instead of the id file. I've made a fix and a MR #7504

wridgers:
This patch seems incompatible with ed25519 keys, because `ssh-add` lists the comment not the path.

```
will@desktop [09:37:45] [~] 
-> % ssh-add -l
4096 SHA256:/XYZA /home/will/.ssh/id_something (RSA)
256 SHA256:XYZB will@desktop (ED25519)
2048 SHA256:/XYZC /home/will/.ssh/id_other (RSA)
```

Currently I'm prompted for the `256 SHA256:XYZB will@desktop (ED25519)` key passphrase every time zsh starts.

My `.zshrc`:

```
will@desktop [09:46:19] [~] 
-> % cat ~/.zshrc | grep -F ssh-agent
plugins=(git vi-mode docker ssh-agent)
zstyle ':omz:plugins:ssh-agent' identities $(ls ~/.ssh | grep ^id_ | grep -v "\.")
```

mopsfelder:
@robbyrussell @mcornella 

I just wanted to report that this pull broke ssh-agent plugin on Fedora 29.  Now, I need to run `ssh-add` every time I login because `~/.ssh/id_rsa` is not automatically added anymore.

`ssh-add -l` returns the following when there are no keys, in case it helps troubleshooting:

```
$ ssh-add -l    
The agent has no identities.
```

mopsfelder:
@robbyrussell @mcornella 

I verified that adding `zstyle :omz:plugins:ssh-agent identities id_rsa` to `~/.zshrc` before loading the `ssh-agent` plugin, it works again.

Perhaps this setting needs to be mandatory and explicitly documented as mandatory in order to use the ssh-agent plugin.

Or the ssh-agent plugin should handle the case where `identities` is empty.

mcornella:
I submitted #7520, see if that fixes it.

muffl0n:
Having the same issue on Arch Linux (OpenSSH_7.9p1). @mcornella thank for the PR, solves this problem for me!

timothystone:
> @robbyrussell @mcornella
> 
> I verified that adding `zstyle :omz:plugins:ssh-agent identities id_rsa` to `~/.zshrc` before loading the `ssh-agent` plugin, it works again.

Coming to this 3 years, 7 months, 14 days later, this worked for me. I was unaware this was happening and just thought it was my new `ed25519` key. Seems that was addressed, but I couldn't figure out why my default `id_rsa` key was loading despite not being in the plugin list. Ok, LaunchD. Thanks. 

It wasn't until I read this, that I was able to address the issue.

macOS Monterey  12.15.1
Terminal (macOS)
zsh 5.8.1 (x86_64-apple-darwin21.0)
OMZ master (6d48309c)

> 
> Perhaps this setting needs to be mandatory and explicitly documented as mandatory in order to use the ssh-agent plugin.
> 
> Or the ssh-agent plugin should handle the case where `identities` is empty.

PR #7176, opened 2018-09-30
Added iwhois readme

DavidMcKissick:
The plugins owner however seems to have discontinued the project.


mcornella:
I made a few changes and deleted the part about functionality being dead because it worked for me. Thanks for your contribution!

Issue #7175, opened 2018-09-30
[Good first PR] Add READMEs to undocumented plugins


mcornella:
_Register at [Hacktoberfest](https://hacktoberfest.digitalocean.com/) before submitting a Pull Request_

----

For some time we've had it in our minds to add a README to every plugin, as a way of documenting what the plugin does and what it requires, among other things. We have some READMEs already but most plugins lack one. That's why this issue exists, so you can take part in this effort.

You can have a look at [existing plugin READMEs](https://github.com/robbyrussell/oh-my-zsh/search?q=extension%3Amd+path%3Aplugins%2F) (the [yarn README is a great example](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/yarn/README.md)). Also you can find [a Markdown guide here](https://guides.github.com/features/mastering-markdown/).

The README should follow a very simple structure:

1. Name and brief description of what the plugin does. Links to relevant, essential websites.
   For example: the README for the `ant` plugin should link to the [ant project](https://ant.apache.org/).

2. How to enable the plugin. You can just copy-paste the excerpt below and substitute the word PLUGIN for the name of the plugin:<pre><code>To use it, add PLUGIN to the plugins array of your zshrc file:
\```
plugins=(... PLUGIN)
\```</code></pre>

3. Any other sections that document what the plugin does. Examples of that are: a l[ist of aliases, functions, or commands created](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/archlinux/README.md); a description of [settings that the plugin allows](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/ssh-agent/README.md#instructions); how to [build themes based on the plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/svn/README.md#usage-on-themes), ...

4. Optionally: a section with a list of contributors.

You'll need to read the `.plugin.zsh` file to know what the plugin does or, if there isn't one, the plugin will probably just add completion. You can also get what the plugin does by looking at the commits, or even the [plugins wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins).

If you want to work on the README for one of the following plugins, comment below and I'll mark it as such.

Thanks for your contribution and happy Hacktoberfest!

- [x] [ant](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/ant)
[...]

PR #7185, opened 2018-10-02
Adding Mix Plugin README

zwhitten:
This contributes to #7175


mcornella:
The completion list wasn't necessary but the rest is awesome. Thanks!

PR #7196, opened 2018-10-03
Add `neofetch` plugin

Ybalrid:
This is a super simple plugin that calls neofetch in any new terminal window you open.
Some people like the eye candy and the display of information in each prompt... ^^"


mcornella:
Sorry @Ybalrid but I don't see the point in having a plugin for just one line of code. You can just put that in your zshrc file, which is equivalent to enabling the plugin, and you don't have to know there's a plugin for OMZ. Cheers

Issue #7195, opened 2018-10-03
Jump: reusing a mark leads to unintended behavior


jonatanskogsfors:
# How to reproduce the problem
```
$ ~> mkdir monkey
$ ~> mkdir chimp
$ ~> cd monkey
$ ~/monkey> mark currentape
Mark /home/me/monkey as currentape? (y/n) y%
$ ~/monkey> cd ..
$ ~ jump> currentape
$ ~/monkey> cd ..
$ ~> cd chimp
$ ~/chimp> mark currentape
Mark /home/me/chimp as currentape? (y/n) y%
$ ~/chimp> cd ..
$ ~> jump currentape
$ ~/monkey> ls -la
total 0
drwxrwxrwx 0 me me 512 Oct  3 09:06 .
drwxr-xr-x 0 me me 512 Oct  3 09:07 ..
lrwxrwxrwx 1 me me  18 Oct  3 09:07 chimp -> /home/me/chimp
```
# What the correct behavior should be
Setting a mark when the name is already in use should either replace the old mark or refusing to mark. The user could either be informed or given a choice. No other side effects should occur.
# What the actual behavior is
Setting a mark when the name is already in use does not create a new mark but creates a symbolic link to the new folder in the old marked folder.

jonatanskogsfors:
That was quick! 🥇

jonatanskogsfors:
Found a problem. See #7205

PR #7197, opened 2018-10-03
Fix behavior when reusing a mark

ruduran:
Force the mark to point to the new dir, replacing the old one.

Fixes the unexpected behavior reported on #7195.


jonatanskogsfors:
Can I suggest another flag? #7205

ruduran:
> Can I suggest another flag? #7205

You are right, I didn't notice that :+1:

PR #7199, opened 2018-10-03
Update to Ubuntu Plugin README

zwhitten:
This contributes to #7175


plugins/ubuntu/readme.md
```diff
@@ -1,21 +1,50 @@
...
+| agu     | `sudo apt-get update`                                                  | Update package list                                                                               | 
+| agud    | `sudo apt-get update && sudo apt-get dist-upgrade`                     | Update packages list and perform a distribution upgrade                                           | 
+| agug    | `sudo apt-get upgrade`                                                 | Upgrade available packages                                                                        | 
+| agar    | `sudo apt-get autoremove`                                              | Remove automatically installed packages no longer needed                                          | 
+| aguu    | `sudo apt-get update && sudo apt-get upgrade`                          | Update packages list and upgrade available packages                                               | 
+| allpkgs | `dpkg --get-selections | grep -v deinstall`                            | Print all installed packages                                                                      |
```
mcornella: The `|` needs to be escaped with backslash (`\|`) or it doesn't render correctly.

plugins/ubuntu/readme.md
```diff
@@ -1,21 +1,50 @@
...
+| agud    | `sudo apt-get update && sudo apt-get dist-upgrade`                     | Update packages list and perform a distribution upgrade                                           | 
+| agug    | `sudo apt-get upgrade`                                                 | Upgrade available packages                                                                        | 
+| agar    | `sudo apt-get autoremove`                                              | Remove automatically installed packages no longer needed                                          | 
+| aguu    | `sudo apt-get update && sudo apt-get upgrade`                          | Update packages list and upgrade available packages                                               | 
+| allpkgs | `dpkg --get-selections | grep -v deinstall`                            | Print all installed packages                                                                      | 
+| kclean  | `sudo aptitude remove -P ?and(~i~nlinux-(ima|hea) ?not(~n`uname -r`))` |Remove ALL kernel images and headers EXCEPT the one in use                                         |
```
mcornella: Same here for tildes (`~`) and vertical bar (`|`). For the backticks, you can use `$()` instead: `$(uname -r)`. Markdown is confusing :sweat_smile:

plugins/ubuntu/readme.md
```diff
@@ -1,21 +1,50 @@
...
+This plugin adds completions and aliases for [Ubuntu](https://www.ubuntu.com/).
 
-acs = Apt-Cache Search  
-acp = Apt-Cache Policy
+To use it, add `ubuntu` to the plugins array in your zshrc file: 
+
+```zsh
+plugins=(... ubuntu)
+```
+### Note:
```
mcornella: This note is no longer valid since we started preferring apt. You can say instead that commands that use `$APT` will automatically choose apt if installed, or defer to apt-get otherwise. This requires changing some of the aliases' commands in the aliases table (just use `$APT` there, people will get it).

zwhitten:
hey @mcornella thanks for the feedback! I think I addressed the issues you mentioned. If there's anything else please let me know. Thanks again

mcornella:
Weirdly on the second version the tildes didn't need to be escaped anymore. Fixed and merged. Thanks!

PR #7205, opened 2018-10-03
More compatible flag for ln

jonatanskogsfors:
The flag '-h' isn't universal across implementation. According to FreeBSD man page for ln you can use 'n'.


mcornella:
True, `-h` didn't exist in Debian either. Thanks!

PR #7209, opened 2018-10-03
feat(agnoster): add remote-tracking status to prompt

fredden:
I find it useful to know when the local branch has diverged from the remote tracking branch. For example, I will often run `gfa` (ie, `git fetch --all --prune`) when starting work on a project. Having this information immediately available means that I no longer need to run `gst` (ie, `git status`). This is also useful if I'm switching branches and may forget to check if there are more commits in the remote branch than I have in my local. As a bonus, if I'm done working on something and have not yet pushed these changes to the remote, the icon is a nice gentle reminder.

This patch adds an icon to the prompt if the local branch is forward, behind, or has diverged from its remote tracking branch. When the branch has no remote branch associated, nothing is displayed.

screenshot_2018-10-03_20-43-22

I'm happy to take feedback on this. I've tried to pick icons that make sense, but am happy to change these and/or the position in the prompt as needed.

This is a modification to the @agnoster theme.


fredden:
Great. How do we proceed? I can see that you've approved this pull request, but I'm not sure what the next step should be. I don't have the power to merge this.

flohw:
If it's not too late and can give my opinion about the icons even if I don't use the theme...

I think there are not very readable at first glance. You may take inspiration from powerlevel9k.
* When there are some commits to pull there is a down circled arrow with the number of commits to pull.
* When there are some commits to push, there is a up circled arrow with the number of commits to push. 
* So when there are commits to push and pull (your dieverge case?) both icons are shown.

[here is an example](https://raw.githubusercontent.com/Falkor/dotfiles/master/screenshots/screenshot_falkor_iterm.png). I can understand that the number of commits to push/pull can be hard/time consumer. But I talk about the icon comprehension.

I found it more readable than upper left corner or upper right corner arrows.

powerlevel uses font awesome but I'm pretty sure it exists a similar unicode character :wink:

agnoster:
@flohw I think powerlevel9k tends towards "more verbose, more tricked-out" and my theme tries to be more on the minimal/subtle side (obviously this is a simplification, but you get the idea). So I actually like the solution @fredden has here for not cluttering the prompt unduly. I agree the glyphs aren't as intuitive as I'd like, but I think it's a suitable balance of signal to noise.

As to how to proceed - I think a maintainer for OMZ needs to actually merge this, I do not have that capability.

agnoster:
I think for me personally the simple up/down arrows don't convey as much meaning - without reading the docs or seeing a number, I don't know if an up arrow means "remote is ahead" or "we have commits that need to be pushed". The arrows @fredden chose indicate clearly a concept of "local is behind remote" and "local is ahead of remote" for me at least - using the fact that intuitively, down is "local", up is "remote", left is "behind" and right is "ahead". So an arrow pointing from up and to the left is showing "local is behind remote", an arrow pointing up and to the right is "local is ahead of remote".

The current iconography is more intuitive and clear to me than just an up/down arrow could ever be. I would support making the icons configurable, but strongly prefer @fredden's choices for the default. Of course there will be personal preference, and what's intuitive to one person will be obscure to another, and that's okay; there is no choice that will make everyone happy. We could try an A/B test and some user studies but that seems like overkill ;-)

fredden:
What can I do to progress this?

53jk1 (approved):

fredden:
@mcornella What can I do to get this merged in?

fredden:
@mcornella is there something I can do to help progress this?

fredden:
@mcornella please can you merge this in.

fredden:
@carlosala please can you merge this in.

carlosala:
After 4 years, we are here! 😊 
Sorry for the long time, and thanks for the contribution!

PR #7210, opened 2018-10-03
Fix coding style on the jump plugin

ruduran:
Use two spaces instead of one tab as said on the [coding style guide](https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide).


ruduran:
The conflicts should now be fixed. Not sure why the checks failed...

ruduran:
@mcornella any chance we can have this merged before more conflicts arise? 👼

mcornella:
Coding style is for now inconsistent throughout the repository. There are places where tabs are more suitable, so I don't really agree with the coding style wiki. Let's just leave it like that for now, and in the future we'll migrate to using editorconfig like the installer and uninstaller already use.

PR #7213, opened 2018-10-03
Added dircycle readme file

grikomsn:
This PR adds a readme file for the `dircycle` plugin, which includes detailed usage examples and an asciinema cast demonstrating directory traversals.


mcornella:
Great job! The dircycle plugin was hard to document and you did it brilliantly, thanks!

PR #7214, opened 2018-10-03
Create README.md for go plugin

MasterOfTheTiger:
This adds a new README.md for the go plugin, which explains to the user that the go plugin is a duplicate of the golang plugin and now symlinks there. 

For #7175


mcornella:
I just used a deprecation notice, we don't want users to use it since we'll remove it eventually. Thanks for the contribution!

PR #7215, opened 2018-10-04
issue: #7175 adding README for vagrant plugin

Bjorn248:
Didn't see any aliases in the file, is there anything else to add to the README?


mcornella:
Nope, that was enough. Thanks!

PR #7218, opened 2018-10-04
git: add an alias for git reflog

komuta:
The `git reflog` is pretty useful, especially with timestamp which are not present by default. So I thought about adding a simple alias for it.


adamchainz:
@carlosala I think this one can be closed, it’s old with few thumbs ups. I would say that using `git reflog` is exceptional enough that an alias probably isn’t required.

carlosala:
I agree. `git reflog` is used (as far as I saw) once in a while.
Thanks for the triaging :)

PR #7226, opened 2018-10-04
Added aliases to readme of docker-compose plugin

kmjayadeep:
Current readme file of docker-compose plugin doesn't have aliases included. It is added.
Contributes to #7175


mcornella:
Thanks for noticing this was needed, great job!

PR #7227, opened 2018-10-04
disable terraform prompt info in home directory

tescalada:
in the home directory there is a .terraform file for modules but it is most likely not an actual directory of terraform scripts. this disables checking for a workspace in the home directory which defaults to [default] if there is no workspace defined, which is unlikely in the home directory. It is possible others have a usecase where this would be desirable though, so maybe this patch should only be enabled with some disable_tf_prompt_in_home_dir type var.

### DIFF
--- a/plugins/terraform/terraform.plugin.zsh
+++ b/plugins/terraform/terraform.plugin.zsh
@@ -1,4 +1,6 @@
 function tf_prompt_info() {
+    # dont show 'default' workspace in home dir
+    [[ "$PWD" == ~ ]] && return
     # check if in terraform dir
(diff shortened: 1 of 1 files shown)


mcornella:
> in the home directory there is a .terraform file for modules 

Is it a file or a directory? Because `-d` only checks if it's a directory, in which case this patch wouldn't apply.

tescalada:
Sorry, yes it is a directory. It contains a plugins directory.

mcornella:
Ok I can see why that is annoying then. Approved :shipit:

PR #7228, opened 2018-10-04
Add sublime-merge plugin

amirmasoudabdol:
I've tried to put together a plugin similar to [`sublime`](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/sublime) but for [Sublime Merge](https://www.sublimemerge.com). I've used the `sublime` as the base and adjusted it with Sublime Merge. In the process, I've changed the `$*` to `"@*"` as I read that it's more robust. It shouldn't be very relevant in the case of Sublime Merge, but I thought why not?

P.S. I'm not a `zsh` expert! Please let me know if I'm missing something, or in any way I can improve this :-)

### DIFF
--- a/plugins/sublime-merge/README.md
+++ b/plugins/sublime-merge/README.md
@@ -1,4 +1,4 @@
-## sublime
+## sublime-merge
 
 Plugin for Sublime Merge, a cross platform text and code editor, available for Linux, Mac OS X, and Windows.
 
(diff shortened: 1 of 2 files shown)


amirmasoudabdol:
I haven't tested this on Windows or Linux, but I made sure that I'm at least picking the correct binaries based on the available `.tar.xz` common binary paths on Linux.

tdegeus:
+1. Would be really great to have!

PR #7229, opened 2018-10-04
Adds Docs for plugins (yum, mosh, themes)

aswathkk:
Issue: #7175 
Adds README.md file for following plugins:
- Yum Plugin
- Mosh Plugin
- Themes Plugin


mcornella:
Thanks!

PR #7232, opened 2018-10-05
Added readme for catimg plugin

grikomsn:
I have created a readme for the `catimg` plugin, which contains how to enable the plugin and an asciinema usage example.


mcornella:
Thanks!

PR #7233, opened 2018-10-05
Themes plugin: custom themes directory fix

aswathkk:
The custom theme directory is `$ZSH_CUSTOM/themes/`
@kevinxucs @r-darwish


mcornella:
Well spotted!

PR #7236, opened 2018-10-05
Adds custom themes directory support to `lstheme`

aswathkk:
`lstheme` command used to list only the themes listed in `$ZSH/themes/` directory. This commit adds themes in `$ZSH_CUSTOM/themes/` also to the theme listings.

### DIFF
--- a/plugins/themes/themes.plugin.zsh
+++ b/plugins/themes/themes.plugin.zsh
@@ -19,6 +19,5 @@ function theme
 
 function lstheme
 {
-    cd $ZSH/themes
-    ls *zsh-theme | sed 's,\.zsh-theme$,,'
+    (ls $ZSH/themes/ && ls $ZSH_CUSTOM/themes/) | sed 's,\.zsh-theme$,,'
 }


plugins/themes/themes.plugin.zsh
```diff
@@ -19,6 +19,5 @@ function theme
 
 function lstheme
 {
-    cd $ZSH/themes
-    ls *zsh-theme | sed 's,\.zsh-theme$,,'
+    (ls $ZSH/themes/ && ls $ZSH_CUSTOM/themes/) | sed 's,\.zsh-theme$,,'
```
mcornella: You can ls multiple directories in the same command: `ls dir1 dir2 | sed ...`

aswathkk: @mcornella
It will create some unwanted outputs (which indicates the directory for which the listing is made) which has to be removed using some other methods.
Here's an example:
```shell
>ls $ZSH/themes/ $ZSH_CUSTOM/themes/ | sed 's,\.zsh-theme$,,'
/Users/aswathkk/.oh-my-zsh/custom/themes/:    # This line is not required in the output
example

/Users/aswathkk/.oh-my-zsh/themes/:   # This line is not required in the output
3den
Soliah
.
.
.
```

What's the issue with `(ls $ZSH/themes/ && ls $ZSH_CUSTOM/themes/)` ? Is it not performant? If it is so, need to write some code to remove the unwanted output.

PR #7243, opened 2018-10-06
Update swift completion script

Zialus:
Script generated by swift 4.2

Fixes bug described [here](https://bugs.swift.org/browse/SR-7053)


mcornella:
Thank you!

PR #7252, opened 2018-10-08
Add readme file for nanoc plugin #7175

soul-wish:
Contributes to #7175


mcornella:
We'll have to do something about those deprecated aliases though. @josh- do you have any ideas?

PR #7258, opened 2018-10-09
Add README for Kubectl in #7175

kaltepeter:
This adds a readme file for the kubectl plugin as requested in #7175.


mcornella:
Thank you!

PR #7264, opened 2018-10-10
Detect aws cli completion file from RPM (#6424).

### LINKED ISSUE #6424: aws plugin does not work for rpm installations
Hi,

the aws plugin doesn't work very well with systems that installed awscli from an RPM.  It is very focused on brew.  On my Fedora machine it fails to activate any completion.

Kind regards,
 Frank.

FrankGruellich:
Hi,

I'd appreciate if someone could have a look and merge this.  It tries to address #6424.  Thanks.

Kind regards,
 Frank.

Fixes #6424

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -21,6 +21,10 @@ _awscli-homebrew-installed() {
   [ -r $_brew_prefix/libexec/bin/aws_zsh_completer.sh ] &> /dev/null
 }
 
+_awscli-rpm-installed() {
+  which rpm &>/dev/null && rpm -q --quiet awscli
+}
+
 export AWS_HOME=~/.aws
 
 function agp {
@@ -45,7 +49,11 @@ if which aws_zsh_completer.sh &>/dev/null; then
   _aws_zsh_completer_path=$(which aws_zsh_completer.sh 2>/dev/null)
 elif _homebrew-installed && _awscli-homebrew-installed; then
   _aws_zsh_completer_path=$_brew_prefix/libexec/bin/aws_zsh_completer.sh
+elif _awscli-rpm-installed; then
+  _aws_zsh_completer_path="/usr/share/zsh/site-functions/aws_zsh_completer.sh"
+else
+  _aws_zsh_completer_path=$(which aws_zsh_completer.sh)
 fi
 
-[ -n "$_aws_zsh_completer_path" ] && [ -x $_aws_zsh_completer_path ] && source $_aws_zsh_completer_path
+[ -n "$_aws_zsh_completer_path" ] && [ -r $_aws_zsh_completer_path ] && source $_aws_zsh_completer_path
 unset _aws_zsh_completer_path


mcornella:
How fast is the call to rpm?

FrankGruellich:
> How fast is the call to rpm?

The call to `rpm -q --quiet awscli` takes approximately:

    rpm -q --quiet awscli  0.01s user 0.00s system 93% cpu 0.012 total

on my system.

mcornella:
I know it's not a lot but just check if `/usr/share/zsh/site-functions/aws_zsh_completer.sh` exists which is . The additional else isn't needed also, we're already using `which aws_zsh_completer.sh` on the first if condition. And you'll have to rebase as well to fix the conflicts.

FrankGruellich:
Hi,

thanks for your comment.

> I know it's not a lot but just check if `/usr/share/zsh/site-functions/aws_zsh_completer.sh` exists

I just tried to follow the exisiting pattern of querying a list of package management systems (eg. `brew`) for the location of the file.  If I follow your argument the whole plugin has to be reworked to just check a list of files for existence and break out as soon as it finds the first one.  Eg. something like:

    for file in /usr/local/opt/awscli/libexec/bin/aws_zsh_completer.sh \
      /usr/share/zsh/site-functions/aws_zsh_completer.sh \
      `which aws_zsh_completer.sh || echo /dev/null` \
      /dev/null
    do
      if [ -r "${file}" ]
      then
        source "${file}"
        break
      fi
    done

I did not want to step that much onto other peoples toes in one PR, though.

> The additional else isn't needed also, we're already using `which aws_zsh_completer.sh` on the first if condition.

Ah, I see.  Yes, that made it over from e273cf0 where my first PR was based on.  Sorry for that.

FWIW, I consider the detection via `which` also a bit broken.  First of all it only detects files marked as executable and there is no requirement for a file to be executable to be`source`d, readabliity of the file is sufficient.  On the topic of performance, iterating over a long `$PATH` with `which` to find an executable file and doing so even two times can be also quite time consuming.

> And you'll have to rebase as well to fix the conflicts.

No problem, I'm willing to do this once the general approach is clarified.

Kind regards,
 Frank.

mcornella:
I think the difference is that with Homebrew the prefix may be in another path, so we check using Homebrew if that's the case. Note that it first looks up in a default location, and only if it's not found it does a call to `brew`.

In rpm that's not an issue, since the location of a file is fixed and will always be the same for a package. So we can just directly probe that location without passing throw the rpm check. 

> FWIW, I consider the detection via `which` also a bit broken. First of all it only detects files marked as executable and there is no requirement for a file to be executable to be`source`d, readabliity of the file is sufficient. On the topic of performance, iterating over a long `$PATH` with `which` to find an executable file and doing so even two times can be also quite time consuming.

The new version only checks for readability, as you rightly point out. The PATH check is almost instantaneous, because zsh keeps a hashtable of where each command is in the filesystem (that's why you sometimes have to run `rehash` after a command has been installed).

So to recap, just probe that location (add a comment that it belongs to an RPM package) and we're good to go. Thanks!

PR #7265, opened 2018-10-10
Also set the `grep` options for `egrep` and `fgrep`

pvdb:
Both BSD and GNU bundle together `grep`, `egrep` and `fgrep`, so it make sense IMO to export the `grep` options for `egrep` and `fgrep` as well.

Note that BSD also bundles `zgrep`, `zegrep`, and `zfgrep`, so we could conditionnally export the same options for the compression-compatible utilities as well, but that's probably a separate PR.


pvdb:
Thanks, @mcornella ... I :heart: the changes in #281! :thumbsup:

PR #7267, opened 2018-10-11
Add README for textmate plugin

BrianMitchL:
Per #7175


BrianMitchL:
Good point, as a single function, a table of aliases doesn't really make sense 🙂

mcornella:
Good job :+1:

PR #7270, opened 2018-10-11
error handling if PackageKit-command-not-found is not installed for Fedora

jeff-chandler:
I was trying to work on a README for this plugin, and discovered that it didn't really do anything with Fedora unless you had the PackageKit-command-not-found package installed.  This just adds some error handling which prompts the user to install PackageKit-command-not-found if it isn't installed.


mcornella:
I'd rather have that requirement documented in the README than having this checked on runtime, unless we can have a check that is only run for Fedora.

The way you've coded it:

1. The command-not-found handler function will be defined in all platforms, not only Fedora, possibly getting in the way of the rest of the command-not-found handlers defined in the plugin.

2. It will spit out the "`PackageKit-command-not-found` not found" every time a command is not found, since zsh runs the function `command_not_found_handler` that way.

Also, the rest of the supported platforms should be amended in a similar fashion to look if the appropriate package is installed or not, leading to added complexity to check what platform we're in.

That's why I'd rather document this in the README; the [Supported platforms](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/command-not-found/README.md#supported-platforms) section is a good place to put it. Right now there is just a link to the relevant packages but it could be reworded to make it more explicit that there is a requirement for each of those platforms.

jeff-chandler:
Thank you for the clarification.  I hadn't thought about the `command_not_found_handler `being run every time on other platforms.  My only issue is that the plugin has no additional functionality in Fedora.  The output is the same with or without the plugin activated.

mcornella:
Then we could add the message at the end of the script saying no command-not-found packages were found, and early return if one has been found. That change I would merge.

PR #7287, opened 2018-10-13
git: add alias for git diff untracked files

repodevs:
Add alias `gdu='git diff /dev/null'`

```
$ gdu common/utils.py
```
```
diff --git a/common/utils.py b/common/utils.py
new file mode 100644
index 0000000..d26a7c8
--- /dev/null
+++ b/common/utils.py
@@ -0,0 +1,5 @@
+# -*- coding: UTF-8 -*-
+import logging
+
+LOGGER = logging.getLogger('myapp')
```


mcornella:
You're just effectively `cat`-ing the file, only in a diff format.

PR #7289, opened 2018-10-14
Create install-zsh-ubuntu.sh

fightercode:
Added in Install and set ZSH to default the Shell in your Terminal
This the bellow in result my Terminal
image


bariscodeid:
image

Link: https://askubuntu.com/questions/131823/cant-make-zsh-the-default-shell

mcornella:
This is documented [in the wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Installing-ZSH), thanks.

PR #7293, opened 2018-10-15
Added Tmux readme

Byte7:
Added Tmux readme #7175


mcornella:
Good job on a hard one. Thanks!

PR #7300, opened 2018-10-16
Add --latest flag for yarn upgrade-interactive

bondz:
Adds autocomplete for `--latest` when doing `yarn upgrade-interactive`


mcornella:
We're using the [zsh-completions](https://github.com/zsh-users/zsh-completions/blob/master/src/_yarn) distribution of yarn completion, can you submit the PR there instead?

bondz:
Will do @mcornella thanks.

PR #7304, opened 2018-10-17
[subl] Fix sublime plugin for cygwin. Typo in path variable was introduced by cea941c

mhennecke:
Fixes broken sublime plugin for cygwin environment. Typo in variable name was introduced in cea941ce42b5d550489367608dd1ac4401151c97.


mcornella:
Well spotted thanks!

PR #7305, opened 2018-10-17
performance improvements for git commands

GrzesiekP:
Improved performance using HEAD in git commands. Without HEAD very slow for bigger git repos.


mcornella:
Many stuff has changed since then, revise if the changes still apply and submit a new pull request. BTW, `git status` doesn't expect a commit, when you pass it `HEAD` you're saying to look for changes in the HEAD path (filer or folder), which is wrong.

PR #7307, opened 2018-10-17
make aws plugin load profile names from ~/.aws/credentials

helloworlddan:
NOTE: It is more feasible to load profile names from .aws/credentials
rather than loading it from .aws/config. the 'credentials' file is a lot
more frequently used when using third party tools to retrieve and store
temporary STS credentials. Therefore, profile names in .aws/credentials
are more likely to be backed with actual 'hot' credentials.

Signed-off-by: Daniel Stamer <daniel.stamer@cloudreach.com>

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -37,7 +37,7 @@ function asp {
 }
 
 function aws_profiles {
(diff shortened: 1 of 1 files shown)


mcornella:
It can be easily modified to read from both files, can you do it?

mcornella:
ping @helloworlddan

PR #3471, opened 2015-01-05
Add completion for invoke https://github.com/pyinvoke/invoke

IvanMalison:

PR #7311, opened 2018-10-19
feat(plugins): add completion plugin for `invoke`

kenden:
This gives completion for the `inv` / `invoke` commands.
https://github.com/pyinvoke/invoke

There already is a PR for invoke (https://github.com/robbyrussell/oh-my-zsh/pull/3471) but it is outdated. Invoke since added a command to generate a completion script.
This PR  simply uses this command.
http://docs.pyinvoke.org/en/latest/invoke.html#tab-completion

Note: [pyinvoke](https://pypi.org/project/pyinvoke/) is a python package that has nothing to do with invoke, but the website for invoke is http://www.pyinvoke.org/.

Closes #3471

### DIFF
--- a/plugins/invoke/README.md
+++ b/plugins/invoke/README.md
@@ -1,0 +1,10 @@
+# Invoke plugin
+
+This plugin adds completion for [invoke](https://github.com/pyinvoke/invoke).
+
+To use it, add `invoke` to the plugins array in your `~/.zshrc` file:
+
+```zsh
+plugins=(... invoke)
+```
+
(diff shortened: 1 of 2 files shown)


kenden:
@mcornella this change is not only documentation, it adds autocompletion for https://www.pyinvoke.org/.
I've been using the code from this pull request since it was created in February 2018.
I'm not sure how to get it merged / looked at, would you know?

53jk1 (approved):

mcornella:
Thanks!

PR #7316, opened 2018-10-20
Added README.md to lighthouse plugin

pasanjayawickrama:


mcornella:
Since this plugin only adds one function, it would be better to document its usage instead of having an alias table. See below for usage instructions: https://github.com/robbyrussell/oh-my-zsh/blob/576ada138fc5eed3f58a4aff8141e483310c90fb/plugins/lighthouse/lighthouse.plugin.zsh#L1-L4

PR #7317, opened 2018-10-20
Change external ip address provider to ifconfig.co

ottnorml:


mcornella:
For the time being icanhazip.com works. If the site goes down or we get a notification to use another site ifconfig.co will be the first one that we check.

PR #7319, opened 2018-10-21
Add alias to Django plugin for manage.py

kronion:
Add `pm`, a simple alias for `python manage.py`.


dimensi0n:
This is useful

mcornella:
Hey there, I've deprecated the plugin because the completion it provides is lacking in features and Zsh already provides completion for it out of the box, so I'm closing this because it doesn't make sense to have a plugin just with one alias.

You can put this alias in your zshrc file, if necessary. Thanks for your contribution.

PR #7321, opened 2018-10-22
fix(archlinux-plugin): Fix yay alias masking

### LINKED ISSUE #7320: [Archlinux plugin] name masking
For example [`yaourt -Syua`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/archlinux/archlinux.plugin.zsh#L32) and [`yay -Syu`](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/archlinux/archlinux.plugin.zsh#L61) have the same alias `yaupg`.

**Possible fix**: prefix all `yay` alias by `yay` instead of `ya`.

Heziode:
Fix #7320

### DIFF
--- a/plugins/archlinux/README.md
+++ b/plugins/archlinux/README.md
@@ -4,8 +4,8 @@
 
 #### YAY
 
-| Alias   | Command                            | Description                                                         |
-|---------|------------------------------------|---------------------------------------------------------------------|
+| Alias    | Command                            | Description                                                         |
+|----------|------------------------------------|---------------------------------------------------------------------|
 | yayconf  | yay -Pg                            | Print current configuration                                         |
 | yayin    | yay -S                             | Install packages from the repositories                              |
 | yayins   | yay -U                             | Install a package from a local file                                 |
--- a/plugins/archlinux/README.md
+++ b/plugins/archlinux/README.md
@@ -6,21 +6,21 @@
 
 | Alias   | Command                            | Description                                                         |
 |---------|------------------------------------|---------------------------------------------------------------------|
-| yaconf  | yay -Pg                            | Print current configuration                                         |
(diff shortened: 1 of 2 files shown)


mcornella:
Can you change the README as well?

Heziode:
Damn I forgot to update this.

Heziode:
@mcornella it is done.

mcornella:
Wait, I think the naming clash was intentional, from #6867:

> I alias `yay` with the same commands as `yaourt`, I think most people will not use them together.

> Definitely `yay` should be above outdated `pacaur` and `yaourt`.

I'm not familiar with Arch Linux but my guess is that if yay is installed it will be to use it instead of yaourt (which, according to the [Arch Linux wiki](https://wiki.archlinux.org/index.php/AUR_helpers#Pacman_wrappers), has become stalled).

_On the one hand_, an alias change allows users who have both yay and yaourt to use whichever they prefer, instead of having to use the one we chose for them.

_On the other_, alias changes means users who update need to first notice the alias don't work anymore and then look at the plugin README to learn the new aliases. In this case, if they have yaourt installed they will run the yaourt commands and possibly get the system in an unusable state.

The alias clashing is a problem I made the mistake of not thinking about before merging #6867, but it will be doubly worse if I merge this PR. To sum up: "if yay is installed, use yay" was ok because yay was newer than yaourt. If we now merge this after yay users are used to the old yaourt aliases, they will end up running the old yaourt commands without their knowledge.

That's why I'm hesitant to merge this. Right now, the only solution if you want to use yaourt aliases is to uninstall yay. Thanks for your contribution.

@specter119 @mati865 @juliuswz thoughts?

Heziode:
Indeed, it is a valid statement that I had not seen.

PR #7323, opened 2018-10-22
Add plugin iterm2-touch-bar-status

jefferickson:
This plugin allows for custom text to be set on the Macbook Pro Touch Bar when using iTerm2. Defaults to the current working directory.


liammccartney:
👍

narukami0324 (approved):

PR #7324, opened 2018-10-22
Create a better please command

nelsonmestevao:
Previously, the please command was just a replacement of `sudo`. Without any new improvement. In fact, it was worst (sudo == 4 letters, please == 6 letters).

This change will add the ability to repeat the previous command with `sudo` if it was needed and you somehow forgot.


mcornella:
See https://github.com/robbyrussell/oh-my-zsh/issues/7527

PR #7326, opened 2018-10-22
mvn-or-mvnw: Changing check from -f to -x

paul-nelson-baker:
The problem that can occur is ocassionally mvnw will not be executable. This can happen if mvnw is included from an archetype, as unix permissions aren't preserved within the jar they're stored in. Only using mvnw if it exists AND is executable


mcornella:
👍

PR #7327, opened 2018-10-23
Don't add nonexistant paths to fpath

kronion:
Oh-my-zsh adds the following two paths to `fpath`, but neither currently exists in the project:
`$ZSH/functions`
`$ZSH/completions`


mcornella:
May use it in a future version.

PR #7333, opened 2018-10-24
 Allow user to hide "Random theme ... loaded"

juniorRubyist:
I'm not so interested in the message that tells me that a random theme was loaded, since I know that the theme is randomly chosen from my picks, so I wanted to be able to hide that message. I added an option that can be set in `.zshrc` to hide the message by uncommenting a line.

As can be seen in `templates/zshrc.zsh-template` at lines 44-45, if `ZSH_THEME_RANDOM_MESSAGE` is set to `false` by uncommenting the line, then the message will be disabled.


mcornella:
I won't add yet another setting variable just to disable a one-line message, sorry.

PR #7339, opened 2018-10-25
Add Alternative Sublime Text Exec Location

ssankauskas:
Added path to Sublime Text executable that was installed via Snap package manager.


mcornella:
Doesn't snap add a `subl` symlink somewhere? Run `which -a subl` to check that out.

ssankauskas:
Yes it does add symlink to `/snap/bin/subl`.

Thanks!

mcornella:
No problem.

PR #7353, opened 2018-10-27
Add poetry plugin, and encapsulate compinit logic

kronion:
This PR does two things:
1. Add a basic plugin to manage tab completions for the Python `poetry` package manager.
2. Moves the `compinit` logic in Oh My Zsh's entry point into its own lib function, so other developers can easily respect the user's security preferences if they ever need to run `compinit` again.

### DIFF
--- a/lib/compfix.zsh
+++ b/lib/compfix.zsh
@@ -42,3 +42,19 @@ function handle_completion_insecurities() {
 
 EOD
 }
+
+
+# Load zsh completions according to the user's security preferences.
+update_completions() {
+  if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
+    # If completion insecurities exist, warn the user
+    if ! compaudit &>/dev/null; then
+      handle_completion_insecurities
+    fi
+    # Load only from secure directories
+    compinit -i -d "${ZSH_COMPDUMP}"
+  else
+    # If the user wants it, load from all found directories
+    compinit -u -d "${ZSH_COMPDUMP}"
+  fi
+}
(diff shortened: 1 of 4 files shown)


mcornella:
The way to do this now is:
1. Create dynamic completion files in `$ZSH_CACHE_DIR/completions`
2. `autoload` the completion file
3. `_comps[<command>]=<completion-function>`

[See example](https://github.com/ohmyzsh/ohmyzsh/blob/eb2eacaf0c46ace8e0eb3543199851a8249338fc/plugins/fnm/fnm.plugin.zsh).

Reopen if there's still interest for this.

kronion:
Hi, I'm still interested in contributing a [poetry](https://python-poetry.org/) plugin. If I'm understanding correctly, there are a few changes you'd want to see:
1. Move the completion and version files into `$ZSH_CACHE_DIR` to be consistent with other plugins.
2. `autoload` the completion function on first use rather than rerunning `compinit`, and wait for `compinit` to run the next time a new shell session is started. Similar to this example: https://github.com/ohmyzsh/ohmyzsh/blob/c6e7f8905fb61b927f12f43fb57f8c514cd48a67/plugins/rust/rust.plugin.zsh#L20 

Would you reopen for that?

mcornella:
`poetry` seems to be widely used, I'd merge a plugin like that yes. Thanks!

mcornella:
Maybe it could be merged with #10028.

kronion:
I've updated my branch. I can make a new PR against `ohmyzsh:master` (don't see a way to reopen and update this PR), but let me know if you'd prefer something else. Not sure how to go about merging into #10028.

mcornella:
You should open a new PR, yes. Let me make the necessary changes then to merge it with #10028.

PR #7361, opened 2018-10-28
Add readme for git-hubflow

mzuehlke:
This adds a readme file for the git-hubflow plugin as requested in #7175


mcornella:
Thanks!

PR #7364, opened 2018-10-28
aws: refactor completion sourcing logic

### LINKED ISSUE #6424: aws plugin does not work for rpm installations

mcornella:
Clean up Homebrew detection and add comments.
Also changed some if flags.

Incorporated the main changes of #7264 

Closes #7264 
Fixes #6424


dimensi0n:
I can test it

PR #7368, opened 2018-10-29
Add auto complete on ng update -Angular 6+

lakinduakash:


mcornella:
Resubmit with the changes.

PR #7372, opened 2018-10-30
transfer: Add deprecation notice (#7371)

### LINKED ISSUE #7371: Deprecate transfer plugin
_Register at [Hacktoberfest](https://hacktoberfest.digitalocean.com/) before submitting a Pull Request_

----

Since [transfer.sh](https://transfer.sh) is shutting down on November 30th we should deprecate the plugin that uses it. The reasoning is that we should encourage users to use a different service, since their files will presumably be gone after it shuts down.

To do this, we should both:

- Add a deprecation warning in the README (explaining the reason)

- Add a deprecation warning in the `.plugin.zsh` file to run when the plugin is loaded. Extra points for printing the warning with a warning-like color, like yellow or orange.

After November 30th we should give a week or so before completely removing the plugin.

a7r3:
Signed-off-by: Arvindraj <arvindultimate7352@gmail.com>

----
Fixes #7371

Co-authored-by: Stephen Ward <stephen@whilenotzero.com>

### DIFF
--- a/plugins/transfer/README.md
+++ b/plugins/transfer/README.md
@@ -1,5 +1,11 @@
 # `transfer` plugin
 
+**NOTICE: The `transfer` plugin will be deprecated soon, since the [transfer.sh](https://transfer.sh) service will be shutdown on 30th November 2018.**
+
+**Please move your files to alternative file sharing service providers**
+
+<hr>
+
 [`transfer.sh`](https://transfer.sh) is an easy to use file sharing service from the command line
 
 ## Usage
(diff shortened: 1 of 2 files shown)


mcornella:
Hey thanks for your contribution! I've reworded a couple of things and will merge this on November 1st to give a full month of notice. Cheers!

mazhigali:
@a7r3  transver will live on!
https://i.imgur.com/mkhMHDz.png

mcornella:
Thanks @mazhigali, reverted.

PR #7374, opened 2018-10-30
Support themes from ZSH_CUSTOM dir in 'random' mode

4O4:
Currently if you specify themes from ZSH_CUSTOM dir in `ZSH_THEME_RANDOM_CANDIDATES`, they will never be picked up because only `$ZSH/themes` dir is being scanned. 

This PR introduces minor refactoring (extracted theme loading logic to new `load_theme()` function) and adds scanning of both internal and custom theme dirs in 'random' mode.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -95,25 +95,39 @@ done
 unset config_file
 
 # Load the theme
+load_theme() {
+  if [ ! "$1" = ""  ]; then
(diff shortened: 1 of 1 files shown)


4O4:
Hi, is there any chance for review / merge? :)

mcornella:
Fixed recently. Thanks for your contribution!

PR #7379, opened 2018-10-31
ADD: lock for starting ssh agent

perfectworks:
ssh-agent plugin will create multiple ssh-agent processes when
concurrent starting multiple shell. This patch add a lock when zsh
plugin try to start a ssh-agent.


perfectworks:
Someone here?

PR #7382, opened 2018-11-01
Added minimum length and ellipsis symbol options for the shrink-path plugin.

krivit:
This PR adds two additional arguments to `shrink_path`, one to specify the (unconditional) minimum length of each truncated path element and the other to specify the symbol indicating truncation if one occurred. For example, here's the old behaviour:
```
$ shrink_path -t  -l ~/.oh-my-zsh/plugins/shrink-path
~/.o/p/shrink-path
```
Here's the new behaviour, with indication of whether an element was truncated:
```
$ shrink_path -t -3 -e '$' -l ~/.oh-my-zsh/plugins/shrink-path
~/.oh$/plu$/shrink-path

$ shrink_path -t -6 -e '$' -l ~/.oh-my-zsh/plugins/shrink-path
~/.oh-my$/plugins/shrink-path
```

This makes the path much more readable, while still saving a lot of space.

### DIFF
--- a/plugins/shrink-path/README.md
+++ b/plugins/shrink-path/README.md
@@ -20,6 +20,8 @@ here are the results of calling `shrink_path <option> /home/me/foo/bar/quux`:
     -s|--short    /h/m/f/b/q
     -t|--tilde    ~/f/ba/q
     -f|--fish     ~/f/b/quux
+    -3            /hom/me/foo/bar/quu
+    -e '$' -3     /ho$/me/foo/bar/qu$
 ```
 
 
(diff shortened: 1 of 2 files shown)


gcox:
@krivit Would be nice to be able to specify truncation location: `start|middle|end`

krivit:
@gcox, the plugin currently truncates the most characters it can while maintaining uniqueness. I am not sure how that would be done under the other regimes. In any case, it seems to be a separate feature from what this PR is trying to accomplish.

gcox:
@krivit I could be misunderstanding, but it seems related. For clarity, here are examples of what I'm suggesting:

```
// Truncate start

$ shrink_path -t -6 --tl start -e '...' -l ~/.short/a longer path component/shrink-path
~/.short/...ponent/...k-path

// Truncate middle

$ shrink_path -t -6 --tl middle -e '...' -l ~/.short/a longer path component/shrink-path
~/.short/a l...ent/shri...ath

// Truncate end (what it currently does)

$ shrink_path -t -6 --tl end -e '...' -l ~/.short/a longer path component/shrink-path
~/.short/a long.../shrink...
```

krivit:
@gcox, it's related only in that it pertains to the same plug-in. However, it's orthogonal to being able to set the maximal amount of truncation and to the choice of the truncation symbol, which is what this PR accomplishes.

mcornella:
I'm going to merge this if you're still interested; a couple of things before though:
1. There is mixed indentation: the original uses spaces so use that instead of tabs. 
2. Document the changes to the README.
Thanks!

krivit:
I've made the changes, as well as handling "ellipsis" characters of length > 1 and handling the non-ellipsis case more consistently.

PR #7385, opened 2018-11-01
Add keychain plugin

jbenden:
This plugin integrates the Keychain tool[1] in to the project.

[1] [keychain](https://www.funtoo.org/Keychain)


robbyrussell:
Thanks for this, @jbenden! I hadn't seen Keychain before...looks like a useful tool

PR #7388, opened 2018-11-04
Update to AWS Plugin - AWS_EB_PROFILE

llbbl:
Added AWS_EB_PROFILE environment variable to the AWS Plugin. The EB CLI uses this variable instead of the primary AWS_PROFILE variable to keep track of what profile it is using. 

I updated the formatting of the README for the AWS Plugin so that it is more in line with the formatting of the other plugins. I added AWS_EB_PROFILE to the README. 

https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-configuration.html


llbbl:
@cristim I think you are the primary maintainer on this ? 🥇

mcornella:
Thanks!

PR #7394, opened 2018-11-07
Add Qwant Lite - Web Search plugin

Zar-rok:
Add the [Qwant Lite](https://lite.qwant.com/) search engine to the web_search plugin.


mcornella:
I don't see the need, the regular qwant looks pretty fast to me.

PR #7410, opened 2018-11-16
Add alias for get latest commit in git repository

draveness:


mcornella:
This is not a useful one. You can just `gsh` for `git show` and it prints much more information about the commit, not just the sha.

PR #7417, opened 2018-11-18
Add aliases for zypper

MasterOfTheTiger:
This adds a simple zypper plugin.


mcornella:
We already have the [suse plugin](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/suse).

PR #7418, opened 2018-11-19
nvm plugin improvements

hanandjun:
After installing nvm with brew, in order for the nvm plugin to work properly.
Although nvm's readme mentions: Homebrew installation is not supported.

I read CONTRIBUTING GUIDELINES, but I may not be doing very well, please point out my mistake, thank you😄

Beginners from China


hanandjun:
Please check

PR #7421, opened 2018-11-20
just command runner oh-my-zsh plugin

linux-china:
Just is a popular command runner to supply a handy way to save and run project-specific commands.
With oh-my-zsh integration, and it's easy to implement code completion for recipes. https://github.com/casey/just

### DIFF
--- a/plugins/just/README.md
+++ b/plugins/just/README.md
@@ -1,0 +1,33 @@
+Just command runner oh-my-zsh plugin
+======================================
+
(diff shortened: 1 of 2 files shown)


steveeJ:
Bump. I'm interested in this.

kenden:
`just --completions` was added in just 0.5.5.
Maybe it could be better to make use of it?
Like for example the helm plugin does:

PR #7425, opened 2018-11-20
change plugin 'sublime' alias 'st' to 'sub'

yuanxin7835:
alias 'st' is not the best for plugin 'sublime'
Whether it's understanding or keyboard manipulation,'sub' is best


mcornella:
You can just redefine it on your own.

PR #7427, opened 2018-11-21
Add alias for brew services list

DmitryOlkhovoi:
I used it a lot, would be nice to have it here


plugins/brew/brew.plugin.zsh
```diff
@@ -4,6 +4,7 @@ alias brewsp='brew list --pinned'
 alias bubo='brew update && brew outdated'
 alias bubc='brew upgrade && brew cleanup'
 alias bubu='bubo && bubc'
+alias brewsl='brew services list'
```
katrinleinweber: How about `bsl` to match the schema? It's [not taken by any other plugin](https://github.com/robbyrussell/oh-my-zsh/search?q=bsl&unscoped_q=bsl) so far :-)

PR #7435, opened 2018-11-24
[OSX] Make iTunes volume adjustment easier and fix Spotify output.

johnsyweb:
#### Context

Setting the volume when you don't know the current volume is hard. It may be expected that not providing a volume would return the current volume (this is how the `playlist` subcommand works). However...

``` zsh
% itunes vol
48:48: syntax error: Expected expression but found end of script. (-2741)
```

#### Change

If no argument is provided to the `vol` subcommand the current volume is displayed.

``` zsh
% itunes vol
38
```

Additionally, added `up` and `down` arguments:

``` zsh
% itunes vol 
Current volume is 48.

% itunes vol up

% itunes vol   
Current volume is 58.

% itunes vol down

% itunes vol     
Current volume is 48.
```


johnsyweb:
plugins/osx/spotify has drifted since I opened this PR, so closing and will open a dedicate PR for iTunes.

PR #7437, opened 2018-11-25
Add alias gcsam, commit everything with message and GPG signature, add nerd-terminal

thbe:
Add an alias to commit everything signed with your assigned GPG key


mcornella:
gpg sign should be a default in `.gitconfig` / global git config.

thbe:
First of all, after 2 years, really?

Secondly, why do you have the switch in gcssm? This doesn't really support the point you've made. Moving this to gcsasm instead of gcasm would make sense ... or removing the gcs* aliases in principle ...

PR #7440, opened 2018-11-26
[kubectl plugin] add namespaces and endpoints management

Shini31:


Shini31:
Not synchronized repo

PR #7446, opened 2018-11-28
[minikube plugin] cache command completions

fokinpv:
Add caching of completions for **minikube** plugin in the same manner as for **kubectl**.


calvinbui:
works for me, greatly speeding up loading times

PR #7448, opened 2018-11-28
z: fix argument handling

blastrock:
If the argument is -t, [ -t ] will test if stdout is a TTY. We should
use -z to test if a string is empty instead.

### DIFF
--- a/plugins/z/z.sh
+++ b/plugins/z/z.sh
@@ -110,9 +110,9 @@ _z() {
 
     else
         # list/go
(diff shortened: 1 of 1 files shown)


mcornella:
Can you submit this upstream? https://github.com/rupa/z

blastrock:
Yes sorry, I opened a PR there https://github.com/rupa/z/pull/254

blastrock:
Upstream seems unmaintained, what should be done in this case?

PR #7455, opened 2018-12-01
Add new vagrant commands

andrewimeson:
I added three commands to the completion that were not there before. I took the wording from the output of `vagrant --help` on Vagrant 2.2.0

Tagging @robbyrussell , since you are listed as the maintainer for the Vagrant plugin


robbyrussell:
Thanks, @andrewimeson!

PR #7456, opened 2018-12-03
Provide even spacing between marks

ahrex:
Before, when typing the `marks` command, longer mark keys would cause
the tabs to spill over to the next tab stop, like so:

    rc      -> /home/ahlex/.rc
    repos   -> /home/ahlex/repos
    a-longer-string         -> /tmp

Implement better key display by running through all of the marks twice,
once to get the longest key length, and the second time to format
everything according to that length:

                 rc -> /home/ahlex/.rc
              repos -> /home/ahlex/repos
    a-longer-string -> /tmp

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -28,11 +28,18 @@ unmark() {
 }
 
 marks() {
+	local max=0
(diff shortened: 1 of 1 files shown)


robbyrussell:
Hey there -- we really appreciate the contribution, @ahrex!

PR #7469, opened 2018-12-14
aws_porfiles completion problem

lidow:
aws_porfiles regex is catching more lines than profiles, thus breaking completion.

Example is when have a profile with assume_role like this:

[profile sandbox]
region = eu-west-1
output = json
role_arn = arn:aws:iam::XXXXXXXXX:role/XXX
source_profile = admin
mfa_serial = arn:aws:iam::XXXXXXXXX:mfa/XXX


robbyrussell:
Thanks for this, @lidow

PR #7486, opened 2018-12-30
[Git plugin] Use stash 'push' instead of the deprecated 'save'

### LINKED ISSUE #7474: Git plugin stash alias should use push instead of save
According to git-stash man page:

git stash save: `This option is deprecated in favour of git stash push`

Currently the git plugin has an alias `gsta='git stash save'`
Should this be changed to `gsta='git stash push'`

brainplot:
CC: @ncanceill 
> Migrating to `git stash push`
> As of late October 2017, there has been extensive discussion on the Git mailing list, wherein the command `git stash save` is being deprecated in favour of the existing alternative `git stash push`. The main reason for this is that `git stash push` introduces the option of stashing selected pathspecs, something `git stash save` does not support.
> 
> `git stash save` is not going away any time soon, so don’t worry about it suddenly disappearing. But you might want to start migrating over to the push alternative for the new functionality.

The Pro Git book, in the chapter about Stashing and Cleaning, now reports the above note. The new command was introduced with Git 2.13, which came out 2 years ago. I'm using the latest version of Git and although `stash save` is still present in the `git-stash`'s man page, `stash push` has taken its place in the command synopsis, as a sign that the latter should be preferred.
```
SYNOPSIS
       git stash list [<options>]
       git stash show [<stash>]
       git stash drop [-q|--quiet] [<stash>]
       git stash ( pop | apply ) [--index] [-q|--quiet] [<stash>]
       git stash branch <branchname> [<stash>]
       git stash [push [-p|--patch] [-k|--[no-]keep-index] [-q|--quiet]
                    [-u|--include-untracked] [-a|--all] [-m|--message <message>]
                    [--] [<pathspec>...]]
       git stash clear
       git stash create [<message>]
       git stash store [-m|--message <message>] [-q|--quiet] <commit>
```
The two commands have mostly the same interface, except that with `stash push` the message is supplied via the `-m`/`--message` flag, while with `stash save` the message is the main argument of the `stash` command and requires no flags.

For this reason I think it's pretty safe to change the `gsta` alias to `git stash push`: the switch should be pretty seamless and hard to misuse as `stash push` will complain if a message is supplied with no flags as it expects a list of pathspecs.

Fixes #7474

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -33,6 +33,11 @@ function work_in_progress() {
   fi
 }
 
+function _omz_git_stash_command() {
+  [[ `git --version 2>/dev/null` =~ '^git version ([[:digit:]]+.[[:digit:]]+)' && "$match[1]" >= '2.13' ]] \
+  && echo push || echo save
+}
+
 #
 # Aliases
 # (sorted alphabetically)
@@ -238,7 +243,7 @@ alias gsps='git show --pretty=short --show-signature'
 alias gsr='git svn rebase'
 alias gss='git status -s'
 alias gst='git status'
-alias gsta='git stash save'
+alias gsta="git stash $(_omz_git_stash_command)"
 alias gstaa='git stash apply'
 alias gstc='git stash clear'
 alias gstd='git stash drop'
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -33,11 +33,6 @@ function work_in_progress() {
   fi
 }
 
-function _omz_git_stash_command() {
-  [[ `git --version 2>/dev/null` =~ '^git version ([[:digit:]]+.[[:digit:]]+)' && "$match[1]" >= '2.13' ]] \
-  && echo push || echo save
-}
-
 #
 # Aliases
 # (sorted alphabetically)
@@ -243,7 +238,12 @@ alias gsps='git show --pretty=short --show-signature'
 alias gsr='git svn rebase'
 alias gss='git status -s'
 alias gst='git status'
-alias gsta="git stash $(_omz_git_stash_command)"
+
+# use the default stash push on git 2.13 and newer
+[[ "$(git --version 2>/dev/null)" =~ '^git version ([0-9]+.[0-9]+)' && "$match" -ge '2.13' ]] \
+  && alias gsta='git stash push'
+  || alias gsta='git stash save'
+
 alias gstaa='git stash apply'
 alias gstc='git stash clear'
 alias gstd='git stash drop'


mcornella:
I'm using 2.11.0 and it's very possible that many other people still use versions below 2.13. As I said in https://github.com/robbyrussell/oh-my-zsh/issues/7474#issuecomment-455894151, maybe we can just use `git stash`, which defaults to `git stash save` and `git stash push`.

I've also checked that flags passed to `git stash` are interpreted as flags passed to `git stash save`. It probably does the same in your version. So if you're ok with it, let's just alias it to `git stash`.

mcornella:
Ok, here's what I came up with, based on `git-stash` from [v2.11.0](https://github.com/git/git/blob/v2.11.0/git-stash.sh) and [v2.13.0](https://github.com/git/git/blob/v2.13.0/git-stash.sh):
```zsh
( cd -q $ZSH; git stash undef_command_to_show_usage 2>&1 | grep -q '\[push' ) && alias gsta='git stash push' || alias gsta='git stash save'
```
It's supper ugly, but it works. I'm not yet satisfied with the solution though. For the moment you can redefine the alias in your zshrc so you don't have to see the deprecation warning.

brainplot:
I'm currently working on a solution. I'm almost there.

brainplot:
I think this is a more flexible solution. The script now parses the git version number and set the `alias` accordingly.

brainplot:
I made a change so that the function no longer spawns a `sed` process now; everything's handled by the shell.

brainplot:
I had changed it already to an inline if comparison, in my later commits.
I personally think that your first solution is better than the other two you suggested since the shell doesn't have to spawn an `awk`/`grep` process. I didn't benchmark it but I expect the first version to be quite a bit faster than the ones using `awk` and `grep` so I'd go with the first:
```zsh
[[ "$(git --version 2>/dev/null)" =~ '^git version ([0-9]+.[0-9]+)' && "$match" -ge "2.13" ]] \
  && alias gsta='git stash push' \
  || alias gsta='git stash save'
```
Also, minor nitpicking, I don't use double-quotes where I don't need stuff to be expanded (i.e. a string literal) so I would personally write it as `-ge '2.13'`. It's very minor but I like to think it's faster :)

mcornella:
> Also, minor nitpicking, I don't use double-quotes where I don't need stuff to be expanded (i.e. a string literal) so I would personally write it as `-ge '2.13'`. It's very minor but I like to think it's faster :)

It's the kind of thing I would care about so don't worry there haha

I've pushed the first version then, thanks for your help!

mcornella:
Meh, I copy-pasted badly and missed a backslash. Fixed!

brainplot:
No problem! Thank you for merging this in!

mcornella:
It turns out it doesn't work for certain versions because it makes floating point comparisons: #7754. I'll push the `is-at-least` version instead.

PR #7496, opened 2019-01-06
make installation script configurable for forks

ayandyl:
This branch replaces the hardcoded `robbyrussell` repo owner with a configurable variable in the `install.sh` script, so that fork maintainers (or private forks) can replace it with their namespace, and install the forked version of `oh-my-zsh` pointing at their fork.

This preserves original behavior by defaulting the said variable to `robbyrussell`.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1,4 +1,7 @@
 main() {
+  # change this for forks, so that it would point to your fork
+  OWNER=robbyrussell
+
   # Use colors, but only if connected to a terminal, and that terminal
   # supports them.
   if which tput >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


mcornella:
Will be fixes by #3648.

ayandyl:
noted thanks @mcornella

mcornella:
#5169 now pulls in the ability to install from forked by setting these variables:

> The installer respects the following environment variables, that can be set up when calling it with `VAR=value sh install.sh`.
> 
> * `REPO`: value in the form of `owner/project`. Useful for forks in GitHub (defaults to `robbyrussell/oh-my-zsh).
> * `BRANCH`: default branch to check out once the clone has finished (defaults to master).
> * `REMOTE`: full remote URL (supersedes `REPO` if set). You could for example choose an ssh remote URL, like `git@github.com:user/oh-my-zsh`, or another fork from GitLab or Bitbucket.

Thanks!

PR #7501, opened 2019-01-08
context prompt needs to respect light/dark config

OddDev:
Changed the hard-coded black to a variable respecting the config whether one uses Solarized light or dark. Previous version was barely readable when using Solarized Light.


robbyrussell:

You were right about the problem... under Solarized Light that segment was hard to read. It got fixed a different way in #12983: agnoster now forces white text when `SOLARIZED_THEME=light`, instead of changing the background.

If you'd rather have your version, that's a one-liner now:

```zsh
AGNOSTER_CONTEXT_BG=white
```

The line your patch touched doesn't exist anymore, so I'm closing this out. Thanks for flagging it back then.

PR #7507, opened 2019-01-13
ssh-agent: autoload identities in one go

### LINKED ISSUE #7506: #7174 requires same SSH key passphrase to be entered multiple times on automatic SSH key loading
The change in #7174 changes the behavior of the `ssh-add` calls to add SSH keys. Before this change all identities were added in a single call to `ssh-add` [1]. The SSH key agent seemed to be smart enough to only ask for the passphrase once and try them on all specified keys. So you only needed to enter it once for keys which are protected with the same passphrase.

Now the identities are added in a for loop and `ssh-add` is called separately for each key which requires me to enter the same passphrase for all identities which share it again and again until all of them are loaded. [2]

[1]: https://github.com/robbyrussell/oh-my-zsh/blob/302270174d8173be35e8c1b464a0d9e731650c15/plugins/ssh-agent/ssh-agent.plugin.zsh#L18
[2]: https://github.com/robbyrussell/oh-my-zsh/blob/fabee55948776e2e4c210e9dcd75e7bc38c02bec/plugins/ssh-agent/ssh-agent.plugin.zsh#L23-L25

G33KatWork:
With this PR the ssh-agent plugin loads all identities which are not yet
loaded in a single call to ssh-add. If a passphrase is shared between
loaded identities it only needs to be entered once.

Fixes #7506

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -13,7 +13,7 @@ function _start_agent() {
 
 function _add_identities() {
 	local id line
-	local -a identities ids
+	local -a identities ids not_loaded
 	zstyle -a :omz:plugins:ssh-agent identities identities
 
 	# get list of loaded identities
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

PR #7514, opened 2019-01-15
Make TEST_CURRENT_SHELL use basename

### LINKED ISSUE #7492: Install fails on Ubuntu 18.04 and 18.10
I'm finding the oh-my-zsh install script fails with `exit 1` and no error message when installing on `ubuntu:18.04` and `ubuntu:18.10`.

```Dockerfile
FROM ubuntu:18.10
ENV TERM screen-256color

RUN apt update && apt install -y curl zsh git

RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"
```

```sh
docker build -t ubuntu-zsh .
...
Cloning into '/root/.oh-my-zsh'...
The command '/bin/sh -c sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)"'
  returned a non-zero code: 1
```

---

I've added the command
```Dockerfile
ENV TERM screen-256color
```
which silences the following error:
```sh
tput: No value for $TERM and no -T specified
```
It appears to be non-fatal.

sabrehagen:
Fixes #7492

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -79,7 +79,7 @@ main() {
   mv -f ~/.zshrc-omztemp ~/.zshrc
 
   # If this user's login shell is not already "zsh", attempt to switch.
-  TEST_CURRENT_SHELL=$(expr "$SHELL" : '.*/\(.*\)')
+  TEST_CURRENT_SHELL=$(basename "$SHELL")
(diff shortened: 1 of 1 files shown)


mcornella:
👍

PR #7515, opened 2019-01-15
plugin is an array so comma is required

etheleon:


mcornella:
Actually no, that would make the argument have a trailing comma. Arrays in zsh are separated by spaces. Thanks for the suggestion though

PR #7518, opened 2019-01-15
Allow mvnw to be eligible to completion suggestions

bric3:
This pull requests enable zsh to suggests completions for the maven wrapper `./mvnw` when used directly. The current plugin only enable completion for `mvn` or for the `mvn-or-mvnw` function.

The [gradle plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/gradle/gradle.plugin.zsh#L182-L183) activates the completion for both `gradle` and `gradlew`, not for the _redirecting_ function. I'd like to align the functionnality so one can benefit of the completion when using `./mvnw` as well.

### DIFF
--- a/plugins/mvn/mvn.plugin.zsh
+++ b/plugins/mvn/mvn.plugin.zsh
@@ -288,5 +288,5 @@ function listMavenCompletions {
 }
 
 compctl -K listMavenCompletions mvn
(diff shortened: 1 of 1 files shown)


mcornella:
Will merge #5876.

PR #7520, opened 2019-01-16
ssh-agent: add default keys if no zstyle identities were set

mcornella:
Also, fixed an if syntax error

Fixes https://github.com/robbyrussell/oh-my-zsh/pull/7174#issuecomment-454741290

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -21,6 +21,16 @@ function _add_identities() {
 		return
 	fi
 
+	# add default keys if no identities were set up via zstyle
+	# this is to mimic the call to ssh-add with no identities
+	if [[ ${#identities} -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


mopsfelder:
Hi, @mcornella.

Thanks for submitting this fix.

In my case, `~/.ssh/id_rsa` had `400` permission, as per `stat` output:

```
➜ stat -c %a "$HOME/.ssh/id_rsa"
400
```

Removing the `stat` check for `600` permission, this pull works and fixes the issue.

I'm just curious, why do we need to check for `600` permission?

mcornella:
> I'm just curious, why do we need to check for `600` permission?

I thought that's what ssh-add did by default, but now I see that it will attempt to load the keys even if permissions are too open (in which case it will display a warning). Also, my check wasn't completely correct: it should have checked that the last two bits were 0, not force it to be `600`.

Changed, thanks for the review!

mopsfelder:
@mcornella 

Everything is working again, i.e.: SSH keys are loaded automatically.

Thanks!

PR #7521, opened 2019-01-16
ssh-agent: check for loaded id filenames first

### LINKED ISSUE #7516: ssh-add always prompts for password
hi,

since the last change in `ssh-add` i always have to enter the password for one of my keys. the problem is, that now `ssh-keygen -lf <mykey>` is used to extract the signatures of keys. but one of my keys is reported as **not a key file**.

i can reproduce the whole thing with a fresh key:
~~~
openssl genrsa -aes256 -out example.key 2048
Generating RSA private key, 2048 bit long modulus (2 primes)
.........................+++++
............+++++
e is 65537 (0x010001)
Enter pass phrase for example.key:
Verifying - Enter pass phrase for example.key:
~~~
but now:
~~~
$ ssh-keygen -l -f example.key 
example.key is not a key file.
~~~
but the key can be loaded with `ssh-add`:
~~~
$ ssh-add example.key 
Enter passphrase for example.key: 
Identity added: example.key (example.key)
$ ssh-add -l
2048 SHA256:PBYUzy+OyFzmRg1hPNBEdv6uuwQf8rRAyVmNLL6lfr8 example.key (RSA)
~~~

the `ssh-add` plugin always tries to load the signatures with `ssh-keygen` but with this new key this fails. and so i always have to enter the passphrase for this one key althoug it is loaded in the agent.

i do not know why `ssh-keygen` thinks this is an illegal key, but since the last changes to the `ssh-add` plugin every new shell asks for the passphrase.

mcornella:
This change makes the plugin check if an identity is loaded by looking first at the key filename reported by `ssh-add -l`. This fixes the use case where ssh-keygen is not able to output the fingerprint of a key, such as the one reported on #7516.
    
Now, for an identity to be passed onto ssh-add, it has to fail both the match for a loaded identity's filename and signature.

Fixes #7516

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -13,7 +13,7 @@ function _start_agent() {
 
 function _add_identities() {
 	local id line sig
-	local -a identities loaded not_loaded signatures
+	local -a identities loaded_sigs loaded_ids not_loaded
 	zstyle -a :omz:plugins:ssh-agent identities identities
 
 	# check for .ssh folder presence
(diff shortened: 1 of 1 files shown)


mcornella:
+1 from https://github.com/robbyrussell/oh-my-zsh/issues/7516#issuecomment-454865550

PR #7524, opened 2019-01-17
safe-paste: support vi mode paste restore

ahrex:
When ending `safe-paste`, we always assumed that we were restoring for emacs mode, which isn't always the case. Perform a rudimentary check to see if we're in vi mode, and if so, restore for vi mode.


ahrex:
Another follow-up: I found that the original issue I was looking for, to fix bracketed paste, was self-inflicted.

I set `KEYTIMEOUT=1`, which is not enough time for zsh to register the beginning of a bracketed paste escape sequence.

PR #7532, opened 2019-01-18
Improve adb -s completion to show helpful info

mohand1993:
Currently it shows for example the following
DEVICE_ID  -- transport_id:2
Which doesn't really ease device selection, I've adapted the awk script to print device name with it's model name, see the example below
DEVICE_ID  -- Pixel_3(blueline)

### DIFF
--- a/plugins/adb/_adb
+++ b/plugins/adb/_adb
@@ -49,7 +49,7 @@ _arguments \
 case "$state" in
 	specify_device)
 	_values -C 'devices' ${$(adb devices -l|awk 'NR>1&& $1 \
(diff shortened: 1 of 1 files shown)


mcornella:
With your change I get 
```
DEVICE_ID -- potter_n(2-1.2)
```
If I change `$3` and `$4` to `$4` and `$5` I get your output. This is my `adb devices -l` output:
```
List of devices attached
ID             device usb:2-1.2 product:potter_n model:Moto_G__5__Plus device:potter_n
```

mohand1993:
I changed it again so that it doesn't use fixed indices

mcornella:
Good job! :+1:

mohand1993:
Thanks

PR #7533, opened 2019-01-19
Plugin "debian": Switch order of "apt" and "aptitude" for detection

Zocker1999NET:
`apt` is installed by default at Debian (maybe Ubuntu too),
while `aptitude` does not seem to be installed by default.
For that, it may be better for most users to prefer `aptitude` if installed.

### DIFF
--- a/plugins/debian/README.md
+++ b/plugins/debian/README.md
@@ -10,7 +10,7 @@ plugins=(... debian)
 
 ## Settings
 
-- `$apt_pref`: use apt or aptitude if installed, fallback is apt-get.
+- `$apt_pref`: use aptitude or apt if installed, fallback is apt-get.
 - `$apt_upgr`: use upgrade or safe-upgrade (for aptitude).
 
 Set `$apt_pref` and `$apt_upgr` to whatever command you want (before sourcing Oh My Zsh) to override this behavior.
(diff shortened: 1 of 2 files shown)


Zocker1999NET:
Just asking, will this PR be merged or is something missing?

mcornella:
Isn't `apt` the preferred tool of the Debian platform going forward? I seem to recall that `aptitude` will be less up-to-date.

Zocker1999NET:
@mcornella `apt` is the recommended tool by Debian, but `aptitude` will still be updated (also it uses the same libraries) because it has a different set of features.

`aptitude` has a complex but powerful search syntax, a TUI and supports forbidding specific package versions (e.g. to prevent installing a buggy version). Also while removing a package, `aptitude` performs an `autoremove` command automatically.

`apt` lacks such features as it is just, but a good, wrapper around existing `apt-get`, `apt-cache`, … commands. `apt` is the recommended tool because it requires less resources and does its job well for the most users.

I proposed to change the order of preference because ´aptitude` is not installed by default and so must be installed by the user if wanted. On a default installation the detection of `aptitude` will fail and `apt` will be used (installed per default), which should be the behavior expected. But if an user installed `aptitude` by hand, he/she/it would expect that `aptitude` and not `apt` will be used. Especially for upgrades this could be essential because the `forbid-version` feature of `aptitude` is not supported by `apt`.

Also `apt` cannot be removed manually (at least without --force) because it is part of the package `apt`, which is marked as essential.

robbyrussell:
Reviewing what @Zocker1999NET has proposed here, I think this is a sensible approach. If people have installed aptitude in their debian environment, we can run with this priority assumption. Let's get it out and see how the community reacts to it. :-)

robbyrussell (approved):

PR #7536, opened 2019-01-20
brew: Simplify keeping least out-dated version

katrinleinweber:
Would this be a useful workflow to prevent unnecessary accumulation of old data, while not deleting the previous version right after upgrading? I'm not 100% sure, but it worked fine for me for about two weeks now.

Feedback welcome! cc @mikemcquaid

PS: This will probably conflict with #7535 & #7427 if merged. I'll take care of pushing the resolution(s) here if necessary.


plugins/brew/README.md
```diff
@@ -16,4 +16,6 @@ plugins=(... brew)
 | brewsp | `brew list --pinned` | Show the versions of pinned formulae, or only the specified (pinned) formulae if formulae are given. |
 | bubo   | `brew update && brew outdated` | Fetch the newest version of Homebrew and all formulae, then list outdated formulae. |
 | bubc   | `brew upgrade && brew cleanup` | Upgrade outdated, unpinned brews (with existing install options), then removes stale lock files and outdated downloads for formulae and casks, and removes old versions of installed formulae. |
+| bcbu   | `brew cleanup && brew upgrade` | Reverse-order of the above so that the least outdated elements of the above list are kept, after an upgrade of outdated, unpinned brews. |
```
katrinleinweber: Maybe [`-s`](https://docs.brew.sh/Manpage#commands) would make sense in that scenario as well: purging everything but the _least_ out-dated stuff?

MikeMcQuaid:
> Would this be a useful workflow to prevent unnecessary accumulation of old data, while not deleting the previous version right after upgrading? I'm not 100% sure, but it worked fine for me for about two weeks now.

Note the `brew cleanup` stuff that will be default in 2.0.0: https://brew.sh/2019/01/09/homebrew-1.9.0/ (most likely released next month)/

katrinleinweber:
In that case, I'm fine with closing this & waiting for a month :-)

katrinleinweber:
Or: Shall we turn this PR into a cleanup of the aliases themselves? `bubc` will no longer be needed since [homebrew 2.0](https://brew.sh/2019/02/02/homebrew-2.0.0/), will it?

katrinleinweber:
@MikeMcQuaid: Finally got around to implementing the "or" suggestion above. Happy New Year!

PR #7538, opened 2019-01-21
readme: update example of plugins definition to single line

dikiaap:


mcornella:
That won't be necessary: while the default configuration has to be single-line for automation purposes, documenting the use of a multi-line approach seems to me to be best and harmless.

PR #7541, opened 2019-01-22
add brew install path to search

jessefarinacci:
recent `brew install jenv` installs to `/usr/local/bin/jenv`, auto-discover it for recent brew installs to avoid secondary, slower search


mcornella:
Hey this introduced a subtle bug where `$FOUND_JENV` is always set because `/usr/local/bin` is always found. When you say `brew install jenv` installs to `/usr/local/bin/jenv`, is that a directory or a binary?

mcornella:
Nevermind, it must be a directory, so the full path needed to be used here (`/usr/local/bin/jenv`).

PR #7544, opened 2019-01-23
Git -> 🐐, Kubectl -> ☁️

haf:
Default theme:

image

TO consider `if [[ command kubectl &2>/dev/null ]]; then; EVEEERYTHING; fi`

This seems to cause issues with spacing when doing reverse-searches; see linked issue. I'd love an explanation as to why this might happen from this change.

### DIFF
--- a/lib/kubectl.zsh
+++ b/lib/kubectl.zsh
@@ -1,0 +1,6 @@
+# Outputs current kubectl in prompt format
+function kubectl_prompt_info() {
+  local ref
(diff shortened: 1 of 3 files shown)


mcornella:
Hi @haf, I'd suggest putting your customization of the robbyrussell theme as a custom theme, see [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes).

PR #7546, opened 2019-01-24
mercurial plugin: add "hgu" alias for "hg update"

michalkielan:
Add `hgu` alias for `hg update` command in the mercurial plugin.


robbyrussell:
@michalkielan would you mind rebasing this?

PR #7548, opened 2019-01-24
Show recent Git branches using `grbr` alias

vernondegoede:


teameh:
Maybe this would be better on line 53 instead as `alias gbr='git branch --sort=-committerdate | head -5'`? All the other `git branch` aliases are located there.

vernondegoede:
The reason I chose `grbr` is because `gbr` is already aliased to `git branch --remote`

manzapanza:
I think that this alias should not start with `grb` because are aliases for git rebase. So IMHO should start with `gb`. Maybe `gbsr` (Git Branch Sort Recent) or some other name.

PR #7555, opened 2019-01-28
Updated Agnoster Theme

SheinH:
The agnoster theme in this repo is outdated. If you go to the original [Agnoster repo ](https://github.com/agnoster/agnoster-zsh-theme) none of the instructions there are applicable because the version included with oh-my-zsh is old.

This PR contains an updated version of Agnoster in themes/ pulled straight from the original repo.


sjaveed:
any word on whether the official agnoster theme should match the one from this repo or not?

robbyrussell:
@carlosala what do you think? Should we be directing people to use the authoratative theme going forward?

carlosala:
Official agnoster theme was "abandoned" a while ago. Our copy diverged a lot from it, I think at this point it can be considered a completely different theme.
I don't think any action is needed.

PR #7571, opened 2019-02-02
ggl giving error, temporarily fixed by commenting the lines

abdullahkhilji:
/Users/my_username/.oh-my-zsh/plugins/git/git.plugin.zsh:120: defining function based on alias `ggl'
/Users/my_username/.oh-my-zsh/plugins/git/git.plugin.zsh:120: parse error near `()'

Error given on: macOS Mojave 10.14.2


mcornella:
Closing since this is not a proper solution. Let's continue the conversation on #7572

Issue #6261, opened 2017-08-22
Up-Arrow behavior different depending on if previous program has finished or not


greggman:
I've noticed that if I press up arrow at a prompt then I get the previous command and up again gets me the command before that.

Whereas if I press up arrow *before the previous program has completed* then instead I get the previous command displayed, the cursor is at the end of the line, but oh-my-zsh is now in "search for lines that start with ... " mode meaning I can't press up to get the previous command.

I'm sure this behavior is well known and expected but just in case you can test like this. 

1. Type `ls` <kbd>return</kbd>  
2. Type `sleep 3` <kbd>return</kbd>
3. wait 3 seconds for prompt to appear
4. press <kbd>up</kbd> (should show `sleep 3`)
5. press <kbd>up</kbd> again (should show `ls`)
6. press <kbd>return</kbd> (to run `ls`)
7. Type `sleep 3` <kbd>return</kbd> <kbd>up</kbd> (press the up arrow before the 3 seconds elapses)
8. It should now be showing `sleep 3`
9. Press <kbd>up</kbd> 
   it will still be showing `sleep 3` but it want it to be showing `ls`. Instead it is in "search for commands that start with `sleep 3` mode instead of just `go to previous command` mode.

To make try to clear in both cases these are the steps

1. `ls`<kbd>return</kbd> 
2. `sleep 3`<kbd>return</kbd> 
3. <kbd>up</kbd> 
4. <kbd>up</kbd> 

But they end up with different results depending on if step3 happens before step2 finished or after.

How do I get oh-my-zsh to be consistent here so that a premature up arrow behaves the same as a normal up arrow? Thank you

greggman:
Does this really not affect anyone but me? I'm really surprised.

greggman:
In any case those looking for a fix you can add this to your .zshrc file sometime after including oh-my-zsh.sh

```
bindkey "^[[A" up-line-or-beginning-search # Up
bindkey "^[[B" down-line-or-beginning-search # Down
```

mcornella:
I can confirm this happens to me too. The reason is that when a command is running, zle is not active and as such the terminal is not in application mode:
https://github.com/robbyrussell/oh-my-zsh/blob/4aa545abedb430df2fa1ae2cc7cc502a5be1a80b/lib/key-bindings.zsh#L5-L16

[...]

PR #7574, opened 2019-02-03
example fix for cursor up/down consistency issue

greggman:
not a real PR but just an example of the fix.

See https://github.com/robbyrussell/oh-my-zsh/issues/6261

Someone more familiar with the terminal could maybe explain

* why this works
* if there is some symbolic terminfo constant that should be used instead
* if it would be good to make this the default
* if not the default, would it be good to add an option like `OH_MY_ZSH_CURSOR_UPDOWN_CONSISTENT_BEHAVIOR=true`

### DIFF
--- a/lib/key-bindings.zsh
+++ b/lib/key-bindings.zsh
@@ -31,13 +31,15 @@ fi
 if [[ "${terminfo[kcuu1]}" != "" ]]; then
   autoload -U up-line-or-beginning-search
   zle -N up-line-or-beginning-search
-  bindkey "${terminfo[kcuu1]}" up-line-or-beginning-search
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry nobody answered your questions at the time... they were good ones.

This did get fixed properly in 8f22370a (Feb 2024). The approach was the one you were circling: bind the raw `^[[A` / `^[[B` sequences alongside the terminfo ones, so the arrows behave the same whether or not application mode is on. Both pairs now map to `up/down-line-or-beginning-search`.

Closing since it's covered. Thanks for taking the time to dig into it.

PR #7581, opened 2019-02-07
Add Givero (supports good causes) as search engine

theill:
Will add "givero" as a keyword for searching Givero, a relative new search engine donating revenue to good causes around the world.


mcornella:
Thanks!

PR #7582, opened 2019-02-07
asdf: fix homebrew installation path

eric-christian:
* The check for the asdf installation directory is more precise.
* Completions, while installed with homebrew, are now expected on an alternative location.

——

The existence of the directory `$HOME/.asdf` does not mean that it is the installation directory of `asdf`. It will also be created after installing at least one asdf plugin.

@robbyrussell  @mcornella @RobLoach


mcornella:
Thanks!

PR #7592, opened 2019-02-14
Add Git alias merge develop

alvince:
Add alias 'gmod' => 'git merge origin/develop'
Add alias 'gmud' => 'git merge upstream/develop'


robbyrussell:
Thanks for the proposed aliases!

Unfortunately, we are going to reject this particular one as it hardcodes a specific git branch name (`develop`). Would suggest this type of alias go into a custom plugin/extension that you could share with your teammates.

PR #7597, opened 2019-02-17
Revert "Fix emacs client terminal"

akinnane:
Reverts robbyrussell/oh-my-zsh#5714


akinnane:
This change breaks my emacs workflow. Instead of the `emacs` alias opening the file in the current emacs frame(window), it always opens emacs in the current terminal instead. There is already the `te` alias to open emacs in the current terminal.

ilyaiqoqo:
This broke my flow too, e always opens file in terminal now

PR #7603, opened 2019-02-18
 update rn plugin with ipad and apple tv sims

esthor:
just adding some more simulator aliases


esthor:
@mcornella can you take a look? 😄

mcornella:
I don't have macOS but I presume these strings work. Can you add the Apple TV ones to the README as well?

esthor:
@mcornella nice catch, totally forgot to add them. Yeah, works on my system 🤷‍♀️ 😄

mcornella:
👍

PR #7606, opened 2019-02-18
[plugins/tmux] Allow to define a custom configuration path

StelFux:
Hello,
Tmux developers do not want to comply with XDG standards ([see here](https://github.com/tmux/tmux/issues/142)). With this variable, those who wish to do so will be able to comply with this standard, without the agreement of these developers.
Thank you very much.


mcornella:
Thanks!

PR #7607, opened 2019-02-19
New alias gcB for git checkout -B

jblejder:
I would like to introduce new git alias for `git checkout -B`.
There already is `gcb` for creating new branches so this is variation with `uppercase B`.

There is common pattern with with `gbd` and `gbD`

This command is useful when you want to reset branch to clean state from remote.

Calling maintainer @ncanceill


mcornella:
This is an obscure command that almost nobody will use.

PR #7608, opened 2019-02-19
vscode: handle command name differences with same aliases

jjangga0214:
- If vscode is installed from snap store(e.g. `snap install vscode`), the default command would be
_vscode_, whereas generally it is _code_. To provide a uniform
interface(aliases) to user with different commands(_vscode_ vs
_code_), 'command resolver' is added at the top of the file. The value
of `$VSCODE` will be **vscode** if vscode command exists in the system.
Otherwise, it will be **code**.
- Reformatted the comments and added an author(https://github.com/jjangga0214)

### DIFF
--- a/plugins/vscode/vscode.plugin.zsh
+++ b/plugins/vscode/vscode.plugin.zsh
@@ -1,19 +1,23 @@
-# VScode zsh plugin
-# author: https://github.com/MarsiBarsi
+# Visual Studio Code aliases
+# Authors:
+#   - https://github.com/MarsiBarsi
+#   - https://github.com/jjangga0214
 
-alias vsc='code .'
-alias vsca='code --add'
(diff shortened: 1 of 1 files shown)


mcornella:
In my system the snap also installed a symlink called `code`:
```
$ ll /snap/bin
total 0
lrwxrwxrwx 1 root root  6 mar 25 19:21 code -> vscode*
lrwxrwxrwx 1 root root 13 mar 25 19:21 vscode -> /usr/bin/snap*
lrwxrwxrwx 1 root root 13 mar 25 19:21 vscode.url-handler -> /usr/bin/snap*
```
which I assume is the same if you have the same snap version (`snap refresh`). You could also use `sudo snap alias vscode code` to add a symlink yourself.

PR #7612, opened 2019-02-19
Gitfast: add spacing before git prompt info

hjdarnel:
It's my strong preference that there be separation in the Git prompt info that indicates branch name, as well as indicating rebase or other state.

This would move from this:
👉 `feature-node-10|REBASE 1/1 ✗` to `feature-node-10 | REBASE 1/1 ✗` 👈 

With clear separation between branch name and "other state"

I noticed this behavior when using the gitfast plugin for the first time. Since I copy/paste this name often, it's a pain to have to carefully deselect the pipe to only get the branch name.

Tagging known maintainers :) 
@felipec

### DIFF
--- a/plugins/gitfast/git-prompt.sh
+++ b/plugins/gitfast/git-prompt.sh
@@ -397,9 +397,9 @@ __git_ps1 ()
 		__git_eread "$g/rebase-merge/msgnum" step
 		__git_eread "$g/rebase-merge/end" total
 		if [ -f "$g/rebase-merge/interactive" ]; then
-			r="|REBASE-i"
+			r="| REBASE-i"
 		else
-			r="|REBASE-m"
+			r="| REBASE-m"
 		fi
 	else
 		if [ -d "$g/rebase-apply" ]; then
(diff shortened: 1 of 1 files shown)


felipec:
Isn't this much easier?

```sh
    [ -n "$r" ] && r="${r/|/ | }"
```

Personally I don't like the format, but it could be configured:

```diff
--- a/plugins/gitfast/git-prompt.sh
+++ b/plugins/gitfast/git-prompt.sh
@@ -458,6 +458,9 @@ __git_ps1 ()
                fi
        fi
 
+       [ -n "${GIT_PS1_EXPAND_REBASE-}" ] &&
+       [ -n "$r" ] && r="${r/|/ | }"
+
        if [ -n "$step" ] && [ -n "$total" ]; then
                r="$r $step/$total"
        fi
```

It would be better if such a patch was sent to upstream git. That's where I get the updates from.

hjdarnel:
I think that that code could patch it, sure, but it seems silly to set the strings and then run a replace against the string instead of just set them properly initially. I wasn't aware this file was a fork, do you have a link to suggest a patch?

If you don't prefer this, I don't mind just using my fork if you want to close it.

felipec:
Well, you use the qualification "properly" to justify your argument, but it's not properly to me: I prefer the code as it is.

If you make the code optional, they adding a check to every single line that changes the variable is stupid; it's much better to change the code in a single place.

Not only it's easier to understand for the developers what you are trying to do, it's easier to maintain, and it's less prone to bugs.

I was the one that wrote the wrapper to make Git's bash completion work perfectly in zsh, I just copied the code to oh-my-zsh so you don't have to manually set it up.

If you want to manually set it up just follow the [instructions](https://github.com/git/git/blob/master/contrib/completion/git-completion.zsh).

To send patches to Git the instructions are clear in the [site](https://git-scm.com/community).

And there's a fourth reason to keep the patch small: if they don't accept it upstream we could just keep it here, and since it's small there would not be much problem to sync with upstream. Of course I would prefer not to have extra changes.

PR #7615, opened 2019-02-21
Update the AWS plugin to support disabling RPROMT display

### LINKED ISSUE #6127: aws plugin - (feat) add a way to clear the asp command
The aws set profile command is great, but there isn't a way to clear it and have no profile set.

dkanejs:
Many popular themes display the current profile inline so this update adds the option to disable the addition to the RPROMT if required.

~~update aws.plugin.zsh to use a $HIDE_AWS_PROMPT option~~
update aws.plugin.zsh to use a $SHOW_AWS_PROMPT option
add usage to the readme

Tagging @cristim (Are you still the primary maintainer?)

----

UPDATED: fixes #6127

### DIFF
--- a/plugins/aws/README.md
+++ b/plugins/aws/README.md
@@ -1,8 +1,7 @@
 # aws
 
 This plugin provides completion support for [awscli](https://docs.aws.amazon.com/cli/latest/reference/index.html)
-and a few utilities to manage AWS profiles: a function to change profiles with autocompletion support
-and a function to get the current AWS profile. The current AWS profile is also displayed in `RPROMPT`.
+and a few utilities to manage AWS profiles and display them in the prompt.
 
 To use it, add `aws` to the plugins array in your zshrc file.
 
@@ -13,7 +12,6 @@ plugins=(... aws)
 ## Plugin commands
 
 * `asp <profile>`: Sets `AWS_PROFILE` and `AWS_DEFAULT_PROFILE` (legacy) to `<profile>`.
-It also adds it to your RPROMPT.
 
 * `agp`: Gets the current value of `AWS_PROFILE`.
 
@@ -22,3 +20,11 @@ It also adds it to your RPROMPT.
 ## Plugin options
 
 * Set `SHOW_AWS_PROMPT=false` in your zshrc file if you want to prevent the plugin from modifying your RPROMPT.
+
+## Theme
+
+The plugin creates a `aws_profile_prompt_info` function that you can use in your theme, which displays the current `$AWS_PROFILE`. It uses two variables to control how that is shown:
+
+- ZSH_THEME_AWS_PREFIX: sets the prefix of the AWS_PROFILE. Defaults to [.
+
+- ZSH_THEME_AWS_SUFFIX: sets the suffix of the AWS_PROFILE. Defaults to ].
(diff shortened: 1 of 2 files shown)


cristim:
@dkanejs thanks for this contribution, LGTM.

Regarding maintainership, I think @robbyrussell is the only one who really matters, I can merely approve the PR but I can't merge it.

plugins/aws/README.md
```diff
@@ -18,3 +18,7 @@ It also adds it to your RPROMPT.
 * `agp`: Gets the current value of `AWS_PROFILE`.
 
 * `aws_profiles`: Lists the available profiles in the file referenced in `AWS_CONFIG_FILE` (default: ~/.aws/config). Used to provide completion for the `asp` function.
+
+## Plugin options
+
+* Set `HIDE_AWS_PROMPT=true` in your zshrc file if you want to prevent the plugin from modifying your RPROMPT.
```
cristim: Why have it called `HIDE_` and use negated logic, when you can use `SHOW_` instead ?

dkanejs: Good point, i'll get this changed 👍

dkanejs: @cristim Got that changed

Also had to move the `local rprompt` declaration back into function scope outside of the if statement as It was causing RPROMPT appending issues when switching profiles, and during profile deselection.

I ended up with all my profiles in my prompt and they couldn't clear 😆

dkanejs:
@cristim let me know if this needs anything else 👍

mcornella:
Let's take a different approach here: since we can set RPROMPT to get the up-to-date value of `$AWS_PROFILE` by using single quotes, we can just set the RPROMPT value once and forget about it in the asp function:

```zsh
asp() {
    ...
}
RPROMPT='<aws:$AWS_PROFILE>'"$RPROMPT"
```

This would render #5821 unnecessary since `asp` could be called without RPROMPT change, no matter the user setting. You can even use the setting to not change RPROMPT if the user doesn't want it to.

vmous:
Unaware of this pull request I have published https://github.com/robbyrussell/oh-my-zsh/pull/7624.
Compared to this one not only it solves the problem of the obligatory setting of `$RPROMPT` info but additionally offers customizable info function (similar to other plugins) to be used in themes plus the ability to show the profile info by just looking at the existence of `$AWS_PROFILE` variable rather than requiring someone to have used the `asp` command to set it.

dkanejs:
> Let's take a different approach here: since we can set RPROMPT to get the up-to-date value of `$AWS_PROFILE` by using single quotes, we can just set the RPROMPT value once and forget about it in the asp function:

Sorry @mcornella, I'm not sure I follow.

By setting the profile in that location it would cause the prompt to display before `asp` has been called, no?

Any and all help would be great 😄

mcornella:
> By setting the profile in that location it would cause the prompt to display before `asp` has been called, no?

That's true. So we can just end up creating a `aws_prompt_info` function that only prints if `AWS_PROFILE` is set, and put that in RPROMPT. Something like what #7624 does. I'll update this PR with a few changes, let me know how it works.

mcornella:
Updated. Let me know how you like it.

dkanejs:
@mcornella yup this is awesome 👍

mcornella:
Perfect then, merged :wink:

PR #7621, opened 2019-02-28
added shortcut for git fetch upstream

odysseus0:
Since `git fetch upstream` is pretty common, we could have an alias for it. It does not conflict with anything else from this plugin.


mcornella:
Is there a reason why `git fetch --all` (`gfa`) doesn't work for you?

odysseus0:
Ah, did not even know that it exists. My bad.

mcornella:
Ok thanks.

PR #7622, opened 2019-02-28
Enable tmux window title updates when using proper tmux* $TERM values.

JadedDragoon:
There are minor but important differences between tmux and screen terminal features. Modern GNU distributions now come with `tmux` and `tmux-256color` termcaps and it is preferable to use these when using tmux. The small change in this PR allows functional tmux window title updates using existing oh-my-zsh functionality when using a "tmux*" `$TERM` .


mcornella:
Thanks!

PR #7624, opened 2019-03-01
Refactoring aws plugin.

vmous:
Adding new utility function for unsetting aws profile.
Renaming old functions with self-explainable names.
Exposing customizable `aws_profile_prompt_info` function to be used in themes.
Showing info if `$AWS_PROFILE` is set, not only when using aws_set_profile.
Removed obligatory addition of aws profile information to `$RPROMPT`.

### DIFF
--- a/plugins/aws/README.md
+++ b/plugins/aws/README.md
@@ -1,8 +1,6 @@
 # aws
 
-This plugin provides completion support for [awscli](https://docs.aws.amazon.com/cli/latest/reference/index.html)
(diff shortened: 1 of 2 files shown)


mcornella:
Too many changes: the function names are ok as they are, and the `asp` function can already be used to clear profile names by not passing a parameter.

Just fixing the `asp` function to not mess up with `$RPROMPT` and having the `aws_prompt_info` function is enough. I'll use some of your changes in #7615 and end up merging that other one instead.

Thanks for your contribution!

PR #7629, opened 2019-03-04
Git Plugin: Add the git alias 'gtl', that list the tags that match the pattern(s)

ameltonyan:
Add the alias 'gtl', which does the following:

1. List the tags that match the pattern(s) passed through the argument.
2. Displays the first line of the annotation message along with the tag, or the line of the first commit message if the tag is not annotated.
3. Sorts and displays tags in descending order.

Suppose there are several tags:
v1.2.1
v2.5.1
v2.5.2
v3.5.1
...

When running gtl v2.5 this will result in

v2.5.2  -commit message 
v2.5.1  -commit message

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -250,7 +250,7 @@ alias gsu='git submodule update'
 
 alias gts='git tag -s'
 alias gtv='git tag | sort -V'
-alias gtl='function _gtl(){ git tag -n -l ${1:-"*"}* --sort=-v:refname; };_gtl'
+alias gtl='gtl(){ git tag --sort=-v:refname -n -l ${1}* }; noglob gtl'
 
 alias gunignore='git update-index --no-assume-unchanged'
 alias gunwip='git log -n 1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'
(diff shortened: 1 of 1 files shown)


mcornella:
How does this play a role in your git workflow and do you use it often?

ameltonyan:
Look, on one real example, we have three branches within one plugin project and each of them is used for working correspondingly with Grails 2.x, 3.x and 4.x versions. From the perspectives of business logic each branch contains the same features. But the implementations differ from Grails version to another. Daily our developers are making some improvements and development on each of them to support our old and new applications on different versions. In cases of some plugins we have more than 300 tags, some of them for 2.x, 3.x, 4.x and so on in future. So every day we face an issue of scrolling and finding the desired tags. At least on example of almost 100 developers in our company it is treated as very useful "life-hack".

mcornella:
I've updated the alias to also use `noglob`. Now you can run it with pattern-matching characters without having to quote them. For example: `gtl *-dev` works instead of throwing a `no matches found` error.

If you're ok with the changes let me know and I'll merge this ASAP.

ameltonyan:
Yes of course. that's cool, Thanks.

mcornella:
:+1:

NimoTh:
The current alias definition does not work for me. If I add quotes around the search string (as also done in git documentation), it works:
```alias gtl=gtl(){ git tag --sort=-v:refname -n -l "${1}*" }; noglob gtl```
Should this be fixed?

mcornella:
You're right, in testing I didn't catch that because I have aliased git to `noglob git`. Fixed by https://github.com/robbyrussell/oh-my-zsh/commit/093b56a7d769d15bdcc2c5300bc4ebedc9d0a837

PR #7631, opened 2019-03-04
ssh-agent: allow using external helper to ask for passwords (and store them in a keychain)

simgunz:
By setting `SSH_ASKPASS=ksshaskpass` for example it is possible to use an helper to ask for the passwords to unlock the SSH keys. The helper may also able to store the password in the system keychain, without having to enter the passwords at every boot.

To make this mechanism work it is necessary to call 
```
ssh-add {identities} < /dev/null
```


abrac (approved):

simgunz:
- Rebased to master to solve conflicts
- Use bash default value for `in_from`

simgunz:
@abrac is this MR still approved? What is needed to get it merged?

PR #7633, opened 2019-03-04
theme/agnoster: Fixed light colors for git dirty prompt

zydeon:
Fixes foreground color for git prompt when it contains changes and light mode is used:

screen shot 2019-03-04 at 2 00 44 pm

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -111,7 +111,7 @@ prompt_git() {
     dirty=$(parse_git_dirty)
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git rev-parse --short HEAD 2> /dev/null)"
     if [[ -n $dirty ]]; then
-      prompt_segment yellow black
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it sat for so long.

Closing it because the line you patched is gone... #12505 moved all of agnoster's colors into `AGNOSTER_*` variables, so `prompt_git` no longer hardcodes `yellow black`.

But I want to be straight with you: your actual complaint is still there in the defaults. The clean and dirty segments aren't symmetric:

```zsh
: ${AGNOSTER_GIT_CLEAN_FG:=${CURRENT_FG}}   # follows light/dark
: ${AGNOSTER_GIT_DIRTY_FG:=black}           # doesn't
```

So with `SOLARIZED_THEME=light` you still get black on yellow, exactly what you reported in 2019. For now:

```zsh
AGNOSTER_GIT_DIRTY_FG=white
```

I'm treating the default asymmetry as its own small fix rather than reopening a seven-year-old branch for it. You were right, and it took us far too long to say so.

PR #7637, opened 2019-03-06
allow kubectl commands against all namespaces

ulrichSchreiner:
hi,

found this snippet in the internet and i use it all the time. i think it would be useful for the `kubectl` plugin


robbyrussell:

Cheers,
Robby

PR #7656, opened 2019-03-14
Improvement to issue number  #7332

chuckstar76:
Improvement to issue number  #7332
updated my oh-my-zsh.sh file - because of the same reason "it looks like this is because regardless of the set $ZSH directory the ZSH_COMPDUMP file is always set to point to the $HOME directory."

```
if [ -z "$ZSH_COMPDUMP" ]; then
  ZSH_COMPDUMP="${ZDOTDIR:-${ZSH}}/cache/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
fi
```

Replacing ${HOME}} with ${ZSH}}/cache
I think this is a much more tidy way of doing things.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -64,7 +64,7 @@ fi
 
 # Save the location of the current completion dump file.
 if [ -z "$ZSH_COMPDUMP" ]; then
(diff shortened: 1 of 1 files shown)


chuckstar76:
@mcornella - have I messed up here with these pull requests?  Please advise... TIA

chuckstar76:
Closing request - stale.

PR #7660, opened 2019-03-17
web-search: fix #7659, update alias for stackoverflow

### LINKED ISSUE #7659: [Plugin] Alias in plugin 'web-search' conflicts with Haskell Tool Stack
### What is it
The newly added alias for `stackoverflow` in the web-search plugin conflicts with the common Haskell Tool Stack command.

### How to reproduce it
After enabling the web-search plugin in oh-my-zsh, then use the Haskell Tool Stack:
```
stack ghci
```
A browser window will pop up with search results of 'ghci' on stackoverflow.
This alias is really confusing to people who use the Haskell Tool Stack.

### Problem code
https://github.com/robbyrussell/oh-my-zsh/blob/9509fd6a9193e32e5c2d460786253ba0e98fd741/plugins/web-search/web-search.plugin.zsh#L55

### Proposed fix
Use a longer alias for 'stackoverflow,' for example 'stackover' to replace the confusing alias.

@iamantu93

LaytonW:
Apply proposed fix in #7659 

@iamantu93

### DIFF
--- a/plugins/web-search/README.md
+++ b/plugins/web-search/README.md
@@ -37,6 +37,7 @@ Available search contexts are:
 | `ecosia`              | `https://www.ecosia.org/search?q=`       |
 | `goodreads`           | `https://www.goodreads.com/search?q=`    |
 | `qwant`               | `https://www.qwant.com/?q=`              |
+| `stackover`           | `https://stackoverflow.com/search?q=`    |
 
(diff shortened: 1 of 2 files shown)


mcornella:
Can you change it to `stackoverflow`? Since we're making it longer we might as well name it properly and let command completion take care of the rest.

LaytonW:
Updated.

PR #7662, opened 2019-03-18
colorize: add ability to colorize multiple files into less

fzuviria:
Fancy new colorize alias plus bugfix


fzuviria:
Hello mcornella.
I was wondering why the PR was closed.
Best of regards.

mcornella:
I just thought this was a PR made by mistake given the lack of information and weird title. What does this do?

fzuviria:
Colorize provides de ccat (colorized cat) im adding the colorized less -
cless

I was checking it and it seems I commented by mistake an unrelated part of
the code, i gotta fix that.

Can the pr be reopened, or would you like a new pr? Whats the branch and pr
naming convention that you use?

mcornella:
Isn't cless the same as `ccat | less`? You can reopen and push-force to the same branch to fix the commented file.

fzuviria:
You can't page through files with ccat|less doing :n and :p

fzuviria:
Ill do that, thanks

mcornella:
> You can't page through files with ccat|less doing :n and :p

You can

mcornella:
> You can't page through files with ccat|less doing :n and :p

You can use something like this instead of generating tmp files:
```
cless() {                        
    local -a args
    for f ($@); do
        args+="<(colorize_via_pygmentize "$f")"
    done
    eval "less -f $args"
}
```

The downside is that it doesn't print the original filenames. It also doesn't parse arguments that should be passed to `pygmentize` or `less`.

fzuviria:
Hi there, I updated the PR with some cleanup, before reading you.

Another downside of your proposal is that it uses eval which can be nasty
and unpredictable, say if files have any character that breaks shell quoting

Maybe I'm wrong, as I know more about bash than anything else

Also, my environment runs with
setopt SH_WORD_SPLIT

which will probably give me even more trouble

My first solution was with an array of named pipes but I found
that python-pygments do no treat correctly signals (not sure if SIGPIPE or
SIGINT) which caused ugly python stack traces and in some cases infinite
loop of failed reads, flooding all the terminal
Yea, I'm not a big fan of tossing around temp files either but oh well...

For some reason your implementation does not make python-pygments fail on
me, yet

What do you want to do?

fzuviria:
I just noticed on some testing that using process substitution has the
dusadvantage that, as the filename is lost, also the file extension is
lost, and colorize fails to recognize the sintax in many cases. For
example, it fails to colorize a wsdl (which is crazy, I would have thought
that recognizing xml would be easy)

Please accept the proposed implementation.

Best of regards,

Francisco.

fzuviria:
Hello @mcornella I was hoping if you have any feedback on my last messages. My team mates are asking me for it and I would rather have them get it from upstream ^_^

plugins/colorize/colorize.plugin.zsh
```diff
@@ -1,7 +1,11 @@
 # easier alias to use the plugin
 alias ccat='colorize_via_pygmentize'
+alias cless='colorize_via_pygmentize_less'  
 
 colorize_via_pygmentize() {
+    # # I have been having problems with pygmentize signal handling
```
mcornella: This should be deleted.

plugins/colorize/colorize.plugin.zsh
```diff
@@ -26,3 +30,25 @@ colorize_via_pygmentize() {
...
     done
 }
+
+colorize_via_pygmentize_less() (   
+    # this function is a subshell so tmp_files can be shared to cleanup function
+    declare -a tmp_files 
+    
+    cleanup () {
+        [[ "${#tmp_files[@]}" != "0" ]] && rm -f "${tmp_files[@]}"
+        if [ -n "$1" ]; then kill -$1 $$; fi #signal propagation
```
mcornella: Why are you doing this?

plugins/colorize/colorize.plugin.zsh
```diff
@@ -26,3 +30,25 @@ colorize_via_pygmentize() {
...
+    }
+    trap 'cleanup' EXIT
+    trap 'cleanup HUP' HUP
+    trap 'cleanup TERM' TERM
+    trap 'cleanup INT' INT
+    
+    while (( $# != 0 )); do     #TODO: filter out less opts
+        tmp_file="$(mktemp --tmpdir tmp.colorize.XXXX.$( sed -E 's/[^a-Z0-9]/./g' <<< "$1"))"
```
mcornella: My `sed` complains about an invalid range end, it should be `a-zA-Z`, not `a-Z`. Why are you doing this sed operation though?

plugins/colorize/colorize.plugin.zsh
```diff
@@ -26,3 +30,25 @@ colorize_via_pygmentize() {
...
+        tmp_files+=("$tmp_file")
+        colorize_via_pygmentize "$1" > "$tmp_file"
+        shift 1;
```
mcornella: Small detail: extra semicolon.

plugins/colorize/colorize.plugin.zsh
fzuviria: I researched for the most robust cleanup function I could come with, and I can't find exactly the stack overflow question where I got this from, but from https://unix.stackexchange.com/questions/57940/trap-int-term-exit-really-necessary this would be the rationale:

This solution isn't robust when another shell calls it. It doesn't handle wait on cooperative exit; you will want to trap - INT TERM; kill -2 $$ as the last line of cleanup, to tell the parent shell that it exited prematurely. If a parent shell foobar.sh calls your script (foo.sh), and then calls bar.sh, you don't want bar.sh to execute if INT/TERM is sent to your foo.sh. trap cleanup EXIT will handle this propagation automatically, so it is IMO the most robust. It also means you wouldn't have to call cleanup at the end of the script. – Nicholas Pipitone Jul 20 '18 at 20:47

$$ Expands to the process ID of the shell. In a () subshell, it expands to the process ID of the current shell, not the subshell.

So basically, signal propagation to the parent. Now that i think about it, I hope it does not mess somehow with the interactive shell. In the current form this is implemented (all sequential), I don't think this is needed. 

It may be useful to have this handy in case of a future refactor, should I comment it out, or delete it entirely? Still learning the coding style of this codebase...

plugins/colorize/colorize.plugin.zsh
fzuviria: Got it

plugins/colorize/colorize.plugin.zsh
fzuviria: Ok, I made it more portable, like this:
`    sed -E 's/[^a-zA-Z0-9-]/./g'`
2 reasons: I made that sed to prevent problems with files that may include unwanted characters like 
`    mark's c**l file~`
to expand, or break quoting, so it replaces any non alphanumerical with a dot. (Do you think this is unnecesarry?)
Second reason, it also replaces directory separators /, so if I do something like
`cless $(find . -name pom.xml) # this work's with my shell opts`
I will be _cless_ing many files with the same name but I will be able to see the path of the file I'm reading.
For comparison, the less status at the bottom would be
`./main/dev/target/pom.xml (file 1 of 8) lines 1-36/151 29%`
and the cless status
`/tmp/tmp.colorize.BdN9...main.dev.target.pom.xml (file 1 of 8) lines 1-36/151 27%`
If I didn't do the sed and I wanted to be able to differentiate them, I would have to do `mkdir -p` on /tmp, which I think would be making too much dirt...

plugins/colorize/colorize.plugin.zsh
fzuviria: I love small details. Done.

plugins/colorize/colorize.plugin.zsh
mcornella: OK I deleted the entire comment: we have this PRs history if we need to go back to it.

plugins/colorize/colorize.plugin.zsh
mcornella: I changed it so that it only replaces slashes in the filename, the rest of characters shouldn't be an issue since the code is properly quoted.

mcornella:
From a few tests I've done, maybe we should exit from the cleanup function so the subshell is closed and any subprocess within it. Here's a sample function to test this:

```zsh
hi() (
local -a array_of_files

cleanup() {
    echo cleanup $1
    rm -vf "${array_of_files[@]}"
    exit
}

trap 'cleanup EXIT' EXIT
trap 'cleanup HUP' HUP
trap 'cleanup TERM' TERM
trap 'cleanup INT' INT

for file in file{1..3}; do
    tmp_file="$(mktemp --tmpdir "tmp.XXXX.$file")"
    array_of_files+=("$tmp_file")
    echo $file > "$tmp_file"
done
sleep 5
less -f "$array_of_files"
)
```

If I CTRL+C when the code is sleeping, the cleanup function runs because of the SIGINT and exits:
```
$ hi
^Ccleanup INT
removed '/tmp/tmp.LqjR.file1'
removed '/tmp/tmp.W7KV.file2'
removed '/tmp/tmp.Vw2b.file3'
cleanup EXIT
```

If I remove the exit, the hi function continues and the less tries to open the tmpfiles:
```
hi    
^Ccleanup INT
removed '/tmp/tmp.vGUS.file1'
removed '/tmp/tmp.q1AL.file2'
removed '/tmp/tmp.Ag1k.file3'
/tmp/tmp.vGUS.file1 /tmp/tmp.q1AL.file2 /tmp/tmp.Ag1k.file3: No such file or directory
cleanup EXIT
```

which to me seems weird; the CTRL+C should IMO stop anything running and that includes the hi function execution. But since this is not what happens I'll add the `exit` call to the cleanup function because we don't know when a signal will trigger.

----
As per the quoting of special characters in filenames, I've made sure through testing that I haven't introduced any bug such as what you warned about.

Let me know what you think of the latest changes and I'll merge this.

fzuviria:
Hello Mark,
Everything looks ok to me. You did a nice cleanup :) I ran a test with spaces and with my shellopts it works like a charm as well :D 
You can merge it. I don't know if I'm supposed to do something at this point (I'm still new to github)

mcornella:
Great, thanks for your contribution! See you around

PR #7675, opened 2019-03-20
Update ys.zsh-theme

danerlt:
change path color to green


mcornella:
This is a breaking change and a personal preference, I suggest you follow the guide for how to customize the theme, [see here](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes).

PR #7676, opened 2019-03-20
Two new alias

juliofalbo:
Adding 2 new alias

- gcfix='git commit --fixup HEAD'
- gprum='git pull --rebase upstream master'


LoLei:
Shouldn't that be `glrum` rather?
Since `gl` is `git pull` and `gp` is `git push`.

juliofalbo:
Perfect @LoLei !

mcornella:
- `commit --fixup HEAD` is basically will have the same effect as `commit --amend`, only after running `git rebase -i --autosquash`.
- `glrum` can be run using the `glum` alias, with `glum -r`.

I really don't see the need to add 2 more aliases for these little changes, see #5721.

PR #7677, opened 2019-03-20
Also indicate git tags properly

er-vin:


robbyrussell:

This was fixed in #11318 in 2022 — `lib/git.zsh` now falls back to `git describe --tags --exact-match HEAD` when HEAD is detached, so the current behavior shows the tag name. Closing this one out.

Sorry you didn't hear back at the time.

PR #7678, opened 2019-03-20
We don't need sudo for some zypper commands

er-vin:
Closes #5564


er-vin:
And now with updated README.
Regards.

mcornella:
I added some other aliases from #5564 as well. Thanks!

PR #7680, opened 2019-03-21
Add support for searching in thewirecutter.com

nilukush:
@robbyrussell Changes add support for doing web search in website, thewirecutter.com . Tested locally by replacing zsh in my plugins folder.


mcornella:
The purpose of the plugin is to allow easy access to search engines, not any website that has search capabilities. The only exception is stackoverflow, and that was appropriate because most users will want to use it at some point or another.
Also, `wc` is a popular command and this alias would override it.

nilukush:
ok thanks @mcornella

PR #7681, opened 2019-03-21
BOFH-Plugin

niwiniwi:
Plugin to lighten up your sysadmin-day

### DIFF
--- a/plugins/bofh-color/README.md
+++ b/plugins/bofh-color/README.md
@@ -4,5 +4,6 @@ A bit of fun to lighten up your day. Get a Error-Message whenever you open your 
 
 Pay a visit to bofh.tips and follow @grauhut on Twitter. 
 
-Note: This is the no-lolcat Version. use bofh-color if you want crazy colors. 
+Note: This is the version that requires lolcat. Each message is displayed with random rainbows.
+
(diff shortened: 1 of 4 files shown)


mcornella:
I won't pull in this plugin due to lack of interest, but here's a review for it so that you can improve future zsh code.

You can replace the whole thing with aliases or a [function](http://zsh.sourceforge.net/Doc/Release/Functions.html), so you don't need to call back to the plugin file. Also, the command for curl already sends the output to stdout by default, you don't need to pass `-o -`. Finally, you don't need two plugins, you can use only one and have two functions.

For example, you'd put this just in one `bofh.plugin.zsh` file:
```zsh
function bofh {
  curl https://bofh.tips 2>/dev/null
}

function bofh-color {
  bofh | lolcat
}

bofh
```

If you'd wanted to allow calling either the plain or the `lolcat` version, you could read a variable used as a setting, for example `$BOFH_COLOR=true`. You'd use that like so:
```zsh
if [[ "$BOFH_COLOR" = true ]]; then
  alias bofh="curl https://bofh.tips 2>/dev/null | lolcat"
else
  alias bofh="curl https://bofh.tips 2>/dev/null"
fi

bofh
```

PR #7700, opened 2019-03-25
ubuntu: plugin cleanup and rename of ag alias to age

mcornella:
- Clean up plugin and README
- Rename ag to age to avoid conflict with The Silver Searcher:
  * Fixes #3866
  * Closes #5752


jjangga0214:
But why should it be ag"**e**"?
Is it because you think it would rarely conflict with existing aliases?

mcornella:
For `apt-get` which is a possible value of the alias. This was done also for the debian plugin.

PR #7705, opened 2019-03-26
remove caractere ','

miqueiaspenha:


mcornella:
That would mean encouraging bad syntax. By getting the error, the user learns how to properly declarate arrays in shell syntax.

PR #7706, opened 2019-03-26
zsh_reload: Respect `$ZDOTDIR` when searching for `.zshrc`

alexander255:
Title says it all. Related to #7052, had originally planned on proposing this after that one is merged, but maybe this one doesn't need as much scrutiny.


mcornella:
Sorry for the delay

PR #7707, opened 2019-03-26
Some fixes for plugins when running the shell with non-default options

ntninja:
Fixes:

  * `z`: Crashes when shell is run with `no_unset` because it tests whether variables are defined using `[ -n "${varname}" ]`
  * `lib/git`: Crashes on detecting working tree state because it expects default expansion (empty string) for some variables without using `${varname:-}`
  * `dirhistory`: Crashes in weird ways when `ksh_arrays` is set


mcornella:
This should be posted upstream: https://github.com/rupa/z/.

PR #7717, opened 2019-03-31
react-native: use new iOS simulator names

ialmoqren:
The simulator names changed after updating to Xcode 10.2 (iOS 12.2 SDK)
I don't know how to prevent this change from affecting those who didn't update yet, but it's worth noting that updating to this version is now required in order to submit iOS apps to the App Store


ialmoqren:
@mcornella sorry I'm not very familiar with `Status: testers needed`,  is there something I can do to get this PR merged or do we need to wait for someone to test it, what if no one volunteered to do so

Thanks for your effort

mcornella:
It's definitely documented that this change in naming has happened ([one instance](https://medium.com/@hacknicity/fixing-xcodes-iphone-xs-xs-max-and-xr-simulator-names-and-ios-versions-f5d7044dc00c)).

What I'm worried about is how this alias change will play out on users that haven't updated Xcode, and see how we can mitigate these side effects. Maybe we could even check the output of `xcrun simctl list devices` to see if these new simulator names are there?

ialmoqren:
Would be awesome, but honestly this is currently beyond my skills level, I don't know how should we proceed on it, we might need to wait until someone contribute in fixing it the same way you mentioned

esthor:
I can try to look at it this week. Should be able to read the device list output, select which version of the naming scheme to use and run the simulators. That will make this plugin significantly more complex, but it could also open up other cool possibilities, like being able to prompt a user if they have multiple iOS SDK versioned sims for the same device, to select which one they want to run, or handle other conditions. 👍

mcornella:
In the meantime and just for your convenience you can override these aliases in your zshrc file (after OMZ is sourced)

```zsh
alias rniosxs='react-native run-ios --simulator "iPhone Xs"'
alias rniosxsm='react-native run-ios --simulator "iPhone Xs Max"'
alias rniosxr='react-native run-ios --simulator "iPhone Xʀ"'
```

Issue #5249, opened 2016-07-24
af-magic improvment


leoheck:
Where can I ask a feature for `af-magic` theme?

Considering it from here, [af-magic.zsh-theme](https://github.com/robbyrussell/oh-my-zsh/blob/master/themes/af-magic.zsh-theme), I would like to know if its is possible to replace the **line 9** by this:

``` bash
SPACER=$(printf '-%.0s' {1..$(tput cols)})
PROMPT='$FG[241]$SPACER%{$reset_color%}
```

This will create a resizable separator line instead of being using a fixed one.
Thanks!

okapia:
A simpler way to create the line is:

```
SPACER="${(l.COLUMNS..-.)}"
```

mcornella:
Implemented in #7739. Go +1 it if it works for you.

PR #7739, opened 2019-04-06
af-magic: make separator window-size-aware

mcornella:
As asked for in #5249.
cc @leoheck


mcornella:
No interest in this.

matthewarmand:
Any chance of reviving this? I implemented a fix for this locally but this solution is much more elegant than mine.

mcornella:
There you go.

matthewarmand:
Thanks!

PR #7742, opened 2019-04-07
edit colorize plugin: add `-f terminal` option

jjangga0214:
The option (`pygmentize -f terminal <...>`) lets pygments use terminal color
scheme. Otherwise, it would use its default colors, which might be
unbalanced with that of terminal (e.g. not harmonious with background color of terminal).


robbyrussell:
I went ahead and got this nicely merged in for us. Also, colorize is a great plugin!

https://twitter.com/robbyrussell/status/1208260529405349889

PR #7746, opened 2019-04-08
fix #7745 by renaming plugin ubuntu to apt

### LINKED ISSUE #7745: isn't plugin "ubuntu" poorly named?
Currently, most features of plugin **ubuntu** are in fact not for ubuntu itself but for "apt".

Apt is not ubuntu-specific. It's used on debian-based distro, and even other distro not based on debian.  

In my opinion, name of plugin ought to imply what it actually does and serves.

So, i think, first, it should be renamed, and second, I'd like to suggest renaming it to **apt**. Though not 100% of this plugin is direct concern of apt (e.g. `kerndeb`, but even it is also indirectly related with apt), its almost entire purpose is for using apt with ease. So this rename might make sense. (Or, we can separate ubuntu-related things and apt-related things each for different plugins, and organize them in future.)

And yes, this change forces users to edit .zshrc. However, this inconvenience might be resolved in a soft and adaptive ways, such as notification after update or fall back to renamed plugin with warning(guide message to change) for a while.

jjangga0214:
Renaming **ubuntu** plugin to **apt** is just one candidate of possible suggestions.
If agreed with #7745 but having a better naming idea, then please make PR that references #7745.


mcornella:
See https://github.com/robbyrussell/oh-my-zsh/issues/7745#issuecomment-481428471

Issue #6346, opened 2017-10-09
RPROMPT evaluation plugin vs. theme


deathjest3r:
This is more of a question then an issue. I have developed a very simple plugin that shows you on the right side ${RPROMT} whether you are in a chroot environment or not. However when I load a theme that also uses RPROMPT then my status does not show up. Is a plugin that messes with "themeing" even wanted? Or should this strictly be separated?

For testing I used the theme 'alanpeabody'. When I modify the RPROMPT from 
`RPROMPT="${return_code} ${git_branch} ${rvm}"` to
`RPROMPT="${RPROMPT} ${return_code} ${git_branch} ${rvm}"` both informations are shown.

But what's the common style guidelines for such a case? Is it ok to have a plugin that maybe does not work if a theme already uses RPROMPT?

mcornella:
The problem here is that themes load after plugins, so any prompt modification done in a plugin is probably bound to be overriden by a theme if the theme doesn't have code like yours. You can add your RPROMPT modification after the theme has loaded, in the zshrc file, but we should probably fix themes to only append modifications to RPROMPT instead.

programmer04:
@mcornella thanks for the advice. I fixed it for theme fishy. But over 40 themes have this type of issue. I do not fix them, because the color of aws message doesn't fit some themes style (probably theme should set color), so I don't want to mess up someone's theme.

PR #7747, opened 2019-04-08
Fix showing aws prompt out of the box for 'fishy' theme

programmer04:
Fix issue related to #7615
After the update, aws prompt (which should be visible out of the box) disappears when a user uses a theme fishy, because of fact that plugins are loaded before themes.
This line:
```
/Users/jw/.oh-my-zsh/plugins/aws/aws.plugin.zsh:47> export RPROMPT='$(aws_prompt_info)${return_status}$(git_prompt_info)$(git_prompt_status)%{$reset_color%}'
```
is overwritten during startup by below line:
```
/Users/jw/.oh-my-zsh/themes/fishy.zsh-theme:17> RPROMPT='${return_status}$(git_prompt_info)$(git_prompt_status)%{$reset_color%}'
```
This pull request fixes issue with not showing aws prompt in theme fishy, by appending RPROMPT in theme fishy instead overwriting (way mentioned in #6346) and adds information about it in documentation.

### DIFF
--- a/plugins/aws/README.md
+++ b/plugins/aws/README.md
@@ -24,7 +24,9 @@ plugins=(... aws)
 
 ## Plugin options
 
-* Set `SHOW_AWS_PROMPT=false` in your zshrc file if you want to prevent the plugin from modifying your RPROMPT. Some plugins overwrite the value of RPROMPT instead of appending to it, so they need to be fixed to see a default aws plugin RPROMPT message (for example theme fishy is fixed).
+* Set `SHOW_AWS_PROMPT=false` in your zshrc file if you want to prevent the plugin from modifying your RPROMPT.
+  Some themes might overwrite the value of RPROMPT instead of appending to it, so they'll need to be fixed to
+  see the AWS profile prompt.
 
 ## Theme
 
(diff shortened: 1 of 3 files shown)


mcornella:
The best approach is to fix the themes that override RPROMPT to append their modifications instead.

programmer04:
@mcornella thanks for the advice. I fixed it for theme fishy. But over 40 themes have this type of issue. I do not fix them, because the color of aws message doesn't fit some themes style (probably theme should set color), so I don't want to mess up someone's theme.

programmer04:
But this way of fixing it provides another issue e.g.
executing a command like that
Screenshot 2019-04-09 at 10 56 51
gives this result in a newly spawned console, because it only appends to RPROMPT
Screenshot 2019-04-09 at 10 57 32
So what should be a generic way of fixing such problem and developing plugins, which change behavior of PROMPT or RPROMPT?
My first solution?

mcornella:
Are you per chance `export`ing `RPROMPT`? Prompt modifications shouldn't pass to children shells like that...

programmer04:
@mcornella this issue occurred, because of aws plugin exported `RPROMPT` variable. I fixed it.

mcornella:
Oh right. Well spotted.

PR #7748, opened 2019-04-08
Rn plugin updates

esthor:
Apple changed the simulator names in 12.2 for

- "iPhone XS" -> "iPhone Xs"
- "iPhone XS Max" -> "iPhone Xs Max"
- "iPhone Xr" -> "iPhone Xr".

This plugin's commands to run those won't work in 12.2 SDK because the sim names are case-sensitive. You will get a `Could not find iPhone XS simulator` or similar error.


mcornella:
Duplicate of #7717.

PR #7755, opened 2019-04-10
Avoid awk in the Git version check

brainplot:
In order to check that the installed version of Git is older than 2.13, a mixture of is-at-least and pattern matching was used; instead of an external program like `awk`, which makes things slower.


mcornella:
In this case I don't think it matters, it's just one call and awk is forked in parallel to the git version command, so there is no noticeable difference. It also makes the code less readable.

PR #7762, opened 2019-04-11
python: Fix and extend pyclean

michael-k:
Deleting `.mypy_cache` did not work, because the directories were not empty.  They contain JSON files instead of compiled byte-code.  (ref PR robbyrussell/oh-my-zsh#7151)

This extends the cleanup by also getting rid of `.pytest_cache` directories.

`-O1` takes care of reordering expressions in order to perform the test based on `-name` first.


michael-k:
> can you remove that?

done

Issue #5569, opened 2016-10-24
How to disable magic plugins


Zert:
How to disable bracketed-paste-magic, url-quote-magic and whatever-magic plugins?

apjanke:
Just don't include them from the `$plugins` list in in your `~/.zshrc`. Only plugins explicitly requested are loaded.

Zert:
This is not oh-my-zsh plugin, it's code from `lib/misc.zsh`, and I don't know how to exclude this file from loading.

apjanke:
Ah, sorry.

OMZ doesn't provide a setting to turn those off. But you can override the provided `lib/misc.zsh` with a customized one by copying it to your `$ZSH_CUSTOM/lib` directory (creating one if necessary) and editing it.

Zert:
May be it worth to move this code to `plugin/` and enable it explicitly?

mlibbey:
Perhaps similar to 5499?

tasmo:
According to the [Wiki](https://github.com/robbyrussell/oh-my-zsh/wiki/Design#customization) it is possible to overwrite internal `lib/*.zsh` files through creating a corresponding `$ZSH_CUSTOM/lib/*.zsh`:

> It's not documented in the _Customization_ page, but `$ZSH_CUSTOM/lib/*.zsh` do override the corresponding internals lib files. If a custom one is present, it is sourced instead of the one in the distribution.

untoreh:
`zle -N bracketed-paste bracketed-paste-magic` causes some serious delays copy-pasting ...an easier option to turn it off might be useful...

AaronFlower:
First, edit the `misc.zsh` file.
```bash
vim $ZSH/lib/misc.zsh
```
Second,  Comment the following code.
```diff
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -5,11 +5,11 @@ if [[ $ZSH_VERSION != 5.1.1 ]]; then
   for d in $fpath; do
        if [[ -e "$d/url-quote-magic" ]]; then
                if is-at-least 5.1; then
-                       autoload -Uz bracketed-paste-magic
-                       zle -N bracketed-paste bracketed-paste-magic
+                       # autoload -Uz bracketed-paste-magic
+                       # zle -N bracketed-paste bracketed-paste-magic
                fi
-               autoload -Uz url-quote-magic
-               zle -N self-insert url-quote-magic
+               # autoload -Uz url-quote-magic
[...]

PR #7763, opened 2019-04-12
#5569 ENV to disable magic

galan:
Adding an environment variable to disable the magic loader instead of manipulating files as workaround for #5569


mcornella:
I submitted a version of this (486fa101) adding you as contributor.

PR #7764, opened 2019-04-13
init: state common reasons for not finding plugins

SomeDer:
Can state if commas are used in plugins array and can state if a plugin is sourced manually. 
#### Example:
~/.zshrc
```
export ZSH="/home/me/.oh-my-zsh"

ZSH_THEME="gallois"

plugins=(
  systemd,
  colored-man-pages
  zsh-syntax-highlighting
)

source $ZSH/oh-my-zsh.sh

source ~/zsh-syntax-highlighting/zsh-syntax-highlighting.plugin.zsh
```
output
```
[oh-my-zsh] the following plugins could not be found:
  systemd, 
  zsh-syntax-highlighting - sourced manually
zsh arrays use spaces, not commas to separate entries
some plugins were sourced manually so don't need to be in the plugins array
```

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -58,7 +58,7 @@ for plugin ($plugins); do
     if echo $plugin | grep ',$' > /dev/null; then
       commasInPluginsArray=true
     fi
-    if grep "source .*/$plugin.zsh" ~/.zshrc > /dev/null; then
+    if grep "^[^#]*source .*/$plugin.zsh" ~/.zshrc > /dev/null; then
       sourced="- sourced manually"
       pluginsSourcedManually=true
     else 
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -47,21 +47,39 @@ is_plugin() {
 
 # Add all defined plugins to fpath. This must be done
 # before running compinit.
+
+pluginsNotFound=()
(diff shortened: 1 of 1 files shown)


mcornella:
Comma separators in the plugins array is already documented to not work: [in the README](https://github.com/robbyrussell/oh-my-zsh#enabling-plugins), in [the template zshrc file](https://github.com/robbyrussell/oh-my-zsh/blob/08da19e3fd27acf5085e207ee7da411e6a5eb96a/templates/zshrc.zsh-template#L60-L63), and in various issues throughout the repository. The only other place were we can document this is in the FAQ.

As per the external plugins, each of them have documented how to install them and enable them in the zshrc file, either within OMZ or out of it.

I won't overcomplicate the init script for something users have received plenty of instructions for.

SomeDer:
But the [first commit](https://github.com/robbyrussell/oh-my-zsh/pull/7764/commits/5f52b063123c7cddd37ec395bbf6c04c8528a557) wasn't complicated. I added the [next commit](https://github.com/robbyrussell/oh-my-zsh/pull/7764/commits/46a798427a63e1a14d20e923bac54240e7357a1b) to avoid repeating the same message multiple times, but I guess it wasn't worth the complexity. There has also been a lot of issues about this, despite all of the instructions.

Issue #5779, opened 2017-01-13
emacs plugin doesn't work correctly on mac


seagle0128:
On my macOS Sierra with zsh 5.2/5.3, emacs and e always create a new frame.

in emacsclient.sh 
```shell
 if [ "$(echo "$frameslist" | sed -n '$=')" -ge 2 ] ;then
        # prevent creating another X frame if there is at least one present.
        emacsclient --alternate-editor "" "$@"
    else
        # Create one if there is no X window yet.
        emacsclient --alternate-editor "" --create-frame "$@"
    fi
```

I think the first condition is incorrect. `>=1` is better. Is it as designed or an issue?
```shell
if [ "$(echo "$frameslist" | sed -n '$=')" -ge 1 ] ;then
```

seagle0128:
Anyone can merge this request?

mcornella:
This is a simple change, can you submit a PR?

akinnane:
@seagle0128 are you able to run `emacsclient --eval '(frame-list)'` with no emacs frames created and then with a frame created. Can you add the output of those commands to this ticket, please?

seagle0128:
- no frame created
```shell
$ emacsclient --eval '(frame-list)'
emacsclient: can't find socket; have you started the server?
To start the server in Emacs, type "M-x server-start".
emacsclient: No socket or alternate editor.  Please use:

	--socket-name
	--server-file      (or environment variable EMACS_SERVER_FILE)
	--alternate-editor (or environment variable ALTERNATE_EDITOR)
```

- With a frame created

```shell
$ emacsclient --eval '(frame-list)'
(#<frame Centaur Emacs - ~/.emacs.d/custom.el 0x10106f5b0>)
```

mcornella:
Hey, I've been revisiting this trying to find a working solution across all platforms, could you tell me what the code below outputs both for:
1. `emacs --daemon` not running
2. daemon running but no frames visible
3. one visible frame

```
emacsclient -n -e '(frame-list-z-order)'
```

mcornella:
I've submitted a PR which theoretically fixes this problem for both situations (one residual frame or none are available). Please test #10227 by running `omz pr test 10227`. It should work when requiring a graphical frame (`--no-wait`) or a tty frame (`-t`/`-nw`/`--tty`).
Whether it works or it doesn't, please comment in #10227.

PR #7765, opened 2019-04-13
Revert cab1ac6e682038b2ceca9d16f8c458c0e79149ca

akinnane:
Revert cab1ac6e682038b2ceca9d16f8c458c0e79149ca

The change from the above commit breaks opening a new emacs frame if
there isn't a existing frame open.

When emacs is running in daemon mode, there will always be a frame
associated with the daemon, even if there are no visible frames.

The lisp function `frame-list`, will always return 1 frame.
```
emacsclient --eval '(frame-list)'
(#<frame F1 0xa3c680>)
```

When an real frame is open, two frames are reported by the `frame-list`
function.
```
emacsclient --eval '(frame-list)'
(#<frame F1 0xa3c680> #<frame emacs Prelude - ~/oh-my-zsh/plugins/emacs/emacsclient.sh 0xf50e10>)
```

See:
https://emacs.stackexchange.com/questions/44537/extra-frame-in-visible-frame-list-when-started-in-daemon-mode-is-causing-prob
https://emacs.stackexchange.com/questions/18859/dont-let-the-daemon-frame-make-a-file-visible

### DIFF
--- a/plugins/emacs/emacsclient.sh
+++ b/plugins/emacs/emacsclient.sh
@@ -5,7 +5,7 @@ _emacsfun()
     # get list of emacs frames.
     frameslist=`emacsclient --alternate-editor '' --eval '(frame-list)' 2>/dev/null | egrep -o '(frame)+'`
 
-    if [ "$(echo "$frameslist" | sed -n '$=')" -ge 1 ] ;then
+    if [ "$(echo "$frameslist" | sed -n '$=')" -ge 2 ] ;then
         # prevent creating another X frame if there is at least one present.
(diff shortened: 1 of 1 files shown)


mcornella:
Is there any way we can detect if there is 1 visible frame?

mcornella:
I'm talking about solving #5779 and #5405 of course.

akinnane:
I would expect Emacs on OSX to behave in the same way: a single frame for the daemon and another frame is a window had been created. We would need more information from the original reporters from #5779 and #5405. I've asked for extra details on both tickets.

PR #7768, opened 2019-04-14
New plugin: alias finder

SomeDer:
This plugin searches the defined aliases before each command and outputs any that match that command. This makes learning new aliases easier.


plugins/alias-finder/alias-finder.plugin.zsh
```diff
@@ -0,0 +1,46 @@
...
+      if [[ $exact = true || $longer = true ]]; then
+        break
+      else
+        cmd=$(sed -E 's/ {0,1}[^ ]*$//' <<< $cmd) # deletes last word
+      fi
+    done
+  fi
+}
+
+preexec() {
```
mcornella: This should be it's own function and be added to `preexec_functions` like so:
```
preexec_functions+=(preexec_alias-finder)
```

Otherwise you might delete an already defined `preexec` function.

plugins/alias-finder/alias-finder.plugin.zsh
```diff
@@ -0,0 +1,46 @@
+alias-finder() {
+  local cmd=""
+  local exact=""
+  local longer=""
+  for i in $@; do
+    case $i in
+      -e|--exact) local exact=true;;
```
mcornella: You don't need to redefine a variable as local, put all your variable definitions at the top and then don't use `local` again. You can also define multiple variables in the same line:
```zsh
local cmd="" exact="" longer=""
```

plugins/alias-finder/alias-finder.plugin.zsh
```diff
@@ -0,0 +1,46 @@
...
+  cmd=$(sed 's/[].\|$(){}?+*^[]/\\&/g' <<< $cmd)
+  if [[ $(wc -l <<< $cmd) == 1 ]]; then
+    while [[ $cmd != "" ]]; do
+      if [[ $longer = true ]]; then
+        local wordStart="'{0,1}"
+      else
+        local wordEnd="$"
+        local multiWordEnd="'$"
+      fi
+      if echo $cmd | grep '^[^ ]*$' > /dev/null; then
```
mcornella: What is this line supposed to do?

plugins/alias-finder/README.md
```diff
@@ -0,0 +1,48 @@
...
+This plugin searches the defined aliases and outputs any that match the command inputted. This makes learning new aliases easier.
+
+## Enabling
+To use it, add `alias-finder` to the `plugins` array of your zshrc file:
+```
+plugins=(... alias-finder)
+```
+
+## Usage
+To see if there is an alias defined for the command, pass it as an argument to `alias-finder`. This can also run automatically before each command you input - add `export ZSH_ALIAS_FINDER_AUTOMATIC=true` to your zshrc if you want this.
```
mcornella: You don't need the `export`.

plugins/alias-finder/README.md
```diff
@@ -0,0 +1,48 @@
+# alias-finder plugin
+
+## Description
+This plugin searches the defined aliases and outputs any that match the command inputted. This makes learning new aliases easier.
+
+## Enabling
```
mcornella: The Description and Enabling sections are normally merged together, without any header:
```markdown
# plugin title

Words of the description

Instructions to enable
```

plugins/alias-finder/alias-finder.plugin.zsh
SomeDer: It checks if `cmd` is a single word (i.e. has no spaces). I have simplified it at https://github.com/robbyrussell/oh-my-zsh/pull/7768/commits/58f7dfd4cb204d644efbdb9135af5d61f71c3c91

mcornella:
👍

veekas:
Hey all, was this plugin confirmed to work before merging? (Asking because of @mcornella's comment about not yet having tested it yet.) My other plugins are working, but `alias-finder` doesn't seem to return anything in my environment. I wanted to confirm that this works for others before investigating whether my setup is faulty. Thanks!

mcornella:
It does work on my end:
image

Can you run `which alias-finder` to see your function definition?

ritzau:
> Hey all, was this plugin confirmed to work before merging? (Asking because of @mcornella's comment about not yet having tested it yet.) My other plugins are working, but `alias-finder` doesn't seem to return anything in my environment. I wanted to confirm that this works for others before investigating whether my setup is faulty. Thanks!

This missing result is most likely due to the fact that wc -l returns output padded by space on macOS Catalina and that causes the == comparison to fail. There is a PR with a fix but it has been closed. #8251

winnochan:
I used this plugin in manjaro linux with zsh version 5.8, and I found value $1 in hook function 'preexec_alias-finder' is empty. I also found use $2 to get command value is ok but I don't know why. I changed code from
`preexec_alias-finder() {
  if [[ $ZSH_ALIAS_FINDER_AUTOMATIC = true ]]; then
    alias-finder $1
  fi
}`
to 
`preexec_alias-finder() {
  if [[ $ZSH_ALIAS_FINDER_AUTOMATIC = true ]]; then
    alias-finder -e $2
  fi
}`
and everything is ok. I think use the -e parameter as default in zsh hook function is better.

mcornella:
This is the documentation for [preexec hooks](http://zsh.sourceforge.net/Doc/Release/Functions.html#index-preexec):
> Executed just after a command has been read and is about to be executed. **If the history mechanism is active (regardless of whether the line was discarded from the history buffer), the string that the user typed is passed as the first argument, otherwise it is an empty string.** The actual command that will be executed (including expanded aliases) is passed in two different forms: the second argument is a single-line, size-limited version of the command (with things like function bodies elided); the third argument contains the full text that is being executed.

This means you have history disabled. Find a way to enable it back, I'm not sure even how to disable it.

Your proposed version would always show an alias suggestion, since `$2` is the command after alias expansion. That means, if you're running `ls` aliased to `ls --color=auto`, the alias-finder function will receive `ls --color=auto` despite having run the alias instead.

winnochan:
@mcornella Yes, I think `alias-finder -e $1` is better, but I have searched in google and didn't get any answer to active the history mechanism.

winnochan:
@mcornella  I have found the problem, when I remove the plugin `per-directory-history` from oh-my-zsh config, and everything is ok, but I don't know why, Maybe these two plugins are conflict.

winnochan:
@mcornella @SomeDer  Do you think these code as default hook better?
`
preexec_alias-finder() {
  if [[ $ZSH_ALIAS_FINDER_AUTOMATIC = true && "$1" = "$2" ]]; then
    alias-finder -e $1
  fi
}
`

mcornella:
> @mcornella I have found the problem, when I remove the plugin `per-directory-history` from oh-my-zsh config, and everything is ok, but I don't know why, Maybe these two plugins are conflict.

The `per-directory-history` plugin must temporarily disable history and that's why `$1` is empty, but I don't know how it does that. Try opening an issue in https://github.com/jimhester/per-directory-history and asking for @jimhester's support.

> @mcornella @SomeDer Do you think these code as default hook better?
> `preexec_alias-finder() { if [[ $ZSH_ALIAS_FINDER_AUTOMATIC = true && "$1" = "$2" ]]; then alias-finder -e $1 fi }`

This change isn't necessary and it also won't fix your issue when using per-directory-history plugin.

jimhester:
https://github.com/jimhester/per-directory-history/issues/10 may be relevant, but I don't really remember how this all interacts, perhaps these two plugins are simply incompatible.

PR #7769, opened 2019-04-15
bwana: change function name to bwana

### LINKED ISSUE #5732: Bwana plugin blocks access to standard man
[Bwana plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/bwana/bwana.plugin.zsh) should have a different name for the function. Currently it's `man` which (after enabling the plugin) make using standard `man` impossible.
I would suggest changing it to `bwana`.
/cc @avit

dantehemerson:
Fix #5732
Change name function from `man` to `bwana`.


mcornella:
Thanks!

PR #7774, opened 2019-04-17
Terraform plugin: formatted prompt and added several aliases

nexxai:
I formatted the prompt to add "Terraform" in front of which Terraform environment is being used, because it was looking strange with both a git branch and Terraform workspace in a single prompt.  I also added several aliases to make common commands simpler.


robbyrussell:
@nexxai Would you be able to rebase this for us given that we've had changes to the Terraform plugin already merged in?

nexxai:
Sorry this PR can be cancelled.  It was submitted 2 years ago and I've moved off using omz since.

PR #7780, opened 2019-04-18
Remove a dot from twitter url in installer

nopticon:
When oh-my-zsh is installed, the welcome message has twitter url, when click it redirects to non existent profile.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -103,7 +103,7 @@ main() {
   echo ''
   echo 'Please look over the ~/.zshrc file to select plugins, themes, and options.'
   echo ''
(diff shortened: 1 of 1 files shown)


mcornella:
That's weird, in Gnome Terminal the URL highlighted doesn't include the dot. What terminal emulator are you using?

Also, the URL below should be similarly fixed. Thanks

nopticon:
It happened in /bin/bash and /bin/zsh with the same behavior (Mac 10.14).

Screen Shot 2019-04-18 at 3 07 00 PM

mcornella:
Ok thanks!

PR #7781, opened 2019-04-19
rid emacs tramp hang

RezoChiang:
https://www.emacswiki.org/emacs/TrampMode#toc8


mcornella:
You can put that in your own zshrc file.

PR #7786, opened 2019-04-22
added a new alias to the golang plugin

RayofLightz:
I added a new alias to the golang plugin to fetch all the dependencies for a project and then build that project.


mcornella:
Can you rebase and rename the alias to `gogwd`? It seems more appropriate given that the alias is still to `go get`.

PR #7789, opened 2019-04-23
mktemp --template doesnt work in macosx.  mktemp -t works in macosx AND linux

danhimalplanet:
"mktemp --template" fails in macosx

"mktemp -t" works in macosx and linux.

this bash function should use "mktemp -t" instead.


mcornella:
This works on my end, @fzuviria thoughts?

PR #7790, opened 2019-04-23
Add non-interactive option to install script

subtlepseudonym:
The install script drops into `zsh` when its finished installing. For this PR, that behavior has been retained as the default, but if the `NO_INTERACTIVE` environmental variable is set, it will exit rather than entering `zsh`.

The purpose of this change is to make the script more amenable to use inside other scripts.


subtlepseudonym:
This change, in its current state, is pretty unwieldy considering the recommended method for retrieving the install script. I've written another change to just check a `NO_INTERACTIVE` env variable, which I'll test / push tomorrow.

mcornella:
#5169 has pulled this PR with a little change: either by setting the environment variable `RUNZSH=no` or passing the flag `--unattended`, which also skips the change of default shell. It has a bunch of other fixes, and it's now ready to test.

Thanks!

mcornella:
Closed by #5169.

PR #7792, opened 2019-04-23
react-native: add simulator aliases according to latest Xcode update

MuhmdRaouf:
Update new simulator names as the new XCode update have update the simulators names and add new ones


mcornella:
Can you rebase and add the new aliases to the README? The Xs, XR and Xs max ones have been updated in #7717.

esthor:
still the problem exists of what to do for people running other version of the iOS SDK before the latest...are we ok with breaking the ability to run Xs, Xr, Xs simulators for those users?

MuhmdRaouf:
> Can you rebase and add the new aliases to the README? The Xs, XR and Xs max ones have been updated in #7717.

i have rebased and updated the readme file, i guess the PR now ready to merge.

mcornella:
The table syntax is messed up now, you don't need to add separators for each section, just have the heading as it was before.

MuhmdRaouf:
Oh sorry i didn't know that i would miss up the table, i have it as it was now, sorry about that.

PR #7794, opened 2019-04-24
docker-compose: add dcip function

maddingo:
function that shows the IP address of a running docker-compose service
Cannot use an alias, since the function takes an argument


maddingo:
I rather put this in a feature branch of my repo

PR #7796, opened 2019-04-24
docker-compose: add function dcip that shows the IP address of a running docker-compose service

maddingo:
…

### DIFF
--- a/plugins/docker-compose/README.md
+++ b/plugins/docker-compose/README.md
@@ -27,3 +27,9 @@ plugins=(... docker-compose)
 | dclf      | `docker-compose logs -f` | Show logs and follow output                                      |
 | dcpull    | `docker-compose pull`    | Pull image of a service                                          |
 | dcstart   | `docker-compose start`   | Start a container                                                |
+
+## Functions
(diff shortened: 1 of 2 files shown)


robbyrussell:
Hi @maddingo, apologies... this one slipped through the cracks for a very long time and nobody ever replied to you.

Still interested in it? Two things would need sorting if so. It conflicts now, and the plugin was reworked for Compose v2 since you opened this, so the `docker-compose ps -q` call would want to go through `$dccmd` like the rest of the aliases do.

One other thought while you're in there: with several services running, the output is a bare list of addresses with nothing saying which is which. Might be worth printing the service name alongside.

No rush and no hard feelings if you'd rather not... just let us know either way and we'll close it out.

maddingo:
I currently don't have the time to fix this. I also archived my fork 6 years ago. If you want to use the code in any way, please do. Otherwise, I am not insulted if you just throw it away :-)

robbyrussell:
Thanks for coming back to say so, and sorry it sat this long. Closing it on your word.

PR #7804, opened 2019-04-26
Safety fix and speed-ups for lib/git.zsh prompt functions

rwe:
My prompt was running slowly, even on non-git directories.
This PR includes three basic fixes:
  * Skips processing `git status` output if the command failed
  * Reduces redundant grep invocations, and speeds them up via the `-q` flag
  * Runs all git commands with `GIT_OPTIONAL_LOCKS=0` to avoid unnecessary (and possibly-unsafe) filesystem activity.

Additional info in the individual commit messages.


erydo:
If someone would be able to look at this, I'd appreciate it. Though I realize there's a large backlog of PRs. The changes here are risk-averse and furthermore add some safety to the execution of `git`. I've been running with them for several months.

erydo:
This PR has been open for nearly a year now. Any chance of it being merged in?

lib/git.zsh
```diff
@@ -127,62 +138,50 @@ function git_prompt_remote() {
...
 # Get the status of the working tree
 function git_prompt_status() {
+  emulate -L zsh
+
   local INDEX STATUS
-  INDEX=$(command git status --porcelain -b 2> /dev/null)
+  INDEX=$(__git_prompt_git status --porcelain -b 2> /dev/null) || return 0
   STATUS=""
-  if $(echo "$INDEX" | command grep -E '^\?\? ' &> /dev/null); then
+  if [[ "${INDEX}" =~ $'(^|\n)\\?\\? ' ]]; then
```
mcornella, line 157: This is wrong, it should be `$'(^|\n)?? '`.

erydo: No, it's not. This is a regex. You can verify with an example status line:
```
INDEX='?? foo'; [[ "${INDEX}" =~ $'(^|\n)\\?\\? ' ]];! echo $?
```
> `0`

Versus the suggestion:
```
INDEX='?? foo'; [[ "${INDEX}" =~ $'(^|\n)?? ' ]];! echo $?
```
> `zsh: failed to compile regex: repetition-operator operand invalid`

I'm very sure that these changes are correct. I've been running with them for a long time.

PR #7805, opened 2019-04-26
Remove extra newline from prompt

arush15june:
There is an extra newline caused by how the prompt format string is written, removed the extra newline for correct prompt output.


mcornella:
That's intended behavior (see the avit theme screenshot in the wiki). If you want to change that you can modify it and put it in the custom folder (see Customization section in the wiki).

PR #7813, opened 2019-04-29
Remove dirty git check from Agnoster theme's prompt on large git repo

hoanghuynh:
Right now Agnoster uses `parse_git_dirty` to check if a git repo is dirty. On a large git repo, this command may takes several seconds to finish, thus slowing the prompt for quite some time.

This PR adds support for a new git config option that's called `oh-my-zsh.hide-dirty`. You can set it to `1` to hide dirty status on your git repo. The `git` command is `git config --add oh-my-zsh.hide-dirty 1`.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -108,7 +108,11 @@ prompt_git() {
 
   if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
     repo_path=$(git rev-parse --git-dir 2>/dev/null)
-    dirty=$(parse_git_dirty)
+    if [[ "$(git config --get oh-my-zsh.hide-status 2>/dev/null)" = 1 ]]; then
+      dirty=1
+    else
+      dirty=$(parse_git_dirty)
+    fi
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and apologies it sat for so long without a reply.

Good news is there's already a switch for exactly this... `git config --add oh-my-zsh.hide-dirty 1` makes `parse_git_dirty` return without shelling out to `git status`, and agnoster picks that up for free. Set it per-repo in the big ones and the prompt stops stalling.

Worth mentioning too: the patch checks `hide-status` rather than `hide-dirty` and sets `dirty=1`, so it would end up pinning the dirty segment on instead of hiding it.

Closing this one out, but thanks for digging into the slowness.

robbyrussell:
Following up on this one: what I told you above was only half right, and the half that was wrong is the half you cared about.

`hide-dirty` does stop `parse_git_dirty` running `git status`, but agnoster then sets `check-for-changes true` and calls `vcs_info` regardless, and that runs a full worktree `git diff` on every render. So on a large repo the stall you were seeing largely stays. Sorry for sending you off with a fix that doesn't do what I said it does.

The real fix is the one in #8930, which skips that scan when there's nothing to report. It's with its author now. I'm reopening this so it stays visible until that lands, rather than leaving it closed on bad information.

PR #7814, opened 2019-04-29
Properly fix `ag` alias/completion conflict for the `debian` plugin.

haaspors:


mcornella:
Fixed in #7698. Please test and +1 that one.

PR #7821, opened 2019-05-01
Updated hub completions

israelfaria:
Simple PR to update hub completions as provided by [hub](https://github.com/github/hub).


mcornella:
:+1:

PR #7828, opened 2019-05-03
Clean up the general aliases in the kubectl plugin

j-griffith:
There's a ton of aliases being added for kubectl commands, which is
awesome, but if you're like me you only remember a handufl of them.
Really the top level aliases are usually the most valuable for me and
easier to remember.

This change adds a few missing `general` aliases, and groups the general
aliases together in the top of the doc.  This way even if "your
specific" alias is missing, all the primitives are in place to provide
the foundation and make kubectl much easier to work with.

Added aliases:
  * kg    - kubectl get
  * kgans - kubectl get $@ --all-namespaces (kga is already taken)
  * kd    - kubectl describe

### DIFF
--- a/plugins/kubectl/README.md
+++ b/plugins/kubectl/README.md
@@ -13,18 +13,21 @@ plugins=(... kubectl)
 
 | Alias   | Command                             | Description                                                                                      |
 |:--------|:------------------------------------|:-------------------------------------------------------------------------------------------------|
+|         |                                   | **General aliases**                                                                              |
 | k       | `kubectl`                           | The kubectl command                                                                              |
 | kca     | `kubectl --all-namespaces`          | The kubectl command targeting all namespaces                                                     |
 | kaf     | `kubectl apply -f`                  | Apply a YML file                                                                                 |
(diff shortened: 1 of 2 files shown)


j-griffith:
I submitted a PR for PVC support as well, then started thinking; rather than just adding every single resource as it comes up (and never remembering most of them) maybe just add some of the missing base/general aliases and that would be good enough.

robbyrussell:
Thanks for this, and sorry it sat here for so long.

You were right that `kg` and `kd` are missing... they still are. The trouble is this diff can't be applied anymore: the plugin has had a couple dozen commits since 2019 and every hunk's context has moved.

Two things I'd flag for whoever picks this up. `kgans` overlaps with the existing `kca` wrapper, which already runs a command across all namespaces. And it defines a bare function named `f`, which would land in the user's namespace and shadow anything they'd defined with that name... the plugin uses `_kca`-style private names to avoid exactly that.

We'll track `kg` and `kd` on #12524, which does the same thing against the current file. Closing this one, but the idea was yours first and I'll say so there.

PR #7830, opened 2019-05-04
frontend-search: add Angular 2 context

### LINKED ISSUE #7802: frontend-search: add Angular 2
Hi. Can you guys add the Angular2+ in the search lsit context?
AngularJS is practically deprecated lol

divbhasin:
### Problem
As described in #7802, the angularjs URL for the frontend-search zsh plugin was outdated. An update was needed to point to the new URL when the user runs "frontend angularjs ...".

### What I did
I updated the README.md of the frontend-search plugin to include the new URL for angularjs, and made the appropriate change in frontend-search.plugin.zsh to point to the new URL 'https://angular.io/?search='. I also updated the code that adds a + between the search terms when appending them to the URL to be encoded (so + now becomes %20). I did this because the new search URL for angular drops the search terms in the angular search bar with the + included, instead of a space. I verified that this does not break any of the existing searches we have.

### What this fixes
Fixes #7802

### DIFF
--- a/plugins/frontend-search/README.md
+++ b/plugins/frontend-search/README.md
@@ -32,7 +32,8 @@ Available search contexts are:
 
 | context       | URL                                                                      |
 |---------------|--------------------------------------------------------------------------|
-| angularjs     | `https://angular.io/?search=`                                             |
+| angular(>=2.0)| `https://angular.io/?search=`                                             |
+| angularjs(1.x)| `https://google.com/search?as_sitesearch=angularjs.org&as_q=`
 | aurajs        | `http://aurajs.com/api/#stq=`                                            |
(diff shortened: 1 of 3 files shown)


mcornella:
I'd keep the old one and add a new one called `angular`. Angular 1.x is still supported albeit in LTS, so there may be users that still use it. This requires adding `angular` to the list of available context in the both help messages, and also modify the `_frontend-search.sh` completion and the README.

I can help out if you don't know how to do some of these things.

+1 to the plus to `%20` change.

divbhasin:
@mcornella Thank you for the feedback. I have made the changes you suggested. Do you have any additional suggestions?

mcornella:
That was perfect, thanks for this!

PR #7849, opened 2019-05-14
kubectl plugin; added alias to get pods from all namespaces

kverstr:
adding an alias for "kubectl get pods --all-namespaces" since this is a rather commonly used kubectl command


robbyrussell:
Thanks for this, and sorry for the very long silence on it.

`kgpa` is already in the plugin... it landed in December 2019 via #8434, which added the `--all-namespaces` variants for pods along with services, ingress, configmaps, secrets, deployments, statefulsets and PVCs. So there's nothing left to merge here.

You had the right idea, someone else's PR just got there first. Appreciate you taking the time.

PR #7853, opened 2019-05-16
Add alias for fetching upstream

cueo:
Add alias `gfu='git fetch upstream'`


mcornella:
This was rejected in favor of `gfa` (`git fetch --all`): https://github.com/robbyrussell/oh-my-zsh/pull/7621#issuecomment-478162756. Thanks for contributing.

PR #7858, opened 2019-05-17
Most of community demand this changing.

### LINKED ISSUE #6385: Rails rg alias conflicts with ripgrep
Perhaps we might want to change the `rails generate` alias in the rails plugin to something other than `rg`, since that conflicts with the executable for ripgrep

mertingen:
Rails rg alias conflicts with ripgrep #6385

I've changed the "rg" alias to "rgen". With this way, it will be more readable and useful.

Fixes #6385


mcornella:
Can you change the README as well?

mertingen:
That was my bad, I wasn't aware of that. I've just updated the README file.

mcornella:
Thanks @mertingen!

PR #7860, opened 2019-05-18
CHANGE Prompt - SSH user@hostname vs LOCAL user

fr32k:
For a remote connection with ssh the user@hostname is displayed and if we are local only the username. This way it is easy to see where you are.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -86,10 +86,13 @@ prompt_end() {
 ### Prompt components
 # Each component will draw itself, and hide itself if no information needs to be shown
 
-# Context: user@hostname (who am I and where am I)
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for the patch, and sorry for the long silence.

The reason agnoster hides that segment locally is `DEFAULT_USER`... people set it specifically so their own username stops taking up prompt space at home, and still shows up over SSH. Always printing the username would undo that for everyone who opted in, so it's not a change we can make to the default.

If you'd rather see it always, you can override `prompt_context` in your `$ZSH_CUSTOM` directory and it'll survive updates.

`prompt_context` has also been rewritten since 2019, so this wouldn't apply cleanly anymore. Closing it out... thanks for taking the time.

PR #7861, opened 2019-05-18
dotenv: add support for custom env file names

### LINKED ISSUE #7579: dotenv: make it possible to define the file you would like to use
It would be good to have the ability to set the file's name  you would like to use for the env (I for example have a problem with .env file because it is not recognized by my pycharm EnvFle plugin)

arshadkazmi42:
Fixes #7579 
-

- Reads env file from exported variable `ZSH_DOTENV_FILE`.
- If file, nothing is exported to the variable it takes `.env` as default file name

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,11 +169,6 @@ You must log out and log back in to see this change.
 
 Once you open up a new terminal window, it should load zsh with Oh My Zsh's configuration.
 
-#### Settings & Configurations
-
-Custom `env` files can be configured by exporting your env file in `ZSH_DOTENV_FILE`.
-If nothing is exported to `ZSH_DOTENV_FILE` then it takes `.env` as default file name
-
 ### Installation Problems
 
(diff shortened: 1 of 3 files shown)


mcornella:
Oh, I meant to modify the [plugin README](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/dotenv/README.md). Now that I see it, you can just add the info to the Usage section.

arshadkazmi42:
> Oh, I meant to modify the [plugin README](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/dotenv/README.md). Now that I see it, you can just add the info to the Usage section. Good job on the rest though!

Thanks for all the help, couldn't have done it without your help. I still need to understand more about the repo, I will be picking up some more issues to understand the project better.

mcornella:
Ok I changed the wording a bit and added an example. Thanks for your contribution and I hope you stick around for more!

PR #7866, opened 2019-05-21
Aliases for 'git push/pull origin master'

tomrlh:
Considering the following git commands are very common:
- git push origin master
- git pull origin master

command aliases to them can be useful.


mcornella:
There is `ggp` for push and `ggl` for pull.

tomrlh:
Thank you @mcornella, I've looked for something exactly matching 'git push origin master' but in git plugin it considers the current branch

mcornella:
If the name of the branch is master it will behave exactly the same. If not, you can pass the name of the destination branch to these functions, so doing `ggl master` and `ggp master` will do it.

PR #7867, opened 2019-05-21
Updated gg and gga aliases in git plugin not to lock console when executed.

krzyczak:
Updated `gg` and `gga` aliases in [plugins/git/git.plugin.zsh](https://github.com/robbyrussell/oh-my-zsh/pull/7867/files#diff-4ad2773a26526d0e361e66139c283099) not to lock console when executed.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -94,8 +94,8 @@ alias gfo='git fetch origin'
 
 alias gfg='git ls-files | grep'
 
-alias gg='git gui citool'
-alias gga='git gui citool --amend'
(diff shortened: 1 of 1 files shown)


robbyrussell:

I'm going to pass on this one. `gg` blocking is quietly doing some work... you find out when git-gui isn't installed or fails to start, and `gg && something` still waits the way you'd expect. Backgrounding it and dropping the output would change that for everyone using the plugin, after a lot of years of it behaving the other way.

Good news is you don't need a patch: `gg &` gets you the same thing whenever you want it, and you can redefine the alias in your own config if you'd rather have it that way always.

Appreciate you taking the time.

PR #7875, opened 2019-05-24
Update colorize.plugin.zsh

rwmitchell:
use -L with less to ignore LESSOPEN, otherwise if also uses pygmentize, the colorizing escape codes are exposed.

### DIFF
--- a/plugins/colorize/colorize.plugin.zsh
+++ b/plugins/colorize/colorize.plugin.zsh
@@ -45,5 +45,5 @@ colorize_via_pygmentize_less() (
         shift 1
     done
 
-    less -f "${tmp_files[@]}"
+    less -L -f "${tmp_files[@]}"
 )


mcornella:
I think the right argument is `-R` but good point. Why did you close this?

rwmitchell:
-L or --no-lessopen
              Ignore the LESSOPEN environment variable


I have LESSOPEN set to:
echo $LESSOPEN
|~/.rc/.lessfilter %s

which calls pygmentize, gotten from: https://bitbucket.org/birkenfeld/pygments-main <https://bitbucket.org/birkenfeld/pygments-main>

Without -L, I get stuff like:
[38;5;241m38;5;136m#[38;5;241m39m[38;5;241m38;5;136minclude[38;5;241m39m [38;5;241m38;5;66m<stdlib>[38;5;241m39m

With, I get:

#include <stdlib.h>

-R also returns:

[38;5;241m38;5;136m#[38;5;241m39m[38;5;241m38;5;136minclude[38;5;241m39m [38;5;241m38;5;66m<stdlib.h>[38;5;241m39m

for my issue, -L appears correct.  I don’t have any reason to use cless, as I already get it automatically with less/LESSOPEN, but I was curious.  I’ve been sitting on the change for a couple weeks, but when oh-my-zsh wanted to upgrade, my local change blocked it.

As for why I closed it, it was my first time submitting a code change to a public GitHub repos through the web interface and I got carried away clicking buttons.

PR #7876, opened 2019-05-24
plugin: mpd (mpc + mpdscribble)

matchaxnb:
This plugin auto-starts an instance of mpdscribble.

I reserve the possibility to add useful aliases at a later point in time depending on my use of it.

Includes a once-per-session warning that mpdscribble is missing.

### DIFF
--- a/plugins/mpd/mpd.plugin.zsh
+++ b/plugins/mpd/mpd.plugin.zsh
@@ -1,0 +1,35 @@
+let _mpdscribble_missing_warned
+let _mpc_missing_warned
+alias m=mpc
+for _ in 1; do
+  pidof mpdscribble > /dev/null
+  test "$?" = 0 && break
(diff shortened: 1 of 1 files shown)


ChloeTigre:
Added alias m=mpc

robbyrussell:
Thanks for this, and apologies for the very long silence.

I'm going to pass on it. The part that decides it for me is starting `mpdscribble` from shell startup... a plugin launching a background daemon every time you open a terminal isn't something we want to take on for people. A couple of smaller things too: `pidof` doesn't exist on macOS, so that check won't do what you want there, and a single-letter alias like `m` is likely to collide with something in a user's own setup.

This is a great fit for `$ZSH_CUSTOM` though. Drop it in `~/.oh-my-zsh/custom/plugins/mpd/` and it'll load exactly the same way and survive updates.

Thanks for taking the time to write it up.

PR #7880, opened 2019-05-25
ng: fix shwordsplit option use

rachelcarmena:
I looked for an option with "word" and "split" and I found this piece of source code:

```
   setopt shwordsplit
   reply=($opts)
   unset shwordsplit
```

However, `unset`doesn't work for inverting the previous zsh option. 

This PR only replaces `unset`by `unsetopt`.

### DIFF
--- a/plugins/ng/ng.plugin.zsh
+++ b/plugins/ng/ng.plugin.zsh
@@ -1,4 +1,3 @@
-
 ng_opts='addon asset-sizes b build completion d destroy doc e2e g generate get github-pages:deploy gh-pages:deploy h help i init install lint make-this-awesome new s serve server set t test update v version -h --help'
 
 _ng_completion () {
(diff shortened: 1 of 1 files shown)


mcornella:
Better yet use `${=var}` to word-split only that variable. That way we don't need to setopt/unsetopt which could mess with the global settings. [From the docs](http://zsh.sourceforge.net/Intro/intro_12.html):

> One possible source of confusion is the fact that in zsh, the result of parameter substitution is not split into words. Thus, this will not work:
> ```
> % srcs='glob.c exec.c init.c'
> % ls $srcs
> glob.c exec.c init.c not found
> ```
> If you like the sh behaviour, zsh can accomodate you:
> ```
> % ls ${=srcs}
> exec.c  glob.c  init.c
> % setopt shwordsplit
> % ls $srcs
> exec.c  glob.c  init.c
> ```

mcornella:
Thanks!

rachelcarmena:
Thanks to you, @mcornella. You made me think about my decision of setting `sh_word_split`in `~/.zshrc` :wink:

PR #7882, opened 2019-05-26
make it explicit that oh-my-zsh.sh needs to be sourced in between

jm6626:
make it explicit that oh-my-zsh.sh needs to be sourced in between plugins and prompt. doesn't work if you do it in a different order.


mcornella:
This was fixed in a later PR (https://github.com/robbyrussell/oh-my-zsh/pull/8142), thanks for your contribution!

PR #7888, opened 2019-05-28
Add support for Symfony Client

auipga:
Symfony has a new [Symfony Client](https://github.com/symfony/cli), which is a binary called `symfony`.
You can use `symfony console` the same way as the old way with `php bin/console` with an incredible difference: It loads a local `php.ini`. [docs](https://symfony.com/doc/current/setup/symfony_server.html#overriding-php-config-options-per-project)

I don't know anything about compdef and never contributed to a zsh plugin. So this is just a typed idea.

I think there could be something like the following, but it doesn't work as expected:
```bash
compdef _symfony2 'symfony console'
```
Could you please implement this?

### DIFF
--- a/plugins/symfony2/symfony2.plugin.zsh
+++ b/plugins/symfony2/symfony2.plugin.zsh
@@ -1,7 +1,13 @@
 # Symfony2 basic command completion
 
 _symfony_console () {
-  echo "php $(find . -maxdepth 2 -mindepth 1 -name 'console' -type f | head -n 1)"
+  if [ -x symfony ]; then
+    echo "./symfony console"
+  else if command -v symfony >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for raising this, and sorry nobody replied for so long.

You were ahead of the curve... preferring the `symfony` binary over `php bin/console` is the right call, and it's how most Symfony work happens now.

Two reasons I'm closing rather than merging. The patch as written doesn't parse — the `else if` opens a second conditional that's closed with a single `fi`, so sourcing the plugin errors out. And #12768 already covers this properly against the current plugin, so that's where the work should land.

No hard feelings on the "could you please implement this" framing either... it turned out to be a real gap, it just took us far too long to notice. Thanks.

PR #7897, opened 2019-05-31
[debian plugin] Add an update alias using apt list -u

BjoernDaase:
Add an update option which also lists the available updates before updating (`apt list -u`).
The new alias would be `alias alu='sudo $apt_pref update && sudo $apt_pref list -u && sudo $apt_pref $apt_upgr'`.
This is necessary to see the exact version change of a package


BjoernDaase:
Closed, because it's part of #7899

PR #7901, opened 2019-06-01
Docker completion with container names on restart

tcoupin:


mcornella:
The docker completion is a bundle of the main project's version, this should be proposed upstream: https://github.com/docker/cli/blob/master/contrib/completion/zsh/_docker.

PR #7905, opened 2019-06-03
Add a completion plugin for minishift

pmacik:


aschbacd:
👍

robbyrussell:
Sorry for the delay, but while reviewing this we noticed that the minishift repo has been moved to read-only/archive. We're going to pass on this for now

PR #7908, opened 2019-06-04
ssh-agent: add `:omz:plugins:ssh-agent ssh_add_args` option

wrboyce:


sergeyklay:
@mcornella Please could you please take a look

elreydetodo:
This change works for me, and I'm using it locally. I need to add the -K argument to `ssh-add` on macOS to have the system keychain manage my passphrase.

Can you add a paragraph to `README.md` explaining how to use this?

miguelgila:
👍

VerosK:
This PR would be very nice to merge.

I prefer to use `ssh-add` with `-c` parameter.  `ssh-agent` then asks for confirmation before using any of stored keys.

What can I do to get this PR merged?

DBX12:
As I understand it, this PR would solve my issue too. I need the SSH_AUTH_SOCK in a specific location so keepassxc finds it, the randomized location in /tmp makes it impossible (seems like an update in the ssh-agent binary). With `-a /run/user/1000/ssh-auth` I would have a static location to reference in keepassxc. So merging this PR would really help me.

mcornella:
Everyone that commented, please test this by fetching the PR with `omz pr test 7908` and then use the `ssh-add-args` setting, like it's now documented in the README.

elreydetodo:
I've confirmed it's working for me.

mcornella:
Thanks!

mikelikespie:
This is awesome, thanks! I was actually just looking for something like thsi today (trying to get the plugin to work with `ssh-add -A` to use keychain on macos). Works great thanks!

PR #7909, opened 2019-06-05
ssh helper

cgerke:
Shortcuts to the UI for System Preferences > Sharing > Remote Login. Thoughts?


mcornella:
These aren't probably used often enough so as to warrant an alias for them. Reopen if you can find users who'd like this change. Thanks for your contribution though.

PR #7910, opened 2019-06-05
Add a completion plugin for Dep - Go dependency management tool 

SagiMedina:


robbyrussell:
Thanks for this, and I'm sorry it sat here this long without a word from anyone.

To be clear about why I'm closing it: it isn't the work. The plugin is well put together and it still merges cleanly seven years later, which says something. It's the timing — `golang/dep` was archived and marked deprecated once modules landed, so a new plugin for it would be something we maintain for a tool almost nobody runs now.

If we'd gotten to this in 2019 it very likely would have been merged. That one's on us, not on you. Thanks for the contribution.

PR #7912, opened 2019-06-05
grep.zsh: add --line-number/-n option if available

thibaultcha:
Hello,

Proposing this change which I think makes sense as a default option. Only worry is backwards-compatibility with scripts processing the output of `grep` commands and splitting lines on `:` separators. If not desired by default, I'd love to be able to customize the default `GREP_OPTIONS` value from my `.zshrc` file or something; let me know what's best.

Cheers,

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -5,6 +5,13 @@ grep-flag-available() {
 
 GREP_OPTIONS=""
 
+# show line numbers in grep results
+if grep-flag-available --line-number; then
+    GREP_OPTIONS+=" --line-number"
+elif grep-flag-available -n; then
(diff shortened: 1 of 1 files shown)


mcornella:
> Only worry is backwards-compatibility with scripts processing the output of `grep` commands and splitting lines on `:` separators.

That's a problem yeah; I've checked that line numbers are also shown even when piping its output so we'd have a problem here.

If you want to customize your grep alias, you can realias it after OMZ is sourced with the following line:
```zsh
alias grep="$aliases[grep] --line-number"
```

thibaultcha:
@mcornella That's exactly what I was hoping for, thanks :+1:

PR #7914, opened 2019-06-06
gitfast: update to upstream v2.21

felipec:
I've added a script to automatically update to the latest upstream, and while at it, I updated to the latest upstream:

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -39,7 +39,7 @@ if [ -z "$script" ]; then
 		test -f $e && script="$e" && break
 	done
 fi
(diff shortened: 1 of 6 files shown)


felipec:
Actually I'm thinking this might not be a good idea. It seems there will be some issues for people running versions of Git before v2.17. I'm not sure how many people that is.

Should I sent another request to downgrade to the safe version? (v2.16). Or shall we wait and see what happens with this one?

mcornella:
Um, if you mean there will be problems with git's completion and a downgrade will work with more versions, then sure, if you can ready a PR I'll merge it right away.

PR #7916, opened 2019-06-07
pip: Fixes no such file or directory: /tmp/zsh_tmp_cache

rgreenblatt:
Zsh may be configured such that ">>" will error if the file doesn't exist.


mcornella:
I'll merge this but just so I know it, what settings trigger this?

rgreenblatt:
The setting which causes this is `setopt noclobber` and `setopt clobber` will allow for `>>` even if the file doesn't exist. See https://unix.stackexchange.com/questions/212127/zsh-disable-file-exists-warning-with-redirection.

mcornella:
Good to know, thanks!

PR #7918, opened 2019-06-08
gitfast: downgrade to upstream v2.16

felipec:
The new completions require a version of Git that is at least v2.17, better use the completions right before that so our users don't run into issues inadvertently.


mcornella:
Thanks!

PR #7923, opened 2019-06-12
Fix typo

vfonic:
I noticed `RED` is being used in example, instead of `REP`, as it says it will migrate the production database.


mcornella:
👍

PR #7924, opened 2019-06-12
Add l1 alias as single column variant of ls

pcdevil:
_As title says_ it would be nice to have a shorthand out-of-the-box for the `ls -1` command.

Nothing fancy but makes piping through grep/other commands easier and quicker.

cc @mcornella


mcornella:
When ls is piped it always uses single-column form. See for example `ls | cat`.

PR #7925, opened 2019-06-12
Add completion plugin for s2i.

pmacik:


robbyrussell:
Thanks for this, and sorry it sat here so long without a reply.

I'm going to pass on it. Seven years open and nobody else chimed in asking for it, which is about as clear a signal as we get that it's not something people have been waiting on. Every plugin is also something a volunteer maintains when it breaks a few years from now, so we're careful about which ones we take on. Nothing wrong with the work itself.

If you'd like to keep using it, drop it in `~/.oh-my-zsh/custom/plugins/s2i/` and it'll load the same way and survive updates.

Thanks for taking the time to write it up.

PR #7935, opened 2019-06-18
pipenv: spawn a shell within virtualenv on a new session

jjangga0214:
When opening a new terminal tab or window (new session) on a directory which has **Pipfile**, new shell within project-specific **virtualenv** is not activated automatically.

So, hereby I let the plugin **pipenv** spawn a new shell on a new terminal session as well.


mcornella:
Done in 4ab273e19ad1768163e662831b3418a2fc9b6eee.

PR #7936, opened 2019-06-18
Add basic gcp-ps1

n0npax:
Add basic PS1 for current gcloud configuration


robbyrussell:
Thanks for this, and sorry it sat here without an answer for so long. Nobody else has asked for a GCP prompt segment in the years since, and a hand-maintained prompt plugin with no upstream behind it is a long commitment for us... so I'd rather pass on it than leave it open indefinitely.

It works well as a custom plugin in `$ZSH_CUSTOM` if you're still using it. Thanks again.

Issue #7642, opened 2019-03-09
Some plugins seem not to load


jkulak:
Clean installation on macOS Mojave 10.14.3 (18D109). 
`git status` inside `~/.oh-my-zsh` shows clean.
`git rev-parse HEAD` shows `52afbf77f63844abd719e565a15af2c34075de7f`

No changes to `.zshrc`.

Plugin for git works fine

```
➜  ~ git 
zsh: do you wish to see all 130 possibilities (130 lines)? 
```

I add `docker` in `.zshrc`. `plugins=(git docker)`. I restart Terminal (or run `zsh`) and...

```
➜  ~ docker
Applications/     Data/             Lota/             Public/         
Archive/          Documents/        Movies/           bin/            
Desktop/          Downloads/        Music/            
Developer/        Library/          Pictures/                   
```

Git plugin still works.

Debug log: https://pastebin.com/RikhtP3g

I tried running it manually.

```
➜  ~ source ~/.oh-my-zsh/plugins/docker/_docker
➜  ~ echo $?
1
```

Also

```
➜  ~ exec ~/.oh-my-zsh/plugins/docker/_docker
/Users/jkulak/.oh-my-zsh/plugins/docker/_docker: line 58: unexpected argument `(' to conditional binary operator
/Users/jkulak/.oh-my-zsh/plugins/docker/_docker: line 58: syntax error near `(s'
/Users/jkulak/.oh-my-zsh/plugins/docker/_docker: line 58: `    [[ $kind = (stopped|all) ]] && args=($args -a)'

[Process completed]
```

jkulak:
It has something to do with the `~/.zcompdump-mycomp-5.3` file. After adding `docker` plugin in `~/.zshrc` I have to remove the mentioned file and restart the session to make `docker` plugin work.

After, removing the `docker` plugin from `.zshrc` and restarting the session, I get

```
➜  ~ docker [TAB]
(eval):1: _docker: function definition file not found
(eval):1: _docker: function definition file not found
(eval):1: _docker: function definition file not found
```

Aftrer `rm ~/.zcompdump-mycomp-5.3` it comes back to normal, plugin is turned off and hitting TAB shows me the files.

andrewmaudsley:
@jkulak Thanks for reporting this and sharing the work around of removing the .zcompdump... files.  This has been driving me crazy!

mcornella:
[...]

PR #7937, opened 2019-06-18
Fix taskwarrior completion

### LINKED ISSUE #3796: Taskwarrior completion doesn't work
I had enabled taskwarrior plugin, but completion doesn't work. 
If I try task[tabtab] then I only received suggest with 'task' word.

seguri:
Fixes #3796


mcornella:
Actually as I said later in https://github.com/robbyrussell/oh-my-zsh/issues/3796#issuecomment-503275743, the compdef line doesn't hurt and what really fixed it was resetting the completion cache (`~/.zcompdump*`), see #7642.

PR #7948, opened 2019-06-23
extract: support unpacking .deb from different directory

kapsh:


kapsh:
TL,DR: `extract` now can only unpack `.deb` files in current directory and misses control files. This commit adds more universal behavior.

Long version.

1. I'll try to explain first change on this example: 
```
% cd /tmp
% x ~/downloads/chromium-codecs-ffmpeg-extra_74.0.3729.169-0ubuntu0.19.04.1_amd64.deb 
ar: ..//home/kapsh/downloads/chromium-codecs-ffmpeg-extra_74.0.3729.169-0ubuntu0.19.04.1_amd64.deb: No such file or directory
tar (child): ../control.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
extract:54: no matches found: ../data.tar.*
```
`$extract_dir` was created in current working directory (e.g. `/tmp`). Plugin did `cd $extract_dir` and `ar` couldn't find file `../$1` there because `$1` is located in completely different path (my `~/downloads`). I resolved this issue using full path to archive. 

2. Second change takes into account that Debian/Ubuntu switched to `tar.xz` archives, but `control.tar.gz` was specified literally. Use wildcard here.

3. Call `rm -f` because user could have `alias rm=rm -i` and this would break non-interactivity when removing intermediate files.

53jk1 (approved):

PR #7950, opened 2019-06-24
Add new n98-magerun2 command

jissereitsma:
Magento 2 has been released more than three years ago as a followup for Magento 1, and the third party CLI tool `n98-magerun` (currently included in oh-my-zsh) also got a new version `n98-magerun2` (https://github.com/netz98/n98-magerun2). This PR adds support for the new CLI tool as well, which is just as popular amongst Magento 2 developers as the original tool for Magento 1.


mcornella:
It seems the bulk of the alias is identical to the n98-magerun plugin, can't you adapt that to support both?

jissereitsma:
@mcornella Good call. That was a dead easy and I should have done that right away. The PR is now updated.

mcornella:
Thanks for your patience!

jissereitsma:
Thanks!

PR #7952, opened 2019-06-24
Ability to exclude ohmyzsh features via a list in zshrc

cbartondock:
Fairly self explanatory, I wanted a simple option to disable certain features from within zshrc.


mcornella:
This has become polluted, please reopen the PR from a branch other than master.

cbartondock:
Oof what the heck happened here. Will do so.

PR #7962, opened 2019-06-27
Some small updates to the Jira plugin

JoshChris:
- Updating `branch` command to be compatible with git-flow style branches by removing any prefix before attempting to go to the ticket number in Jira.
- Removing the switch on the `branch` command between `/browse/` and `/issues/` based off of the `JIRA_RAPID_BOARD` setting - as all issues are under `/browse/` now it seems.


jmfayard:
@JoshChris my co-workers like to have descriptive names in git branches, so instead of `TAP-355` my branches are called something like `TAP-355_sync_updates_on_handled_by_akelius_cases`. Could you handle this use case?

PR #7964, opened 2019-06-28
Extract from password protected zip, 7z, rar files

lixin9311:
Although I'm not quite good at shell programming, I wanna share my modifications.

For example, I have a bunch of RAR files protected by the same password. I won't bother typing the passwords each time for every single file.

```shell
extract -r -p the_secret_code *.rar
```

And the order of these arguments doesn't matter now.

```shell
extract a.rar b.zip c.7z -r -p my_dark_secret
```

I fell I didn't write the auto-completion the right way, feel free to edit.

Tested on Linux, not sure about the BSD OSs.

### DIFF
--- a/plugins/extract/extract.plugin.zsh
+++ b/plugins/extract/extract.plugin.zsh
@@ -4,68 +4,93 @@ extract() {
 	local remove_archive
 	local success
 	local extract_dir
+	local files
+	local unrar
+	local _7za
+	local unzip
 
 	if (( $# == 0 )); then
 		cat <<-'EOF' >&2
 			Usage: extract [-option] [file ...]
 
 			Options:
-			    -r, --remove    Remove archive after unpacking.
+			    -r, --remove                Remove archive after unpacking.
+			    -p, --password {password}   Use the password to extract.
 		EOF
 	fi
 
 	remove_archive=1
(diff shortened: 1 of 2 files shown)


bartekpacia:
Hello @lixin9311, thank you for this modification. Please follow our [PR template](https://github.com/ohmyzsh/ohmyzsh/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and fix merge conflicts.

plugins/extract/extract.plugin.zsh
```diff
@@ -4,68 +4,93 @@ extract() {
...
+			shift 2
+			continue
+		fi
+
 		if [[ ! -f "$1" ]]; then
 			echo "extract: '$1' is not a valid file" >&2
 			shift
 			continue
 		fi
+		files+="$1"
```
TriMoon, line 49: This should be `files+=( "$1" )` because you are adding an array element....

plugins/extract/extract.plugin.zsh
```diff
@@ -4,68 +4,93 @@ extract() {
...
+		shift
+	done
+
+	
 
+	for file in ${files[*]}; do
```
TriMoon, line 55: use `"${files[@]}"` here instead of `${files[*}` to expand all elements as separate elements to iterate over.

PR #7969, opened 2019-06-30
agnoster: fixed fg color for context and status

fabb:
it was using `default` which did result in black text on black background for light themes


robbyrussell:
Thanks for this, and sorry it waited so long. This same fix landed in 92da3108 — context and status both use white on light backgrounds now, and they're overridable with `AGNOSTER_CONTEXT_FG` and `AGNOSTER_STATUS_FG` if you want something else.

Closing since it's already in. Thanks again.

PR #7970, opened 2019-07-01
Silence Errors

sabrehagen:
Redirect stderr to avoid printing errors to the console when the ssh key is not unlocked.


mcornella:
Thanks!

PR #7971, opened 2019-07-01
add --skip-runzsh option

boazca:
Useful if inserted inside a script:

```
sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh)" --skip-runzsh

echo "$some_alias">>~/.zshrc 
```

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -25,6 +25,7 @@
 #
 # You can also pass some arguments to the install script to set some these options:
 #   --skip-chsh: has the same behavior as setting CHSH to 'no'
+#   --skip-runzsh: has the same behavior as setting RUNZSH to 'no'
 #   --unattended: sets both CHSH and RUNZSH to 'no'
 # For example:
 #   sh install.sh --unattended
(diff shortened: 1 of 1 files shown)


mcornella:
You can prepend `RUNZSH=no` to the `sh` install line like so:
```
RUNZSH=no sh -c "$(curl ...)"
```

boazca:
But then if you change the variable name it will break the script. 
Argument option is an API and less like to change or at least developers aware of the risk when the change them.
I can also do `sh -c "$(curl -fsSL https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/tools/install.sh | sed 's:exec zsh -l::g')"` 
but thats also not bullet proof. it broke my script because you recently changed `env zsh` to `exec zsh`

mcornella:
Variable-based settings were recently introduced and, as it says in the install script comments, are not subject to change. Trust me on this, I made it.

Any reason why you're not using the `--unattended` option? The `chsh` part requires user input, so maybe you don't need it?

boazca:
I use it in [this site](https://www.mac-installer.com)
It creates a script for setting developer's workstation and It requires user input. 
If I leave the `exec zsh` it will stop the script from running mid-way. 
putting Oh My Zsh at the end is not good as well because if I want to add lines to the `.zshrc` file it has to happen after Oh My Zsh installation.

PR #7973, opened 2019-07-02
add duckduckgo as an option to frontend-search

jackbrown:


jackbrown:
@willmendesneto this is a small addition to the plugin you created, do you mind taking a look at this for me?

plugins/frontend-search/README.md
```diff
@@ -62,6 +62,10 @@ Available search contexts are:
 
 If you want to have another context, open an Issue and tell us!
 
+## Changing the fallback search behaviour
+
+You can set the fallback search engine to DuckDuckGo by setting `FRONTEND_SEARCH_FALLBACK_SEARCH_ENGINE='duckduckgo'` in your `~/.zshrc` file.
```
willmendesneto: Would you mind to add the available options in this documentation, sharing that by default is using google, but it can uses `duckduckgo` if added in `FRONTEND_SEARCH_FALLBACK_SEARCH_ENGINE`, please? It will be better knowing in this readme than having to look at the code

jackbrown: No problem, I've updated the readme to include this.

willmendesneto:
Hey @mcornella , could you check this pull request, please?

mcornella:
👍, just a couple things: I would rename the option to `FRONTEND_SEARCH_FALLBACK` (the search engine is implied), and the function name to something like `_frontend_search_url` (the underscore is used for functions not used directly by users, and we don't use camel-case for functions).

PR #7974, opened 2019-07-02
Prettier output for AWS EKS

anton-yurchenko:
Hey,
When working against Kubernetes on AWS the output looks like that: `arn:aws:eks:us-west-2:123456789012:cluster/unicorn:default`, which is obviously very long.

This PR introduces new variable `KUBE_PS1_EKS`, which whenever set to *true* will format the output to: `unicorn:default`.

### DIFF
--- a/plugins/kube-ps1/kube-ps1.plugin.zsh
+++ b/plugins/kube-ps1/kube-ps1.plugin.zsh
@@ -121,7 +121,7 @@ _kube_ps1_get_context_ns() {
   KUBE_PS1_LAST_TIME=$EPOCHSECONDS
 
   if [[ "${KUBE_PS1_EKS}" == true ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
Please make the changes in the [upstream project](https://github.com/jonmosco/kube-ps1).

jonmosco:
The kubectl command allows for renaming contexts: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-rename-context-em-

```bash
$ kubectl config rename-context CONTEXT_NAME NEW_NAME
```
So in this case, the command would be similar to the following:
```bash
$ kubectl config rename-context arn:aws:eks:us-west-2:123456789012:cluster/unicorn unicorn
```

This would be the preferred method, and is also able to apply to other contexts created by external providers.

PR #7975, opened 2019-07-02
use `mafredri/zsh-async` to accelerate

pre1ude:
use `mafredri/zsh-async` to accelerate the nvm plugin, now works faster

https://github.com/nvm-sh/nvm/issues/539#issuecomment-403661578


mscottnelson:
I like the idea of zsh-async, and I like this commit, but as currently committed this will fail (unless zsh-async is already present).

mcornella:
This also assumes the location of nvm.sh instead of using `$NVM_DIR`. Plans to implement zsh-async are in the near future though.

PR #7977, opened 2019-07-03
Add ggpush! and gfu

davialexandre:
I realized I use these two quite often, so I thought it would be helpful to have aliases for them.


robbyrussell:
Thanks for this, and sorry for the long silence. Both of these are covered now: `gpf` force-pushes the current branch with `--force-with-lease --force-if-includes`, and `gfa` does the fetch-and-prune across all remotes. The plugin also picked up `gpf!` and `ggfl` along the way.

Issue #7610, opened 2019-02-19
Vi mode breaks special characters ( Home, End, Delete, Backspace)


gudrutis:
Experienced on: MacOs and Fedora26.
Zsh works fine untill I turn on Vi mode ( either by `set -o vi` or `bindkey -v`).

No `bindkey` solutions worked for me.

aymanbagabas:
try using `-M vicmd` in your `bindkey`
I have the following in my config
```
# allow delete, home and end keys to work in normal mode
if [[ "${terminfo[khome]}" != "" ]]; then
  bindkey -M vicmd "${terminfo[khome]}" beginning-of-line      # [Home] - Go to beginning of line
fi
if [[ "${terminfo[kend]}" != "" ]]; then
  bindkey -M vicmd "${terminfo[kend]}"  end-of-line            # [End] - Go to end of line
fi
if [[ "${terminfo[kdch1]}" != "" ]]; then
  bindkey -M vicmd "${terminfo[kdch1]}" delete-char            # [Delete] - delete forward
fi
```

Backspace works fine for me

mcornella:
I can no longer reproduce this. I think we fixed that in `lib/key-bindings.zsh` in ec70c0c393e53004b885d146dd77b9a3f136971a.

PR #7978, opened 2019-07-04
vi-mode: fix behaviour on ctrl-c and define default keybindings

aymanbagabas:
Sending a ctrl-c interrupts resets VI_KEYMAP and as a result, we get a
false vi mode since that variable is not set. A workaround is to redefine
both zle-keymap-select and zle-line-init to perform the same task.

Use some default key bindings for home, end, and delete keys. #7610

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -27,14 +27,14 @@ bindkey -v
 bindkey -M vicmd '^J' vi-accept-line
 bindkey -M vicmd '^M' vi-accept-line
 
-# allow v to edit the command line (standard behaviour)
+# allow ctrl-v to edit the command line (standard behaviour)
 autoload -Uz edit-command-line
 zle -N edit-command-line
-bindkey -M vicmd 'v' edit-command-line
+bindkey -M vicmd '^V' edit-command-line
 
-# allow ctrl-p, ctrl-n for navigate history (standard behaviour)
-bindkey '^P' up-history
-bindkey '^N' down-history
+# allow ctrl-p, ctrl-n for navigate history and fuzzy find history
+bindkey '^P' up-line-or-beginning-search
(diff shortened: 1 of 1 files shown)


whilefalse:
👍 

I've just started using Zsh with Vi mode, and the delete key in particular caught me off guard. For me it would also be nice to add:

`bindkey "${terminfo[kdch1]}" delete-char`

(and equivalent for home/end) so the delete key works in INSERT mode as well.

I hope this can be merged so I can remove these extra bindings in my .zshrc.

aymanbagabas:
> I've just started using Zsh with Vi mode, and the delete key in particular caught me off guard. For me it would also be nice to add:
> 
> `bindkey "${terminfo[kdch1]}" delete-char`
> 
> (and equivalent for home/end) so the delete key works in INSERT mode as well.
> 
> I hope this can be merged so I can remove these extra bindings in my .zshrc.

thank you @whilefalse and I hope this gets merge as well. I think @erydo did a better job in vi-mode for omz you might wanna check that out #8004 but apparently, he didn't include the del key in his pr.

erydo:
Thanks for the note @aymanbagabas. I must have overlooked that binding (https://github.com/robbyrussell/oh-my-zsh/blob/de3b14cf69eeb43797bae8320af6972032b604ff/lib/key-bindings.zsh#L60-L66) in my PR's commit (https://github.com/robbyrussell/oh-my-zsh/pull/8004/commits/042dd3933a430941b0cd49bf91a24913e70c1743). Let me see if I can add that into my PR as well so people have something to cover all their bases.

sergekir:
Thanks for PR, I hope It'll be merged :)
If somebody interested in temporary solution based on this PR you can add this to the end of `.zshrc` : 

```
function zle-line-init zle-keymap-select() {
  VI_KEYMAP=$KEYMAP
  if (( ${+terminfo[smkx]} )); then
    echoti smkx
  fi
  zle reset-prompt
  zle -R
}

```

PR #7980, opened 2019-07-04
Add alias for tmuxinator and tmuxinator stop

heidar:
This PR adds an alias for `tmuxinator` as `tx` as well as an alias for calling `tmuxinator stop` as `txx`.

I've been using this plugin for a while now and feel like this is missing from it.


jampow:
this was exactly the same what I did in my local copy!

jampow:
btw, you could request review from the plugin maintainer @mcornella

heidar:
> btw, you could request review from the plugin maintainer @mcornella

Hmm I don't seem to be able to request reviews

PR #7982, opened 2019-07-06
Automatic title: Replace fg with description from jobs

### LINKED ISSUE #7981: Avoid "fg" in terminal title (omz_termsupport_preexec)
**Is your feature request related to a particular plugin or theme? If so, specify it.** Base ZSH, `lib/termsupport.zsh`, specifically `omz_termsupport_preexec()`

When I suspend my editor (`vim`) and return via `fg`, then my terminal window's title changes to `fg` instead of the initial title (e.g. `vim paper.tex`).  I'd argue that this is never a useful title.

I'd like `fg` to trigger setting the title to the actual job name, as reported by `jobs`.

[This superuser.com post](https://superuser.com/questions/497715/keep-tmux-title-from-fg-to-original-jobname) describes one approach for fixing this problem, but I think it should be in base zsh, as I can't imagine `fg` being a useful window title.

I'm working on a PR; will post once complete.

edemaine:
Fix #7981 by using actual job title instead of `fg` in window title.  Based on part on [this superuser.com discussion](https://superuser.com/questions/497715/keep-tmux-title-from-fg-to-original-jobname).

I'm new to zsh scripting, so please let me know if this could be written any cleaner/shorter.  I found it particularly odd that I needed to check for the no-argument case separately; otherwise, `${${(z)1}[2]}` would return `g` (the second character of `fg`).

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -68,24 +68,47 @@ function omz_termsupport_preexec {
     return
   fi
 
+  # split command into array of arguments
+  local -a cmdargs
+  cmdargs=("${(z)2}")
+  # if running fg, extract the command from the job description
+  if [[ "${cmdargs[1]}" = fg ]]; then
+    # get the job id from the first argument passed to the fg command
+    local job_id jobspec="${cmdargs[2]#%}"
+    # logic based on jobs arguments:
+    # http://zsh.sourceforge.net/Doc/Release/Jobs-_0026-Signals.html#Jobs
+    # https://www.zsh.org/mla/users/2007/msg00704.html
+    case "$jobspec" in
+      <->) # %number argument:
+        # use the same <number> passed as an argument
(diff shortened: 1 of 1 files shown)


edemaine:
Is there anything I can do to help get this merged? It's still working well for me, 4 months later, on the current master branch.

mcornella:
While your version worked reasonably well for `fg` and `fg %number`, it didn't cover the other invocations of `fg` using `%string` and `%?string` because `jobs %string` dies inside `$()` with a `job not found` error.

I replaced most of it with a version that uses the parameters `$jobstates` and `$jobtexts`, and while it significantly more verbose every case is covered. I got some inspiration from here: https://www.zsh.org/mla/users/2007/msg00704.html

edemaine:
Awesome, thanks for the improvements! (and review and merging)

PR #7983, opened 2019-07-08
Make unattended installation command a one-liner

MEHColeman:
The original suggestion for an unattended install downloads the installation script to a file, then runs that file with the --unattended argument. The install.sh file would be left behind after the suggested command was run.
This change pipes the file from curl into bash while passing the --unattended argument into bash. So, it's a nice one-liner like the default installation script, and it doesn't leave a dangling install.sh script.


mcornella:
I've changed it to use the `sh -c` construct because using pipes gave problems in the past. Thanks!

PR #7841, opened 2019-05-11
git: add git revert alias

SomeDer:

PR #7984, opened 2019-07-09
git: Add missing `grev` alias to docs

Addono:
#7841 adds support for the `grev` alias for `git revert`, however this is not reflected in the documentation.


mcornella:
Thanks!

PR #7988, opened 2019-07-11
bugfix in colorize mktemp

fzuviria:
Hey Mark, long time no see.

with this update, 

cless README.TXT

will now generate the temp name 

/tmp/tmp.colorize.HGg.README.TxT

Turns out that mktemp treats the X in TXT as part of the template and fails. I did not find a way to escape it from mktemp's format :(

It is still very easy to guess which file I'm looking at the terminal :)

Also, I reduced the random component to the minimum 3 chars, to get more screen space for long paths...


fzuviria:
Also, this is another bugfix/improvement:

`-e 's/^.\///'`

To prevent a triple dot in the tmp filename when doing `cless ./filename`
In other words, so the tmp file is the same if I do `cless ./filename` or `cless filename`

fzuviria:
Hello @mcornella , can this be merged please? Is there anything that should be improved? Best regards

mcornella:
#8392 adds support for not having to create the tmp files, can you check that out and see if we maybe don't need this one anymore?

mcornella:
We merged #8392, this is no longer needed. Thanks!

PR #7991, opened 2019-07-12
Quasar framework plugin

sajt:
I wrote a plugin for quasar framework (quasar.dev)


Qoyyuum (approved):
Can confirm that the Quasar plugin for oh-my-zsh works.

robbyrussell:
Thanks for putting this together, and sorry it sat here this long without a reply. We closed a second Quasar plugin (#11351) in July for lack of demand, so I'm giving this one the same answer... it works well as a custom plugin in `$ZSH_CUSTOM` if you're still using it.

PR #7992, opened 2019-07-12
used kubectl alias everywhere

sanderv:
I've changed the aliases to use the root-alias 'k'. In this way I can easily switch between standard kubectl and microk8s.kubectl just by redefining the 'k' alias. I see you've reused the 'kgp' alias as well, so I think this would be all right too.
See https://microk8s.io/docs/ for more info


sanderv:
Found another way without changing the plugin. Oh wel... ;)

PR #7995, opened 2019-07-12
Added kube-ps1 CLUSTER_FUNCTION and NAMESPACE_FUNCTION

genebean:
This adds the ability to set `KUBE_PS1_CLUSTER_FUNCTION` and `KUBE_PS1_NAMESPACE_FUNCTION` like are done in the kube-ps1 program. Code is a near copy/paste from https://github.com/jonmosco/kube-ps1


mcornella:
Closing in favor of pulling all the upstream changes in #8009.

genebean:
The PR above doesn’t address the cluster or namespace functions... can we still add these? Do I need to rebase for that to happen?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

maxdanilov:
@mcornella Do you have any ETA on when this change will be available?

PR #7996, opened 2019-07-12
lib: support additional clipboard types

erydo:
This implements essentially the same heuristic as neovim, preserving the additional special support for Cygwin.

See: https://github.com/neovim/neovim/blob/e682d799fa3cf2e80a02d00c6ea874599d58f0e7/runtime/autoload/provider/clipboard.vim#L55-L121

The following clipboard types are tried in order:
 - pbcopy, pbpaste (macOS)
 - cygwin (Windows running Cygwin)
 - wl-copy, wl-paste (if `$WAYLAND_DISPLAY` is set)
 - xclip (if `$DISPLAY` is set)
 - xsel (if `$DISPLAY` is set)
 - lemonade (for SSH) https://github.com/pocke/lemonade
 - doitclient (for SSH) http://www.chiark.greenend.org.uk/~sgtatham/doit/
 - win32yank (Windows)
 - tmux (if `$TMUX` is set)


lib/clipboard.zsh
```diff
@@ -67,20 +47,55 @@ function clipcopy() {
...
+      local clipcmd="${1}"; shift
+      if detect-clipboard; then
+        "${clipcmd}" "$@"
+      else
+        print "${clipcmd}: Platform $OSTYPE not supported or xclip/xsel not installed" >&2
+        return 1
+      fi
+    }
+    function clipcopy() { _retry_clipboard_detection_or_fail clipcopy "$@"; }
+    function cilppaste() { _retry_clipboard_detection_or_fail clippaste "$@"; }
```
mcornella: Typo here

erydo: Good eye! Thanks. Fixed, and verified no similar typos.

erydo:
@mcornella — I've tested:
  * `pbcopy` (existed before)
  * `xclip` (existed before)
  * `xsel` (existed before)
  * `tmux` (new)

The first three existed before, and continue to work. I don't have an easy environment set up for testing the other ones:
  * `cygwin` (existed before; should be working the same)
  * `wl-copy`, `lemonade`, `doitclient`, and `win32yank`

Their implementations were based on the `nvim` implementation, so at least their argument types/orders should be correct. But untested.

PR #7999, opened 2019-07-14
Add ZSH_TMUX_NEW_SESSION_ARGS option to tmux plugin

wedaly:
When the tmux plugin auto-starts tmux, it calls `tmux new-session`
with no arguments.  It can be useful to pass additional arguments
to this command to set the session name or session group.
This commit adds a new environment variable ZSH_TMUX_NEW_SESSION_ARGS
that can be used to pass additional arguments to the tmux new-session
command.

This is similar to, but more general than, this (closed) PR to set the session name: https://github.com/robbyrussell/oh-my-zsh/pull/5792


larson-carter:
@wedaly Hey, can you please go into a little more detail with that this will do? Why would we want it to pass in 0 arguments? What benefit would this provide?

PR #8000, opened 2019-07-14
Enhanced clipboard detection and management

rwe:
This is based on #7996, which can be merged on its own. This is split off on its own to simplify things if there are any issues with this branch.

This adds several new commands and capabilities:
## New Command: `clipboard-list`
Users can list the names of the available clipboard types. Example:
> ```
> $ clipboard-list
> macos
> tmux
> file
> var
> null
> ```

## New Command: `clipboard-set`
Users can set the selected clipboard driver by name. Example:
> ```
> clipboard-set tmux
> ```

## Feature: Custom clipboard list
Users may customize the list and order of autodetected clipboards by setting the array variable `ZSH_CLIPBOARDS`.
> ```
> ZSH_CLIPBOARDS=(macos tmux)
> ```

## Feature: File-based clipboard
Users can fall back to a file-based clipboard buffer. This will either be a tempfile which gets cleaned up automatically, or a filename specified by the user as `ZSH_CLIPBOARD_FILENAME`.

## Feature: `var` clipboard
Just uses an environment variable (`ZSH_CLIPBOARD_CONTENTS`) for copying and pasting. Works fine for vi-mode copying and pasting.

And added a `null` fallback clipboard, which does nothing and pastes empty.

### DIFF
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -175,8 +175,7 @@ function __omz_clipboard_file_setup() {
       autoload -U add-zsh-hook
 
       # NOTE: This is not exported. That way we can be sure *we* created it.
-      __omz_clipboard_file="$(mktemp -t "omz-clipboard-${USER}.XXXXXXX")"
-      typeset -rg __omz_clipboard_file
+      typeset -rg __omz_clipboard_file="$(mktemp -t "omz-clipboard-${USER}.XXXXXXX")"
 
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

carlosala:
Closing this, we just changed the way clipboard is loaded (on usage) and I agree with Marc's comment, it seems to be a lot of stuff for something so simple.
Thanks for your contribution!

PR #8009, opened 2019-07-17
merge updates from upstream kube-ps1 project

jonmosco:
Fixes issue reported on upstream project: 
https://github.com/jonmosco/kube-ps1/issues/92


mrrsm:
I was testing this out and it looks like it is missing the check to see if the global disable flag was set.

From the upstream:
`[[ -f "${KUBE_PS1_DISABLE_PATH}" ]] && KUBE_PS1_ENABLED=off`

mcornella:
Are there any more changes we should pull in?

PR #8010, opened 2019-07-17
fabric: support fabric 2+ completion targets

cybergrind:
Hi @robbyrussell / @mcornella ! 
New major fabric version is using another option to list targets. This commit adds the fallback to the newer version listing in case of errors.


mcornella:
This PR is \*chef's kiss\* 👌

cybergrind:
thank you, @mcornella

PR #8011, opened 2019-07-18
random theme loading message removed

sauravmanoj:


mcornella:
The random theme logic is now encapsulated in its own theme: https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/random.zsh-theme. If you want this changed, make a custom random theme which sources the default random theme and redirect its output to `/dev/null`. See [the wiki for customization](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes).
```zsh
# Put this in $ZSH_CUSTOM/themes/random.zsh-theme
source $ZSH/themes/random.zsh-theme &>/dev/null
```

PR #8014, opened 2019-07-18
fix: ignore .DS_Store

DaniSpringer:
Ignore .DS_Store (MacOS)


mcornella:
This hasn't been a problem so far, but thanks for the contribution.

PR #5855, opened 2017-02-07
Followup to #5299 much more dynamic completion for angular cli

catull:
This PR updates the completion for Angular CLI.

PR #8018, opened 2019-07-19
ng: suggest projects in build, test and lint

yannickglt:
Now, in addition of the default command options, the `build`, `test` and `lint` commands suggest the list of available projects for the Angular workspace.

Example with the [nx-examples](https://github.com/nrwl/nx-examples) repository:
```sh
 ~/nx-examples: $ ng test
--browsers                    --reporter                    -r                            shared-cart-state
--colors                      --server                      -s                            shared-e2e-utils
--config-file                 --silent                      -tp                           shared-header
--environment                 --test-page                   -w                            shared-jsxify
--filter                      --test-port                   cart                          shared-product-data
--host                        --watch                       cart-cart-page                shared-product-state
--launch                      -H                            cart-e2e                      shared-product-types
--log-level                   -c                            products                      shared-product-ui
--module                      -cf                           products-e2e                  shared-styles
--path                        -e                            products-home-page
--port                        -f                            products-product-detail-page
--query                       -m                            shared-assets
```

### DIFF
--- a/plugins/ng/ng.plugin.zsh
+++ b/plugins/ng/ng.plugin.zsh
@@ -1,5 +1,7 @@
 ng_opts='addon asset-sizes b build completion d destroy doc e2e g generate get github-pages:deploy gh-pages:deploy h help i init install lint make-this-awesome new s serve server set t test update v version -h --help'
 
+projects=($(ng config projects | grep -E '^\s{2}"' | sed -E 's/.*"([^"]*)".*/\1/'))
+
 _ng_completion () {
   local words cword opts
   read -Ac words
@@ -16,7 +18,7 @@ _ng_completion () {
       ;;
 
     b | build )
-      opts='--environment --output-path --suppress-sizes --target --watch --watcher -dev -e -prod'
+      opts="$projects --environment --output-path --suppress-sizes --target --watch --watcher -dev -e -prod"
       ;;
 
     d | destroy )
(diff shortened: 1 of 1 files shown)


yannickglt:
The current version is not perfect as it calculates once (at zsh startup) the projects dependencies. 
It means we need to manually reload zsh when entering in a new Angular workspace, in order to get benefits from projects completion.
Another solution would have been to always extract projects on command completion, but as the projects calculation is very slow (due to `ng config`), it makes usage really poor...
If someone gets a better solution to get projects completion, I would appreciate its help 🙂

mcornella:
The solution is to stop using a custom completion function and start using `ng completion --zsh`. Closing in favor of #5855.

mcornella:
Nevermind, they [deleted the command in v6](https://github.com/angular/angular-cli/releases/tag/v6.0.0) (see `Breaking changes` section). This can still make it in, just put the `projects=()` line inside the function, only when the completion requires it.

yannickglt:
As I said, moving the projects call inside the function slows down the completion and make it unusable.

mcornella:
Even if we only do that when using `$projects`? Something like
```zsh
if [[ "$words[cword]" = b|build|l|lint|t|test ]]; then
  projects=($(ng config projects | grep -E '^\s{2}"' | sed -E 's/.*"([^"]*)".*/\1/'))
  opts+=" $projects"
fi
```

PR #8019, opened 2019-07-19
Update _cargo completion to 0.37.0 version

yotamofek:


mcornella:
Fixed in https://github.com/ohmyzsh/ohmyzsh/commit/05dfd0ae8520e58e852c574f55c83fe6f49fb189.

PR #8021, opened 2019-07-19
Created Fire Theme

esfoliante:
# Fire theme
### Just avoid burning your codes

Just created this theme based on (basically copying) the cloud theme.


mcornella:
[We can't accept more themes](https://github.com/ohmyzsh/ohmyzsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/ohmyzsh/ohmyzsh/wiki/External-themes) wiki page.

You can also add your theme as a custom theme. See [customization](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes) for instructions.

PR #8027, opened 2019-07-22
Removing plugin that was named after an external tool

planetargon:
This plugin's name has a history as a racist phrase. Was reported by a user of OMZ who shared some background on the terminology and how it made them feel. Understandably, I've decided to immediate remove it from the project


bomm:
I'm not interested in this plugin as I don't have a Mac, but can you share some background information? (Preferably some links.)
The word "bwana" seems to be mainly used for normal purposes, see https://glosbe.com/sw/en/bwana and the explanation on https://www.bruji.com/bwana/
There is even a Kenyan Pop song "Jambo Bwana", see https://en.wikipedia.org/wiki/Jambo_Bwana

robbyrussell:
Am open to sharing how this information came to my attention.

https://www.linkedin.com/feed/update/urn:li:activity:6558508037055881216/

bomm:
IMHO the complaint is not very clear about where the racism is supposed to be in the description of the Bwana tool. The explanation of the name even acknowledges that the oldest drawing of a man was found in a cave in Namibia.

robbyrussell:
@bomm I'm going to trust their perspective on it.

PR #8034, opened 2019-07-24
Store zshrc in `custom`

b-:
This commit will store `zshrc` in `${ZSH}/custom/zshrc` by default, and put something simply sourcing it in `~/.zshrc`.

I did this so that I could `git init` inside my `${ZSH}/custom/` and track that separately, including my zshrc, in one repo.

### DIFF
--- a/custom/.gitignore
+++ b/custom/.gitignore
@@ -1,0 +1,2 @@
+*example*
+*.swp
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -33,7 +33,7 @@ set -e
 
 # Default settings
 ZSH=${ZSH:-~/.oh-my-zsh}
-REPO=${REPO:-robbyrussell/oh-my-zsh}
+REPO=${REPO:-robbyrussel/oh-my-zsh}
 REMOTE=${REMOTE:-https://github.com/${REPO}.git}
 BRANCH=${BRANCH:-master}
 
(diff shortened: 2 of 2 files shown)


mcornella:
This use case won't be supported. Note that you can add the `source $ZSH_CUSTOM/zshrc` line yourself or put it in an init file to be run when provisioning a new machine. Also, OMZ automatically sources files in $ZSH_CUSTOM so you probably will run into a loop were OMZ continually sources itself.

b-:
I actually was worried about such a loop based on the documentation saying that it sources files from there itself before picking `${ZSH}/custom/zshrc` as a location. I was hoping I could store it in `${ZSH}/custom`, as it being in `.gitignore` allows me to make that a separate repository to store my customizations along with my `zshrc` without forking https://github.com/robbyrussel/oh-my-zsh

In order to test if there would be a loop issue, I checked and added `set -x` to top of `oh-my-zsh.sh` to trace execution and created a few dummy files with varying names and executable bits on or off to test, and it only sourced files ending `.zsh`, ignoring the executable bit. It did not touch anything with any name like (and including) zshrc. I did this directly after editing the installer to point to my own repo and branch, and installing a clean install that way.

Therefore, I am confident that it will not run into a loop if a file named `zshrc` is in `${ZSH}/custom`.

If there's any suggestion as to how I can implement what I desire (creating a repo for dotfiles that includes my zshrc as well as `${ZSH}/custom`) I'd certainly appreciate it, whether it involves changing oh-my-zsh or not.

mcornella:
> and it only sourced files ending `.zsh`

Oh yes that's true.

> If there's any suggestion as to how I can implement what I desire (creating a repo for dotfiles that includes my zshrc as well as `${ZSH}/custom`) I'd certainly appreciate it, whether it involves changing oh-my-zsh or not.

You can set the ZSH_CUSTOM variable to the path of the dotfiles directory, then add [those lines](https://github.com/robbyrussell/oh-my-zsh/pull/8034/files#diff-55b046be8e081f97f37b7667e1c4bfa9R133-R135) to `~/.zshrc`. The OMZ directory can be created with a simple `git clone`, so you can automate this easily.

b-:
Great, that helps a lot, and you're right, it only took a couple changes to what I already did. Thanks!

PR #8048, opened 2019-07-30
core: remove inc_append_history option

B-Galati:
The doc says it is useless to have both:

>This option both imports new commands from the history file, and also causes your typed commands to be appended to the history file (the latter is like specifying INC_APPEND_HISTORY, which should be turned off if this option is in effect)

http://zsh.sourceforge.net/Doc/Release/Options.html


curusarn:
Wouldn't it be better to keep current behavior and to remove `SHARE_HISTORY` option instead?

Issue #7985, opened 2019-07-09
schrodingers VIRTUAL_ENV_DISABLE_PROMPT 


thenoid:
**Describe the bug**
the `virtualenv` command generates a `bin/activate` script that modifies `$PS1` to include the virtual environment name.

oh-my-zsh does something similar with `prompt_virtualenv`

each script decides whether to display virtualenv information based on the `VIRTUAL_ENV_DISABLE_PROMPT` variable. 

the virtualenv's activate script uses the `-z` test (string is null, that is, has zero length).  oh-my-zsh uses the `-n` (string is not null)....  not sure how to craft a variable that is both null and not null.

```
prompt_virtualenv () {
    local virtualenv_path="$VIRTUAL_ENV"
    if [[ -n $virtualenv_path && -n $VIRTUAL_ENV_DISABLE_PROMPT ]]
    then
        prompt_segment blue black "(`basename $virtualenv_path`)"
    fi
}
```

```
if [ -z "${VIRTUAL_ENV_DISABLE_PROMPT-}" ] ; then
    _OLD_VIRTUAL_PS1="${PS1-}"
    if [ "x" != x ] ; then
        PS1="${PS1-}"
    else
        PS1="(`basename \"$VIRTUAL_ENV\"`) ${PS1-}"
    fi
    export PS1
fi
```
**To Reproduce**
1. Install oh-my-zsh
2. Install python2's virtualenv
3. create a virtual env `virtualenv test`
4. activate virtualenv `source test/bin/activate`
5. No matter what value you put for `VIRTUAL_ENV_DISABLE_PROMPT` always end up with a munged prompt

**Expected behavior**
When VIRTUAL_ENV_DISABLE_PROMPT is set, don't display the virtual env in PS1

**System:**
 - OS: OSX 10.14.5
 - Zsh version zsh 5.3 (x86_64-apple-darwin18.0)
 - Terminal emulator: iterm2
 - Virtualenv version: /usr/local/bin/virtualenv --version 16.6.1
 - Python version: Python 2.7.16

mcornella:
You're right, we need to change the flag in agnoster to `-z`. Can you put together a PR?

thenoid:
Sure!  Sorry for the delay - really need to figure out where my GitHub notifications are getting filtered to.

PR #8050, opened 2019-07-30
Bug fix for schrodingers VIRTUAL_ENV_DISABLE_PROMPT (7985)

thenoid:
This PR fixes the logic bug here: https://github.com/robbyrussell/oh-my-zsh/issues/7985

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -204,7 +204,7 @@ prompt_dir() {
 # Virtualenv: current working virtualenv
 prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
-  if [[ -n $virtualenv_path && -n $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
+  if [[ -n $virtualenv_path && -z $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
     prompt_segment blue black "(`basename $virtualenv_path`)"
   fi
 }


mcornella:
Thanks!

mcornella:
Hmm actually on closer inspection, it looks like the old way was specially designed so that it would only show virtualenv info *if and only if* the virtualenv built-in prompt was disabled. Otherwise, setting `VIRTUAL_ENV_DISABLE_PROMPT` would disable both the built-in and the agnoster prompt. See: https://github.com/robbyrussell/oh-my-zsh/pull/2184#issuecomment-27846156, https://github.com/robbyrussell/oh-my-zsh/pull/2239, https://github.com/robbyrussell/oh-my-zsh/commit/59c1ec80aa9b018db34078e69a27e57aea5ad761#r34515377.

thenoid:
@mcornella hmmm, seems like perhaps we need improved logic?  As-was there was no way to disable the prompt entirely. Which I wanted to I could display the venv in my `RPROMPT`.

Da-Juan:
I can confirm, if the virtualenv plugin is enable, no virtualenv displays in the prompt anymore and if I disable the plugin, I get both virtualenv prompts.

The virtualenv plugin sets `VIRTUAL_ENV_DISABLE_PROMPT` to 1
https://github.com/robbyrussell/oh-my-zsh/blob/6d2221e697ea4a2445202b1c32283610f84585ef/plugins/virtualenv/virtualenv.plugin.zsh#L7

thenoid:
@mcornella Perhaps we should revert this and discuss what would be appropriate logic?  Being forced into having one-or-the-other or none-or-both seems bad. 

Perhaps oh-my-zsh should use a different boolean?  or inspect the value of `VIRTUAL_ENV_DISABLE_PROMPT`?  Since python-virtualenv is just checking for existence, oh-my-zsh could check if it is set to `disable` or something and not enable it's own virtualenv plugin?  any other value it does?

jlucier:
I updated to the newest oh-my-zsh last night and now I'm getting the same behavior as described above (both prompts or nothing). Any fix?

mcornella:
I've reverted the change. Perhaps we should just check if the value is 1, instead of just checking if it has a value.

thenoid:
Yeah, representing it with Boolean logic isn't enough here - we need
 1. Disable python prompt, disable OMZ prompt
 1. Disable python prompt, enable OMZ prompt
 1. Enable python prompt, disable OMZ prompt
 
Rather than `1` I'd personally prefer something very explicit like `"YES_DISABLE_OMZ_PROMPT_TOO"` so we don't collide with people that are setting it to `1` currently.

thenoid:
Okay - New PR: https://github.com/robbyrussell/oh-my-zsh/pull/8063

Issue #4365, opened 2015-09-18
How to use docker-compose plugin


berdyshev:
Hi, I have enabled `docker` and `docker-compose` plugins. Here is the line in ~/.zshrc

```
plugins=(git github history symfony2 docker docker-compose)
```

`docker` completions works by tab, but `docker-compose` doesn't work. What I'm doing wrong?

darkwinternight:
Funny enough, for me it's the other way round.

evsign:
Not working too

mcornella:
Try `rm ~/.zcompdump*` and restarting the shell when enabling a plugin that provides completion.

evsign:
Thats work! thanks!

PR #8051, opened 2019-07-31
Mention FAQ in the README

butla:
I couldn't make docker-compose completions to work. I wondered if I misconfigured something in ZSH. Turned out, I wasn't the only one with that problem: #4365
I thought that if this is a thing that happens with plugins, it should be mentioned in the README part dealing with enabling plugins, as it would be nice if it had all the info you need to get going.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ ZSH_THEME_RANDOM_CANDIDATES=(
 
 ### FAQ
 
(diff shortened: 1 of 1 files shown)


mcornella:
This is mentioned in the [FAQ section](https://github.com/robbyrussell/oh-my-zsh/wiki/FAQ#i-have-enabled-a-completion-plugin-but-the-completion-doesnt-work). I see we don't have a section in the README pointing them to the FAQ, having that would be better. Can you take care of that?

butla:
@mcornella OK, I've changed my commit to mention the FAQ. Does it look ok?

PR #8054, opened 2019-08-02
A couple of new git aliases

Favna:
File changes pretty much say everything there is to say. `grau` and `grad` are just there because the `upstream` and `downstream` names are common usecase along with `origin`.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -192,6 +192,8 @@ alias gpv='git push -v'
 
 alias gr='git remote'
 alias gra='git remote add'
+alias grau='git remote add upstream'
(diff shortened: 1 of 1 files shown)


Favna:
Since this PR is seemingly never going to get merged and I wanted to add more aliases I have released my own version of the plugin to supercede the build in one: https://github.com/favware/zsh-git-enhanced

mcornella:
Hi Jeroen, we already have a lot of aliases and these commands don't seem to be very mainstream. Thank you for your contribution though! You can add your plugin to the [external plugins wiki page](https://github.com/ohmyzsh/ohmyzsh/wiki/External-plugins).

PR #8056, opened 2019-08-03
Add hg to af-magic, and tweak the virtualenv support a bit

matthewarmand:
I've been using the `af-magic` theme for a while now and I love it, just had a couple tweaks inspired by my workflow that I thought would be widely useful.

- Added mercurial support to af-magic, so now the vcs prompt will show in either a git or hg repository
- The virtualenv prompt was white and bumped up against the user@hostname output. Fixed that so its green (which I thought highlighted it more thematically) and has a space before user@hostname

Before virtualenv fix:
image

After virtualenv fix:
image

cc @mcornella as last one to touch this theme and @andyfleming as original theme-writer


matthewarmand:
@mcornella or others, does anything else need to happen with this in order for it to be merged?

andyfleming:
Looks good to me 👍

PR #8062, opened 2019-08-07
Allow plugins to be loaded from custom configurations

ndrone:
Moved the existing pluging loading set to a function, and added an unset.
The unset is so that if there are plugins defined in custom configurations the
previous plugins don't get double loaded. Then called the load_plugins function
for every custom configuration file.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -36,6 +36,17 @@ is_plugin() {
     || test -f $base_dir/plugins/$name/_$name
 }
 
+load_plugins() {
+  for plugin ($plugins); do
+    if [ -f $ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh ]; then
+      source $ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh
+    elif [ -f $ZSH/plugins/$plugin/$plugin.plugin.zsh ]; then
+      source $ZSH/plugins/$plugin/$plugin.plugin.zsh
+    fi
+  done
+  unset $plugins
+}
+
 # Add all defined plugins to fpath. This must be done
 # before running compinit.
 for plugin ($plugins); do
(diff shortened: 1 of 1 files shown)


ndrone:
When can this be merged?

mcornella:
I feel this complicates unnecessarily the init process and I can't see a reason behind it. Discuss otherwise.

ndrone:
In the bottom of the `.zshrc` file states

```
# Set personal aliases, overriding those provided by oh-my-zsh libs,
# plugins, and themes. Aliases can be placed here, though oh-my-zsh
# users are encouraged to define aliases within the ZSH_CUSTOM folder.
# For a full list of active aliases, run `alias`.
#
# Example aliases
# alias zshconfig="mate ~/.zshrc"
# alias ohmyzsh="mate ~/.oh-my-zsh"
```
I found that while trying to load plugins from the ZSH_CUSTOM folder. That I couldn't load them the same way as the .zshrc file `plugins(git)` they had to be loaded `source ${ZSH}/plugins/autojump/autojump.plugin.zsh`. Why I find this necessary is that you are asking users to use the ZSH_CUSTOM folder to load their custom themes and aliases, but not their plugins the same way as defined in the .zshrc file. This pull request makes it so that I as a user don't have to change the .zshrc file at all, and can completely customize my zsh settings via the ZSH_CUSTOM folder.

Thoughts?

mcornella:
> This pull request makes it so that I as a user don't have to change the .zshrc file at all, and can completely customize my zsh settings via the ZSH_CUSTOM folder.

Why do that though? What's the goal in not having to modify zshrc?

ndrone:
```
# Set personal aliases, overriding those provided by oh-my-zsh libs,
# plugins, and themes. Aliases can be placed here, though oh-my-zsh
# users are encouraged to define aliases within the ZSH_CUSTOM folder.
```
With the statement above, why are we encouraged to define our aliases with the ZSH_CUSTOM folder? And if I am going to set aliases there why not do all my aliases, functions, and plugins there as well there? And if it's not truly the way to work in oh-my-zsh why have the statement 
`though oh-my-zsh users are encouraged to define aliases within the ZSH_CUSTOM folder.`

PR #8063, opened 2019-08-07
Allow disabling virtualenv prompt in agnoster theme

thenoid:
Bug fix for schrodingers VIRTUAL_ENV_DISABLE_PROMPT (7985)

This PR fixes the logic bug here: https://github.com/robbyrussell/oh-my-zsh/issues/7985

This is version two of the fix.  v1: https://github.com/robbyrussell/oh-my-zsh/pull/8050

We need to support three use cases:

 1. Disable python prompt, disable OMZ prompt
 2. Disable python prompt, enable OMZ prompt
 3. Enable python prompt, disable OMZ prompt

This can be achieved by:

 1. Setting `$VIRTUAL_ENV_DISABLE_PROMPT` to `"YES_DISABLE_OMZ_PROMPT_TOO"`
 1. Setting `$VIRTUAL_ENV_DISABLE_PROMPT` to anything else
 1. Unsetting `$VIRTUAL_ENV_DISABLE_PROMPT`

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -204,7 +204,7 @@ prompt_dir() {
 # Virtualenv: current working virtualenv
 prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
-  if [[ -n $virtualenv_path && -z $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
+  if [[ -n $virtualenv_path && -n $VIRTUAL_ENV_DISABLE_PROMPT ]]; then
     prompt_segment blue black "(`basename $virtualenv_path`)"
   fi
 }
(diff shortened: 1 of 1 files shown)


mcornella:
If your goal is to be able to disable the virtualenv prompt in agnoster separately from the other variables, I think something like `AGNOSTER_DISABLE_VIRTUALENV` or a variant of it seems more fitting.

mcornella:
Though maybe `prompt_virtualenv(){}` just does the trick, as we've been using for `prompt_context`. A dirty hack but temporary if we're meant to make it mirror [the official repo](https://github.com/agnoster/agnoster-zsh-theme).

bartekpacia (changes requested):
Thank you for your contribution.

This looks fine to me, but I'd like to ask you to follow our [PR template](https://github.com/ohmyzsh/ohmyzsh/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and responds to Marc's comments above.

robbyrussell:
Thanks for this, and sorry for the long wait. The ask at the time was for a plain `AGNOSTER_DISABLE_VIRTUALENV` toggle rather than overloading `VIRTUAL_ENV_DISABLE_PROMPT` with a magic value, and it never got picked back up... the theme has changed a fair bit since, so I'd rather close it than leave it sitting.

PR #8065, opened 2019-08-08
cargo: dynamically generate sub-command list

Frederick888:
Tested against cargo 1.36.0 & 1.38.0 under Linux.


mcornella:
We're using the [upstream completion](https://github.com/rust-lang/cargo/blob/master/src/etc/_cargo), you should direct these changes to the project.

PR #8066, opened 2019-08-09
kube-ps1: Add an option to hide the namespace when value is `default`

eherot:


mcornella:
Please make the changes in the [upstream project](https://github.com/jonmosco/kube-ps1).

PR #8068, opened 2019-08-09
Add option to change shell quietly

bagrat:
Provide an option to skip the `change default shell?` prompt. It would be done with the addition of `ask` value of `CHSH`:

`no` - the default shell will not be changed
`ask` (default, current behavior) - the installer will prompt for a yes/no choice
`yes` (or anything else) - the installer will change the default shell without asking

This would be useful when you are running the installer from inside another script and don't want to have any prompts (like quiet run).

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -20,7 +20,7 @@
 #   BRANCH  - branch to check out immediately after install (default: master)
 #
 # Other options:
-#   CHSH    - 'no' means the installer will not change the default shell (default: yes)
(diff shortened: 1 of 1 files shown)


mcornella:
What about using 
```
if [ -t 0 ]; then
   # ask 
fi
```
It checks to see if stdin is open, will that work?

larson-carter:
@bagrat Can you please resolve these merge conflicts that way me and Marc can take another look at this PR?

PR #8069, opened 2019-08-10
Polishing the emoji plugin

lemeb:
Hi everyone 👋, and thank you so much for this kick-ass project! 

The emoji plugin hasn't been updated in about a year, and the emojis that it was supporting were pretty outdated. I spent the afternoon trying to update it, but in the process, I ended up re-writing significant parts of the plugin. Let me know what you think of it.

Here is the (relatively complete) list of changes:
1. **`emoji-data.txt`**: The previous file was very old, and has been replaced by the [current](https://unicode.org/Public/emoji/12.0/emoji-test.txt) Emoji spec (Emoji 12, January 2019). This new spec, however, was formatted very differently from the previous one. This led me...
2. **New `update_emoji` script**: ... to rewrite the parsing script in Python (I am more comfortable with Python than Perl). The mechanism is the same, but I added a few things to it.
3. **Keeping up with Emoji 12**: To state the obvious, _a lot_ has changed in the emoji world over the past few years. This means that the current Emoji spec is much more complex than before. 
   - **Redundant Emojis**: Consider, for instance, that the same emoji can be expressed with different code points. To take an extreme example, the emoji "Woman detective" can be expressed with four (!) different combinations of unicode symbols. (🕵️‍♀️,🕵‍♀️,🕵️‍♀,🕵‍♀ for the record.) In that kind of case, the Unicode consortium determines that one of these code points is the gold standard ("full-qualified", in their parlance), and the other not so much ("minimally-qualified", "unqualified", and so on.) I made sure that the script took that into account. Concretely, all the code points are registered in the `$emoji` array, but with a qualification attached to their names if necessary. (`$emoji[woman_detective] $emoji[woman_detective_unqualified] $emoji[woman_detective_unqualified_1] $emoji[woman_detective_unqualified_2]`, in this case.)
   - **Skin tone / Hair style**: Integrating skin tone and hair style is relatively straightforward. One is still able to make custom combinations by hand, but it gets tricky quickly. All the listed combinations on the Unicode spec are here, though. Moreover, I took the liberty of removing some experimental variables that were planning for this feature (`$emoji2`, for instance).
   - **Non-compatible systems**: Not every OS has been updated with the new emojis, so there will be some weird stuff on your screen if you try to display the newest additions. But hopefully, with this integration, Emoji 12 will be available to the torrent of new macOS users who will be coming this fall.
4. **Automated `$emoji_groups`**: Previously, the `$emoji_groups` associative array, the one that triages emojis into groups for the `display_emoji` command, was hand-coded in `emoji.plugin.zsh`. I found it much easier to automate the process, so these groups are now defined in the python script.
   - **Three groups added**: In addition to programmatically define the emoji groups, I added two new ones: flags, hands, and people. But the latter two are pretty massive in size, which defeats the purpose of the `display_emoji <group>` command a little...
    - **Still more work**: There is still work to be done on which groups to define though. The classification until now was OK when the number of emojis was manageable, but with almost 4,000 emojis nowadays, the classification should be re-thought.
   - **But there is more data**: On the plus side, the current Emoji spec classifies everything into groups and subgroups, which means more granularity to classify if we want to.
5. **Gemoji integration**: It was listed a relative priority on the README, and it was relatively easy to do. I downloaded their JSON database, which is now located at `gemoji_db.json`. Every time an emoji is considered by the script, it looks into that database, and if it finds a nickname, an alias that was not already in the spec, it takes it into account (lines 155-158, 184-187 of `update_emoji.py`). Concretely, that means that the same emoji can be accessed through multiple labels: 😋, for instance, is `$emoji[face_savoring_food]` and `$emoji[yum]`.
6. **Country Codes**: Also high on the priority list, adding country codes as aliases for flags. I added them with the help of an external Python library, `country_converter`. It is an annoying library with a lot of logging, but it works well and has almost every flag covered. This, in turn, means that 🇲🇿, for instance, can be accessed through multiple labels: `$emoji_flags[Mozambique] / [MZ] / [MOZ] / [mozambique]`
7. **README**: I updated the README accordingly, although I tried to leave as untouched as possible.

And I think that's it! Tell me what you think. 

Thank you so much for making this plugin in the first place, I really just polished it.


bartekpacia:
Hi @lemeb, thank you for your work🎉

Sorry for not responding, we have a ton of issues/PRs over here. 
I used `omz pr test 8069` to test this PR and it seems to work fine, but a long time has passed since this was submitted.

I'd like to ask you to:
- make sure this works fine
- follow our [PR template](https://github.com/ohmyzsh/ohmyzsh/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

mcornella:
Merged in 0613232202b92349cee79804fc46e0495152ef12. TO DO: fix update script which removes 2 and 3-letter code flag emojis. Example:
```diff
...
 emoji_flags[st_vincent_grenadines]=$'\U1F1FB\U1F1E8'
 emoji_flags[Venezuela]=$'\U1F1FB\U1F1EA'
-emoji_flags[VE]=$'\U1F1FB\U1F1EA'
-emoji_flags[VEN]=$'\U1F1FB\U1F1EA'
 emoji_flags[venezuela]=$'\U1F1FB\U1F1EA'
 emoji_flags[British_Virgin_Islands]=$'\U1F1FB\U1F1EC'
-emoji_flags[VG]=$'\U1F1FB\U1F1EC'
-emoji_flags[VGB]=$'\U1F1FB\U1F1EC'
 emoji_flags[british_virgin_islands]=$'\U1F1FB\U1F1EC'
 emoji_flags[U_S_Virgin_Islands]=$'\U1F1FB\U1F1EE'
-emoji_flags[VI]=$'\U1F1FB\U1F1EE'
-emoji_flags[VIR]=$'\U1F1FB\U1F1EE'
...
```

PR #8073, opened 2019-08-12
kubectl: add aliases for kustomization and explain

mapshen:
- add kustomization aliases
- add the explain alias
- misc cleanup


DocEmmetBrown (approved):

DocEmmetBrown:
These changes would be super helpful, especially now that Kustomize is getting more traction.

robbyrussell:
Thanks for this, and sorry it waited so long. The kustomize half landed in 2024 as `kapk` and `kdelk`, so merging this would give us two spellings of the same thing... and `kk` and `ke` on their own don't quite clear the bar we hold new aliases to.

PR #8074, opened 2019-08-12
autojump: fix autojump loading

mapshen:
`commands[autojump]` block errs out when autojump is not found, and the rest, which is intended to be used for loading `autojump`, does not get executed.

### DIFF
--- a/plugins/autojump/autojump.plugin.zsh
+++ b/plugins/autojump/autojump.plugin.zsh
@@ -1,8 +1,3 @@
-(( $+commands[autojump] )) || {
-  echo '[oh-my-zsh] Please install autojump first (https://github.com/wting/autojump)'
-  return
-}
-
 declare -a autojump_paths
 autojump_paths=(
   $HOME/.autojump/etc/profile.d/autojump.zsh         # manual installation


mcornella:
autojump needs to be in the path, otherwise the functions that the autojump scripts define will throw an error. They don't "load" the autojump command.

mapshen:
@mcornella Not sure if there is a misunderstanding.

I believe what this plugin does is to find a valid `autojump.sh` file and source it.

```
declare -a autojump_paths
autojump_paths=(
  $HOME/.autojump/etc/profile.d/autojump.zsh         # manual installation
  $HOME/.autojump/share/autojump/autojump.zsh        # manual installation
  $HOME/.nix-profile/etc/profile.d/autojump.sh       # NixOS installation
  /run/current-system/sw/share/autojump/autojump.zsh # NixOS installation
  /usr/share/autojump/autojump.zsh                   # Debian and Ubuntu package
  /etc/profile.d/autojump.zsh                        # manual installation
  /etc/profile.d/autojump.sh                         # Gentoo installation
  /usr/local/share/autojump/autojump.zsh             # FreeBSD installation
  /opt/local/etc/profile.d/autojump.sh               # macOS with MacPorts
  /usr/local/etc/profile.d/autojump.sh               # macOS with Homebrew (default)
)

for file in $autojump_paths; do
  if [[ -f "$file" ]]; then
    source "$file"
    found=1
    break
  fi
done
``` 

The lines above loop over all the possible locations of `autojump.sh`, which adds `autojump` to `$PATH`.

```
 head autojump.zsh
export AUTOJUMP_SOURCED=1

# set user installation paths
if [[ -d ~/.autojump/bin ]]; then
    path=(~/.autojump/bin ${path})
fi
if [[ -d ~/.autojump/functions ]]; then
    fpath=(~/.autojump/functions ${fpath})
fi
```

If `autojump` is already in `$PATH,` it seems that we don't need this plugin at all?

mcornella:
You're right, I overlooked that part of the script. It looks like the check for autojump in the path was added in #801 7 years ago. [At that time](https://github.com/wting/autojump/blob/390bca4291cb814ade7d3d94f924a70ee9af8739/autojump.zsh) the script didn't yet change the path so it made sense, but as you pointed out, it is no longer necessary. Thanks for sticking to your guns 👍

PR #8083, opened 2019-08-16
Updated ripgrep completion to latest version

jounathaen:
As before, just the official completion.

### DIFF
--- a/plugins/ripgrep/README.md
+++ b/plugins/ripgrep/README.md
@@ -8,6 +8,6 @@ To use it, add `ripgrep` to the plugins array in your zshrc file:
 plugins=(... ripgrep)
 ```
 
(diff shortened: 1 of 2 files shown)


plugins/ripgrep/README.md
```diff
@@ -10,4 +10,4 @@ plugins=(... ripgrep)
 
 Completion is taken from the ripgrep release [`0.10.0`](https://github.com/BurntSushi/ripgrep/releases/tag/0.10.0).
```
kaykayehnn: This link should also be updated to point to the latest release [`11.0.2`](https://github.com/BurntSushi/ripgrep/releases/tag/11.0.2)

jounathaen: Good find. I have updated it

mcornella:
Thanks!

PR #8088, opened 2019-08-18
feat(git): add `grbo` alias for `git rebase --onto`

pexison:


plugins/git/README.md
```diff
@@ -10,6 +10,7 @@ plugins=(... git)
 
 ## Aliases
 
+
```
pexison, line 13: this has to go

PR #8089, opened 2019-08-19
Aliases for git switch and restore

jjanvier:
With git 2.23, the `checkout` command has been divided into `switch` and `restore`. Which is much more clear. 

This PR creates aliases for those 2 commands.

More info here https://github.blog/2019-08-16-highlights-from-git-2-23/.


anaelChardan:
Thanks @jjanvier, :+1:

jjanvier:
Could somebody explain me why the CI is red please? 
I don't understand :thinking:

mcornella:
It's an issue on GitHub's end, you can ignore it.

plugins/git/README.md
```diff
@@ -131,7 +131,9 @@ plugins=(... git)
 | grmc                 | git rm --cached                                                                                                               |
 | grmv                 | git remote rename                                                                                                             |
 | grrm                 | git remote remove                                                                                                             |
+| grs                  | git restore                                                                                                                   |
 | grset                | git remote set-url                                                                                                            |
+| grs                  | git restore --source                                                                                                          |
```
mcornella: Should read `grss`

mcornella:
I'm hesitant to allow these yet since [they're marked as experimental](https://github.blog/2019-08-16-highlights-from-git-2-23/#experimental-alternatives-for-git-checkout).

----
Nevermind, it looks like the names are here to stay, only the parameters *could* change: https://github.com/git/git/commit/4e43b7ff1ea4b6f16b93a432b6718e9ab38749bd.

plugins/git/README.md
jjanvier: fixed!

jjanvier:
@mcornella we can wait a little bit if you prefer. As you wish, no problem :)

The 2.23.0 has been officially released a few days ago https://github.com/git/git/releases.
So I guess, it _shouldn't_ change.

jjanvier:
Thanks @mcornella :)

PR #8098, opened 2019-08-22
Update aws.plugin.zsh

paleite:
The plugin wasn't working, because it couldn't find profiles that were actually available.

This fix finds an exact matching line, which achieves the same intended behaviour.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -11,7 +11,7 @@ function asp() {
   fi
 
   local available_profiles=($(aws_profiles))
-  if [[ -z "${available_profiles[(r)$1]}" ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
Can you post what `aws_profiles` returns?

mcornella:
I realize that may contain sensitive info, just the format of what it returns is enough for me. Example:
```
profile-one
profile-two
```

paleite:
```
my-company-user
my-company-dev
my-company-test
my-company-prod
eb-cli

```

mcornella:
This isn't a problem in my end, the extra newline just gets ignored and the `available_profiles` array is correctly formed, so that the conditional works correctly. I'm going to close this PR until there is evidence of a problem.

Issue #7432, opened 2018-11-23
History Wrapper in omz_history breaks with zsh nounset and no options are passed


dpremy:
The history wrapper function omz_history throws an error if nounset is set within your zsh environment and you use history without passing any options or numbers. Using history and passing any numbers or options will work as expected, but when no parameters are passed the current using of $@ is undefined and throws an error.

This started in commit 5a729f66672d84f63c9f295008387d675a87b795 due to the use of ${@[-1]} in line 15.

I have a pull request in process to resolve this. This will be my first PR so please let me know if I've messed something up.

1. How to reproduce the problem.
```
% set -o nounset
% history
omz_history:13: @[-1]: parameter not set
```

2. What the correct behavior should be
```
% set -o nounset
% history
..... (list of your history commands)
```

dpremy:
Sourcing this file directly also give an error for an unset parameter so I will include this fix as well.

```
% . lib/history.zsh
lib/history.zsh:20: HIST_STAMPS: parameter not set
```

PR #8101, opened 2019-08-23
typo fixed

padlyuck:
HIST_STAMPS config property ignoring fixed


mcornella:
This was introduced in #7433 to fix #7432, it's actually there on purpose and it serves to use `""` if `HIST_STAMPS` wasn't defined.

PR #8102, opened 2019-08-23
plugins/git: Fix `gbda` trying to delete worktree branches

rmstyrczula:
Git learned to add a `+` in front of branches that are
checked out in other worktrees.

See: https://github.com/git/git/blob/745f6812895b31c02b29bdfe4ae8e5498f776c26/Documentation/RelNotes/2.23.0.txt#L252-L256


mcornella:
👍

PR #8103, opened 2019-08-23
homestead: repair sed regex

pOmelchenko:
I'm so sorry. i broke the plugin but now it' work correct

### DIFF
--- a/plugins/homestead/homestead.plugin.zsh
+++ b/plugins/homestead/homestead.plugin.zsh
@@ -1,6 +1,6 @@
 # Homestead basic command completion
 _homestead_get_command_list () {
-  homestead --no-ansi | sed "1,/(Available|Common) commands/d" | awk '/^ +[a-z]+/ { print $1 }'
+  homestead --no-ansi | sed "1,/[Available|Common] commands/d" | awk '/^ +[a-z]+/ { print $1 }'
 }
 
 _homestead () {
--- a/plugins/homestead/homestead.plugin.zsh
+++ b/plugins/homestead/homestead.plugin.zsh
@@ -1,6 +1,6 @@
 # Homestead basic command completion
 _homestead_get_command_list () {
-  homestead --no-ansi | sed "1,/[Available|Common] commands/d" | awk '/^ +[a-z]+/ { print $1 }'
+  homestead --no-ansi | sed -E "1,/(Available|Common) commands/d" | awk '/^ +[a-z]+/ { print $1 }'
 }
 
 _homestead () {


pOmelchenko:
```
➜  ~ homestead --no-ansi | sed "1,/[Available|Common] commands/d" | awk '/^ +[a-z]+/ { print $1 }'
box
cloud
destroy
global-status
halt
help
hostmanager
init
login
package
plugin
port
powershell
provision
push
rdp
reload
resume
snapshot
ssh
ssh-config
status
suspend
up
upload
validate
version
winrm
winrm-config
➜  ~ 
```

mcornella:
Hmm I would've guessed that syntax was correct, turns out [we have to escape the parentheses and or operator](https://superuser.com/a/112000/545959) but that is not cross-platform compatible; sed is hilariously bad at that if we look at [the rest of the answers](https://superuser.com/questions/111996/use-of-alternation-in-seds-regex). 

There's one answer that says the macOS's sed has the option `-E` so that we don't have to escape the parentheses and or operator, and in my sed version (GNU) it also works, so we have two possibilities:
```
sed -E '1,/(Available|Common) commands/d'
sed '1,/\(Available\|Common\) commands/d'

Your version will match any one-letter that is equal to what is inside the square brackets, so it's just a coincidence that it works. Can you test the above commands and also tell me your system info?

**Comment by pOmelchenko** [comment]
Yes of course.

I never used `sed` with regex and did not suspect that they may differ from those used in php or js.

system: macOS 10.14.6 (18G87)

```
Last login: Sat Aug 24 00:33:46 on ttys000
➜  ~ homestead --no-ansi | sed -E '1,/(Available|Common) commands/d'
     box             manages boxes: installation, removal, etc.
     cloud           manages everything related to Vagrant Cloud
     destroy         stops and deletes all traces of the vagrant machine
     global-status   outputs status Vagrant environments for this user
     halt            stops the vagrant machine
     help            shows the help for a subcommand
     hostmanager     plugin: vagrant-hostmanager: manages the /etc/hosts file within a multi-machine environment
     init            initializes a new Vagrant environment by creating a Vagrantfile
     login           
     package         packages a running vagrant environment into a box
     plugin          manages plugins: install, uninstall, update, etc.
     port            displays information about guest port mappings
     powershell      connects to machine via powershell remoting
     provision       provisions the vagrant machine
     push            deploys code in this environment to a configured destination
     rdp             connects to machine via RDP
     reload          restarts vagrant machine, loads new Vagrantfile configuration
     resume          resume a suspended vagrant machine
     snapshot        manages snapshots: saving, restoring, etc.
     ssh             connects to machine via SSH
     ssh-config      outputs OpenSSH valid configuration to connect to the machine
     status          outputs status of the vagrant machine
     suspend         suspends the machine
     up              starts and provisions the vagrant environment
     upload          upload to machine via communicator
     validate        validates the Vagrantfile
     version         prints current and latest Vagrant version
     winrm           executes commands on a machine via WinRM
     winrm-config    outputs WinRM configuration to connect to the machine

For help on any individual command run `vagrant COMMAND -h`

Additional subcommands are available, but are either more advanced
or not commonly used. To see all subcommands, run the command
`vagrant list-commands`.
➜  ~ 
```

```
Last login: Sat Aug 24 00:34:22 on ttys000
➜  ~ homestead --no-ansi | sed '1,/\(Available\|Common\) commands/d'
➜  ~ 
```

mcornella:
Ok then, let's go with `-E`. Doesn't work on Solaris as it seems but what does anyway...

PR #8110, opened 2019-08-28
agnoster.zsh-theme/git_prompt: red prompt if git repo is ahead/unpush…

ferbar:
Change prompt to red if repository contains unpushed changes. (atm it is green in this case)

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -104,13 +104,16 @@ prompt_git() {
     local LC_ALL="" LC_CTYPE="en_US.UTF-8"
     PL_BRANCH_CHAR=$'\ue0a0'         # 
   }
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry for the long silence. Agnoster shows this now — c189e8b added a branch status indicator that marks ahead, behind or both, and it's on by default.

Closing since it's covered. Thanks again.

ferbar:
@robbyrussell: Thanks a lot for your work!
after 100 years of using git, I still forget the git push every now and then.

My change resulted in this when there are unpushed commits:
image

this is the current behavoir for unpushed commits (just the icon is changed):
image

I just see green = work done :-)

robbyrussell:
Appreciate the quick response!

PR #8119, opened 2019-08-30
git: add glumff for glum with --ff-only

leighmcculloch:
When pulling or merging from `upstream/master` into a local `master` branch it can often be desirable and best practice to maintain a local `master` that purely tracks `upstream/master`, and to do all fork work in branches.

If I've accidentally committed to local `master` when I pull from `upstream/master` I don't want to start creating merge commits, I want to be warned of the situation so I can clean it up and keep my local `master` pure.


leighmcculloch:
The check failure looks like a problem with the GitHub Action, not a result of the change in the PR.

leighmcculloch:
Closing this as I no longer need it. I found a better way that suits my workflow really well which is to instead set in my `gitconfig` the `pull.ff = only` value which makes all `git pull`s use the `--ff-only` option.

PR #8127, opened 2019-09-03
Kubectl command HPA in kubectl plugin

rahultagadghar:
Added horizontal pod autoscaler command to kubectl plugin

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -133,3 +133,5 @@ alias kepvc='kubectl edit pvc'
 alias kdpvc='kubectl describe pvc'
 alias kdelpvc='kubectl delete pvc'
 
+# List all horizontal pod autoscaler
+alias kgh='kubectl get hpa'


rahultagadghar:
@robbyrussell

kbakdev (approved):

laststem:
It's been a while since pr was created, and there are command differences.
Can i improve this pr with adding command by create new pr?
@53jk1 @mcornella

53jk1:
Pull this branch, and create your own PR with this one if you want to :)

robbyrussell:
Thanks for this, and sorry it sat here so long. `kgh` doesn't quite follow the shape the rest of the plugin uses — the siblings all spell out the resource abbreviation, so this would want to be `kghpa`. #10249 covers the same ground following that convention, so I'm closing this one in favor of it.

robbyrussell:
Quick correction to what I said above: we decided not to take the HPA aliases at all, so #10249 is closed now too. It wasn't that your version needed different naming... the plugin's alias count is the thing we're holding the line on. Sorry for the mixed message.

Issue #8105, opened 2019-08-24
git branch completion not working for git switch


kleinfreund:
**Describe the bug**
When typing `git switch ` and pressing <kbd>Tab</kbd> after the space, the tab completion produces a list of file system objects of the current working directory rather than a list of git branch names (i.e. what happens when pressing <kbd>Tab</kbd> after typing `git checkout `).

**To Reproduce**

1. Enable plugin gitfast
2. Go to a directory that is a git repository with more than one local branch
3. Type `git switch ` (including the space at the end) and press <kbd>Tab</kbd>

**Expected behavior**

It shows a list of local branch names

**Actual behavior**

It shows a list of file system objects

**System:**
 - OS: Ubuntu
 - Zsh version 5.5.1
 - Terminal emulator gnome-terminal

**Additional context**

Requires git version 2.23.

It looks like the the git completion files need to updated (e.g. the gitfast plugin’s completion files are not up to date).

**Workaround**

Run the following commands. They place the completion files from the git repository into the `.zsh` directory.

```sh
mkdir -p ~/.zsh
curl -o ~/.zsh/git-completion.bash https://raw.githubusercontent.com/git/git/master/contrib/completion/git-completion.bash
curl -o ~/.zsh/_git https://raw.githubusercontent.com/git/git/master/contrib/completion/git-completion.zsh
```

Add the following lines to your `.zshrc` file.

```
# Load completion files from the ~/.zsh directory.
zstyle ':completion:*:*:git:*' script ~/.zsh/git-completion.bash
fpath=(~/.zsh $fpath)

autoload -Uz compinit && compinit
```

felipec:
The proper way to fix this is to backport the changes from v2.23. The pull request introduced more issues.

kleinfreund:
This has stopped working for me at some point in the last days. The behavior is now back to how it was initially reported.

edrpls:
Latest version also broke this for me, I am using the git and gitfast plugins, now none work.

citosid:
@felipec do you know the status of this issue? It seems to still be broken, however the bug is closed.

mcornella:
The gitfast plugin was updated to support oficial git completion. Make sure you have the latest OMZ version and open a new issue with details if you're still having issues.

felipec:
> @felipec do you know the status of this issue? It seems to still be broken, however the bug is closed.

Should be fixed since December 2020.

jaklan:
[...]

PR #8130, opened 2019-09-04
plugins/gitfast: update completions

Prajjwal:
Includes completions for `git branch`.

Tested on `git version 2.23.0`.

Fixes: #8105

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -30,7 +30,7 @@ if [ -z "$script" ]; then
 	local -a locations
 	local e
 	locations=(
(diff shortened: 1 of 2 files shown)


Prajjwal:
Hi @mcornella,

Can you take a look at this?

mcornella:
Thanks!

felipec:
@mcornella I reverted to upstream v2.16 for a reason. If we use v2.23 a lot more issues could arise. The best way to fix this issue is to backport whatever changes are needed.

In addition, if we update the completion, it should be through the `update` script. I wrote it precisely to avoid synchronization issues, which were introduced with this pull request.

This pull request reverts changes already fixed in `updates.patch`.

PR #8131, opened 2019-09-04
Remove local variable in robbyrussell theme

de1acr0ix:
Now it can be sourced in a function, like zplug.


nv035674:
@mcornella out of curiosity why did you remove the extra space in 0671e9e?

I just got the update and I personally prefer the old look in iTerm2:

Screen Shot 2019-09-13 at 11 10 55 AM

Would you be opposed to adding that space back in?

mcornella:
I thought it looked uneven. I guess it can be considered a breaking change so if you want to submit a quick PR I'll be fast to merge it.

nv035674:
Opened #8148. I agree, it's a matter of opinion and looks better in some places than others.

PR #8132, opened 2019-09-06
Fixed error in git-prompt.sh

AddisonG:
An error was thrown ( `bash: [: =: unary operator expected` ) when using the `__git_ps1_colorize_gitstring` function outside of the one place it's called (line 512), because the `$detached` variable was not quoted, and was unset.

### DIFF
--- a/plugins/gitfast/git-prompt.sh
+++ b/plugins/gitfast/git-prompt.sh
@@ -219,7 +219,7 @@ __git_ps1_show_upstream ()
 		if [[ -n "$count" && -n "$name" ]]; then
 			__git_ps1_upstream_name=$(git rev-parse \
 				--abbrev-ref "$upstream" 2>/dev/null)
-			if [ $pcmode = yes ] && [ $ps1_expanded = yes ]; then
+			if [ "$pcmode" = "yes" ] && [ "$ps1_expanded" = "yes" ]; then
 				p="$p \${__git_ps1_upstream_name}"
 			else
 				p="$p ${__git_ps1_upstream_name}"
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks @AddisonG.

felipec:
@AddisonG First, you did not explain why you need to call `__git_ps1_colorize_gitstring` outside `__git_ps1`.

Second, the fix could easily be one line of code that adds `"$detached"`; there's no need to modify $pcmode, or anything else.

Third, any custom changes should be added to `updates.patch` in order to avoid synchronization issues in the future. Even better would be to send those changes to Git upstream so we don't have to apply them every time we update the scripts.

@mcornella I am the maintainer of the gitfast plugin, why am I not consulted on these changes?

AddisonG:
@felipec - Sorry for creating strife here.

I admit, the change quoted a lot more variables than it may have needed to, but I find that the safest and best practice (the majority will agree) is to _always_ quote variables before comparison.

These are some of the reasons I always quote everything:
- It prevents mistakes that may be missed when someone does something foolish or unusual (spaces in file-path's is a common example)
- It ensures that code is re-usable outside of very specific contexts
- It allows code to be extended and modified, without other parts of the program breaking

felipec:
@AddisonG Sure, but we are borrowing this code from Git upstream. The more changes we make the harder it will be to keep track of those changes when we want to update from upstream.

If you want to quote all the variables I would suggest you send a patch to Git upstream.

And you still haven't explained why you want to call `__git_ps1_colorize_gitstring` separately.

AddisonG:
@felipec - I wanted to extend/borrow functionality, for a personal solution.

How can I update this upstream? Is there another branch/repo? I'm new to open source - please don't bite me.

felipec:
@AddisonG You can send your patch to git@vger.kernel.org, however, you will need to explain everything that I already asked: why do you want to call `__git_ps1_colorize_gitstring` separately?

Issue #8137, opened 2019-09-08
yarn plugin add yarn version


frani:
**Is your feature request related to a particular plugin or theme? If so, specify it.**

is for Yarn Plugin

**Is your feature request related to a problem? Please describe.**

There is no alias for yarn version

**Describe the solution you'd like**

add an alias for yarn version

**Describe alternatives you've considered**

**Additional context**

mcornella:
Closing, please do not open an issue if you have a PR ready.

PR #8138, opened 2019-09-08
Yarn Plugin - add yarn version alias

frani:
#8137  yarn plugin - add yv alias for "yarn version"
@robbyrussell


mcornella:
Can you add the alias to the README?

PR #8142, opened 2019-09-11
Add a note where to put the `PROMPT=...` line

jonasholtkamp:
Found this while investigating why the prompt wouldn't show up in my Terminal.

The change in the contributor section wasn't intended; guess this is a Github issue.


mcornella:
I put the extra note in the settings example so that there aren't two contradicting lines.

PR #8143, opened 2019-09-11
git: add gfroh for fetch and reset to origin

desimaniac:
We use this a lot to catch up to origin:

`git fetch && git reset --hard @{u}`

(`@{u}` shortcut comes from [here](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html))

which is equivalent to the alias in this PR:

`git fetch origin && git reset origin/$(git_current_branch) --hard`

So we hope it gets added in :)


mcornella:
I think chaining together a couple of aliases (`gf && groh`) isn't that much of a hassle to warrant another alias in the git plugin. You can still define the alias in your zshrc file though.

PR #8150, opened 2019-09-16
add flutter plugin

lesnitsky:


mcornella:
Can you add a basic README documenting what it does and what aliases it defines? The [npm plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/npm/README.md) is a good example.

lesnitsky:
@mcornella readme added

mcornella:
Good job thanks!

PR #8151, opened 2019-09-16
timer: threshold to show timers only for time-consuming commands

Frederick888:


mcornella:
Hey I know this is a bit old, but can you add this setting to the README?

Frederick888:
@mcornella Done. Btw i also changed `-gt` to `-ge` in `${tdiff} -ge "${TIMER_THRESHOLD}"`, so that `TIMER_THRESHOLD=0` can make some sense.

mcornella:
LGTM. Thanks!

PR #8156, opened 2019-09-19
bar.zsh-theme

anki-code:
deleted

### DIFF
--- a/themes/bar.zsh-theme
+++ b/themes/bar.zsh-theme
@@ -30,7 +30,7 @@ export ZSH_FIRST=1
 
 statusbar () {
   # Sections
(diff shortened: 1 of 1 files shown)


anki-code:
> [We can't accept more themes](https://github.com/robbyrussell/oh-my-zsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/robbyrussell/oh-my-zsh/wiki/External-themes) wiki page. You can also add your theme as a custom theme. See [customization](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes) for instructions.

Got it.

anki-code:
@mcornella I've reviewed all themes in zsh pack and external. I found nothing like my concept. Most of themes looks like christmas tree and I've found this so noizy for attention. Could you please show me the theme which has visual features like in my list above? If there is no theme like this I suggest to add bar theme to the main pack! 😃 So I'm going to reopen the PR back ☕️

PR #8155, opened 2019-09-18
added support for sysctl

927589452:
this is just copy paste and move from acpi to sysctl for FreeBSD

PR #8157, opened 2019-09-19
Update README.md for FreeBSD

emilengler:
Tested it and using it.
Works like a charm.
Also macOS is based on FreeBSD which just proves this

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ To learn more, visit [ohmyz.sh](https://ohmyz.sh) and follow [@ohmyzsh](https://
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


927589452:
though a few plugins need tweaks for freebsd

See for example #8155

mcornella:
I clarified on which platforms it works and got rid of the disclaimer since it's double information. I added you as [co-author of the commit](https://github.com/robbyrussell/oh-my-zsh/commit/a6d71ddadb5c9c1f0ded8ecce228dcde1a23ad05) but it doesn't appear correctly for some reason. Thanks!

PR #8160, opened 2019-09-20
Added ecr_login function to aws plugin

kujon:
Hi 👋 

I thought that logging in to ECR is a common enough use case and in my case quite hard to remember by heart that it could be an additional function in the plugin.

Let me know what you think.


kujon:
@mcornella

kbakdev (approved):

robbyrussell:
Thanks for this, and sorry it sat here so long without a reply. AWS CLI v2 removed `aws ecr get-login` back in 2020, so this doesn't work on any current CLI.

PR #8161, opened 2019-09-20
Jira Plugin: Use last part of branch (basename).

jhonny007:
We divide our branches in fixes and features:
-  feature/IOT-4461
-  feature/IOT-4490
-  fix/IOT-4643

To open  jira with the right path only the basename is needed.

The change is transparent to existing behaviour, because if you add anything except jira issue number it cannot be opened anyway.

PR #8162, opened 2019-09-21
Jira: Add command to search branch name as is

ASittingLapp:
Add branch-as-is command to take branch name without applying prefix. My team puts our prefix in branch names as a convention but I still want a prefix set for looking up other tickets. 

By adding a new command I also won't conflict with other PRs wanting to change the beviour of the branch command, such as [jhonny007's](https://github.com/robbyrussell/oh-my-zsh/pull/8161).

### DIFF
--- a/plugins/jira/README.md
+++ b/plugins/jira/README.md
@@ -17,14 +17,15 @@ This plugin supplies one command, `jira`, through which all its features are exp
 ```
 jira            # performs the default action
 
-jira new        # opens a new issue
-jira dashboard  # opens your JIRA dashboard
+jira new          # opens a new issue
+jira dashboard    # opens your JIRA dashboard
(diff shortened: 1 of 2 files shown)


mcornella:
Can you put some example branches with / without prefixes and the URL it should hit?

ASittingLapp:
When I made this PR I didn't realize the `jira branch` command was checking if the prefix was already present. The actual issue I was seeing was that the check was case sensitive, so if I had my "PREFIX-1234" branch and my `$JIRA_PREFIX` is "PREFIX" then it's fine, but a branch named "prefix-1234" would try and open "PREFIX-prefix-1234". 

I've made a new PR that tackles the actual problem I was seeing, found here: #8799

PR #8163, opened 2019-09-21
plugins/do_sudo: alias expansion handling for sudo

Frederick888:
Usually we may add `alias sudo='sudo '` to `.zshrc` so that shell can expand the following alias for us. This shouldn't be on by default cos of security reasons but it's indeed quite handy. One problem about it tho is that if the following alias is preceded by e.g. `nocorrect` `noglob` sudo will complain that the command cannot be found.

There was a workaround for this from [Zsh mailing list](https://www.zsh.org/mla/users/2008/msg01229.html) and I reckon it can be quite helpful to have it in oh-my-zsh. I removed the hard-coded `-u` support and `/bin/zsh` fallback so that the alias can work more similarly as `sudo` in the first commit.

The second commit is what I'm actually not that sure about. I tried to cover some fringe cases but ended up with something that tends to be overly complex. Compared with the original one, it can handle all sudo options and recursively expand aliases until it find a real command. For example:
```zsh
$ alias vim='nvim'
$ sudo -u michael -g dunder-mifflin vim  # not supported by the original one

$ alias xvim='nvim'
$ alias yvim='noglob xvim'
$ alias vim='yvim'
$ sudo vim *  # noglob is effective

$ alias vim=vim1
$ alias vim1=vim2
$ alias vim2=vim
$ sudo vim --version  # prevent infinite loop
```
It gets most of my use cases covered but not sure whether it can be simplified a little?

### DIFF
--- a/plugins/do_sudo/do_sudo.plugin.zsh
+++ b/plugins/do_sudo/do_sudo.plugin.zsh
@@ -10,6 +10,7 @@ function _do_sudo() {
     local -a args
     local -a cmd_alias_arr
     local cmd_alias
+    local return_value
     while (($#)); do
         case "$1" in
         command|exec|-) shift; break ;;
@@ -43,8 +44,10 @@ function _do_sudo() {
         else
             PATH="/sbin:/usr/sbin:/usr/local/sbin:$PATH" command sudo "${args[@]}" $==*
         fi
+        return_value=$?
         unset __do_sudo_glob
         unset __do_sudo_expanded
+        return $return_value
     fi
 }
 
--- a/plugins/do_sudo/do_sudo.plugin.zsh
+++ b/plugins/do_sudo/do_sudo.plugin.zsh
@@ -17,14 +17,10 @@ function _do_sudo() {
         noglob) __do_sudo_glob=0; shift ;;
         *)
             cmd_alias="$(command -v 2>/dev/null -- "$1")"
-            if [[ "$?" -eq 0 ]]; then
-                if [[ "$cmd_alias" == 'alias'* ]] && [[ -z "$__do_sudo_expanded["$1"]" ]]; then
-                    __do_sudo_expanded["$1"]=1
-                    IFS=' ' read -A cmd_alias_arr <<< "$(sed -e "s/[^=]*=//" -e "s/^'//" -e "s/'$//" <<< "$cmd_alias")"
-                    args+=( "${cmd_alias_arr[@]}" )
-                else
-                    args+=( "$(sed "s/[^=]*=//" <<< "$(hash -v 2>/dev/null -- "$1")")" )
-                fi
+            if [[ "$?" -eq 0 ]] && [[ "$cmd_alias" == 'alias'* ]] && [[ -z "$__do_sudo_expanded["$1"]" ]]; then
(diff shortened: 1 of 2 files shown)


TriMoon (changes requested):
https://github.com/ohmyzsh/ohmyzsh/pull/8163/commits/3f4f4adce52220ddd95fc42e44e985e6b881328f :-1: 
In terms of functionality and logic, this is a **null-edit** because it does not change anything...
It only increases the size of the script unnecessarily.

plugins/do_sudo/do_sudo.plugin.zsh
```diff
@@ -38,13 +38,14 @@ function _do_sudo() {
     if [[ ${#cmd_alias_arr[@]} -gt 0 ]]; then
         _do_sudo "${args[@]}" $==*
     else
+        unset __do_sudo_expanded
         if ((__do_sudo_glob)); then
+            unset __do_sudo_glob
```
TriMoon: this

plugins/do_sudo/do_sudo.plugin.zsh
```diff
@@ -38,13 +38,14 @@ function _do_sudo() {
...
             PATH="/sbin:/usr/sbin:/usr/local/sbin:$PATH" command sudo "${args[@]}" $~==*
         else
+            unset __do_sudo_glob
```
TriMoon: and this

plugins/do_sudo/do_sudo.plugin.zsh
```diff
@@ -38,13 +38,14 @@ function _do_sudo() {
...
             PATH="/sbin:/usr/sbin:/usr/local/sbin:$PATH" command sudo "${args[@]}" $==*
         fi
-        unset __do_sudo_glob
```
TriMoon: Do the same as this was doing...

TriMoon (changes requested):
Unneeded changes

plugins/do_sudo/do_sudo.plugin.zsh
Frederick888: Any suggested solution? Before this change the function always returned 0, which was the exit code of `unset`. A workaround firstly popped up in my mine was to save the exit code of `command sudo ...` in a variable and return it in the end but that's 1 more line than this one.

TriMoon: ohhh i see now what you mean.
I didn't catch the fact you were running  a command on those lines after the ones i mentioned...
Those two lines look almost identical, thats why i didn't catch the 1 char diff :smile_cat:

Frederick888: No prob. Just force pushed a new commit using a local variable to handle it :)

kbakdev (approved):

robbyrussell:
Thanks for this, and sorry nobody replied for six years. The bug you found is real.

Wrapping the global `sudo` alias in a recursive function is more machinery than we want around that particular command, though, and nobody else has asked for it in the time this has been open.

Closing for that reason... it's a good fit for `$ZSH_CUSTOM`. Thanks again.

PR #8170, opened 2019-09-24
react-native: add aliases for newer iPhones

KevinHu2014:
Added aliases for `iPhone 11`, `11 Pro`, `11 Pro Max`


mcornella:
Thanks!

PR #8173, opened 2019-09-25
update git plugin for new restore and switch commands

mrgrauel:
* add alias for `git restore` and `git switch` which has been added to git in version 2.23
* https://github.blog/2019-08-16-highlights-from-git-2-23/


mcornella:
They were added in https://github.com/robbyrussell/oh-my-zsh/pull/8089.

PR #8174, opened 2019-09-25
Feature/plugin git merge develop

mrroot5:
Added because develop is a common branch in git-flow workflow and other workflows.

This PR is an improvementto this: https://github.com/robbyrussell/oh-my-zsh/pull/7592

- Improvement: added `git merge --log origin/develop`


robbyrussell:
Thanks for this, and sorry it sat so long.

We've declined the `develop` aliases a few times... #7592 (which this one builds on) for hardcoding the branch name, and #12742 more recently because the git plugin is already bloated and this fits one specific flow.

They work well in `$ZSH_CUSTOM`. Thanks again for contributing.

Issue #8154, opened 2019-09-18
git-flow plugin doesn't support the "bugfix" flow


Edo78:
**Is your feature request related to a particular plugin or theme? If so, specify it.**
It's related to the git-flow plugin

**Is your feature request related to a problem? Please describe.**
The plugin doesn't provide autocompletion for the 'bugfix' flow

**Describe the solution you'd like**
It should be nice if the plugin can provide autocompletion for all the git-flow flows

Id doesn't look too complicated to add this flow to the plugin so I can be able to make a PR if you are open to consider it (obviously after reading your guideline for collaboration).

mcornella:
See https://github.com/ohmyzsh/ohmyzsh/pull/8176#issuecomment-582040288.

PR #8176, opened 2019-09-26
add support for 'bugfix' flow in the 'git-flow' plugin

Edo78:
Just added the `bugfix` flow support as I request in the issue #8154 
Let me know if there's something to fix


mcornella:
[The original `git-flow` tool](https://github.com/nvie/gitflow) doesn't support bugfix, you're thinking of the [AVH Edition](https://github.com/petervanderdoes/gitflow-avh), which requires the [`git-flow-avh` plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/git-flow-avh/README.md).

Edo78:
Ops, my bad. Sorry for the confusion.

PR #8185, opened 2019-09-30
added readme for thor and pip plugins

Kullsno2:


mcornella:
I also documented the lesser known pip plugin cache as well. Thanks for your contribution!

Kullsno2:
> I also documented the lesser known pip plugin cache as well. Thanks for your contribution!

Always here to contribute boss!

PR #8191, opened 2019-10-01
Add bindings for Ctrl+Backspace and Ctrl+Delete

### LINKED ISSUE #7609: Why don't Ctrl+Backspace and Ctrl+Delete work?
It is standard in UI applications to have Ctrl+Backspace delete a word before the cursor (`backwards-kill-word`) and Ctrl+Delete delete a words after the cursor (`kill-word`). In Oh-My-ZSH I can use Alt+Backspace and Ctrl+W for the former and there doesn't seem to be an alternative for the latter. Why is that?

It is very easy to implement the standard behavior, I could just add these lines to `lib/key-bindings.zsh`:
```
bindkey '^H' backward-kill-word
bindkey '^[[3;5~' kill-word
```
But before doing so, I'm interested whether there is a reason not to add this.

m93a:
Delete words with Ctrl+Backspace and Ctrl+Delete, like it's standard in GUI applications. Fixes #7609.

### DIFF
--- a/lib/key-bindings.zsh
+++ b/lib/key-bindings.zsh
@@ -52,6 +52,9 @@ bindkey ' ' magic-space                               # [Space] - do history exp
 bindkey '^[[1;5C' forward-word                        # [Ctrl-RightArrow] - move forward one word
 bindkey '^[[1;5D' backward-word                       # [Ctrl-LeftArrow] - move backward one word
 
+bindkey '^[[3;5~' kill-word                           # [Ctrl-Backspace] - delete word backward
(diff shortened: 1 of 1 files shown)


BridgeAR:
I just want to point out that one of the key bindings already exist: https://github.com/ohmyzsh/ohmyzsh/pull/8191/files#diff-79ee521883d01c4cd6367cc6e962af8dR67

davidmogar:
@BridgeAR I'd remove that one in favor of the more standard one (remove the whole word).

PR #8196, opened 2019-10-02
Add Readme to cakephp3 plugin

Rexben001:
**What does this PR do?**
Add README to cakephp3 plugin

This contributes to #7175


mcornella:
Thanks!

PR #8199, opened 2019-10-03
Adding README for rvm

zwhitten:
This adds the README file for `rvm`, Ruby Version Manager, as part of #7175


mcornella:
Thanks!

PR #8201, opened 2019-10-03
Added autocomplete to docker-compose plugin

alexmanno:
Added service autocomplete for the following aliases on docker-compose plugin:
- dcb
- dce
- dcps
- dcrestart
- dcrm
- dcr
- dcstop
- dcup
- dcupd
- dcl
- dclf
- dcpull
- dcstart


mcornella:
The plugin now has the upstream docker-compose completion.

PR #8203, opened 2019-10-03
parameter not set error when using set -u option in interactive shell

archf:
Sometimes in an interactive shell I find myself using `set -u` or `set -e` options but `oh-my-zsh`  components are found exiting because of unbound variables (z plugin and oh-my-zsh git prompt):

```
archf@wm:~/d/oh-my-zsh master>
_z:5: _Z_OWNER: parameter not set                                                                                                                      
parse_git_dirty:8: DISABLE_UNTRACKED_FILES_DIRTY: parameter not set
```


mcornella:
The z changes should be directed upstream: https://github.com/rupa/z/. The other one is already fixed in master.

PR #8209, opened 2019-10-04
Add option to install OMZ without replacing .zshrc

### LINKED ISSUE #8208: Allow install without replacing .zshrc

I have an install script which installs my dotfiles for me on new machines. When extending it to install OMZ I realized that there is no option in the OMZ installer to disable replacing `.zshrc`.

I would like there to be either:
1. A command line option, perhaps `--noreplace-zshrc`, or
2. An environment variable, perhaps `REPLACE_ZSHRC`, which could be set to `no`

...or both, which would disable the backup and replacement of `.zshrc`.

The best workaround I have is to simply overwrite the generated `.zshrc` script after the fact, but since the OMZ installer prints a message about backing up and replacing `.zshrc`, it'll be confusing to anyone who uses the script. I could hide the output of the OMZ installer, but frankly I really like the way it looks and it'd be a shame to hide it.

mfdorst:
This makes `tools/install` respect the `REPLACE_RC` environment variable. If `REPLACE_RC='no'`, and a `.zshrc` exists, the installer will not replace it. If there is no `.zshrc` the variable will have no effect (a new `.zshrc` will still be generated).

This also adds the `--noreplace-rc` option, i.e. `tools/install --noreplace-rc`, which has the effect of setting `REPLACE_RC='no'`.

Closes #8208


mcornella:
+1, but it'd be better using this:
- `KEEP_ZSHRC=yes`
- `--keep-zshrc`

Also, if this option is passed, a minimal OMZ setup should be appended to the existing zshrc file. Something like this:
```sh
cat <<-EOF >> ~/.zshrc

export ZSH="$ZSH"
ZSH_THEME=robbyrussell
plugins=(git)
source $ZSH/oh-my-zsh.sh
EOF
```
I haven't checked that it works.

mfdorst:
@mcornella the use-case for this option (for me, anyway) is [my dotfiles installer], symlinks all my dotfiles, then installs OMZ. It has to be repeatable though, so I needed a setting in OMZ that would prevent it from touching my `.zshrc` file, which is already configured the way I want it. So I don't like your suggestion of appending a minimal OMZ setup.

[my dotfiles installer]: https://github.com/mfdorst/dotfiles/blob/master/install

mcornella:
Thanks Michael

githorse:
For anyone arriving here, like me, wondering why their `--keep-zshrc` (or other option) doesn't work, be sure that if you're using `sh -c`, you pass an empty string argument before the flags:
 
```
sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)" "" --keep-zshrc
```
(This is documented in the installer comments but not immediately obvious if you're using curl or whatever.)

Explanation: with `-c` (which uses a command string instead of stdin), `sh` interprets the second argument as the name of the command/script, i.e. `$0`. Remaining arguments are the standard positional parameters. In this case the script doesn't need `$0` so we can just use `""`.

mcornella:
[Also kinda documented in the README](https://github.com/ohmyzsh/ohmyzsh#unattended-install).

PR #8210, opened 2019-10-04
Added README describing mysql-macports plugin in oh-my-zsh

adityakaria:
### Adds a README for the [mysql-macports](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/mysql-macports) plugin 
Describes the aliases provided by the script for easy use of mysql on MacOS when installed using MacPorts

References issue #7175


mcornella:
Thanks!

PR #8213, opened 2019-10-05
Missing virtualenv on af-magic.zsh-theme

AchmadGozali8:
Missing position on virtualenv if activated.


mcornella:
You don't need that, when doing `bin/activate`, the virtualenv information will be automatically added to the prompt. But yeah, using the virtualenv plugin also works.

PR #8220, opened 2019-10-06
Requested alias #8216.

### LINKED ISSUE #8216: Add a git alias to clone and change directory
git

No

Add an alias which clones a repo and then `cd`s into it.

Manually `cd` into the directory after `git clone` or `gcl`.

**Additional context**

fanuch:
This function can `cd` after `clone` with and without the `.git` extension at the end
Can't be one with an alias, as it needs to take a parameter for both the `clone`, and then the following `cd`

Sourced from this [StackExchange question/answer](https://unix.stackexchange.com/questions/9123/is-there-a-one-liner-that-allows-me-to-create-a-directory-and-move-into-it-at-th)

Fixes #8216

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -64,6 +64,7 @@ alias gcam='git commit -a -m'
 alias gcsm='git commit -s -m'
 alias gcb='git checkout -b'
 alias gcf='git config --list'
+function gccd() { git clone $1; cd "$(basename $_ .git)" }
 alias gcl='git clone --recurse-submodules'
 alias gclean='git clean -id'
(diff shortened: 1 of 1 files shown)


mcornella:
This version would allow for specifying a target directory:
```zsh
function gccd {
  git clone --recurse-submodules "$@"
  [[ -d "$_" ]] && cd "$_" || cd "$(basename $_ .git)"
}
```
Also added the `--recurse-submodules` to mirror the `gcl` alias.

fanuch:
While it isn't directly in the vain of the original request, I can see that as a valuable addition!

53jk1 (approved):

PR #8221, opened 2019-10-06
docs: add README.md for Dash plugin

sk1u:
addresses part of https://github.com/robbyrussell/oh-my-zsh/issues/7175


mcornella:
Thanks!

PR #5826, opened 2017-01-28
add go-swagger completion plugin

GlenDC:
This PR adds the completion plugin for go-swagger.

More information can be found on the official repository of go-swagger: http://github.com/go-swagger/go-swagger

PR #8222, opened 2019-10-06
Add go-swagger autocompletion

titusjaka:
Add go-swagger autocompletion plugin.
The completion is taken from [official go-swagger repo](https://github.com/go-swagger/go-swagger/blob/master/cmd/swagger/completion/swagger.zsh-completion).

### DIFF
--- a/plugins/go-swagger/README.md
+++ b/plugins/go-swagger/README.md
@@ -1,0 +1,13 @@
+# Go-swagger plugin
+
+This plugin adds completion for [go-swagger](https://github.com/go-swagger/go-swagger).
+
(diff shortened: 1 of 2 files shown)


kbakdev (approved):

robbyrussell:
Thanks for this, and sorry for the very long delay.

We declined the same plugin back in #5826, and the reasoning still holds... go-swagger ships and maintains its own zsh completion, so vendoring a copy here means we'd be shipping a snapshot that drifts out of date. Anyone who wants it can source `cmd/swagger/completion/swagger.zsh-completion` from upstream directly, or drop it in `$ZSH_CUSTOM`.

Closing for that reason. Thanks again for taking the time.

PR #8223, opened 2019-10-06
Detect gems.rb in _rake_command

agorf:
Apart from `Gemfile`/`Gemfile.lock`, Bundler supports `gems.rb`/`gems.locked` since [version 1.8.0.pre](https://github.com/bundler/bundler/blob/1-8-stable/CHANGELOG.md#180pre-2015-01-26)

This commit tweaks the `_rake_command` to pick up `gems.rb` as well.


agorf:
A related PR that does a similar thing for the Bundler plugin #6594

PR #8224, opened 2019-10-06
Add documentation for pyenv plugin

joeratt:
Partially addresses #7175 

Adds a README for the `pyenv` plugin. Plugin already has documentation on the wiki: https://github.com/robbyrussell/oh-my-zsh/wiki/Plugins#pyenv


mcornella:
Thanks!

PR #8226, opened 2019-10-06
docs: add README.md for virtualenvwrapper

sk1u:
fixes part of https://github.com/robbyrussell/oh-my-zsh/issues/7175


mcornella:
Good docs, thanks!

PR #8230, opened 2019-10-07
#7175 add README.md for Terminalapp plugin

otherpaco:
#7175 add README.md for Terminalapp plugin


mcornella:
I just deleted the whole thing since it's been deprecated for so long.

PR #8232, opened 2019-10-07
#7175 Create README.md for Singlechar plugin

otherpaco:
#7175


mcornella:
I fixed the markdown formatting so it was readable in plain text and reworded some descriptions.

PR #8243, opened 2019-10-10
Fix showing aws prompt out of the box in af-magic

ToROxI:
Fix issue related to #7615, #7747 and #6346
After the update, aws prompt (which should be visible out of the box) disappears when a user uses a theme af-magic, because of fact that plugins are loaded before themes.
This pull request fixes issue with not showing aws prompt in theme af-magic, by appending RPROMPT in theme af-magic instead overwriting.


ToROxI:
@mcornella, could you please review this PR?

Thanks in advance!

PR #8251, opened 2019-10-12
Update alias-finder.plugin.zsh

ego:
Use `((` for numeric comparisons.


ego:
The very useful plugin, can you review please @SomeDer  @mcornella

mcornella:
This isn't necessary, but thanks.

ritzau:
This causes a bug on macOS Catalina. I solved it myself by replacing == for -eq, but this works as well. Please re-open and accept.

PR #8255, opened 2019-10-14
Better 'bira' theme and added functions

shriniwas26:
- Add a useful mkcd command
- Better working of bira theme with virtualenv

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -222,3 +222,5 @@ function omz_urldecode {
 
   echo -E "$decoded"
 }
+
+function mkcd() { mkdir -p "$@" && cd "$@";   }
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -2,13 +2,13 @@
 # export PATH=$HOME/bin:/usr/local/bin:$PATH
 
 # Path to your oh-my-zsh installation.
-export ZSH=$HOME/.oh-my-zsh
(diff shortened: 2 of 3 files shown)


mcornella:
* You shouldn't modify the zshrc template. Instead, you should copy it to `~/.zshrc` and make your modifications there.
* `mkcd` already exists with the name `take`.
* The bira theme change modifies the order of the displayed information, which is IMO a breaking change. Instead, you want to copy it to the custom folder and modify it there. See [Overriding and adding themes](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization#overriding-and-adding-themes)

shriniwas26:
Thanks for your reply. All your points are valid, but I did not understand
why the theme change would be breaking. Could you explain in a bit more
detail?
Best regards,
Shriniwas

mcornella:
Because it modified the order of display of the virtualenv. I thought the original had the virtualenv information shown before the user information, but it looks like this was a recent change. It does make sense to put it after the git prompt. Open a new PR with just the bira change please.

shriniwas26:
Done!
Best regards,
Shriniwas

--
Shriniwas Sharma
Software Developer, Sioux Embedded Systems, Eindhoven, The Netherlands
M.Sc. (Computer Science & Engineering), TU Eindhoven, The Netherlands
Phone & WhatsApp : +31 6 45872727

PR #8262, opened 2019-10-15
Print Git Remote status when details are not required

SscSPs:
Earlier the git_remote_status function would return nothing if "ZSH_THEME_GIT_PROMPT_REMOTE_STATUS_DETAILED" was not set to true in theme and other theme properties were not present.
This patch changes that behavious by returning the remote branch if details are not needed.

Signed-off-by: Sahil Soni <SscSPs@gmail.com>


kbakdev (approved):

robbyrussell:

The catch is that this changes what the prompt prints for anyone already using the non-detailed mode. The themes we ship that use it (michelebologna, intheloop) define their own ahead/behind symbols and render fine today, so this would add the remote branch name to their prompts without them asking for it. That's a change we'd rather not make on everyone's behalf.

If you want the remote name in your own prompt, overriding `git_remote_status` in `$ZSH_CUSTOM` gets you there and survives updates.

Closing for that reason... thanks again for the contribution.

PR #8268, opened 2019-10-16
Fix 'sudo: [cmd]: command not found' when using `_` alias

ShahinSorkh:
fix 'sudo: [cmd]: command not found' error when executing commands like `_ ff somefile -print`
(see http://superuser.com/questions/605417/ddg#605460)


ShahinSorkh:
Though it does not affect `_ -u someuser somealias`

PR #8272, opened 2019-10-16
Add alias for 'git add --interactive' for git plugin

moyuanhuang:
I sometimes use the big brother of `git add --patch` which is more powerful. Wonder if you are interested in adding this alias. Difference between `--patch` and `--interactive` can be found [here][article].

[article]: https://nuclearsquid.com/writings/git-add/


moyuanhuang:
Cleaning up my opened PRs.

PR #8281, opened 2019-10-18
Update README.md

finch185277:
- fix typos


mcornella:
That's proper English, there's a suttle distinction between both phrases. Thanks for your contribution anyhow.

PR #8290, opened 2019-10-19
Add Kustomize apply

jufab:
Add an alias "kak" for apply a kustomize file (kubectl apply -k)


jufab:
Add in Readme

kbakdev (approved):

truearken:
this got added in #13033

robbyrussell:

`kubectl apply -k` landed in #13033 as `kapk`, so this one is covered now. The different name is probably for the best... `kak` is Kakoune's command, so the alias would have shadowed it.

Closing since it's already in. Thanks for the contribution.

PR #8292, opened 2019-10-20
add readme for safe paste plugin

Netoun:
I've add a readme for safe-paste plugin with a description of this plugin

Refer #7175


mcornella:
Thanks

PR #8297, opened 2019-10-21
Add alias gdh='git diff HEAD~ HEAD'

DavidDeprost:
This is useful to quickly check which modifications the last commit added.


mcornella:
You can just do `gsh` for `git show`, which defaults to HEAD.

PR #8299, opened 2019-10-22
add hasura plugin

tristola:
When using hasura-cli there are some repeated commands that would be faster to use with aliases.

### DIFF
--- a/plugins/hasura/README.md
+++ b/plugins/hasura/README.md
@@ -1,0 +1,17 @@
+# Hasura
+
+this plugins provides aliases for frequent hasura-client commands.
+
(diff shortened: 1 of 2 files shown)


kbakdev (approved):

robbyrussell:

We ended up shipping a hasura plugin in #11278 in 2022 — a completion plugin, at the same path — so this can't be merged on top of it.

Worth mentioning for the future: `h` and `hc` are already taken by a few other plugins here (`history`, `helm`, `heroku`, `hcloud`), so single-letter aliases like those tend not to make it in regardless. They work well in `$ZSH_CUSTOM` if you want them.

Closing for that reason. Thanks for the contribution.

PR #8303, opened 2019-10-22
core: add title support for mlterm

joshuashaffer:
Add mlterm.


mcornella:
Thanks!

peterjschroeder:
mlterm should be mlterm* to accommodate mlterm-256color.

mcornella:
Fixed in 95218372402afc9aeafc89c0984310387ce79599.

PR #8304, opened 2019-10-22
[PLUGINS][GIT]: Adds `git branch -v` alias

theodorosidmar:
Hello there!

I'm a heavy user of `git branch -v`. So, I decided to create `gbv` alias. I hope ou enjoy it!


keithbrianptomo:
?

theodorosidmar:
@keithbrianptomo ? lol

53jk1 (approved):

carlosala:
Hi @theodorosidmar!
I think adding that alias does not add a lot of value, since you only avoid two keystrokes. Going from `gbv` to `gb -v` is almost the same stuff. Even though, if you use it a lot I recommend you add it to your zshrc.
Thanks for your contribution!

theodorosidmar:
I have it! Thank you for your feedback :)

PR #8305, opened 2019-10-23
Fix README, fix yolo, truncate useless symlink

45gfg9:
Nah I guess I just have to write descriptions

* `whatthecommit.com` doesn't appear to be an HTTPS site, and I kept getting "Aborted due to empty message" thing. `http://` just works for this.
* There is a symlink `per-directory-history.plugin.zsh` which pointed to `per-directory-history.zsh`. So I just made the symlink the file, and removed the symlink.
* `phing`'s README added `eecms` to plugin array.


keithbrianptomo:
?

mcornella:
I fixed the phing README, the rest of the changes aren't necessary. Thanks for your contribution.

PR #8309, opened 2019-10-24
add kubectl edit secret also

Moulick:
have to use kubectl edit secret many times...


Moulick:
@mcornella Hi, conflict resolved :)

kbakdev (approved):

Moulick:
@53jk1  @mcornella  who can merge this ?
@53jk1 Thanks for approving on the 1 Year anniversary of this MR  🎉

npetrovdx:
Hello, 
How we can have that change :pray:

robbyrussell:
Thanks for this, and sorry it sat for six years.

The alias was worth having... secret was the only resource in the plugin with get, describe and delete but no edit. I've reapplied it in #14048 with a `Co-Authored-By` trailer crediting you, moved after `kgseca` to match the ordering the other blocks use, and added the README row.

Nothing needed from you. This will close when that merges.

PR #8316, opened 2019-10-25
nvm: add Homebrew's nvm setup

### LINKED ISSUE #8307: Load nvm.sh in Homebrew install
nvm

The nvm plugin will call nvm.sh for you, but it points at the wrong directory for a Homebrew install.

I would like the plugin to check for the standard nvm.sh location in a Homebrew install, and source it.

I have a customized version of this plugin to do this, and I've seen others put the source directive in .zshrc.

**Additional context**
None

davidfmatheson:
Homebrew has a slightly different nvm setup, and this change checks for it and sources the right files.

Fix #8307 
@saschabratton

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -7,8 +7,11 @@ if ! type "nvm" &> /dev/null; then
     if [[ -f "$NVM_DIR/nvm.sh" ]];
         source "$NVM_DIR/nvm.sh"
     else
-        # Load nvm from Homebrew location if it exists
+        # User can set this if they have an unusual Homebrew setup
         [[ -z "$NVM_HOMEBREW" ]] && export NVM_HOMEBREW="/usr/local/opt/nvm"
+        # Load nvm from Homebrew location if it exists
         [[ -f "$NVM_HOMEBREW/nvm.sh" ]] && source "$NVM_HOMEBREW/nvm.sh"
+        # Load nvm bash completion from Homebrew location if it exists
+        [[ -f "$NVM_HOMEBREW/etc/bash_completion.d/nvm" ]] && source "$NVM_HOMEBREW/etc/bash_completion.d/nvm"
     fi
 fi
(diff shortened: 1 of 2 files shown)


davidfmatheson:
Fixes #8307

plugins/nvm/nvm.plugin.zsh
```diff
@@ -4,5 +4,14 @@
...
-    [[ -f "$NVM_DIR/nvm.sh" ]] && source "$NVM_DIR/nvm.sh"
+    if [[ -f "$NVM_DIR/nvm.sh" ]]; then
+        source "$NVM_DIR/nvm.sh"
+    else
+        # User can set this if they have an unusual Homebrew setup
+        [[ -z "$NVM_HOMEBREW" ]] && export NVM_HOMEBREW="/usr/local/opt/nvm"
+        # Load nvm from Homebrew location if it exists
+        [[ -f "$NVM_HOMEBREW/nvm.sh" ]] && source "$NVM_HOMEBREW/nvm.sh"
+        # Load nvm bash completion from Homebrew location if it exists
+        [[ -f "$NVM_HOMEBREW/etc/bash_completion.d/nvm" ]] && source "$NVM_HOMEBREW/etc/bash_completion.d/nvm"
```
steinybot: Why load bash completions for zsh?

davidfmatheson: The bash completions work in zsh, and there are no zsh-specific completions. I'm a bit new to zsh, though, are bash completions not typically sourced?

diegogsann:
currently this plugin isn't working for mac users using zsh and changes proposed by @davidfmatheson are working here.
@steinybot could you take a look and see if this change breaks your use case? what we need to go ahead with this change? do you think in a better way to do this?
thanks.

plugins/nvm/nvm.plugin.zsh
steinybot: To be honest I do not know if they are normally or not but I never do. I either use ZSH specific ones or none at all.

Just going from the name I would expect that there could be bash specific syntax in there. In saying that, I took a look at it and it appears as though they have specifically accounted for ZSH so seems like it is ok.

arlyon: Yeah having just tested this PR it seems to be no problem on zsh, haven't encountered any serious issues after some rigorous manual fuzzing

gabmontes:
@steinybot any plans to merge this soon? Thanks!

sihil:
Would also love to see this merged.

PR #8317, opened 2019-10-26
Adds README for codeclimate plugin

SSchwartz214:


mcornella:
This plugin only adds autocompletion for codeclimate. The way it is phrased it looks like it is the whole cli, but that must be installed separately. The documentation of the commands isn't necessary either. Can you change that?

PR #8319, opened 2019-10-26
agnoster: makes mercurial consistent with git on icon displays

bmarsaud:
I noticed that in the agnoster theme, the mercurial's icons for staged/unstaged changes and branches are not consistent with the git ones.
This PR fixs that.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -162,36 +162,42 @@ prompt_bzr() {
 
 prompt_hg() {
   (( $+commands[hg] )) || return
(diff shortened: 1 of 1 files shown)


TheZopo:
> ?

Hmm, can you be more specific ?

kbakdev (approved):

robbyrussell:
Thanks for this, and sorry nobody replied at the time.

`prompt_hg` has been rewritten since you opened this, so the patch no longer applies to the current theme.

Closing for that reason... thanks for taking the time on it.

PR #8326, opened 2019-10-28
feat(helm): cache completion file in the background

mostafahussein:
This PR updates the way of using `helm completion  zsh` command by creating a file that will be used as a cache to speed things up.


sagikazarmark:
Would love to get this feature!

53jk1 (approved):

Issue #8332, opened 2019-10-29
Old update.lock will block update


empire:
**Describe the bug**
If update was interrupted, lock will not be removed and block future updates.

**To Reproduce**

1. Wait for update prompt
2. Close terminal
3. Update will be block forever and update.lock will not removed

**Expected behavior**
Lock must be removed next time (or after some time)

**Screenshots or recordings**
https://asciinema.org/a/7KhfLzsSPWZgxK19npv6qiGmc

**System:**
 - OS: Manjaro (linux)
 - Zsh version zsh 5.7.1 (x86_64-pc-linux-gnu
 - Terminal emulator xterm

**Additional context**

empire:
Hi @mcornella
I like to work on this bug, could you please give some feedback on my pull request.

Thanks

mcornella:
https://github.com/ohmyzsh/ohmyzsh/commit/40b013f5f119be27bd2fdece431cf4979193bd25 gives a different solution, which is to run `upgrade_oh_my_zsh` and that will force a removal of the update.lock directory.

empire:
@mcornella, the problem is not solved by that commit (that just removes the lock directory) and needs some manual fix from a user but I think it can be handled automatically (without users notice).

jwarkentin:
Something needs to be done. I've had this issue for such a long time and only just discovered it was the `update.lock` directory preventing update prompts. Unfortunately, for anyone experiencing the issue they will simply have to eventually notice that it's not auto-updating and figure out how to force it manually before they can even get a future fix, assuming one is eventually made.

empire:
Another solution is to change the path of update.lock to a temporary location (/tmp as an example). This way we sure the last lock will be removed on reboot.

mcornella:
That won't work because `/tmp` isn't always volatile in all systems.

The bug in the auto-update code was already fixed, and a manual upgrade means that the code is fixed and the `update.lock` folder is removed. That said, I think we can also add a check to see if it is a day old and remove it in that case.

empire:
I change the locking strategy in the pull request and use the flock command instead.

PR #8333, opened 2019-10-29
Use flock to lock update.lock

empire:
If update process was interrupted, lock will not be removed and no any update will be occurred.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -29,18 +29,16 @@ fi
 # Cancel upgrade if git is unavailable on the system
 whence git >/dev/null || return 0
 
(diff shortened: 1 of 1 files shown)


empire:
Fix https://github.com/robbyrussell/oh-my-zsh/issues/8332

mcornella:
We should add a trap function to make sure the lock is removed even on CTRL+C and such. When running `upgrade_oh_my_zsh` the lock should also be removed if it exists. Let me work on an alternative.

mcornella:
[`flock` isn't available in Solaris](https://unix.stackexchange.com/a/13025), but we will use `trap` in combination with `mkdir`. Thanks!

PR #8338, opened 2019-10-30
Add Command List and Update Readme.md

kxgcayh:
With ubuntu-clinux function user doesnt need to open https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/ubuntu or file because they can see aliases in terminal!


53jk1 (approved):

mcornella:
I'm sorry for being blunt but this is wrong on many levels.

PR #8341, opened 2019-10-31
agnoster-theme: distinguish untracked from modified git prompt color

XavierAraujo:
When analyzing the git prompt color it is useful to know if the git status is dirty due to modified files or only due to untracked files. This PR tries to solve this issue by setting the git prompt color to yellow if there are untracked files but not modified files, and to red if there are modified files.


53jk1 (approved):

XavierAraujo:
Can you merge this PR @53jk1 ? 😂

53jk1:
Nope, I am not allowed to

PR #8347, opened 2019-11-01
extract: add RPM archive support

poettler-ric:
Not sure whether I should put the files in a newly created subdirectory.
@mcornella


mcornella:
If this extracts more than one file in the current working directory you can use the `$extract_dir` variable as a directory to extract to. Does it need that or is this ready to merge?

poettler-ric:
Thanks for the hint. Added `$extract_dir`. I think it's good now.

PR #8351, opened 2019-11-03
Support BSD mktemp in emacsclient.sh

dobbs:
Piping stdin to emacs alias on MacOS was breaking (--tmpdir is not
supported in BSD flavored mktemp).

Tested in MacOS 10.14 and debian:buster to confirm it still works in
linux.

```
ls -al | e -
ls $TMPDIR/emacs-stdin*
```

@mcornella, I've read these guidelines which ask contributors to copy anyone relevant & found you as the most recent contributor to this script:
https://github.com/robbyrussell/oh-my-zsh/blob/master/CONTRIBUTING.md#you-have-a-solution


mcornella:
👍 thanks!

Issue #5765, opened 2017-01-09
Completion waiting dots cause prompt to disappear with multi line prompts


chrishoage:
With completion waiting dots off:
![screen shot 2017-01-09 at 1 47 51 pm](https://cloud.githubusercontent.com/assets/1056886/21784998/b15156e8-d672-11e6-8699-b8fd80de909f.png)

With completion waiting dots on:
![screen shot 2017-01-09 at 1 48 08 pm](https://cloud.githubusercontent.com/assets/1056886/21784997/b1508fb0-d672-11e6-9ebe-67313db39311.png)

This issue is not specific to my custom theme. I have tried other multi line prompts and all have the same issue. 

In the meantime I have disabled `COMPLETION_WAITING_DOTS`

daneah:
I didn't experience this issue until upgrading to macOS High Sierra. I asked [this StackOverflow question](https://stackoverflow.com/questions/46440121/different-multiline-prompt-behavior-using-zsh-oh-my-zsh-iterm2-high-sierra) before finally tracing the issue to `COMPLETION_WAITING_DOTS` myself.

mcornella:
Can you test with this modification?
```
expand-or-complete-with-dots() {
    print -Pn "%{%F{red}......%f%}"

    zle expand-or-complete
    zle redisplay
}
zle -N expand-or-complete-with-dots
bindkey '^I' expand-or-complete-with-dots
```

daneah:
@mcornella I've just tried this and I'm still seeing the same behavior. I believe the `zle redisplay` is what causes this affliction, as commenting it out makes things behave roughly as I'd expect.

mcornella:
Sadly I can't reproduce this. Can you share your theme so I can try on my end? Also, what `zsh --version` are you running?

daneah:
I'm on `zsh 5.5.1 (x86_64-apple-darwin17.5.0)` on macOS 10.13.4.

My theme is a fairly heavily customized one, originally started from `kolo`. Here's [the gist](https://gist.github.com/daneah/6b85b4aabb4bbc6e806b0c443bfcf350).

daneah:
I attempted to make an `asciinema` recording of this behavior to help you see a little more clearly what's happening, but `asciinema` drops you into a subshell where it seems that the completion dots don't execute (and therefore the behavior is not exhibited). I don't quite understand why that is, but sorry if that observation is just noise 😅

shilov:
[...]

PR #8352, opened 2019-11-04
chore: caution against `COMPLETION_WAITING_DOTS` in template

sudo-suhas:
Enabling this setting causes issues for multiline prompts when tab
completion is triggered. Add a caution in template warning against the
same.
See https://github.com/robbyrussell/oh-my-zsh/issues/5765


53jk1 (approved):

mcornella:
The current version should work on multiline prompts, doesn't it?

mcornella:
Specified in the Settings wiki as well https://github.com/ohmyzsh/ohmyzsh/wiki/Settings#completion_waiting_dots

PR #8353, opened 2019-11-05
sharing this shortcut i've been using

levensailor:


mcornella:
We already have a few plugins for dealing with virtualenvs. Thanks for your contribution though! Feel free to create a new repository and add it to the [external plugins](https://github.com/ohmyzsh/ohmyzsh/wiki/External-plugins) wiki page.

PR #8361, opened 2019-11-06
Rename LICENSE.txt to LICENSE

DaniSpringer:
So GitHub shows it automatically in its License UI.


mcornella:
The reason GitHub's UI doesn't parse it correctly is because of the extra line below the copyright, but thanks for bringing this to our attention.

DaniSpringer:
Actually for my LICENSE, which had no extra text over than the standard one, the only thing holding it back - and which fixed it - was such a naming change.

🙏 Thanks
📱 On the App Store <https://apps.apple.com/us/developer/daniel-springer/id1402417666>
💻 GitHub <https://github.com/DaniSpringer>
🔗 LinkedIn <https://www.linkedin.com/in/imdanielspringer/>

PR #8372, opened 2019-11-11
osx: modify itunes function to support Apple Music

### LINKED ISSUE #8373: Cannot control Apple Music with `itunes`
**Describe the bug**

**To Reproduce**

1. Enable plugin `osx`
2. Run command `itunes vol`.
3. See error `39:45: execution error: The variable volume is not defined. (-2753)`

**Expected behaviour**

Control Apple Music volume.

**System:**
 - OS: _**macOS* Catalina 10.15.1_
 - Zsh version 5.7.1
 - Terminal emulator iTerm2

**Additional context**
In _**macOS** Catalina 10.15_, the music section of _iTunes_ was discontinued, its features split into _Music_, _Podcasts_, and _TV_. The behaviour of the _Music_ app is pretty much the exact same for music, including AppleScript commands.

jajaperson:
Change `itunes` to `music`, as iTunes was renamed to Music in _**macOS** Catalina 10.15_.

Pretty much all functionality is the same, except for changing the name and using AppleScript to communicate with `Music` instead of `iTunes`.

Perhaps both versions should be supported instead?

Fix #8373

### DIFF
--- a/plugins/osx/README.md
+++ b/plugins/osx/README.md
@@ -42,20 +42,20 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ## Commands
 
-| Command         | Description                                         |
-| :-------------- | :-------------------------------------------------- |
-| `tab`           | Open the current directory in a new tab             |
-| `split_tab`     | Split the current terminal tab horizontally         |
-| `vsplit_tab`    | Split the current terminal tab vertically           |
-| `ofd`           | Open the current directory in a Finder window       |
-| `pfd`           | Return the path of the frontmost Finder window      |
-| `pfs`           | Return the current Finder selection                 |
-| `cdf`           | `cd` to the current Finder directory                |
-| `pushdf`        | `pushd` to the current Finder directory             |
-| `quick-look`    | Quick-Look a specified file                         |
-| `man-preview`   | Open a specified man page in Preview app            |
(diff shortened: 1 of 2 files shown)


larson-carter:
@jajaperson Hello, does this PR fix this issue? https://github.com/robbyrussell/oh-my-zsh/issues/8373

jajaperson:
@larson-carter yep 👍

mcornella:
Hmm.. the upgrade path here is a bit tricky, we don't want people using an older macOS to be left out of the plugin. Can we make it configurable somehow, or even chosen automatically by finding the macOS version?

mcornella:
We could use [`sw_vers` to check the macOS version](https://www.cyberciti.biz/faq/mac-osx-find-tell-operating-system-version-from-bash-prompt/), or an [osascript to check if Music or iTunes are installed](https://stackoverflow.com/a/14297870/5798232). Based on this latest link, we could also check the appID, it might be the same for "iTunes" and "Music". Then, have a variable either set to "iTunes" or "Music" based on the results.

As far as changing the name of the function to music, we should smoothly deprecate the `itunes` name based on the results of the previous checks, creating an `itunes` alias to the music function and checking `$0`, displaying a deprecation notice.

mcornella:
I've pushed a fix to allow backwards compatibility with older versions than Catalina, can you check that it still works on Catalina?

jajaperson:
Yep, 👍 it still works.

PR #8374, opened 2019-11-11
Fix bad tmux config syntax and logical expression error in 86b39cf

yhchen0906:
1. The tmux command that allows you to include config file should be `source-file` not `source`.
2. since `$_ZSH_TMUX_FIXED_CONFIG` already includes the custom config with `source-file`, it is unnecessary to include the config manually with `tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")`.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -67,9 +67,8 @@ function _zsh_tmux_plugin_run() {
 
   # If failed, just run tmux, fixing the TERM variable if requested.
   if [[ $? -ne 0 ]]; then
-    [[ "$ZSH_TMUX_FIXTERM" == "true" ]] && tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG") || \
-    [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")
-    $tmux_cmd new-session  
+    [[ "$ZSH_TMUX_FIXTERM" == "true" ]] && tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG")
+    $tmux_cmd new-session
   fi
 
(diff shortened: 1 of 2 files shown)


plugins/tmux/tmux.plugin.zsh
```diff
@@ -67,9 +67,8 @@ function _zsh_tmux_plugin_run() {
 
   # If failed, just run tmux, fixing the TERM variable if requested.
   if [[ $? -ne 0 ]]; then
-    [[ "$ZSH_TMUX_FIXTERM" == "true" ]] && tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG") || \
-    [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")
```
mcornella: This line is there in case the fixed config isn't include, notice the extra `|| \` at the end of the previous line.

yhchen0906:
Really appreciate for your prompt reply!
This is actually another bug I didn't report in the former comment, since I thought that manually include the config is unnecessary. But now I understand what you trying to do.
While I was debugging this script, I found a interesting fact that there is no precedence of logical operators in zsh, so `true && echo "first" || false && echo "second"` will be interpreted as `( ( true && echo "first" ) || false ) && echo "second"`, I guess this is not the order you expected. I've update the script, please check it. Thanks for your help :D


Notice that adding the parenthesis directly will not work because it fails the variable append operation

```
# this won't work
( [[ "$ZSH_TMUX_FIXTERM" == "true" ]] && tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG") ) || \
( [[ -e "$ZSH_TMUX_CONFIG" ]] && tmux_cmd+=(-f "$ZSH_TMUX_CONFIG") )
```

Currently, I don't know how to achieve this through one-liner, so I decide to fallback to if-else statement. If you figured out how to done this, please tell me! Thanks a lot.

mcornella:
Thanks!

PR #8375, opened 2019-11-12
fix(spotify): fix exit on unknown command

PeterHeja:
When you run `spotify` with an unknown command, then the default case runs in the function, which exits with return code 1 (`exit 1`).
I think this is a not very nice behaviour as it closes my shell session when I mistype a command.
The trivial fix is to just `break`.


mcornella:
Thanks!

PR #8390, opened 2019-11-15
Fix / Export ZSH_TMUX_CONFIG again to provide to tmux

paulodiovani:
#8374 introduced a bug in tmux autostart.
After those changes, `~/.tmux.conf` was not being loaded anymore when setting `ZSH_TMUX_FIXTERM="true"`.

This is due to tmux removing variables that are not present (exported) in the source environment.

> Tmux' man page states that update-environment will remove variables "that do not exist in the source environment [...] as if -r was given to the set-environment command".
>
> -- https://unix.stackexchange.com/a/76256/27284

This can be checked in the terminal with:

```bash
$ echo $ZSH_TMUX_CONFIG                          
/home/diovani/.tmux.conf
$ grep ZSH_TMUX_CONFIG
# nothing                     
$ tmux show-environment -g | grep ZSH_TMUX_CONFIG
# nothing
``` 

This pull request fixed that by exporting again `ZSH_TMUX_CONFIG` before sourcing `tmux.extra.conf` (which will source `$ZSH_TMUX_CONFIG`).

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -46,6 +46,7 @@ fi
 
 # Set the correct local config file to use.
 if [[ "$ZSH_TMUX_ITERM2" == "false" && -e "$ZSH_TMUX_CONFIG" ]]; then
+  export ZSH_TMUX_CONFIG="$ZSH_TMUX_CONFIG"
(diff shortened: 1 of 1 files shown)


koxu1996:
I was wondering why my tmux config stopped working...

PR #8391, opened 2019-11-15
add alias gpot

fmigneault:
Adds the `gpot` alias to *only* push tags to origin.

Other existing `gpoat` alias has the side effect of also pushing *all branches* to remote origin, which is not always desired, for example, when you have many develop branches still being worked on.

Side edit, just aligned the table line of one entry that was misaligned.


robbyrussell:

We've turned down the push-with-tags aliases before, in #7931 and #10747.

Closing for the same reason. Thanks for contributing.

PR #8392, opened 2019-11-15
Update colorize to support less options

ProbstDJakob:
I improved the `colorize_via_pygmentize_less` methode to be able to use less options within `cless` and as an side effect, no temporary files have to be created.

### DIFF
--- a/plugins/colorize/README.md
+++ b/plugins/colorize/README.md
@@ -41,12 +41,12 @@ ZSH_COLORIZE_STYLE="colorful"
 ## Usage
 
 * `ccat <file> [files]`: colorize the contents of the file (or files, if more than one are provided).
-  If no arguments are passed it will colorize the standard input or stdin.
+  If no files are passed it will colorize the standard input.
 
-* `cless <file> [files]`: colorize the contents of the file (or files, if more than one are provided) and
-  open less. If no arguments are passed it will colorize the standard input or stdin.
+* `cless [less-options] <file> [files]`: colorize the contents of the file (or files, if more than one are provided) and open less.
+  If no files are passed it will colorize the standard input.
+  The LESSOPEN and LESSCLOSE will be overwritten for this to work, but only in a local scope.
 
-Note that `cless` will behave as less when provided more than one file: you have to navigate files with
-the commands `:n` for next and `:p` for previous. The downside is that less options are not supported.
-But you can circumvent this by either using the LESS environment variable, or by running `ccat file1 file2|less --opts`.
-In the latter form, the file contents will be concatenated and presented by less as a single file.
+## Requirements
+
+You have to install Pygments first: [pygments.org](https://pygments.org/download/)
--- a/plugins/colorize/README.md
+++ b/plugins/colorize/README.md
@@ -17,10 +17,10 @@ plugins=(... colorize)
 
 ### Requirements
 
-This plugin requires that either of the following tools be installed:
+This plugin requires that at least one of the following tools is installed:
 
-* Chroma: [https://github.com/alecthomas/chroma](https://github.com/alecthomas/chroma)
-* Pygments be installed: [pygments.org](https://pygments.org/)
+* [Chroma](https://github.com/alecthomas/chroma)
+* [Pygments](https://pygments.org/download/)
 
 ### Colorize tool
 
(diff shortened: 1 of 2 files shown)


ProbstDJakob:
If you merge this and #8393, you probaly should move line 6 to 25 (based on commit 5d5d202794ad62c51435798fdd464245f928e5db) in `colorize_via_pygmentize_check_requirements` to prevent `less` from being opened in case something is not correctly set or installed. Thus the previous lines in `colorize_via_pygmentize_check_requirements` should be removed.

mcornella:
@fzuviria can you review this?

robbyrussell:
Hi @ProbstDJakob -- I ended up merging  #8393. Would it be possible to rebase your PR?

ProbstDJakob:
@robbyrussell of course I can. I will do it tomorrow. Should be done within the next 24 hours.

ProbstDJakob:
Hi @robbyrussell the pull request is ready to merge

plugins/colorize/README.md
```diff
@@ -41,12 +41,12 @@ ZSH_COLORIZE_STYLE="colorful"
...
-  open less. If no arguments are passed it will colorize the standard input or stdin.
+* `cless [less-options] <file> [files]`: colorize the contents of the file (or files, if more than one are provided) and open less.
+  If no files are passed it will colorize the standard input.
+  The LESSOPEN and LESSCLOSE will be overwritten for this to work, but only in a local scope.
 
-Note that `cless` will behave as less when provided more than one file: you have to navigate files with
-the commands `:n` for next and `:p` for previous. The downside is that less options are not supported.
-But you can circumvent this by either using the LESS environment variable, or by running `ccat file1 file2|less --opts`.
-In the latter form, the file contents will be concatenated and presented by less as a single file.
+## Requirements
```
mcornella: The Requirements section is now above, this isn't needed.

plugins/colorize/colorize.plugin.zsh
```diff
@@ -61,22 +67,41 @@ colorize_via_pygmentize() {
...
+    #   while already the next file will be displayed.
+    local LESSOPEN="|zsh -c 'source \"$ZSH_COLORIZE_PLUGIN_PATH\"; colorize_cat %s 2> /dev/null'"
 
-    less -f "${tmp_files[@]}"
-)
+    # LESSCLOSE will be set to prevent any errors by executing an user script which assumes that his LESSOPEN has been executed.
+    local LESSCLOSE=""
+
+    _cless() {
+      LESS="$LESS" LESSOPEN="$LESSOPEN" LESSCLOSE="$LESSCLOSE" less "$@"
```
mcornella: This `LESSCLOSE="$LESSCLOSE"` is the same as `LESSCLOSE=""`, without having to redefine LESSCLOSE as a local variable. I would also put the variables inside the `_cless` function as well because it's where they're used.

plugins/colorize/README.md
ProbstDJakob: Sorry I missed that. I will correct that.

plugins/colorize/colorize.plugin.zsh
ProbstDJakob: Your right, the variables should be inside the function. But I created the variable LESSCLOSE on purpose, so it is consistent to the others and can be directly commented. If you wish, I'll remove it.

mcornella: I guess we can keep it for consistency, yeah.

ProbstDJakob:
@mcornella the changes have been applied

mcornella:
Thanks!

ProbstDJakob:
Thanks formatting the comments and merging into master.
One thing. The ZSH_COLORIZE_TOOL and ZSH_COLORIZE_STYLE variables in LESSOPEN should be escaped to prevent splitting. Something like this.
```
local LESSOPEN="| zsh -c 'source \"$ZSH_COLORIZE_PLUGIN_PATH\"; \
ZSH_COLORIZE_TOOL='$ZSH_COLORIZE_TOOL' ZSH_COLORIZE_STYLE='$ZSH_COLORIZE_STYLE' \
colorize_cat %s 2> /dev/null'"
```

mcornella:
They don't contain spaces so we're good for now.

ProbstDJakob:
Yeah for now and `ZSH_COLORIZE_TOOL` maybe never contain spaces but
1. Consitency
2. Safty First, not just wait for an issue on GitHub
3. No (noteworthy) downsides
4. Future proof and in case someone adds a style with spaces, it will work properly
5. No huge changes must be made

Furthermore we never check `ZSH_COLORIZE_STYLE` for spaces and this, in case someone uses spaces, will produce no errors and only results in plaine text. An `if` is not the solution either, it's only bulky and not future proof.

No offense only my personal opinion ;)

PR #8393, opened 2019-11-15
Add support for chroma for the colorize plugin

AaronNBrock:
[Chroma](https://github.com/alecthomas/chroma/tree/master) is a reimplementation of pygment in GoLang which is _much_ faster.

In this pull request I add support for optionally using `chroma` instead of `pygmentize` when either `ZSH_COLORIZE_TOOL="chroma"` or when `chroma` is installed & pygmentize is not.

This should have no effect on those currently using colorize, just an option to make it faster.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ custom/
 # temp files directories
 cache/
 log/
+*.swp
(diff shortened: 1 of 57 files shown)


robbyrussell:
Spent some time playing with this PR and Chroma does seem to be snappier than pygmentize. Was able to test switching back and forth in my ~/.zshrc file and appears to be working. Going to get this merged.

Thanks @AaronNBrock for the contribution!

PR #8396, opened 2019-11-16
Update terraform completion

umglurf:
*  Add 0.12upgrade commands
* Add state commands
* Add state completion to -target option in commands
* Add state completion where used as argument

@mcornella


mcornella:
Thanks!

PR #8397, opened 2019-11-16
Reimplement #6703

zeertzjq:
#6703 needs to be reimplemented to resolve new conflicts.


plugins/tmux/tmux.plugin.zsh
```diff
@@ -76,7 +80,11 @@ function _zsh_tmux_plugin_run() {
     elif [[ -e "$ZSH_TMUX_CONFIG" ]]; then
       tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")
     fi
-    $tmux_cmd new-session
+    if [[ "$ZSH_TMUX_SESSIONGROUP" == "true" ]]; then
```
bsiegel: Looks good, but the same conditional needs to happen when we call `tmux attach` on L74.

qjzTreeZ: It seems `attach-session` doesn't allow specifying a session group name.
```
attach-session [-dErx] [-c working-directory] [-t target-session]
```

zeertzjq:
I haven't used ohmyzsh for long. Closed due to inactivity.

PR #8399, opened 2019-11-16
Stops git-auto-fetch plugin from asking for ssh-key passphrase

### LINKED ISSUE #8381: [BUG]`git-auto-fetch` not able to handle passphrase protected ssh-keys that are not in ssh-agent
**Describe the bug**
After cd into a git repository with `git-auto-fetch` enabled it asks for the ssh passphrase but it's not possible to enter it.

**To Reproduce**
1. Enable plugin `git-auto-fetch`
2. Get ssh key with passphase that is used to pull git repos
3. Add `AddKeysToAgent yes` to `~/.ssh/config`
4. `exec zsh`
5. start `ssh-agent` (e.g. `eval $(ssh-agent)`) (Do not add ssh key!)

**Expected behavior**
Either soft fail (1) or be able to enter passphrase (2).

**Screenshots or recordings**
image

**System:**
 - OS: Manjaro
 - Zsh version: zsh 5.7.1 (x86_64-pc-linux-gnu)
 - Terminal emulator: Irrelevant but Tilix

Syphdias:
This takes care of not prompting for the ssh-key passphrase that can then not be entered.
This fixes #8381

PS: I also prefixed the git commands with `command` to not run into any issues with aliases or functions.

PPS: @slavaGanzin, it looks like you wrote most of this code, maybe you can take a look at it :smile:

### DIFF
--- a/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh
+++ b/plugins/git-auto-fetch/git-auto-fetch.plugin.zsh
@@ -1,16 +1,17 @@
 GIT_AUTO_FETCH_INTERVAL=${GIT_AUTO_FETCH_INTERVAL:=60}
 
 function git-fetch-all {
-  (`git rev-parse --is-inside-work-tree 2>/dev/null` &&
-  dir=`git rev-parse --git-dir` &&
+  (`command git rev-parse --is-inside-work-tree 2>/dev/null` &&
+  dir=`command git rev-parse --git-dir` &&
   [[ ! -f $dir/NO_AUTO_FETCH ]] &&
   (( `date +%s` - `date -r $dir/FETCH_LOG +%s 2>/dev/null || echo 0` > $GIT_AUTO_FETCH_INTERVAL )) &&
-  git fetch --all 2>/dev/null &>! $dir/FETCH_LOG &)
+  GIT_SSH_COMMAND="command ssh -o BatchMode=yes" \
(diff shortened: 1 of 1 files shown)


slavaGanzin:
@Syphdias looks good for me.

towbes:
So did this commit enable the plugin to ask for the SSH passphrase or disable that feature?  I am not getting a prompt to input my SSH key passphrase, which is causing the update to fail.  The original issue post seems to indicate the fix should have allowed the plugin to prompt the user to input the SSH Key passphrase, but the title is the opposite.  Which was this meant to accomplish?

Syphdias:
The commit turns on `BatchMode` for the ssh connection which is kind of telling it to know that there will not be any interactive input. This will lead to the git fetch failing without trying to ask for the passphrase. Which is the expected behavior.

> This takes care of not prompting for the ssh-key passphrase that can then not be entered.

I can see how this can be misunderstood. The key here being that "that can then not be entered" is being dealt with. I view soft failing as a handled error state in this case.

> The original issue post seems to indicate the fix should have allowed the plugin to prompt the user to input the SSH Key passphrase, but the title is the opposite.

I suggested two different expected behaviors and implemented (1) from #8381. It was easy to implement and is unintrusive to the user experience. I would probably also use (2) if it was implemented.

towbes:
Gotcha thanks for the reply.  I misunderstood what the expected behavior was.  I ended up using #ssh-add -K to put my key in the keychain and auto-fetch is able to function the way I needed.

PR #8401, opened 2019-11-17
added git prune local branches

sapirgolan:


sapirgolan:
@ncanceill 
@mcornella

PR #8402, opened 2019-11-17
fzf: change debian completion file path

mengbo:
Debian package install completions file in /usr/share/doc/fzf/examples/completion.zsh.


mcornella:
I added a fallback to the old path in case it wasn't found, to cover the case where the release is stretch (oldstable) or buster (stable).

PR #8403, opened 2019-11-17
Improved Git and JENV plugins

brunoborges:


mcornella:
This makes a ton of changes, can you put them in separate PRs and explain what they do? Thanks

PR #8405, opened 2019-11-18
Make 'agnoster'-theme colours themeable, add more status to prompt

sndrsmnk:
I made all colours in the 'agnoster' theme ... themeable!

The defaults are preserved, but can be overridden from your `.zshrc`

Also added extra info to the 'status' part:
* When SSH Agent Forwarding is active
* When screen is active (displays window number)


robbyrussell:

Making agnoster's colours configurable landed in #12505, under the `AGNOSTER_*` names, so the bulk of this is in the theme now.

PR #8406, opened 2019-11-18
Use safer append to hook function arrays

flatiron32:
When zsh config and thus omz is reloaded `preexec_functions` can start to have duplicates. I noticed this first with alias-finder. There are three other places in the codebase where this is done. If this solution is acceptable I would like to encapsulate it in a function that could be used in any plugin and then update the existing places to use this function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ custom/
 # temp files directories
 cache/
 log/
+*.swp
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -75,8 +75,9 @@ function omz_termsupport_preexec {
   title '$CMD' '%100>...>$LINE%<<'
 }
 
-precmd_functions+=(omz_termsupport_precmd)
-preexec_functions+=(omz_termsupport_preexec)
+autoload -U add-zsh-hook
+add-zsh-hook precmd omz_termsupport_precmd
+add-zsh-hook preexec omz_termsupport_preexec
 
 
 # Keep Apple Terminal.app's current working directory updated
@@ -99,7 +100,7 @@ if [[ "$TERM_PROGRAM" == "Apple_Terminal" ]] && [[ -z "$INSIDE_EMACS" ]]; then
   }
 
   # Use a precmd hook instead of a chpwd hook to avoid contaminating output
-  precmd_functions+=(update_terminalapp_cwd)
+  add-zsh-hook precmd update_terminalapp_cwd
   # Run once to get initial cwd set
   update_terminalapp_cwd
 fi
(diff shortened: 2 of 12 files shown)


mcornella:
[`add-zsh-hook` was added 12 years ago](https://github.com/zsh-users/zsh/commit/0d101890d0faeaefa863099042e3510f5021fbc3), I think we're OK to use it now. It needs to be autoload-ed first. 

That said, resourcing the zshrc file is not the appropriate way to restart zsh; use `exec zsh`.

flatiron32:
@mcornella You comment is a big off-putting to a newcomer to the project (and this shell). 

exit
It is not clear to me if you are suggesting I modify my PR to use add-zsh-hook or if you are suggesting the core maintainers should do something. The use of "we" confuses me and the commend suggest I might have more than a week of experience with the project. Happy to update, just don't want to do work that will never be merged.

Thanks for correcting me on appropriate way to restart zsh. Given that `source ~/.zshrc` appears more often in the omz codebase than `exec zsh` and that `source ~/.zshrc` is references twice as often than `exec zsh` according to Google results it is important to support this path even if it is inappropriate. Perhaps we could identify when this happens and nudge users to use `exec` (after correcting our docs).

mcornella:
Those are good points and sorry for the terse comment, I was writing from a phone. Also, you didn't come off as a newcomer to the shell with the implemented solution, so kudos for the research you must have had to do!

> Given that `source ~/.zshrc` appears more often in the omz codebase than `exec zsh` and that `source ~/.zshrc` is references twice as often than `exec zsh` according to Google results it is important to support this path even if it is inappropriate. **Perhaps we could identify when this happens and nudge users to use `exec`** (after correcting our docs).

This is a great point I never thought of, we could do more in the init script to catch these things (by 'we' I mean the royal 'we', I'm a part of the project). I also wasn't aware of having it in the documentation, so thanks for the heads up.

As far as mentioning `add-zsh-hook`, we started not preferring it years ago because of some old zsh versions that didn't have it yet, can't pinpoint the exact issue. Since there are several places in the codebase where we use the array append syntax, having to add the check to see if it's already in the `_functions` (and has also weird syntax) isn't optimal. The `add-zsh-hook` syntax is better and saves us from the weird syntax.

_TL;DR_, this is the code that we'd have to put in place:
```zsh
autoload add-zsh-hook
add-zsh-hook preexec preexec_alias-finder
```

The autoload could be omitted since it could be put in `lib/termsupport.zsh`, which also makes use of the `*_functions` hooks. Since these are big changes and you don't yet know the codebase, I'll put together a commit changing the whole thing and modify this PR.

PR #8408, opened 2019-11-18
Remove current directory from sys.path in python invocations

### LINKED ISSUE #8400: Python invocations are not secure
`python -c` invocations in plugins are not secure, because Python will include the current working directory in `sys.path`. As such, Python files in the current directory could be imported and executed, instead of the expected module. This does not apply to some built-in modules, such as `os` or `sys` (as these are imported before `sys.path` is configured), but is still applicable to a number of Oh My Zsh plugins.

In the following case, any Python code in a file named `json.py` in the current directory would be executed.

https://github.com/robbyrussell/oh-my-zsh/blob/451fb257032a37a7d9cae4631c03f65fd3aa7f1c/plugins/jsontools/jsontools.plugin.zsh#L13-L31

In this example, any Python code in a file named `urllib.py` (or its dependencies, e.g. `re.py` or `enum.py`) would be executed.

https://github.com/robbyrussell/oh-my-zsh/blob/451fb257032a37a7d9cae4631c03f65fd3aa7f1c/plugins/urltools/urltools.plugin.zsh#L14-L19

Ditto, but for `salt.py`.

https://github.com/robbyrussell/oh-my-zsh/blob/451fb257032a37a7d9cae4631c03f65fd3aa7f1c/plugins/salt/_salt#L274

mcornella:
Fixes #8400.

Uses [this solution](https://bugs.python.org/issue33053#msg313966) for `python -c` invocations, and [the `-I` option](https://docs.python.org/3/using/cmdline.html#id2) (added in Python 3.4) for module loading.

### DIFF
--- a/plugins/django/django.plugin.zsh
+++ b/plugins/django/django.plugin.zsh
@@ -374,7 +374,8 @@ _managepy-commands() {
 _applist() {
   local line
   local -a apps
-  _call_program help-command "python -c \"import os.path as op, re, django.conf, sys;\\
+  _call_program help-command "python -c \"import sys; sys.path.remove("");\\
+                                          import os.path as op, re, django.conf;\\
                                           bn=op.basename(op.abspath(op.curdir));[sys\\
                                           .stdout.write(str(re.sub(r'^%s\.(.*?)$' %
                                           bn, r'\1', i)) + '\n') for i in django.conf.settings.\\
--- a/plugins/django/django.plugin.zsh
+++ b/plugins/django/django.plugin.zsh
@@ -374,7 +374,7 @@ _managepy-commands() {
 _applist() {
   local line
   local -a apps
-  _call_program help-command "python -c \"import sys; sys.path.remove("");\\
+  _call_program help-command "python -c \"import sys; del sys.path[0];\\
                                           import os.path as op, re, django.conf;\\
                                           bn=op.basename(op.abspath(op.curdir));[sys\\
                                           .stdout.write(str(re.sub(r'^%s\.(.*?)$' %
--- a/plugins/jsontools/jsontools.plugin.zsh
+++ b/plugins/jsontools/jsontools.plugin.zsh
@@ -11,9 +11,10 @@ if [[ $(whence node) != "" && ( "x$JSONTOOLS_METHOD" = "x"  || "x$JSONTOOLS_METH
 	alias urlencode_json='xargs -0 node -e "console.log(encodeURIComponent(process.argv[1]))"'
 	alias urldecode_json='xargs -0 node -e "console.log(decodeURIComponent(process.argv[1]))"'
 elif [[ $(whence python) != "" && ( "x$JSONTOOLS_METHOD" = "x" || "x$JSONTOOLS_METHOD" = "xpython" ) ]]; then
-	alias pp_json='python -mjson.tool'
+	alias pp_json='python3 -Im json.tool'
 	alias is_json='python -c "
-import json, sys;
+import sys; sys.path.remove("");
+import json;
 try: 
 	json.loads(sys.stdin.read())
 except ValueError, e: 
@@ -22,11 +23,13 @@ else:
 	print True
 sys.exit(0)"'
 	alias urlencode_json='python -c "
-import urllib, json, sys;
+import sys; sys.path.remove("");
+import urllib, json;
 print urllib.quote_plus(sys.stdin.read())
 sys.exit(0)"'
 	alias urldecode_json='python -c "
-import urllib, json, sys;
+import sys; sys.path.remove("");
+import urllib, json;
 print urllib.unquote_plus(sys.stdin.read())
 sys.exit(0)"'
 elif [[ $(whence ruby) != "" && ( "x$JSONTOOLS_METHOD" = "x" || "x$JSONTOOLS_METHOD" = "xruby" ) ]]; then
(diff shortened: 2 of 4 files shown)


mcornella:
This solution can be phased out in the near future in favor of `-I`, once Python 2 is dead and we can be sure people are using a version older or equal to 3.4 (which [was released 5 years ago](https://docs.python.org/3/whatsnew/3.4.html)).

plugins/jsontools/jsontools.plugin.zsh
```diff
@@ -11,9 +11,10 @@ if [[ $(whence node) != "" && ( "x$JSONTOOLS_METHOD" = "x"  || "x$JSONTOOLS_METH
 	alias urlencode_json='xargs -0 node -e "console.log(encodeURIComponent(process.argv[1]))"'
 	alias urldecode_json='xargs -0 node -e "console.log(decodeURIComponent(process.argv[1]))"'
 elif [[ $(whence python) != "" && ( "x$JSONTOOLS_METHOD" = "x" || "x$JSONTOOLS_METHOD" = "xpython" ) ]]; then
-	alias pp_json='python -mjson.tool'
+	alias pp_json='python3 -Im json.tool'
```
saleemrashid: If you don't want to use `-I`, the following would work:

```shell
alias pp_json="python -c 'import sys; sys.path.remove(\"\"); import runpy; runpy._run_module_as_main(\"json.tool\")'"
```

`runpy._run_module_as_main` doesn't appear to be documented but, if it were to prove to be a problem in the future, you could easily switch to `-I`.

plugins/urltools/urltools.plugin.zsh
```diff
@@ -12,11 +12,11 @@ if [[ $(whence node) != "" && ( "x$URLTOOLS_METHOD" = "x"  || "x$URLTOOLS_METHOD
     alias urlencode='node -e "console.log(encodeURIComponent(process.argv[1]))"'
     alias urldecode='node -e "console.log(decodeURIComponent(process.argv[1]))"'
 elif [[ $(whence python3) != "" && ( "x$URLTOOLS_METHOD" = "x" || "x$URLTOOLS_METHOD" = "xpython" ) ]]; then
-    alias urlencode='python3 -c "import sys, urllib.parse as up; print(up.quote_plus(sys.argv[1]))"'
-    alias urldecode='python3 -c "import sys, urllib.parse as up; print(up.unquote_plus(sys.argv[1]))"'
+    alias urlencode='python3 -c "import sys; sys.path.remove(""); import urllib.parse as up; print(up.quote_plus(sys.argv[1]))"'
```
saleemrashid: As with the Django plugin, I am not sure if these aliases are escaped correctly.

plugins/jsontools/jsontools.plugin.zsh
saleemrashid: If you change the alias to `python3`, it will no longer match up with the `whence` invocation in the line above.

plugins/django/django.plugin.zsh
```diff
@@ -374,7 +374,8 @@ _managepy-commands() {
 _applist() {
   local line
   local -a apps
-  _call_program help-command "python -c \"import os.path as op, re, django.conf, sys;\\
+  _call_program help-command "python -c \"import sys; sys.path.remove("");\\
```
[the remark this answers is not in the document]
mcornella: You're right, I made a simple copy-paste without looking at that. Thanks for the extra pair of eyes!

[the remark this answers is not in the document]
mcornella: I've been playing around with it and we can just do
```py
import sys; del sys.path[0];
```
no quotes to deal with.

plugins/jsontools/jsontools.plugin.zsh
mcornella: Used your numpy suggestion. We can refactor the whole thing once Python 2 is dead.

plugins/django/django.plugin.zsh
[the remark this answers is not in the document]
saleemrashid: I don't think that is a secure approach:

- `sys.path.remove("")` will remove **all** occurrences of the empty string, whereas `del sys.path[0]` would only remove one.

- I don't think it is safe to assume `sys.path[0] == ""`. For example, the user may have configured their Python not to include the empty string, or a module or site customization may have prepended a path to `sys.path`.

saleemrashid:
Another, potentially more robust solution (that doesn't require quotes!):

```python
import sys; sys.path = list(filter(None, sys.path))
```

This will remove all falsy values (including the empty string), so it is more robust than `sys.path.remove("")` (which will only remove the first occurence).

mcornella:
[The Python docs say](https://docs.python.org/3/using/cmdline.html#interface-options) the current directory / empty string will be added at the start of `sys.path`:

> If this option is given, the first element of sys.argv will be "-c" and **the current directory will be added to the start of sys.path** (allowing modules in that directory to be imported as top level modules).

> If this option is given, the first element of sys.argv will be the full path to the module file (while the module file is being located, the first element will be set to "-m"). **As with the -c option, the current directory will be added to the start of sys.path.**

> If this option is given, the first element of sys.argv will be "-" and **the current directory will be added to the start of sys.path.**

> If the script name refers directly to a Python file, the directory containing that file is added to the start of sys.path,

saleemrashid:
I understand that, however users may have different environment setups that cause paths to prepended to `sys.path` after the empty path is prepended. I think we should err on the side of caution when it comes to potential security issues.

mcornella:
From what I've read in the [docs re: environment variables](https://docs.python.org/3/using/cmdline.html#environment-variables), the "current directory"/"script directory" will always be put first in `sys.path`. See for example [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH):
> An additional directory will be inserted in the search path in front of PYTHONPATH as described above under Interface options.

I'm comfortable enough with my assessment here. I do not like the solution, but as I said before, this is only temporary until we can get rid of Python 2 and use `-I` which will be secure by default.

saleemrashid:
Python has had numerous regressions in the past, so I am incredibly wary about relying on close reading of the documentation. I also wouldn't rely on specifics about the user's environment (especially when working on a framework intended for users who heavily customize their environment!).

But I don't use Oh My Zsh, so I'm not going to argue this any further ¯\_(ツ)_/¯

PR #8417, opened 2019-11-21
feat(agnoster): add git tag to prompt

### LINKED ISSUE #8587: Show current git tag in prompts (if there is one on the currently checked out commit)
Not really, but it seems that acquiring VCS information is implemented in themes, not centrally :/ I'm using agnoster.

I'd love if there was a config flag to display a tag instead of a commit hash, if the current commit is exactly on a tag.

**What I tried so far**
I searched a bit for how I would get the current tag conveniently, but there does not seem to be a "stable" git command that only outputs the current tag's name. So there would be some string parsing involved.

**Additional context**
My use case for this is for quickly seeing which version of projects I'm only building/using (not developing) I checked out. I like using tags if they're available to stay on marked releases instead of whatever HEAD is.

jabunovoty:
Fixes #8587

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -109,7 +109,9 @@ prompt_git() {
   if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
     repo_path=$(git rev-parse --git-dir 2>/dev/null)
     dirty=$(parse_git_dirty)
-    ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git rev-parse --short HEAD 2> /dev/null)"
+    ref=$(git symbolic-ref HEAD 2> /dev/null) || \
+    ref="◈ $(git describe --exact-match --tags HEAD 2> /dev/null)" || \
+    ref="➦ $(git rev-parse --short HEAD 2> /dev/null)" 
     if [[ -n $dirty ]]; then
       prompt_segment yellow black
     else


themes/agnoster.zsh-theme
```diff
@@ -109,7 +109,9 @@ prompt_git() {
   if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
     repo_path=$(git rev-parse --git-dir 2>/dev/null)
     dirty=$(parse_git_dirty)
-    ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git rev-parse --short HEAD 2> /dev/null)"
+    ref=$(git symbolic-ref HEAD 2> /dev/null) || \
+    ref=" $(git describe --exact-match HEAD 2> /dev/null)" ||\
```
jabunovoty: The unknown character is `nf-fa-tag` (`f02b`). Working in the command line when the powerline fonts are installed.

dan1994: I had the same solution in mind, but for me I had to add `--tags` to the `git describe` command. Otherwise it searches only for annotated tags, and fails on normal tags.

dan1994: Also, theoretically the `formats` and `actionformats` can be used to get `ref` more cleanly. Also, if you checked out a commit that is pointed to by both a branch and a tag it will choose what to display based on how you checked it out.

The downside is that it needs to be formatted conditionally anyways (depending if it's a commit hash, branch or tag) so I'm not sure it's worth it.

mcornella: That is not actually a Powerline font glyph, but a Font Awesome one. Powerline font glyphs are only the ones starting with `nf-pl-` in https://www.nerdfonts.com/cheat-sheet. We should use a different glyph, preferably one in the UTF-8 spec because Powerline glyphs are very few.
Also, to keep this consistent, there should be a space between `||` and `\`:
```suggestion
    ref="◈ $(git describe --exact-match HEAD 2> /dev/null)" || \
```

mcornella: Also, the proper command is `git describe --exact-match --tags HEAD`, otherwise non-annotated tags aren't matched:
```suggestion
    ref="◈ $(git describe --exact-match --tags HEAD 2> /dev/null)" ||\
```

53jk1 (approved):

nwu63:
Just a casual bump on this PR---this is a feature that I would love to have in the theme.

52326125:
Where is it blocked now?

carlosala:
Thanks for all your contributions!

PR #8419, opened 2019-11-21
Added role delegation support and MFA support as per IAM Best Practices

maksyms:
Inspired by antonbabenko/awsp@6588191, I added Role Delegation and MFA support as per [AWS IAM Best Practices](https://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html#delegate-using-roles) to AWS plugin. No changes to the command syntax, everything happens behind the scenes. This is especially useful when MFA is enforced for AWS accounts.

The newly supported use case is when a user has one main IAM account protected by MFA that can assume multiple roles elsewhere, and those roles are configured in .aws/config. The script will now assume a role and set additional environment variables automatically, including support for [external id] and MFA authentication with session length. (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) and [MFA](https://aws.amazon.com/premiumsupport/knowledge-center/authenticate-mfa-cli/).


haidaraM:
I'm having an issue testing your changes. Looks like the function `aws_profiles` is broken somehow. The profiles are not correctly parsed.
Example with your changes:
```shell
$ aws_profiles
[profile vwt-nprod]
[profile vision-dev]
$ asp vwt-nprod
Profile 'vwt-nprod' not found in '/Users/mohamed.haidara/.aws/config'
Available profiles: [profile, vwt-nprod], [profile, vision-dev]
```
Before:
```shell
$ aws_profiles
vwt-nprod
vision-dev
$ asp vwt-nprod # no error
```

And I'm also wondering if we should add such logic to the AWS plugin. We should keep it really simple as it is now. If the main purpose is for Terraform, we may let Terraform assume the roles in the accounts by specifying them in the provider. 

For MFA and related, there is [aws-mfa](https://github.com/broamski/aws-mfa) that can automatically created a profile with the required keys. You can reference this profile in your `~/.aws/config` to assume a role.

maksyms:
Thanks for your comment. I tested it just now on zsh 5.4.2 on Linux Mint 19.2, and it works just fine. It seems that the syntax of sed regex somehow works differently. I will find a Mac and test.

As for the logic, this is not only for Terraform. Here is what AWS says: https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_environment.html
This is the "right" way, as far as AWS is concerned, to securely access AWS resources in other accounts.

maksyms:
@haidaraM I think I know why you are seeing the issue. Could you please send me the result of:
`which grep`

I've noticed the line doesn't work if grep has colour enabled.

maksyms:
@haidaraM I have now got a Mac with its ancient sed, which supports REG_EXTENDED, but not REG_ENHANCED mode, despite both being described in documentation `man re_format`. Now I tested it both on Linux and MacOS - all works, ready to go.

maksyms:
@kujon This can now be used on a Mac.

maksyms:
We are finding this change quite valuable when enforcing MFA on AWS.

maksyms:
@mcornella It seems the bot misclassified this PR - it has nothing to do with documentation, but everything to do with using AWS CLI in a secure way. What else can I do to promote merging this PR, if anything?

maksyms:
I've resolved the conflict with `master`, which now allows `@` in the profile name, as per #9099 fixed by @lutzhorn .

53jk1 (approved):

PR #8426, opened 2019-11-24
Update sbt plugin to adopt with new versions

drmatrix0:
for example: https://www.scala-sbt.org/1.x/docs/Console-Project.html

`publish-local` is updated to `publishLocal`


mcornella:
Can you also change the README and the options offered in the `_sbt` completion file? This last one is just a find and replace.

PR #8430, opened 2019-11-25
Fix change to old shell on uninstall

### LINKED ISSUE #8252: "uninstall_oh_my_zsh" command not working properly
`uninstall_oh_my_zsh` command returns to **bash** instead of **zsh**.

reproduce steps;
- install oh-my-zsh
- run `uninstall_oh_my_zsh`

System;
Macos Catalina 10.15 with default zsh shell

I think it's because of this part. There is no .`shell.pre-oh-my-zsh` file in the home directory;

https://github.com/robbyrussell/oh-my-zsh/blob/687c50bdf999f8efd45f3c8f578a62329b0633da/tools/uninstall.sh#L28-L32

giuscri:
Close #8252

### DIFF
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -25,12 +25,8 @@ if [ -e "$ZSHRC_ORIG" ]; then
   echo "Your original zsh config was restored."
 fi
 
-if hash chsh >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

PR #8432, opened 2019-11-27
Add README.md to root of Plugins

bcdady:
To make the power of Oh My ZSH Plugins more approachable for new users, a Plugins root level README could serve as an handy index and introduction for the numerous plugins.

This early draft at least links to each (expected) README for each Plugin (as of [commit 671d21a](https://github.com/ohmyzsh/ohmyzsh/tree/671d21ac5305e815668500563a6b7ef61057b659)
I hope other Plugin fans (or even maintainers) might fork a copy of this README and update it with any Plugin 'Intro' that they'd like to add or improve.


mcornella:
I don't think it necessary to list the plugins here while there is a folder list displayed by GitHub. Furthermore, the documentation of what each plugin does is and should remain in the wiki, where it is much easier to keep up to date. Thanks for your contribution!

Issue #8380, opened 2019-11-12
Keychain plugin fails with .local domain name.


matthewhembree:
Keychain seems to write the environment files with the fqdn (which includes .local, in my case).

The Keychain plugin sources $HOME/.keychain/$SHORT_HOST-sh, which is not found. This leaves the SSH_AUTH_SOCK as the default macos one set by launchd. Identities never get added to the Keychain instantiated agent.

ssh-add -L (keychain -L)
"The agent has no identities."

I have added a line to my .zshrc as a workaround:
source "$HOME/.keychain/$HOST-sh"

macos 10.15.1

larson-carter:
Hello @matthewhembree Can you give me some more details on your enviroment? Such as DNS IP-Ranges DHCP?

mcornella:
You're right, as pointed out in [this commit comment](https://github.com/ohmyzsh/ohmyzsh/commit/95a3b2768033627d9ba3e347f368dffae92906b7#commitcomment-32794010) they use `$HOSTNAME` instead:

https://github.com/funtoo/keychain/blob/7e37e4b63529f4a1caae9ca5e6db42b10a6fd2f3/keychain#L1373-L1375

This is a simple change, can you submit a PR?

matthewhembree:
I'm not sure if it's relevant to go down the troubleshooting path. I'll compare may work with my home environment later.

```$HOSTNAME``` is empty on my system. ```uname -n``` returns the ```.local``` suffix.

It seems that the plugin could just call keychain as below and avoid all ambiguity when sourcing the pidfile:
```sh
keychain ${^options:-} --agents ${agents:-gpg} ${^identities} --host $SHORT_HOST
```
I'm not sure what the [NFS compatibility](https://github.com/funtoo/keychain/blob/7e37e4b63529f4a1caae9ca5e6db42b10a6fd2f3/keychain#L1368) issue would be.

mcornella:
Right, [`$hostopt` is set when passing `--host`](https://github.com/funtoo/keychain/blob/7e37e4b63529f4a1caae9ca5e6db42b10a6fd2f3/keychain#L1271-L1274). That looks like a cleaner solution.

I'm not sure if we need to think through the behavior when upgrading the plugin. What would happen if keychain is running and the plugin is upgraded? If not on a `.local` domain, `$SHORT_HOST` and the keychain method are equivalent, so there wouldn't be a problem? I don't have a Mac to test this.

[...]

PR #8436, opened 2019-11-28
Added hostopt to keychain execution.

matthewhembree:
Fixes issue #8380 .

What ```keychain``` creates will now match up with what the plugin sources.


mcornella:
Thanks!

Issue #8444, opened 2019-11-30
$ZSH/lib/grep.zsh is too slow for what it does - proposal to use $ZSH_CACHE_DIR


mattmc3:
**Is your feature request related to a particular plugin or theme? If so, specify it.**
$ZSH/lib/grep.zsh

**Is your feature request related to a problem? Please describe.**
$ZSH/lib/grep.zsh is way too slow for its simple function - namely to generate a single grep alias.

Running this script highlights how inefficient $ZSH/lib/grep.zsh is compared to the rest of the scripts in lib:

```zsh
#!/usr/bin/env zsh
zmodload zsh/zprof
for config_file ($ZSH/lib/*.zsh); do
  source $config_file
done
zprof
```

This is the profiler output:
```
num  calls                time                       self            name
-----------------------------------------------------------------------------------
 1)    2           9.95     4.97   62.36%      9.95     4.97   62.36%  grep-flag-available
 2)    2           2.81     1.41   17.65%      2.81     1.41   17.65%  env_default
 3)    1           1.72     1.72   10.80%      1.72     1.72   10.80%  colors
 4)    2           0.94     0.47    5.88%      0.94     0.47    5.88%  add-zsh-hook
 5)    1           0.29     0.29    1.81%      0.29     0.29    1.81%  is-at-least
 6)    1           0.24     0.24    1.52%      0.15     0.15    0.97%  omz_termsupport_preexec
 7)    1           0.09     0.09    0.55%      0.09     0.09    0.55%  title
```

**Describe the solution you'd like**
I'd propose the following simple change to $ZSH/lib/grep.zsh, which would cache the results of calculating the grep alias for one day and source those.

```zsh
# get any cached grep alias file if it's less than 24 hours old
_grep_alias_cache=($ZSH_CACHE_DIR/grep_alias.zsh(Nm-24))
if (( $#_grep_alias_cache )); then
    source $ZSH_CACHE_DIR/grep_alias.zsh
else
    # ... do all the same things $ZSH/lib/grep.zsh already does in the else part
    # then cache the result
    echo "alias grep=\"grep $GREP_OPTIONS\"" > $ZSH_CACHE_DIR/grep_alias.zsh
fi
unset _grep_alias_cache
```

After implementing this, the performance profile looks much better (after the first run of the day obviously):

```zsh
num  calls                time                       self            name
-----------------------------------------------------------------------------------
[...]

PR #8445, opened 2019-11-30
Performance enhancement for lib/grep

mattmc3:
- Use $ZSH_CACHE_DIR to store the grep alias with all the right features
- Expire the cache after 24 hours
- See issue #8444


mattmc3:
Don't forget to ignore whitespace differences in the diff with "?w=1". Due to putting the existing lib/grep.zsh contents in an else block, the indentation changed. Here's the easier to read diff link:

https://github.com/ohmyzsh/ohmyzsh/pull/8445/files?w=1

PR #8446, opened 2019-12-01
Added Completion for Todoman

a0n:
See: https://github.com/pimutils/todoman


robbyrussell:
Hi there, thanks for this. Could you provide a README file in this directory, too?

PR #8447, opened 2019-12-02
Created Watson Completion Plugin

a0n:
See: https://tailordev.github.io/Watson/


robbyrussell:
Hi there, thanks for this. Could you provide a README file in this directory, too?

PR #8448, opened 2019-12-02
Add kubectl log since aliases

alaa:
Add kubectl log since aliases
```
--since 1h
--since 1m
--since 1s
```


robbyrussell:

PR #8453, opened 2019-12-04
feat(ys): add virtualenv prompt info

MrAureliusR:
Now this theme will show if you are in a virtualenv, and if so, what the name of the virtualenv is.

### DIFF
--- a/themes/ys.zsh-theme
+++ b/themes/ys.zsh-theme
@@ -35,18 +35,16 @@ ys_hg_prompt_info() {
 	fi
 }
 
-local exit_code="%(?,,C:%{$fg[red]%}%?%{$reset_color%})"
-
 # Virtualenv
 local venv_prompt='$(virtenv_prompt)'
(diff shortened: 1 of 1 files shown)


MrAureliusR:
I have been using this theme with this addition for a few months now, and it works perfectly with all the plugins I use.

53jk1 (approved):

MrAureliusR:
Is going to be merged eventually? It's been almost a year, and it's a fairly trivial change.

MrAureliusR:
Should I close this PR? I don't think I've ever had one sit _this_ long before.
I thought it was useful, but since it hasn't been merged I'm assuming there's a reason why? Did I not match formatting? Is there a better way I could have achieved the same result? I am happy to refactor or re-write the commit if that's the issue.

TheTechRobo:
@mcornella is very busy with the hundreds of PRs. Don't fret - it will eventually be reviewed.

mcornella:
Sorry it took so long.

PR #8460, opened 2019-12-07
use DCO_FILE env inside docker-compose aliases

ShahinSorkh:


ShahinSorkh:
Any thoughts on this?

mcornella:
This was a good addition but I was sure it was implemented already into docker-compose, and it is: [`COMPOSE_FILE` env variable](https://docs.docker.com/compose/reference/envvars/#compose_file). Thanks for your contribution though!

PR #8462, opened 2019-12-09
Run pyenv plugin only when necessary

2tunnels:
Wrapped pyenv plugin in condition, so it will only run if `pyenv` command is not already loaded.
Some software (Poetry for example), may spawn shell multiple times.
In this scenario `.zshrc` will be read again, running all plugins and changing $PATH again.
Seems like other plugins are aware of that, nvm for example.
May be it's not the best solution and plugin should create some environment variable that it has been already used.


mcornella:
This is needed if in between the update of `$PATH` the `$commands` parameter hasn't been updated, i.e. if there was no `rehash` or `hash -rf` or automatic rehash. LGTM

PR #8474, opened 2019-12-13
Added FZF base path for freebsd

pprugger:
Added the fzf base path for freebsd to fzf.plugin.zsh.

@maxbrunet 
@mcornella


mcornella:
Have you tested that this works? I don't have a FreeBSD machine to test this on.

pprugger:
Heya!
I wrote the patch on a freebsd machine yes.
That's why i found the issue on the first place :)
Can send you screenshots if you want.

mcornella:
Thanks!

PR #8479, opened 2019-12-17
extract: Add tar.lz support

mustaqimM:


plugins/extract/extract.plugin.zsh
```diff
@@ -45,6 +45,7 @@ extract() {
 				&& tar --zstd -xvf "$1" \
 				|| zstdcat "$1" | tar xvf - ;;
 			(*.tar) tar xvf "$1" ;;
+			(*.tar.lz) (( $+commands[lzip] )) && tar xvf "$1" ;;
```
mcornella: Shouldn't this use `lunzip`? https://apple.stackexchange.com/a/301020

mustaqimM: It could be either. On my system it's not available. `lunzip` and `plzip` decompress to `tar`. With `lzip` installed calling `tar` is the only command required to fully extract it.

mcornella: You're right, I just tested it.

PR #8480, opened 2019-12-17
Adding in support for kubernetes context prompt

guice:
Following in the stype of AWS, adding in a cli prompt for current kubernetes context, using `kccc` alias.


larson-carter:
@guice Thank you for the PR. I am going to need about 2-4 testers to test this PR out. Since I do not have a way of testing this out in my specific environment.

Thanks  -- Larson

53jk1 (approved):

PR #8484, opened 2019-12-18
git: add `gla` alias for top git graph recipe on StackOverflow

abitrolly:
https://stackoverflow.com/questions/1057564/pretty-git-branch-graphs

Compared to current `git log --graph` aliases (`glola` is the closest)
this one easy to remember - `gla`(nce) or `git log --all`, more
readable - puts dates first and moves ref names to the end of the line
making comments aligned.

image

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -89,6 +89,7 @@ plugins=(... git)
 | gk                   | gitk --all --branches                                                                                                            |
 | gke                  | gitk --all $(git log -g --pretty=%h)                                                                                             |
 | gl                   | git pull                                                                                                                         |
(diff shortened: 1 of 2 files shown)


woestedroid:
###

carlosala:
Closing this! We already have lots of aliases for different kinds of `git log` purposes, and this one seems a bit particular; probably the best place for it is in the user's configs!
Thanks for you contribution!

abitrolly:
@carlosala would be nice to document them properly, i.e. with screens, because https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/git doesn't really give an idea if those aliases are useful.

PR #8485, opened 2019-12-18
xdg_config_home and --no-use

hesxenon:
honor xdg_config_home if present and make use of the `--no-use` flag to lazy load it for nvm versions that support it


mcornella:
Duplicate of #7138 and #7807.

PR #8486, opened 2019-12-18
added mercurial to gnzh and robbyrussell themes

TousledSmile:


anran758:
why stop the PR? Because added the variable will affect the performance of robbyrussell?

PR #8487, opened 2019-12-19
Replace %{$fg[color]%} by %F{color} in themes

ericbn:
that only change the foreground color. %F works at least since Zsh 4.3.


ericbn:
These prompt sequences work since Zsh 4.3. At last since Zsh 4.3.11, released in 2010!

Screen Shot 2019-12-18 at 19 48 47

awasum:
Looking good... Thanks @ericbn

mcornella:
Needs automated theme testing.

PR #8495, opened 2019-12-26
Created new theme with streamlined variant of xiong-chiamiov-plus.

geekmdtravis:
Created new theme with streamlined variant of xiong-chiamiov-plus where the directory 
displayed is only the current working directory rather than the full path relative to the user home directory. This reduces the footprint of the first line in the prompt, accommodating smaller view-port sizes for terminals without having that top line break, which normally results in increasing bulkiness and reducing readability.


mcornella:

PR #8498, opened 2019-12-26
Fix for tab completion eating lines when using af-magic theme

### LINKED ISSUE #8497: Tab completion is eating above lines when using af-magic theme
**Describe the bug**
Tab completion is eating above lines when using af-magic theme.

**To Reproduce**
1. Use the af-magic theme
2. Press tab for autocomplete at the terminal
3. The line immediately above was deleted

**Expected behavior**
Tab completion should not delete any lines from the display.

**Screenshots or recordings**
#### Before Tab:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic) » date                                                                                                                                                zach@zach-desktop
Thu Dec 26 09:53:46 EST 2019
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic) »

```
#### After Tab:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic) » date                                                                                                                                                zach@zach-desktop
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic) » ls themes/
```

#### Before Tab:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » source ~/.zshrc                                                                                                                                    zach@zach-desktop
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » date                                                                                                                                               zach@zach-desktop
Thu Dec 26 09:56:50 EST 2019
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » 

```
#### After Tab:
```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » source ~/.zshrc                                                                                                                                    zach@zach-desktop
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » date                                                                                                                                               zach@zach-desktop
Thu Dec 26 09:56:50 EST 2019
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/.oh-my-zsh(tab-complete-af-magic*) » ls themes/ 
```

**System:**
 - OS: Ubuntu
 - zsh 5.4.2 (x86_64-ubuntu-linux-gnu)

**Additional context**
I started seeing the error after [this change](https://github.com/ohmyzsh/ohmyzsh/commit/17f4cfca99398cb5511557b8515a17bf1bf2948a).  
It looks like the root cause was introduced by [zsh 5.3](http://www.zsh.org/mla/workers/2017/msg00021.html) when using multiline prompts.
It would be nice to fix here for compatibility. I have a solution and will open a pull request.

zachscheider:
Fixes #8497 

I'm new to this, so let me know if there's anything I should be doing differently.  
The description of the problem is in the issue I referenced above.

### DIFF
--- a/themes/af-magic.zsh-theme
+++ b/themes/af-magic.zsh-theme
@@ -6,7 +6,7 @@ if [ $UID -eq 0 ]; then NCOLOR="red"; else NCOLOR="green"; fi
 local return_code="%(?..%{$fg[red]%}%? ↵%{$reset_color%})"
 
 # primary prompt
-PROMPT='$FG[237]${(l.COLUMNS..-.)}%{$reset_color%}
+PROMPT='$FG[237]${(l.COLUMNS-1..-.)}%{$reset_color%}
 $FG[032]%~\
 $(git_prompt_info)$(hg_prompt_info) \
 $FG[105]%(!.#.»)%{$reset_color%} '


themes/af-magic.zsh-theme
```diff
@@ -6,7 +6,7 @@ if [ $UID -eq 0 ]; then NCOLOR="red"; else NCOLOR="green"; fi
 local return_code="%(?..%{$fg[red]%}%? ↵%{$reset_color%})"
 
 # primary prompt
-PROMPT='$FG[237]${(l.COLUMNS..-.)}%{$reset_color%}
+PROMPT='$FG[237]${(l.COLUMNS-1..-.)}%{$reset_color%}
```
mcornella: Can you try this instead?
```suggestion
PROMPT='%{$FG[237]%}${(l.COLUMNS..-.)}%{$reset_color%}
```

mcornella:
I can reproduce it with `COMPLETION_WAITING_DOTS` turned on, but neither your solution nor mine change the behavior. This has been reported other times, so I'll close the other issue as duplicate.

themes/af-magic.zsh-theme
zachscheider: I tried it and it unfortunately doesn't fix the issue.

zachscheider:
Hi @mcornella . Are using the af-magic theme when trying to reproduce this? They both have similar effects but different causes. The issue #8497 that this is trying to fix is occurring whether or not COMPLETION_WAITING_DOTS is turned on.

mcornella:
I can't reproduce it without `COMPLETION_WAITING_DOTS` and zsh 5.3. What other plugins are you using? This has to be a `zle redisplay` bug for sure.

zachscheider:
You're right it does only happen with COMPLETION_WAITING_DOTS set to true. I must have forgotten to source my .zshrc after changing it. 

The only plugin I have is git. I don't think it's related to that.

I think it occurs when a multiline prompt fills every column in the terminal. The reason I tried this as a fix is because of a [stackoverflow comment](https://stackoverflow.com/questions/23544078/hitting-tab-in-zsh-eats-above-lines#comment82330429_41645262) I found.

I've started from a fresh installation and am able to reproduce the issue and fix it easily. I'll attach the full logs.

To summarize the steps:

1. Docker [containers](https://github.com/zsh-users/zsh-docker) for zsh versions.
docker run -it zshusers/zsh:5.3
This allows me to start from a clean slate and not mess with my implementation. 
2. Install dependencies and oh-my-zsh
sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
3. Edit /root/.zshrc, change the theme to af-magic, and source it
Tab completion works as expected.
4. Set COMPLETION_WAITING_DOTS to true and source it
Tab completion now eats the above line
5. Edit /root/.oh-my-zsh/themes/af-magic.zsh-theme and source it
Tab completion is working as expected again

[log.txt](https://github.com/ohmyzsh/ohmyzsh/files/4006503/log.txt)

mcornella:
Can't reproduce your fix, it still happens:

image

mcornella:
Works without problems in the latest version of zsh, without the need for a fix:
image

The solution then is to use a newer zsh version.

juanibiapina:
To be more specific: this is fixed for me on zsh 5.7.1, but not when inside tmux 3.0a. I'd appreciate any ideas here.

PR #8501, opened 2019-12-26
Create "printc" plugin for abstracting ANSI color codes. 

philFernandez:
This is an interface, or wrapper if you will, around the ANSI color code system that our terminal emulators use to print colorized text. It provides a clean interface, in the form of a `printc` statement, that allows users to print colorized, italicized, bold, and/or underlined text.


kbakdev (approved):

robbyrussell:
Thanks for putting this together, and sorry for the very long delay.

zsh covers this natively now... `print -P "%F{#ff8800}text%f"` gives you truecolor directly.

PR #8503, opened 2019-12-27
Feature/add dotnet plugin

xontab:


robbyrussell:
Never thought I'd see the day! Thanks for bringing the .NET world to OMZ, @xontab!

cra0zy:
Ugh, dotnet already has a zsh plugin: https://github.com/dotnet/cli/blob/master/scripts/register-completions.zsh

cra0zy:
Here: https://github.com/ohmyzsh/ohmyzsh/pull/8518

PR #8507, opened 2019-12-30
Readd ret_status many local configs would still use

dszakal:
Many people just set up their prompts once - and removing this ret_status broke things for them with the auto update. At least it will just exist again and work the way it used to.


mcornella:
As I said in the [commit thread](https://github.com/ohmyzsh/ohmyzsh/commit/c44569f06eef2a3b4a1666686e7040c33e59e49c#commitcomment-37003030), this won't make it into the repository. You should never rely on internal implementation that is subject to change. Also, this change wouldn't do nothing, as it doesn't modify the `PROMPT` variable.

PR #8508, opened 2019-12-31
automatically install zsh if not installed already

0xdaya:
During installation, I thought it would be nice if we automatically install zsh in user's system if they haven't installed already instead of exiting the script. It will work only for Linux users as macOS have zsh preinstalled. The only thing which don't seem right in my proposed changes is I have added support only for most common Linux distros but not for all and I am not sure how do to it.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -117,7 +117,7 @@ setup_zshrc() {
 			mv "$OLD_ZSHRC" "${OLD_OLD_ZSHRC}"
 
 			echo "${YELLOW}Found old ~/.zshrc.pre-oh-my-zsh." \
-				"${GREEN}Backing up to ${OLD_OLD_ZSHRC}${RESET}"
+			"${GREEN}Backing up to ${OLD_OLD_ZSHRC}${RESET}"
 		fi
 		echo "${YELLOW}Found ~/.zshrc.${RESET} ${GREEN}Backing up to ${OLD_ZSHRC}${RESET}"
 		mv ~/.zshrc "$OLD_ZSHRC"
@@ -127,8 +127,8 @@ setup_zshrc() {
 
 	cp "$ZSH/templates/zshrc.zsh-template" ~/.zshrc
 	sed "/^export ZSH=/ c\\
-export ZSH=\"$ZSH\"
-" ~/.zshrc > ~/.zshrc-omztemp
+	export ZSH=\"$ZSH\"
+	" ~/.zshrc > ~/.zshrc-omztemp
 	mv -f ~/.zshrc-omztemp ~/.zshrc
 
 	echo
@@ -148,8 +148,8 @@ setup_shell() {
 	# If this platform doesn't provide a "chsh" command, bail out.
 	if ! command_exists chsh; then
 		cat <<-EOF
-			I can't change your shell automatically because this system does not have chsh.
-			${BLUE}Please manually change your default shell to zsh${RESET}
+		I can't change your shell automatically because this system does not have chsh.
+		${BLUE}Please manually change your default shell to zsh${RESET}
 		EOF
 		return
 	fi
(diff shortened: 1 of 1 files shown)


Kattus:
Please don't just go ahead and install software on a users system, instead make it ask them if they want the script to do it for them (or quit).

For the problem about only supporting specific distros, well first of all you should probably target package managers instead of distros (else the list will get insanely long), and simply fail and quit.

0xdaya:
> Please don't just go ahead and install software on a users system, instead make it ask them if they want the script to do it for them (or quit).

Actually I was previously thinking about this, but then I thought _If_ user is willing to install oh-my-zsh then _why_ he would refrain from installing zsh? So that's why in my opinion It will be unnecessary to ask for user permission. Of course I can add that but let's see what other contributors have to say about this :)

0xdaya:
> For the problem about only supporting specific distros, well first of all you should probably target package managers instead of distros (else the list will get insanely long), and simply fail and quit.

After some googling and research I found that unfortunately it is not possible to find the package manager of a Linux distro by simply a command. We can only find _distro name_ but that is also not the solution.

ModdedGamers:
Neofetch can figure out what package manager a system is using. I'll go do some hunting to figure out how it does that.

0xdaya:
> Neofetch can figure out what package manager a system is using. I'll go do some hunting to figure out how it does that.
> Edit: Lines 1379 - 1527 have what you're looking for.

I am not sure what you mean by this? Default output of neofetch don't tell about package manager.
Do you mean I should modify /usr/bin/neofetch(from 1379-1527) to display the name of package manager? And what if user have not installed neofetch already?

TriMoon:
@0xdaya i think he means the logic can be found between those line numbers inside the code of neofetch, so the user does not need to have neofetch installed for that purpose :wink:

0xdaya:
> @0xdaya i think he means the logic can be found between those line numbers inside the code of neofetch, so the user does not need to have neofetch installed for that purpose 😉

Ah I see, but why would I do that? I mean we need to find the pkg manager of _user's_ system to install zsh. So how could anyone can possibly do that by modifying neofetch _while_ he installs oh-my-zsh through installer script?

ModdedGamers:
@Trimoon that's exactly what I was thinking. The logic can be found there, and we could probable just take the code from there as it's freely available on GitHub.

TriMoon:
> @0xdaya ,
> So how could anyone can possibly do that by modifying neofetch _while_ user is installing oh-my-zsh through installer script?

:rofl: that's exactly the point @ModdedGamers was making, you **don't need** to interact with neofetch while installing oh-my-zsh through installer script.
You just need to implement the **logic** you find inside neofetch's code in the installer script of oh-my-zsh :wink:

ModdedGamers:
@TriMoon Thanks for the explanation! That's exactly what I was thinking.

0xdaya:
> > @0xdaya ,
> > So how could anyone can possibly do that by modifying neofetch _while_ user is installing oh-my-zsh through installer script?
> 
> 🤣 that's exactly the point @ModdedGamers was making, you **don't need** to interact with neofetch while installing oh-my-zsh through installer script.
> You just need to implement the **logic** you find inside neofetch's code into the installer script of oh-my-zsh 😉
> Psssttt.... [neofetch](https://github.com/dylanaraps/neofetch) is a shell script....
> 
> > @ModdedGamers:
> > > @TriMoon that's exactly what I was thinking. The logic can be found there, and we could probable just take the code from there as it's freely available on GitHub.
> 
> For quick access: [the mentioned lines](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L1379-L1527).
> 
> Screenshot20200116_171114 neofetch

I see what you mean, I misunderstood it before  😅 . But I got confused while reading functions `has(), dir(), pac(), tot()` because I only know basics of BASH and I don't understand what is meant by  those symbols. So if you guys can figure it out then I will add a new commit :)

ModdedGamers:
has(), dir (), pac(), and tot() are just functions that are really complicated. I wish I could explain them, but I can't.

53jk1 (approved):

mcornella:
This is not a good idea. Thanks for trying though!

PR #8489, opened 2019-12-23
add plugin of zsh-syntax-highlighting zsh-autosuggestions

qingwei9001:

PR #8514, opened 2020-01-01
Add the zsh-autosuggestions, zsh-completions and zsh-syntax-highlighting plugins

ProbstDJakob:
After the Pull-Request #8489 has been closed without further commenting why. So I opened a new one with the additional zsh-completions plugin.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,9 @@
+[submodule "zsh-syntax-highlighting"]
+	path = plugins/zsh-syntax-highlighting
+	url = https://github.com/zsh-users/zsh-syntax-highlighting.git
+[submodule "zsh-autosuggestions"]
+	path = plugins/zsh-autosuggestions
+	url = https://github.com/zsh-users/zsh-autosuggestions.git
+[submodule "zsh-completions"]
+	path = plugins/zsh-completions
+	url = https://github.com/zsh-users/zsh-completions.git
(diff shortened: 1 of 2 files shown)


ModdedGamers:
As you might notice, all of the other plugins are bundled with ohmyzsh, so it's probably better if you just put the source files into the source tree.

ProbstDJakob:
They are referencing the master branch and this is in my opinion the better way to choose. I am not the owner of the code and if it would be bundled, it has to be updated every time the original code updates. This also means, that the version history and the author is not correct.

One thing that has to be done, is to write a descritpion in the wiki about the plugins.

ModdedGamers:
Well, we probably shouldn't write anything on the wiki until everything's merged.

ProbstDJakob:
Of course, that’s what I meant.

xiruizhao:
Any update? Since I want to git manage my dotfiles, ohmyzsh has to be added as a submodule. But I don't want to fork ohmyzsh so that these plugins can be added as submodules to ohmyzsh.

ProbstDJakob:
@mcornella will this ever get merged or should this request be closed?

mcornella:
Using git submodules is a complex choice. At the moment there isn't a plan to implement it.

marcthe12:
zsh-completions is very important. It is often in distro's repos and has high quality completions. It is abetter place to move some of the completion plugins.

xiruizhao:
I agree with @marcthe12. git submodule brings more inconvenience than its worth. People should preferably install from their distribution's repository. Although zsh-completions is not provided in official Ubuntu repository (the other two are), people can still add its OBS repo. Another good thing is that these packages contain only the plugin files without git files or another stuff. I think you should close this PR @ProbstDJakob.

PR #8516, opened 2020-01-03
Creates a new alias ggpush! with no-verify and force flags

edersonbadeca:
Creates a new alias ggpush! with no-verify and force flags

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -146,6 +146,7 @@ compdef _git ggu=git-checkout
 alias ggpur='ggu'
 alias ggpull='git pull origin "$(git_current_branch)"'
 alias ggpush='git push origin "$(git_current_branch)"'
+alias ggpush!='git push origin "$(git_current_branch)" --no-verify --force'
 
 alias ggsup='git branch --set-upstream-to=origin/$(git_current_branch)'
 alias gpsup='git push --set-upstream origin $(git_current_branch)'


kbakdev (approved):

Addono:
Hmm, would rather have the `--no-verify` option be explicitly part of the alias. Skipping the Git hooks can be quite impactful - e.g. locally run unit tests, automated code formatting, and so on. So magically adding it seems suboptimal.

(Context, locally I have been using the following aliases to add the `--no-verify` flag, and I was looking if someone made a PR with more of them:
```bash
alias gpn="gp --no-verify"
alias gpfn="gpf --no-verify"
alias gpfn!="gpf! --no-verify"
alias gpsupn="gpsup --no-verify"
```
)

In my experience, adding the `n` is an easily generalizable pattern, which can be usefull for many other commands as well. So why not split out the `ggpush!` command into 
- `ggpush!`, which basically is `ggpush --force`, and
- `ggpushn!`, which would be `ggpush --no-verify --force` 

(Obviously, the expanded form of `ggpush` should be used when putting it into code, but just for breviety I write it like this.)

robbyrussell:
Thanks for this, and sorry nobody followed up at the time.

Force-pushing the current branch is already covered by `ggf` and `ggfl`, and I agree with the point @Addono raised here about `--no-verify` — pre-push hooks are usually running tests or formatters, and skipping them is better typed out deliberately than hidden in an alias.

PR #8518, opened 2020-01-04
Use official dotnet zsh plugin

harry-cpp:
## Standards checklist:

## Changes:

- Replaced the custom dotnet plugin with the official one
  - Official autocomplete has several more features than the custom one, like autocompletion of nuget packages (try doing `dotnet add package MonoGame.~` and it will autocomplete possible nuget packages)

## Other comments:

Link to the official plugin (the code base is MIT License): https://github.com/dotnet/toolset/blob/master/scripts/register-completions.zsh

### DIFF
--- a/plugins/dotnet/README.md
+++ b/plugins/dotnet/README.md
@@ -1,6 +1,6 @@
 # .NET Core CLI plugin
 
-This plugin provides completion for [.NET Core CLI](https://dotnet.microsoft.com/).
+This plugin provides completion and useful aliases for [.NET Core CLI](https://dotnet.microsoft.com/).
 
 To use it, add `dotnet` to the plugins array in your zshrc file.
 
@@ -21,4 +21,3 @@ plugins=(... dotnet)
 | da    | dotnet add       | Add a package or reference to a .NET project.                     |
 | dp    | dotnet pack      | Create a NuGet package.                                           |
 | dng   | dotnet nuget     | Provides additional NuGet commands.                               |
-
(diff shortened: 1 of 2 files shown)


cra0zy:
Hmm, so it turns out that there is a negative side to the script, and that is that it does not select filepaths... ever. Simple solution is to do:

```zsh
# zsh parameter completion for the dotnet CLI

_dotnet_zsh_complete() 
{
  local completions=("$(dotnet complete "$words")")

  # If the completion list is empty, just continue with filename selection
  if [ -z "$completions" ]
  then
    _arguments '*::arguments: _normal'
    return
  fi

  # This is not a variable assigment, don't remove spaces!
  _values = "${(ps:\n:)completions}"
}

compdef _dotnet_zsh_complete dotnet
```

harry-cpp:
Submitted the PR to the dotnet repo containing the plugin, it got merged, this should be good to go now as well.

larson-carter:
@harry-cpp Hello, can you please rebase to the master branch. Then upstream all of your improvements that way I can do a review on it. Also while you are at it can you please follow the new Github PR template?

Thanks -- Larson

harry-cpp:
> Hello, can you please rebase to the master branch.

Done.

> Then upstream all of your improvements that way I can do a review on it.

If you mean submit the few changes to the official dotnet plugin? Than that is already done:
https://github.com/dotnet/toolset/blob/master/scripts/register-completions.zsh

> Also while you are at it can you please follow the new Github PR template?

Done.

harry-cpp:
> After getting to looking into the code of this PR. This PR removes a bunch of functionality. Is there a specific reason for this?

The only real functionality lost are aliases for commands and descriptions, however it at the same time provides a much better autocomplete experience as it has several commands / arguments and custom stuff it can autocomplete now.

harry-cpp:
@mcornella done.

On a related note I feel like plugins should never come with aliases by default, its not something a plugin should set for you. Aliases are there so you can set them yourself for the commands you yourself often use.

PR #8519, opened 2020-01-05
Fix setting title via terminfo (e.g. on vt320)

franrogers:
The `title` function in `lib/termsupport.zsh` tries to use terminfo sequences to set the title if not using one of various known terminal emulators, but it's broken: it only issues a `tsl` with no argument (it takes one argument, and the status line also has to be cleared). On my VT320, this causes garbage to pile up in the status line. This patch fixes it.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -32,7 +32,8 @@ function title {
         # Try to use terminfo to set the title
         # If the feature is available set title
         if [[ -n "$terminfo[fsl]" ]] && [[ -n "$terminfo[tsl]" ]]; then
-	  echoti tsl
+	  echoti tsl 0
+	  echoti dl1
 	  print -Pn "$1"
 	  echoti fsl
 	fi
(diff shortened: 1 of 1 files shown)


franrogers:
My second commit makes it so the long window title is used instead of the short tab title when setting via terminfo, since what's being set is a status line, normally terminal-wide.

robbyrussell:
Thanks for this, and sorry it took six years to get an answer.

Your diagnosis was right... running prompt expansion over the capability string mangles a parameterized `tsl`, and on vt320 the working directory ends up inside the escape sequence. I've reimplemented it in #14052 with a `Co-Authored-By` trailer crediting you.

One change from your patch: it uses `echoti` for both capabilities so terminfo does the substitution itself, and leaves out the `echoti dl1`. Ghostty didn't exist when you opened this and it reaches the same branch, where `dl1` would inject `\e[M` into the middle of its OSC sequence.

Nothing needed from you. This closes when that merges.

PR #8521, opened 2020-01-05
virtualenvwrapper: rewrite init script location code

### LINKED ISSUE #8513: virtualenvwrapper plugin should search default location for virtualenvwrapper.sh on Debian
**Describe the bug**
The Debian virtualenvwrapper package installs the virtualenvwrapper.sh script into `/usr/share/virtualenvwrapper`. Unfortunately this location is not searched by the virtualenvwrapper plugin.

Adding this location to the path before invoking oh-my-zsh does not fix the problem because these scripts are installed without execute permission, hence are not located by the plugin.

Also, the bash completion lookup does not work either because again the wrong location is searched. On debian at least, bash completions are installed into `/usr/share/bash-completion/completions`.

From looking at the script I was unable to determine another workaround.

**To Reproduce**

1. Install virtualenvwrapper via `sudo apt-get install virtualenvwrapper`
2. Enable virtualenvwrapper plugin
3. Observe error when new shell opened

**Expected behavior**
Plugin should locate virtualenvwrapper.sh file and load it without issuing an error.

**System:**
 - OS: Ubuntu 19.10 (virtualenvwrapper 4.8.4-4)
 - Zsh version 5.7.1
 - Terminal emulator GNOME Terminal

randomphrase:
Rewrote the init script location code to remove duplication. Also added the default debian location for the file.

Fixes #8513

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -1,55 +1,29 @@
-virtualenvwrapper='virtualenvwrapper.sh'
-virtualenvwrapper_lazy='virtualenvwrapper_lazy.sh'
+function {
+    # search in these locations for the init script:
+    for f in $commands[virtualenvwrapper_lazy.sh] \
+             $commands[virtualenvwrapper.sh] \
+             /usr/share/virtualenvwrapper/virtualenvwrapper{_lazy,}.sh \
+             /usr/local/bin/virtualenvwrapper{_lazy,}.sh \
+             /etc/bash_completion.d/virtualenvwrapper \
+             /usr/share/bash-completion/completions/virtualenvwrapper
+    do
+        if [[ -f $f ]]; then
+            source $f
+            return
+        fi
+    done
+    print "[oh-my-zsh] virtualenvwrapper plugin: Cannot find virtualenvwrapper.sh.\n"\
+          "Please install with \`pip install virtualenvwrapper\`" >&2
+}
 
-if (( $+commands[$virtualenvwrapper_lazy] )); then
-  function {
-    setopt local_options
-    unsetopt equals
-    virtualenvwrapper=${${virtualenvwrapper_lazy}:c}
-    source ${${virtualenvwrapper_lazy}:c}
-    [[ -z "$WORKON_HOME" ]] && WORKON_HOME="$HOME/.virtualenvs"
-  }
-elif (( $+commands[$virtualenvwrapper] )); then
-  function {
-    setopt local_options
-    unsetopt equals
-    source ${${virtualenvwrapper}:c}
-  }
-elif [[ -f "/usr/local/bin/virtualenvwrapper.sh" ]]; then
-  function {
-    setopt local_options
-    unsetopt equals
-    virtualenvwrapper="/usr/local/bin/virtualenvwrapper.sh"
-    source "/usr/local/bin/virtualenvwrapper.sh"
-  }
-elif [[ -f "/usr/share/virtualenvwrapper/virtualenvwrapper.sh" ]]; then
-  function {
-    setopt local_options
-    unsetopt equals
-    virtualenvwrapper="/usr/share/virtualenvwrapper/virtualenvwrapper.sh"
-    source "/usr/share/virtualenvwrapper/virtualenvwrapper.sh"
-  }
-elif [[ -f "/etc/bash_completion.d/virtualenvwrapper" ]]; then
-  function {
-    setopt local_options
-    unsetopt equals
-    virtualenvwrapper="/etc/bash_completion.d/virtualenvwrapper"
-    source "/etc/bash_completion.d/virtualenvwrapper"
-  }
-else
-  print "[oh-my-zsh] virtualenvwrapper plugin: Cannot find ${virtualenvwrapper}.\n"\
-        "Please install with \`pip install virtualenvwrapper\`" >&2
-  return
-fi
 if ! type workon &>/dev/null; then
   print "[oh-my-zsh] virtualenvwrapper plugin: shell function 'workon' not defined.\n"\
         "Please check ${virtualenvwrapper}" >&2
   return
 fi
 
-if [[ "$WORKON_HOME" == "" ]]; then
-  print "[oh-my-zsh] \$WORKON_HOME is not defined so plugin virtualenvwrapper will not work" >&2
-  return
+if [[ -z "$WORKON_HOME" ]]; then
+  WORKON_HOME="$HOME/.virtualenvs"
 fi
 
 if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then


plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -1,48 +1,30 @@
-virtualenvwrapper='virtualenvwrapper.sh'
-virtualenvwrapper_lazy='virtualenvwrapper_lazy.sh'
+function {
+    setopt nullglob
```
Aloxaf: I suggest also set `local_options` here, so that it won't change user's shell options.

Alternatively, don't set `nullglob` here. Your code does not seem to use this feature.

randomphrase: Good point, will remove it.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -1,55 +1,28 @@
...
+    for f in $commands[virtualenvwrapper_lazy.sh] \
+             $commands[virtualenvwrapper.sh] \
+             /usr/share/virtualenvwrapper/virtualenvwrapper{_lazy,}.sh \
+             /usr/local/bin/virtualenvwrapper{_lazy,}.sh \
+             /etc/bash_completion.d/virtualenvwrapper \
+             /usr/share/bash-completion/completions/virtualenvwrapper
+    do
+        [[ -f $f ]] || continue
+        source $f
+        return
```
mcornella: You can use `break` instead so that you can remove the self-executing anonymous function.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -1,55 +1,28 @@
...
+function {
+    # search in these locations for the init script:
+    for f in $commands[virtualenvwrapper_lazy.sh] \
+             $commands[virtualenvwrapper.sh] \
+             /usr/share/virtualenvwrapper/virtualenvwrapper{_lazy,}.sh \
+             /usr/local/bin/virtualenvwrapper{_lazy,}.sh \
+             /etc/bash_completion.d/virtualenvwrapper \
+             /usr/share/bash-completion/completions/virtualenvwrapper
+    do
+        [[ -f $f ]] || continue
```
mcornella: This would be more elegant turned into a plain old `if`.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -1,55 +1,28 @@
...
   print "[oh-my-zsh] virtualenvwrapper plugin: shell function 'workon' not defined.\n"\
         "Please check ${virtualenvwrapper}" >&2
   return
 fi
 
-if [[ "$WORKON_HOME" == "" ]]; then
-  print "[oh-my-zsh] \$WORKON_HOME is not defined so plugin virtualenvwrapper will not work" >&2
-  return
+if [[ -z "$WORKON_HOME" ]]; then
+  WORKON_HOME="$HOME/.virtualenvs"
```
mcornella: The logic is slightly different now, I should check that it still works as expected.

mcornella:
We also have to make sure we don't reintroduce https://github.com/ohmyzsh/ohmyzsh/issues/3893 after removing the `setopt no_equals` option.

randomphrase:
Great feedback, thanks. Will look at these issues soon.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
randomphrase: Hi, can you explain this comment? If a break were used here it would result in the error message being displayed, incorrectly.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
randomphrase: Previously `WORKON_HOME` would only be set if the `virtualenvwrapper_lazy.sh` script was found. I think the reasoning here is that the non-lazy version of the script will set this variable if found. However, I think it is simpler to just set the variable if it had not already been set, regardless of whether the lazy or non-lazy version was found.

randomphrase:
> We also have to make sure we don't reintroduce #3893 after removing the `setopt no_equals` option.

I was not able to replicate #3893 with my install of virtualenvwrapper. However I don't see the harm in unsetting the equals option prior to sourcing the init script. Perhaps this can be removed in the future.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
mcornella: I didn't see that, you're right.

mcornella:
> I was not able to replicate #3893 with my install of virtualenvwrapper. However I don't see the harm in unsetting the equals option prior to sourcing the init script. Perhaps this can be removed in the future.

I downloaded the [jessie version of the Debian distribution](https://packages.debian.org/jessie/all/virtualenvwrapper/download) of the package and I can get an error when sourcing `/etc/bash_completion.d/virtualenvwrapper`. This are the file contents, which doesn't error out:
```
# bash completion snippet for virtualenvwrapper

USE_FULL=no

if [ "$USE_FULL" = "yes" ]; then
    . /usr/share/virtualenvwrapper/virtualenvwrapper.sh
else
    . /usr/share/virtualenvwrapper/virtualenvwrapper_lazy.sh
fi
```
If I double the equal operator in the conditional, I get:
```
virtualenvwrapper:5: = not found
```

Since the version of the file distributed is already fixed, and it is in fact an error in syntax (a single `[` expects a single equal, as the POSIX standard demands), I think we can get rid of the hack we added to "fix" it and proceed with the current code.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
mcornella: Looks like it does the same, yes.

plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
```diff
@@ -1,55 +1,35 @@
...
-    source "/usr/local/bin/virtualenvwrapper.sh"
-  }
-elif [[ -f "/usr/share/virtualenvwrapper/virtualenvwrapper.sh" ]]; then
-  function {
+function {
+    # The equals option must be unset in order for virtualenvwrapper's init
+    # script to load (see #3893). Note that recent versions of virtualenvwrapper
+    # no longer seem to have this issue, so possibly this is no longer needed
     setopt local_options
     unsetopt equals
```
mcornella: Sorry, it looks like [we can get rid of this](https://github.com/ohmyzsh/ohmyzsh/pull/8521/files#issuecomment-581049715).

randomphrase: OK, great

mcornella:
Thanks @randomphrase!

PR #8523, opened 2020-01-06
Fix Symfony Console Selection Priority

huebs:
This change selects the last (instead of the first) `console` executable found. This has the effect of selecting `bin/console` instead of `app/console` when both exist. The `app/console` executable has been deprecated and displays a warning when it's called. Without this change, using any of the plugin's aliases will also cause the warning to display.

### DIFF
--- a/plugins/symfony2/symfony2.plugin.zsh
+++ b/plugins/symfony2/symfony2.plugin.zsh
@@ -1,7 +1,7 @@
 # Symfony2 basic command completion
 
 _symfony_console () {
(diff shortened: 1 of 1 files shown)


bartekpacia (approved):
thank you!

kbakdev (approved):

robbyrussell:

Symfony 2 has been end-of-life since 2018, and we now ship `symfony` and `symfony6` alongside it, so `symfony2` isn't a plugin we're still fixing.

PR #8537, opened 2020-01-14
Add Testing for PRs and pushes

ModdedGamers:
This was big, and thanks to @larson-carter for his help along the way.


ModdedGamers:
This PR also include differently styled badges, which I think look better than the current badges. These changes can be discarded if needed.

larson-carter:
@ModdedGamers 

See - "Run sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)" "" --unattended && exit 0"

I think every PR that comes in it is going to try to install from the master branch. This is extremely close to being correct. We just need to make it work on every new PR branch.

mcornella:
I agree with Larson, this is a big improvement over not having any testing. To make it work for Pull Requests we can use the [`checkout` action](https://github.com/actions/checkout) to have access to the install.sh script for that particular commit, be it in a push event or a pull_request one. We can also taylor the pull_request event so that it only fires on a `opened` or a `synchronize` event, which is when new code is posted.

Another optimization is to only try to run the installer when a change has been made to it. It is trickier, so I'll start with integrating the checkout action and work from there.

> This PR also include differently styled badges, which I think look better than the current badges. These changes can be discarded if needed.

The style is very personal and everyone will have a different opinion. I personally chose these ones to be consistent with what the majority of other open source projects seem to be using. Let's take these out of the PR and focus on getting the workflow just right.

ModdedGamers:
Alright. I'll modify the script to take these changes into account and remove the badge changes from the PR.

ModdedGamers:
I've updated the script so that it takes all suggested changes into account.

ModdedGamers:
Alright!

ModdedGamers:
I've made a new PR( #8569) to help with making a cleaner commit history, and I'll be closing this PR now.

PR #8538, opened 2020-01-14
asdf: fix completions if ASDF_DIR is already exported

aserrallerios:
Asdf completions are broken if the plugin is loaded after the `$ASDF_DIR/asdf.sh` application is loaded (asdf.sh exports ASDF_DIR) or if the ASDF_DIR variable is exported at all (manually for example) before the plugin is loaded.


RobLoach:
This should be good. Thanks a lot!

PR #8542, opened 2020-01-15
Add MSYS2 support to system clipboard integration

svraka:
This pull request enables the clipboard features on the Cygwin-based [MSYS2](https://www.msys2.org/). MSYS2 has the same `/dev/clipboard` device as Cygwin but MSYS2 reports `msys` in `$OSTYPE`.

### DIFF
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -24,7 +24,7 @@ function clipcopy() {
     else
       cat $file | pbcopy
     fi
-  elif [[ $OSTYPE == cygwin* ]]; then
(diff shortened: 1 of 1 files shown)


lib/clipboard.zsh
```diff
@@ -24,7 +24,7 @@ function clipcopy() {
     else
       cat $file | pbcopy
     fi
-  elif [[ $OSTYPE == cygwin* ]]; then
+  elif [[ $OSTYPE == (cygwin*|msys)* ]]; then
```
mcornella: It should read `(cygwin|msys)*`. Same below.

svraka: Thanks, done.

PR #8549, opened 2020-01-17
changed apple icon to awesomefont compatible

xxKeith20xx:
The old apple icon didn't appear once I installed a https://github.com/ryanoasis/nerd-fonts/blob/master/patched-fonts/UbuntuMono. Once I copied in the character taken from https://fontawesome.com/icons/apple , it appears properly.

### DIFF
--- a/themes/apple.zsh-theme
+++ b/themes/apple.zsh-theme
@@ -1,5 +1,5 @@
 function toon {
-  echo -n ""
+  echo -n ""
 }
 
 get_git_dirty() {


chris-martinelli:
```
 ~/ echo  | hexdump
0000000 ef 85 b9 0a
0000004
```

mcornella:
@tricon you made the `apple` theme first (#552); do you remember the font you used?

tricon:
I am pretty sure it was Inconsolata.

xxKeith20xx:
looking at this again I'm not sure if you'd want this to work without special fonts or w/ special fonts only

```
 ~/.oh-my-zsh/ [master*] echo  | hexdump
0000000 ef a3 bf 0a
0000004
```

tricon:
In retrospect, I agree. The Apple logo is one that Apple places in the
Unicode PUA (F8FF specifically): https://en.wikipedia.org/wiki/Private_Use_Areas

xxKeith20xx:
so technically I guess I was asking for a new theme :p - I'll close this PR

mcornella:
Note that the way @tricon coded the theme, you can redefine the `toon` function with a different character, in your zshrc file (after OMZ is sourced).

xxKeith20xx:
oh nice! didn't realize that @mcornella

PR #8550, opened 2020-01-18
gitfast: proper synchronization

felipec:
There were some improper updates, this series of patches should address the issues that arised from that.

I added an `update` script precisely to avoid synchronization issues, and this script was completely ignored when updates were issued.

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -39,7 +39,7 @@ if [ -z "$script" ]; then
 		test -f $e && script="$e" && break
 	done
 fi
-GIT_SOURCING_ZSH_COMPLETION=y . "$script"
+ZSH_VERSION='' . "$script"
 
 __gitcomp ()
 {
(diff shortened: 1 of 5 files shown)


larson-carter:
I think that we need to break this down into a few more PRs rather than just being in one. 

For example
1 - Issues
2 - Update Script
3 - Patch for Bash upgrades

That would make this migration process so much easier than having to merge 6 commits ontop of +772 −610 changes.

larson-carter:
@felipec 

I've looked into this in way more detail today. I would do one PR for this: plugins/gitfast/git-completion.bash then another pertaining to issues and patches put together. It appears that plugins/gitfast/git-completion.bash has all of the syntax chances that we would need to have a clutter free PR to really dive into.

This PR however is amazing on its own and I want to start up front and say thank you for all of the hard work. So lets see what @mcornella has to say.

mcornella:
LGTM

PR #8551, opened 2020-01-18
fix(colored-man-pages): Suppress `env` shell lookup

jrwrigh:
If `env` is aliased or there is a function with a name conflict, this may cause `man` to fail. Prepending `env` with `command` "bypasses" any aliases or functions. See the issue I ran into in this [StackOverflow answer](https://unix.stackexchange.com/a/562814/297140).

### DIFF
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -17,7 +17,7 @@ EOF
 fi
 
 function colored() {
-	env \
+	command env \
 		LESS_TERMCAP_mb=$(printf "\e[1;31m") \
 		LESS_TERMCAP_md=$(printf "\e[1;31m") \
 		LESS_TERMCAP_me=$(printf "\e[0m") \


TriMoon:
The real problem is that the `grc` plugin you mentioned overloads `env`, which is the sole reason for all problems related to any use of `env` in **ANY** program because `env` is a base OS command...
So instead of trying to fix something that is supposed to be working as expected, just because some plugin does something wrong, i would advice to contact the maker of that plugin to fix their bad coding...
I would have said the same in case it would overload `cat` etc...

Bottom-line IMHO: Do not use `grc` until it is fixed...

jrwrigh:
So I agree with the bottom line. I would never had made this PR if It were that simple. The issue is that `grc` isn't maintained anymore... and neither is any command colorizer that I can find either (suggestions open). But that's neither here nor there. If this PR doesn't go through, I'll just end up ditching using `grc`. :man_shrugging: 

So the reason I am proposing the changes is really two fold:
1) It fixes my personal immediate problem (duh)
2) Looking at the commit history of the plugin, there seems to be an appetite for making command references more robust. 

Take PR #4702. They could've simply just referenced `less` directly and made the it search `$PATH` for the binary, (in analogy to directly referencing `env` before my additional commit). In fact, my default value for `$PAGER` is just `less`, not `/usr/bin/less`. Instead, they indirectly reference`less` via `$commands[less]`. I could argue the same thing could/should be done with `env` for robustness as well.

Granted, the equivalent of my 1st reason for changing in #4702 is quite a bit more broad (it fixes usage on an entire distro).

mcornella:
This is an easy merge and, while I agree that `env` shouldn't be overriden (especially not if it breaks the original behavior), we've made similar changes in the past. Thanks for your contribution!

jrwrigh:
Welcome!

PR #8554, opened 2020-01-20
Fix 3796: Taskwarrior completion doesn't work

### LINKED ISSUE #3796: Taskwarrior completion doesn't work

Wabri:
Compdef executions don't work without this change.

Any of this not work to solve the issue #3796:

- Try deleting the zcompdump files: rm ~/.zcompdump* and restart the terminal.
- Try running compaudit | xargs chmod g-w,o-w
- Start zsh, then run autoload -T _task, and then try task [TAB].
- Try switching your locale temporarily: run LC_ALL=C and then try again to autocomplete.

Fix #3796


mcornella:
See https://github.com/ohmyzsh/ohmyzsh/pull/8002#issuecomment-577376136.

53jk1 (approved):

PR #8558, opened 2020-01-21
Add kgcrd and kgapi aliases to kubectl plugin

alaa:
Add kubectl aliases to:

- get custom resource definitions using `kgcrd`
- get k8s api services using `kgapi`


bartekpacia:
@alaa Forgive us the late response. Thank you for the contribution, it'll be probably reviewed and merged.

PR #8563, opened 2020-01-25
git: add convenience aliases for `git apply` and `git am`

larumbe:
These are common operations when applying patches from mbox files on
mailing list based projects like U-Boot or the Linux
Kernel.


mcornella:
Can you document this in the README?

PR #8564, opened 2020-01-25
feat(fossil): add completion for `fossil add`

CGenie:
For fossil it only makes sense to 'add' files which aren't currently
tracked in the repo.

http://fossil-scm.org/home/help?cmd=add


CGenie:
Ah, files with spaces in them are split into 2 entries in the completion. Any ideas why? :)

53jk1 (approved):

PR #7370, opened 2018-10-29
Add adonis plugin

dimensi0n:
This plugin provides autocompletion for the AdonisJs framework

PR #8566, opened 2020-01-25
Added AdonisJS plugin with aliases

jonaselan:
I know the exist another pull request (#7370) for add autocompletion. Maybe we can merge the two to build a better plugin.

### DIFF
--- a/plugins/adonisjs/README.md
+++ b/plugins/adonisjs/README.md
@@ -1,0 +1,54 @@
+# Laravel
+
+This plugin adds aliases for [AdonisJS Framework](https://adonisjs.com/).
+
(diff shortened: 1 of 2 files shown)


dimensi0n:
Good idea

robbyrussell:
Thanks for putting this together, and sorry it waited so long.

These aliases wrap the old `adonis` CLI, which AdonisJS retired in v5 — current projects run `node ace` instead, so the plugin wouldn't do much on a modern install.

Closing for that reason. If you're still working with AdonisJS and want to write one against `node ace`, we'd take a look at that. Thanks for contributing.

PR #8568, opened 2020-01-25
Add calling VS Code Insiders

babakks:
The current [VS Code](https://github.com/microsoft/vscode) plugin (at `plugins/vscode`) only invokes the *Stable* release installation, unless the user make changes to `.zshrc` (which are already explained in the README.md of the plugin).

However, some enthusiastic users, including myself, just install the *Insiders* release to follow up on the upcoming features of VS Code. For those users to be able to use this great plugin out of the box, I've modified the script to use `code-insidiers` (instead of `code`), **if the Insiders release was the only available installation** on the host machine.

### DIFF
--- a/plugins/vscode/README.md
+++ b/plugins/vscode/README.md
@@ -1,6 +1,6 @@
-# VS code
+# VS Code
 
-This plugin makes interaction between the command line and the code editor easier.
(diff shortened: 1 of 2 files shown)


babakks:
I've also updated `README.md` file to assure readers about the automatic binding to the Insiders version.

PR #8569, opened 2020-01-27
Create Github Action to run tests

ModdedGamers:
In order to maintain a cleaner commit history, I made a new branch and will be closing the other PR. This is the only way I see of giving this work a cleaner commit history.
@larson-carter

### DIFF
--- a/.github/workflows/check-suite.yml
+++ b/.github/workflows/check-suite.yml
@@ -1,0 +1,25 @@
+name: Check Suite
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - master
+  push:
+    branches: 
+      - master
+
+jobs:
+  verify-installer:
+    name: Install on Ubuntu
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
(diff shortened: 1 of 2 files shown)


mcornella:
I reorganized some stuff and added a step to check zsh syntax of known zsh files, inspired mostly by the [Homebrew tests workflow file](https://github.com/Homebrew/brew/blob/master/.github/workflows/tests.yml). It's a bit wasteful for now, because for instance, it doesn't skip testing the installer step if there weren't any changes made to it. It doesn't seem to take too long though, so for now it's ok to ship it IMO. I also replaced the workflow badge from the one GitHub provides.

Finally, I had to enable ['Local and third party Actions'](https://github.blog/2019-10-03-github-actions-adds-policy-controls-for-organization-and-repository-owners/) to be run so that GitHub would run the workflow for this PR. It's unclear to me if that has to be kept enabled after the workflow is merged or not; will investigate later.

So to recap, to me this is now ready to be merged. Let me know what you think.

ModdedGamers:
Fine with me. I think that we should turn it on as this is pretty much an alpha for Actions integration, so if we need to update this(which we will need to) we'll need to keep local and third-party actions enabled. @mcornella

PR #8582, opened 2020-01-30
fix: Update tmux plugin to use modern terminfo.

hermitmaster:
The standard terminfo for modern tmux is `tmux` or `tmux-256color`. Updating to reflect the change.

### DIFF
--- a/plugins/tmux/README.md
+++ b/plugins/tmux/README.md
@@ -35,7 +35,7 @@ The plugin also supports the following -
 | `ZSH_TMUX_AUTOQUIT`                 | Automatically closes terminal once tmux exits (default: `ZSH_TMUX_AUTOSTART`) |
 | `ZSH_TMUX_FIXTERM`                  | Sets `$TERM` to 256-color term or not based on current terminal support       |
 | `ZSH_TMUX_ITERM2`                   | Sets the `-CC` option for iTerm2 tmux integration (default: `false`)          |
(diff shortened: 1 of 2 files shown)


mcornella:
Can you also fix the comments?

hermitmaster:
Will do. I’ll update the PR in a few mins.

mcornella:
Thanks!

kennith:
@hermitmaster Do you have problem with OSX Catalina with Tmux 3.0a? I am having problem with the backspace key and Vim is generating this error:

```
E558: Terminal entry not found in terminfo
'tmux-256color' not known. Available builtin terminals are:
    builtin_ansi
    builtin_xterm
    builtin_iris-ansi
    builtin_dumb
defaulting to 'ansi'
```

I set it back to use 
```
ZSH_TMUX_FIXTERM_WITH_256COLOR=screen-256color
ZSH_TMUX_FIXTERM_WITHOUT_256COLOR=screen
```
is fine.

mcornella:
Reverted the change: #8597. Update with `upgrade_oh_my_zsh`.

hermitmaster:
I tested with  tmux 2.9a on Fedora.
See: https://github.com/tmux/tmux/wiki/FAQ#why-do-you-use-the-screen-terminal-description-inside-tmux

mcornella:
For what I searched, macOS doesn't have a terminfo entry for tmux and tmux-256color. I'm on my phone, but it was an issue in the tmux github repository.

hermitmaster:
MacOS probably uses an ancient version of curses. /shrug

PR #8584, opened 2020-01-30
feat: Introduce exa plugin.

hermitmaster:
Introduce a plugin that adds commonly used `exa` aliases.


EvanCarroll (commented):
lgtm

EvanCarroll:
Why wasn't this merged?

mcornella:
> Why wasn't this merged?

No testers, no +1s. There are a lot of new plugins submitted and we have no way of knowing which ones will be actually used. So they stay in limbo.

PR #8594, opened 2020-02-04
✨ feat(alias): add git checkout track

bouteillerAlan:
add alias for git checkout ``alias gct='git checkout --track'``


mcornella:
I don't believe you use this often enough to require an alias. Either way, we already have a ton of aliases in the git plugin and this one doesn't seem worth it.

Thanks for your contribution though

PR #8601, opened 2020-02-06
Add bitwarden useful aliases

xjunior:
Adds two bitwarden aliases:

* `bwunlock` to unlock the vault and keep the session open in the current shell
* `bwpass` to get a password (useful for scripts) and keep or reuse the shell's session

### DIFF
--- a/plugins/bitwarden/README.md
+++ b/plugins/bitwarden/README.md
@@ -1,0 +1,32 @@
+# Bitwarden plugin
+
+This plugin provides basic bitwarden aliases to manage your bitwarden session
+
+-------
+
+
+This adds two aliases:
+
+### `bwpass`
+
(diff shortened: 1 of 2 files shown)


jmeridth (approved):

casaper:
@jmeridth @xjunior 

Recently I created the PR #11181  and only now I've noticed that this one here is here since 2020.

#11181 also handles completions, and it does the same as this PR's aliases with functions.
The functions are obviously more complicated, but have the advantage that the BW_SESSION env variable would be available to new terminals too.

However: How about combining/mergeing #8601 and #11181 in one, since they tackle the same thing?

xjunior:
@casaper I'm ok with closing this one and keeping #11181 if you think it does everything this one does. I haven't been using these aliases myself for a while now so I'm not entirely sure they still work.

xjunior:
Close in favor of https://github.com/ohmyzsh/ohmyzsh/pull/11181

PR #8606, opened 2020-02-08
[Security] Update dotenv plugin to prompt before executing env

mazen160:
Hi all,

I have encountered an issue on the usage of .dotenv plugin.  Sourcing arbitrary .env files can lead a direct remote code execution on user machines. source doesn't soley read environment variables, it executes the code.

If I download a malicious repo, with a .env file. Then it would be executed with informing me about the execution of the source file.
```bash
$> cat .env             
touch /tmp/hacked
head -n 1 /etc/passwd
```
```bash
$> cd downloaded-dir
root:x:0:0:root:/root:/bin/bash
```

Fix: we can prompt the user before the execution.
```
$> cd x   
zsh: source .env file in the directory [Y/n]?: n
```

### DIFF
--- a/plugins/dotenv/dotenv.plugin.zsh
+++ b/plugins/dotenv/dotenv.plugin.zsh
@@ -1,13 +1,12 @@
 source_env() {
   if [[ -f $ZSH_DOTENV_FILE ]]; then
-  	
-    # confirmation before sourcing .env file
-    echo -n "dotenv: source .env file in the directory [Y/n]?: "
-    read confirmation
-    if [[ ($confirmation == "n") || ($confirmation == "N") ]]; then
+    # confirm before sourcing .env file
+    local confirmation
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks, Mazin

zyhou:
It is possible to override the confirmation config? I don't want to answer the question every time. Maybe something like `ZSH_DOTENV_DISABLE_CONFIRMATION=false` and active the confirmation by default

mcornella:
Sure, we could add a ZSH_DOTENV_PROMPT=false condition, do you think you can submit a PR?

zyhou:
Yes I can :)

mazen160:
Hi all,
I published a blog post regarding the finding: https://mazinahmed.net/blog/ohmyzsh-dotenv-rce/
Thanks for fixing the issue quickly!

PR #8607, opened 2020-02-09
Update OSX to macOS

charlesrocket:
This updates `osx` plugin and docs.


mcornella:
This will break people's setups. It might not be "correct" to use `osx`, but it really doesn't matter as much as you think it does.

charlesrocket:
I wouldn’t even notice if there weren't macos lines already in docs. Only reason why :) And also felt like it wouldn’t be easier to open such PR later.

mcornella:
I understand. I'll close this, but we might revisit this in the near future. Thanks for your contribution!

PR #8610, opened 2020-02-10
arcanist: add `arc diff --create` alias

gomesalexandre:


larson-carter:
@gomesalexandre Hello. Can you please fill out the proper PR template. After that is done I can take a look at this PR.

PR #8618, opened 2020-02-12
Support getting fzf for NixOS and Void Linux

### LINKED ISSUE #8585: fzf plugin: Add support for voidlinux
On voidlinux the fzf base directory is `/usr/share/doc/fzf`. Like in Arch there is no `shell` directory in under fzf_base. 
This causes the `setup_using_base_dir` function to fail when initializing the plugin. 

**To Reproduce**
On voidlinux:
1. Enable the plugin
2. Start a new session
3. See the error

**Expected behavior**
The expected behaviour would be that the plugin initializes correctly. 

**System:**
 - OS: voidlinux
 - Zsh 5.7.1
 - Terminal emulator: kitty

**Additional context**
I already fixed this locally and could provide a pull request, if that would help.

r-darwish:
EDITED by @mcornella:
This PR also works for Void Linux, so I modified it to mention that as well.

Closes #8548
Fixes #8585
Closes #8623

### DIFF
--- a/plugins/fzf/fzf.plugin.zsh
+++ b/plugins/fzf/fzf.plugin.zsh
@@ -10,6 +10,7 @@ function setup_using_base_dir() {
     if [[ -z "${fzf_base}" ]]; then
         fzfdirs=(
           "${HOME}/.fzf"
+          "${HOME}/.nix-profile/share/fzf"
           "/usr/local/opt/fzf"
           "/usr/share/fzf"
           "/usr/local/share/examples/fzf"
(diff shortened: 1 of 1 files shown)


abctaylor:
Is the NixOS package updated? It appears not (https://github.com/ohmyzsh/ohmyzsh/issues/9146). I still get the problematic logic when installing oh-my-zsh from Nix that doesn't cover NixOS. The conjunction for `-f /etc/arch-release` should be gone, per the master :) 
```
# Fix fzf shell directory for Archlinux package
        if [[ ! -d "${fzf_base}/shell" ]] && [[ -f /etc/arch-release ]]; then
          fzf_shell="${fzf_base}"
        else
          fzf_shell="${fzf_base}/shell"
        fi
```

PR #8620, opened 2020-02-12
Executing gradle literally, even if an alias exists (gradle plugin)

### LINKED ISSUE #8046: gradle-or-gradlew:5: maximum nested function level reached; increase FUNCNEST?
**Describe the bug**

function recursion 

**To Reproduce**

1. Enable plugin gradle and mvn 
2. gradle-or-gradlew:5: maximum nested function level reached; increase FUNCNEST?
3. the same problem for mvn plugin 

**Expected behavior**

**Screenshots or recordings**

**System:**
 - OS: Manjaro 18.0.4
 - Zsh version 5.7.1 
 - Terminal emulator Qterminal

**Additional context**

like the gradle and mvn plugin

```sh
gradle-or-gradlew() {
	if [ -f ./gradlew ] ; then
		echo "executing gradlew instead of gradle";
		./gradlew "$@";
	else
		gradle "$@";
	fi
}

alias gradle=gradle-or-gradlew;
```

remove the alias will fix this problem, or change alias

jonatan-ivanov:
- Similarly to the mvn plugin
- Without this fix, the shell crashes in some cases
- fixes #8046

### DIFF
--- a/plugins/gradle/gradle.plugin.zsh
+++ b/plugins/gradle/gradle.plugin.zsh
@@ -7,7 +7,7 @@ gradle-or-gradlew() {
 		echo "executing gradlew instead of gradle";
 		./gradlew "$@";
 	else
-		gradle "$@";
+		command gradle "$@";
 	fi
 }
 


mcornella:
Thanks!

PR #8622, opened 2020-02-13
added way to rename branch

ujwaldhakal:
## Standards checklist:

## Changes:

- [x] Added method to rename GitHub branch locally and remotely

## Usage
`grename oldbranch newbranch`


larson-carter:
@ujwaldhakal Can you please follow the Proper Pull Request template that we have setup. It makes the process of reviewing PRs so much easier and quicker. Otherwise PRs that have followed the proper layout will get looked at first.

ujwaldhakal:
hi @larson-carter I have re-implemented your default PR layout can you please review and tell if it's ok

plugins/git/git.plugin.zsh
```diff
@@ -255,3 +255,21 @@ alias glum='git pull upstream master'
...
+  then
+    echo "use command like 'grename old_branch new_branch'"
+    return
+  fi
+
+  if [ -z "$2" ]
+  then
+    echo "use command like 'grename old_branch new_branch'"
+    return
+  fi
```
mcornella: This check can be added to the previous if

plugins/git/git.plugin.zsh
```diff
@@ -255,3 +255,21 @@ alias glum='git pull upstream master'
 
 alias gwch='git whatchanged -p --abbrev-commit --pretty=medium'
 alias gwip='git add -A; git rm $(git ls-files --deleted) 2> /dev/null; git commit --no-verify --no-gpg-sign -m "--wip-- [skip ci]"'
+
+function grename() {
+  if [ -z "$1" ]
+  then
+    echo "use command like 'grename old_branch new_branch'"
```
mcornella: The message here could be better, i.e. follow common conventions. This is better:
```zsh
echo "Usage: $0 old_branch new_branch"
```
Also, use `return 1` to indicate a failure to run the command.

plugins/git/git.plugin.zsh
ujwaldhakal: I have added next if on a single line

plugins/git/git.plugin.zsh
ujwaldhakal: Thank you for review. I have done some changes please let me know if there is anything i would need to do

PR #8630, opened 2020-02-16
Set oh-my-zsh.sh script as executable

aaoferreira:
## Standards checklist:

## Changes:

- Set `oh-my-zsh.sh` script as executable

## Other comments:

I noticed that oh-my-zsh.sh is being checked-out as a non executable file (specially annoying because I have to do/undo this on every update), so I'm submitting this PR to update it on the git tree so it's automatically set as executable on new checkouts.


mcornella:
The file is sourced in the zshrc file, which is not the same as executed. Setting as executable would encourage a wrong usage of the init script.

aaoferreira:
Indeed! Intriguing that I really had issues with this, probably me doing something stupid then 😁

PR #8645, opened 2020-02-18
Add an additional 'git pull' alias

boxxroom:
## Standards checklist:

- [√] The PR title is descriptive.
- [√] The PR doesn't replicate another PR which is already open.
- [√] I have read the contribution guide and followed all the instructions.
- [√] The code follows the code style guide detailed in the wiki.
- [√] The code is mine or it's from somewhere with an MIT-compatible license.
- [√] The code is efficient, to the best of my ability, and does not waste computer resources.
- [√] The code is stable and I have tested it myself, to the best of my abilities.

## Changes:

- Add an additional 'git pull' alias as `gpl` keeping the `gl` for backwards compatibility. This seems more instinctive as an alias and more inline with the git alias abbreviations as opposed to the current git `gl` alias, while also saving time  having to look it up and t

- With `gp` alias referring to `git push` it seems better fitted that `gpl` refers `git pull`

- `gl` would be more inline with/suited to 'git log' with the rest of the `gl` aliases referring to `git log`

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -160,6 +160,7 @@ alias gk='\gitk --all --branches'
 alias gke='\gitk --all $(git log -g --pretty=%h)'
 
 alias gl='git pull'
+alias gpl='git pull'
 alias glg='git log --stat'
 alias glgp='git log --stat -p'
 alias glgg='git log --graph'


adamchainz:
I agree that `gl` for `git pull` is a bit of a weird choice. But since then we have `gup` for `git pull --rebase`, which is kind of the new “base“ `git pull` alias, as other `gup*` aliases extend it. `--rebase` is an easier option for new Git users to understand, so I think that’s fine.

So perhaps we shouldn't add another `git pull` “base“ alias, to avoid complicating things.

I would say we can close this one @carlosala .

carlosala:
I agree with Adam. It's kinda true that `gl` is a weird choice, but it has its place in our workflows.
Closing, thanks for the contribution!

boxxroom:
Thanks for coming back to this all, it's much appreciated. 

While I still believe `gpl` is better suited to `git pull` and `gl` better suited to the `git log` family. It has been so long since this pull request I am now accustom to the `gl` for `git pull`.  :) 

Thanks again. 👍🏼

PR #8651, opened 2020-02-19
Add random theme and consolidate logic from init and themes plugin

### LINKED ISSUE #3036: Random theme picker should skip themes that rely on uninstalled commands
Feature request: The random theme picker should do some sort of check to determine whether a given theme has all the dependencies it needs, and skip themes for which this is not the case.

It's annoying to open a new shell and see something like
`Can't open perl script "/home/scp1/bin/foopath": No such file or directory`
or 
`prompt_jnrowe_precmd:3: command not found: zsh_path`

### LINKED ISSUE #3704: Feature request: Allow user to blacklist themes from being picked by the random selector

### LINKED ISSUE #7696: Add custom themes to the random theme picker
The random theme picker searches only the themes from the $ZSH/themes folder. It will be nice to make it choose from the $ZSH_CUSTOM/themes folder also.

mcornella:
## Standards checklist:

## Changes:

- Put the random theme logic in a theme of the name random, taking it out of the init script and the theme plugin (closes #2962, closes #3337).
- Change the logic such that themes in `$ZSH_CUSTOM` are also taken into account (fixes #7696, closes #7697).
- Add array variable to blacklist themes from the random pool (`$ZSH_THEME_RANDOM_BLACKLIST`) (fixes #3704, closes #3703).
- In the themes plugin, the `theme` function errors when it can't find the theme provided.
- In the themes plugin, the `lstheme` function uses wildcard expansion and looks in the `$ZSH_CUSTOM` folder as well. Also, the themes are printed in one line with spaces in between, instead of in a column separated by newlines.
- In the themes plugin, the completion function has been placed inside the `.plugin.zsh` file.

## Other comments:

Closes #4605.
Closes #3036 (the blacklist setting solves this, plus now most default themes check for their dependencies and gracefully fail).

## Tested

- For the `random` theme:
  + `ZSH_THEME_RANDOM_CANDIDATES` works as expected.
  + `ZSH_THEME_RANDOM_BLACKLIST` works as expected (only applied when `ZSH_THEME_RANDOM_CANDIDATES` isn't set up, *should probably document this in the wiki).
  + When a theme isn't found (such as when it was specified in `ZSH_THEME_RANDOM_CANDIDATES`, but didn't exist), the current behavior is to fail with an error:
    ```
    $ ZSH_THEME=random zsh
    [oh-my-zsh] Random theme 'nonexistent' not found
    ```
    **Should nonexistent themes be purged from the list of candidates?**
- For the `themes` plugin:
  + The completion for the `theme` function has been tested and it works.
  + The `lstheme` function finds themes both in the default folder and the `$ZSH_CUSTOM/*` and `$ZSH_CUSTOM/themes/*.zsh-theme`.

## Not working

- Theme names that contain paths, such as the powerlevel10k that tells users to put it in `$ZSH_CUSTOM` under a `powerlevel10k` folder, such that the theme name becomes `powerlevel10k/powerlevel10k`:
  + The `random` theme doesn't look in subfolders, but if `powerlevel10k/powerlevel10k` is specified in the candidates variable it will be correctly picked up.
  + The `themes` plugin's completion doesn't look in subfolders either, but the `theme` function will correctly source the theme if specified in the first argument (`theme "powerlevel10k/powerlevel10k"` works).

### DIFF
--- a/themes/random.zsh-theme
+++ b/themes/random.zsh-theme
@@ -1,0 +1,10 @@
+if [[ "${(t)ZSH_THEME_RANDOM_CANDIDATES}" = "array" ]] && [[ "${#ZSH_THEME_RANDOM_CANDIDATES[@]}" -gt 0 ]]; then
+  themes=($ZSH/themes/${^ZSH_THEME_RANDOM_CANDIDATES}.zsh-theme)
+else
+  themes=($ZSH/themes/*zsh-theme)
+fi
+N=${#themes[@]}
+((N=(RANDOM%N)+1))
+RANDOM_THEME=${themes[$N]}
+source "$RANDOM_THEME"
+echo "[oh-my-zsh] Random theme '$RANDOM_THEME' loaded..."
--- a/themes/random.zsh-theme
+++ b/themes/random.zsh-theme
@@ -5,19 +5,23 @@ if [[ "${(t)ZSH_THEME_RANDOM_CANDIDATES}" = array && ${#ZSH_THEME_RANDOM_CANDIDA
   # Use ZSH_THEME_RANDOM_CANDIDATES if properly defined
   themes=($ZSH_THEME_RANDOM_CANDIDATES)
 else
-  # Look for themes in $ZSH_CUSTOM and $ZSH and add only the theme name (:t)
+  # Look for themes in $ZSH_CUSTOM and $ZSH and add only the theme name
   themes=(
     "$ZSH_CUSTOM"/*.zsh-theme(N:t:r)
     "$ZSH_CUSTOM"/themes/*.zsh-theme(N:t:r)
     "$ZSH"/themes/*.zsh-theme(N:t:r)
   )
+  # Remove blacklisted themes from the list
+  for theme in ${ZSH_THEME_RANDOM_BLACKLIST[@]}; do
+    themes=("${(@)themes:#$theme}")
+  done
 fi
 
 # Choose a theme out of the pool of candidates
 N=${#themes[@]}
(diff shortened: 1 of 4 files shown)


DArtagan:
You are terrific. Thank you for putting this one to bed. 🎉

themes/random.zsh-theme
```diff
@@ -0,0 +1,38 @@
...
+  themes=($ZSH_THEME_RANDOM_CANDIDATES)
+else
+  # Look for themes in $ZSH_CUSTOM and $ZSH and add only the theme name
+  themes=(
+    "$ZSH_CUSTOM"/*.zsh-theme(N:t:r)
+    "$ZSH_CUSTOM"/themes/*.zsh-theme(N:t:r)
+    "$ZSH"/themes/*.zsh-theme(N:t:r)
+  )
+  # Remove blacklisted themes from the list
+  for theme in ${ZSH_THEME_RANDOM_BLACKLIST[@]}; do
```
liyishuai: Is it worth documenting this somewhere?

mcornella: Thanks for the ping, documented in https://github.com/ohmyzsh/ohmyzsh/commit/9ce7de9f16ba894423bc5105c7e959894ca2e751.

PR #8667, opened 2020-02-21
Add Insecure completion-dependent directories detected section to README

gliptak:
## Standards checklist:


mcornella:
There's a very long and informative error message printed when this happens, which already points out the same. It's better to put that in the terminal than expecting the user to go to the README for a solution to their specific problem, isn't it?

gliptak:
I do not recall seeing ```ZSH_DISABLE_COMPFIX=true``` in the informative message.

mcornella:
It is there: https://github.com/ohmyzsh/ohmyzsh/blob/40b013f5f119be27bd2fdece431cf4979193bd25/lib/compfix.zsh#L40-L41

gliptak:
@mcornella thank you. maybe I had an older version

PR #8668, opened 2020-02-22
Simplify installer by removing a `cp` command

cccntu:
## Standards checklist:

## Changes:

- [x] simplify installer by removing a `cp` command

## Other comments:
- It can be further simplified by also removing the following `mv` command, 
but if `sed` is interrupted, it might exit with an incomplete .zshrc, so I decided
not to do so.
...


mcornella:
Thanks!

PR #8669, opened 2020-02-22
Added "doas" plugin to support a "sudo" alternative

NeoTheFox:
## Changes:

- Added "doas" plugin that is just a carbon copy of "sudo" plugin.

## Other comments:

This will help OpenBSD users and Linux users that use doas instead of sudo for privilege escalation. You can also use doas on OSX. There isn't really any new code here or major changes.

### DIFF
--- a/plugins/doas/README.md
+++ b/plugins/doas/README.md
@@ -1,0 +1,57 @@
+# doas
+
+Easily prefix your current or previous commands with `doas` by pressing <kbd>esc</kbd> twice
+
+## Enabling the plugin
+
+1.  Open your `.zshrc` file and add `doas` in the plugins section:
+
+    ```zsh
+    plugins=(
+        # all your enabled plugins
+        doas
+    )
+    ```
+
+2.  Restart your shell or restart your Terminal session:
+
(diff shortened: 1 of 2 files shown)


NeoTheFox:
@dongweiming can you please check this as an author of the original plugin?

mcornella:
The problem with a "rip-off plugin" is that once changes are made to one, the other becomes obsolete. A better alternative is to make the `sudo` plugin configurable to allow setting the `sudo` word and the `sudo -e`/`sudoedit` word.

The replacement for `sudo` would be `doas`, but does `doas` have an "edit" equivalent?

154pinkchairs:
@mcornella, well, you can use doas $EDITOR. The risk of plugin becoming obsolete is much less of an issue than with sudo, because doas is dead simple and will remain like this, that's it's objective. Will try working on a solution though.

carlosala:
@154pinkchairs actually `sudo $EDITOR` does not do exactly the same as `sudo -e`

carlosala:
Let's close this in favour of #11395

PR #8670, opened 2020-02-23
Add support for AWS CLI v2 autocompletion

ArjenSchwarz:
## Standards checklist:

## Changes:

- Check if the autocompletion installed by AWS CLI v2 is present, and if so use that. Otherwise fall back to the original autocompletion code.

## Other comments:

Version 2 of the AWS CLI was released on [February 10](https://aws.amazon.com/blogs/developer/aws-cli-v2-is-now-generally-available/) and comes with its own autocompletion built-in. Per [the README for the v2 branch](https://github.com/aws/aws-cli/blob/v2/README.rst#command-completion), this can be enabled with a wrapper in zsh, or with the `complete` command that ohmyzsh already supports. 

I realise I have a hardcoded path in the check, but both the [Linux](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html) and [macOS](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-mac.html) installation instructions highlight that this is where a symlink to the actual binary will be.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -76,22 +76,27 @@ function _awscli-homebrew-installed() {
   fi
 }
 
-# get aws_zsh_completer.sh location from $PATH
-_aws_zsh_completer_path="$commands[aws_zsh_completer.sh]"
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks for your contribution!

PR #8681, opened 2020-02-26
Add vi-mode support for interactive-cd

wallace11:
## Standards checklist:

## Changes:

- Add vi-mode support for interactive-cd

## Other comments:

Used [sudo.plugin.zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/sudo/sudo.plugin.zsh) as a reference.  
However, decided to skip on `vicmd` since it's probably not the intentional use, nor intuitive for the user.


mcornella:
Good job thanks!

PR #8688, opened 2020-02-27
[arcanist] Allow for easier copy pasting of URLS via functions and add arc diff --preview

AlexeyGy:
## Standards checklist:

## Changes:
 - adds the alias
  - ardcp= 'arc diff --preview'
 - adds the functions
  - ardu: With this, both, `ardu https://arcanist-url.com/<REVISION>`, and `ardu <REVISION>` work.
  - arpa: With this, both, `arpa https://arcanist-url/.com/<REVISION>`, and `arpa <REVISION>` work.

Both accept a parameter which can either be the revision id or the whole URL to the revision.

 - updates the README

### DIFF
--- a/plugins/arcanist/README.md
+++ b/plugins/arcanist/README.md
@@ -36,5 +36,5 @@ E.g., `ardu` accepts `https://arcanist-url.com/<REVISION>` as well as `<REVISION
 
 | Function                  | Command                                  |
 | ------------------------- | ---------------------------------------- |
-| ardu [URL or revision_id] | `arc diff --update` [URL or revision_id] |
-| arpa [URL or revision_id] | `arc patch` [URL or revision_id]         |
+| ardu [URL or revision_id] | `arc diff --update` [revision_id] |
+| arpa [URL or revision_id] | `arc patch` [revision_id]         |
--- a/plugins/arcanist/README.md
+++ b/plugins/arcanist/README.md
@@ -31,10 +31,11 @@ plugins=(... arcanist)
 
 ## Functions
 
-The following functions make copy pasting revision ids from the URL bar of your browser easier as they allow for copy pasting the whole URL.
-E.g., `ardu` accepts `https://arcanist-url.com/<REVISION>` as well as `<REVISION>`.
+The following functions make copy pasting revision ids from the URL bar of your browser
+easier, as they allow for copy pasting the whole URL. For example: `ardu` accepts
+both `https://arcanist-url.com/<REVISION>` as well as `<REVISION>`.
 
-| Function                  | Command                                  |
-| ------------------------- | ---------------------------------------- |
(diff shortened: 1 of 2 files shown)


plugins/arcanist/arcanist.plugin.zsh
```diff
@@ -13,10 +13,26 @@ alias ardc='arc diff --create'
...
-alias arpa='arc patch'
+
+#
+# Functions
+# (sorted alphabetically)
+#
+
+ardu() {
+# With this, both, `ardu https://arcanist-url.com/<REVISION>`, and `ardu <REVISION>` work.
+arc diff --update "$(echo "$1" | awk '{n=split($1,a,"/");print a[n]}')"
```
mcornella: You can use `${(s./.)var}` to split by `/` a variable, then `${var[-1]}` to get the latest element of the var array. All together:
```suggestion
arc diff --update "${${(s./.)1}[-1]}"
```
Don't forget to indent as well

plugins/arcanist/README.md
```diff
@@ -27,3 +28,13 @@ plugins=(... arcanist)
...
+
+## Functions
+
+The following functions make copy pasting revision ids from the URL bar of your browser easier as they allow for copy pasting the whole URL.
+E.g., `ardu` accepts `https://arcanist-url.com/<REVISION>` as well as `<REVISION>`.
+
+| Function                  | Command                                  |
+| ------------------------- | ---------------------------------------- |
+| ardu [URL or revision_id] | `arc diff --update` [URL or revision_id] |
+| arpa [URL or revision_id] | `arc patch` [URL or revision_id]         |
```
mcornella: The command part gets a revision only, no URL.

plugins/arcanist/arcanist.plugin.zsh
AlexeyGy: The suggested update will not work when you just want to provide the revision id. i.e.,

`ardu http://example.com/D33333` will work

`D33333` won't work

I suggest using awk for readability and the above point. 

Another point for not using awk would be that awk may not be available, however, using rip grep, I can see it being used widely, e.g., the ubuntu plugin in line 88, the gradle plugin in line 124 among others.

mcornella: We can make it a two-part algorithm then:
```suggestion
local -a URL=("${(s:/:)1}")
arc diff --update "${URL[-1]}"
```

AlexeyGy: Thanks for the suggestion, this one works great 👍 . I am extracting this into a shared function to prevent code duplication.

Test result
image

plugins/arcanist/README.md
AlexeyGy: ✅

PR #8692, opened 2020-03-01
Add `shell-proxy` plugin

septs:
## Standards checklist:

## Changes:

- add a plugin

## Other comments:

Ability to quickly set proxy for user.


septs:
@mcornella @robbyrussell

larson-carter:
@septs 
Hello thank you for your PR. I however do not know how to give feedback since I have never had to  setup a proxy directly in my shell. Can you explain how this benefits others?

septs:
@larson-carter In Mainland China, it is often affected by GFW
All kinds of software will not work properly

e.q:
not use proxy, clone github repo (Mainland China ONLY), speed very slow
image

larson-carter:
@septs 

Alright that makes sense, however isn't that against policy? Trying to get around the GFW?

septs:
No, No

SukkaW:
> Alright that makes sense, however isn't that against policy? Trying to get around the GFW?

There are already many tools used to get around GFW on the GitHub like known [shadowsocks](https://github.com/shadowsocks/shadowsocks-libev).
IMHO, the Tools here is only used to setup proxy in environment thus neither law nor policy is broke.

septs:
@larson-carter Hope to be merged. Thansk

larson-carter:
@septs 

Okay, I'm fine with this being merged. However I do not have a way of testing to validate that it works. I'm going to mark this as testers needed so when I have about 2 or 3 people validate that it works I'll sign off on it!

bdbai:
image
(Tested in China Mainland)
Before applying proxy, git could not even start cloning. It only started working when the proxy is on.

xiao-po:
3C00591D-B5AB-48AF-8083-E258A6AC63FE

(test in china Mainland)

Because of GFW , i can't open Google search website。After set proxy url and apply proxy ， i can access Google search 。And it's easy to disable or enable， pretty good。

VictorNanka:
Screen Shot 2020-05-19 at 18 03 18

(Tested in China Mainland)
I can connect to google after I enable the proxy. This plugin is really convenient !

mcornella:
This seems to be really useful to others, a couple of notes:
- Why did you choose to write this in Python? The things it does are better suited for shell scripting.
- If you do want it to remain in Python, it should be listed as a requirement in the README.
- There is nothing written about which proxy programs to use, this only redirects calls through the proxy. Can you add a little bit of documentation about that?

SukkaW:
@mcornella 

> Why did you choose to write this in Python? The things it does are better suited for shell scripting.

There is already [a zsh alternative](https://github.com/SukkaW/zsh-proxy). But it seems to be too complicated for a shell to implement this feature.

> There is nothing written about which proxy programs to use, this only redirects calls through the proxy. Can you add a little bit of documentation about that?

The plugin here is to setup environment and configuration, not for running a proxy client. Any proxy software providing socks / http support will be compatible.

septs:
@mcornella 
> If you do want it to remain in Python, it should be listed as a requirement in the README.

"This a pure user-space program, shell-proxy setter, written Python3 and Bash." in README file

PR #8694, opened 2020-03-02
Extract hardcoded $HOME to a configurable variable

### LINKED ISSUE #7332: Allow creation of zcompdump in a custom directory
Hey all 👋

I tried clearing my home directory of all things ZSH
(bar the `.zshrc` of course, and a new `.all-zsh-config` directory)
but a `.zcompdump-James’s MacBook Pro-5.3` file was always present
(and if deleted would be recreated on opening of a new terminal)

As everything else ZSH-related - including the `.oh-my.zsh/` directory - was successfully moved to the `.all-zsh-config` directory by setting `$ZSH` to `~/.all-zsh-config/.oh-my-zsh`, this was confusing.

After a look at the `oh-my-zsh.sh` file it looks like this is because regardless of the set `$ZSH` directory the `ZSH_COMPDUMP` file is always set to point to the `$HOME` directory.

https://github.com/robbyrussell/oh-my-zsh/blob/745b6550a6a8415e2c6df0006aad40549c3bc56a/oh-my-zsh.sh#L65-L68

If this could be changed to reference `$ZSH` it would allow a custom config folder to be created (to de-clutter the home directory) without affecting completion.

``` sh
if [ -z "$ZSH_COMPDUMP" ]; then
  ZSH_COMPDUMP="${ZDOTDIR:-${ZSH}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
fi
```

Things seem to be working for me having replaced it, I'm just not sure whether all plugins will respect the `$ZSH_COMPDUMP` and use the new zcompdump file (or just revert to one in the home directory).

Is this something that would be helpful / worth including?

minterior:
Fixes #7332

## Standards checklist:

## Changes:

The new variable `ZSH_COMPDUMP_DEFAULT_DIR` can now be configured alone, having
$HOME as the default, keeping the same behaviour as until now.
For instance, one could set `ZSH_COMPDUMP_DEFAULT_DIR=${ZSH_CACHE_DIR}` to have
the .zcompdump files in the cache directory of her preference.

## Other comments:

Read further: https://github.com/ohmyzsh/ohmyzsh/issues/7332#issuecomment-593308026

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -58,7 +58,8 @@ fi
 
 # Save the location of the current completion dump file.
 if [ -z "$ZSH_COMPDUMP" ]; then
-  ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
+  ZSH_COMPDUMP_DEFAULT_DIR="${ZSH_COMPDUMP_DEFAULT_DIR:-${HOME}}"
+  ZSH_COMPDUMP="${ZDOTDIR:-${ZSH_COMPDUMP_DEFAULT_DIR}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
 if [[ $ZSH_DISABLE_COMPFIX != true ]]; then


mcornella:
As I said in https://github.com/ohmyzsh/ohmyzsh/issues/7332#issuecomment-624221366, this is not a use case that requires yet another setting. `ZSH_COMPDUMP` can be redefined already, quite easily, so there is no need for this. We'll eventually move the zcompdump file inside `$ZSH_CACHE_DIR` though, so that issue will remain open until then.

PR #8699, opened 2020-03-05
kubectl add alias to decode secret data

hkiang01:
## Standards checklist:

  Mine with inspiration from https://github.com/mveritym/kubedecode

## Changes:
 - Add `kdecsec` alias to `kubectl` plugin to base64 decode kubernetes secrets

  ```
  $ kubectl create secret generic my-secret --from-literal=key1=supersecret --from-literal=key2=topsecret
  $ kdecsec my-secret
  {
    "key1": "supersecret",
    "key2": "topsecret"
  }
  ```
## Other comments:

[Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) have their data encoded in base64. More often than not, when wanting to know what's in a given secret, we want the data.

The `kdecsec` alias is pretty much the same thing as [kubedecode](https://github.com/mveritym/kubedecode), which solves the same problem. Making it part of the `kubectl` plugin is just super convenient. :grin:

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -86,6 +86,7 @@ alias kdecs='
 _kdecs() {
   if [ $(jq --version) != "jq-1.6" ]; then
       echo "kdecs requires jq-1.6 to use the base64d format";
+      unset -f _kdecs
       return 1
   fi
   kubectl get secret "$@" -o json | jq ".data | map_values(. | @base64d )";
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -82,18 +82,18 @@ alias kgsec='kubectl get secret'
 alias kgseca='kubectl get secret --all-namespaces'
 alias kdsec='kubectl describe secret'
 alias kdelsec='kubectl delete secret'
-alias kdecs='
-_kdecs() {
+alias kdecsec='
+_kdecsec() {
   if [ $(jq --version) != "jq-1.6" ]; then
-      echo "kdecs requires jq-1.6 to use the base64d format";
-      unset -f _kdecs
+      echo "kdecsec requires jq-1.6 to use the base64d format";
+      unset -f _kdecsec
       return 1
   fi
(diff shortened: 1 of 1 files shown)


kbakdev (approved):

alemairebe:
I think the following alias would be better/do not depends on jq
alias kgsecp='kubectl get secret -o go-template='\''{{range $k,$v := .data}}{{"### "}}{{$k}}{{"\n"}}{{$v|base64decode}}{{"\n\n"}}{{end}}'\'''
code is from https://kubernetes.io/docs/reference/kubectl/cheatsheet/#viewing-finding-resources and go-template could be edited to match a json if that's what the output should look like, but I prefer plaintext as json may need to escape some characters

hkiang01:
Not depending on jq is definitely a plus, nice!

robbyrussell:
Thanks for this, and sorry for the very long delay in replying. The idea is a good one... `kgsec` gives you the base64 blob and there's nothing in the plugin that decodes it.

One thing needs fixing before I can take it further. The 2023 commit that removed the `jq` dependency left a stray `alias kdecsec='` line above the real alias, so the plugin no longer parses and won't load at all:

```zsh
alias kdecsec='
kdecsec_go_template='{{range $k,$v := .data}}...{{end}}'
alias kdecsec="kubectl get secret -o go-template='$kdecsec_go_template'"
```

Dropping that first line should be all it needs. A README row for the new alias would be good too, since the table there lists all the others.

One heads up: #14048 adds a `kesec` alias in the same secrets block, so there may be a small conflict depending on which lands first.

If you'd rather not pick this back up after this long, that's completely fine... just say so and I'll take it from here.

hkiang01:
Go ahead

robbyrussell:
Thanks, and sorry about the wait. I took you up on "go ahead" and pushed the fix to your branch.

The old version had a stray `alias kdecsec='` line above the template, which left a quote open and stopped the kubectl plugin from loading at all. It's one line now, so there's no temp variable to unset at startup either.

I left the README out of it. That table got reformatted on master since 2020 and any edit to it conflicts, so the doc row will come as a separate commit.

PR #8703, opened 2020-03-05
Plugin for vscodium users

msmafra:
## Standards checklist:

- [ x  ] The code is stable and I have tested it myself, to the best of my abilities.

@mcornella @robbyrussell @dimensi0n

### DIFF
--- a/plugins/atom/README.md
+++ b/plugins/atom/README.md
@@ -1,0 +1,84 @@
+# Atom Editor Zsh shell Plugin
+
+This plugin will define aliases for the few Atom command line options but not for `apm`.
+
+To enable, add it to the `plugins` array inside your zsh resources file in `~/.zshrc`. Look for 'plugins', which is the name of the plugins array and add it to the end of the list. This can help to discover which line is it via your shell's terminal:
+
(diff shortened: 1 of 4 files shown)


mcornella:
The vscodium plugin you add is pretty much the same as the vscode one, it'd be better to modify the vscode plugin to allow for configuring which one to use.

This also adds a plugin for atom (which also looks very similar to the vscode plugin), which should be in a separate PR. `force-push` to the PR branch to delete that or open a new PR.

msmafra:
> The vscodium plugin you add is pretty much the same as the vscode one, it'd be better to modify the vscode plugin to allow for configuring which one to use.
Yes, boath based on VSCode one's idea.
 
> This also adds a plugin for atom (which also looks very similar to the vscode plugin), which should be in a separate PR. `force-push` to the PR branch to delete that or open a new PR.
Not sure how to do that.

mcornella:
> Not sure how to do that.

Start again, use a branch other than master and submit a new PR with just the changes to allow vscodium in the vscode plugin. This is the loading logic, it shouldn't be too difficult to extend it:
https://github.com/ohmyzsh/ohmyzsh/blob/2560f04982dfc2693221c7c0f47d1a71df09a794/plugins/vscode/vscode.plugin.zsh#L6-L12

PR #8716, opened 2020-03-08
rsync: update -z to -zz

soredake:
```
This rsync lacks old-style --compress due to its external zlib.  Try -zz.
Continuing without compression.
```

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/rsync/rsync.plugin.zsh
+++ b/plugins/rsync/rsync.plugin.zsh
@@ -1,4 +1,4 @@
-alias rsync-copy="rsync -avz --progress -h"
-alias rsync-move="rsync -avz --progress -h --remove-source-files"
-alias rsync-update="rsync -avzu --progress -h"
-alias rsync-synchronize="rsync -avzu --delete --progress -h"
+alias rsync-copy="rsync -avzz --progress -h"
+alias rsync-move="rsync -avzz --progress -h --remove-source-files"
+alias rsync-update="rsync -avzzu --progress -h"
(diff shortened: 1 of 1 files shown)


mcornella:
We've had some troubles before when changing flags, so I've done a little research:

[`-zz` was introduced in 3.1.1](https://download.samba.org/pub/rsync/rsync.html):
> Note that this option typically achieves better compression ratios than can be achieved by using a compressing remote shell or a compressing transport because it takes advantage of the implicit information in the matching data blocks that are not explicitly sent over the connection. This matching-data compression comes at a cost of CPU, though, and can be disabled by repeating the -z option, but **only if both sides are at least version 3.1.1.**
>
> Note that if your version of rsync was compiled with an external zlib (instead of the zlib that comes packaged with rsync) then it will not support the old-style compression, only the new-style (repeated-option) compression. **In the future this new-style compression will likely become the default.**

[3.1.1 was released in June 22nd, 2014](https://rsync.samba.org/), therefore this change is a risky one.

If you need it, just add an extra `-z` to your rsync commands, until this is widespread enough so we can change it to `-zz`. You can also do this persistently by changing the rsync aliases in your zshrc file, after OMZ is sourced. For example:
```zsh
alias rsync-copy="$aliases[rsync-copy] -z"
alias rsync-move="$aliases[rsync-move] -z"
alias rsync-update="$aliases[rsync-update] -z"
alias rsync-synchronize="$aliases[rsync-synchronize] -z"
```

PR #8747, opened 2020-03-15
man: Removed everything but the command itself before prepending man

wurza:
## Standards checklist:

## Changes:

- Changed man plugin, so it removes all the arguments form the previous command and prepends man to the command itself without arguments

### DIFF
--- a/plugins/man/man.plugin.zsh
+++ b/plugins/man/man.plugin.zsh
@@ -14,7 +14,8 @@
 
 man-command-line() {
     [[ -z $BUFFER ]] && zle up-history
-    [[ $BUFFER != man\ * ]] && LBUFFER="man $LBUFFER"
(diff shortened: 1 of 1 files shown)


mcornella:
Good job, thanks!

mcornella:
I had a problem now when trying to look for the man page of `git rebase`; if we remove everything the function looks for the man page of `git` instead. Maybe there's another possible solution; what was the problem you tried to solve?

wurza:
It was less of a problem, but more of an inconvenience: man would try to open each argument you wrote before. I didn't thought of git, when coding this patch. Are there more such commands? I think I only know git and the ip command. 

Maybe we could resolve this issue by firing up man beforehand with the first two arguments and checking the return value?

wurza:
I created a new pull request https://github.com/ohmyzsh/ohmyzsh/pull/8798, in which i resolved the issue.

PR #8748, opened 2020-03-16
Add lando plugin

JoshuaBedford:
## Standards checklist:


JoshuaBedford:
Added drush support and synced with master to ensure it is up-to-date.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,161 @@
...
+      return "1"
+    fi
+
+    if [[ "$flag" == "1" ]]; then
+      echo "Could not find $1."
+    fi
+
+  fi
+
+  return $flag
```
mcornella, line 159: This part will only be reached if the code exits from the first if. You can return in there instead.

plugins/lando/README.md
```diff
@@ -0,0 +1,20 @@
...
+#### WARNING:
+- **Note:** This assumes your lando projects are located at `$HOME/Sites` (E.G. /Users/username/Sites).
+- This could conflict with any aliases previously installed. If you have any of the CLIs installed globally (outside of Lando). 
+- The `if` statements *should* account for this by calling the non-lando command for you, but not all have been tested fully.
+
+## USE
+- Clone this project into your `custom/plugins` folder:
+	- `cd ~/.oh-my-zsh/custom/plugins`
+	- via SSH: 		`git clone git@github.com:JoshuaBedford/lando-zsh.git lando`
+	- via HTTPS: 	`git clone https://github.com/JoshuaBedford/lando-zsh.git lando`
```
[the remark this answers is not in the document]
JoshuaBedford: Thanks for the input! That was from early stages before submitting to ohmyzsh. I'll get that updated.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,161 @@
...
+function wp(){
+
+  if checkForFile $CONFIG_FILE $SITES_DIRECTORY ; then
+    # Run Lando wp
+    lando wp "$@"
+  else
+    # Run System wp
+    command wp "$@"
+  fi
+}
```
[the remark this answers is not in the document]
JoshuaBedford: Whoa, thanks for the tip! I don't deal with bash scripts a whole lot. This started as a convenience thing for me so i just cobbled something together at the time. Definitely going to use that.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,161 @@
...
+# $2: The directory to search up to.
+checkForFile(){
+
+  local current_directory="$PWD"
+
+  # Bash is backwards. 0 is true 1 (non-zero) is false.
+  flag="1"
+
+  # Only bother checking for lando within the Sites directory.
+  if [[ ":$PWD:" == *":$2"* ]]; then
```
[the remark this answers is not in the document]
JoshuaBedford: @mcornella The goal there is to check if the current directory is within the sites directory. E.G. Check to see if its a substring. Im not sure that the colons are necessary, and the first star in `*":$2"* is probably unnecessary since the goal is to check the end of the path not the beginning.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,161 @@
...
+      echo "Found it"
+      return 0
+    else
+      echo "Not Found"
+      return "1"
+    fi
+
+    if [[ "$flag" == "1" ]]; then
+      echo "Could not find $1."
+    fi
```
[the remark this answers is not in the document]
JoshuaBedford: I think i had left these in from debugging. Removing alot of the useless stuff.

JoshuaBedford:
> Hi there, thanks for your first contribution! You did a good job, but there are still some changes needed. Let me know if you need more guidance!

I made the initial changes and implemented some of your tips (thanks for those!) Let me know if anything else is needed.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
...
+}
+
+# Check for the file in the current and parent directories.
+checkForFile(){
+
+  local current_directory="$PWD"
+
+  # Bash is backwards. 0 is true 1 (non-zero) is false.
+  # Only bother checking for lando within the Sites directory.
+  if [[ "$PWD" == "$LANDO_ZSH_SITES_DIRECTORY"* ]]; then
```
mcornella, line 28: Ok I tested the conditional and here's how it should be written:
```suggestion
  if [[ "$PWD/" == "$LANDO_ZSH_SITES_DIRECTORY"/* ]]; then
```
This makes sure the current working directory is not `$HOME/Sitesandsomethingelse` but still matches the conditional.

mcornella: One more thing: you can flip the logic of the conditional (`!=`) and make an early return. That way you can flatten the rest of the code and it isn't nested inside an if.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
...
+    while true; do
+      if [ $current_directory != "$LANDO_ZSH_SITES_DIRECTORY" ]; then
+          if [ -f "$current_directory/$LANDO_ZSH_CONFIG_FILE" ]; then
+            echo "Using Lando"
+            return
+          fi
+        current_directory="$(dirname $current_directory)"
+      else
+        break;
+      fi
```
mcornella, line 41: Same thing with the early return. And use `return 1` instead of break here. Also use `[[` instead of `[`.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
...
+
+  local current_directory="$PWD"
+
+  # Bash is backwards. 0 is true 1 (non-zero) is false.
+  # Only bother checking for lando within the Sites directory.
+  if [[ "$PWD" == "$LANDO_ZSH_SITES_DIRECTORY"* ]]; then
+
+    # "Checking for file: $LANDO_ZSH_CONFIG_FILE within $LANDO_ZSH_SITES_DIRECTORY..."
+
+    while true; do
```
mcornella, line 32: This probably should be `[[ "$current_directory" != / ]]` instead of true. Otherwise if code isn't tight there may be an infinite loop.

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
...
+      if [ $current_directory != "$LANDO_ZSH_SITES_DIRECTORY" ]; then
+          if [ -f "$current_directory/$LANDO_ZSH_CONFIG_FILE" ]; then
+            echo "Using Lando"
+            return
+          fi
+        current_directory="$(dirname $current_directory)"
```
mcornella, line 38: Instead of dirname you can use `${current_directory:h}` which gets the dirname (zsh specific).

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
...
+      else
+        break;
+      fi
+    done
+
+    # File not found in loop above
+    # "Could not find $LANDO_ZSH_CONFIG_FILE in the current directory or in any of its parents up to $LANDO_ZSH_SITES_DIRECTORY."
+    return 1;
```
mcornella, line 46: No ending semicolons ;)

plugins/lando/lando.plugin.zsh
```diff
@@ -0,0 +1,54 @@
+LANDO_ZSH_SITES_DIRECTORY="$HOME/Sites"
+
+LANDO_ZSH_CONFIG_FILE=./.lando.yml
```
mcornella, line 3: You can leave out the `./`, just make it the filename.

plugins/lando/README.md
```diff
@@ -0,0 +1,30 @@
...
+| `artisan`  | `lando artisan`  |
+| `composer`  | `lando composer`  |
+| `drush`  | `lando drush`  |
+| `gulp`  | `lando gulp`  |
+| `npm`  | `lando npm`  |
+| `wp`  | `lando wp`  |
+| `yarn`  | `lando yarn`  |
+
+#### How It Works:
+This plugin removes the requirement to type `lando` before a command. It prepends 'lando' for commands executed in the same directory as a .lando.yml config file, but preserves the use of non-lando commands with the same name.
```
mcornella, line 22: This explanation isn't accurate. It will also use lando if a lando.yml file is found in a parent directory, always inside `$LANDO_ZSH_SITES_DIRECTORY`.

plugins/lando/README.md
```diff
@@ -0,0 +1,30 @@
+
+# Lando ZSH (lando-zsh)
```
mcornella, line 2: The header goes in the first line, and always leave an empty newline after it (same for the rest of headers).

plugins/lando/README.md
```diff
@@ -0,0 +1,30 @@
...
+#### ALIASES:
+| Alias | Description |
+|:-:|:-:|
+| `artisan`  | `lando artisan`  |
+| `composer`  | `lando composer`  |
+| `drush`  | `lando drush`  |
+| `gulp`  | `lando gulp`  |
+| `npm`  | `lando npm`  |
+| `wp`  | `lando wp`  |
+| `yarn`  | `lando yarn`  |
```
mcornella, line 19: In rendered Markdown this looks fine, but it's more pleasant for the eyes to format the table so that the lines coincide:
```md
| Alias      | Description      |
|:----------:|:----------------:|
| `artisan`  | `lando artisan`  |
| `composer` | `lando composer` |
| `drush`    | `lando drush`    |
| `gulp`     | `lando gulp`     |
| `npm`      | `lando npm`      |
| `wp`       | `lando wp`       |
| `yarn`     | `lando yarn`     |
```

mcornella (changes requested):
Some more changes needed. Getting there! 👍

plugins/lando/lando.plugin.zsh
mcornella: > One more thing: you can flip the logic of the conditional (`!=`) and make an early return. That way you can flatten the rest of the code and it isn't nested inside an if.

This may not be clear enough. What I mean is:

```
if condition; then
  do something
else
  return 1
fi
```

can be changed to

```
if ! condition; then
  return 1
fi

do something
```

plugins/lando/lando.plugin.zsh
JoshuaBedford: sheesh. I cant tell you how many times ive forgotten a semicolon and now NOT forgetting is the problem! good catch. it didnt throw an error locally.

plugins/lando/README.md
JoshuaBedford: Agreed!

plugins/lando/lando.plugin.zsh
JoshuaBedford: Well... i only half understood the early return. I just swapped it but did the following:
```
if ! condition; then
  return 1
else
do something
fi
```
Which, in my defense does the same thing but is unnecessarily verbose. I'll update that.

plugins/lando/README.md
```diff
@@ -1,30 +1,37 @@
...
-| `composer`  | `lando composer`  |
-| `drush`  | `lando drush`  |
-| `gulp`  | `lando gulp`  |
-| `npm`  | `lando npm`  |
-| `wp`  | `lando wp`  |
-| `yarn`  | `lando yarn`  |
+
+| Alias 	| Description 	   |
+|:---------:|:----------------:|
+| `artisan` | `lando artisan`  |
```
JoshuaBedford, line 15: This was lined up locally perfectly. Thats unfortunate.

mcornella (approved):

mcornella:
I made some light cleanups and now it's ready to be merged. Look it up one last time and let me know if this still works ok.

By the way, is it by designed that `lando` won't be called if in the `$LANDO_ZSH_SITES_DIRECTORY`, i.e. only in subdirectories will it be called?

JoshuaBedford:
> I made some light cleanups and now it's ready to be merged. Look it up one last time and let me know if this still works ok.
> 
> By the way, is it by designed that `lando` won't be called if in the `$LANDO_ZSH_SITES_DIRECTORY`, i.e. only in subdirectories will it be called?

Yes, that is correct. This plugin operates under the assumption that you have a directory (E.G. `~/Sites`) that hosts your various project directories (E.G. `~/Sites/lando-project-1` or `~/Sites/lando-project-2`).

mcornella:
Ok thanks!

JoshuaBedford:
> Ok thanks!

@mcornella Thank you for all of your help and input! I don't do much bash scripting and was able to learn a lot today thanks to you. Kudos!

PR #8754, opened 2020-03-20
gitfast: add mergetool completion

remcohaszing:
## Standards checklist:

## Changes:

- `mergetool` has been added to the `git` completion in the `gitfast` plugin.

## Other comments:

I switched to `gitfast` because I started using `git switch`. `mergetool` was the one completion I really missed and started typing wrong because `git m<tab>` would complete to `git merge `. (Note the trailing space.)

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -144,6 +144,7 @@ __git_zsh_cmd_common ()
 	init:'create an empty Git repository or reinitialize an existing one'
 	log:'show commit logs'
 	merge:'join two or more development histories together'
+	mergetool:'run merge conflict resolution tools to resolve merge conflicts'
 	mv:'move or rename a file, a directory, or a symlink'
 	pull:'fetch from and merge with another repository or a local branch'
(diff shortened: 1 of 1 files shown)


mcornella:
The gitfast plugin uses the [official git completion](https://github.com/git/git/blob/master/contrib/completion/git-completion.zsh), you should direct this change to the upstream project. @felipec

felipec:
@mcornella It uses the official completion, but not quite; I have patches on top of it.

Right now the projects are not properly synchronized (there's pull/8550 pending), we need more communication to decide how to proceed.

mcornella:
Ok let me see what needs to be done to fix everything in the gitfast plugin.

larson-carter:
@felipec I'm reviewing pull 8550, however it needs to be broken into 2 or more different PRs. It looks like you are trying to do so much into one PR.

See my exact reply here: https://github.com/ohmyzsh/ohmyzsh/pull/8550#issuecomment-606371743

PR #8759, opened 2020-03-22
Added macOS plugin

robvadai:
## Standards checklist:

## Changes:

- Added new `macos` plugin to wrap utilities solely usable (and of course helpful) on the macOS platform
- Currently only has one command, that is, to force the cleaning up of purgeable disk space on a selected macOS disk, because macOS does this in an unpredictable way therefore if one needs to free up system caches etc it is not deterministic when it will run

## Other comments:

- The disk clean up command needs to be run as an `administrator`: the plugin does not make a judgement on who is running it; for non-administrator users a password prompt will appear on the screen
...

### DIFF
--- a/plugins/macos/README.md
+++ b/plugins/macos/README.md
@@ -1,0 +1,22 @@
+# macOS plugin
+
+## Introduction
+
+The `macos plugin` adds useful command-line utilities for macOS.
+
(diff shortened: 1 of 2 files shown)


mcornella:
There's a macOS plugin already, only it's named OSX for backwards compatibility.

PR #8762, opened 2020-03-22
Added osx plugin command to clean purgeable disk space

robvadai:
## Standards checklist:

## Changes:

- Added new command to enforce cleaning up purgeable disk space on the selected disk device

## Other comments:

macOS doesn't clean up purgeable disk space in a deterministic fashion. In some cases the user wants to enforce this to kick-off in order to free up disk space. Using this command the user can trigger cleaning up the purgeable disk space that in my experience can easily free up tens of GBs of disk space.


mcornella:
Can you explain more in detail what's considered purgeable disk space and what are the possible side effects of this command?

robvadai:
Hi @mcornella 

purgeable disk is a concept by Apple, more info: https://support.apple.com/en-gb/guide/disk-utility/dskutl1005/mac#:~:text=Amount%20of%20space%20available%2C%20which,them%20as%20space%20is%20required.

I've used this command on my Mac without any side effects, so I'm not aware of any.

Reading the man pages of the `diskutil` command, I think leaving out the `freespace` parameter could erase the actual contents of the disk. However my PR does add the `freespace` parameter.

bartekpacia:
You'll probably have to fix the merge conflicts @robvadai

mcornella:
Ok, one more thing: from what I understood, it actually wipes the free space so that the old contents of that space can't be retrieved, it doesn't actually free up space in the hard drive, right (meaning, there won't be any extra available space for data after this runs)?

robvadai:
I think it fills up the space with NULL bytes which is essentially freeing up space (that is, free disk space is literally made up of NULL bytes). Something like that.

robvadai:
As I said there was more free disk available when I tried it.

53jk1 (approved):

robvadai:
Hi guys, any chance to merge this?

TheTechRobo:
How does it free up already free disk space?

robvadai:
Please check the above link to the Apple docs. All I did was basically adding an alias to an already existing macOS command.

rwmitchell:
Checking: https://discussions.apple.com/thread/250765829  it seems to be of limited, if any, value.  It doesn't work on SSDs and main purpose seems to be to overwrite free space with zeros to make it more secure.  If you saw space being freed, it was more likely due to TimeMachine clearing its work area.

TheTechRobo:
so this command is based for SECURITY, not STORAGE.

robvadai:
I have a Mac with a SSD and it works :)
otherwise I wouldn't have implemented it

TheTechRobo:
Oh, I know what it is.

https://bglug.ca/pipermail/group_bglug.ca/2021-March/002407.html

robvadai:
Could be yes.

In my case, I removed like 30GB worth of cached files, checked Finder, like 1GB less.
Ran above command, voila: I saw the 30GB disk saving.

Given everything relies on the OS' ability to report available space, this trick might come handy for people like me who have very small free disk space left.

mcornella:
Hey Rob, I'm sorry to ask for more but I have two suggestions:

1. Rename to something less hard to remember. Something like `freespace` sounds right to me.
2. Is it possible to offer completion for it, i.e. listing the valid disks you can pass it, so that it's more user-friendly?

robvadai:
Hi @mcornella ,

yes and yes!

Let me update the PR

PR #8765, opened 2020-03-23
Enhance Git performance for the `bureau`theme

laarmen:
## Standards checklist:

## Changes:

This is a partial of the bureau_git_status function to be much less resource-hungry. I'm fairly certain I'm reinventing the wheel at times, but so did the previous version, so no regression on that front. I borrowed from lib/git.zsh on the dirty part, as I couldn't really see how to make the parse_git_dirty function fit with the current features of the theme.

The bad news : this removes whatever support there was for "divergence" with the remote. AFAICT this support wasn't functional as the variable wasn't defined, though.

## Other comments:

I very rarely write shell scripts, and never had written anything for zsh, which means there's  probably room for improvements in there. However, I might not have much time to refine it further, so if you feel like these changes might be valuable but need some tweaks, do feel free to do whatever you feel necessary.

Thank you for your work!

### DIFF
--- a/themes/bureau.zsh-theme
+++ b/themes/bureau.zsh-theme
@@ -26,22 +26,29 @@ bureau_git_status() {
   _STATUS=""
 
   # check status of files
-  _INDEX=$(command git status --porcelain 2> /dev/null)
-  if [[ -n "$_INDEX" ]]; then
-    if $(echo "$_INDEX" | command grep -q '^[AMRD]. '); then
-      _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_STAGED"
+  if [[ "$(command git config --get oh-my-zsh.hide-dirty 2> /dev/null)" != "1" ]]; then
+    local -a FLAGS
+    FLAGS=('--porcelain')
+    if [[ "$DISABLE_UNTRACKED_FILES_DIRTY" == "true" ]]; then
+      FLAGS+="--untracked-files=no"
     fi
-    if $(echo "$_INDEX" | command grep -q '^.[MTD] '); then
-      _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_UNSTAGED"
+    _INDEX=$(command git status $FLAGS 2> /dev/null)
+    if [[ -n "$_INDEX" ]]; then
+      if $(echo "$_INDEX" | command grep -q '^[AMRD]. '); then
+        _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_STAGED"
+      fi
+      if $(echo "$_INDEX" | command grep -q '^.[MTD] '); then
+        _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_UNSTAGED"
+      fi
+      if $(echo "$_INDEX" | command grep -q -E '^\?\? '); then
+        _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_UNTRACKED"
+      fi
+      if $(echo "$_INDEX" | command grep -q '^UU '); then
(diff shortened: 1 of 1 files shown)


themes/bureau.zsh-theme
```diff
@@ -23,44 +23,51 @@ bureau_git_branch () {
...
+      fi
+      if $(echo "$index" | command grep -q -E '^\?\? '); then
+        gstatus="$gstatus$ZSH_THEME_GIT_PROMPT_UNTRACKED"
+      fi
+      if $(echo "$index" | command grep -q '^UU '); then
+        gstatus="$gstatus$ZSH_THEME_GIT_PROMPT_UNMERGED"
+      fi
+    else
+      gstatus="$gstatus$ZSH_THEME_GIT_PROMPT_CLEAN"
     fi
```
Aloxaf: I suggest using `case` here. Avoid forks, it is slow.
```suggestion
    case $index in
      [AMRD]?*) gstatus+=$ZSH_THEME_GIT_PROMPT_STAGED ;;
      ?[MTD]*) gstatus+=$ZSH_THEME_GIT_PROMPT_UNSTAGED ;;
      "?? "*) gstatus+=$ZSH_THEME_GIT_PROMPT_UNTRACKED ;;
      "UU "*) gstatus+=$ZSH_THEME_GIT_PROMPT_UNMERGED ;;
      "") gstatus+=$ZSH_THEME_GIT_PROMPT_CLEAN ;;
    esac
```

themes/bureau.zsh-theme
```diff
@@ -23,44 +23,51 @@ bureau_git_branch () {
...
-  _INDEX=$(command git status --porcelain -b 2> /dev/null)
-  if $(echo "$_INDEX" | command grep -q '^## .*ahead'); then
-    _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_AHEAD"
-  fi
-  if $(echo "$_INDEX" | command grep -q '^## .*behind'); then
-    _STATUS="$_STATUS$ZSH_THEME_GIT_PROMPT_BEHIND"
+  local ref
+  ref=$(command git rev-parse --abbrev-ref HEAD 2> /dev/null)
+  index=$(command git for-each-ref --format="%(upstream:track)" refs/heads/${ref} 2> /dev/null)
+  if $(echo "$index" | command grep -q 'ahead'); then
```
Aloxaf: The same as above. Or you can also use `if [[ ]]` here considering there are only two branches.
```suggestion
  if [[ "$index" == *"ahead"* ]]; then
```

kbakdev (approved):

laarmen:
Hello,

Have I missed something to get these merged, should this be closed as out of scope, or is it "just" a matter of not having the time/energy to getting to it? I'll be fine either way (do some more work, closing, or waiting some more).

Cheers,
Simon

robbyrussell:
Thanks for this, and apologies for the long silence. The work you were after here got done in the meantime, just not through this PR... 43be5ea (Dec 2021) dropped the global variables and 74a3db7 (Feb 2022) removed the grep calls. `bureau_git_status` is now a single `git status --porcelain -b` with no forks, which is the same win you were chasing, so I'm closing this one out.

Two ideas in your patch aren't in master and would still be worth having: using `git for-each-ref` for the upstream check, and honoring `oh-my-zsh.hide-dirty` / `DISABLE_UNTRACKED_FILES_DIRTY`. If you're still interested, those would make a good small PR against the current file.

Sorry you didn't hear back at the time.

laarmen:
Hi,

Sorry, I don't use ZSH anymore so don't have any interest in picking
this back up. Thanks for finally getting to this, though :)

PR #8776, opened 2020-03-28
colorized-man-pages: Added hooks for commands dman and debman

Zocker1999NET:
## Standards checklist:

## Changes:

- Adds hooks for `dman` and `debman` similar to existing hook of `man` to colorize its output using `less`
- README of the plugin also states that `dman` and `debman` are hooked.

## Other comments:

`dman` and `debman` are tools from the package `debian-goodies` for
Debian systems (see https://packages.debian.org/stable/debian-goodies).
They can display man pages from manpages.debian.org or local debian
packages respectively.

I thought about adding a new plugin for these but then code would either
be duplicated or a dependency to this plugin would be there and as I
think non Debian users will not be disturbed by these additions this
solution should be fine. If you see it otherwise, please let me know.

### DIFF
--- a/plugins/colored-man-pages/README.md
+++ b/plugins/colored-man-pages/README.md
@@ -8,6 +8,8 @@ To use it, add `colored-man-pages` to the plugins array in your zshrc file:
 plugins=(... colored-man-pages)
 ```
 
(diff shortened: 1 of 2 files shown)


mcornella:
Thanks!

PR #8780, opened 2020-03-29
Remove redundant information

ploth:


mcornella:
This is not redundant, and keeping it there makes it easier to pull upstream changes.

ploth:
image

PR #8787, opened 2020-03-31
[FEATURE: CHTH] change theme by command line

calvetalex:
The script added in this feature aims to allow the modification of the theme without editing the .zshrc itself.
I know it's not necessarily a useful addition, but I thought it might be interesting to be able to select our theme by using command line.
To be 100% user friendly, I would like to add error handling to avoid choosing a theme that cannot be found. For that I had thought of using the _find_ command and check the result but I would first like to know if it is possible to add autocompletion in order to propose names of themes found in "oh-my-zsh / themes" directory.
chth_look

### DIFF
--- a/chth.sh
+++ b/chth.sh
@@ -6,15 +6,31 @@
 DEFAULT="\033[00m"
 YELLOW="\033[1;33m"
 TEAL="\033[1;36m"
+RED="\033[1;31m"
+GREEN="\033[1;32m"
 
 #
 ## function chth for change theme
 ## change the current ZSH_THEME by the one readed on stdin
-## TODO: add error handling on theme selected || autocompletion for theme choice
+## src:
+## - https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/themes/themes.plugin.zsh
+##           plugin which ables you to change your theme on your current term
+## - man sed
 #
 chth() {
 	echo -e $YELLOW'Current Theme: '$TEAL $ZSH_THEME $DEFAULT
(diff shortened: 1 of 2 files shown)


mcornella:
This is already implemented in the [`themes` plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/themes/README.md), only it doesn't make the `ZSH_THEME` change in the zshrc file.

calvetalex:
> This is already implemented in the [`themes` plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/themes/README.md), only it doesn't make the `ZSH_THEME` change in the zshrc file.

Indeed, I had not seen this plugin before.
I apologize for the inconvenience and I close the pull request

calvetalex:
> This is already implemented in the [`themes` plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/themes/README.md), only it doesn't make the `ZSH_THEME` change in the zshrc file.

The plugin don't change the theme permanently. With this version it's permanent, so you will not have to do it for each terminal. Could we upgrade the plugin ?

calvetalex:
I used the error handling from the plugin (giving it in src for sure). New look is this one: 
chth_look

mcornella:
Hi @calvetalex, `sed -i` doesn't work in all platforms, see 1246c3e08bfb7d999bac2ae3e9933084cffbc4eb. There are other issues with the PR, but just in general, changing the zshrc file is something we're not comfortable doing yet.

PR #8005, opened 2019-07-15
Upgrade correct branch

ngaro:
If oh-my-zsh is installed with "BRANCH=somebranch" then upgrade_oh_my_zsh will try to use the (incorrect) branch "master" instead of "somebranch".
This commit fixes this.

PR #8788, opened 2020-03-31
bugfix: upgrade correct branch when running upgrade_oh_my_zsh

ngaro:
## Standards checklist:

## Changes:

If oh-my-zsh is installed with `BRANCH=somebranch` then `upgrade_oh_my_zsh` tried to use the (incorrect) branch `master` instead of `somebranch`.
This PR fixes this.

## Other comments:

This does the same as PR #8005 which i closed myself because it didn't follow contribution guide


mcornella:
`git symbolic-ref --short HEAD` in `git pull` is redundant, since a `git pull` already uses the current branch if not passed, so it could just be removed to only say `git pull origin`.

ngaro:
You are correct, the effect is the same.

I prefer using my more explicit method because It emphasizes that you are upgrading the current branch without touching any other branches.

ngaro:
Merge conflict is now fixed, (although github doesn't seem to know this)

53jk1 (approved):

wac-:
This seems to have gotten stalled and seems like the conflicts haven't been resolved.

Can I suggest something like https://github.com/ohmyzsh/ohmyzsh/compare/master...wac-:upgrade-stay-on-branch to move things forward?

ngaro:
Your method is indeed better, thanks !
Now let's hope it's gets merged soon before it conflicts again...

ngaro:
I just noticed that the same problem exits when using the new `omz update` (which didn't exist when i created this bugfix). I'll fix it there also _after_ this gets **finally**  merged...

mcornella:
098bcda6910aa2f734fb0ad59ac089157a239c87 adds `oh-my-zsh.remote` and `oh-my-zsh.branch` settings when installed, which are then used by the updater to figure out from where to `git pull`.

If you need to change that, you can do `git -C "$ZSH" config oh-my-zsh.remote <remote-name>` or `git -C "$ZSH" config oh-my-zsh.branch <branch-name>` to change the behavior manually.

PR #8793, opened 2020-04-02
add /usr/bin/ as fzf location

pjvds:
This adds the default installation directory that Ubuntu uses in the default fzf distribution to the list of default locations. This makes this plugin compatible with the default fzf installation for Ubuntu users.

## Standards checklist:

## Changes:

- Adds a known fzf location for the default installation location for modern Ubuntu distributions.


mcornella:
The [ubuntu package](https://packages.ubuntu.com/disco/amd64/fzf/filelist) doesn't have the same structure as in other systems, it should be correctly picked up by the `setup_using_debian_package` function, no?

53jk1 (approved):

PR #8798, opened 2020-04-04
man: look for man page for subcommands

wurza:
## Standards checklist:

## Changes:

Example: git rebase ...
- Execute man beforehand, to determine if command manual for "git rebase"
- If exists: write "man git rebase" into buffer
- Else: write "man git" into buffer


mcornella:
There are many other man pages in my system that use the second and even the third and fourth arguments, even though man only tries the first two (i.e. `man git rebase` uses `git-rebase`, but `man docker container attach` doesn't look `docker-container-attach`). 

So I think your solution mimics what `man` does well, let me fix a few code style things and I'll merge this.

PR #8801, opened 2020-04-05
place default compdump in ZSH_CACHE_DIR instead of ZDOTDIR

holymonson:
## Standards checklist:

## Changes:

_compdump_ should be consided as a cache file.  Don't pollute user's home, and `ZDOTDIR` maybe not writable.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -58,7 +58,7 @@ fi
 
 # Save the location of the current completion dump file.
 if [ -z "$ZSH_COMPDUMP" ]; then
-  ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
+  ZSH_COMPDUMP="${ZSH_CACHE_DIR}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
 if [[ $ZSH_DISABLE_COMPFIX != true ]]; then
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -58,7 +58,7 @@ fi
 
 # Save the location of the current completion dump file.
 if [ -z "$ZSH_COMPDUMP" ]; then
-  ZSH_COMPDUMP="${ZSH_CACHE_DIR}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
+  ZSH_COMPDUMP="${ZSH_CACHE_DIR}/zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
 if [[ $ZSH_DISABLE_COMPFIX != true ]]; then


mcornella (approved):
Generally I approve this, but it needs to be documented and a migration plan needs to be thought. Most docs and issue comments talk of `rm ~/.zcompdump*`, not `rm $ZSH_COMPDUMP`.

holymonson:
There is only one `rm -f ~/.zcompdump` in source code, the others in docs or comments are out of our control.
https://github.com/ohmyzsh/ohmyzsh/blob/62648d71bb05116287206d83181f9daa5a59ba67/plugins/wd/README.md#L98

I suggest move `plugins/zsh_reload/zsh_reload.plugin.zsh` into `lib/` (loaded by default), and guide users using the `src` function (or rename to a better name) to clean up zcompdump file and reload, getting rid of `rm ~/.zcompdump*` or `rm $ZSH_COMPDUMP`.

kbakdev (approved):

robbyrussell:
@mcornella @carlosala ... reviving this one because it keeps coming back, and I think it needs a decision from us rather than another year of waiting.

Sorry for the long silence, @holymonson. This is still open because we never settled the question, not because the change was wrong.

Where we landed last time, on the twin PR #9090:

- Marc's blocker (Jan 2021) was that all our old "just `rm ~/.zcompdump*`" advice would stop working if the file moves.
- Carlo closed it (Apr 2023) because the zsh manual puts the dump in `$ZDOTDIR`, or `$HOME` if that's unset.

Two things have changed since:

- `omz reload` now deletes the dump wherever it lives (`command rm -f $_comp_dumpfile $ZSH_COMPDUMP` in `lib/cli.zsh`), so there's a location-independent answer to give people. I think that retires Marc's blocker.
- `$ZSH_COMPDUMP` is already overridable... we just don't document it anywhere in the repo, so nobody finds it.

And people are still asking: #13406 is open with a handful of thumbs-up from last November.

So the question is narrow: does the manual's `$ZDOTDIR` describe a default we're free to override, or one we should keep? I lean toward moving it to `$ZSH_CACHE_DIR`. The manual states where the file goes by default, not where it has to go, and we already let people point it elsewhere.

Whichever way we go also decides #9074 and #9092, so it'd be good to answer all three at once.

One line from each of you is enough to unblock it.

mcornella:
@robbyrussell 

I agree with #9092 approach, the metadata to check if zcompdump needs to be recreated should be separate from the file. Including the zsh version and host metadata. 

I'd choose a format to store this metadata that is easier to check for and doesn't require forks, which also touches on your previous PR changing the approach to make the check more performant. Additional points if a zsh built-in function is used for this format (check in builtin functions, I think one was store_cache and retrieve_cache). But if nothing suits our specific approach, it's a good idea to "invest" in a custom function that is performant and works well for our use case: storing the metadata / list of variables in some file and checking the stored file for the desired state (boolean approach).

Also moving to ZSH_CACHE_DIR instead of HOME/ZDOTDIR is better, I agree. This solves all 3 PRs.

PR #8807, opened 2020-04-06
lib: enable diff color

holymonson:
## Standards checklist:

## Changes:

enable diff color if possible.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -39,6 +39,11 @@ if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
   fi
 fi
 
+# enable diff color if possible.
+if diff --color . . &>/dev/null; then
+  alias diff='diff --color'
+fi
+
 setopt auto_cd
 setopt multios
 setopt prompt_subst


mcornella:
`diff --color=auto` is already the default in my system, what's your `diff` version and in what platform?

holymonson:
> `diff --color=auto` is already the default in my system, what's your `diff` version and in what platform?

diff (GNU diffutils) 3.7 , in both archlinux and macOS.
Anyway, `diff --color=auto --color`, with an extra color option, seems no harm to you?

mcornella:
> Anyway, `diff --color=auto --color`, with an extra color option, seems no harm to you?

I mean that `diff` (without arguments) already adds color automatically if not piped. Does that not happen for you?

holymonson:
No, the default is no color. As stated in https://www.gnu.org/software/diffutils/manual/html_node/diff-Options.html
>--color [=when]
Specify whether to use color for distinguishing different contexts, like header, added or removed lines. when may be omitted, or one of:
> * none Do not use color at all. **This is the default when no –color option is specified.**
> * auto Use color only if standard output is a terminal.
> * always Always use color.
>
>Specifying --color and no when is equivalent to --color=auto.

mcornella:
You're right, it doesn't color it, I probably ran the wrong command. LGTM

kattjevfel:
Since this was merged my shell takes a full 6 seconds to load, not quite sure how to debug this but reverting the commit fixes it.

holymonson:
@kattjevfel Place your alias after loading ohmyzsh libs? Custom settings are usually placed at the end so you can overwrite anything you want.

mcornella:
We should prefix the diff call with `command` to bypass his alias when checking for the color argument.

It might also be a good idea to cache the check like we do in the grep.zsh file.

PR #8808, opened 2020-04-06
ZSH_CACHE_DIR should respect XDG_CACHE_HOME first

holymonson:
## Standards checklist:

## Changes:

In some distro, i.e. archlinux, oh-my-zsh is packaged system-wide, which causes the whole `ZSH` readonly. Since `XDG_CACHE_HOME` has been defined in _XDG Base Directory Specification_, we should respect it first.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -1,7 +1,7 @@
 # Set ZSH_CACHE_DIR to the path where cache files should be created
 # or else we will use the default cache/
 if [[ -z "$ZSH_CACHE_DIR" ]]; then
-  ZSH_CACHE_DIR="$ZSH/cache"
+  ZSH_CACHE_DIR=${XDG_CACHE_HOME:-"$ZSH/cache"}
 fi
 
 # Migrate .zsh-update file to $ZSH_CACHE_DIR


bartekpacia (approved):
seems valid, thank for the contribution!

mcornella:
This should be `$XDG_CACHE_HOME/ohmyzsh` so as not to put all files outside a folder, and it should downgrade to `$HOME/.cache` if `$XDG_CACHE_HOME` doesn't exist.

holymonson:
`$XDG_CACHE_HOME/ohmyzsh` may not exist so we need to check if it exists first. I tend to use `$XDG_CACHE_HOME`, because nobody cares its mess of files, as long as filenames won't conflict with other applications. I believe `zcompdump-${SHORT_HOST}-${ZSH_VERSION}` is distinct enough.

Downgrading to `$HOME/.cache` will change the default that it's to `$ZSH/cache` now, and we can't assert `$HOME/.cache` existing. Anyway, I'm fine with either.

kbakdev (approved):

ZerdoX-x:
what is status of this?

robbyrussell:

Two things have changed since you opened it. The read-only install case you describe is already handled... `oh-my-zsh.sh` falls back to `${XDG_CACHE_HOME:-$HOME/.cache}/oh-my-zsh` when `$ZSH/cache` isn't writable, which covers the packaged-install situation. And we're not changing the XDG defaults, for the reasons in #9543: `~/.oh-my-zsh` and what lives under it is baked into too many existing installs to move safely. This PR changes exactly that default.

One other thing worth knowing either way: the patch points at the bare `$XDG_CACHE_HOME` rather than a subfolder under it, so it would put our cache files straight into the user's cache root.

If you want the cache somewhere else, `export ZSH_CACHE_DIR=...` before sourcing `oh-my-zsh.sh` does it and is fully supported.

PR #8809, opened 2020-04-06
direnv: check if direnv is installed

mjklemm:
Instead of printing that the direnv hook failed for every command
prompt, this fix changes the plugin to only load the hook if the
'direnv' command is available on the system.

## Standards checklist:

## Changes:

- Extend the direnv plugin with a simple `if` statement to check if the `direnv` command is installed on the system.

## Other comments:

- Testing performed: touch testing on two systems (w/ and w/o direnv)

### DIFF
--- a/plugins/direnv/direnv.plugin.zsh
+++ b/plugins/direnv/direnv.plugin.zsh
@@ -1,15 +1,16 @@
-if type "direnv" > /dev/null; then
-  _direnv_hook() {
-    trap -- '' SIGINT;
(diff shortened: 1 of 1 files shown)


mjklemm:
RFC: Should the direnv plugin print a warning if direnv is not available when the plugin is loaded?  Right now, it silently ignores the missing direnv and simply does not install the proper shell hooks.

mjklemm:
Thanks for improving and merging my patch!

PR #8810, opened 2020-04-06
kubectl: Add persistent volumes (pv) commands

alzubaidi:
## Standards checklist:

## Changes:

- Add commands to get, edit, describe and delete pv

## Other comments:

...


robbyrussell:

The alias names follow the plugin's convention correctly, so this isn't about how it was written. Persistent volumes are cluster-scoped objects that most people using this plugin never touch directly... the plugin already covers PVCs, which is the side of it developers work with day to day. Every alias we add is permanent surface, so the bar is whether many people would reach for it often, and PVs sit below that line.

These work well in `$ZSH_CUSTOM`, which is a supported setup and survives updates.

PR #8816, opened 2020-04-08
aws: fix aws_change_access_key function with awscli v2

### LINKED ISSUE #8815: Since v2 of AWS CLI the default pager may prevent AWS plugin from working
**Describe the bug**
Since the update to version 2 of the AWS CLI all requests are formatted with a pager (e.g. `less`).
This is useful to prevent long responses to clutter the terminal history but makes much more difficult to use some features of the aws plugin.

**To Reproduce**
1. Update your AWS CLI version to 2.*
2. The default pager might already be `less` (check running `aws sts get-caller-identity`), otherwise set it through your `~/.aws/config` file or through `export AWS_PAGER="less"`
3. Install the _aws_ plugin
4. Try running `aws_change_access_key`

**Expected behavior**
All calls should be done without blocking the scripts, some of them have to print the result on screen.

 - OS / Distro: macOS
 - Latest ohmyzsh Update?: Yes
 - ZSH Version: [e.g. 5.6]

**Proposed solution**
I will submit shortly a PR to unset any pager for all aws requests in the scripts.

tom139:
After the upgrade to AWS CLI v2, if the default client side pager is set
the aws_change_access_key stopped working.

Fixes #8815

## Standards checklist:

## Changes:

The default pager is set before any call to the aws cli:
```zsh
# before
aws iam create-access-key

# now
AWS_PAGER="" aws iam create-access-key
```

## Other comments:

This prevents the aws cli to paginate any request, regardless of the default pager of the user.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -31,12 +31,12 @@ function aws_change_access_key() {
 
   echo Insert the credentials when asked.
   asp "$1" || return 1
-  aws iam create-access-key
-  aws configure --profile "$1"
+  AWS_PAGER="" aws iam create-access-key
+  AWS_PAGER="" aws configure --profile "$1"
 
   echo You can now safely delete the old access key running \`aws iam delete-access-key --access-key-id ID\`
   echo Your current keys are:
-  aws iam list-access-keys
+  AWS_PAGER="" aws iam list-access-keys
 }
 
 function aws_profiles() {


tom139:
It would be great if @mcornella and/or @contrun could take a look at this.

contrun:
I just checked aws-cli release history https://pypi.org/project/awscli/#history which shows v2 is currently not released yet. So I have did not check if your problem exists and the proposed solution works. I personally think if your pr is reasonable if that is indeed the problem.

tom139:
Hi, thank you for checking out. The cli v2 is GA and the new default (see: https://aws.amazon.com/blogs/developer/aws-cli-v2-is-now-generally-available/).

PyPI is _behind_ because the cli is no longer installed using pip.

Other common package manager (like brew for mac https://formulae.brew.sh/formula/awscli) already install v2 and up.

Additionally the fix has full backward compatibility.

tom139:
Hi, sorry to bother you @contrun but could you take another look at this issue and the docs about AWS CLI v2 being the current version?

Pagination options are well explained in [this guide](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html).

Thank you 🙏🏽

contrun:
@tom139 I am in no position of merging this PR. I think it is reasonable and should be merged.
Judging from the github status of @mcornella and the current covid situation, @mcornella could be occupied with some other business for a while. Maybe we should ping @larson-carter ?

larson-carter:
Hey @tom139 and @contrun thank you for the ping.

I'm going to put this on hold right now since we are internally working on a plan between me and @mcornella we will review this as soon as possible

mcornella:
Thanks for your patience folks.

PR #8819, opened 2020-04-10
Added a plugin to support the SCCS version control system

dertuxmalwieder:
## Standards checklist:

## Changes:

- Added support for the SCCS version control system: Adds a "dirty" and a "clean" sign.


kbakdev (approved):

robbyrussell:
Thanks for putting this together, and sorry it sat this long without an answer. Six years is not a reasonable wait for a yes or a no.

The answer is no, and it's about reach rather than the code or the age of SCCS. Every plugin we take on is permanent maintenance surface for a small volunteer team, so the test is how many people would enable it. Debian's popularity contest puts `cssc` at 82 installs and zero regular users, and in six years nobody else commented here, opened an issue, or asked for SCCS support any other way. I can't justify carrying it against that.

You'll have noticed we do ship `fossil`, `svn` and `mercurial`. Those predate the current bar rather than clearing it... `mercurial` was last touched in 2013.

It works well as a custom plugin in `$ZSH_CUSTOM/plugins/sccs`, which is a supported setup and survives updates. One small thing if you keep using it: the plugin `export`s `PROMPT` and `RPROMPT`, where the fossil plugin it's based on assigns them plainly. Exporting them can leak the prompt into child processes.

PR #8824, opened 2020-04-12
colorize: add $ZSH_COLORIZE_CHROMA_FORMATTER config env var

johnw188:
In order to get colorize to work with my setup I needed to change the formatter chroma was using to `terminal256`. This updates colorize to look for a new $ZSH_COLORIZE_CHROMA_FORMATTER env var to set this value.

## Changes:

- Check for the existence of `$ZSH_COLORIZE_CHROMA_FORMATTER` and if present add a `--format` flag to the `chroma` call

## Other comments:

Super simple low risk change.

### DIFF
--- a/plugins/colorize/colorize.plugin.zsh
+++ b/plugins/colorize/colorize.plugin.zsh
@@ -46,7 +46,11 @@ colorize_cat() {
         if [[ "$ZSH_COLORIZE_TOOL" == "pygmentize" ]]; then
             pygmentize -O style="$ZSH_COLORIZE_STYLE" -g
         else
-            chroma --style="$ZSH_COLORIZE_STYLE"
+            if [ -z "$ZSH_COLORIZE_CHROMA_FORMATTER" ]; then
+                chroma --style="$ZSH_COLORIZE_STYLE"
+            else
+                chroma --style="$ZSH_COLORIZE_STYLE" --formatter="$ZSH_COLORIZE_CHROMA_FORMATTER"
+            fi
         fi
         return $?
     fi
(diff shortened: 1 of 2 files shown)


plugins/colorize/colorize.plugin.zsh
```diff
@@ -46,7 +46,11 @@ colorize_cat() {
         if [[ "$ZSH_COLORIZE_TOOL" == "pygmentize" ]]; then
             pygmentize -O style="$ZSH_COLORIZE_STYLE" -g
         else
-            chroma --style="$ZSH_COLORIZE_STYLE"
+            if [ -z "$ZSH_COLORIZE_CHROMA_FORMATTER" ]; then
+                chroma --style="$ZSH_COLORIZE_STYLE"
+            else
+                chroma --style="$ZSH_COLORIZE_STYLE" --formatter="$ZSH_COLORIZE_CHROMA_FORMATTER"
+            fi
```
mcornella, line 53: [The default `--formatter` is `terminal`](https://github.com/alecthomas/chroma/blob/477ad4aefdf2501f92e0a2261572bfc2cc2c9cb2/cmd/chroma/main.go#L49). We can use that to simplify the code by always providing a `--formatter` argument:
```suggestion
            chroma --style="$ZSH_COLORIZE_STYLE" --formatter="${ZSH_COLORIZE_CHROMA_FORMATTER:-terminal}"
```
Same as for the line below.

johnw188:
@mcornella Thank you for the review, used your suggestion. Updated the readme as well, let me know if you have other comments but otherwise this should be good to go now.

mcornella:
Great job, John. Thanks!

PR #8834, opened 2020-04-15
 Plugin Git: Add alias to clone git by pasting clipboard contents

Viranchee:
## Changes:

- [x] Git Plugin - Git Clone alias to clone repo from clipboard


mcornella:
This only works on macOS and really the `gcl` alias plus CMD+V is short enough.

PR #8837, opened 2020-04-16
Load asdf completions

filipegiusti:
## Standards checklist:

## Other comments:

May fix https://github.com/asdf-vm/asdf/issues/692

It's a short term solution for https://github.com/ohmyzsh/ohmyzsh/issues/8779

Fix https://github.com/ohmyzsh/ohmyzsh/pull/8820

Tagging listed Maintainer of OMZ asdf plugin: @RobLoach

### DIFF
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -1,17 +1,18 @@
 # Find where asdf should be installed
 ASDF_DIR="${ASDF_DIR:-$HOME/.asdf}"
-ASDF_COMPLETIONS="$ASDF_DIR/completions"
 
 # If not found, check for Homebrew package
 if [[ ! -f "$ASDF_DIR/asdf.sh" ]] && (( $+commands[brew] )); then
    ASDF_DIR="$(brew --prefix asdf)"
-   ASDF_COMPLETIONS="$ASDF_DIR/etc/bash_completion.d"
 fi
 
 # Load command
 if [[ -f "$ASDF_DIR/asdf.sh" ]]; then
     . "$ASDF_DIR/asdf.sh"
 
-    fpath=($ASDF_COMPLETIONS $fpath)
-    autoload -Uz compinit && compinit
+    # If no Homebrew package, load completions
+    if (( ! $+commands[brew] )); then
+        fpath=("$ASDF_DIR/completions" $fpath)
+        autoload -Uz compinit && compinit
+    fi
 fi
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -12,8 +12,6 @@ fi
 if [[ -f "$ASDF_DIR/asdf.sh" ]]; then
     . "$ASDF_DIR/asdf.sh"
 
-    # Load completions
-    if [[ -f "$ASDF_COMPLETIONS/asdf.bash" ]]; then
-        . "$ASDF_COMPLETIONS/asdf.bash"
-    fi
+    fpath=($ASDF_COMPLETIONS $fpath)
+    autoload -Uz compinit && compinit
 fi


filipegiusti:
@jthegedus may also be interested.

jthegedus:
Thanks @filipegiusti I am indeed interested.

plugins/asdf/asdf.plugin.zsh
```diff
@@ -12,8 +12,6 @@ fi
 if [[ -f "$ASDF_DIR/asdf.sh" ]]; then
     . "$ASDF_DIR/asdf.sh"
 
-    # Load completions
-    if [[ -f "$ASDF_COMPLETIONS/asdf.bash" ]]; then
-        . "$ASDF_COMPLETIONS/asdf.bash"
-    fi
+    fpath=($ASDF_COMPLETIONS $fpath)
+    autoload -Uz compinit && compinit
```
jthegedus, line 16: This shouldn't be performed for any brew installed version of `asdf` as Homebrew has it's own setup for completions, which I believe I implemented in https://github.com/Homebrew/homebrew-core/blob/6e728e63ab37243b6832d2995e7efdbac7f0d612/Formula/asdf.rb#L25 (the correctness of my implementation needs to be validated).

jthegedus: I believe #L8 can be removed also as the Homebrew install should now handle the completions.

filipegiusti: I'm not familiar with brew, so I may be completely wrong here.

The documentation suggest that `FPATH=$(brew --prefix)/share/zsh/site-functions:$FPATH` is needed to install all zsh completions when using brew, not just one plugin. ZSH [brew plugin](https://github.com/ohmyzsh/ohmyzsh/blob/7d6f754b6807e312089179e3a4719145c58c7986/plugins/brew/brew.plugin.zsh) also doesn't do anything related to completions.

I think that your changes on brew formulae for asdf should sync link the _asdf file and as users of brew follow the instructions on https://docs.brew.sh/Shell-Completion#configuring-completions-in-zsh, _it just work_ ®.

filipegiusti: I made a few changes to not add the completion code on brew, however I can't test it on brew.

filipegiusti: I have a suspicion that we should also be handling here different versions of asdf. Which asdf version added the zsh completion?

jthegedus: I can perform a test of brew tomorrow, though am also not particularly familiar.

`0.7.6` was the last version this plugin properly supports.

`0.7.8` added the ZSH completions properly.

mcornella:
Running compinit again in the plugin will cause duplicate work and slow down the startup significantly. My suggestion in https://github.com/ohmyzsh/ohmyzsh/issues/8779#issuecomment-605785210 was to change `$fpath`, then `autoload` the completion and `compdef`, something like this:
```zsh
fpath=("$ASDF_DIR/completions" $fpath)
autoload -Uz _asdf
compdef _asdf asdf
```

That said, if the bash completion is present in all present asdf versions, the proper solution is to use it in conjunction with bashcompinit, which is already loaded by OMZ, so https://github.com/asdf-vm/asdf/issues/692 shouldn't be an issue here.

I'm closing this then, in wait of new information.

jthegedus:
@mcornella our goal with this plugin would be to use the ZSH completions and not rely on the bash completions. The confusion in that linked issue is people expecting this to load the ZSH completions with compinit, but getting the bash completions and not having the deps of our bash completions script.

Without the call to compinit in this plugin, would this work without users having to call compinit after they initialise OMZ and it's plugins? My understanding was compinit ran before any plugins :thinking:

Quintasan:
Soooooo this was closed and the only solution for now is to either source asdf completions BEFORE o-m-z without using the plugin or using the bash completions?

mcornella:
[As I said](https://github.com/ohmyzsh/ohmyzsh/pull/8837#issuecomment-624238738), the bash completion (which is perfectly functional) is present in all versions, unlike the zsh one. So it makes sense to keep it that way.

jthegedus:
@Quintasan That's exactly what I do https://github.com/jthegedus/dotfiles/blob/c58a76734caa2d7bbba6acdeadcc439326dde0bb/dotfiles/.zshrc#L71

I agree not ideal.

@mcornella I feel 

>bash completion is present in all versions, unlike the zsh one

Is not a sound reason to avoid solving for ZSH native completions. The fact the Bash works and the `compinit` issues are a cause for headaches is a fair point though.

I hope there's a solution to this in the future, but until then I agree with @mcornella in using the Bash completions if you use this plugin.

nsutelbd1 (approved):

digitalmoksha:
Argh! This should work out of the box, I've already wasted an hour trying to get this to work

gaving:
Came here from https://asdf-vm.com/#/core-manage-asdf

Confused 😵 

I now have:-

```
plugins=(
  asdf
  ....
```

To get things working.

TheTechRobo:
@gaving The way I understand this PR is, you have to specifically load the PR in omz. Its not included by default.

Try `omz pr test #8837` and see how that works. :nod:

jthegedus:
@gaving try what I suggest here - https://github.com/ohmyzsh/ohmyzsh/pull/8837#issuecomment-688615068

The reason the documentation does not just say "Do X" is because there are many different ways to manage the ZSH completions depending on Shell setup. You shouldn't need to run `autoload` etc with this OMZSH plugin as long as you include the Shell completions with `fpath` before the OMZSH plugins are listed.

gaving:
Ah thanks, thats much cleaner!

fxsalazar:
I tried exactly https://github.com/ohmyzsh/ohmyzsh/pull/8837#issuecomment-790980106
```
fpath=(${ASDF_DIR}/completions $fpath)
plugins=(
  asdf
)

source $ZSH/oh-my-zsh.sh
```
and still got the error `.asdf/completions/asdf.bash:68: command not found: complete` even when deleting the `.zcompdump` file. and restarting the terminal. Only solution was adding the autoload/compinit lines.

Version:
```
asdf --version
v0.8.0-c6145d0
```

jthegedus:
@fxsalazar Check if `ASFD_DIR` is populated at that point in your shell setup, I would wager it is not set and therefore the completions dir not being found properly

fxsalazar:
```
fpath=(${ASDF_DIR}/completions $fpath)
plugins=(
  asdf
)
```
```
>> la .asdf/completions
total 40
-rwxr-xr-x  1  staff   6.2K Mar 25 12:42 _asdf
-rw-r--r--  1  staff   2.0K Mar 25 12:42 asdf.bash
-rw-r--r--  1  staff   6.9K Mar 25 12:42 asdf.fish
------------------------------------------------------------
```

Same result.

jthegedus:
> ```
> fpath=(${ASDF_DIR}/completions $fpath)
> plugins=(
>   asdf
> )
> ```
> 
> ```
> >> la .asdf/completions
> total 40
> -rwxr-xr-x  1  staff   6.2K Mar 25 12:42 _asdf
> -rw-r--r--  1  staff   2.0K Mar 25 12:42 asdf.bash
> -rw-r--r--  1  staff   6.9K Mar 25 12:42 asdf.fish
> ------------------------------------------------------------
> ```
> 
> Same result.

What did you do differntly? If `ASDF_DIR` is not set before you use it in the `fpath` code then what you are doing is not going to work. 

Also, you said

>I tried exactly #8837 (comment)

But `fpath=(${ASDF_DIR}/completions $fpath)` != `fpath=($HOME/.asdf/completions $fpath)`. Please try the exact code with `$HOME` and not `$ASDF_DIR` and report your result.

fxsalazar:
``` shell
fpath=($HOME/.asdf/completions $fpath)
plugins=(  
  asdf
)

source $ZSH/oh-my-zsh.sh

# . $HOME/.asdf/asdf.sh
# fpath=(${ASDF_DIR}/completions $fpath)
# autoload -Uz compinit && compinit
```
==>
```
# init terminal
Last login: Sat Apr  3 15:01:41 on ttys005
~/.asdf/completions/asdf.bash:68: command not found: complete

# asdf [tab]
~ » asdf _asdf:60: command not found: compgen   
```

devinrhode2:
@jthegedus do you know if the docs on asdf-vm.com are outdated or not?
CleanShot 2022-12-20 at 15 34 10

CleanShot 2022-12-20 at 15 35 20

jthegedus:
@devinrhode2 The docs were written when this PR was pending, not closed. I am not a ZSH user so not sure if the issue has been resolved elsewhere or not. Happy to update the docs should that caveat no longer be the case.

devinrhode2:
Seems like it's still buggy in zgen

GraemeF:
The completion is working fine for me (oh-my-zsh plugin on macos)

owpac:
Hey guys! 

The completions for ZSH is not yet supported by asdf plugin. I opened a PR [here](https://github.com/ohmyzsh/ohmyzsh/pull/11143) 6 months ago to resolve this. TBH, I'm not sure to perfectly understand all the issues around the ZSH completions problems. Is my PR resolving the issues? Is my PR not relevant?

iloveitaly:
This is fixed in the asdf + direnv plugin https://github.com/redxtech/zsh-asdf-direnv/pull/10

asadakbar:
@carlosala Has this issue been corrected by your changes here https://github.com/ohmyzsh/ohmyzsh/commit/278bcfc93bed91d1095a9ea7981c4a1109da7abd? asdf is still referencing this issue in its zsh and git guide https://asdf-vm.com/guide/getting-started.html#_3-install-asdf.

carlosala:
It should work, yes!

PR #8838, opened 2020-04-16
update ng completion to work with angular-cli 9

famaridon:
## Standards checklist:

## Changes:

- update possible option and aliases
- fix completion of multiple args


mcornella:
How does this fare with older angular-cli versions? Also, this is in a very rough state because it defines lots of variables, was this automatically generated?

famaridon:
Hi I have started from the previous script skeleton and add missing options and remove removed options.

I think this work fine with angular-cli 4+. But the script can suggest options only available on newer version for exemple ```ng add``` whose have been added in angular-cli 6. 

Yes the script is generated from angular-cli help command. You can found the generation script into this github repository : https://github.com/famaridon/ohmyzsh-ng-completion

53jk1 (approved):

PR #8843, opened 2020-04-19
[docker-compose] Added alias for docker-compose up --build

numToStr:
Sometimes building images before starting the containers comes in handy.

Alias: `dcupb`
Command: `docker-compose up --build`


robbyrussell:
Much appreciated @numToStr !

PR #8848, opened 2020-04-21
feat(azure): add `azure` plugin

### LINKED ISSUE #8847: Support for Azure cli
ohmyzsh is missing support for the tool azure-cli
https://docs.microsoft.com/en-us/cli/azure/

A plugin like the one in https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/aws with similar utilities (profiles --> subscriptions)

It looks like people have been living without a plugin for azure for a while, but is that fun ?

jrottenberg:
## Standards checklist:

## Changes:

- add an Azure plugin

### New aliases

* `az_subscriptions`: lists the available subscriptions.
* `azgs`: gets the current value of `$azure_subscription`.
* `azss [<subscription>]`: sets the `$azure_subscription`.

## Other comments:

Fixes  #8847


jrottenberg:
I've worked with my buddy @hagridaaron on this one
It's our first PR against this project, we read https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md but let us know if we missed anything

awasum:
Please squash the commit into one so the git history is clean

awasum:
@jrottenberg , if you have a co-author, you can squash and add the co-author like so: https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

jrottenberg:
absolutely.

53jk1 (approved):

CoaxVex:
Can this be merged? I'd love this feature. Thx!

oooookk7:
Please merge this. Need this. Thank you!

mcornella:
@jrottenberg does `az` provide completion support for `az account set --subscription`? If so, a simple alias would be better.

jrottenberg:
Hello !

There is completion support but it relies heavily on bash : https://github.com/Azure/azure-cli/blob/dev/az.completion

Not sure how to extract / alias it for `az  account set --subscription`

mcornella:
Zsh has `bashcompinit` to read bash completion files, which is automatically called by OMZ already. Regardless, you'd do
```zsh
source /path/to/az.completion
```

That last line looks like it's dependent on how you installed `az`. For example, using `apt`, the file is put in `/etc/bash_completion.d/azure-cli`. Perhaps it would be better to load it from the plugin.

jrottenberg:
Thank you for the feedback, yeah that makes sense :

1.  I'd kill the whole function, since it can be aliased, the only difference is the list now contains the subscriptionId, along with their name, but I'd call that a feature ;-)
1. Not sure to follow you, I'm looking for the most probable location for the plugin, with a fallback to `/etc/bash_completion.d/azure-cli`. You meant to "not try to load the completion altogether" ?

doubleyewdee:
Does this still need testers? I'd be happy to validate some functionality :)

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+  az account list  --all --output tsv --query '[*].name'
+}
+
+function _az_subscriptions() {
+  reply=($(az_subscriptions))
+}
+compctl -K _az_subscriptions azss
+
+# Az prompt
+function prompt_azure() {
```
tmoschou, line 21: Convention seems to be `<plugin>_prompt_info()`
```suggestion
function azure_prompt_info() {
```

And would probably need to add the stub at `lib/prompt_info_functions.zsh`

plugins/azure/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+NOTE : because azure keeps the state of ative subscription in ${AZURE_CONFIG_DIR:-$HOME/.azure/azureProfile.json}, the prompt command requires `jq` to be enabled. If jq is not int the path the prompt will show nothing
+
+## Theme
+
+The plugin creates an `azure_prompt_info` function that you can use in your theme, which displays
+the current `$azure_subscription`. It uses two variables to control how that is shown:
+
+- ZSH_THEME_azure_PREFIX: sets the prefix of the azure_subscription. Defaults to `<az:`.
```
tmoschou, line 37:
```suggestion
- ZSH_THEME_AZURE_PREFIX: sets the prefix of the azure_subscription. Defaults to `<az:`.
```

plugins/azure/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+- ZSH_THEME_azure_SUFFIX: sets the suffix of the azure_subscription. Defaults to `>`.
```
tmoschou, line 39:
```suggestion
- ZSH_THEME_AZURE_SUFFIX: sets the suffix of the azure_subscription. Defaults to `>`.
```

plugins/azure/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+## Plugin options
+
+* Set `SHOW_AZURE_PROMPT=false` in your zshrc file if you want to prevent the plugin from modifying your RPROMPT.
+  Some themes might overwrite the value of RPROMPT instead of appending to it, so they'll need to be fixed to
+  see the azure subscription prompt.
+
+
+NOTE : because azure keeps the state of ative subscription in ${AZURE_CONFIG_DIR:-$HOME/.azure/azureProfile.json}, the prompt command requires `jq` to be enabled. If jq is not int the path the prompt will show nothing
```
tmoschou, line 30:
```suggestion
NOTE : because azure keeps the state of ative subscription in ${AZURE_CONFIG_DIR:-$HOME/.azure/azureProfile.json}, the prompt command requires `jq` to be enabled. If jq is not in the path the prompt will show nothing
```

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+  [[ ! -f "${AZURE_CONFIG_DIR:-$HOME/.azure/accessTokens.json}" ]] && return
+  # azgs is too expansive, if we have jq, we enable the prompt
```
tmoschou, line 23:
```suggestion
  # azgs is too expensive, if we have jq, we enable the prompt
```

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+  [[ ! -f $(which jq) ]]  &&  return
```
tmoschou, line 24: I think this is preferred?

```suggestion
  (( $+commands[jq] )) || return 1
```

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+  azgs=$(cat ${AZURE_CONFIG_DIR:-$HOME/.azure/azureProfile.json} | jq -r '.subscriptions[] | select(.isDefault==true) .name')
```
tmoschou, line 25: This should avoid an additional subprocess
```suggestion
  azgs=$(jq -r '.subscriptions[] | select(.isDefault==true) .name' ${AZURE_CONFIG_DIR:-$HOME/.azure/azureProfile.json})
```

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+  echo "${ZSH_THEME_AZURE_PREFIX:=<az:}$(azgs)${ZSH_THEME_AZURE_SUFFIX:=>}"
```
tmoschou, line 26:
```suggestion
  echo "${ZSH_THEME_AZURE_PREFIX:=<az:}${azgs}${ZSH_THEME_AZURE_SUFFIX:=>}"
```

tmoschou (commented):
Thanks for raising a PR. I'm not a contributor, but a user also interested in an azure plugin. Here are some comments

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+}
+
+if [ "$SHOW_AZ_PROMPT" != false ]; then
+  RPROMPT='$(prompt_azure)'"$RPROMPT"
```
tmoschou, line 30: As a side, perhaps a question for the ohmyzsh maintainers, should new plugins generally be allowed to munge the prompt? It seems only three plugins actually do (`aws`, `fossil`, and `git-prompt` - though the last is specifically a plugin for the prompt). The rest rather document the `*_prompt_info()` and leave it up to the user and/or themes to customise. For instance [plugins/jenv/README.md](https://github.com/ohmyzsh/ohmyzsh/tree/0ab87c26c17171ae6162ff379a0c704fa57dff2e/plugins/jenv#theme-example), [plugins/virtualenv/README.md](https://github.com/ohmyzsh/ohmyzsh/tree/0ab87c26c17171ae6162ff379a0c704fa57dff2e/plugins/virtualenv#virtualenv)

Perhaps that munging behaviour is legacy in those plugins, and not removed for compatibility reasons?

jrottenberg:
Thank you very much @tmoschou ! I didn't know you could push patch to the branch like that in Github, super nice. All look correct, but I want to test some locally before to commit.

plugins/azure/azure.plugin.zsh
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function _az_subscriptions() {
+  reply=($(az_subscriptions))
+}
+compctl -K _az_subscriptions azss
+
+# Az prompt
+function prompt_azure() {
+  [[ ! -f "${AZURE_CONFIG_DIR:-$HOME/.azure/accessTokens.json}" ]] && return
```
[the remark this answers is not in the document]
jrottenberg: Very interesting, we need a file check to make sure the user has a valid session, at the same time the fallback is not too bad:

```
$ az account list  --all --output tsv --query '[*].name'
Please run "az login" to access your accounts.
```

I'll add a 2>/dev/null

plugins/azure/azure.plugin.zsh
jrottenberg: Yeah unless I'm missing something, I don't see a canonical way to update a prompt, I"m using POWERLEVEL9K (damn deprecated too, needs 10k)

jrottenberg: oh wow, yeah it seems to be a theme decision, the powerlevel10k integration is pretty nice (loads the current subscriptions in the right prompt only if I type `az`)

I'll keep `azgs` as variable to use regardless in a RPROMPT, but that simplifies a lot.

plugins/azure/azure.plugin.zsh
jrottenberg: Absolutely, the plugin won't make assumption anymore on where it will show up, it will just load aliases, prepare functions and delegate the prompt to PROMPT or the theme explicitly.

sdavis3:
Can we please get this merged into the master branch?

Etienne-Schmitt:
Is there any news about this?

carlosala:
Thanks everyone for the contributions, and sorry for being so late!

PR #8859, opened 2020-04-26
update alias

P-Shakibafar:
Adding a number of shortcuts that I think are very useful in a project.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/laravel/laravel.plugin.zsh
+++ b/plugins/laravel/laravel.plugin.zsh
@@ -1,5 +1,6 @@
 #!zsh
 alias artisan='php artisan'
+#alias pa='php artisan'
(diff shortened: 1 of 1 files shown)


plugins/laravel/laravel.plugin.zsh
```diff
@@ -24,6 +25,20 @@ alias pamj='php artisan make:job'
...
+
+# Routes
+alias parl="php artisan route:list"
+
+# UI
+alias paua="php artisan ui:auth"
+alias paua="php artisan ui:controllers"
+
+# Test 
+alias test="php artisan test"
```
mcornella, line 39: This will break many scripts as `test` is already a command.

PR #8860, opened 2020-04-26
Sort completed refs by most recent committerdate

thomaseizinger:
Resolves https://github.com/ohmyzsh/ohmyzsh/issues/8767.

## Standards checklist:

## Changes:

Sort refs for completion by most recent committer date.

## Other comments:

This behaviour is currently not configurable, mainly due to my limited knowledge around writing zsh completion scripts.

I noticed that with the current design, the `__git_refs` function calls `git for-each-ref` three times. This is suboptimal because it means that I had to make the same change three times.
Also, the refs returned by each of those three invocations are concatenated and only sorted by committerdate internally but not overall.

I can't fix this myself but maybe someone can refactor this function to only call `git for-each-ref` once and just assign the arguments correctly based on the parameters to the function.

### DIFF
--- a/plugins/gitfast/_git
+++ b/plugins/gitfast/_git
@@ -73,7 +73,13 @@ __gitcomp_direct ()
 
 	local IFS=$'\n'
 	compset -P '*[=:]'
-	compadd -Q -- ${=1} && _ret=0
+
+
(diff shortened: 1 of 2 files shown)


mcornella:
See https://github.com/ohmyzsh/ohmyzsh/issues/8767#issuecomment-625835287

PR #8862, opened 2020-04-27
-Plugin: current-time-with-zone.

thiensubs:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...


53jk1 (approved):

mcornella:
Hi @thiensubs, thank you for the submission.

At this point we're not going to go with this plugin, as we have a lot of them already and we need to consider the priority of the other ones pending.

I'd like to provide some feedback though, just in case you find it useful:
- The plugin hardcodes a list of many timezones and then hardcodes the time offset (`GMT +HH:MM`) on the output with relation to the variables created.
- All the variables created are not `local`, so every time this runs the variables will pollute the global namespace (i.e. after execution `$GMT_` variables will be defined and visible).

You could use an associative array to relate the timezone string (`Pacific/Midway`) to its time offset (`GMT -11`), and then a `for` loop to print all of them in a few lines of code.

Thank you for the contribution though. Wish you the best in your next endeavors!

PR #8866, opened 2020-04-28
Add alias git checkout master && git pull updates

sb-dev:
Add alias `git checkout master && git pull`

## Standards checklist:

## Changes:

- Added new alias to git plugin

## Other comments:

This change is pretty awesome 🎉


kbakdev (approved):

robbyrussell:

The plugin moved under this one. `gcm` has meant `git checkout $(git_main_branch)` since #9049 landed in July 2020, a couple of months after you opened this... it resolves main or master at runtime, so the hardcoded version can't come back. In the diff as it stands your `gcm` sits directly above that line and would be overwritten by it anyway.

For `gcml`, `gcm` followed by `gl` does the same thing, and we keep the alias list to single commands rather than chains where we can.

Closing for that reason. I'm closing #8888 at the same time so you get both answers together rather than another wait... the reasoning there is different and it's written up on that PR. Thanks for both of them, and sorry again that they sat this long.

PR #8884, opened 2020-05-04
feat(1password): add plugin for 1password

r-medina:
## Standards checklist:

## Changes:

- adds a [1password](https://1password.com/) plugin

### DIFF
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -1,0 +1,33 @@
+# opswd puts the password of the named service into the clipboard. If there's a
+# one time password, it will be copied into the clipboard after 5 seconds. The
+# clipboard is cleared after another 10 seconds.
+function opswd() {
+    if [[ $# < 1 ]]; then
+	echo "Usage: opswd <service>"
+	return 1
+    fi
+
+    if ! type pbcopy; then
+	if ! type xclip; then
+	    echo "must have pbcopy or xclip"
+	    return 1
+	else
+	    alias pbcopy='xclip -selection clipboard'
+	fi
+    fi
+
+    local service=$1
+
+    (( ! oploggedin )) && opsignin
+
+    op get item $service \
+	| jq -r '.details.fields[] | select(.designation=="password").value'\
+	| pbcopy
+
+    ( sleep 5 && op get totp $service | pbcopy
+      sleep 10 && pbcopy < /dev/null 2>&1 & ) &!
+}
+
+function oploggedin() {
+    op list users &> /dev/null
+}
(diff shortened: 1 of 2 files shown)


r-medina:
just realized that `pbcopy` will only work on os x. going to update when i have time later this week

bartekpacia:
Any update @r-medina?

plugins/1password/1password.plugin.zsh
```diff
@@ -0,0 +1,33 @@
+# opswd puts the password of the named service into the clipboard. If there's a
+# one time password, it will be copied into the clipboard after 5 seconds. The
+# clipboard is cleared after another 10 seconds.
+function opswd() {
```
ianchesal, line 4: Was about to write this plugin when I saw you had it here already. One change that would be good would be to add completion loading into the plugin. So, at the top of the file:

```
if type op; then
     eval "$(op completion zsh)"; compdef _op op
fi
```

ianchesal:
I have left a comment for adding in `op` completion into the plugin. Otherwise this looks excellent and would love to see it get merged.

ianchesal:
Actually, I think all your `if type...` calls need stderr redirect to stdout and stdout dumped to `/dev/null` otherwise you end up with lots messages like:

```
op is /usr/local/bin/op
```

in your terminal with you use this plugin, right?

plugins/1password/1password.plugin.zsh
```diff
@@ -0,0 +1,33 @@
...
+    if [[ $# < 1 ]]; then
+	echo "Usage: opswd <service>"
+	return 1
+    fi
+
+    if ! type pbcopy; then
```
ianchesal, line 10: Should be:

```
if ! type pbcopy  >/dev/null 2>&1; then
```

plugins/1password/1password.plugin.zsh
```diff
@@ -0,0 +1,33 @@
...
+	if ! type xclip; then
```
ianchesal, line 11: Should be:

```
if ! type xclip  >/dev/null 2>&1; then
```

53jk1 (approved):

bartekpacia:
I'd love to see it merged!

plugins/1password/1password.plugin.zsh
```diff
@@ -0,0 +1,36 @@
...
+
+  (
+    sleep 5 && op get totp "$service" 2>/dev/null | clipcopy 2>/dev/null
+    sleep 10 && clipcopy < /dev/null 2>/dev/null &
+  ) &!
+}
+
+function _opswd() {
+  local -a services
+  services=("${(@f)$(op list items --categories Login 2>/dev/null | jq -r '.[].overview.title')}")
```
mcornella, line 32: The `jq` call could also be removed with
```sh
op list items --categories Login 2>/dev/null | op get item - --fields title 2>/dev/null
```

mcornella (changes requested):
I made code style changes and added a completion function for `opswd`, but there are other possible improvements.

PR #8888, opened 2020-05-05
Add alias to assign execute permissions

sb-dev:
## Standards checklist:

## Changes:

- Add alias to assign execute permissions

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -158,6 +158,7 @@ alias git-svn-dcommit-push='git svn dcommit && git push github master:svntrunk'
 
 alias gk='\gitk --all --branches'
 alias gke='\gitk --all $(git log -g --pretty=%h)'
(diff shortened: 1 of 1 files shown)


kbakdev (approved):

sb-dev:
Any thoughts @mcornella ?

robbyrussell:
Thanks for this, and sorry for the very long delay in replying... and sorry that the ping to Marc in 2021 went unanswered. That one is on us.

Three things put this below the bar. `git update-index --chmod=+x` is a situational command rather than an everyday one. The `+` breaks the naming pattern that the other ~197 aliases in the plugin follow, where the letters after `g` are the initials of the subcommand. And the git plugin is enabled by default in the zshrc template, so anything added here lands in a very large number of shells that never asked for it.

It shadows nothing, so it works well in `$ZSH_CUSTOM` if you still use it.

PR #8897, opened 2020-05-07
Provide completion for DDEV

### LINKED ISSUE #7589: ddev plugin?
How would you feel about a ddev plugin for oh-my-zsh? 

So that command auto-completion works? 

I would be happy to provide the plugin.

felmab:
Fixes #7589

## Standards checklist:

## Changes:

- added new plugin with completion file and README.

## Other comments:

See [more general issue about ZSH completion](https://github.com/drud/ddev/issues/327).

### DIFF
--- a/plugins/ddev/README.md
+++ b/plugins/ddev/README.md
@@ -1,0 +1,12 @@
+# DDEV plugin
+
+This plugin adds auto-completion for [DDEV](https://www.ddev.com/), a
+container-based development solution.
+
+To use it add `ddev` to the plugins array in your zshrc file.
+
+```zsh
+plugins=(... ddev)
+```
+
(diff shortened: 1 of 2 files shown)


tobiasquadflieg (approved):

robbyrussell:
Thanks for putting this together, and sorry it took six years to get you an answer.

The problem is that this is a static copy of ddev's command set as it stood in 2020, rather than something that regenerates. ddev has grown a lot since... `add-on`, `dotenv`, `push`, `mutagen` and others aren't in here, and `restore-snapshot` became `snapshot restore`. A hand-maintained completion file for a fast-moving CLI goes stale quietly, which is worse than having none.

ddev also generates and ships its own zsh completion now, installed automatically by Homebrew and apt, and their docs have a section specifically for Oh My Zsh users:
https://ddev.readthedocs.io/en/stable/users/install/shell-completion/

PR #8924, opened 2020-05-14
Create go_prompt_info to add go version to prompts

grantstephens:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/lib/golang.zsh
+++ b/lib/golang.zsh
@@ -1,0 +1,7 @@
+# get the golang version
+function go_prompt_info() {
+  local go_prompt
+  go_prompt=$(go version | awk '{ print substr($3, 3) }')
+  [[ "${go_prompt}x" == "x" ]] && return
+  echo "${ZSH_THEME_GO_PROMPT_PREFIX}${go_prompt}${ZSH_THEME_GO_PROMPT_SUFFIX}"
+}
--- a/lib/golang.zsh
+++ b/lib/golang.zsh
@@ -1,7 +1,0 @@
-# get the golang version
-function go_prompt_info() {
-  local go_prompt
-  go_prompt=$(go version | awk '{ print substr($3, 3) }')
-  [[ "${go_prompt}x" == "x" ]] && return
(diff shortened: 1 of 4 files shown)


kbakdev (approved):

grantstephens:
@53jk1 can this be merged?

53jk1:
I'm not allowed

robbyrussell:
Thanks for this, and sorry for the long silence... and sorry that when you followed up in 2022 the only reply you got was from someone who couldn't act on it.

We'd take this in `plugins/golang/golang.plugin.zsh` rather than `lib/`. Everything in `lib/` is sourced for every user on every shell start, whether or not they write Go, and a prompt helper that only Go developers use belongs with the rest of the Go tooling.

Three things it needs alongside the move:

- a `(( $+commands[go] ))` guard, so people without Go don't spawn a process and get a stderr error on every prompt render
- the output quoted as `${go_prompt:gs/%/%%}`, the same retrofit the rvm and nvm helpers got, so a `%` in the version can't be interpreted as a prompt escape
- a matching stub in `lib/prompt_info_functions.zsh`, which is how the other `*_prompt_info` helpers are declared for themes that call them when the plugin isn't loaded

Would you be up for that? If you'd rather not after this long, say so, or say nothing, and I'll do it with you credited.

grantstephens:
No worries about the silence, thanks for picking it back up!

Pushed the requested changes:

- Moved `go_prompt_info` from `lib/golang.zsh` into `plugins/golang/golang.plugin.zsh`
- Added a `(( $+commands[go] ))` guard so users without Go installed don't spawn a process
- Quoted the output as `${go_prompt:gs/%/%%}`, matching the rvm/nvm retrofit
- Added `go_prompt_info` to the dummy stub list in `lib/prompt_info_functions.zsh`

Also rebased onto current master since this branch was quite a few commits behind. Ready for another look.

robbyrussell (approved):
Thanks for turning this around so quickly, and for the clean rebase. Ran it here: the guard, the stub, and the percent escaping all behave as expected, and it costs under 10ms per prompt on my machine. Merging.

PR #8930, opened 2020-05-15
agnoster: better respect hide-dirty git config by plainly coloring prompt and not running checks for changes

eliblock:
## Standards checklist:

## Changes:

- In the git prompt: do not re-check entire working tree for changes if `parse_get_dirty` indicates a clean working tree. This will avoid re-processing clean working trees, and ensure that we respect `oh-my-zsh.hide-dirty` to avoid slow loading on large repos.
- In the git prompt: color the prompt segment white if `oh-my-zsh.hide-dirty` is set. Previously the prompt would be colored green (indicating a clean tree) whenever this flag is set, even if the tree is dirty. This new color indicates the lack of status, rather than misleadingly indicating a clean tree.

## Other comments:

* Similar to #6926, but with clearer prompt color and improved fallback behavior for staged changes.
* Relatedly, #7813 may be closed since it appears to duplicate a check of the git config

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -112,6 +112,8 @@ prompt_git() {
     ref=$(git symbolic-ref HEAD 2> /dev/null) || ref="➦ $(git rev-parse --short HEAD 2> /dev/null)"
     if [[ -n $dirty ]]; then
       prompt_segment yellow black
+    elif [[ "$(git config --get oh-my-zsh.hide-dirty 2>/dev/null)" = 1 ]]; then
+      prompt_segment white black
     else
       prompt_segment green $CURRENT_FG
     fi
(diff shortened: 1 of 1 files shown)


kbakdev (approved):

robbyrussell:
Thanks for this, and sorry for the very long delay in replying. You were right, and more right than the title suggests.

We'd been telling people that `oh-my-zsh.hide-dirty` was enough to stop agnoster stalling on big repos. It isn't. `parse_git_dirty` honors it and skips its `git status`, but `prompt_git` then sets `check-for-changes true` unconditionally and calls `vcs_info`, and zsh's git backend runs `git diff --no-ext-diff --ignore-submodules=dirty --quiet --exit-code` plus a `diff-index --cached` on every render regardless. Your second hunk is what actually fixes that, and it helps more people than just `hide-dirty` users... when `parse_git_dirty` already came back clean, the full worktree diff has nothing to find, so skipping it is free for everyone.

Would you be up for rebasing just that half? The `zstyle` lines are unchanged in master, so it should apply nearly as-is.

The colour half is the part we'd rather leave out. It changes visible output, and #10897 did something similar and had to be reverted in #10927.

If you'd rather not pick this back up after six years, that's completely fine... say so, or say nothing, and I'll reapply the perf half myself with you credited in the commit.

PR #8931, opened 2020-05-15
pipenv: fix error when Pipfile is not a file

rmax:
## Standards checklist:

## Changes:

- `Pipfile` check from `-a` to `-f`.

## Other comments:

This fixes a bug when `Pipfile` is a directory which cuases `pipenv` to
crash.

To reproduce it:

* Create a directory `Pipfile`.
* Enter the directory.

You are going to see an output similar to this:

```
  File "/Users/rmax/.pyenv/versions/3.8.2/bin/pipenv", line 10, in <module>
    sys.exit(cli())
  ...
  File "/Users/rmax/.pyenv/versions/3.8.2/lib/python3.8/site-packages/pipenv/project.py", line 677, in touch_pipfile
    with open("Pipfile", "a"):
IsADirectoryError: [Errno 21] Is a directory: 'Pipfile'
```

The current code does a `-a` check which is true for any file type,
including directories. A better way is to use `-f` to ensure `Pipfile`
is a regular file.

### DIFF
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -7,7 +7,7 @@ compdef _pipenv pipenv
 # Automatic pipenv shell activation/deactivation
 _togglePipenvShell() {
   # deactivate shell if Pipfile doesn't exist and not in a subdir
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

PR #8937, opened 2020-05-17
builtin cd overrides user aliases and functions

MenkeTechnologies:
## Standards checklist:

## Changes:

builtin cd to override user alias and functions in system OMZ functions. Important when capturing output and writing to zcompdump

~/oh-my-zsh/oh-my-zsh.sh: 66 #omz revision: $(builtin cd -q "$ZSH"; git rev-parse HEAD 2>/dev/null)## Other comments:

## Other comments:

tested

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -103,7 +103,7 @@ EOF
 function _omz::pr::clean {
     (
         set -e
-        cd -q "$ZSH"
+        builtin cd -q "$ZSH"
 
         _omz::log info "removing all Oh My Zsh Pull Request branches..."
         command git branch --list 'ohmyzsh/pull-*' | while read branch; do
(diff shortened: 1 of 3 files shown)


tools/upgrade.sh
```diff
@@ -17,7 +17,7 @@ else
   RESET=""
 fi
 
-cd "$ZSH"
+builtin cd "$ZSH"
```
mcornella, line 20: `upgrade.sh` is ran with `sh`, `builtin` doesn't exist here.

MenkeTechnologies: ok, i will remove this one.

MenkeTechnologies: done

mcornella:
Thanks!

PR #8944, opened 2020-05-18
Handle unset variables in various parts of the codebase

patmoore:
DISABLE_UNTRACKED_FILES_DIRTY, DISABLE_AUTO_TITLE, GIT_STATUS_IGNORE_SUBMODULES are not set

Handle these variables not being set with conditional access.
    
If the user has set -u option to report attempts to use undeclared / unassigned variable, accessing the variables needs to be conditional.

## Standards checklist:

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -14,10 +14,10 @@ function parse_git_dirty() {
   local -a FLAGS
   FLAGS=('--porcelain')
   if [[ "$(command git config --get oh-my-zsh.hide-dirty)" != "1" ]]; then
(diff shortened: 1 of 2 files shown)


bartekpacia:
Thank you for this PR. To me, it looks good.

@larson-carter wdyt?

larson-carter:
@bartekpacia I can see a use for it. I added the testers needed label so we can get a little bit of feedback!

mcornella:
These are OK changes since they use common zsh syntax. Thanks for your contribution Patrick!

PR #8945, opened 2020-05-18
Fix/emacs from terminal

PeterMosmans:
## Standards checklist:

## Changes:

- This PR adds emacs plugin support for non-X users. This allows one to start emacs always with the `-nw` flag.
The change is backwards compatible. Onlyl if the variable `ZSH_EMACS_NOX` is defined and set to `true`, the new option will be enforced.

### DIFF
--- a/plugins/emacs/README.md
+++ b/plugins/emacs/README.md
@@ -15,6 +15,9 @@ To use it, add emacs to the plugins array in your zshrc file:
 plugins=(... emacs)
 ```
 
+For non-X users, set the variable `ZSH_EMACS_NOX` to `true`. This will ensure
+that Emacs is started with the `-nw` flag (no windows).
+
 ## Aliases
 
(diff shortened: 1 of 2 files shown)


PeterMosmans:
Hi maintainers, any news on this? Can I do anything else at this point?

PeterMosmans:
Ping - anything that can be done in order to get it merged?

53jk1 (approved):

mcornella:
You can use `te` which opens `emacs` with `-nw`.

PeterMosmans:
Hi @mcornella , thanks - I know you can start emacs with the `-nw` flag using `te`. However, this is about aliasing the `emacs` command itself in a standardized way: This way, one can always use emacs, as it will default to the proper variant (either using X or just terminal). Please reconsider this PR, thanks

PR #8947, opened 2020-05-19
Fix a bug in upgrade.sh where it fails to reset terminal color

romkatv:
The bug was introduced in https://github.com/ohmyzsh/ohmyzsh/commit/27f4e079329dabf8e7008b98c8bddf443116eb54#diff-6d2fa595805de6f141a64f905f4db641. This commit has renamed `NORMAL` to `RESET` everywhere except one place.


romkatv:
Thanks for the quick merge!

PR #8962, opened 2020-05-25
tmux: add support for the XDG Base Directory Specification

commiyou:
#  Changes:
add support for the XDG Base Directory Specification


bartekpacia (approved):

robbyrussell:

This landed in 2023 through #11103, so the plugin already looks for your tmux config under `$XDG_CONFIG_HOME`. The version that shipped keeps `~/.tmux.conf` first and falls back to the XDG path, where yours preferred XDG... that ordering is the safer one for anyone who happens to have both files, so it's just as well it went that way round.

Worth knowing too: tmux itself reads `$XDG_CONFIG_HOME/tmux/tmux.conf` natively since 3.2, so on a current tmux this works even without the plugin's help.

Closing as already done. Sorry you didn't hear back at the time.

PR #8965, opened 2020-05-25
Add Kube Context to the default prompt of Agnoster theme

austinbv:
## Standards checklist:

## Changes:

- Adds a kube context line to the Agnoster theme
- If the user doesn't have a current context set or does not have kubectl installed the prompt is a noop

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -242,11 +242,27 @@ prompt_aws() {
   esac
 }
 
+prompt_kubecontext() {
+  [[ -z "$(which kubectl)" ]] && return
+  local kube_prompt="k8s: "
+  local current_context="$(kubectl config current-context)"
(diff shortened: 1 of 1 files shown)


kbakdev (approved):

robbyrussell:

The segment goes into `build_prompt` unconditionally, so every agnoster user would get it. agnoster is one of the most-used themes in the project, and adding something to everyone's prompt needs demand behind it, which this hasn't got.

The guard also doesn't do what it looks like it does. zsh's `which` prints "kubectl not found" to stdout and exits 0, so `[[ -z "$(which kubectl)" ]]` is never true and the early return never fires... people without kubectl would pay two process spawns on every prompt render. `(( $+commands[kubectl] ))` is the check that works, if you're still using this locally.

There's a later PR doing the same thing with more interest behind it, #11510, so that's the thread to follow. The route we'd prefer for segments like this is #7355, which makes agnoster's segment list configurable rather than hardcoding another one.

Issue #8959, opened 2020-05-22
Add .nvmrc loading to nvm plugin


endquote:
Can we add [this bit](https://github.com/nvm-sh/nvm#zsh) to the [nvm plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/nvm), to auto-load `.nvmrc` files?

I added it to my `.zshrc` file with the `nvm` plugin enabled, and it seems to work correctly.

I could do a PR, but wanted to make sure I wasn't missing something.

@nicholasjpaterno

nicholasjpaterno:
@endquote I'm not actually a maintainer of ohmyzsh, sorry for the confusion. Definitely make a PR! :smile:

endquote:
@nicholasjpaterno Gotcha, I just tagged you since it looked like you did the `nvm` plugin.

endquote:
Actually it looks like @drGrove did it already in #5782 in 2017, but it was never merged.

PR #8976, opened 2020-05-30
nvm: change node version when .nvmrc is found

endquote:
## Standards checklist:

## Changes:

In the `nvm` plugin, I added code which detects if `.nvmrc` is present and switches to the requested node version if so. It's copied from here:
https://github.com/nvm-sh/nvm#zsh

I opened the issue #8959 as well.

## Other comments:

Upon searching, I see another PR #5782 which made the same change, but it's been waiting to be merged for over three years. So merging that one would be fine too.

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -6,3 +6,26 @@ if ! type "nvm" &> /dev/null; then
     # Load nvm if it exists
     [[ -f "$NVM_DIR/nvm.sh" ]] && source "$NVM_DIR/nvm.sh"
 fi
+
+# Calling nvm use automatically in a directory with a .nvmrc file
+# https://github.com/nvm-sh/nvm#zsh
+autoload -U add-zsh-hook
+load-nvmrc() {
+  local node_version="$(nvm version)"
+  local nvmrc_path="$(nvm_find_nvmrc)"
+
+  if [ -n "$nvmrc_path" ]; then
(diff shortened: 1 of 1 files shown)


xaqrox:
+1 on this feature! The repo that #5782 is from seems to be deleted, so I think we should favor this PR. I also had the idea that perhaps this could include a `ZSH_NVM_AUTOLOAD` option (defaulted to enabled) so that this behavior could be easily disabled if needed (the terminal feels a bit slow with this on sometimes). I'm not sure how to add commits to someone else's PR so I'll leave it up two @endquote or the maintainers if anyone thinks that is a good idea.

plugins/nvm/nvm.plugin.zsh
```diff
@@ -6,3 +6,26 @@ if ! type "nvm" &> /dev/null; then
...
+
+# Calling nvm use automatically in a directory with a .nvmrc file
+# https://github.com/nvm-sh/nvm#zsh
+autoload -U add-zsh-hook
+load-nvmrc() {
+  local node_version="$(nvm version)"
+  local nvmrc_path="$(nvm_find_nvmrc)"
+
+  if [ -n "$nvmrc_path" ]; then
+    local nvmrc_node_version=$(nvm version "$(cat "${nvmrc_path}")")
```
Aloxaf, line 18: There is not need to use cat here.
```suggestion
    local nvmrc_node_version=$(nvm version "$(<$nvmrc_path)")
```

PR #8981, opened 2020-06-01
Add missing dependency - cmp (GNU Diffutils)

xerrni:
## Standards checklist:

## Changes:

- Add missing dependency - cmp which is a part of [GNU Diffutils](https://www.gnu.org/software/diffutils/). cmp is used here: [oh-my-zsh.sh:71](https://github.com/ohmyzsh/ohmyzsh/blob/93cc3964e2d265ab0571298d69d2eed0a65d13f2/oh-my-zsh.sh#L71)


mcornella:
Isn't cmp available everywhere? If not, we can just use diff.

xerrni:
No it's not, for example Arch Linux docker image does not have it.
To test it:
```shell
sudo docker run -it archlinux bash
```
Then inside docker container:
```shell
command -v cmp || echo "not installed"
command -v diff || echo "not installed"
```
Result:
```shell
not installed
not installed
```
diff is also from GNU Diffutils package.

To check which package provides file (e.g. cmp):
```shell
pacman -Fy
pacman -F cmp
```

mcornella:
I've changed the logic to use `grep` instead. Thanks for the report!

PR #8984, opened 2020-06-02
Add option for classic green logo text on upgrade

gildartsirc:
## Standards checklist:

  - It doesn't, but neither does the code around my insertion. I matched the local style.

## Changes:

- Adds an option to use the classic green logo text to match one of the sticker options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,12 @@ To disable automatic upgrades, set the following in your `~/.zshrc`:
 DISABLE_AUTO_UPDATE=true
 ```
 
+To use classic green logo after upgrade, set the following in your `~/.zshrc`:
+
+```shell
+UPDATE_GREEN=true
+```
+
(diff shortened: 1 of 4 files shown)


mcornella:
Why, no love for the rainbow version? I don't really want to add yet another setting for this.

gildartsirc:
Just prefer the green color. And it matches one of the sticker designs that is sold. Which was the reasoning behind changing it to a rainbow in the first place as far as I can tell.

If adding another option is a no-go, I can deal with it obviously. Just thought I would add the option.

mcornella:
I can't accept this sorry, thanks for contributing!

timka:
For those who still want the classic green it's safe to just change the logo printing part in `~/.oh-my-zsh/tools/upgrade.sh` by replacing $RAINBOW with $GREEN.

PR #9008, opened 2020-06-07
Fix per-directory-history plugin to use local history in new shell

### LINKED ISSUE #9007: per-directory-history plugin incorrectly uses global history in newly created shell
**Describe the bug**
The per-directory-history plugin does not use the per-directory history immediately after creating a new shell. The per-directory history is only used when changing to global and back to per-directory mode (by pressing Ctrl+G twice) or when changing directories (e.g. `cd .`).

**To Reproduce**
1. Enable the per-directory-history plugin.
2. Create a new directory for the experiments (`/tmp/test` in this example).
3. Change to the test directory (`cd /tmp/test`).
4. Run several commands in the test directory in order to create a history:
  `echo 1`
  `echo 2`
  `echo 3`
5. Close the shell and open a new one in a different directory.
6. In that directory, run a few other commands:
  `echo 4`
  `echo 5`
7. Open a new shell in the test directory.
  **Caution:** The bug will only appear when the new shell is opened while already in that directory. It will not appear when changing to the directory *after* creating the shell.
8. Browse through the history using the Arrow Up key. The commands that were run in the other directory will be visible in the history.

**Expected behavior**
Instead of using the global history, the newly created shell should use the history of the directory in which it was created.

**Desktop**
 - OS / Distro: macOS 10.15.5
 - Latest ohmyzsh Update?: Yes
 - ZSH Version: zsh 5.7.1 (x86_64-apple-darwin19.0)
 - Terminal emulator: iTerm2

**Additional context**
Theme: powerlevel10k
Other active plugins: bgnotify command-not-found copydir dirhistory git safe-paste z

smarsching:
This patch fixes the problem that in a newly created shell, the global
history would be used instead of the per-directory on. It does this by
delaying the loading of the per-directory history until just before the
prompt is created.

## Standards checklist:

## Changes:

- Fixes #9007.
- The bug seems to be caused by ZSH loading the history from HISTFILE *after* the plugin is initialized, thus overriding the local history loaded by the plugin with the global history.
- This PR fixes this problem by delaying the plugin initialization.

## Other comments:
- This might also fix https://github.com/jimhester/per-directory-history/issues/5.
- @jimhester might want to review this.

### DIFF
--- a/plugins/per-directory-history/per-directory-history.zsh
+++ b/plugins/per-directory-history/per-directory-history.zsh
@@ -118,6 +118,15 @@ function _per-directory-history-addhistory() {
   fi
 }
 
+function _per-directory-history-precmd() {
+  if [[ $_per_directory_history_initialized == false ]]; then
+    # start in directory mode
+    mkdir -p ${_per_directory_history_directory:h}
+    _per_directory_history_is_global=true
+    _per-directory-history-set-directory-history
+    _per_directory_history_initialized=true
+  fi
+}
 
 function _per-directory-history-set-directory-history() {
   if [[ $_per_directory_history_is_global == true ]]; then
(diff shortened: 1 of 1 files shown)


jimhester:
LGTM, @smarsching would you be willing to open a PR with the same patch over at https://github.com/jimhester/per-directory-history?

smarsching:
@jimhester I created a PR for you: https://github.com/jimhester/per-directory-history/pull/37

It’s a slightly adjusted version of this PR because the initialization logic in the version of the plugin in the per-directory-history repository was slightly different.

53jk1 (approved):

last-partizan:
Could this be merged please?

TheTechRobo:
They have a huge backlog of PRs; if you really want it to be merged, try asking on the Discord server.

PR #9021, opened 2020-06-10
updated .gitignore

bartekpacia:
## Standards checklist:

## Changes:

- Added a wildcard (`*`) to ignore all `.DS_Store` files. I really don't know what else to say.

## Other comments:
It's annoying for me to have all those nested `.DS_Store` included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ custom/
 cache/
 log/
 *.swp
-.DS_Store
+*.DS_Store


mcornella:
I was under the impression that these files were all named `.DS_Store`. Has that changed?

bartekpacia:
It might be just me, but I had nested `.DS_Stores` getting added to git's staging area and adding that wildcard fixed that🤷‍♂️

mcornella:
When you say nested, what do you mean? What're the filenames?

bartekpacia:
I mean files not in the root (`.oh-my-zsh/`), but in its children directories.

mcornella:
The way it's written in the `.gitignore` file, any file named `.DS_Store` anywhere in the repository will be ignored. There has to be something else happening here:
image

bartekpacia:
Okay, so it was some other problem on my side. Thanks for response!

PR #9029, opened 2020-06-14
Add plugin clipboard

bb2020:
## Standards checklist:

## Changes:

- Adding a new plugin.

## Other comments:

This plugin replaces kill-ring of zsh with system clipboard by overriding default bindings. Copying and pasting between any application and zsh becomes much more seamless.

### DIFF
--- a/plugins/clipboard/README.md
+++ b/plugins/clipboard/README.md
@@ -1,0 +1,9 @@
+# clipboard plugin
+
+This plugin integrates zsh kill-ring with system clipboard.
+
+It makes it possible to kill in zsh and paste into browser/editor or vice versa.
+
+Key bindings that rebound by the plugin are `C-y`, `C-k`, `C-u`, `M-d`, `M-backspace` and `M-w`.
+
+The plugin should not change behaviour of these keys.
--- a/plugins/clipboard/README.md
+++ b/plugins/clipboard/README.md
@@ -7,3 +7,12 @@ It makes it possible to kill in zsh and paste into browser/editor or vice versa.
 Key bindings that rebound by the plugin are `C-y`, `C-k`, `C-u`, `M-d`, `M-backspace` and `M-w`.
 
 The plugin should not change behaviour of these keys.
+
+| Key binding | Description                                                |
+|-------------|------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


TriMoon:
I would suggest a different name for this plugin, the current title suggest **clearing** the clipboard.
Maybe a good alternative would be just plain "clipboard"?

bb2020:
I agree. I was thinking the same thing.

TriMoon:
PS: why the strange extension on the README?
`.org` should be `.txt` because it's a plain text file...

bb2020:
It is not plain text by the way. Org-mode is rendered by Github, similar to Markdown. It is documented [here](https://github.com/github/markup).

bb2020:
Is Markdown the only accepted format?

TriMoon:
> @bb2020 
> It is not plain text by the way. Org-mode is rendered by Github, similar to Markdown. It is documented [here](https://github.com/github/markup).
> Is Markdown the only accepted format?

First time i hear about `org` mode, no idea if markdown is the **only** accepted format but it is the most used one for sure... :wink:

bb2020:
Yeah, that's for sure. `org-mode` is the awesome Emacs note taking format. As this plugin modifies zsh Emacs bindings, `org` is more relative to the context.

mcornella:
Please use Markdown for consistency and document what each key binding does.

bb2020:
Done. Actions of the key bindings are unchanged. It tries to modify clipboard or cutbuffer while executing the same action.

mcornella:
This is not as easy to understand as you think it is. A little background into what a kill-ring or cutbuffer are would go a long way for this plugin to be used. Plus, I know the key bindings are the same as the default (only adding the text to the clipboard), but it's always a good idea to describe what they do. A Markdown table would be great, for example:
```markdown
| Key binding | Description                                       |
|-------------|---------------------------------------------------|
| CTRL+K      | Cuts the line and puts it in the system clipboard |
```

bb2020:
Done with the table.

53jk1 (approved):

bb2020:
Any updates on this getting merged?

bb2020:
Any updates on this getting merged?
Ping @mcornella

PR #9049, opened 2020-06-19
git: change main branch naming scheme

adam2k:
## Standards checklist:

## Changes:

- Changed the git plugin commands from master to main.


mcornella:
This is a signifcant change, I'd wait until a backwards compatible solution is ready to merge this. One possibility is to check whether main exists, and use that in that case. For example, we could use this function:
```zsh
git_main_branch() {
  if [[ -n "$(git branch --list main)" ]]; then
    echo main
  else
    echo master
  fi
}
```
and substitute the word master for a call to this function in the aliases that use it.

adam2k:
@mcornella good idea!  I'll make that update.

adam2k:
@mcornella let me know if that looks good to you.

mcornella:
Thanks Adam!

ouuan:
I think it's better to make this customizable. e.g. use an environment variable.

The current implementation is a breaking change if there's the main branch but it's not used as the default branch. Also, the user can't use names besides main and master.

ouuan:
It will be better if we can configure it per repository, but I've no idea how to do it :thinking:

ouuan:
Even if we don't make it configurable, why not "if master doesn't exist, use main"? I think anyone using "main" instead of "master" doesn't use "master” anymore, but someone using "master" as the default branch may use "main" as a normal branch.

mcornella:
I pushed your suggestion (1db79f97f5523cd2c76b5dfb8185a6d3b7a789d9) instead of the other PR, with attribution. Makes more sense, thanks.

Poyoman39:
Hey guys, not really happy about this PR. It looks like a really good idea, ... but 'git_main_branch' is used in kind of a lot of tools other than oh my zsh ... which mean a lot of people may have defined it something really different than a master branch. In my case ... git_main_branch is develop ... Since i want developers on my project to land on it after clone. But then gcm become the same as gcd. And also no more git checkout master shortcut anymore :-1:  

This is based in my personal projects. May be useful to have other users giving their feelings about it too.

mcornella:
Huh, so would it work if the remote main branch is only checked if no master nor main branches are found, instead of the way it is now where it checks first the remote main branch? Or at that point `master` doesn't exist yet, so it'll use `develop` anyway? It's tricky to make everyone happy about this, apologies for that.

Poyoman39:
@mcornella no problem ;) Since it's a breaking change it will for sure break some users habits. Wanted to share my case ... and yes i've no idea how to make both solutions work. I was thinking about adding a gcma alias ... but looks more like a quick patch than a real solution.

ouuan:
> It's tricky to make everyone happy about this

I think #9091 is a solution :thinking:  Now here's one more case that configurable main branch name could be useful.

@Poyoman39 Even if #9091 is rejected, you could use the code in that PR to override `git_main_branch` in your `.zshrc`. I hope it helps.

PR #9052, opened 2020-06-20
encode64: fix typo in README

insign:


lutzhorn:
Yes, this seems to be a typo. Both `decode64` and `d64` should be mentioned in the documentation example. LGTM

PR #9059, opened 2020-06-23
archlinux: fix pacweb breaking when multiple packages found

kattjevfel:
…repos

## Standards checklist:

## Changes:

- Fix so that `pacweb` doesn't produce an invalid URL when package is available in multiple repos.

## Other comments:

Tested with rsync and works as intended.
Previously: <https://www.archlinux.org/packages/testing%0Aextra/x86_64%0Ax86_64/rsync/>
This PR: <https://www.archlinux.org/packages/testing/x86_64/rsync/>


mcornella:
Thanks!

PR #9062, opened 2020-06-24
update: only update on a valid affirmative input

### LINKED ISSUE #9057: Update defaults to Yes on any input other than N or n
**Describe the bug**
When prompted to update Oh-My-Zsh the prompt defaults to Yes for any keypress other than `N` or `n`. 

**To Reproduce**
1. Start shell when there is an update available.
2. At the update prompt type `b`
3. Oh My Zsh begins to update.

**Expected behavior**
Oh My Zsh should only update when I've explicitly told it to by selecting `Y`, `y` or pressing enter to select the default value of `Y`.

 - OS / Distro: macOS
 - Latest ohmyzsh Update?: Yes
 - ZSH Version: 5.8
 - Terminal emulator iTerm2

alexagnelli:
## Standards checklist:

## Changes:

- When prompt to update ohmyzsh, the input accepted is: yY and enter to proceed and nN to deny, if an invalid input is given the user gets prompted again

## Other comments:

It fixes #9057

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -75,8 +75,15 @@ function update_ohmyzsh() {
         while read -t -k 1 option; do true; done
         [[ "$option" != ($'\n'|"") ]] && echo
 
-        echo -n "[oh-my-zsh] Would you like to update? [Y/n] "
-        read -r -k 1 option
+        # reset option to avoid autoaccept if enter was the last character
+        option=""
+
+        while [[ ! $option =~ [nNyY$'\n'] ]]
+        do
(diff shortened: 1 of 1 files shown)


mcornella:
Can you instead only change the `case` logic so that [Yy\n] triggers an update, and everything else doesn't? I'm not a fan of asking again if the input isn't valid, I think it'll be a support requests nightmare.

alexagnelli:
Done! @mcornella

mcornella:
Ok, I also changed it to make sure the `last-updated` file isn't updated if any other character was pressed by accident. Thanks @alexagnelli!

PR #9066, opened 2020-06-27
Update to scd 1.4.0

pavoljuhas:
## Standards checklist:

  (except of 4-space indent to stay identical with upstream version)

## Changes:

- Add command-line completion for zsh.
- Support for path exclusion according to patterns in `~/.scdignore`.
- New option `-p,--push` to invoke `pushd` instead of `cd`.
- Implicit indexing of parent directories to all added paths.
- Special patterns `./` to match under PWD, `^` and `$` for
  start and end anchors, `:PAT` to match in the tail component of target path.
- Add cleanup of obsolete aliases with `scd --unalias OLD`.
- Use smart case matching with `(#l)` zsh globbing flag.

## Other comments:

Catch up with https://github.com/pavoljuhas/smart-change-directory/releases/tag/v1.4.0.


pavoljuhas:
ping @mcornella - here is an update of my omz plugin.  Please let me know if there are any questions or suggestions.

pavoljuhas:
@mcornella - PTAL.

PR #9067, opened 2020-06-28
VI-Mode Plugin: Keybindings 

ssbanerje:
## Changes:

- Introduces a helper function to map key bindings to different vim modes in zsh (e.g., `vicmd`, `visual`)
- Adds support for `ysiw"` kind of operators like the vim-surround plugin.
- Makes some of the old Emacs keybindings (e.g., `C-a`, `C-e`) available only a subset of modes.
- Moves the `edit-command-line` command to `^v` and allows entering `visual` mode from `vicmd` mode using the `v`.


mcornella:
There were recent changes to the `vi-mode` plugin. Figure out whether this PR still stands and rebase if it does.

PR #9072, opened 2020-07-01
agnoster: fix git working tree check

kylelaker:
## Standards checklist:

## Changes:

`git rev-parse --is-inside-work-tree` does not have a different exit
code when not within a work tree; the difference is in the output,
either "true" when in a work tree, or "false" when not.

The if was modified similarly to the check for the hide-status config,
which also parses command output rather than exit code.

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -106,7 +106,7 @@ prompt_git() {
   }
   local ref dirty mode repo_path
 
-  if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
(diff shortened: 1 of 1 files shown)


mcornella:
To clarify, here are the possible outputs:
```console
# not in a git repository (128 exit code)
$ git rev-parse --is-inside-work-tree
fatal: not a git repository (or any of the parent directories): .git

# inside .git folder
$ git rev-parse --is-inside-work-tree
false

# inside work tree of git repository
$ git rev-parse --is-inside-work-tree
true
```
So checking if the output is `true` is a good solution.

PR #9078, opened 2020-07-03
introduce reload_zsh

holymonson:
## Standards checklist:

## Changes:

Introduce `reload_zsh()`, which could easily remove the zcompdump file and reload zsh.

## Other comments:

This should be associated with #8802 and #8801, and better with #9074.
Close #8800.

### DIFF
--- a/plugins/zsh_reload/README.md
+++ b/plugins/zsh_reload/README.md
@@ -1,23 +1,3 @@
 # zsh_reload plugin
 
-The zsh_reload plugin defines a function to reload the zsh session with
-just a few keystrokes.
-
-To use it, add `zsh_reload` to the plugins array in your zshrc file:
-
-```zsh
-plugins=(... zsh_reload)
-```
-
-## Usage
-
-To reload the zsh session, just run `src`:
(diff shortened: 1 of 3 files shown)


bartekpacia:
After investigating the topic this looks like a good idea. Thank you.

plugins/zsh_reload/README.md
```diff
@@ -1,23 +1,3 @@
 # zsh_reload plugin
 
-The zsh_reload plugin defines a function to reload the zsh session with
```
Fhiltyphil88, line 3:
```suggestion
The zsh_reload plugin defines a function to reload the zsh session with
```

Fhiltyphil88 (commented):
@atmos

TheTechRobo (approved):

mcornella:
I'd prefer this be in the `omz` CLI so that if this gets documented somewhere external to OMZ at least it is made very clear that it comes from Oh My Zsh and therefore needs it. You don't need to do anything, I'll quickly make the edits myself.

PR #9080, opened 2020-07-03
vscode: add support for VSCodium

SteelShot:
## Standards checklist:

## Changes:
* Added support to detect VSCodium, the plugin will now choose in the order of VS Code > VS Code Insiders > VSCodium which executable to use for the aliases, ofc this can still be overriden as before by adding the variable manually to .zshrc file. 
* Added a check in case all 3 editors are missing, then the aliases wont be added. 
* Refactored the README quite a bit to match the changes and hopefully improve it.

## Other comments:
@MarsiBarsi, @babakks

### DIFF
--- a/plugins/vscode/README.md
+++ b/plugins/vscode/README.md
@@ -31,7 +31,7 @@ the Command Palette via (F1 or ⇧⌘P) and type shell command to find the Shell
 
 ## Using multiple flavours
 
-If for any reason, you ever require to use multiple flavours of VS Code i.e. VS Code and VS Code Insiders, you can 
+If for any reason, you ever require to use multiple flavours of VS Code i.e. VS Code (stable) and VS Code Insiders, you can 
 manually specify the flavour's executable. Add the following line to the .zshrc file (between the `ZSH_THEME` and the `plugins=()` lines).
 This will make the plugin use your manually defined executable.
 
--- a/plugins/vscode/README.md
+++ b/plugins/vscode/README.md
@@ -1,6 +1,6 @@
 # VS Code
 
-This plugin makes interaction between the command line and the VS Code editor easier.
+This plugin provides useful aliases to simplify the interaction between the command line and VS Code or VSCodium editor.
 
 To start using it, add the `vscode` plugin to your `plugins` array in `~/.zshrc`:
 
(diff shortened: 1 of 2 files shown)


plugins/vscode/vscode.plugin.zsh
```diff
@@ -3,28 +3,33 @@
 #   https://github.com/MarsiBarsi (original author)
 #   https://github.com/babakks
```
babakks, line 5: You can add your profile, if you like.

SteelShot: Done

plugins/vscode/README.md
```diff
@@ -10,23 +10,36 @@ plugins=(... vscode)
...
-## VS Code Insiders
+[For VSCodium](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#how-do-i-open-vscodium-from-the-terminal), open
+the Command Palette via (F1 or ⇧⌘P) and type shell command to find the Shell Command:
+> Shell Command: Install 'codium' command in PATH
 
-🍏 **If you are only using [VS Code Insiders](https://code.visualstudio.com/insiders/), the plugin will automatically bind to your Insiders installation.**
+## Using multiple flavours
 
-But, if you have both Stable and Insiders versions and want to configure the plugin to just use the Insiders version, add the following line in the oh-my-zsh settings section (between the `ZSH_THEME` and the `plugins=()` line). This will make the plugin use the Insiders version instead.
+If for any reason, you ever require to use multiple flavours of VS Code i.e. VS Code and VS Code Insiders, you can
```
babakks, line 34:
```suggestion
i.e. VS Code and VS Code Insiders
i.e. VS Code (stable) and VS Code Insiders
```

babakks: I feel the title _insiders_ fails to communicate with non-native English readers, so they won't get informed of the unstable nature of the release on the spot. Adding the "stable" hint helps them figure out the difference.

SteelShot:
@MarsiBarsi and @babakks, if you could review this final commit, then it is be ready to be merged into master.

plugins/vscode/vscode.plugin.zsh
```diff
@@ -1,30 +1,44 @@
...
+  alias vscl="$VSCODE --log"
+  alias vscde="$VSCODE --disable-extensions"
+}
 
-alias vsced="$VSCODE --extensions-dir"
-alias vscie="$VSCODE --install-extension"
-alias vscue="$VSCODE --uninstall-extension"
+# Verify if any manual user choice of VS Code exists first.
+# Otherwise, try to detect a flavour of VS Code.
+if ! [ -z ${VSCODE+x} ] && which $VSCODE > /dev/null; then
```
babakks, line 28: Sorry, but I think it's better to let the user see the error messages if the aliases failed to work. Our goal is to make things simple, not undetectable or hard to trace.

What do you think @MarsiBarsi?

mcornella:
I changed that because there isn't an alias for just the VS Code flavor, but if there's such strong opposition I'll leave it like that.

I added a warning for when `$VSCODE` is not found. If this is OK with everybody this is ready to be merged.

SteelShot:
Yeah, merge away.

As a side note: @mcornella, for future (i am happy to do the initial/full work too) would it not be better for ohmyzsh to introduce some plugin convention in handling errors. i.e making use of return values as they are being sourced. Hence why i returned a value when there was no valid executable defined/found. ohmyzh could print something like "[ohmyzsh] plugin x failed/had an error" etc. etc?

babakks:
@mcornella Totally agree with the warning. Let's merge this.

mcornella:
Thanks folks!

PR #9090, opened 2020-07-07
Change default ZSH_COMPDUMP directory to ZSH_CACHE_DIR

ctranstrum:
## Standards checklist:

## Changes:

- Fixes issue #7332
- Approach taken from a comment by @mcornella at https://github.com/ohmyzsh/ohmyzsh/issues/7332#issuecomment-624221366
- Seemed like an easy fix that I was also interested in, so I thought I'd contribute

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -53,7 +53,7 @@ fi
 
 # Save the location of the current completion dump file.
 if [ -z "$ZSH_COMPDUMP" ]; then
-  ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
+  ZSH_COMPDUMP="${ZSH_CACHE_DIR}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
 # Construct zcompdump OMZ metadata


gaving:
Any chance of seeing this merged?

Would save setting our own variable such as the workarounds in https://github.com/ohmyzsh/ohmyzsh/issues/7332.

53jk1 (approved):

tw-space:
Smart change, I'd like to see this merged too. Is the next action on @ctranstrum?

mcornella:
The blocker for this is that all the previous suggestions to fix completion issues by doing `rm ~/.zcompdump*` will be obsolete. I'm not sure yet how to proceed with this part of the problem.

Drumstickx:
While this push request has been stale for more than a year, I would like to "revive" it for two reasons:

1. I like the idea that applications use commonly agreed upon directories (e.g., `.cache`, `.config`) for their files. Personally, I strive for a minimal amount of dotfiles and -directories directly in my home directory
2. I agree that it would be somehow a breaking change with regards to past documentation. However, sometimes breaking changes are required to move forward and improve mistakes from the past. The longer we wait on this, the more anwers will refer to the "old" location.

VikalpRusia:
+1

carlosala:
It seems like, [as stated in the manual](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Initialization), compdump file should go in `$ZDOTDIR`, or `$HOME` if it's not defined, as well as all other shell configuration files. We are [already respecting that](https://github.com/ohmyzsh/ohmyzsh/blob/b4f9698733d7b29cc495e649e26fd6c3a5dcfcae/oh-my-zsh.sh#L105), so it should not be changed.
Thanks for your contribution!

bavalpey:
@carlosala 

The manual specifies where these are placed by default. It does not specify where they _should_ go. Not allowing them to use an environment variable or something similar means this violates XDG compliance.

PR #9091, opened 2020-07-07
git: make git_main_branch configurable

ouuan:
## Standards checklist:

## Changes:

Make the function `git_main_branch` configurable. You can set the environment variable `GIT_MAIN_BRANCH_PRIORITY` which is a list of branch names separated by `:`. The function `git_main_branch` checks for all branches in the list and returns the first existing branch in the list.

The default `GIT_MAIN_BRANCH_PRIORITY` is `master:main` instead of `main:master`, and `git_main_branch` returns `master` if no branch in `GIT_MAIN_BRANCH_PRIORITY` exists, so that it's not a breaking change (compared to the version before #9049).

## Other comments:

This is an enhancement of #9049. The contributor of that PR: @adam2k. The reviewer of that PR: @mcornella.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -181,10 +181,12 @@ plugins=(... git)
 
 ### Main branch preference
 
-Following the recent push for removing racially-charged words from our technical vocabulary, the git plugin favors using
-a branch name other than `master`. In this case, we favor the shorter, neutral and descriptive term `main`. This means
-that any aliases and functions that previously used `master`, will use `main` if that branch exists. We do this via the
-function `git_main_branch`.
+Following the recent push for removing racially-charged words from our technical vocabulary, users might favor using
+a branch name other than `master`. Now you can configure this by the environment variable `GIT_MAIN_BRANCH_PRIORITY`.
+It is a list of branch names separated by `:`. The function `git_main_branch` checks for all branches in the list and
+returns the first existing branch in the list. `git_main_branch` is used in many aliases which use `master` as the main
+branch before. The default `GIT_MAIN_BRANCH_PRIORITY` is `master:main` instead of `main:master`, and `git_main_branch`
+returns `master` if no branch in `GIT_MAIN_BRANCH_PRIORITY` exists, so that it's not a breaking change.
 
 ### Deprecated aliases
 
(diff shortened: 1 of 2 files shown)


mcornella:
I like the current solution, I think projects with clout like this one should take a stand and push for a common alternative to the master term. If we let this be configurable we won't all converge towards a common term, and I think `main` is one of the strongest candidates.

As for `main` being present in existing repositories (and not being used as a `master` equivalent), we'll deal with that when the situation presents itself, but in the meantime I don't think it's necessary. For what it's worth though, you can customize the `git_main_branch` yourself and put it in your zshrc file if you don't agree.

@robbyrussell thoughts?

ouuan:
> For what it's worth though, you can customize the `git_main_branch` yourself and put it in your zshrc file if you don't agree.

That's a good point :thinking: Maybe we can document it?

ouuan:
> should take a stand and push for a common alternative to the master term. If we let this be configurable we won't all converge towards a common term

I think the default setting is a "common term"? The name of the default branch on GitHub is configurable, but "master" as the default name of the default branch on GitHub makes nearly everyone use it.

ouuan:
It's also possible that using a name besides "master" and "main" is the best. Let's consider this case:

I have a repository hosting GitHub Pages that contains both the source file of the website and the files to be deployed. I can only use `master` as the branch which contains the files to be deployed because "user pages must be built from the master branch". Then it's strange to have both "master" and "main", I may use the branch "source" or something else to store the source files.

mcornella:
You can use a [`gh-pages` branch](https://docs.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source), but either way, this is something GitHub is actively working on so it's very possible that the reference to `master` will just become the default branch of the repository.

ouuan:
> You can use a [`gh-pages` branch](https://docs.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source)

image

It can only use the master branch if the repository is `<username>.github.io`.

And I'm not asking for solutions, I'm providing an example where the name of the default branch can be something besides master and main.

However, I'm overriding `git_main_branch` in my `.zshrc` now. Hopefully, someone interested in a configurable main branch name can find this PR.

mcornella:
It turned out that the ['simple' algorithm failed](https://github.com/ohmyzsh/ohmyzsh/issues/9103), so I used part of your logic to fix it, with co-authorship: 55a98fc06d9ea9d003683dec5aae6cb6b0130ab0.

PR #9092, opened 2020-07-07
separate metadata file

holymonson:
## Standards checklist:

## Changes:

Separate metadata into an independent file. This is an alternative of #9074.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -51,46 +51,29 @@ else
   SHORT_HOST=${HOST/.*/}
 fi
 
-# Save the location of the current completion dump file.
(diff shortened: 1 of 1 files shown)


bartekpacia (approved):

bartekpacia (approved):

kbakdev (approved):

robbyrussell:

`oh-my-zsh.sh` has changed a lot since 2020, and the metadata check you were reworking here got reworked upstream in #14061. The other thing that stops this one is that it drops `$ZSH_COMPDUMP`, which people set to move the dump file, and `omz cache clean` still relies on it.

Closing for that reason rather than because the idea was wrong. Splitting the fingerprint out of the dump file was a reasonable thing to want.

PR #9094, opened 2020-07-07
cli: add update command

### LINKED ISSUE #8003: Make the upgrade command more easily discoverable
As I move around various networks, auto upgrade sometimes doesn't work because of proxy settings.  As such, I like to perform a manual update when convenient.  It's frustrating, though, because the upgrade command isn't immediately discoverable because it's not an option to oh-my-zsh itself, and doesn't start with ohmyzsh, which are the first things I look for when trying to find a new command.

I'd like an option (in addition to the existing upgrade_oh_my_zsh that is more easily discoverable from the command line, rather than me having to google it every 6 months or so. :)

I would think one of the following would be appropriate:
- Add an oh_my_zsh_upgrade (or ohmyzsh_upgrade) script along side the existing script 
- Add a --upgrade option to ohmyzsh 
- Add an install-time alias similar to the zshconfig alias
- Add some text to the ohmyzsh script listing the upgrade command

I know it's possible to create my own alias on my machines, but it would be nicer if I didn't have to remember to create it myself on each machine I install on. :)

vladimyr:
## Standards checklist:

## Changes:

Extend `omz` CLI (https://github.com/ohmyzsh/ohmyzsh/pull/8876) with `update` command that invokes `lib/upgrade.sh`.

Fixes #8003

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -23,7 +23,7 @@ function _omz {
     local -a cmds subcmds
     cmds=(
         'help:Usage information'
-        'upgrade:Upgrade Oh My Zsh'
+        'update:Update Oh My Zsh'
         'pr:Commands for Oh My Zsh Pull Requests'
     )
 
@@ -49,7 +49,7 @@ Usage: omz <command> [options]
 Available commands:
 
     help                Print this help message
-    upgrade             Upgrade Oh My Zsh
+    update              Update Oh My Zsh
     pr <command>        Commands for Oh My Zsh Pull Requests
 
 EOF
@@ -80,11 +80,6 @@ function _omz::log {
         warn) print -P "%S%F{yellow}$logname%f%s: $@" ;;
         error) print -P "%S%F{red}$logname%f%s: $@" ;;
     esac >&2
-}
-
-function _omz::upgrade {
-    trap "{ rc=\$?; rm -rf '\$ZSH/log/update.lock'; return \$rc; }" EXIT INT QUIT
-    env ZSH="$ZSH" sh "$ZSH/tools/upgrade.sh"
 }
 
 function _omz::pr {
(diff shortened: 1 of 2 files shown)


mcornella:
This should also update the `.zsh-update` file and delete the lock file if it exists, see:

https://github.com/ohmyzsh/ohmyzsh/blob/97cf0bf4bd7655767e9e5ea27ddb6fbe550a305d/tools/check_for_upgrade.sh#L17-L29

and

https://github.com/ohmyzsh/ohmyzsh/blob/0caae9082a3bbb14b05938a7d0f292c2e4d161f4/lib/functions.zsh#L9-L12

Also, add the `update` command to the `_omz` completion function.

vladimyr:
@mcornella Thank you for your review. After closer examination, I figured out the are important differences between `update_ohmyzsh` and `upgrade_oh_my_zsh`.
Former is called as part of auto-update feature and if you disable it with `DISABLE_AUTO_UPDATE="true"` it doesn't even get defined. It updates `.zsh-update` file while lockfile gets deleted by trap set inside `check_for_upgrade.sh`: https://github.com/ohmyzsh/ohmyzsh/blob/97cf0bf4bd7655767e9e5ea27ddb6fbe550a305d/tools/check_for_upgrade.sh#L50-L54
OTOH `upgrade_oh_my_zsh` does not touch `.zsh-update` file and lockfile gets deleted by `rm` invocation.
It brings me to conclusion that update and upgrade are in fact two completely different concepts. Former does not depend nor touch `.zsh-update` file while the latter does. Because `README` clearly states one should use `upgrade_oh_my_zsh` for manual updates I decided to rename command from `update` to `upgrade` and:
1. do not update `.zsh-update` file
2. remove lockfile using a trap

mcornella:
> OTOH `upgrade_oh_my_zsh` does not touch `.zsh-update` file and lockfile gets deleted by `rm` invocation.
> It brings me to the conclusion that update and upgrade are in fact two completely different concepts. Former does not depend nor touch `.zsh-update` file while the latter does.

This is just a side effect of the codebase not being consistent enough, they should both do the same thing and there is no semantic difference between `update` and `upgrade`.

> Because `README` clearly states one should use `upgrade_oh_my_zsh` for manual updates I decided to rename command from `update` to `upgrade`

The README talks about receiving updates and the settings also use `UPDATE`; the `upgrade` word is in the process of being replaced by `update`, so please change that back.

Also, a manual update should reset the side effects of the automatic update, so even if the `upgrade_oh_my_zsh` function doesn't, this should both update the last updated file (`.zsh-update`) and remove the lock dir when finishing the update. The `trap` isn't necessary since we're not using the lock dir at all, you can just remove it afterwards.

PR #9095, opened 2020-07-07
fix(vim-interaction): look up the newest GVim instance

johannfr:
## Standards checklist:

## Changes:

Instead of relying on a hardcoded instance-name existing for the default-value, let's look up the latest instance by getting the latest line from `gvim --serverlist`. This also resolves my problem where MacVim uses a servername of "VIM" instead of "GVIM".

### DIFF
--- a/plugins/vim-interaction/vim-interaction.plugin.zsh
+++ b/plugins/vim-interaction/vim-interaction.plugin.zsh
@@ -22,7 +22,8 @@ EOH
   local cmd=""
   local before="<esc>"
   local after=""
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

LunarWatcher (approved):

mcornella:
Thanks and apologies for the delay!

mcornella:
Can you take a look at #10209 (run `omz pr test 10209`). I've refactored the code and added logic to open gvim if no instance is running, please let me know if everything works on your end.

PR #9107, opened 2020-07-12
Suppress the problematic trap output in check_for_upgrade.sh

### LINKED ISSUE #9109: Upgrade fails to delete lock file with rm alias
On Mac, I have `alias rm="rmtrash"` in my `.zprofile`. Sometimes when I start a shell and ohmyzsh tries to upgrade it prints an error like this:

    rmtrash: illegal option -- r
    USAGE: rmtrash [-h] [-v] [-u USERNAME] FILENAME

This is because it runs this command:

    rm -rf ~/.oh-my-zsh/log/update.lock

It should do something more robust than using `rm` directly.

**To Reproduce**

1. Add `alias rm="rmtrash"` to `.zprofile`
2. Start a sufficient number of shells to have oh my zsh update. Maybe there's a way to force this.

**Expected behavior**
I believe there is a way to ignore aliases when running a command, something like `\rm`. Alternatively it should use a proper programming language to do updates.

 - OS / Distro: MacOS
 - Latest ohmyzsh update?: Yes
 - ZSH Version: 5.7.1
 - Terminal emulator: iTerm2

MYDavoodeh:
The newly added trap, in systems where `rm` is aliased to `rm="rm -v"`,
shows a message stating that "update.lock" has been removed each time `zsh` is called.

## Changes:

- Suppressed the trap `rm` with directing the output to `/dev/null`. (prevents extra alerts in more verbose systems)

## Standards checklist:

Fixes #9109

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -51,7 +51,7 @@ function update_ohmyzsh() {
     #  The return status from the function is handled specially. If it is zero, the signal is
     #  assumed to have been handled, and execution continues normally. Otherwise, the shell
     #  will behave as interrupted except that the return status of the trap is retained.
-    trap "rm -rf '$ZSH/log/update.lock' >/dev/null; return 1" EXIT INT QUIT
+    trap "command rm -rf '$ZSH/log/update.lock'; return 1" EXIT INT QUIT
 
     # Create or update .zsh-update file if missing or malformed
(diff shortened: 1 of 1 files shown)


mcornella:
Hey thanks for pointing this out. A better solution is to prefix the rm call with `command`. Can you change that?

mcornella:
Thanks!

PR #9116, opened 2020-07-16
battery: add acpi requirement to README

samuelterra22:
## Standards checklist:

## Changes:

- Update read from battery plugin. Adds information for installing the acpi tool to display the battery percentage


mcornella:
The `acpi` command is only necessary under Linux-based distros (including WSL). Can you add that bit, and also say instead that `acpi` is needed, and "here's an example of how to install with apt"?

PR #9123, opened 2020-07-18
aws: improve check for aws_completer - unhardcode path

### LINKED ISSUE #9122: aws_completer path assumes global installation
**Describe the bug**
aws completion does not work when installed in non-global location

**To Reproduce**
* enable aws plugin
* type `aws ` *TAB*

**Expected behavior**
autocompletion works

**Additional context**
This seems to be a repeat of #7811 and #7812 from back when it was a shell script command

amirdt22:
Closes #9122

The current check is assuming that the aws_completer is installed
globally, it then runs the command without that path

I have aws_completer in ~/.local/bin/aws_completer

This changes the check to see if the command exists instead of
concerning itself with where it in installed. This test matches the
behavior of just running the command.

## Standards checklist:

## Changes:

- use `command` to check for aws_completer instead of hardcoded path

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -63,7 +63,7 @@ fi
 # Load awscli completions
 
 # AWS CLI v2 comes with its own autocompletion. Check if that is there, otherwise fall back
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks Amir

PR #9134, opened 2020-07-23
Plugin for Tarantool's cartridge framework

pnvasanth:
## Standards checklist:

## Changes:

- New plugin added for Tarantool's [cartridge](https://www.tarantool.io/en/cartridge/) framework which supports commands completion


kbakdev (approved):

robbyrussell:
Thanks for putting this together, and sorry it took so long to get you an answer.

The deciding factor is upstream rather than anything about the code: `tarantool/cartridge-cli` was archived, and its README now opens with a deprecation notice pointing at `tarantool/tt`. That successor ships its own zsh completion, so there's nothing here for us to hand-maintain.

PR #9135, opened 2020-07-24
Update laravel.plugin.zsh

masihjay-z:
add tinker command

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


robbyrussell:

The idea is a good one and a maintainer has already said yes to it... in the `pat` form, on #12847. What we can't take is the bare `tinker` name: every other artisan alias in the plugin is `pa` plus the subcommand initials (`pas`, `pats`, `pam`, `paopc`), and a plain `tinker` in everyone's shell is broad enough to get in the way of other things.

#10797 is open with the `pat` spelling, so that's the thread that'll carry this in.

PR #9140, opened 2020-07-25
git: delete local tracking branches for pruned remote branches

ricokareem:
## Standards checklist:

## Changes:

- Add `gbp` and `gbpf`

## Other comments:

...


andreas-taranetz:
I was in search for this functionality and wondered why it was not yet part of the git plugin.

I too found the approach this PR tries to implement, yet **I think there is a simpler way** to achive the same thing.
`git remote prune origin`

This will remove all local branches which have an upstream branch defined that is no longer existing. Meaning, untracked local branches won't be affected.

**Why should this exist?**
As much as I don't want to overload this plugin with more and more aliases, this specific action is a very common one for developers working in a team, creating multiple branches a day. Usually their branches get deleted after being merged but the local clones clutter up with old references which make the use of autocompletion harder in other git related commands.

Edit:
**Why this is already (kind of) covered**
There already is the `gfa` alias (`git fetch --all --tags --prune`), which also prunes local branches that do not exist on the remote any longer.

Issue #8749, opened 2020-03-16
'safe-paste' plugin breaks 'history-substring-search' plugin


mcurly:
**Bug**
Every time the 'safe-paste' plugin is activated it breaks 'history-substring-search' plugin

**To Reproduce**
1. Enable this plugin 'safe-paste' along with 'history-substring-search'
2. Try to use the normal keybindings of history-substring-search to autocomplete any given command in the zsh-history (up key)
3. Always get latest command in history (which is the previous one which is 'source ~/.zshrc' even if there is no equivalent character/string in the current line that would match that previous one) 

**Expected behavior**
To fetch the latest command with matching character/string.

**Desktop**
 - Operating System: Manjaro-ARM 
 - KDE Plasma Version: 5.18.3
 - KDE Frameworks Version: 5.67.0
 - Qt Version: 5.14.1
 - Kernel Version: 4.19.97-1-MANJARO-ARM
 - OS Type: 64-bit
 - Processors: 4 × 
 - Memory: 3.7 GiB of RAM
 - Latest ohmyzsh from distro.
 - ZSH 5.8
 - Terminal emulator: konsole

muriiloandrade:
Same problem here...

**Desktop**
OS: Arch Linux
Kernel: x86_64 Linux 5.6.15-arch1-1
Shell: zsh 5.8
DE: KDE 5.70.0 / Plasma 5.18.5
CPU: Intel Core i5-3210M @ 4x 3.1GHz [59.0°C]
RAM: 11,6GB

carlocab:
Same problem:
 - OS: 64bit Mac OS X 10.15.6 19G2021
 - Kernel: x86_64 Darwin 19.6.0
 - Shell: zsh 5.8
 - Latest version of OMZ

Issue #7883, opened 2019-05-26
safe-paste breaks zle vi editing


HaleTom:
## Description

Given:
* [`safe-paste`](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/safe-paste) is loaded
* Some text is pasted

Then:
 * It is no longer possible to use the `zle` vi editing commands

## To reproduce

**Common steps to reproduce:**

1. Copy the text `clipboard` into the clipboard
1. Open new terminal
1. Start a new `zsh` with no user system or user configuration:

       VISUAL=vi zsh -f

1. Paste the clipboard
1. Press keys: <kbd>esc</kbd> <kbd>I</kbd> <kbd>:</kbd> <kbd>space</kbd> 

**Default (and expected) behaviour:**

The prompt shows: 

    svelte% : |clipboard

Where `|` is the cursor

**Load `safe-paste`:**

    source <(curl -s 'https://raw.githubusercontent.com/robbyrussell/oh-my-zsh/master/plugins/safe-paste/safe-paste.plugin.zsh')

**Actual behaviour after loading `safe-paste`:**

The command line shows:
    
    svelte% clipboard: |

Where `|` is the cursor

**System:**
 - OS: `Linux svelte 5.0.9-2-MANJARO #1 SMP PREEMPT Sun Apr 21 07:11:08 UTC 2019 x86_64 GNU/Linux`
 - Zsh version: `zsh 5.7.1 (x86_64-pc-linux-gnu)`
 - Terminal emulator: rxvt

**Related Issues:**

* [Safe-paste plugin breaks Home/End keys #5237](https://github.com/robbyrussell/oh-my-zsh/issues/5237)
* [Plugin 'safe-paste' messes with 'history-substring-search' #5149](https://github.com/robbyrussell/oh-my-zsh/issues/5149)

Both of the above claim to be resolved by [Ditch smkx in favor of custom bindkey extension #5113](https://github.com/robbyrussell/oh-my-zsh/pull/5113) which is still unmerged after being raised on  on 24 May 2016.

Other, possible related issues:

* [[Tracking] Terminal and terminfo related issues #4344](https://github.com/robbyrussell/oh-my-zsh/issues/4344) - A meta issue tracking issue
* [The safe-paste plugin does not make pasting safe and breaks other features. #1742](https://github.com/robbyrussell/oh-my-zsh/issues/1742) - On a flaw in the code as written

PR #7887, opened 2019-05-28
safe-paste: refresh plugin (update for zsh 5.1 and vi keymaps)

HaleTom:
### Functionality

* Prefer built-in bracketed paste goodness if zsh version >= 5.1
* Save/restore active keymap before/after paste (resolves #7883)
* Attempt to not clobber `zle_line_{init,finish}`
  * Partially addresses [Plugins breaks if `zle -N zle-line-init` is called #70](https://github.com/zsh-users/zsh-history-substring-search/issues/70)
* Clear paste content queue before first use

### Style

* [Style guide](https://github.com/robbyrussell/oh-my-zsh/wiki/Coding-style-guide)
  * Wrap lines at 80 columns
  * Remove `function` before function definitions
* Make paste keymap range more readable
* Simplify `$TERM` tests
* Ensure all functions have comments
* Prefix all globals with `bracketed`
* Improve punctuation in comments

Fixes #5237
Fixes #7883
Fixes #8749
Fixes #8886

PR #9148, opened 2020-08-01
Add: allow keymap other than emacs after safe-paste

### LINKED ISSUE #8886: safe-paste breaks vi bindings
**Describe the bug**

When using a terminal with `vi` bindings the mode will be set back to `emacs` unexpectedly. [The code mentions the issue](https://github.com/ohmyzsh/ohmyzsh/blob/dd1a72696f162fd5285a104b5f63970deb66e8b0/plugins/safe-paste/safe-paste.plugin.zsh#L35), but when using a lot of plugins and other config files it's not obvious why this is happening.

**To Reproduce**
1. Enable safe-paste
2. Run `bindkey -v`
3. Paste some things
4. Notice that you're no longer in vi mode.

SleepyBag:
Close #8886 #8749 #7883
## Standards checklist:

## Changes:

Currently safe-paste set keymap to `emacs` after paste. It is annoying.
So I add two variables for better control:
 - `KEYMAP_AFTER_PASTE`, which specify a keymap after paste
 - `KEYMAP_VAR_AFTER_PASTE`, which specify a defined variable tracking keymap

### DIFF
--- a/plugins/safe-paste/safe-paste.plugin.zsh
+++ b/plugins/safe-paste/safe-paste.plugin.zsh
@@ -30,9 +30,16 @@ function _start_paste() {
 # command line. this has the nice effect of making the whole paste be
 # a single undo/redo event.
 function _end_paste() {
-#use bindkey -v here with vi mode probably. maybe you want to track
-#if you were in ins or cmd mode and restore the right one.
-  bindkey -e
+  # by default, safe-paste set keymap to emacs after paste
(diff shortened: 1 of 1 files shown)


mcornella:
We can log the `$KEYMAP` before changing to the paste keymap, and then use it to toggle back to it with `bindkey -M $keymap`. I'll push a fix.

mcornella:
Actually, #7887 is much more advanced, I'll close this in favor of that one.

PR #9159, opened 2020-08-06
Ruby gem no doc tweak

Rajpratik71:
Optimization tweak for ruby gems no doc while gems installation using
 "--no-document" . Also removal of depricated "--no-ri" and "--no-rdoc".

 More detail can be found at
 https://guides.rubygems.org/command-reference/#installupdate-options-2

 Signed-off-by: Pratik raj <rajpratik71@gmail.com>


mcornella:
[You can set that globally for all `gem install`s](https://stackoverflow.com/a/1386014/5798232). I'm not sure of making it a default for all users of the `ruby` plugin since I'm not a member of the Ruby community.

Rajpratik71:
> [You can set that globally for all `gem install`s](https://stackoverflow.com/a/1386014/5798232). I'm not sure of making it a default for all users of the `ruby` plugin since I'm not a member of the Ruby community.

ok

PR #9167, opened 2020-08-10
possible avoid 3 times of git execution per cd

fr233:
## Standards checklist:

## Changes:
- possible avoid 3 times of git execution per cd

## Other comments:
- Now, every time we cd, we don't have to fork and execute git for 3 times, unless current directory seems to be within a git repository. The probe sequence is identical to git (as show by bpftrace), so we won't get false negatives.
- The command use for inspect: 
   - sudo bpftrace -e 'tracepoint:syscalls:sys_enter_newstat /comm=="git"/ {printf("%s\n", str(args->filename))}'
   - sudo bpftrace -e 'tracepoint:syscalls:sys_enter_newstat /comm=="zsh"/ {printf("%s\n", str(args->filename))}'

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -11,6 +11,17 @@ function __git_prompt_git() {
 
 # Outputs current branch info in prompt format
 function git_prompt_info() {
+  local cwd=$(pwd)
+  while [[ "$cwd" != '' ]];do
+    if [[ -f "$cwd/.git/HEAD" ||  -f "$cwd/HEAD" ]];then
+      break
+    fi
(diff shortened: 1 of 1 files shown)


kbakdev (approved):

robbyrussell:

You were right at the time... at the commit you branched from, the prompt really did fork `git` three times outside a repository. It doesn't any more: an early `git rev-parse --git-dir` guard landed since, which cuts that to a single call, so the saving you were after is mostly already there.

Closing because the cost is gone rather than because the idea was wrong. Thanks for digging into it.

PR #9171, opened 2020-08-12
added new aliases

raafideveloper:
## Standards checklist:

## Changes:

- [...]Just Added New Aliases

## Other comments:

...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -63,6 +63,11 @@ alias gbsg='git bisect good'
 alias gbsr='git bisect reset'
 alias gbss='git bisect start'
 
+alias gullom= 'git pull origin master'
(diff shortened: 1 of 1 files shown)


RaafiHaswoto:
Just added some new aliases. 
Thankyou

HackAttack:
I’m not sure these warrant their own aliases given you can already do `ggpull` and `ggpush`.

robbyrussell:

`ggpull` and `ggpush` already do this, and they work against whatever branch you're actually on rather than a hardcoded one... that's the part that matters, since not every repo uses `master` or `develop`.

One thing worth knowing: there's a stray space after the `=` in each line, so zsh fails to parse them and leaves the aliases empty. They wouldn't have worked even if we'd merged this.

PR #9174, opened 2020-08-13
velero autocompletion plugin

s1msn:
## Standards checklist:

## Changes:

- added velero autocompletion plugin to the plugins/ directory

## Other comments:

...

### DIFF
--- a/plugins/velero/README.md
+++ b/plugins/velero/README.md
@@ -1,0 +1,22 @@
+# velero-zsh-autocomplete-plugin
+Provides [Velero](https://velero.io/) CLI autocompletion for ZSH
+
+To use it, copy the `velero.plugin.zsh` file to your zsh plugin directory and add `velero` to the plugins array of your zshrc file:
+
+```bash
+
+plugins=(
+
+        ...
+
+        velero
+
+        ...
+
+        )
+
+```
+## Contributors
+
++ [s1msn](https://github.com/s1msn) - Plugin Author
++ Based on [kevinkirkup](https://github.com/kevinkirkup)'s oc autocomplete plugin.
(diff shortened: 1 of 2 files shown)


kbakdev (approved):

marioaparcero:
I need add velero autocompletion plugin to the plugins/ directory :package:

marioaparcero:
Not working for me https://github.com/ohmyzsh/ohmyzsh/pull/9649/commits/ba77fa538f2949ec97946fd34165d17bf11f824d

robbyrussell:
Thanks for this, and sorry for the very long delay in replying. The approach is right... sourcing velero's own generator means it can't drift the way a hand-written completion file would.

One thing needs updating before it can go in. `source <(velero completion zsh)` runs velero on every shell start, and we've since standardised on caching the generated file and regenerating it in the background. `plugins/helm/helm.plugin.zsh` is the model:

```zsh
if (( ! $+commands[velero] )); then
  return
fi

if [[ ! -f "$ZSH_CACHE_DIR/completions/_velero" ]]; then
  velero completion zsh | tee "$ZSH_CACHE_DIR/completions/_velero" >/dev/null
  source "$ZSH_CACHE_DIR/completions/_velero"
else
  source "$ZSH_CACHE_DIR/completions/_velero"
  zmodload -F zsh/files b:zf_mv
  () {
    local TMPPREFIX="$ZSH_CACHE_DIR/completions/_velero"
    zf_mv -f -- =( velero completion zsh ) "$TMPPREFIX"
  } &|
fi
```

The README could also use a trim... the heading and the spacing in the plugins block don't match the other plugin READMEs. Keep the credit to @kevinkirkup, that's good practice.

If you'd rather not pick this back up after this long, that's completely fine... say so, or say nothing, and I'll do it with you credited.

PR #9188, opened 2020-08-18
feat(lib): show upstream branch in `git_prompt_info`

brunoborges:
This is a redo of #7073 

To try this PR, perform these commands:

```zsh
~ $ cd ~/.oh-my-zsh

~/.oh-my-zsh $ git remote add brunoborges https://github.com/brunoborges/oh-my-zsh.git

~/.oh-my-zsh $ git fetch brunoborges

~/.oh-my-zsh $ git checkout brunoborges/no-remote-git
```

This will put your `.oh-my-zsh` folder into detached mode. Once you do that, start a new terminal window and access folders with a git repository to check the change. Ensure the theme you use displays `git_prompt_info`. 

**Important**: You must also change the following setting:

```sh
$ git config --global --add oh-my-zsh.info-mode enhanced
```

To restore, just checkout master/main, or whatever is the default branch used by oh-my-zsh on your system.


juliofalbo:
Really useful! Thanks!!!

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
 # Outputs current branch info in prompt format
 function git_prompt_info() {
   local ref
-  if [[ "$(__git_prompt_git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
-    ref=$(__git_prompt_git symbolic-ref HEAD 2> /dev/null) || \
-    ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null) || return 0
-    echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
+  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
```
mcornella, line 15: This should use `__git_prompt_git` instead of `command git`, which adds safety features (see #7804).

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
...
+    branches=$(command git branch --list -a 2> /dev/null)
+    if [[ "$branches" == "" && "$?" == "0" ]]; then
+        # Look for initialized folders
+        info="<initialized>"
```
mcornella, line 19: Does this work the same way as before? Now it'll just say `<initialized>`, before it says whatever the `git symbolic-ref HEAD` command returns, which is normally `master`.

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
...
+        # Branch name in the form "<upstream branch name>/<brach name>"
+        ref=$(command git for-each-ref --format='%(upstream:short)' ${branch} 2> /dev/null) || return 0
+        info=$(command echo "${ref#refs/heads/}")
+        if [[ "${info}" == "" ]]; then
+            # Handle for no remote branch
+            info=$(command echo "<no-remote>/$(command git rev-parse --abbrev-ref HEAD)" 2> /dev/null) || return 0
+        fi
+    elif [[ "$(command git tag --points-at 2>/dev/null)" != "" ]]; then
+        # Detached at a tag
+        info=$(command git tag --points-at 2>/dev/null)
```
mcornella, line 30: This makes two calls unnecessarily. Also, that command may return more than one tag, so we should use `head -1` or use the `git describe --tags --exact-match` command, which always return only one tag.

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
...
-    echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
+  if [[ "$(command git config --get oh-my-zsh.hide-status 2>/dev/null)" != "1" ]]; then
+    branches=$(command git branch --list -a 2> /dev/null)
+    if [[ "$branches" == "" && "$?" == "0" ]]; then
+        # Look for initialized folders
+        info="<initialized>"
+    elif branch=$(command git symbolic-ref HEAD 2> /dev/null); then
+        # Branch name in the form "<upstream branch name>/<brach name>"
+        ref=$(command git for-each-ref --format='%(upstream:short)' ${branch} 2> /dev/null) || return 0
+        info=$(command echo "${ref#refs/heads/}")
```
mcornella, line 23: You don't need to `echo` this, you can just assign it directly to `$info`:
```
info="${ref#refs/heads/}"
```

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
...
+        if [[ "${info}" == "" ]]; then
+            # Handle for no remote branch
+            info=$(command echo "<no-remote>/$(command git rev-parse --abbrev-ref HEAD)" 2> /dev/null) || return 0
```
mcornella, line 26: Same as before, there's no use using `echo`. Also, I bet there's a way to use only one git call to get this information either with or without a remote branch. `git` calls are expensive!

lib/git.zsh
```diff
@@ -12,10 +12,27 @@ function __git_prompt_git() {
 # Outputs current branch info in prompt format
 function git_prompt_info() {
   local ref
```
mcornella, line 14: Make all the variables you use `local`

mcornella (changes requested):
There are some things that need revisiting. Congrats on getting support for this though!

NOTE: if there is a remote branch, I would only show it the user wants it, for backwards compatibility with the old behavior. That means looking for a variable like `ZSH_GIT_PROMPT_SHOW_REMOTE_BRANCH=true`, or something like that.

lib/git.zsh
brunoborges: Done

lib/git.zsh
brunoborges: I added back the branch name to be displayed even on `<initialized>` state.

lib/git.zsh
brunoborges: @mcornella  I couldn't reproduce a case where more than one tag is returned. Could you provide an example?

brunoborges:
I've also added a new flag `oh-my-zsh.classic-info` so users can go back to showing only the branch.

brunoborges:
Finally, I noticed that a lot of the cost of using git statuses and infos on PROMPT comes from the cost of processing `git_prompt_status` and `parse_git_dirty`. It is important to note that `git_prompt_info` calls `parse_git_dirty`.

On a large git repository such as the OpenJDK, this is how much it costs:

```
work/openjdk/jdk  origin/master ✗
▶ time ( git_prompt_status )
%{%}◒
( git_prompt_status; )  0.22s user 2.44s system 284% cpu 0.936 total

work/openjdk/jdk  origin/master ✗
▶ time ( parse_git_dirty )
 %{%}✗%{%}
( parse_git_dirty; )  0.22s user 2.51s system 284% cpu 0.959 total

work/openjdk/jdk  origin/master ✗
▶ time ( git_prompt_info )
%{%}origin/master %{%}✗%{%}%{%}
( git_prompt_info; )  0.24s user 2.60s system 227% cpu 1.246 total
```

Themes like `Avit` show a lot of git information by default, and because of this, it gets extremely slow to get a shell ready.

The enhanced `git_prompt_info` proposed here does not add any extra cost. The real culprit is `parse_git_dirty` as you can see below:

```
work/openjdk/jdk  origin/master ✔
▶ git config --add oh-my-zsh.hide-dirty 0

work/openjdk/jdk  origin/master ✗
▶ time ( git_prompt_info )
%{%}origin/master %{%}✗%{%}%{%}
( git_prompt_info; )  0.23s user 2.63s system 223% cpu 1.280 total

work/openjdk/jdk  origin/master ✗
▶ git config --add oh-my-zsh.hide-dirty 1

work/openjdk/jdk  origin/master ✔
▶ time ( git_prompt_info )
%{%}origin/master %{%}✔%{%}%{%}
( git_prompt_info; )  0.02s user 0.04s system 77% cpu 0.072 total
```

brunoborges:
@mcornella marc, are these changes enough?

brunoborges:
@megbxk79 want to review this one instead?

megbxk79 (approved):

lib/git.zsh
brunoborges: See my comment below: https://github.com/ohmyzsh/ohmyzsh/pull/9188#issuecomment-678034444

brunoborges:
@mcornella marc, any chance for a second look into this?

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+    return 0
+  fi
+
+  # If we are on a folder not tracked by git, get out.
+  if [[ "$(__git_prompt_git rev-parse --git-dir 2> /dev/null)" == "" ]]; then
+    return 0
+  fi
+
+  # If oh-my-zsh.classic-info is set, use classic info output of branch only
+  if [[ "$(__git_prompt_git config --get oh-my-zsh.classic-info 2>/dev/null)" == "1" ]]; then
```
mcornella, line 30: I don't like that you have to opt-in to the old behavior, it should be other way around. Also, putting this under another git config option isn't ideal, I'd prefer a `ZSH_GIT_PROMPT_*` variable.

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+  local info
+  local deatchedtag
+
+  # If oh-my-zsh.hide-status equals to '1', get out.
+  if [[ "$(__git_prompt_git config --get oh-my-zsh.hide-status 2>/dev/null)" == "1" ]]; then
+    return 0
+  fi
+
+  # If we are on a folder not tracked by git, get out.
+  if [[ "$(__git_prompt_git rev-parse --git-dir 2> /dev/null)" == "" ]]; then
```
mcornella, line 25: Don't check the output, check the return code:
```suggestion
  if ! __git_prompt_git rev-parse --git-dir 2> /dev/null; then
```

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
     echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
+    return 0
+  fi
+
+  branches=$(__git_prompt_git branch --list -a 2> /dev/null)
+  if [[ "${branches}" == "" && "$?" == "0" ]]; then
+      # Look for initialized repos that have zero commits
+      ref=$(__git_prompt_git symbolic-ref HEAD 2> /dev/null) || \
+      ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null)
+      info="<initialized>/${ref#refs/heads/}"
```
mcornella, line 42: I dislike that this adds `<initialized>` to the prompt. This is unnecessary and verbose. Just print the branch.

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+  # If oh-my-zsh.classic-info is set, use classic info output of branch only
+  if [[ "$(__git_prompt_git config --get oh-my-zsh.classic-info 2>/dev/null)" == "1" ]]; then
     ref=$(__git_prompt_git symbolic-ref HEAD 2> /dev/null) || \
     ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null) || return 0
     echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
+    return 0
+  fi
+
+  branches=$(__git_prompt_git branch --list -a 2> /dev/null)
+  if [[ "${branches}" == "" && "$?" == "0" ]]; then
```
mcornella, line 38: You're asking git for a list of all branches only for this condition. There must be a better way.

lib/git.zsh
mcornella, line 42: This is essentially the same as the "`classic-info`" version. Don't repeat yourself.

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+      # Look for initialized repos that have zero commits
+      ref=$(__git_prompt_git symbolic-ref HEAD 2> /dev/null) || \
+      ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null)
+      info="<initialized>/${ref#refs/heads/}"
+  elif branch=$(__git_prompt_git symbolic-ref HEAD 2> /dev/null); then
```
mcornella, line 43: Running the same command again. DRY

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+      # Show upstream remote and current branch
+      ref=$(__git_prompt_git for-each-ref --format='%(upstream:short)' ${branch} 2> /dev/null) || return 0
+      info="${ref#refs/heads/}"
+      if [[ "${info}" == "" ]]; then
+          # Handle for no remote branch
+          info="<no-remote>/$(__git_prompt_git rev-parse --abbrev-ref HEAD)" || return 0
```
mcornella, line 49: You don't need to specify there's no remote, just print the information you have. If there's no remote displayed, it's already known that there's no remote.

lib/git.zsh
```diff
@@ -12,11 +12,53 @@ function __git_prompt_git() {
...
+      fi
+  else
+      deatchedtag=$(__git_prompt_git tag --points-at 2>/dev/null)
```
mcornella, line 52: Typo

mcornella (changes requested):
There are many opportunities for improvement. I think you need to rethink your approach and simplify the algorithm. You're also trying to do too much. Why not use the old version and just ask for the remote? That seems like a more straightforward solution. Thanks!

lib/git.zsh
brunoborges: Fixed on 8a749810.

lib/git.zsh
brunoborges: It will display `<initialized>` only when in a repository without any commit. Literally, just got initialized. It helps differentiate against other repos that already have something in.

It will display `<no-remote>` only when in a repository *already* initialized (has at least one commit), but no remote.

In my experience with this enhanced git prompt, I have found these quite useful to help differentiate between multiple local folders/copies of the same project.

lib/git.zsh
brunoborges: There are two things to discuss here:

1. Opt-out: I think you are right that it is better to have this change as `opt-in`. Well documented and advertised in the release notes, so people can give it a try. Later on this could change to an `opt-out`, if the feedback is positive enough.

2. git config: The reason I went with a `git config` option is to allow the user to switch this feature _per repository_ if they want. You can have it on globally, but off on specific repositories. I found this useful for some _very large_ repositories (e.g. github.com/openjdk/jdk).

All said, I believe we can go with `opt-in` and the git config option. What do you think @mcornella ?

lib/git.zsh
brunoborges: The idea here is to really differentiate once people enable the enhanced git prompt info.

See this: https://github.com/ohmyzsh/ohmyzsh/pull/9188/commits/b787c278b482220854500167301b40693238d929#diff-b67797c4a19e91a6eebdd5b97df83dff900a5ea4cacd63d1df8abbe86839c663L41-L46

lib/git.zsh
brunoborges: Not really running it again. Yes, repeated, but not sure this needs to be externalized to its own function.

lib/git.zsh
brunoborges: So far I have not found a better way. Doesn't cost anything, though.

lib/git.zsh
brunoborges: I moved the classic info to a separate function, but the idea now is that people have to opt in to the enhanced mode by flipping the info-mode config flag.

mcornella: Why would you need to say it's initialized? You already know it is, you just created the repo (?) Again, it's unnecessary and verbose (prompt line space is sacred, this adds 13 characters.

lib/git.zsh
mcornella: Yes it does.

mcornella:
As I said previously, you wanted to do too much that was outside of the scope of the PR (at least by looking at the PR title). The following is a version that uses the old code, refactored, and without creating more functions:
```zsh
function git_prompt_info() {
  [[ "$(__git_prompt_git config --get oh-my-zsh.hide-status 2>/dev/null)" != 1 ]] || return 0

  local ref
  ref=$(__git_prompt_git symbolic-ref --short HEAD 2> /dev/null) \
  || ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null) \
  || return 0

  local upstream
  if (( ${+ZSH_THEME_GIT_SHOW_UPSTREAM} )); then
    upstream=$(__git_prompt_git rev-parse --abbrev-ref --symbolic-full-name "@{upstream}" 2>/dev/null) \
    && upstream="->${upstream}"
  fi

  echo "${ZSH_THEME_GIT_PROMPT_PREFIX}${ref}${upstream}$(parse_git_dirty)${ZSH_THEME_GIT_PROMPT_SUFFIX}"
}
```
As you can see, here I'm using the variable setting approach (`ZSH_THEME_GIT_SHOW_UPSTREAM=1`). There's really no need to toggle it on and off for different repositories, but if we were to use the git-config approach it would be a one line change.

Also this uses `->` to indicate that the branch is tracking the upstream branch, but it could be changed to another symbol (`@` was another candidate I thought off).

This is what it looks like (first, without an upstream branch; second, with one):

image

brunoborges:
Marc, your take here nails it. Yes, it was doing more than it was "supposed to", but that's because when I got this started, I got some people asking for more information (e.g. `<detached>/` or `<initialized>/`).

One thing that your code made me realize is that the original code did not account properly for local branches that could have a different name than the one on upstream, as this is obviously possible with git.

By displaying `master -> origin/master`, it is clear whether there is a remote upstream and what branch is that, even if the name is different.

One change I am submitting is regarding the `oh-my-zsh.hide-status` check. This flag should be limited to the `git_prompt_status` function. Because of that, I changed to `oh-my-zsh.hide-info`. This is because `git_prompt_status` can take a long time to process compared to `git_prompt_info` and depending on the repository size. Therefore, we should consider letting users control which of these two they want to see/hide.

```zsh
work/java-engineering/openjdk-jdk  main -> origin/main ✔       1d5h  java:15.0
▶ time ( git_prompt_info )
main -> origin/main ✔
( git_prompt_info; )  0.02s user 0.03s system 79% cpu 0.065 total

work/java-engineering/openjdk-jdk  main -> origin/main ✔       1d5h  java:15.0
▶ time ( git_prompt_status )
( git_prompt_status; )  0.23s user 2.61s system 226% cpu 1.251 total
```

I also added a quicker check for whether the shell is in a git repository at all, and exit early if not.

I think this is a solid change, and truly appreciate you taking the time to hack with it a bit.

lib/git.zsh
```diff
@@ -9,14 +9,27 @@ function __git_prompt_git() {
   GIT_OPTIONAL_LOCKS=0 command git "$@"
 }
 
-# Outputs current branch info in prompt format
 function git_prompt_info() {
+  # If we are on a folder not tracked by git, get out.
+  # Otherwise, check for hide-info at global and local repository level
+  if ! $(__git_prompt_git rev-parse --git-dir &> /dev/null) \
+     || [[ "$(__git_prompt_git config --get oh-my-zsh.hide-info 2>/dev/null)" == 1 ]]; then
```
mcornella, line 16: This is a breaking change; in OMZ the concept of `git status` is equivalent to `git info`, so the current config key is already OK.

lib/git.zsh
```diff
@@ -9,14 +9,27 @@ function __git_prompt_git() {
   GIT_OPTIONAL_LOCKS=0 command git "$@"
 }
 
-# Outputs current branch info in prompt format
 function git_prompt_info() {
+  # If we are on a folder not tracked by git, get out.
+  # Otherwise, check for hide-info at global and local repository level
+  if ! $(__git_prompt_git rev-parse --git-dir &> /dev/null) \
```
mcornella, line 15: You can remove the subshell syntax:
```suggestion
  if ! __git_prompt_git rev-parse --git-dir &> /dev/null \
```

lib/git.zsh
```diff
@@ -9,14 +9,27 @@ function __git_prompt_git() {
...
-    echo "$ZSH_THEME_GIT_PROMPT_PREFIX${ref#refs/heads/}$(parse_git_dirty)$ZSH_THEME_GIT_PROMPT_SUFFIX"
+  ref=$(__git_prompt_git symbolic-ref --short HEAD 2> /dev/null) \
+  || ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null) \
+  || return 0
+ 
+  # Use global ZSH_THEME_GIT_SHOW_UPSTREAM=1 for including upstream remote info
+  local upstream
+  if (( ${+ZSH_THEME_GIT_SHOW_UPSTREAM} )); then
+    upstream=$(__git_prompt_git rev-parse --abbrev-ref --symbolic-full-name "@{upstream}" 2>/dev/null) \
+    && upstream=" -> ${upstream}"
```
mcornella, line 29: Is the arrow ok? I used it as a first draft but maybe there's a better symbol? Maybe we should ask the users who +1'd the PR.

mcornella (commented):
Alright, this is almost ready. Thanks for applying the suggestions!

lib/git.zsh
brunoborges: I think the arrow is OK, but frankly, it duplicates content (the branch appears twice), and it is something you suggested earlier on that it should not repeat information.

Maybe go back to the `[$remote/]$branch` format?

lib/git.zsh
brunoborges: Yes @mcornella  it is a breaking change, but please keep in mind that  `git_prompt_info` and `git_prompt_status` display different data at different places in the prompt, and these two functions take considerably different amount of time to process depending no the size of the repository.

On the OpenJDK git repository:

- git_prompt_info takes 0.02s
- git_prompt_status takes 2.63s

So I think it is really worthy the breaking change.

lib/git.zsh
brunoborges: The upside is that the local branch may have a different name, as we also discussed before.

mcornella: > The upside is that the local branch may have a different name, as we also discussed before.

Also a branch may not be at the same commit as the upstream branch, for example:
```
$ git log --all | head -3
* 452586e7 (upstream/master, origin/master, origin/HEAD) chore: add Josh Medeski as `wp-cli` plugin maintainer
* 865d2d8a refactor(archlinux): prevent leaking local variables (#9476)
* e48be7df (master) feat(wp-cli): add aliases for DB management (#9469)
$ ZSH_THEME_GIT_SHOW_UPSTREAM=1 git_prompt_info
git:(master -> upstream/master)
```
If we only show the upstream branch, that could be misleading.

lib/git.zsh
mcornella: Alright, I can see it being merged.

mcornella:
Users who +1'd the PR: @AndreyGFranca @ademarizu @bmuskalla @deepu105 @denuno @fabiohenriquebayma @jpe42 @juliofalbo @lahiguera @marcphilipp @msmagnanijr @omeuerodrigofreitas @plateauu @sebastienblanc @smaziano @spencergibb @sureshg @utix @vegegoku

This is the current status: you need to set `ZSH_THEME_GIT_SHOW_UPSTREAM=1` to enable showing the remote branch, and it's displayed like so `<local-branch> -> <remote-branch>`:
image

Any thoughts?

lib/git.zsh
brunoborges: Cool. Let's go with `$branch [-> $remote/$branch]` then 👍

utix:
If the upstream is origin and the targeted branch is the same what about not showing it.
This behaviour can be activate via a flag.
For basic repo this will avoid too heavy prompt, my need is mainly to know when I am targeting something else than origin.

saadanerdetbare:
> Any thoughts?

+1
Showing ` master -> origin/master `  is the way to do it

juliofalbo:
+1 I agree with @saadanerdetbare

mcornella (approved):

brunoborges:
Oops! 😂

brunoborges:
@mcornella good to go?

mcornella:
Thanks @brunoborges

PR #9189, opened 2020-08-20
[git autocomplete] Git version is taken from system, instead of hard-coded

Gvald:
## Standards checklist:

## Changes:

- Git autocomplete was missing some commands of new version. Quick check reveals that autocomplete uses specific version of git.
Changed the git `version` variable to take it from the installed version, and autocomplete is working as expected now.
- I'm aware of #7918, but no reason is given to *not* use the installed version.

### DIFF
--- a/plugins/gitfast/update
+++ b/plugins/gitfast/update
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 url="https://git.kernel.org/pub/scm/git/git.git/plain/contrib/completion"
-version="2.16.0"
+version=$(git --version | awk '{print $3}')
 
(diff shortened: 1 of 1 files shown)


mcornella:
cc @felipec

mcornella:
On further review, this isn't the intention of the update script, which is to always use the newest possible version *without breaking backwards compatibility*, so it should be a manual process.

felipec:
@mcornella That's right. It should be manually updated to the latest safe version.

It was also the intention that if the `updates.patch` failed, that should be updated too. But in the latest pull request that patch isn't needed anymore.

PR #9206, opened 2020-08-25
pipenv: change directories after exiting shell

tschaub:
## Standards checklist:

## Changes:

- Adds a second `chpwd` hook to the pipenv plugin to change directories after exiting the pipenv shell
- Makes it so when you change from one pipenv to another, you exit the first and activate the second

## Other comments:

This uses the filesystem to keep track of the intended target directory when a `cd` results in exiting the pipenv shell.

See #6679 and #6698 where users noted that attempting to change directories out of a pipenv directory exits the shell but does not succeed in changing the directory.

I've confirmed this works in the following scenarios:
 * Changing from one normal directory (not containing a `Pipfile`) to another; nothing happens
 * Changing from a normal directory to a directory containing a `Pipfile`; pipenv shell launched
 * Changing from a `Pipfile`-containing directory to a normal directory; pipenv shell exits, change to target directory
 * Changing from one `Pipfile`-containing directory to a different `Pipfile`-containing directory; fist shell exits, change to target directory, pipenv shell launched for new directory

cc @gangleri

### DIFF
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -4,14 +4,16 @@ _pipenv() {
 }
 compdef _pipenv pipenv
 
+POST_EXIT_DIR="${ZSH_CACHE_DIR:-$ZSH/cache}/pipenv-post-exit-dir"
+
 # Automatic pipenv shell activation/deactivation
 _togglePipenvShell() {
-  # deactivate shell if Pipfile doesn't exist and not in a subdir
-  if [[ ! -f "$PWD/Pipfile" ]]; then
-    if [[ "$PIPENV_ACTIVE" == 1 ]]; then
-      if [[ "$PWD" != "$pipfile_dir"* ]]; then
-        exit
-      fi
+  # deactivate shell if active and and not in a subdir of original pipenv
+  if [[ "$PIPENV_ACTIVE" == 1 ]]; then
+    if [[ "$PWD" != "$pipfile_dir"* ]]; then
+      echo "$PWD" > "$POST_EXIT_DIR"
+      unset pipfile_dir
+      exit
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

austinnebel:
This works well, however, I noticed that if the shell is started within a pipenv directory, the first `cd` out of the directory does not work, but all of them after that work flawlessly. This may be specific to my setup which is Windows 10 + cygwin. I designed my own code that fixed that issue on my system if anyone else is interested:

```
# Pipenv completion
_pipenv() {
  eval $(env COMMANDLINE="${words[1,$CURRENT]}" _PIPENV_COMPLETE=complete-zsh pipenv)
}
compdef _pipenv pipenv

export pipenv_exit_pwd=$ZSH/plugins/pipenv/pwd.txt

# Automatic pipenv shell activation/deactivation
_togglePipenvShell() {
  # deactivate shell if Pipfile doesn't exist and not in a subdir
    if [[ "$PIPENV_ACTIVE" == 1 ]]; then
      if [[ "$PWD" != "$pipfile_dir"* ]]; then
        touch $pipenv_exit_pwd
        echo $PWD | tee $pipenv_exit_pwd > /dev/null
        exit
      fi
    fi

  # activate the shell if Pipfile exists
  if [[ -f "$PWD/Pipfile" ]]; then
     if [[ "$PIPENV_ACTIVE" != 1 ]]; then
      export pipfile_dir="$PWD"
      pipenv shell
      #post-fork operations
      #after pipenv exits, cd to location stored in $pipenv_exit_pwd file and remove it
      dir=`cat $pipenv_exit_pwd`
      cd $dir;
      if [ -f $pipenv_exit_pwd ]; then; rm $pipenv_exit_pwd; fi
    fi
  fi
}
autoload -U add-zsh-hook
add-zsh-hook chpwd _togglePipenvShell
_togglePipenvShell
```

PR #9218, opened 2020-08-31
emacs: add ansi-term directory tracking

IVAN-DPP:
## Standards checklist:

## Changes:
I used the prompt into emacs, with the commands M-x term, or others, when I began to use zsh, the commands as C-x C-f to find files in emacs  doesn't followed the path to file directory of the prompt, but with this change doesn't matter in which directory I stay, always the commands will follow the path of the directory.
## Other comments:

...

### DIFF
--- a/plugins/emacs/emacs.plugin.zsh
+++ b/plugins/emacs/emacs.plugin.zsh
@@ -26,6 +26,11 @@ if "$ZSH/tools/require_tool.sh" emacsclient 24 2>/dev/null ; then
     # create a new X frame
     alias eframe='emacsclient --alternate-editor "" --create-frame'
 
+    # The commands like to C-x C-f follow the directories path of zsh prompt 
+    chpwd() { print -P "\033AnSiTc %d" }
+    print -P "\033AnSiTu %n"
+    print -P "\033AnSiTc %d"
+
 
     # Write to standard output the path to the file
     # opened in the current buffer.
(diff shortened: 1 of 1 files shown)


mcornella:
Can you provide an example, preferrably via [asciinema](https://asciinema.org/)? I couldn't understand your explanation. Also if you can point to documentation to the weird strings printed that would be awesome as well.

IVAN-DPP:
All this was made in emacs, I open the term into Emacs with multi-term package
Ctrl-x Ctrl-f in emacs is used for open files into emacs
[https://www.gnu.org/software/emacs/manual/html_node/emacs/Basic-Files.html](url)

<a href="https://asciinema.org/a/LpPGS2iGaWC58PSSnGVKwzWb9" target="_blank"></a>

You can see that when I put Ctrl-x Ctrl-f in the bottom of the screen appears an other screen with directories, but when I move to other dirs, the comand Ctrl-x Ctrl-f doesn't follow the path of the of zsh prompt  

<a href="https://asciinema.org/a/ydx3XUlS2cP2w7bMhEtXXAXYK" target="_blank"></a>

But with the code, the command Ctrl-x Ctrl-f follow the path of zsh prompt

I'll hope that I've be clear

mcornella:
You're explanation was great thanks! I changed the code a bit and added comments, can you check that it still works?

IVAN-DPP:
I don't know if the line 36 is necessary (print -P "\033AnSiTh %m") because this line call a connection with an own localhost with ssh, without this line emacs work fine

mcornella:
Ok I thought that just displayed the hostname. I've removed it so this is effectively the same you posted with comments. Thanks for your contribution!

PR #9223, opened 2020-08-31
dotenv plugin: Hide "[a]lways" and "n[e]ver" when the list files do not exist

pedromss:
# Plugin: dotenv

## Standards checklist:

## Changes:

- Hide "[a]lways" and "n[e]ver" when the list files do not exist
- Updates the readme

## Other comments:

So I had my `$ZSH` and `$ZSH_CACHE_DIR` undefined, and I'm perfectly happy with that.
Now the `dotenv` plugin assumes that a `touch $ZSH_DOTENV_[DIS]ALLOWED_LIST` will create the files. But if the env vars I mentioned are undefined the touch command will assume root directory and in my case it won't have permissions for it.
This will print:

```
touch: /cache/dotenv-allowed.list: No such file or directory
touch: /cache/dotenv-disallowed.list: No such file or directory
dotenv: found '.env' file. Source it? ([Y]es/[n]o/[a]lways/n[e]ver) e
source_env:24: no such file or directory: /cache/dotenv-disallowed.list
```

And since it doesn't record this failure, this will happen everytime.

This PR is a less obstrusive way of dealing with this: if the files don't exist don't try to append to them.

For users like me who don't need to keep a state regarding allowed or not, having the prompt every time is perfectly fine as it will.

### DIFF
--- a/plugins/dotenv/README.md
+++ b/plugins/dotenv/README.md
@@ -61,6 +61,11 @@ will be added to an allowed list; if you choose Never, it will be added to a dis
 If a directory is found in either of those lists, the plugin won't ask for confirmation and will
 instead either source the .env file or proceed without action respectively.
 
+If both `$ZSH_CACHE_DIR` and `$ZSH` are undefined the `[a]lways` and `n[e]ver`
+options will not be displayed as attempting to append to a missing file would
+result in an error everytime the plugin ran until you define the variables
+mentioned
+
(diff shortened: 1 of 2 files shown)


mcornella:
`$ZSH` should always be defined, therefore `$ZSH_CACHE_DIR` as well since it's defined from `$ZSH`. The real solution is to define `$ZSH` if it's not defined.

I imagine you don't really want an always / never option anyways, though that being there doesn't really hurt you in any way. I'd advise you to define `$ZSH` regardless, but I'll push a quick fix for cases like yours where it isn't.

PR #9238, opened 2020-09-06
update: return git error code on failure

### LINKED ISSUE #9237: when upgrade fails, it shouldn’t return `0`
**Describe the bug**
If `omz update` fails and this message appears,
https://github.com/ohmyzsh/ohmyzsh/blob/708ea42384d378343a590fc34a3dee536a1651df/tools/upgrade.sh#L71 
Oh My Zsh should return a non-zero exit status, but it doesn’t.

**To Reproduce**
1. turn off the internet
1. run `omz update`
1. run `echo $?` to see the last exit status

**Expected behavior**
when update fails, it should return a non-`0` exit status

 · OS / Distro: macOS
 · Latest ohmyzsh update?: Yes
 · ZSH Version: 5.8
 · Terminal emulator: iTerm2

LucasLarson:
##  Standards checklist:

## Changes:
Add `exit 1` to return an error status if `omz update` fails. This replaces an implicit `0`.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -79,4 +79,4 @@ case "$resetAutoStash" in
 esac
 
 # Exit with `1` if the update failed
-exit $status && unset status
+exit $status
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -68,10 +68,8 @@ then
   printf "${BLUE}${BOLD}%s ${UNDER}%s${RESET}\n" "Want to get involved in the community? Join our Discord:" "https://discord.gg/ohmyzsh"
   printf "${BLUE}${BOLD}%s ${UNDER}%s${RESET}\n" "Get your Oh My Zsh swag at:" "https://shop.planetargon.com/collections/oh-my-zsh"
 else
+  status=$?
   printf "${RED}%s${RESET}\n" 'There was an error updating. Try again later?'
-
-  # Prepare to return an error exit code
-  status=1
 fi
 
 # Unset git-config values set just for the upgrade
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -69,6 +69,7 @@ then
   printf "${BLUE}${BOLD}%s ${UNDER}%s${RESET}\n" "Get your Oh My Zsh swag at:" "https://shop.planetargon.com/collections/oh-my-zsh"
 else
   printf "${RED}%s${RESET}\n" 'There was an error updating. Try again later?'
+  exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


tools/upgrade.sh
```diff
@@ -69,6 +69,7 @@ then
   printf "${BLUE}${BOLD}%s ${UNDER}%s${RESET}\n" "Get your Oh My Zsh swag at:" "https://shop.planetargon.com/collections/oh-my-zsh"
 else
   printf "${RED}%s${RESET}\n" 'There was an error updating. Try again later?'
+  exit 1
 fi
 
 # Unset git-config values set just for the upgrade
```
mcornella, line 75: These lines below should still be called, so the exit condition would need to go after that.

LucasLarson: You’re right. I updated the PR so that we can still undo the temporary `autoStash` settings, but return the proper exit status:
https://github.com/ohmyzsh/ohmyzsh/blob/0a3bb71a36df6da563a4d1deaee437fa2a28e490/tools/upgrade.sh#L73-L84

tools/upgrade.sh
```diff
@@ -69,10 +69,16 @@ then
...
 case "$resetAutoStash" in
   "") git config --unset rebase.autoStash ;;
   *) git config rebase.autoStash "$resetAutoStash" ;;
 esac
+
+# Exit with `1` if the update failed
+exit $status && unset status
```
mcornella, line 84: The `unset` command will never fire after running `exit` (try it), and since this script is called with sh anyways it's not necessary to "clean up" afterwards.

tools/upgrade.sh
```diff
@@ -69,10 +69,16 @@ then
   printf "${BLUE}${BOLD}%s ${UNDER}%s${RESET}\n" "Get your Oh My Zsh swag at:" "https://shop.planetargon.com/collections/oh-my-zsh"
 else
   printf "${RED}%s${RESET}\n" 'There was an error updating. Try again later?'
+
+  # Prepare to return an error exit code
+  status=1
```
mcornella, line 74: We can do one better by saving the `$?` immediately after the else, to capture the error code returned by the git command. So:
```suggestion
  status=$?
  printf "${RED}%s${RESET}\n" 'There was an error updating. Try again later?'
```
Remember to update the comment above the exit below.

mcornella (approved):
Small cosmetic changes required.

tools/upgrade.sh
LucasLarson: This is much more elegant – I love it!

PR #9239, opened 2020-09-06
optimize Git plugin’s `grename` logic by reducing `if` conditions

LucasLarson:
## Standards checklist:

## Changes:

This is an optimization. The function to rename a Git branch using @ujwaldhakal’s `grename` (#8622, e8609b857caf730c626ab27910ece30420bc6693) requires both arguments `$1` and `$2`, but the existence of `$2` occurs only when `$1` exists. Only one check for the existence of `$2` (the new name of the branch) is needed.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -279,7 +279,7 @@ alias gama='git am --abort'
 alias gamscp='git am --show-current-patch'
 
 function grename() {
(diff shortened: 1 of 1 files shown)


ujwaldhakal:
Thanks, it looks good !

mcornella:
This is more frail (`$1` might be empty string) and it's not worth the risk for the null benefits of this optimitzation. Thanks for your contribution though.

PR #9245, opened 2020-09-08
Add a basic plugin for pipx

cristim:
## Standards checklist:

## Changes:

- Add a basic plugin for [pipx](https://github.com/pipxproject/pipx) that merely implements its documented shell completions for ZSH.

### DIFF
--- a/plugins/pipx/README.md
+++ b/plugins/pipx/README.md
@@ -1,0 +1,13 @@
+# pipx
+
+[`pipx`](https://pipxproject.github.io/pipx/) - Install and Run Python Applications in Isolated Environments
+
+To use it, add `pipx` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... pipx)
+```
+
+## Installation
+
+Please find detailed installation guide [`here`](https://pipxproject.github.io/pipx/installation/)
--- a/plugins/pipx/pipx.plugin.zsh
+++ b/plugins/pipx/pipx.plugin.zsh
@@ -3,6 +3,4 @@ if (( ! ${+commands[pipx]} )); then
   return
 fi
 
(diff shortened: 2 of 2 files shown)


kbakdev (approved):

killjoy1221:
This might not work if [pipx-in-pipx](https://pypi.org/project/pipx-in-pipx) is used. In that case, `register-python-argcomplete` won't be on the path, but in the venv containing pipx.

Something like this should be used instead.

```shell
if (( ! ${+commands[pipx]} )); then
  return
fi
if (( ! ${+commands[register-python-argcomplete]} )); then
  # pipx might be installed with pipx-in-pipx and is managing itself
  if pipx list --short | grep '^pipx ' > /dev/null; then
    venvs=$(pipx environment --value PIPX_LOCAL_VENVS)
    commands[register-python-argcomplete]="$venvs/pipx/bin/register-python-argcomplete"
  fi
fi
eval "$(register-python-argcomplete pipx)"
```

robbyrussell:

The implementation is the right shape... wrapping pipx's own argcomplete output rather than hand-writing a completion file is exactly how we'd want it done. The reason I'm closing is that there isn't much left for a plugin to add: pipx documents the same `eval "$(register-python-argcomplete pipx)"` line in its own setup guide, so anyone who wants it can add one line to their `.zshrc` without us carrying a file for it.

If you do run it locally, it's worth guarding on `register-python-argcomplete` as well as on `pipx`. Homebrew's pipx doesn't link that binary, so on those installs the current version prints "command not found" on every shell start. `plugins/pipenv/pipenv.plugin.zsh` has the pattern. @mattmess1221 pointed this out on the thread back in 2022 and deserved an answer at the time.

PR #9256, opened 2020-09-14
Add plugin for history tool called hstr

KrzysztofHajdamowicz:
## Standards checklist:

## Changes:

This PR adds a plugin for history helper tool called `hstr` (https://github.com/dvorka/hstr).  
Provided is shortcut alias and key binding.

### DIFF
--- a/plugins/hstr/README.md
+++ b/plugins/hstr/README.md
@@ -1,0 +1,5 @@
+# hstr plugin
+
+This is a helper plugin for https://github.com/dvorka/hstr
+
+Includes necessary aliases and key bindings for hstr in ZSH.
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry for the very long delay in replying. hstr is a good tool and this isn't about that.

The blocker is that `bindkey -s "\C-r"` runs unconditionally, so anyone who enables the plugin without hstr installed loses Ctrl-R entirely... it becomes a key that types a command that isn't there. Taking over Ctrl-R is a big thing for a plugin to do, and it would need to be conditional at minimum.

hstr also ships `hstr --show-zsh-configuration`, which emits this setup itself, so it can be wired up in one line with upstream's own binding.

PR #9257, opened 2020-09-14
Move compinit logic out of core oh-my-zsh.sh into lib so it can be customized using $ZSH_CUSTOM

mattmc3:
## Standards checklist:

## Changes:

- Moves compinit logic into lib allowing advanced users to override compinit logic with standard $ZSH_CUSTOM methods
- Fixes longstanding issue of lib/compfix.zsh being double loaded
- Simplifies oh-my-zsh.sh into core functionality, without changing its logic
- Minor whitespace formatting changes for consistency

## Other comments:
See issue for discussion: https://github.com/ohmyzsh/ohmyzsh/issues/9258


53jk1 (approved):

seivan:
This is actually pretty great and solves a lot of problems, is there any chance we can get this merged?

mattmc3:
It is somewhat of an impactful change to add compinit as another file in lib, so I am also open to an alternative implementation to achieve the same result if there are concerns. One other option that occurs to me is leaving everything in oh-my-zsh.sh where it resides today, but wrapping the compinit logic in a function and only defining that function if it’s not already defined. That would have the upside of still letting a user define the function themselves, and this kind of change would be less impactful to the overall OMZ workings, but the downside that it would be yet another way to customize outside of what $ZSH_CUSTOM already provides. Certainly open to feedback from the devs, but this is a really important missing area of customization that a lot of people concerned about performance would like to see implemented.

mattmc3:
I am going to close this PR in favor of the simpler and lower risk approach of https://github.com/ohmyzsh/ohmyzsh/pull/9537

PR #9265, opened 2020-09-18
Add git mv alias

HackAttack:


sandipsubedi (approved):

plugins/git/README.md
```diff
@@ -108,6 +108,7 @@ plugins=(... git)
 | gmt                  | git mergetool --no-prompt                                                                                                        |
 | gmtvim               | git mergetool --no-prompt --tool=vimdiff                                                                                         |
 | gmum                 | git merge upstream/$(git_main_branch)                                                                                                          |
+| gmv                  | git mv                                                                                                                           |
```
Mr-Ker, line 111: Should this line be moved one line lower to not break the alphabetical order?

Mr-Ker (approved):

robbyrussell:

The name is free and it follows the convention, so this is a judgement call rather than a technical one... and the call is that `git mv` doesn't quite clear the bar. Most people rename with plain `mv` and let git work out what happened, so it isn't reached for often enough to earn a permanent alias in a plugin that nearly every user has enabled. #11462 was declined on the same grounds.

It works well in `$ZSH_CUSTOM`, which is supported and survives updates.

PR #9268, opened 2020-09-19
git: run gfa with --jobs=10 (fetch remotes in parallel)

Reperator:
## Standards checklist:

## Changes:

This PR changes `gfa` to run with up to 10 jobs in parallel using `--jobs=10`. This speeds up fetching from many remotes significantly. I arbitrarily chose 10 jobs; I'm using up to 7 remotes on the projects I work with.

I can't think of any reason not to fetch remotes in parallel besides limiting bandwidth or CPU load. In my opinion the speedup is very much worth it.

I'm happy to add a separate alias (e.g. `gfaj` or `gfap`) for parallel fetching if there's opposition to my PR.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -1,3 +1,7 @@
+# Git version checking
+autoload -Uz is-at-least
+git_version="${(As: :)$(git version 2>/dev/null)[3]}"
+
 #
 # Functions
 #
@@ -105,8 +109,7 @@ compdef _git gdv=git-diff
 
 alias gf='git fetch'
 # --jobs=<n> was added in git 2.8
-autoload -Uz is-at-least
-is-at-least 2.8 "$(git --version 2>/dev/null | awk '{print $3}')" \
+is-at-least 2.8 "$git_version" \
   && alias gfa='git fetch --all --prune --jobs=10' \
   || alias gfa='git fetch --all --prune'
 alias gfo='git fetch origin'
@@ -244,7 +247,7 @@ alias gss='git status -s'
 alias gst='git status'
 
 # use the default stash push on git 2.13 and newer
-is-at-least 2.13 "$(git --version 2>/dev/null | awk '{print $3}')" \
+is-at-least 2.13 "$git_version" \
   && alias gsta='git stash push' \
   || alias gsta='git stash save'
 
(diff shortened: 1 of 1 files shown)


mcornella:
[`git fetch --jobs` was introduced in v2.8.0](https://github.com/git/git/commit/62104ba14af4845c6c1ba1dab05fad4a289d806f), [March 28, 2016](https://github.com/git/git/releases/tag/v2.8.0). There are still older versions in use out there, this change would make them break.

Reperator:
How about this then?
```diff
diff --git a/plugins/git/git.plugin.zsh b/plugins/git/git.plugin.zsh
index 5a9ccff5..910358a1 100644
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -104,7 +104,11 @@ function gdv() { git diff -w "$@" | view - }
 compdef _git gdv=git-diff
 
 alias gf='git fetch'
-alias gfa='git fetch --all --prune'
+# --jobs=<n> was added in git 2.8
+autoload -Uz is-at-least
+is-at-least 2.8 "$(git --version 2>/dev/null | awk '{print $3}')" \
+  && alias gsta='git fetch --all --prune --jobs=10' \
+  || alias gsta='git fetch --all --prune'
 alias gfo='git fetch origin'
 
 alias gfg='git ls-files | grep'
@@ -240,7 +244,6 @@ alias gss='git status -s'
 alias gst='git status'
 
 # use the default stash push on git 2.13 and newer
-autoload -Uz is-at-least
 is-at-least 2.13 "$(git --version 2>/dev/null | awk '{print $3}')" \
   && alias gsta='git stash push' \
   || alias gsta='git stash save'
```

Reperator:
I changed it to my suggestion; opinion?

Reperator:
@mcornella Not sure what `"${(As: :)$(git version 2>/dev/null)[3]}"` does, but on my machine `echo "${(As: :)$(git version 2>/dev/null)[3]}"` outputs `t` and therefore `is-at-least 2.8 $git_version` doesn't succeed.

mcornella:
What about `echo "${${(As: :)$(git version 2>/dev/null)}[3]}"`?

mcornella:
Fixed in c5493877.

Reperator:
Works, thanks!

PR #9272, opened 2020-09-20
gco- and gm- aliases are added for "git checkout -" and  "git merge -" respectively.

raxetul:
## Standards checklist:

## Changes:

- gco- alias is added for "git checkout -"
- gm- alias is added for "git merge -"

## Other comments:

...


plugins/git/git.plugin.zsh
```diff
@@ -65,6 +65,7 @@ alias gbss='git bisect start'
 
 alias gc='git commit -v'
 alias gc!='git commit -v --amend'
+alias gc-='git checkout -'
```
sandipsubedi, line 68: Just out of curiosity, will this be an issue since we already have `gc` for `git commit`? 

Also, we also have an alias for `git checkout` as `gco`.

raxetul: There is no issue git existing `gc` alias. I've tested it.
This is one of my alias that I use always. Just wanted to share that it might be liked and used by others.

raxetul:
I realized that gm- can be added as "git merge -" If both command is OK, this PR can be ignored and I will open a single PR for both. I'm just waiting for maintainer responses/decisions for this PR. I'm not in a hurry.

53jk1 (approved):

raxetul:
@mcornella, could you check this one?

carlosala:
I'd say this aliases do not help to save that many clicks. Instead of `gco-`, you could do `gco -` which is almost the same amount of chars.
Thanks for your contribution!

PR #9277, opened 2020-09-22
Git alias added: `gc-` for `git checkout -`

gitaarik:
`git checkout -` checks out the last branch/commit. It is very useful if you are switching back and forth between two branches. Therefore, the `git checkout` command has the `-` shortcut, which is a synonym for `"@{-1}`. See `man git-checkout` section `OPTIONS` and then option `<branch>` (https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-ltbranchgt).

This MR adds an alias to `git checkout -`. The alias will be: `gc-`. It makes sense to use `-` because that is the character that the original git command uses. Which is also in line with other shell commands, like `cd -`, which changes the working directory to the last working directory.

## Standards checklist:

## Changes:

- Added alias `gc-` for `git checkout -`

## Other comments:

...


jrodva (approved):
👍

raxetul:
Clone of #9272. I consider adding `gm-` for `git merge -` and waiting responses from community.

kbakdev (approved):

robbyrussell:

We looked at the same request on #9272 and closed it there, with the reasoning that `gco -` already does this and is almost the same number of characters. That holds here too.

Worth mentioning that no alias in the plugin ends in a hyphen, so `gc-` would be the only one of its shape... that's a small thing on its own, but it's the kind of inconsistency that gets noticed once it ships.

PR #9280, opened 2020-09-23
brew: add alias for brew to avoid upgrading casks

paour:
Since July 2020, `brew upgrade` is upgrading casks by default, which causes painful side-effects such as reseting the quarantine flag on executables and some permissions, such as notifications.

The Homebrew maintainers insist this is intended, and the only way to avoid it is to use `brew upgrade --formula`.

https://discourse.brew.sh/t/brew-upgrade-now-upgrades-casks-but-i-dont-want-that/8459/7
https://github.com/Homebrew/brew/issues/8228

## Standards checklist:

## Changes:

- add `buf` alias to run `brew update --formula`.

## Other comments:

...


larson-carter:
Great changes (Personally I will be using this) I'm passing it on internally to see if we can merge it!

mcornella:
LGTM. Can you also update the comments on `brew upgrade` aliases to clarify that they'll also update casks?

mcornella:
Thanks!

PR #9308, opened 2020-10-03
virtualenv plugin: pull into lib/prompt_info_functions

matsievskiysv:
## Standards checklist:

## Changes:

This function should be in lib. Many themes depend on it or copy its functionality.

### DIFF
--- a/lib/prompt_info_functions.zsh
+++ b/lib/prompt_info_functions.zsh
@@ -40,3 +40,13 @@ ZSH_THEME_RVM_PROMPT_OPTIONS="i v g"
 function ruby_prompt_info() {
   echo $(rvm_prompt_info || rbenv_prompt_info || chruby_prompt_info)
 }
+
+# use this to enable users to see current python virtualenv name
+# `ZSH_THEME_VIRTUALENV_PREFIX`: sets the prefix of the VIRTUAL_ENV. Defaults to `[`.
+# `ZSH_THEME_VIRTUALENV_SUFFIX`: sets the suffix of the VIRTUAL_ENV. Defaults to `]`.
+# Note: to disable prompt mangling in virtual_env/bin/activate
+# use `export VIRTUAL_ENV_DISABLE_PROMPT=1``
+function virtualenv_prompt_info(){
+  [[ -n ${VIRTUAL_ENV} ]] || return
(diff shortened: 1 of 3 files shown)


lib/prompt_info_functions.zsh
```diff
@@ -40,3 +40,14 @@ ZSH_THEME_RVM_PROMPT_OPTIONS="i v g"
...
+# use this to enable users to see current python virtualenv name
+# `ZSH_THEME_VIRTUALENV_PREFIX`: sets the prefix of the VIRTUAL_ENV. Defaults to `[`.
+# `ZSH_THEME_VIRTUALENV_SUFFIX`: sets the suffix of the VIRTUAL_ENV. Defaults to `]`.
+function virtualenv_prompt_info(){
+  [[ -n ${VIRTUAL_ENV} ]] || return
+  echo "${ZSH_THEME_VIRTUALENV_PREFIX:=[}${VIRTUAL_ENV:t}${ZSH_THEME_VIRTUALENV_SUFFIX:=]}"
+}
+
+# disables prompt mangling in virtual_env/bin/activate
+export VIRTUAL_ENV_DISABLE_PROMPT=1
```
mcornella, line 53: The fact that this needs to be set is enough to warrant this being its own separate plugin. Otherwise, people who don't use the `virtualenv_prompt_info` function won't see their prompt changed when running the activate function.

matsievskiysv: This may be disabled by default.

matsievskiysv:
[This](https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/pygmalion-virtualenv.zsh-theme#L4-L17) is a fine example of using virtualenv plugin. This is just to use (almost) one line function.

matsievskiysv:
Quick grep showed that at least 4 themes use their own version of `virtualenv_prompt_info` because having it as a plugin is unreliable.

kbakdev (approved):

robbyrussell:
Thanks for this, and sorry it sat so long without a real answer.

We ended up going the other way on this. `lib/prompt_info_functions.zsh` keeps only dummy stubs so themes don't error, and the real implementation stays in the plugin... `virtualenv_prompt_info` is already in that stub list. We did the same thing again this week in #8924, which added `go_prompt_info` as a stub in lib with the actual code in `plugins/golang`.

PR #9312, opened 2020-10-04
i made a simple and minimalistic theme for the zsh shell

HeroGlaucoP:
it's really simple but i like it

### DIFF
--- a/themes/cool-simple-zsh-theme.zsh-theme
+++ b/themes/cool-simple-zsh-theme.zsh-theme
@@ -8,15 +8,15 @@ else
     local user_symbol='$'
 fi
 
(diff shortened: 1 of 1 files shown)


mcornella:
[We can't accept more themes](https://github.com/ohmyzsh/ohmyzsh#do-not-send-us-themes). Please make a separate repo with your theme and link to it from the [external themes](https://github.com/ohmyzsh/ohmyzsh/wiki/External-themes) wiki page.

You can also add your theme as a custom theme. See [customization](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes) for instructions.

PR #9315, opened 2020-10-06
feat(plugins): add grc plugin for Generic Colouriser

mamercad:
## Standards checklist:

## Changes:

- No changes, adds a new plugin (grc)

## Other comments:

```zsh
❯ grep ^plugins ~/.zshrc
plugins=(direnv fzf git kubectl pipenv ssh-agent tmux pyenv grc)

❯ whence ping
grc --colour=auto /sbin/ping
```

### DIFF
--- a/plugins/grc/README.md
+++ b/plugins/grc/README.md
@@ -1,0 +1,9 @@
+# Generic Colouriser plugin
+
+This plugin adds completion for [Generic Colouriser](https://github.com/garabik/grc):
+
+To use it, add `grc` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... grc)
+```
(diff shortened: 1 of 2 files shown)


mcornella:
Looks like the project is dead?

mamercad:
> Looks like the project is dead? Last commit was April 2018.

Possibly, I *still* use the utility. Is there a stipulation that "a plugin needs to be backed by a project that has a certain level of activity"?

53jk1 (approved):

mcornella:
I fixed the code style and switch to using function wrappers instead of aliases, which enables having user aliases to the commands and preserves command completion. I removed `ls` from the list as well because I couldn't get it to work with `grc`.

PR #9316, opened 2020-10-06
kubectl: add wrappers for colorized output in JSON and YAML

raphink:
## Standards checklist:

## Changes:

- Add two wrapper functions to the kubectl plugin for colorized output

### DIFF
--- a/plugins/kubectl/README.md
+++ b/plugins/kubectl/README.md
@@ -109,10 +109,8 @@ plugins=(... kubectl)
 
 ## Wrappers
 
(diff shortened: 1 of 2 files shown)


mcornella:
`jq` supports colorization with `-C`, maybe use that instead of another tool?

raphink:
@mcornella I'm using `jq` precisely. I've only added `fx` today because it provides an interactive way of parsing the JSON, so it's not only colorization.

mcornella:
Sorry, I misread that. `jq` does colorize by default, also. +1

PR #9328, opened 2020-10-11
Dirhistory plugin: add the cde command

jeffwilliams:
The cde command is used to preserve future/forward directories when
changing to an arbitrary directory, instead of losing the future directories like when using 'cd DIR'.

For example say the user has navigated the directory history such that the past directories are (/tmp /tmp/a) and the future directories are (/usr /usr/bin). At this point if the user did 'cd /bin', then the past directories would become (/tmp /tmp/a /bin). That is, the future directories (/usr /usr/bin) are lost.

Instead if we were in the original state and 'cde /bin' was used then the past directories would become (/tmp /tmp/a /usr /usr/bin /bin); it is as if the user went forward in history as far as possible and then did 'cd /bin' so that the "future" directories are preserved. 

## Standards checklist:

## Changes:

- Adds the cde command to the dirhistory plugin as described above.

## Other comments:

None

### DIFF
--- a/plugins/dirhistory/README.md
+++ b/plugins/dirhistory/README.md
@@ -19,8 +19,7 @@ plugins=(... dirhistory)
 
 ## Usage
 
-This plugin allows you to navigate the history of previous current-working-directories using ALT-LEFT and ALT-RIGHT. ALT-LEFT moves back to directories that the user has changed to in the past, and ALT-RIGHT undoes ALT-LEFT. MAC users may alternately use OPT-LEFT and OPT-RIGHT.
-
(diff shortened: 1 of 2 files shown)


53jk1 (approved):

mcornella:
I wonder if it wouldn't be better to fix the default behavior, instead of introducing a new command. Also, maybe the future directories would be better kept as future directories. Thoughts?

mcornella:
Closed in favor of fixing the current behavior of preserving future directories.

mcornella:
Decided to push this instead, as the other option would be a breaking change.

PR #9330, opened 2020-10-11
ask with lowercase y in update dialogue

Destaq:
## Standards checklist:

## Changes:

- changed `oh-my-zsh` update dialogue to two lowercase options

## Other comments:

It makes more sense having both yes *and* no represented in lower case, especially because `y` can be used in response to the dialogue:
```
[oh-my-zsh] Would you like to update? [Y/n] y
```


mcornella:
The uppercase option is traditionally used to indicate that it's the default one if no option is entered (if Enter is pressed, it'll be as if y is entered).

I don't know when this tradition started but [it's replicated everywhere in a console application](https://askubuntu.com/a/322601).

PR #9335, opened 2020-10-12
add new search path /usr/bin/virtualenvwrapper

GannonTdW:
Community ArchLinux package install virtualenvwrapper{_lazy,}.sh in /usr/bin
https://www.archlinux.org/packages/community/any/python-virtualenvwrapper/


mcornella:
This should be covered by lines 3 and 4 which use `$commands[virtualenvwrapper_lazy.sh]` and `$commands[virtualenvwrapper.sh]`, provided that `/usr/bin` is in `$PATH` (which it always is, unless you messed with `$PATH` in a weird way). Does it not work for you? If it doesn't, can you post the output of `ls -l /usr/bin/virtualenvwrapper{,_lazy}.sh`?

GannonTdW:
? I have test on a other ArchLinux and I have no problem.
May be one my "Hack", I don"t understand.

Sorry, I close

PR #9339, opened 2020-10-13
Add rke command completion

weh:
## Standards checklist:

## Changes:

- Adds the Rancher Kubernetes Engine `rke` CLI completion


weh:
@mcornella do you want completions like this? i plan to add k3s also, since i use it regularly.

kbakdev (approved):

robbyrussell:

The reason is upstream rather than the code: `rancher/rke` is archived now, with an end-of-life notice pointing people at RKE2. rke also never shipped a completion generator, so this had to be hand-written, which means someone would need to keep re-transcribing it as the CLI changed... for a tool that has stopped changing because it has stopped.

PR #9340, opened 2020-10-14
Add Fedora/RHEL/CentOS to Installation Problems

summonholmes:
## Standards checklist:

## Changes:

For the last few releases of Fedora Workstation and for other RHEL-like environments, the `chsh` command is not available by default.  Guidance should be provided on installing `util-linux-user`.  I think it makes sense to merge basic instructions into the README Installation Problems section, perhaps even adding a prompt in the `install.sh` if it can detect that yum/dnf exists.


53jk1 (approved):

mcornella:
This should be documented in the [wiki](https://github.com/ohmyzsh/ohmyzsh/wiki/Installing-ZSH).

PR #9346, opened 2020-10-15
kubectl: verify kubectl is installed before running compdef

### LINKED ISSUE #9345: kubectl plugin: compdef: unknown command or service: kubectl
**Describe the bug**
recent [PR](https://github.com/ohmyzsh/ohmyzsh/pull/9316) introduced syntax coloring but implementation expects kubectl is installed prior - which may not always be the case. In a fresh deployment, kubectl may not be installed prior to configure the shell environment - or the user may not have kubectl installed at all (in my case, i have enabled the same plugins from host to host and only some have kubectl installed (rest of the plugins enabled have no issue).

In this case, if we check for kubectl, prior to running 'compdef' (which triggers the error output), that will alleviate the messages getting printed at shell init.

**To Reproduce**
1. enable the kubectl plugin
2. make sure kubectl is not installed
3. open a new shell

**Expected behavior**
no error messages are printed on shell login/init

**Screenshots and/or Recordings**

image

causalityloop:
## Standards checklist:

## Changes:
This change fixes #9345.

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -150,22 +150,12 @@ alias kepvc='kubectl edit pvc'
 alias kdpvc='kubectl describe pvc'
 alias kdelpvc='kubectl delete pvc'
 
-# Colored JSON output
-kj() {
-  kubectl "$@" -o json | jq
-}
+# Only run if the user actually has kubectl installed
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

causalityloop:
thanks for the tip on `$+commands[]` - i had to look that one up. cool.

PR #9348, opened 2020-10-15
Allow dependabot to check GitHub actions daily

jlosito:
## Standards checklist:

## Changes:

- This allows dependabot to check any GitHub actions which this project uses and submit PRs automatically with version bumps to any GitHub action which is out-of-date.

## Other comments:

Here's a link to GitHub's documentation with essentially the same snippet.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


mcornella:
We don't use external GitHub Actions though (that's not entirely true, we use [`actions/checkout@v2`](https://github.com/actions/checkout), but that isn't supposed to change automatically).

Thanks for the info, if and when we need it we'll be sure to credit you. Cheers

PR #9351, opened 2020-10-16
Use GRADLE_USER_HOME for completion cache if set

cwrau:
## Standards checklist:

## Changes:

- Use the `$GRADLE_USER_HOME` environment variable if set, otherwise use `$HOME/.gradle`


53jk1 (approved):

mcornella:
The `_gradle` file is a bundle of the [upstream project](https://github.com/gradle/gradle-completion/blob/master/_gradle), you should make the PR there.

PR #9362, opened 2020-10-19
gitfast: upgrade completions to v2.28.0

ThisIsMissEm:
# Standards checklist:

## Changes:

- updates the completions in the gitfast plugin to 2.28.0 of git, which is the latest stable version of git.

## Other comments:

In more modern versions of git, additional commands have been added to improve user experience, such as `git switch` and `git restore` instead of using `git checkout` for everything

### DIFF
--- a/plugins/gitfast/update
+++ b/plugins/gitfast/update
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 url="https://git.kernel.org/pub/scm/git/git.git/plain/contrib/completion"
-version="2.16.0"
+version="2.28.0"
 
 curl -s -o _git "${url}/git-completion.zsh?h=v${version}" &&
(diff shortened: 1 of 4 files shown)


plugins/gitfast/update
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 url="https://git.kernel.org/pub/scm/git/git.git/plain/contrib/completion"
-version="2.16.0"
+version="2.28.0"
```
ThisIsMissEm, line 4: This is the only line in this PR that I actually changed, all other files are automatically generated

53jk1 (approved):

felipec:
@ThisIsMissEm This pull request is correct, but I already sent one with a lot more changes.

PR #9364, opened 2020-10-20
fix(battery): use `pmset` for battery percentage in macOS

aimuz:
returns the battery charge consistent with the system battery manager

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/battery/battery.plugin.zsh
+++ b/plugins/battery/battery.plugin.zsh
@@ -18,10 +18,7 @@ if [[ "$OSTYPE" = darwin* ]]; then
   }
 
   function battery_pct() {
-    local battery_status="$(ioreg -rc AppleSmartBattery)"
-    local -i capacity=$(sed -n -e '/MaxCapacity/s/^.*"MaxCapacity"\ =\ //p' <<< $battery_status)
-    local -i current=$(sed -n -e '/CurrentCapacity/s/^.*"CurrentCapacity"\ =\ //p' <<< $battery_status)
-    echo $(( current * 100 / capacity ))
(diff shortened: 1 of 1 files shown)


mcornella:
I don't have a Mac to test this, so can you post the following?
- Output with old version
- Output with new version
- Output of `pmset -g batt`

aimuz:
### old version

```bash
function battery_pct() {                                                                                                                                                                                                                             
    local battery_status="$(ioreg -rc AppleSmartBattery)"
    local -i capacity=$(sed -n -e '/MaxCapacity/s/^.*"MaxCapacity"\ =\ //p' <<< $battery_status)
    local -i current=$(sed -n -e '/CurrentCapacity/s/^.*"CurrentCapacity"\ =\ //p' <<< $battery_status)
    echo $(( current * 100 / capacity ))
}
$ battery_pct                                                                                                                                                                                                                                            
97
```

### new version

```bash
function battery_pct() {                                                                                                                                                                                                                             
    pmset -g batt | grep -Eo "\d+%" | cut -d% -f1
}
$ battery_pct                                                                                                                                                                                                                                            
100
```

### pmset -g batt

```
Now drawing from 'Battery Power'
 -InternalBattery-0 (id=8388707)	100%; discharging; 2:28 remaining present: true
```


The output of the old version may be closer to the actual situation of the battery, because it is calculated by its own algorithm, while the new version directly uses the system battery management value. There will be differences between them.

53jk1 (approved):

aimuz:
@mcornella Hello, what is the PR status

Recently, I haven't seen any state

mcornella:
Thanks!

PR #9373, opened 2020-10-22
[kubectl] Added alias for several kubectl commands

szarroug3:


robbyrussell:

`kd` and `kg` are the two we want, but they're already the whole content of #8851, which is open and with its author for a rebase. Merging both would collide, so I'm keeping that one as the canonical PR for those two rather than splitting the credit.

`kdelf!` is the only part that's unique here, and I don't think it works as written: `-f` takes a filename, so `kdelf! app.yaml` would pass `--grace-period=0` where the file should go. kubectl also wants `--force` alongside `--grace-period=0` for that to do what it looks like it does.

Closing for that reason. Thanks for putting it up, and the timing is just bad luck... you asked for `kd` and `kg` around the same time as the other PR.

PR #9379, opened 2020-10-24
[git] Added alias for several git commands

szarroug3:


robbyrussell:

Four of the six wrap `git cat-file`, which is plumbing... it's mostly reached for from scripts rather than typed at a prompt, and the bar for adding to the git plugin is that a lot of people use it often.

PR #9386, opened 2020-10-24
Shorten installation links

Davoodeh:
## Changes:
- Installation links in `README.md` are replaced with shorter equivalents.


mcornella:
There's really no need, we gain very few characters and add a redirect. But thanks for the contribution!

PR #9393, opened 2020-10-27
Add kubectl kcvm alias

blacksails:
## Standards checklist:

## Changes:

- Add an alias `kcvm` for `kubectl config view --minify`

## Other comments:

This commands is very useful for finding out what user/cluster the current
context resolves to. It also makes it easy to lookup the api server hostname.

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -31,6 +31,9 @@ alias kccc='kubectl config current-context'
 # List all contexts
 alias kcgc='kubectl config get-contexts'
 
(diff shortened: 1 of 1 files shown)


blacksails:
I could also add an alias for `kubectl config view` without the `--minify`. That prints out the entire kubeconfig. Personally I don't find that very useful, which is why I didn't include it.

I decided to include the `m` in the alias for `--minify`.

blacksails:
I also considered adding an alias which returns the URL for the current api server:
`kubectl config view --minify -ojsonpath="{.clusters[0].cluster.server}"`

PR #9397, opened 2020-10-28
aws: get and set session token if available

setu4993:
## Standards checklist:

## Changes:

Get and set session token if available and already configured when `role_arn` is not set.

## Other comments:

Changes from https://github.com/ohmyzsh/ohmyzsh/pull/8419/ stopped setting the value for `AWS_SESSION_TOKEN` for me. Traced it back to this line of change.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -61,7 +61,7 @@ function asp() {
     else
       aws_access_key_id="$(aws configure get aws_access_key_id --profile $1)"
       aws_secret_access_key="$(aws configure get aws_secret_access_key --profile $1)"
-      aws_session_token=""
+      aws_session_token="$(aws configure get aws_session_token --profile $1)"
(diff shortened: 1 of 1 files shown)


mcornella:
cc @maksyms

maksyms:
@setu4993 Thanks for your submission!

Initially, this use case wasn't considered, as the session token has a limited time span. But if the session token is set externally in your case somehow (i.e., using another script), that's fine.  So this change is👍 from me.

@mcornella - good to go from my side, thanks for flagging up!

mcornella:
@maksyms would you want to be made plugin maintainer? This would mean automatic mentions on PRs that modify the aws plugin.

maksyms:
@mcornella Yes, sure, thank you. We rely on this plugin a fair bit, and for now, I'm happy to help maintain it. 👍

mcornella:
Ok done. Thanks!

PR #9412, opened 2020-11-01
git: Enable customization of main branch name (non-breaking)

tywmick:
## Standards checklist:

## Changes:

Hello! Following up on #9049 and #9091, I had an idea for a non-breaking way to let users specify the name of their main branch in a given repo: using a custom `git config` field called `omz.git.mainBranch` (which I don't think anyone would already be using for any other purpose).

I adjusted `git_main_branch` to check this field first before falling back to its previous functionality, and I added a new function `set_main_branch` as a shortcut for `git config --local omz.git.mainBranch`. And I adjusted the plugin's readme to reflect these changes.

Is this a good solution? Should `set_main_branch` be an alias instead of a function? Should it have a nicer, more abbreviated name? I figured it'd be a command people wouldn't be typing too often, since you can pretty much set it and forget it for any individual repo that doesn't already use `main` or `master`.

### Updated `git_main_branch`:

```zsh
# Check if user specified main branch in config, then check if "main" exists,
# then default to "master"
function git_main_branch() {
  local main_branch=$(git config omz.git.mainBranch)
  if [[ -n "$main_branch" ]]; then
    echo "$main_branch"
  elif [[ -n "$(git branch --list main)" ]]; then
    echo main
  else
    echo master
  fi
}
```

### New `set_main_branch`:

```zsh
# Set the name of the main branch in a local repository
function set_main_branch() {
  git config --local omz.git.mainBranch "$1"
}
```

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -208,14 +208,14 @@ These are aliases that have been removed, renamed, or otherwise modified in a wa
 
 ### Current
 
-| Command                | Description                                                                                     |
-|:-----------------------|:------------------------------------------------------------------------------------------------|
-| `grename <old> <new>`  | Rename `old` branch to `new`, including in origin remote                                        |
-| current_branch         | Return the name of the current branch                                                           |
-| git_current_user_name  | Returns the `user.name` config value                                                            |
-| git_current_user_email | Returns the `user.email` config value                                                           |
+| Command                | Description                                                                                  |
+|:-----------------------|:---------------------------------------------------------------------------------------------|
+| `grename <old> <new>`  | Rename `old` branch to `new`, including in origin remote                                     |
+| current_branch         | Return the name of the current branch                                                        |
+| git_current_user_name  | Returns the `user.name` config value                                                         |
+| git_current_user_email | Returns the `user.email` config value                                                        |
 | git_main_branch        | Returns the name of the main branch: the value of `git config omz.git.mainBranch` if that key has been set, or `main` if a `main` branch exists, or `master` otherwise |
-| set_main_branch        | Sets the name of the main branch in your local repository using `git config omz.git.mainBranch` |
+| set_main_branch        | Sets the name of the main branch in a local repository using `git config omz.git.mainBranch` |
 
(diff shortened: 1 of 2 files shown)


mcornella:
I dislike this kind of freedom mainly because it promotes non-standard naming. I recently merged a PR that added `trunk` as another option, but only because it's already used in a relatively popular open source project.

I prefer having to manually change the `git_main_branch` function, and only if there's evidence in favor of doing so, because I believe we should all move towards a common naming scheme.

tywmick:
That's a good reason. 👍🏼 The [GitHub CLI](https://github.com/cli/cli/)'s `trunk` branch was actually the reason I made this PR, too, so #9417 solves all my problems!

dsayling:
for those that get to this as I - was wondering if there was some sort of customization for the ohmyzsh git aliases `gcm, gmom, etc...`. The easiest you can do is just override the ref in your zshrc
```
function git_main_branch() {
  if [[ -n "$(git branch --list develop)" ]]; then
    echo develop
  elif [[ -n "$(git branch --list main)" ]]; then
    echo main
  else
    echo master
  fi
}
```
if you dont mind waiting on the remote connection:
```
function git_main_branch() {
  def=`git remote show origin | sed -n '/HEAD branch/s/.*: //p'`
  echo $def
}
```

mogsdad:
> git remote show origin | sed -n '/HEAD branch/s/.*: //p'

You can avoid waiting on the remote connection via:
```bash
git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'
```
so:
```bash
function git_main_branch() {
  def=`git remote show origin | sed -n '/HEAD branch/s/.*: //p'`
  echo $def
}
```

ref: [SO/git - how to get default branch?](https://stackoverflow.com/a/44750379/1677912)

NeilGirdhar:
> ```shell
> git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'
> ```

Why isn't this the default setting for the `git_main_branch` function?

PR #9413, opened 2020-11-01
Support using mvnw in multi-module projects

### LINKED ISSUE #8557: Look for mvnw up the file system
mvn

When I run `mvn`, this uses the alias `mvn-or-mvnw`. If I do this in a directory that contains `./mvnw` then this works exactly as needed. If I do this in a subdirectory - e.g. for a multi-module project - then this does *not* work.

I'd really like it if the `mvn-or-mvnw` alias could walk up the tree to find `mvnw` and run that if it's found.

For example: If running in /Users/coxg/source/project/submodule/subsubmodule
* Look for /Users/coxg/source/project/submodule/subsubmodule/mvnw
* Failing that, look for /Users/coxg/source/project/submodule/mvnw
* Failing that, look for /Users/coxg/source/project/mvnw
* Finding that, run that script to execute Maven

The script already works correctly if I run it manually - e.g.:
```
$ ../../mvnw
```

but the alias doesn't go looking for it.

amaltson:
## Standards checklist:

## Changes:

Frequently Maven projects are multi-module, so there's a top level directory that has the `mvnw `script and a parent POM file and a number of subdirectories that have the their own POM files but no `mvnw` wrapper. Walk one directory up to see if there's a `mvnw` and execute it, supporting this use case.

Fixes #8557

## Other comments:

I had attempted to make a recursive function that walks up to the `/` directory looking for a `mvnw` wrapper script but couldn't get it working. I figure since a 1 level deep multi-module is the most common scenario, this should be sufficient.

### DIFF
--- a/plugins/mvn/mvn.plugin.zsh
+++ b/plugins/mvn/mvn.plugin.zsh
@@ -1,8 +1,12 @@
 # Calls ./mvnw if found, otherwise execute the original mvn
 mvn-or-mvnw() {
-	if [ -x ./mvnw ]; then
+	if [ -x ./mvnw ] || [ -x ../mvnw ]; then
 		echo "executing mvnw instead of mvn"
-		./mvnw "$@"
+		if [ -x ./mvnw ]; then
+			./mvnw "$@"
+		else
+		  ../mvnw "$@"
+		fi
 	else
 		command mvn "$@"
 	fi


sazzer:
This looks like it will find `mvnw` in the current or parent directory, but no further than that.

I.e. if I have the following tree:
```
├── project
│   ├── moduleA
│   │   ├── submoduleB
├── mvnw
```

Then it will work from `project` or `project/moduleA` but not from `project/moduleA/submoduleB`. And feasibly the project structure could be any arbitrary depth.

There are some (Bash) examples for walking up the directory tree all the way to the root that you can see here: https://stackoverflow.com/questions/9376904/find-file-by-name-up-the-directory-tree-using-bash

PR #9417, opened 2020-11-03
feat(git): support trunk branches in `git_main_branch`

alissonbrunosa:
## Standards checklist:

## Changes:
It adds `trunk` branch name as option in case the repository uses it as default branch.
- [...]

## Other comments:

...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -33,7 +33,7 @@ function work_in_progress() {
 function git_main_branch() {
   local branch
   for branch in main trunk; do
-    if git show-ref -q --verify refs/heads/$branch; then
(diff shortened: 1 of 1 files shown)


mcornella:
Is there a reason you're not using `main`? That seems to be the future standard, so reliance on this function is going to bring problems down the line.

alissonbrunosa:
Hey @mcornella, yeah there is, I am contributing for https://github.com/cli/cli, and they're using `trunk` as default branch, and I believe there are more out there.
Being honest, I don't think the community has defined a standard :sweat_smile:

mcornella:
Weird since GitHub itself had defined `main` as the new default. I've changed the logic to a for in case we need to support other cases in the future.

alissonbrunosa:
@mcornella Awesome.

PR #9423, opened 2020-11-04
'.' no longer special, '..' only sometimes - makes completion easier

omertuc:
## Standards checklist:
- [?] The code is mine or it's from somewhere with an MIT-compatible license. - It was taken from http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Completion-Directories so I think it's fine?

## Changes:

Modified the treatment of `..` and `.` as special-dirs:
- `.` no longer a special dir
- `..` a special dir only under specific circumstances - only when the prefix is empty, `.` or `../`

First I'd like to say I did not come up with this solution on my own - I found [this (ctrl-f special-dirs)](http://zsh.sourceforge.net/Doc/Release/Completion-System.html) while looking how to solve it.

Here's the original quote:

> The following example sets special-dirs to ‘..’ when the current prefix is empty, is a single ‘.’, or consists only of a path beginning with ‘../’. Otherwise the value is ‘false’.
```
zstyle -e ':completion:*' special-dirs \ 
   '[[ $PREFIX = (../)#(|.|..) ]] && reply=(..)'
```

The motivation for this is as follows:
- When the completion menu pops up, it's always much more useful to have actual files suggested and not special dirs that can anyway be easily typed (it's just `..` or `.`) and don't require any completion. It's annoying to always have them at the beginning of the list, being forced to skip them using `tab` twice to get to the actual files.

- When there's only a single file in a directory, it's much nicer having it completed directly without a menu. Instead, because `.` and `..` are always special directories, they always get suggested along with the file, forcing you to either type the first letter of the file to have it completed or skip `.` and `..` by pressing tab twice, once for each.

- The above point makes "descending" (via completion) into long chains of nested single-directories so much nicer - just tab tab tab tab. Since there are no special dirs being suggested - the only option is the single directory at each level - and it gets completed without a menu.

- The original motivation for even adding them as special dirs, according to the PR that added the line modified by this PR, is to have `..<TAB>` complete to `../` which is definitely nice. This is why if the prefix is empty, `.` or `..` then `..` is still suggested as a special dir, and thus the `/` still gets added automatically. Also, `../` is another prefix that causes `..` to be special - this makes it easier to go back `../../../` a lot of times, just . and tab . and tab to get `../../../`.

I couldn't find ways where this breaks existing behaviour in substantial ways, other than `.` and `..` being gone from completion menus... who uses the completion menu to get `..` and `.` anyway? I think the improvements are so substantial that this "breaking change" is worth it.

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -25,8 +25,8 @@ else
 fi
 unset CASE_SENSITIVE HYPHEN_INSENSITIVE
 
-# Complete . and .. special directories
-zstyle ':completion:*' special-dirs true
+# Only treat .. as a special directory when the prefix is empty, '.' or '../'
+zstyle -e ':completion:*' special-dirs '[[ $PREFIX = (../)#(|.|..) ]] && reply=(..)'
 
 zstyle ':completion:*' list-colors ''
 zstyle ':completion:*:*:kill:*:processes' list-colors '=(#b) #([0-9]#) ([0-9a-z-]#)*=01;34=0=01'


robbyrussell:
Thanks for this, and sorry it sat this long without a reply.

I tried to reproduce the problem before deciding, on zsh 5.9, comparing both settings across `ls`, `cd`, and a few prefixes. For an empty or ordinary prefix the results were byte-identical, including descending through nested single directories. The one real difference I found is `ls .<TAB>`: today it offers `./` and `../`, and with your change it goes straight to `../` while `./` stops completing.

That's a change to a default every user gets, and I couldn't observe the benefit on a current zsh. It's quite possible the annoyance was real on zsh 5.8 back when you wrote this and has since been fixed upstream... I had no way to test that, so I'm not claiming you were wrong at the time.

For anyone who prefers the behavior, setting `zstyle ':completion:*' special-dirs false` in `.zshrc` works and is what we've pointed people to on #13069.

Closing for that reason. Thanks for the careful write-up.

omertuc:
Out of curiosity, tried to spin up an ubuntu 20.04 container with zsh 5.8, didn't reproduce. Tried to move to an omz commit from around the time I created this issue, still nothing. Tried to use my .zshrc from that time, right before I added the zstyle directive above, and it did reproduce.

Tried my .zshrc from that time on a fresh modern machine - also reproduced.

So next I tried to narrow down what about my .zshrc from that time caused this behavior and I found:

```bash
_comp_options+=(globdots)
```

Which I guess I wanted at the time to make hidden files show up in the autocomplete menu, as I often `cd ~/`<kbd>Tab</kbd> just to remember what directories I have *including* hidden ones (otherwise I'm not even aware they're there enough to know I should type `.`<kbd>Tab</kbd> to see them). I just want to see all directories. I never have so many hidden directories to the point I don't want to not see them or have them listed.

But it has the unwanted side effect of now always suggesting `.` and `..` like I complained in my original issue (complained while unaware `globdots` was causing it) as if they're part of the "hidden" subdirectories I want to see

So with the zstyle suggestion enabled alongside `globdots` I get the best of both worlds - hidden files showing up in my tab autocomplete, and `.` and `..` not being suggested when I otherwise just wanna quickly descend into nested single directories

As to what oh-my-zsh should do about it... Maybe it's good to have this zstyle for everyone so once they do try to enable `globdots`, they will not run into the same problem as I did? But at this point I'm not sure what's your policy on putting in things vs documenting how things should be done, I'm not that familiar with omz enough to say

So up to you if you wanna reopen this issue, or maybe if you want me to open a new issue rephrasing it more correctly to be about `globdots`, or do nothing at all. lmk

PR #9428, opened 2020-11-07
feat(updater): add changelog and restart session on successful update

### LINKED ISSUE #2971: Introducing "What's new" after upgrade
Upgrades happens quite frequently, but to the end user it is very unclear what actually changed and why one should upgrade. Having a changelog or a "what's new" chunk of text would be awesome.

mcornella:
## Standards checklist:

## Changes:

- refactor(updater): switch to Zsh execution and fix git remote detection logic
- fix(updater): correctly restart the zsh session when the update pulled changes
- feat(updater): add changelog display by parsing the commit list
- feat(CLI): add `omz changelog` command

## Other comments:

**NOTE:** after updating, the changelog is displayed briefly with `less` if available, or directly `cat`-ed to the console if `less` isn't available. This might not be the best UI, so I'll appreciate any feedback on this.

The changelog functionality is dependent on commits following the [Conventional Commits spec](https://www.conventionalcommits.org/).

Sample screenshot (works with new and old commit formats):

image

In this screenshot, the call to `omz changelog` expects a commit-ish (branch, hash or tag work). In the future, this will only be needed for looking at the changelog of particular versions, so you'll just run `omz changelog v1.0.0`, for example.

Fixes #2971

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -64,6 +64,13 @@ if git pull --rebase --stat origin master; then
     ret=80 # non-zero exit code to indicate no changes pulled
   else
     message="Hooray! Oh My Zsh has been updated!"
+
+    # Display changelog with less if available, otherwise just print it to the terminal
+    if (( $+commands[less] )); then
+      command less -R <("$ZSH/tools/changelog.sh" HEAD "$last_commit")
+    else
+      "$ZSH/tools/changelog.sh" HEAD "$last_commit"
+    fi
   fi
 
   printf '%s         %s__      %s           %s        %s       %s     %s__   %s\n' $RAINBOW $RESET
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -1,13 +1,23 @@
-# Use colors, but only if connected to a terminal, and that terminal
-# supports them.
+#!/usr/bin/env zsh
+
+cd "$ZSH"
+
+# Use colors, but only if connected to a terminal
+# and that terminal supports them.
+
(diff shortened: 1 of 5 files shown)


mcornella:
After properly supporting revert commits this is ready to go.

tools/upgrade.sh
```diff
@@ -1,13 +1,23 @@
...
-# supports them.
+#!/usr/bin/env zsh
+
+cd "$ZSH"
+
+# Use colors, but only if connected to a terminal
+# and that terminal supports them.
+
+local -a RAINBOW
+local RED GREEN YELLOW BLUE UNDER BOLD RESET
```
mati865, line 9: Those 2 lines are wrong:
```
/home/mateusz/.oh-my-zsh/tools/upgrade.sh: 8: local: not in a function
```

mcornella: How are you running the file? It should use zsh instead of `sh`.

mati865: Sorry my bad, you are right.

mcornella: This was changed recently. Where you running it manually? If so you should use `omz update` if at all possible.

I also added a check now to make sure it's always run with zsh if it was called with `sh`.

mati865: I was using Topgrade, the fix was committed recently: https://github.com/r-darwish/topgrade/commit/a68387abed301221b7fef05b800059af2e87d38e and published soon after my comment here.

PR #9431, opened 2020-11-09
feat(sudo): support aliases to $EDITOR

subnut:
If you use some alias for your $EDITOR, say `e`. You type `e /to/file`.
But that file needs sudo permission. You naturally do `<Esc><Esc><Enter>`.
But, to your dismay, it errors out `command not found`, because it
didn't detect your alias to $EDITOR, thereby failing to convert it to
`sudoedit`.

This commit tries to detect your alias and convert it to sudoedit.

## Standards checklist:

## Changes:

- Removed a trailing whitespace
- Added vim modeline
- Added my name and GitHub user URL under Authors
- Added the actual functionality this PR is intended for

## Other comments:

NONE

### DIFF
--- a/plugins/sudo/sudo.plugin.zsh
+++ b/plugins/sudo/sudo.plugin.zsh
@@ -14,6 +14,16 @@
 #
 # ------------------------------------------------------------------------------
 
+__sudo-replace-buffer() {
+    local old=$1 new=$2
+    if [[ ${#LBUFFER} -le ${#old} ]]; then
+        RBUFFER=" ${BUFFER#$old }"
+        LBUFFER="$new"
+    else
+        LBUFFER="$new ${LBUFFER#$old }"
+    fi
+}
+
 sudo-command-line() {
     [[ -z $BUFFER ]] && LBUFFER="$(fc -ln -1)"
 
@@ -33,28 +43,14 @@ sudo-command-line() {
         fi
     fi
 
-
-    if [[ -n $EDITOR && $BUFFER == $EDITOR\ * ]]; then
-        if [[ ${#LBUFFER} -le ${#EDITOR} ]]; then
-            RBUFFER=" ${BUFFER#$EDITOR }"
-            LBUFFER="sudoedit"
-        else
-            LBUFFER="sudoedit ${LBUFFER#$EDITOR }"
-        fi
-    elif [[ $BUFFER == sudoedit\ * ]]; then
-        if [[ ${#LBUFFER} -le 8 ]]; then
-            RBUFFER=" ${BUFFER#sudoedit }"
-            LBUFFER="$EDITOR"
-        else
-            LBUFFER="$EDITOR ${LBUFFER#sudoedit }"
-        fi
-    elif [[ $BUFFER == sudo\ * ]]; then
-        if [[ ${#LBUFFER} -le 4 ]]; then
-            RBUFFER="${BUFFER#sudo }"
-            LBUFFER=""
-        else
-            LBUFFER="${LBUFFER#sudo }"
-        fi
+    if [[ -n $EDITOR && $BUFFER = $EDITOR\ * ]]; then
+        __sudo-replace-buffer "$EDITOR" "sudoedit"
+    elif [[ -n $EDITOR && $BUFFER = \$EDITOR\ * ]]; then
+        __sudo-replace-buffer "\$EDITOR" "sudoedit"
+    elif [[ $BUFFER = sudoedit\ * ]]; then
+        __sudo-replace-buffer "sudoedit" "$EDITOR"
+    elif [[ $BUFFER = sudo\ * ]]; then
+        __sudo-replace-buffer "sudo" ""
     else
         LBUFFER="sudo $LBUFFER"
     fi
--- a/plugins/sudo/sudo.plugin.zsh
+++ b/plugins/sudo/sudo.plugin.zsh
@@ -15,12 +15,12 @@
 # ------------------------------------------------------------------------------
 
 __sudo-replace-buffer() {
-    local old=$1 new=$2
+    local old=$1 new=$2 space=${2:+ }
     if [[ ${#LBUFFER} -le ${#old} ]]; then
-        RBUFFER=" ${BUFFER#$old }"
-        LBUFFER="$new"
+        RBUFFER="${space}${BUFFER#$old }"
+        LBUFFER="${new}"
     else
-        LBUFFER="$new ${LBUFFER#$old }"
+        LBUFFER="${new}${space}${LBUFFER#$old }"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


subnut:
Ping @mcornella

mcornella:
Good idea but a small tip: you can use the `$aliases` variable to get the command.

subnut:
@mcornella Why is this PR not merged? Any problems?

> you can use the $aliases variable to get the command

How? I am unable to understand... It would be better if you show me how to use it in which line.....

mcornella:
> How? I am unable to understand... It would be better if you show me how to use it in which line.....

Right sorry, I thought it was clear. The `$aliases` variable is an associative array that holds, for every alias, the command it runs. So if you have an alias `e=$EDITOR`, `$aliases[e]` is `$EDITOR`. You can use this instead of checking the output of `alias`.

mcornella:
I've changed the logic to use the $aliases variable and significantly reduced the complexity. I've tested on my end and it seems to work, but do try it yourself before I merge it.

subnut:
Oh... so that's why my local commit wasn't pushing! :smile:

This is the way I had tried to implement it -
```diff
diff --git a/plugins/sudo/sudo.plugin.zsh b/plugins/sudo/sudo.plugin.zsh
index c092e81..6869bbd 100644
--- a/plugins/sudo/sudo.plugin.zsh
+++ b/plugins/sudo/sudo.plugin.zsh
@@ -28,7 +28,6 @@ sudo-command-line() {
         if [[ -z $EDITOR ]]; then                       # EDITOR not set, so no work
             return
         fi
-        _EDITOR=$EDITOR                                 # To later restore the original value
         _CURRENT=$(echo -n $BUFFER | cut -d' ' -f1)     # Get current command (to check for alias)
         if [[ $_CURRENT == $EDITOR ]]; then             # If current command is same as EDITOR, no need to check further.
             return 0
@@ -36,40 +35,13 @@ sudo-command-line() {
         if [[ $_CURRENT == 'sudoedit' ]]; then          # If current command is sudoedit, no need to check further.
             return 0
         fi
-        if [[ $EDITOR =~ ' ' ]]; then
-            EDITOR="'$EDITOR"                           # If EDITOR has spaces, then alias will be wrapped in ''
-        fi
 
-        # Check for aliases that match the value of $EDITOR
-        if
-            () {
-                for alias in $(alias | grep "=$EDITOR" | cut -d= -f1)
-                do  if [[ $alias = $_CURRENT ]]; then
-                        EDITOR=$alias   # set EDITOR to the alias
-                        return 0
-                    fi
-                done
-                return 1    # none of these aliases match
-            };
-        then
-            return
-
-        # Maybe alias is set to the variable instead of it's value
-        elif
-            () {
-                for alias in $(alias | grep \=\'\$EDITOR | cut -d= -f1)
-                do  if [[ $alias = $_CURRENT ]]; then
-                        EDITOR=$alias   # set EDITOR to the alias
-                        return 0
-                    fi
-                done
-                return 1    # none of these aliases match
-            }
+        # Check if current command is aliased to $EDITOR
+        if echo -n $aliases[$_CURRENT] | grep -e ^$EDITOR -e '^$EDITOR' &> /dev/null
         then
+            EDITOR=$_CURRENT
             return
         fi
-
-        EDITOR=$_EDITOR     # Nothing matched. Restore $EDITOR
     }
 
     if [[ -n $EDITOR && $BUFFER == $EDITOR\ * ]]; then
```

<br>
Anyway.. Okay, I shall be testing it a bit, and reporting back to you.

_off-topic_: Do you know any (free) good resources for learning zsh? I am a noob (in case you haven't figured it out already..)

subnut:
well.... just as I am about to report it, you push `e38f004`

mcornella:
> _off-topic_: Do you know any (free) good resources for learning zsh? I am a noob (in case you haven't figured it out already..)

I always end up going to the [Zsh documentation](http://zsh.sourceforge.net/Doc/Release/zsh_toc.html#SEC_Contents), most frequently the [expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion) docs. Also completion can be very useful sometimes. For example, when trying to find a way to modify a variable with `${(modifier)var}` you can type `${([TAB]` and you get many options. Reading official Zsh code (completion functions) also gives me a lot of ideas. And if that isn't enough, you have the [zsh-workers mailing list](https://www.zsh.org/mla/) at your disposal.

subnut:
@mcornella Thanks!
BTW there's a bug... aliases work well, but if the command is the actual value of $EDITOR, then it doesn't work

subnut:
See GIF - 
gifcast_201112011639

mcornella:
So `$EDITOR` is `nvim` in that case?

subnut:
Yes

mcornella:
Oh, so your setup is like
```zsh
EDITOR=nvim
alias n=nvim
```
and the problem you find is that when the typed command uses the alias, triggering the function twice goes back to using `nvim` instead of `n` which is the alias?

subnut:
No, the problem is that when the command is `nvim sudo.plugin.zsh` and the function is triggered, the result is `sudo nvim` instead of `sudoedit`

subnut:
I added this line to see the values of $LBUFFER and $EDITOR
`zle -M "$LBUFFER $EDITOR"`

See this GIF -
gifcast_201112012225

subnut:
Seems like the value of $EDITOR is set to the alias, even when the current command isn't an alias

subnut:
you can see that the value of $EDITOR is `n` when the command is `nvim`, causing `sudo nvim` instead of `sudoedit`

subnut:
UPDATE: seems like `$cmd` is only taking the first character of the buffer -
gifcast_201112012923

mcornella:
Right, I had it working before. Try with this patch:
```patch
diff --git a/plugins/sudo/sudo.plugin.zsh b/plugins/sudo/sudo.plugin.zsh
index c75821fc..f2445a76 100644
--- a/plugins/sudo/sudo.plugin.zsh
+++ b/plugins/sudo/sudo.plugin.zsh
@@ -37,7 +37,7 @@ sudo-command-line() {
     # Get the first part of the typed command and check if it's an alias to $EDITOR
     # If so, locally change $EDITOR to the alias so that it matches below
     if [[ -n "$EDITOR" ]]; then
-        local cmd="${(Az)BUFFER[1]}"
+        local cmd="${${(Az)BUFFER}[1]}"
         if [[ "${aliases[$cmd]} " = (\$EDITOR|$EDITOR)\ * ]]; then
             local EDITOR="$cmd"
         fi
```

subnut:
That was it!

mcornella:
Ok, thanks for debugging! Do you think this is ready to merge?

subnut:
Yeah, I think so...

subnut:
Don't you think these commits should be sqashed before merging?

mcornella:
Yeah but I couldn't do it without force-pushing to the branch, which automatically closes the PR because I'm not the PR owner. Instead, I'll squash it now, when merging from the GitHub UI.

PR #9437, opened 2020-11-12
refactor(colored-man-pages): move nroff wrapper and refactor logic in `colored` function

### LINKED ISSUE #9436: [Enhancement] colored-man-pages plugin

The [colored-man-pages](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/colored-man-pages) plugin.

Plugin code is needlessly complicated and undocumented. And I know how to make it better!

Also, unless someone can provide me with a reason for turning on Nroff Unsafe mode (`_NROFF_U=1`), I'm in favor of removing it. I'm not an expert in Groff/Nroff, but it just sounds dangerous.

And what's the deal with silently creating executable scripts in `$HOME/bin`? Thanks God, it does not affect me — only solaris users. But that is certainly **not** what I would expect from a plugin like this. At the very least, it should ship that executable script with the plugin, and extend PATH to its parent directory. It's not like plugins aren't allowed to provide additional supporting files, do they?

Please, take a look at [my rewrite](https://github.com/ratijas/dotfiles/commit/81e83202c2a852c2e53f1727f87652054645874a) @ my personal dotfiles repository! Sure, I stripped down parts I don't need on my system, but otherwise it can be copy-patched into oh-my-zsh.

My solution requires `autoload colors` for named colors, which is loaded by `lib/theme-and-appearance.zsh` anyway. I'm looking forward to any feedback.

**Additional context**

None

**Related Issues**

None

ratijas:
## Standards checklist:

## Changes:

1. Convert tabs to spaces
2. Extract Solaris hack
  Also refactors and documents existing behavior.
3. Move PAGER to the env array
4. Use named colors and associative arrays

## Other comments:

@mcornella, please, take a look.

Closes #9436

### DIFF
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -4,6 +4,10 @@ typeset __colored_man_pages_dir="${0:A:h}"
 function colored() {
   local -a environment
 
+  # Prefer `less` whenever available, since we specifically configured
+  # environment for it.
+  environment+=( PAGER="${commands[less]:-$PAGER}" )
+
   # See ./nroff script.
   if [[ "$OSTYPE" = solaris* ]]; then
     environment+=( PATH="${__colored_man_pages_dir}:$PATH" )
@@ -17,7 +21,6 @@ function colored() {
     LESS_TERMCAP_so=$(printf "\e[1;44;33m") \
     LESS_TERMCAP_ue=$(printf "\e[0m") \
     LESS_TERMCAP_us=$(printf "\e[1;32m") \
-    PAGER="${commands[less]:-$PAGER}" \
     $environment \
       "$@"
 }
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -1,39 +1,39 @@
 if [[ "$OSTYPE" = solaris* ]]
 then
-	if [[ ! -x "$HOME/bin/nroff" ]]
-	then
-		mkdir -p "$HOME/bin"
-		cat > "$HOME/bin/nroff" <<EOF
+  if [[ ! -x "$HOME/bin/nroff" ]]
+  then
+    mkdir -p "$HOME/bin"
+    cat > "$HOME/bin/nroff" <<EOF
 #!/bin/sh
 if [ -n "\$_NROFF_U" -a "\$1,\$2,\$3" = "-u0,-Tlp,-man" ]; then
-	shift
-	exec /usr/bin/nroff -u\$_NROFF_U "\$@"
+  shift
+  exec /usr/bin/nroff -u\$_NROFF_U "\$@"
 fi
 #-- Some other invocation of nroff
 exec /usr/bin/nroff "\$@"
 EOF
-		chmod +x "$HOME/bin/nroff"
-	fi
+    chmod +x "$HOME/bin/nroff"
+  fi
 fi
 
 function colored() {
-	command env \
-		LESS_TERMCAP_mb=$(printf "\e[1;31m") \
-		LESS_TERMCAP_md=$(printf "\e[1;31m") \
-		LESS_TERMCAP_me=$(printf "\e[0m") \
-		LESS_TERMCAP_se=$(printf "\e[0m") \
-		LESS_TERMCAP_so=$(printf "\e[1;44;33m") \
-		LESS_TERMCAP_ue=$(printf "\e[0m") \
-		LESS_TERMCAP_us=$(printf "\e[1;32m") \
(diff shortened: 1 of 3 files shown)


plugins/colored-man-pages/colored-man-pages.plugin.zsh
```diff
@@ -1,39 +1,51 @@
...
+
+  # Prefer `less` whenever available, since we specifically configured
+  # environment for it.
+  environment+=( PAGER="${commands[less]:-$PAGER}" )
+
+  # See ./nroff script.
+  if [[ "$OSTYPE" = solaris* ]]; then
+    environment+=(
+      PATH="${__colored_man_pages_dir}:$PATH"
+      _NROFF_U=1
```
mcornella, line 39: This can be removed and fix the nroff binary so that it's 1, since it's always 1.

plugins/colored-man-pages/nroff
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# The whole point of this wrapper is to replace emboldening factor -u0 with
+# the provided $_NROFF_U (which is always 1) under certain circumstances on
+# Solaris.
```
mcornella, line 5: Change the comment to say `-u1` instead of `-u0`.

plugins/colored-man-pages/colored-man-pages.plugin.zsh
```diff
@@ -1,39 +1,51 @@
...
+  done
+
+  # Prefer `less` whenever available, since we specifically configured
+  # environment for it.
+  environment+=( PAGER="${commands[less]:-$PAGER}" )
+
+  # See ./nroff script.
+  if [[ "$OSTYPE" = solaris* ]]; then
+    environment+=(
+      PATH="${__colored_man_pages_dir}:$PATH"
```
mcornella, line 38: I'd prefer if we hardcode `$ZSH/plugins/colored-man-pages` here. I don't love creating a hidden global variable just for this (the path only changes if the plugin is moved to a different location).

mcornella (commented):
Generally +1 to everything. I know using an associative array and the for loop is used so that users can customize it, so we should also document this in the README.

plugins/colored-man-pages/colored-man-pages.plugin.zsh
ratijas: Makes sense.

plugins/colored-man-pages/nroff
ratijas: Wait, what? But it's `-u0` which gets replaced with `-u1`! Or do you mean just get rid of mentioning `_NROFF_U`?

plugins/colored-man-pages/colored-man-pages.plugin.zsh
ratijas: I won't prefer that, and the reason is very simple. As it is, the plugin is perfectly drag-and-dropable into any blank zsh configuration without bringing in the whole oh-my-zsh bloatware. Its only dependency is `autoload colors`.

AFAIK, unfortunately, there is no way to retrieve the file path of a function once it is parsed and loaded, and there are no local variables at the file level (when calling `source ...`, at least); but if you know something that I don't — I'm all ears.

plugins/colored-man-pages/nroff
mcornella: Sorry, "replace `-u0` with `-u1`". Brain fart 😂

ratijas: Makes sense. [x2]

plugins/colored-man-pages/colored-man-pages.plugin.zsh
mcornella: Ok

ratijas:
Also, in the NEWS (or whatever is the closest analogue in oh-my-zsh world) we should instruct Solaris people to delete their `$HOME/bin/nroff` if it was created by this plugin.

NOTE: I am **strongly** against any attempts to remove it automatically upon update or whatever.

ratijas:
Hold on, README is not there yet

mcornella:
Alright, thanks for your hard work!

ratijas:
Thank you too!  It was one of my fastest merges for sure :)

plugins/colored-man-pages/colored-man-pages.plugin.zsh
vimpostor: @ratijas If `autoload colors` is a dependency, shouldn't this plugin autoload that function?

I noticed that recently colored-man-pages broke for me and I don't know why. I can fix it by putting `autoload colors; colors` in my zshrc, but I never had that in my zshrc before and it was working.

I noticed that `plugins/zsh-navigation-tools/n-help` autoloads colors too, so shouldn't this plugin do the same thing so that it works without configuration?

PR #9441, opened 2020-11-13
fix(systemd): remove `sudo` from power-related aliases

scrouthtv:
All systemd power commands I've used so far (e. g. `suspend`, `hibernate`, ...) do work without root access (as in I can call `systemctl suspend` instead of `sudo systemctl suspend`).
This PR would alias those commands to not use sudo.
Also there is no user variant of the power commands so I removed that as well.

## Standards checklist:

## Changes:

 - I changed the systemd module to include a list of subcommands that are power-relevant (suspend, hibernate, hybrid-sleep, poweroff, reboot, restart, suspend)
- I changed the systemd module to remove the aliases from `scu-(any power-relevant command)` to `systemctl --user (any power-relevant command)` and changed the aliases from `sc-(power-relevant command)` which orginally was aliased to `sudo systemctl (prc)` to `systemctl (prc)` as the sudo is not needed. 

## Other comments:

 - hibernate currently does not work on my pc since I upgraded the ram recently, so I wasn't able to test it. If hibernate does require `sudo`, feel free to leave this change out.

### DIFF
--- a/plugins/systemd/README.md
+++ b/plugins/systemd/README.md
@@ -3,7 +3,8 @@
 The systemd plugin provides many useful aliases for systemd.
 
 To use it, add systemd to the plugins array of your zshrc file:
-```
+
+```zsh
 plugins=(... systemd)
 ```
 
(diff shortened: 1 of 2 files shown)


G4Zz0L1:
I think this commit has to be fixed, one of the useful aliases are now removed.
I used to use `scu-restart` to restart a user unit, now it has wrongly moved to a power command which is not.
On this page you can find the difference between the real power management and the unit management systemd command.
https://wiki.archlinux.org/title/Systemd#Using_units (Arch is just an example, systemd is the same everywhere AFAIK) 
At least return the `restart` command to the correct group (sudo and user).
Another suggestion: to enable the use case described above (the power aliases without sudo), one shoud have polkit installed (which maybe not everyone has).
I think there's a difference between how this commands works and how they are aliased here.
The user and sudo group can be called with system or user units alike, so they both should have a `sudo systemctl` and a `systemctl --user` alias.
I can fix this with another PR if it's needed, but the suggested page should add enough context.
Thanks in advance.

scrouthtv:
Yes, that's a miss on my part. I pushed a fix to my branch, however, I can't create a PR for it for some reason.

PR #9443, opened 2020-11-14
Option to disable loading of pyenv-virtualenv in the "pyenv" plugin

pepoluan:
Some people does not like automatic activation of virtualenv, something
that `pyenv virtualenv-init` implements.

The default behavior when PYENV_VIRTUALENV is not set remains the same;
the env var just adds a knob so people who dislike the auto-activation
behavior can disable that.

## Standards checklist:

## Changes:

- Add test for an env var `PYENV_VIRTUALENV` to allow users to (optionally) disable the loading of the pyenv-virtualenv plugin (for pyenv)

## Other comments:

None.

### DIFF
--- a/plugins/pyenv/README.md
+++ b/plugins/pyenv/README.md
@@ -14,3 +14,10 @@ plugins=(... pyenv)
 
 - `pyenv_prompt_info`: displays the Python version in use by pyenv; or the global Python
   version, if pyenv wasn't found.
+
+
+## Options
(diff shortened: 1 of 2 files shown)


pepoluan:
Bump?

mcornella:
Thanks for your contribution! Should be available now with `ZSH_PYENV_VIRTUALENV=false`.

PR #9446, opened 2020-11-16
Add kubectl kga(a)w alias

Preisschild:
## Standards checklist:

## Changes:

- Add alias kgaw to use watch(1) to monitor changes in `kga` 
- Do the same with kgaa

## Other comments:
I often need to monitor multiple things in a namespace while doing another thing in another terminal.
Before i had this alias, i always had to craft the command myself manually.

kgaaw is probably a little less useful since all the information in a big k8s cluster wouldn't fit on the screen, but i added it anyways in case someone ever needs it

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -121,7 +121,9 @@ alias kpf="kubectl port-forward"
 
 # Tools for accessing all information
 alias kga='kubectl get all'
+alias kgaw='watch "kubectl get all"'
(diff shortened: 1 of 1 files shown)


robbyrussell:

The `w` suffix in this plugin means kubectl's own `--watch` flag... `kgpw`, `kgsw`, `kgdw` and the rest all use it. These two shell out to `watch(1)` instead, which is a different thing and isn't installed on macOS by default, so the suffix would mean two different things and a good share of users would get "command not found".

PR #9451, opened 2020-11-17
Three new aliases for git plugin : add deleted files, commit fixup and rebase autosquash

ThamosIO:
## Standards checklist:

## Changes:

- `gad`:  git add only deleted files
- `gcfix`: commit fixup
- `grbas`: rebase autosquash, allows to auto squash fixup commits

## Other comments:

`gcfix` and `grbas`  are used by adding a SHA-1:

```
$ gcfix 63b9724b
$ grbas eef04c62
```


florentVgn:
Hello ! I would be happy if this PR can be resume. I can take care of the PR, make it clean (aliases are not in alphabetical order) and merge it

rwmitchell:
How often do people need to add deleted files that it warrants a 3 letter alias?

Maybe there should be a "git pro" plugin for aliases like this and keep regular "git" with completions and more basic and very frequently used aliases.   There should be a place for it, but the number of current aliases in this plugin is one of the more overwhelming ones.

ThamosIO:
@florentVgn The aliases are now in alphabetical order. Other aliases weren't in order so I corrected it.

@rwmitchell I'm failing to see the problem in aliases abundancy. Concerning the deleted files, I see the aliases not only as a quicker way to type commands but also as a registry of useful commands that people might not know.

rwmitchell:
@ThamosIO Adding unnecessary / unused aliases only pollute the user's workspace and increases the chance of a mistype getting auto-corrected (and not noticed) to something the user absolutely does not want to run.  More aliases also increases the chance of a naming conflicts with other aliases/programs/etc.

Having new aliases just magically appear without the user having a direct need is not useful as the aliases, be definition, are too short to give any sort of clue as to their use.

Having lists of suggested aliases for various programs is a great!  They give ideas for suggesting workflow improvements.  

I have a ton aliases I use and don't want ones I don't - hence my suggestion of having layered levels for git aliases (as git is one of the biggest offenders for having too many aliases).  The newbie git user needs git completions and some basic aliases to get them started.  As their knowledge and use increases, more extensive aliases are great, which is why I suggested a 'gitpro' plugin.

adamchainz:
1. `gad` - I agree with @rwmitchell that adding only deleted files is a bit niche. Normally a vanilla `git add` is sufficient, if `git rm` wasn't used in the first place.

2. `gcfix` - Git allows prefix matching with the `--fixup` option, so you can already type `gc --fix` to achieve the same as the alias. Given that you still need to provide a commit SHA to fixup, I think the new alias wouldn't save much typing. Also, using `--fix` directly allows combining it with other aliases like `gcam`.

3. `grbas`: Autosquash can be turned on by default by enabling [`rebase.autoSquash`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt-rebaseautoSquash). Given that will affect all invocations of rebase from existing aliases, I don't think there's value in adding an alias for autosquash specifically.


@dotkas I don't see how your suggestion is any different from `git commit --amend` (`gc!` alias, and others)

ThamosIO:
>     2. `gcfix` - Git allows prefix matching with the `--fixup` option, so you can already type `gc --fix` to achieve the same as the alias. Given that you still need to provide a commit SHA to fixup, I think the new alias wouldn't save much typing. Also, using `--fix` directly allows combining it with other aliases like `gcam`.

I see your point, but how is that any different from all the aliases already present, such as:
- `alias grba='git rebase --abort'`
- `alias grbc='git rebase --continue'`
- `alias grbi='git rebase --interactive'`
- `alias grbo='git rebase --onto'`
- `alias grbs='git rebase --skip'`

Following your logic, we could get rid of all these, and other similar aliases.

>     3. `grbas`: Autosquash can be turned on by default by enabling [`rebase.autoSquash`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt-rebaseautoSquash). Given that will affect all invocations of rebase from existing aliases, I don't think there's value in adding an alias for autosquash specifically.

Personally, I wouldn’t use `rebase.autoSquash`, I might not want to autosquash everytime. I might want to rebase my branch and autosquash only when my feature is complete.

adamchainz:
> I see your point, but how is that any different from all the aliases already present, such as:

Those save significant typing. On the other hand, the difference between `gcfix` and `gc --fix` is three characters.

Also, I think `git commit --fixup` is not such a commonly used feature.

carlosala:
I have the same point of view. Users should configure `rebase.autoSquash` with the option they use the most and use `--no-autosquash` or `--autosquash` for those other times.
We're also trying not to overcrowd even more the git plugin. That's why we are more picky with the aliases we add nowadays.
Thanks for your contribution!

PR #9453, opened 2020-11-18
feat(aws): respect optional parameters from the AWS CLI config file

RomanDanyk:
## Standards checklist:

## Changes:

If there are `duration_seconds` and/or `role_session_name` parameters in the AWS CLI config file then use them.

## Other comments:

...

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -50,32 +50,39 @@ function acp() {
 
   # First, if the profile has MFA configured, lets get the token and session duration
   local mfa_serial="$(aws configure get mfa_serial --profile $profile)"
+  local sess_duration="$(aws configure get duration_seconds --profile $profile)"
 
   if [[ -n "$mfa_serial" ]]; then
     local -a mfa_opt
-    local mfa_token sess_duration
+    local mfa_token
     echo -n "Please enter your MFA token for $mfa_serial: "
     read -r mfa_token
-    echo -n "Please enter the session duration in seconds (900-43200; default: 3600, which is the default maximum for a role): "
-    read -r sess_duration
+    if [[ -z "$sess_duration" ]]; then
+      echo -n "Please enter the session duration in seconds (900-43200; default: 3600, which is the default maximum for a role): "
+      read -r sess_duration
+    fi
     mfa_opt=(--serial-number "$mfa_serial" --token-code "$mfa_token" --duration-seconds "${sess_duration:-3600}")
 
     # Now see whether we need to just MFA for the current role, or assume a different one
     local role_arn="$(aws configure get role_arn --profile $profile)"
+    local sess_name="$(aws configure get role_session_name --profile $profile)"
 
     if [[ -n "$role_arn" ]]; then
       # Means we need to assume a specified role
       aws_command=(aws sts assume-role --role-arn "$role_arn" "${mfa_opt[@]}")
 
       # Check whether external_id is configured to use while assuming the role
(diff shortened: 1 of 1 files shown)


plugins/aws/aws.plugin.zsh
```diff
@@ -50,32 +50,38 @@ function acp() {
 
   # First, if the profile has MFA configured, lets get the token and session duration
   local mfa_serial="$(aws configure get mfa_serial --profile $profile)"
+  local sess_duration="$(aws configure get duration_seconds --profile $profile)"
 
   if [[ -n "$mfa_serial" ]]; then
     local -a mfa_opt
-    local mfa_token sess_duration
```
maksyms, line 56: Doesn't this line have to be `local mfa_token`, as otherwise the var isn't declared? I understand `sess_duration` is declared higher up.

maksyms (commented):
Thanks a lot for your contribution, it is a great addition! I've left a comment in the review. I think once that is fixed, I'll do a quick test and we can merge.

plugins/aws/aws.plugin.zsh
RomanDanyk: Oh, yes it should be here. My bad.

maksyms:
Fantastic, thanks, @RomanDanyk!

@mcornella This is a great change, making the use of the `acp` function even easier by reading some additional parameters from the `.aws/config` file. Good to merge.

mcornella:
According to [this comment](https://superuser.com/questions/1260311/is-there-a-way-to-list-available-configured-aws-cli-profiles#comment2347391_1544640) the `aws configure list-profiles` command was introduced in awscli v2.x. Is that widespread enough to enable the change? If so, is the output compatible with parsing the `aws_profiles` function into an array, like we do in `asp`, `acp` and the completion function for these?

plugins/aws/aws.plugin.zsh
```diff
@@ -13,7 +13,7 @@ function asp() {
   local -a available_profiles
   available_profiles=($(aws_profiles))
   if [[ -z "${available_profiles[(r)$1]}" ]]; then
-    echo "${fg[red]}Profile '$1' not found in '${AWS_CONFIG_FILE:-$HOME/.aws/config}'" >&2
```
mcornella, line 16: Why remove the quotes? They help delimit where the name of the profile ends, in case there are spaces in them for example.

RomanDanyk: I'll revert this change.

RomanDanyk:
I forgot to check if `list-profiles` command works with awscli v1.x. You are right, it's better to leave `grep | sed` for backward compatibility.

maksyms:
> According to [this comment](https://superuser.com/questions/1260311/is-there-a-way-to-list-available-configured-aws-cli-profiles#comment2347391_1544640) the `aws configure list-profiles` command was introduced in awscli v2.x. Is that widespread enough to enable the change? If so, is the output compatible with parsing the `aws_profiles` function into an array, like we do in `asp`, `acp` and the completion function for these?

At the point of approving changes, `aws configure list-profiles` wasn't in the commit - and yes, I completely agree that grep/sed for now is still a better option.

RomanDanyk:
@maksyms, @mcornella 
It was my bad, sorry. I've removed all changes not related to the PR subject.

mcornella:
Not bad at all, it could be a nice change if we can assert it works for everybody, especially since it also supports the credentials file. Maybe open a new PR with that so we can discuss it?

PR #9477, opened 2020-11-27
shar format support for extract plugin

agkozak:
## Standards checklist:

## Changes:

- The `.shar` format is still occasionally found in the wild, and it is quite easy to add support for it to the `extract` plugin.

## Other comments:

...


mcornella:
I don't know, it doesn't look safe, but at the same time I don't know if this is that common to support it.

agkozak:
The `.shar` format is a relic of a more innocent time. On second thought, I agree - let's not support it.

PR #9484, opened 2020-12-02
fix(git): silently fail in `git_main_branch` if not in a git repo

jdheyburn:
## Standards checklist:

## Changes:

- `git_main_branch` silently returns when executed outside of a git repo. 

## Other comments:

Mimics the same behaviour as `git_current_branch`
...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -31,6 +31,10 @@ function work_in_progress() {
 
 # Check if main exists and use instead of master
 function git_main_branch() {
+  local ref
+  ref=$(__git_prompt_git symbolic-ref --quiet HEAD 2> /dev/null)
+  local ret=$?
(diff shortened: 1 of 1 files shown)


mcornella:
Would silencing stderr on the `git show-ref` command be enough?

jdheyburn:
@mcornella fair point - but if we're running the command in a legit git repo you may want the stderr for debugging. I only took the approach another function had for consistency.

mcornella:
Changed it to [`git rev-parse --git-dir`](https://stackoverflow.com/a/2180367/5798232) which is the proper way to do it. Works for me but test it first before merging.

jdheyburn:
> Changed it to [`git rev-parse --git-dir`](https://stackoverflow.com/a/2180367/5798232) which is the proper way to do it. Works for me but test it first before merging.

Nice one 👍  Confirmed it works as expected on my end

PR #9502, opened 2020-12-11
feat(plugins): add genpass plugin with 3 distinct password generators

atoponce:
## Standards checklist:

## Changes:

- 
    Add genpass plugin providing `gen-apple-pass`, `gen-monkey-pass`, and `gen-xkcd-pass` password generators

## Other comments:
These password generators help minimize the mistake people make thinking they can "design" good passwords in the shell. Instead, each password generator provided by this plugin generates passwords with at least a 128-bit security margin, suitable even for cryptographic keys. The user can provide a numerical argument to generate more passwords if needed. No external 3rd party software is required.
...

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,6 @@
 # Plugin owners
 plugins/aws/                        @maksyms
+plugins/genpass/                    @atoponce
 plugins/git-lfs/                    @vietduc01100001
 plugins/gitfast/                    @felipec
 plugins/sdk/                        @rgoldberg
--- a/plugins/genpass/README.md
+++ b/plugins/genpass/README.md
@@ -15,46 +15,46 @@ Install:
 
     plugins=(... genpass)
 
-### gen-apple-pass
+### gennpass-apple
 Generates a pronounceable pseudoword passphrase of the "cvccvc" consonant/vowel
 syntax, inspired by [Apple's iCloud Keychain password generator][1]. Each
 pseudoword has exactly 1 digit placed at the edge of a "word" and exactly 1
 capital letter to satisfy most password security requirements.
 
-    % gen-apple-pass 
+    % gennpass-apple
     gelcyv-foqtam-fotqoh-viMleb-lexduv-6ixfuk
 
-    % gen-apple-pass 3
+    % gennpass-apple 3
     japvyz-qyjti4-kajrod-nubxaW-hukkan-dijcaf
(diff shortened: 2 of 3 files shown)


atoponce:
This replaces #9473 where I clobbered my Git repository, then by extension the PR.

zacchiro:
I've just tested this (deploying it locally via `omz pr test`) and it works great.
I also find it very useful functionality in general — but, disclaimer, I'm no crypto expert so I can't review the implementation on that front (Aaron is though! :-))
As a minor quirk, I wonder if you could consider renaming the functions from `gen-TYPE-pass` to `gen-pass-TYPE` (or maybe even `genpass-TYPE` while you're at it). Rationale would be being more TAB-completion friendly, as IME the _type_ of password one wants to generate doesn't necessarily come to finger immediately, while the fact you want to generate any password (`genpass-TAB`) does.

Thanks for this anyway!

atoponce:
> As a minor quirk, I wonder if you would consider renaming the functions from `gen-TYPE-pass` to `gen-pass-TYPE` (or maybe even `genpass-TYPE` while you're at it). Rationale would be being more TAB-completion friendly, as IME the *type* of password one wants to generate doesn't necessarily come to finger immediately, while the fact you want to generate any password (`genpass-TAB`) does. `genpass-TYPE` would also be more consistent with the plugin name, namespacing all its functions behind that prefix.

Ah, fair point. I agree the `genpass-TYPE` naming scheme is both more consistent with the name of the plugin itself and more tab-completion friendly.

zacchiro:
> Ah, fair point. I agree the `genpass-TYPE` naming scheme is both more consistent with the name of the plugin itself and more tab-completion friendly.

Thanks!

Typo in the commit you just pushed: it's now `gennpass-*` (double "n") instead of `genpass-*`

zacchiro:
Perfect now, thanks !

atoponce:
Yeah. Sorry about that. I tested `gen<TAB>` to verify execution before pushing, but failed to notice that two "n"s were completed. Heh

mcornella (commented):
LGTM, but can you add some comments to know what it's doing? Mainly in `genpass-apple` and `genpass-xkcd`, those `log` go over my head.

mcornella:
I made some light changes, if you agree I'll merge this right away.

atoponce:
Yeah, looks great. Thanks!

mcornella:
Awesome, thanks!

PR #9516, opened 2020-12-14
command fix(genpass): Add compatibility for MacOS paste command in `genpass-xkcd`

mestizo:
"paste" on MacOS requires a '-' to signify that the standard input is used.  Without the '-' character, the command errors out.

## Standards checklist:

## Changes:

- Check if command is running under MacOS (darwin*) and use a mac compatible format instead.

## Other comments:

"paste" on MacOS requires a '-' to signify that the standard input is used.  Without the '-' character, the command errors out.


atoponce:
There is no need for the `$OSTYPE` check. GNU `paste(1)` can also use "-" for `[FILE]`. Per the manpage:

    SYNOPSIS
           paste [OPTION]... [FILE]...
    DESCRIPTION
           ...(snip)...
           With no FILE, or when FILE is -, read standard input.

atoponce:
Thanks @mcornella

PR #9520, opened 2020-12-15
fix(genpass): improve performance and usability and fix bugs

romkatv:
This is a rewrite of `genpass` plugin. The new implementation has better performance and usability and fixes several bugs. It is backward-compatible.

## Checklist

## Bugs

The following bugs have been fixed:

- All generators ignored errors from external commands. For example, if `/usr/share/dict/words` was unreadable, `genpass-xkcd` would print "0-" as a password and return success.
- `genpass-xkcd` was generating passwords with less than 128 bits of security margin in contradiction to documentation. The smaller the dictionary size, the weaker the passwords it was generating. For a dictionary with 27 words, `genpass-xkcd` was generating passwords with 93 bits of security margin (`log2(27!)`).
- The source of random data used by `genpass-xkcd` was not cryptographically secure in contradiction to documentation. See: https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html
- `genpass-apple` could generate a password with non-ascii characters depending on user locale. For example, passwords could contain 'İ' for users with Turkish locale.
- `genpass-apple` didn't work with `ksh_arrays` shell option.
- `genpass-xkcd` was printing spurious errors with `ksh_arrays` shell option.
- `genpass-xkcd` was producing too short (weak) or too strong (long) and/or printing errors when `IFS` was set to non-default value.
- All generators were printing fewer passwords than requested and returning success when passed a very large number as an argument.

## Usability

Generators are now implemented as self-contained executable files. They can be invoked from scripts with no additional setup.

Generators no longer depend on external commands. The only dependencies are `/dev/urandom` and, for `genpass-xkcd`, `/usr/share/dict/words`.

All generators used to silently ignore all arguments after the first, and the first argument if it wasn't a number. For example, both `genpass-apple -2` and `genpass-apple -n 2` were generating one password and not printing any errors. Now these print an error and fail.

## Performance

The time it takes to load the plugin has been greatly reduced. This translates into faster zsh startup when the plugin is enabled.

Incidentally, two generators out of three have been sped up to a large degree while one generator (`genpass-xkcd`) has gotten slower. This is unlikely to matter one way or another unless generating a very large number of passwords. In the latter case `genpass-xkcd` is now also faster than it used to be.

The following table shows benchmark results from Linux x86-64 on i9-7900X. The numbers in the second and third columns show how many times a given command can be executed per second. Higher numbers are better.

command                     | before (Hz) | after (Hz) | speedup |
----------------------------|------------:|-----------:|--------:|
`source genpass.plugin.zsh` |        4810 |      68700 |  +1326% |
`genpass-apple`             |        30.3 |        893 |  +2846% |
`genpass-monkey`            |         203 |       5290 |  +2504% |
`genpass-xkcd`              |        34.4 |       14.5 |    -58% |
`genpass-xkcd 1000`         |       0.145 |      0.804 |   +454% |


atoponce:
> * All generators silently ignored the argument if it wasn't a number. For example, `genpass-apple -2` was generating one password and not printing any errors.
> * All generators silently ignored extra arguments. For example, `genpass-apple -n 2` was generating one password and not printing any errors.

Both by design. In the former, generating "-2" or "abc" number of passwords is nonsensical. In the latter, the documentation clearly identified the only argument to supply is a positive number. Regardless, your `usage: genpass-apple [NUM]` on STDERR is more elegant, indeed.

>* `genpass-xkcd` was generating passwords with less than 128 bits of security margin in contradiction to documentation. The smaller the dictionary size, the weaker the passwords it was generating. For a dictionary with 27 words, `genpass-xkcd` was generating passwords with 93 bits of security margin (log2(27!)).

Ah, fair point. `shuf(1)` is indeed shuffling the dictionary, without the possibility of repeats. Thanks for catching that.

> * The source of random data used by `genpass-xkcd` was not cryptographically secure in contradiction to documentation. See: https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html

GNU `shuf(1)` can use an external randomness source via `--random-source=file`, but internally it uses the ISAAC CSPRNG using `/dev/urandom` as a seed. See https://www.maizure.org/projects/decoded-gnu-coreutils/shuf.html which breaks down the source code. GNU `shuf(1)` is internally cryptographically secure.

Regardless, I was working on removing the `shuf(1)` dependency, relying strictly on `/dev/urandom` to pick elements from the dictionary.

Regarding your `genpass-apple` patch, you have the following source:

    # Replace the last character in one of the words with a
    # random digit.
    -$0-rand $#digits || return
    local digit=$digits[REPLY]
    -$0-rand $#words || return
    words[REPLY]=${words[REPLY]%?}$digit

The digit can also randomly be the first character in one of the pseudowords. IE "Dvccvc" or "cvccvD".

romkatv:
@atoponce Thanks for the rewview!

> > * All generators silently ignored the argument if it wasn't a number.
> > * All generators silently ignored extra arguments.
> 
> Both by design. [...] Regardless, your `usage: genpass-apple [NUM]` on STDERR is more elegant, indeed.

I've updated the PR description. This change is now listed under usability improvements rather than bug fixes. I've also added one more entry to the bug fixes while at it:

- All generators were printing fewer passwords than requested and returning success when passed a very large number as an argument.

(This is unlikely to matter to anyone.)

> The digit can also randomly be the first character in one of the pseudowords. IE "Dvccvc" or "cvccvD".

Thanks! I've missed that. Fixed.

atoponce:
> Thanks for the rewview!

Thanks for the improvements, even if it is a complete rewrite! Right now I'm verifying the implemented RNG is uniform, but other than that, it looks good.

atoponce:
@romkatv Can you explain why in `genpass-xkcd` the upper limit is 31 bits instead of 32?

Also, can you explain this line? It's executed 4 times for every word, but I can't convince myself why:

    (( rnd = (~(1 << 23) & rnd) << 8 | #c ))

romkatv:
> @romkatv Can you explain why in `genpass-xkcd` the upper limit is 31 bits instead of 32?

Zsh can be compiled without 64-bit numbers. Numbers in zsh are signed, so there may be only 31 value bits per number (plus one bit for the sign).

I could branch on 32/64-bit support but I figured there is no need for it. It's unlikely that anyone will want to generate passwords with a dictionary containing over 2 billion words. If someone attempts it, they'll get a clear error (although they will have to wait for it for a very long time).

> Also, can you explain this line? It's executed 4 times for every word, but I can't convince myself why:
> 
> ```
> (( rnd = (~(1 << 23) & rnd) << 8 | #c ))
> ```

The following piece of code generates a random number in `[0, 1 << 31)` with uniform distribution:

```zsh
local -i rnd=0
repeat 4; do
  sysread -s1 c || return
  (( rnd = (~(1 << 23) & rnd) << 8 | #c ))
done </dev/urandom
print -r -- $rnd
```

If we knew that zsh uses 64-bit numbers, the code could be as simple as this:

```zsh
local -i rnd=0
repeat 4; do
  sysread -s1 c || return
  (( rnd = rnd << 8 | #c ))
done </dev/urandom
(( rnd &= ~(1 << 31) ))
print -r -- $rnd
```

However, if zsh has 32-bit numbers, this code has implementation-defined behavior under c89. To stay within the realm of well-defined behavior, we need to ensure that the result of `<<` is always in `[0, 1 << 31)`. To achieve that, the code is clearing the bit that would become the sign bit after the shift.

Does it make sense?

romkatv:
FYI: I've made a small change to simplify the code in `genpass-apple`:

```diff
@@ -58,11 +58,7 @@ zmodload zsh/system zsh/mathfunc || return
     -$0-rand $#digits || return
     local digit=$digits[REPLY]
     -$0-rand $((2 * $#words)) || return
-    if (( REPLY % 2 )); then
-      pwd[REPLY/2*7+2*(REPLY%2)-1]=$digit
-    else
-      pwd[REPLY/2*7-1]=$digit
-    fi
+    pwd[REPLY/2*7+2*(REPLY%2)-1]=$digit
 
     # Convert one lower-case character to upper case.
     while true; do
```

I've also updated the description of the commit (and force-pushed) and of the PR. I've added one more bug to the list of fixes:

- `genpass-xkcd` was producing too short (weak) or too strong (long) passwords and/or printing errors when `IFS` was set to non-default value.

atoponce:
> However, if zsh has 32-bit numbers, this code has implementation-defined behavior under c89. To stay within the realm of well-defined behavior, we need to ensure that the result of << is always in [0, 1 << 31). To achieve that, the code is clearing the bit that would become the sign bit after the shift.

Oh, I see. Yeah, I'm good. Thanks for that explanation.

mcornella:
Is this ready then?

atoponce:
Yeah, it looks good. Thanks @romkatv for the rewrite. It's both humbling and inspiring.

mcornella:
Thanks @romkatv!

romkatv:
Thanks for the super speedy reviewing and merging, @atoponce and @mcornella :100:

PR #9529, opened 2020-12-20
refactor(install.sh): fix static analysis warnings

leagris:
## Standards checklist:

## Changes:

Clear all warnings and errors raised by https://shellcheck.net static analysis.

- Replace non-POSIX shell use of `$OSTYPE` by POSIX compliant check on `uname -o`.
- Move variables out of `printf` format string.
- Refactor/simplify string formatters for error and underline.
- Fix expansion of arguments to a single string `$*` rather than individual elements `$@` within the error and underline formatters.

## Other comments:

none

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -214,7 +214,7 @@ EOF
     # Get the path to the right zsh binary
     # 1. Use the most preceding one based on $PATH, then check that it's in the shells file
     # 2. If that fails, get a zsh path from the shells file, then check it actually exists
-    if ! zsh=$(which zsh) || ! grep -qx "$zsh" "$shells_file"; then
+    if ! zsh=$(command -v zsh) || ! grep -qx "$zsh" "$shells_file"; then
       if ! zsh=$(grep '^/.*/zsh$' "$shells_file" | tail -1) || [ ! -f "$zsh" ]; then
         fmt_error "no zsh binary found or not present in '$shells_file'"
         fmt_error "change your default shell manually."
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -103,7 +103,7 @@ setup_ohmyzsh() {
     exit 1
   }
 
-  ostype=$(uname -o)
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -102,7 +103,8 @@ setup_ohmyzsh() {
     exit 1
   }
 
-  if [ "$OSTYPE" = cygwin ] && git --version | grep -q msysgit; then
+  ostype=$(uname)
```
mcornella, line 106: Does it complain without this? `$OSTYPE` is guaranteed to exist in cygwin which is the only reason we're doing this.

leagris:
```sh
  if [ "$OSTYPE" = cygwin ] && git --version | grep -q msysgit; then
```
> `        ^-----^ SC2039: In POSIX sh, OSTYPE is undefined.`

Since the shebang is pointing to POSIX sh, `OSTYPE` cannot be used:
```sh
#!/bin/sh
```

About detecting Cygwin because it has this OSTYPE available, it has because its default sh shell points to bash. But since `ostype=$(uname)` is a portable POSIX compliant replacement, no reason to not do this instead.

mcornella: So it'll be an empty string and the condition will fail. Is it really worth it for the extra fork?

mcornella: Actually don't reply to that, it doesn't matter. Ready to merge isn't it?

leagris: If forking was in a loop, I'd worry about performance. Otherwise, shell is not a general programming language but a good choice for portable installation scripts.

Yes ready to merge from my point, but I have no say, except that:
I tested my version of the script locally on a Linux Ubuntu box and it did not show any issue.
Now I did not test it on anything else, not even exotic BSD or antique SunOS or Xenix, or Minix ^^

mcornella:
Thanks Léa!

PR #9530, opened 2020-12-20
feat(lib): show commit hash in `git_prompt_info`

hossam-magdy:
## Standards checklist:

## Changes:

- Use `ZSH_THEME_GIT_SHOW_COMMITHASH=1` 

Like `ZSH_THEME_GIT_SHOW_UPSTREAM=1`, utilizing `ZSH_THEME_GIT_SHOW_COMMITHASH=1` to show commit hash of the current HEAD commit alongside the `ref` name.

I.e: `master[394774b]` instead of `master`

This can be very useful in case of mistakenly losing or modifying local commits, as the hash will be easily retrievable from the prompt, and hence can help recovering from lost commits, or also revert amended ones.

It can help recovering from commands or use cases like:
- `git reset --hard HEAD~1`
- `git commit --amend`

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -29,7 +29,14 @@ function git_prompt_info() {
     && upstream=" -> ${upstream}"
   fi
 
-  echo "${ZSH_THEME_GIT_PROMPT_PREFIX}${ref}${upstream}$(parse_git_dirty)${ZSH_THEME_GIT_PROMPT_SUFFIX}"
+  # Use global ZSH_THEME_GIT_SHOW_COMMITHASH=1 for including current HEAD commit short hash
+  local commitHash
+  if (( ${+ZSH_THEME_GIT_SHOW_COMMITHASH} )); then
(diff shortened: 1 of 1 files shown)


robbyrussell:

The shape is right: it's opt-in, it mirrors the `ZSH_THEME_GIT_SHOW_UPSTREAM` block directly above it, and the extra `git rev-parse --short HEAD` measures around 7.5ms here, which is roughly what the prompt's existing `git status` costs and it's on the async path now anyway. So there's no performance objection.

Four things it needs before it can go in:

- a rebase... `git_prompt_info` was renamed to `_omz_git_prompt_info` in #12257 and the echo line gained `%` quoting since
- the hash quoted the same way, `${commitHash//\%/%%}`, so a `%` can't be read as a prompt escape
- the `[...]` around the hash made themeable rather than hardcoded, following how the other markers here work
- a detached-HEAD case: `ref` is already the short SHA there, so today you'd get `394774b[394774b]`

If you'd rather not pick this back up after this long, that's completely fine... say so, or say nothing, and I'll finish it with you credited.

PR #9532, opened 2020-12-21
refactor: use +commands in check_for_upgrade

SukkaW:
## Standards checklist:

## Changes:

- refactor: use "+commands" in check_for_upgrade

## Other comments:

Use `$+commands[git]` instead of `command -v git`.


mcornella:
The condition is wrong, it should be `(( ! ${+commands[git]} ))` (if not exists). That said, I see no reason to merge this since both forms are correct. Thanks for your contribution though!

PR #9537, opened 2020-12-24
Add function wrapper around compinit logic in oh-my-zsh.sh

mattmc3:
By adding run_compinit as a function wrapper that first checks for its existence, it then becomes possible to override the run_compinit function by defining it before OMZ is sourced.

## Standards checklist:

## Changes:

- Moves compinit logic into lib allowing advanced users to override compinit logic with standard $ZSH_CUSTOM methods
- Fixes longstanding issue of lib/compfix.zsh being double loaded
- Simplifies oh-my-zsh.sh into core functionality, without changing its logic
- Minor whitespace formatting changes for consistency

## Other comments:
See issue for discussion: https://github.com/ohmyzsh/ohmyzsh/issues/9258
This is a much simpler and lower risk change, and replaces the need for https://github.com/ohmyzsh/ohmyzsh/pull/9257
Be sure to view the diff ignoring whitespace changes, as that better shows the simplicity of this change (https://github.com/ohmyzsh/ohmyzsh/pull/9537/files?diff=unified&w=1)

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -95,7 +95,7 @@ $zcompdump_fpath
 EOF
   fi
 }
-if [[ "$DISABLE_COMPINIT" != "true" ]]; then
+if [[ "$ZSH_DISABLE_COMPINIT" != "true" ]]; then
   run_compinit
 fi
 
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -95,9 +95,9 @@ $zcompdump_fpath
 EOF
   fi
 }
-
-run_compinit
-
+if [[ "$DISABLE_COMPINIT" != "true" ]]; then
+  run_compinit
+fi
 
 # Load all of the config files in ~/oh-my-zsh that end in .zsh
 # TIP: Add files you don't want in git to .gitignore
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -59,40 +59,44 @@ if [ -z "$ZSH_COMPDUMP" ]; then
   ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
-# Construct zcompdump OMZ metadata
-zcompdump_revision="#omz revision: $(builtin cd -q "$ZSH"; git rev-parse HEAD 2>/dev/null)"
-zcompdump_fpath="#omz fpath: $fpath"
+(( $+functions[run_compinit] )) ||
+function run_compinit() {
+  # Construct zcompdump OMZ metadata
+  local zcompdump_revision="#omz revision: $(builtin cd -q "$ZSH"; git rev-parse HEAD 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


seivan:
If I understood it correctly, you either add your own implementation from scratch that `omz`calls, or you just use the pre-defined `run_compinit`? 

What if I just want the default behaviour of `run_compinit`, but I need to decide when it's called? 
There's  a lot of logic in there, related to manifest, deleting old stuff, and I think it's all good, so it feels weird having to reimplement that whole thing. 

Could there be a flag to indicate to omz not to `run_compinit`, and let me manually call it later down the line?

mattmc3:
> If I understood it correctly, you either add your own implementation from scratch that omzcalls, or you just use the pre-defined run_compinit?

Yes, this implementation lets you define your own run_compinit function prior to sourcing OMZ, and if you did it calls that.

> What if I just want the default behaviour of run_compinit, but I need to decide when it's called? Could there be a flag...

Yes, that's a great callout. I'll add a simple commit that introduces a variable so that you could use OMZ's run_compinit, but call it on your own timeframe.

> so it feels weird having to reimplement that whole thing

That's how overriding anything in the `lib/` folder works, so it's not out of band with how OMZ already handles things. The alternative implementation of this feature would be to move the compinit logic to the `lib\` folder, but that is a much bigger change and introduces some risk. Ideally, this logic moves out of oh-my-zsh.sh into lib someday, but that requires some more thought and input by the core devs.

seivan:
@mattmc3 Any progress on this?

I got a few suggestions, some intrusive and some less, lets start with the latter. 

I think this part should also be move to the function: 
```sh
if [[ "$OSTYPE" = darwin* ]]; then
  # macOS's $HOST changes with dhcp, etc. Use ComputerName if possible.
  SHORT_HOST=$(scutil --get ComputerName 2>/dev/null) || SHORT_HOST=${HOST/.*/}
else
  SHORT_HOST=${HOST/.*/}
fi
```

There were two plugins that used `SHORT_HOST` but they also default to `${(%):-%m}`. It's not super necessary. 


Then there's the discussion of opting out while still retaining the logic of handling OMZ metadata generation and comparison, which is my main gripe with the current state of the PR. 
I want to opt out, to make sure `compinit` is called last, but I want to retain the logic, such as naming convention, revision control via metadata and so on. 

To explain further why this is important (mostly thinking out loud): 
Right now OMZ uses metadata to decide if it needs new a new `compdump` or not.
ZSH already has a builtin feature for this, that uses the number of files at the top, however that doesn't take into account changes made. There could be changes, but the number of files remained the same. That would causes issues and is why the OMZ metadata logic is there. 


So my suggestion is to 
add functions for these two
- `# Construct zcompdump OMZ metadata`
- `# Append zcompdump metadata if missing`

This would let people opt out OMZ calling `compinit` but use its logic in their own `.zshrc` to keep the same features.

mattmc3:
@seivan I switched to Prezto for performance reasons and because it was easier to get PRs merged, so these changes will need to be done by someone else if it's still desirable for the OMZ project.

PR #9544, opened 2020-12-25
Cleanup git aliases

ratijas:
## Standards checklist:

## Changes:

- fix(git): change git-diff aliases
- fix(git): update deprecation notice for git aliases
- fix(git): replace remaining current_branch with git_current_branch

## Other comments:

Removal of an existing alias, albeit a duplicate, might not be desirable for old users. However it's good for new ones as it would avoid some confusion. And you shouldn't mindlessly update your zsh config without reviewing changes anyway.

### DIFF
--- a/plugins/git-lfs/README.md
+++ b/plugins/git-lfs/README.md
@@ -19,6 +19,6 @@ plugins=(... git-lfs)
 
 ## Functions
 
-| Function | Command                                         |
-| :------- | :---------------------------------------------- |
-| `gplfs`  | `git lfs push origin "$(current_branch)" --all` |
+| Function | Command                                             |
+| :------- | :-------------------------------------------------- |
+| `gplfs`  | `git lfs push origin "$(git_current_branch)" --all` |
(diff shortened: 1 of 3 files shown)


ratijas:
Rebased onto latest master.

ratijas:
Looking at it now, I'd also rename `git_main_branch` to `git_default_branch`, because that's what it is. Maybe there's even a git API for that?

robbyrussell:
Thanks for this, and I'm sorry it waited so long for a reply.

You were right about the docs. The README still documents `$(current_branch)` for `ggl`, `ggf`, `ggfl` and `ggp`, plus `gplfs` in the git-lfs README, and that function doesn't exist any more... they all call `git_current_branch`. That's worth fixing and I'm going to reimplement it against master and credit you in the commit. Nothing needed from you, but if you'd rather do it yourself, say so and it's yours.

The one thing I'm going to leave out is removing `gds`. It duplicates `gdca`, but it's in a lot of people's muscle memory and we don't have a soft-deprecation path for aliases. We reverted a similar removal last December in #13485 for that reason.

PR #9546, opened 2020-12-25
refactor(archlinux): fix name & scripts; reorder aliases

ratijas:
## Standards checklist:

## Changes:

* Fix naming.
* Pacman goes first. AUR helpers, sorted alphabetically, follow it.
* Add footnote to explain identically named *upg aliases.
* Convert upgrade function to aliases, put them alongside other aliases.
* Rewrite paclist function: it was unreliable and over-complicated.
* Add --help to pacweb function instead of silently failing.

## Other comments:

This refactoring is a prerequisite for #9467

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -1,3 +1,127 @@
+#######################################
+#               Pacman                #
+#######################################
+
+# Pacman - https://wiki.archlinux.org/index.php/Pacman_Tips
+alias pacupg='sudo pacman -Syu'
+alias pacin='sudo pacman -S'
+alias pacins='sudo pacman -U'
+alias pacre='sudo pacman -R'
+alias pacrem='sudo pacman -Rns'
+alias pacrep='pacman -Si'
+alias pacreps='pacman -Ss'
+alias pacloc='pacman -Qi'
+alias paclocs='pacman -Qs'
+alias pacinsd='sudo pacman -S --asdeps'
+alias pacmir='sudo pacman -Syy'
+alias paclsorphans='sudo pacman -Qdt'
+alias pacrmorphans='sudo pacman -Rs $(pacman -Qtdq)'
+alias pacfileupg='sudo pacman -Fy'
+alias pacfiles='pacman -F'
+alias pacls='pacman -Ql'
+alias pacown='pacman -Qo'
+alias upgrade='sudo pacman -Syu'
+
+if (( $+commands[abs] && $+commands[aur] )); then
+  alias pacupd='sudo pacman -Sy && sudo abs && sudo aur'
+elif (( $+commands[abs] )); then
+  alias pacupd='sudo pacman -Sy && sudo abs'
+elif (( $+commands[aur] )); then
+  alias pacupd='sudo pacman -Sy && sudo aur'
+else
+  alias pacupd='sudo pacman -Sy'
+fi
+
+function paclist() {
+  # Based on https://bbs.archlinux.org/viewtopic.php?id=93683
+  pacman -Qqe | \
+    xargs -I '{}' \
+      expac "${bold_color}% 20n ${fg_no_bold[white]}%d${reset_color}" '{}'
+}
+
+function pacdisowned() {
+  local tmp db fs
+  tmp=${TMPDIR-/tmp}/pacman-disowned-$UID-$$
+  db=$tmp/db
+  fs=$tmp/fs
+
+  mkdir "$tmp"
+  trap 'rm -rf "$tmp"' EXIT
+
+  pacman -Qlq | sort -u > "$db"
+
+  find /bin /etc /lib /sbin /usr ! -name lost+found \
+    \( -type d -printf '%p/\n' -o -print \) | sort > "$fs"
+
+  comm -23 "$fs" "$db"
+}
+
+alias pacmanallkeys='sudo pacman-key --refresh-keys'
+
+function pacmansignkeys() {
+  local key
+  for key in $@; do
+    sudo pacman-key --recv-keys $key
+    sudo pacman-key --lsign-key $key
+    printf 'trust\n3\n' | sudo gpg --homedir /etc/pacman.d/gnupg \
+      --no-permission-warning --command-fd 0 --edit-key $key
+  done
+}
+
+if (( $+commands[xdg-open] )); then
+  function pacweb() {
+    if [[ $# = 0 || "$1" =~ '--help|-h' ]]; then
+      local underline_color="\e[${color[underline]}m"
+      echo "$0 - open the website of an ArchLinux package"
+      echo
+      echo "Usage:"
+      echo "    $bold_color$0$reset_color ${underline_color}target${reset_color}"
+      return 1
+    fi
+
+    local pkg="$1"
+    local infos="$(LANG=C pacman -Si "$pkg")"
+    if [[ -z "$infos" ]]; then
+      return
+    fi
+    local repo="$(grep -m 1 '^Repo' <<< "$infos" | grep -oP '[^ ]+$')"
+    local arch="$(grep -m 1 '^Arch' <<< "$infos" | grep -oP '[^ ]+$')"
+    xdg-open "https://www.archlinux.org/packages/$repo/$arch/$pkg/" &>/dev/null
+  }
+fi
+
+#######################################
+#             AUR helpers             #
+#######################################
+
+if (( $+commands[pacaur] )); then
+  alias paupg='pacaur -Syu'
+  alias pasu='pacaur -Syu --noconfirm'
+  alias pain='pacaur -S'
+  alias pains='pacaur -U'
+  alias pare='pacaur -R'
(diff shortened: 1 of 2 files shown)


ratijas:
Also, yaourt seems to be unmaintained, archived and deprecated. I wonder how many people are still using it? If not much, them maybe we could remove support for it? At least, popularity and votes are pretty low: https://aur.archlinux.org/packages/yaourt/.

ratijas:
Bump.

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -1,3 +1,130 @@
...
+
+function paclist() {
+  # Based on https://bbs.archlinux.org/viewtopic.php?id=93683
+  pacman -Qqe | \
+    xargs -I '{}' \
+      expac "${bold_color}% 20n ${fg_no_bold[white]}%d${reset_color}" '{}'
+}
+
+function pacdisowned() {
+  tmp=${TMPDIR-/tmp}/pacman-disowned-$UID-$$
```
mcornella, line 43: This removed the `local` line.

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -1,3 +1,130 @@
...
+  alias pacupd='sudo pacman -Sy && sudo aur'
+else
+  alias pacupd='sudo pacman -Sy'
+fi
+
+function paclist() {
+  # Based on https://bbs.archlinux.org/viewtopic.php?id=93683
+  pacman -Qqe | \
+    xargs -I '{}' \
+      expac "${bold_color}% 20n ${fg_no_bold[white]}%d${reset_color}" '{}'
```
mcornella, line 39: Is `expac` installed by default?

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -1,3 +1,130 @@
...
+    \( -type d -printf '%p/\n' -o -print \) | sort > "$fs"
+
+  comm -23 "$fs" "$db"
+}
+
+function pacmanallkeys() {
+  curl -s https://www.archlinux.org/people/{developers,trustedusers}/ | \
+    awk -F\" '(/pgp.mit.edu/) { sub(/.*search=0x/,""); print $1}' | \
+    xargs sudo pacman-key --recv-keys
+}
```
mcornella, line 62: We fixed this function in #9569, please rebase.

mcornella (approved):
+1 with some comments

plugins/archlinux/archlinux.plugin.zsh
ratijas: No, it is in community repository. Although, I got it as a dependency of pacaur. Anyway, it is a very handy utility to query over pacman database.

Some moderator person on aforementioned thread replied that the script that I replaced did not make any sense, and you should be using `checkupdates` from `community/pacman-contrib` if you absolutely must. (before he proceeded to trash-talk oh-my-zsh, but that's a story for another time).

Honestly, this function is just-for-fun, and I don't see any real-world use cases, except for teaching newbies how to script Arch.

plugins/archlinux/archlinux.plugin.zsh
ratijas: My bad. Lost it while merging and rebasing.

ratijas:
Fixed `local` line. Now rebasing... hold on.

ratijas:
Manually cherry-picked...

ratijas:
Aaaand... I think it's done.

cascadingstyletrees:
Which AUR helper is chosen when using `$ upgrade`?

plugins/archlinux/archlinux.plugin.zsh
mcornella: +1

ratijas:
> Which AUR helper is chosen when using `$ upgrade`?

The one that you have, I can tell you that much. If you have more than one installed at the same time, but you want to invoke a specific one, you probably shouldn't be using generic aliases in the first place.

cascadingstyletrees:
> but you want to invoke a specific one, you probably shouldn't be using generic aliases in the first place.

Alphabetic is weird for this context, could instead give the choice to the user instead, maybe using an environment variable like `$AUR_HELPER` (pacui does this), or similar. Not sure on the best practice.

To be fair, this is a very generic alias which people will have to unset very often to use `upgrade` for something else, Is upgrade used in other plugins that would cause a clash? or consider a slight different alias, like `archupgrade`?

ratijas:
> Alphabetic is weird for this context, could instead give the choice to the user, maybe using an environment variable like $AUR_HELPER (pacui does this), or similar. Not sure on the best practice.

What's the pacui you are referring to?  I never heard of it. Anyway, I don't think it's a 'best practice' to over-engineer simple things. If you do multiple AUR helpers, please, configure you own aliases in the way you like them in your own ~/.zsh_aliases or whatever. Personally, the one I use is quite... [distinctive](https://github.com/ratijas/dotfiles/blob/master/tag-zsh/config/zsh/zshrc.d/70.aliases.d/autism.zsh), I would never suggest to include it in oh-my-zsh.

Regarding naming, I have no opinion. These were in oh-my-zsh for as long as I'm using arch (btw), which is more than 4 years. If I were ever using them, and they was suddenly changed, I'd be so mad. That's actually the reason I migrated from oh-my-zsh to copying my favorite plugins into my personal dotfiles repository — to prevent any surprises from upstream.

mcornella:
> Is upgrade used in other plugins that would cause a clash?

I checked, and no. I'd rather keep breaking changes out of this PR though. After this is merged you're free to submit a new one updating every obsolete part of the plugin.

cascadingstyletrees:
> > Is upgrade used in other plugins that would cause a clash?
> 
> I checked, and no. I'd rather keep breaking changes out of this PR though. After this is merged you're free to submit a new one updating every obsolete part of the plugin.

What do you mean here by breaking changes? The current proposed changes in this PR are the ones I think are invasive, and using `upgrade` for all AUR helpers is the dangerous part. They all have different behaviours, and handle dangerous flags differently:

mcornella:
The current `upgrade` function is not documented, but it is definitely easier to understand which AUR helper is used *in the code*. Now, the order of priority is as follows:

1. `trizen`
2. `pacaur`
3. `yaourt`
4. `yay`
5. `pacman`

After this is merged, the order is:

1. `yay`
2. `yaourt`
3. `trizen`
4. `pacaur`
5. `pacman`

I'd keep the code logic to make it more discoverable, as well as the order. In fact, I'd convert the if chain to a for which is easier to read and code:
```zsh
for pkgmgr in trizen pacaur yaourt yay pacman; do
  if (( ${+commands[$pkgmgr]} )); then
    alias upgrade="$pkgmgr -Syu"
    break
  fi
done
unset pkgmgr
```

And document it in the README.

> Just trying to understand, which breaking changes are you referring to?

Removing `yaourt` support and removing useless functions. Also I thought your suggestion was to remove the `upgrade` alias.

ratijas:
> The fact that if you have multiple AUR_helpers, and not know which one is going to be called is bad practice, since it may have been installed as a dependency for another package.

Now, that's an interesting thought. However, I'm not convinced it is relevant. Here's why:

1. Alias `upgrade` is clearly intended to be used as-is, without any additional flags. If you need more detailed upgrade, why would you be using [generic] alias in the first place?
2. If some AUR helper (installed without user knowing about it — WTF on its own) were to include dangerous flags by default, it would get kicked out of AUR in a blink of an eye. That's what happened to pacui, after all.

> Now, the order of priority is as follows:
> After this is merged, the order is:

Do you really think it's a good idea to rely on order of alias shadowing? I could barely call that ordering 'logical', but rather an 'accidental'. Then again, if you are using more than one AUR helper -> don't use aliases.

ratijas:
Or let's just remove `upgrade` aliases altogether, and append "they were considered harmful" to the changelog. I have no preference regarding them whatsoever.

ratijas:
Replaced `pacmanallkeys` with alias. Tweaked commit description.

ratijas:
So, what do we do now?

mcornella:
> Do you really think it's a good idea to rely on order of alias shadowing? I could barely call that ordering 'logical', but rather an 'accidental'.

I'd rather make the order explicit with code like the above, and keep the previous order. If we have to, we can remove the alias on a later PR. Also:

> And document it in the README.

in a separate section.

mcornella:
Ok I documented the `upgrade` alias. One more thing, what's with the whole `sudo abs && sudo aur`? Is that still used?

mcornella:
This was open too long, it was ready so I merged it. Do any of you want to be the maintainer for the archlinux plugin? That'd mean automatic mentions on PRs that change the plugin.

ratijas:
Oh man, forgot about this one. Sorry. Good to see it moving again. Thank you!

> Do any of you want to be the maintainer for the archlinux plugin? That'd mean automatic mentions on PRs that change the plugin.

I wouldn't mind being mentioned at all.

mcornella:
OK done, thanks!

PR #9547, opened 2020-12-27
Allow user to exclude zsh components.

cbartondock:
## Standards checklist:

## Changes:

* Allow users to exclude certain parts of oh-my-zsh that they may not need or that interfere with part of their setup. 
* Document said ability in the readme under Advanced Topics.

## Other comments:
Full quote from readme:

If you want to simply exclude any of the default behaviors, create a zsh list named excludeconfigs of the filenames you wish to exclude in your `.zshrc`. For example, if you wanted to exclude `lib/theme-and-appearance.zsh` and `lib/nvm.zsh` you would add the following to your `.zshrc`:
```
excludeconfigs=(
theme-and-appearance
nvm
)
```


mattmc3:
Isn't this already easily accomplished via `$ZSH_CUSTOM`? You can exclude configs just by having an empty file in `$ZSH_CUSTOM/lib` like so:

```zsh
ZSH_CUSTOM=${ZSH_CUSTOM:-$ZSH/custom}
mkdir -p $ZSH_CUSTOM/lib
touch $ZSH_CUSTOM/lib/theme-and-appearance.zsh
touch $ZSH_CUSTOM/lib/nvm.zsh
```

cbartondock:
Ah that's quite clever; not sure it's obvious from the user perspective but if it's well documented that's fine by me !

mattmc3:
Well documented? No, it's definitely not. You're right that it's totally non-obvious for a regular user, and something you'd never know unless you look in the oh-my-zsh.sh code: https://github.com/ohmyzsh/ohmyzsh/blob/66e0438d4788d0a0cdc107b3975418f1a572e354/oh-my-zsh.sh#L100-L104

It's mentioned in [some places](https://stackoverflow.com/questions/21812864/oh-my-zsh-disable-lib-history-zsh) if you know what to search for, but not very easily discoverable.

Perhaps a PR adding a `lib` directory in [ohmyzsh/custom](https://github.com/ohmyzsh/ohmyzsh/tree/master/custom) with comments similar to "themes" and "plugins" would make this override method way more obvious to a user.

mcornella:
It's also documented [in the wiki](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-internals).

PR #9552, opened 2020-12-28
fix(lib): mark changes as MODIFIED on 'MM' in `git_prompt_status`

seamile:
## Standards checklist:

## Changes:

- modify a status mapping in function `git_prompt_status`.

## Other comments:

The 'MM ' prefix means some file modified after added, so the right status should be `'MODIFIED'`.

Like this:

```shell
$ echo 'foo' >> README.md
$ git add README.md
$ echo 'bar' >> README.md
$ git status --short

MM README.md
```

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -170,7 +170,7 @@ function git_prompt_status() {
     '\?\? '     'UNTRACKED'
     'A  '       'ADDED'
     'M  '       'ADDED'
-    'MM '       'ADDED'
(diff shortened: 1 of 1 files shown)


mcornella:
Should `'M  '` also be MODIFIED?

seamile:
> Should `'M '` also be MODIFIED?

Yes. 
The first 'M' means thad some code has been modified and added.
The second 'M' means that some code has been modified **but not added yet**.

Issue #6954, opened 2018-06-28
Can not complete docker commands after alias with grc


j16180339887:
Can not get docker completion work after alias with [grc](https://github.com/garabik/grc).

```sh
$ alias docker='grc -es --colour=auto /usr/bin/docker'
$ docker <tab> # Not working anymore
```

Any ideas? thanks.

mcornella:
You should specify that you want to use docker completions with the docker alias with `compdef _docker docker`. But you have to also run `setopt completealiases` to disable zsh's behavior of looking at the alias definition to know which completion function to run (in your case, it would try to find completions for the grc command).

Enabling this option will mess up the rest of aliases that don't have an explicit completion defined, so here's a better option: use a function instead of an alias, like so:

```zsh
function docker { grc -es --colour=auto /usr/bin/docker "$@" }
compdef _docker docker
```

PR #9553, opened 2020-12-28
feat(grc): source `grc.zsh` instead of hard-coding its content

hjpotter92:
## Standards checklist:

## Changes:

Instead of maintaining a copy for grc plugin, let grc installation define the aliases

The grc project is still under active development, despite lagging in stable releases to the package repositories. For users's with the devel installation, the plugin doesn't make use of [the newly defined aliases][1].

Since grc installs its alias config to `/etc/`, the plugin should offload to the same.

  [1]: https://github.com/garabik/grc/blob/devel/grc.zsh


hjpotter92:
The only downside to this could be the completion, which is covered in #6954

hjpotter92:
With the successful merge of garabik/grc#172, the above issue would also not be there in the next release of grc.

mcornella:
How often does `grc` release? The `master` branch hasn't seen activity since 2018, as well as the releases tab. I'd love to just do this but I don't know if this would work for all currently used versions.

I also made some comments on your changes, see https://github.com/garabik/grc/commit/98f483068a817afccb29572c25f120ea06b4c6b6#r45477503.

hjpotter92:
@mcornella  The plan is to release the next version soon, just that the maintainer has had a little backlog: garabik/grc#149

however, the `/etc/grc.zsh` is present for past 3 years, so it is safe to assume that the same would be found on all distributions with the installation.

PS: Thank you for the review on my PR over the other project. I just have one question about the point you raised regarding `path` variable: _does it conflict even if it is defined as a `local` value inside the function_? I was originally planning to use `$path` instead of `${commands[$0]}`, but I think it does not get replaced as I had hoped and the stub was just left in place.

mcornella:
> _does it conflict even if it is defined as a `local` value inside the function_?

When the function exits, the global `$path` is still there, but inside the function you'll get weirdness if you call an external command. In that case nothing happens, but it's a bad habit using `$path` in zsh: sooner or later you'll forget about it and there will be problems.

> I was originally planning to use `$path` instead of `${commands[$0]}`

`${commands[$0]}` works well. An alternative is to use `eval` to define the function and make the path static, like I did in the plugin, but your solution is also good.

mcornella:
We should do something with the README, remove the wrapped commands list (since it'll depend on the grc version) and make it explicit that it sources `/etc/grc.zsh`.

Also, I'd substitute the `[ -s ... ]` for
```zsh
if [[ -r /etc/grc.zsh ]]; then
  source /etc/grc.zsh
fi
```
Only source it if the file is readable, and doesn't propagate the return code if the condition fails. I know the docs say otherwise, but this is a better approach. Another option, if you want to keep the oneliner, is to use `||` instead of `&&` and negate the condition:
```zsh
[[ ! -r /etc/grc.zsh ]] || source /etc/grc.zsh
```
But it's harder to understand.

hjpotter92:
Updated both as per suggestion.

Thanks

mcornella:
I corrected the README and added support for Homebrew path (it seems to be `/usr/local/etc/grc.zsh` but feel free to correct me if I'm wrong). If you're ok with the changes I'll merge this.

hjpotter92:
Thanks @mcornella , the changes look ok. :+1:

mcornella:
Alright thanks!

PR #9554, opened 2020-12-28
Add aliases for git diff --exit-code

leighmcculloch:
## Standards checklist:

## Changes:

- Add aliases for git diff --exit-code:
   `gde` = `git diff --exit-code`
   `gdce` = `git diff --cached --exit-code`

## Other comments:

Composing git diff often requires forking continuing or not on whether there are changes, and the exit code is helpful here. An alias that includes this would be helpful when chaining commands.


mcornella:
Can you provide an example? Is this used often enough to require an alias?

leighmcculloch:
@mcornella Thanks for the quick response.

The use case I've most commonly found myself using this is with workflows similar to:
```
go test -v && (gde || gdce || (ga . && gcn! && gp -f))
```

It probably isn't used often enough to establish as an alias here. Even my own use is sparing, it's just an alias I use locally and thought I'd contribute upstream. Happy to close if that's the right call.

kayobruno (approved):

mcornella:
OK, so I'll close this, seeing as this is a niche use case. I have #5721 in mind. What I'd suggest is that you shorten it for yourself, if you always end up using `gde || gdce`, you can use `alias gdnochanges="git diff --exit-code && git diff --cached --exit-code"` or something like that. You can also put it in a one-line alias or function, in your zshrc file.

Thanks!

PR #9557, opened 2020-12-29
feat(plugins): add `gh` plugin for GitHub CLI

jilleJr:
~~Copied from the kubectl plugin but changed to use the `gh` command.~~

**EDIT:** I changed implementation after some insights taken from https://github.com/cli/cli/issues/2741 and https://github.com/cli/cli/issues/2321#issuecomment-733031189.

How it works:

- pipe `gh completion --shell zsh` into `$ZSH/completions/_gh`.
- pipe `gh --version` into `$ZSH_CACHE_DIR/gh_version`.
- Once per 24h, compare `gh --version` with `$ZSH_CACHE_DIR/gh_version`, and if they differ then regenerate the `_gh` and `gh_version` files.

## Standards checklist:

## Changes:

- Added new plugin under `plugins/gh/gh.plugin.zsh`

## Other comments:

~~I only tested it through my zinit setup, and had trouble with getting it to load properly. Initially I copied the code from the kubectl plugin and did the equivalent thing for the gh command, but it did not want to register the completions. Spend an hour on debugging without prevail.~~

~~This method works, though it does not take advantage of any caching.~~

**EDIT:** As mentioned above; with some help from @mislav and @jetersen I changed the implementation to rely on writing to a file in `$fpath` instead. The implementation now also take use of caching, so users may only experience weird behaviour after an update for up to 24h.


jetersen:
You mentioned zinit, how did you try to load the plugin?

I current use, which should actually handle calling compinit.
```zsh
zinit depth"1" wait lucid atload"zicompinit; zicdreplay" blockf for \
  pick"asdf.sh" src"completions/_asdf" @asdf-vm/asdf \
  OMZ::plugins/kubectl/kubectl.plugin.zsh \
  OMZ::plugins/helm/helm.plugin.zsh \
  OMZ::plugins/git/git.plugin.zsh \
```

jilleJr:
@jetersen I loaded it in like so:

```zsh
zinit https://github.com/jilleJr/ohmyzsh/blob/feature/gh-plugin/plugins/gh/gh.plugin.zsh

zinit as'null' \
    atinit'zicompinit; zicdreplay' \
    id-as'zicompinit-zicdreplay' \
    for zdharma/null
```

I omitted all of my other configs from the snippet above ^, but you can see my whole setup here: https://github.com/jilleJr/dotfiles/blob/d777ef48ba44ea9543cd43e472e4c818220c067b/zinit/.zshrc#L71

mcornella:
Hey this looks great! Can you follow the same advice as in https://github.com/ohmyzsh/ohmyzsh/pull/9847#issuecomment-825847100 and put the completion file inside the plugin folder, with a .gitignore? The `/completions` folder will disappear and it isn't gitignore'd. Thanks!

jilleJr:
> Hey this looks great! Can you follow the same advice as in [#9847 (comment)](https://github.com/ohmyzsh/ohmyzsh/pull/9847#issuecomment-825847100) and put the completion file inside the plugin folder, with a .gitignore? The `/completions` folder will disappear and it isn't gitignore'd. Thanks!

@mcornella I hope I understood you correctly here. I basically did a search and replace on `$ZSH/completions` to `$ZSH/plugins/gh`, as @carlosala did in https://github.com/ohmyzsh/ohmyzsh/pull/9847/commits/b470a24d1c7ec8c7d3c77a046a81a67a7eaf5f0d

carlosala:
Hey! @jilleJr 
There's no point in doing this hook each 24h, it's better to just run the function each time ohmyzsh is loaded (each time you open a terminal). Look at #9847

jilleJr:
@carlosala It was a performance thing. I sometimes have a lot of plugins loaded and it could slow down my zsh startup quite a bit, so I hotfixed some plugins to not register new plugins each time and did the same here. But I've since moved away from my toaster and so I could remove the zsh-hook now without waiting another second for a terminal to become interactive.

carlosala:
The `mkdir -p $ZSH/plugins/gh` and the `fpath` adding are not necessary.

jilleJr:
@carlosala They were necessary for my zinit setup. But perhaps I'm using zinit wrong then. Will look into it further

carlosala:
@jilleJr Ok, but I think that ohmyzsh doesn't care about which plugin manager are you using. I think that plugins should be adapted to ohmyzsh, not others.

jilleJr:
@carlosala valid point. I'll change it

carlosala:
Hey!
You need to do the same that in this PR #9863, I discovered this bug in the deno plugin this morning.
Thanks!

carlosala:
@mcornella I think it's ready to merge!

PR #9561, opened 2020-12-30
feat(lerna): adds a lerna plugin w/ aliases

pgollucci:
Aliases come from the aws/aws-cdk project but are generic


TheTechRobo:
Why was it closed?

pgollucci:
Gave up hope on attention.

PR #9569, opened 2021-01-02
fix(archlinux): update URL and key server in `pacmanallkeys`

cascadingstyletrees:
Error and investigation:

```
❯ pacmanallkeys                                                                                                                                 ─╯
[sudo] password for lagbrie: 
==> ERROR: No targets specified
❯ which pacmanallkeys                                                                                                                           ─╯
pacmanallkeys () {
	curl -s https://www.archlinux.org/people/{developers,trustedusers}/ | awk -F\" '(/pgp.mit.edu/) { sub(/.*search=0x/,""); print $1}' | xargs sudo pacman-key --recv-keys
}
❯ curl -s https://www.archlinux.org/people/{developers,trustedusers}/ | awk -F\" '(/pgp.mit.edu/) { sub(/.*search=0x/,""); print $1}'
❯ curl -svo /dev/null https://www.archlinux.org/people/\{developers,trustedusers\}/ |& grep -i "< HTTP"
< HTTP/2 301 
< HTTP/2 301 
❯ curl -Lsvo /dev/null https://www.archlinux.org/people/\{developers,trustedusers\}/ |& grep -i "< HTTP\|location:"
< HTTP/2 301 
< location: https://archlinux.org/people/developers/
< HTTP/2 200 
< HTTP/2 301 
< location: https://archlinux.org/people/trustedusers/
< HTTP/2 404 

```

After the above investigation, I've updated the command to function as originally intended.

## Standards checklist:

## Changes:

- Added follow redirects curl flag to fix the 301 response.
- Updated trusted users URL format.
- Changed from `pgp.mit.edu` keyserver, since the archlinux page has moved to `keyserver.ubuntu.com` keyservers.

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -188,8 +188,8 @@ function pacdisowned() {
 }
 
 function pacmanallkeys() {
-  curl -s https://www.archlinux.org/people/{developers,trustedusers}/ | \
-    awk -F\" '(/pgp.mit.edu/) { sub(/.*search=0x/,""); print $1}' | \
+  curl -sL https://www.archlinux.org/people/{developers,trusted-users}/ | \
+    awk -F\" '(/keyserver.ubuntu.com/) { sub(/.*search=0x/,""); print $1}' | \
     xargs sudo pacman-key --recv-keys
 }
 


mcornella:
Thanks!

ratijas:
By the way, why this script is needed in the first place?

mcornella:
Related: https://github.com/ohmyzsh/ohmyzsh/pull/1456.

cascadingstyletrees:
aww I should have added myself as a contributor haha.


> By the way, why this script is needed in the first place? Can we have a comment or a `--help` text?

Honestly, I was just fixing because it wasn't working, but I'm pretty sure `sudo pacman-key --refresh-keys` accomplishes the same thing, and would work for arch derivatives like manjaro.

mcornella:
> Honestly, I was just fixing because it wasn't working, but I'm pretty sure `sudo pacman-key --refresh-keys` accomplishes the same thing, and would work for arch derivatives like manjaro.

@ratijas `pacman-key --refresh-keys` looks and works better, can you substitute the function for an alias to that in #9546?

ratijas:
To the best of my understanding, is could only be useful to the Trusted Users and Developers of Arch Linux. Which are like... about 60 of them total? Is oh-my-zsh really an appropriate place to store tooling and scripts for such specific group of people?

cascadingstyletrees:
> To the best of my understanding, is could only be useful to the Trusted Users and Developers of Arch Linux. Which are like... about 60 of them total? Is oh-my-zsh really an appropriate place to store tooling and scripts for such specific group of people?

That's what @mcornella is saying, instead of using the function, to replace it with `pacman-key --refresh-keys`.

cascadingstyletrees:
> To the best of my understanding, is could only be useful to the Trusted Users and Developers of Arch Linux. Which are like... about 60 of them total? Is oh-my-zsh really an appropriate place to store tooling and scripts for such specific group of people?

Maybe I misunderstood your question. Importance, since this script does similar to `pacman-key --refresh-keys`, it's useful because very important to have your key database up to date, or else pacman won't trust the packages built by the maintainer (TUs and Develoeprs), and it will accuse invalid gpg signature.

ratijas:
Isn't that is what upgrades to pacman-keyring are doing? I never refreshed keys manually for the 4+ years. Should I be worried? :eyes: 

> instead of using the function, to replace it with pacman-key --refresh-keys.

You mean, the whole 3 lines of curl | awk | xargs, right?

cascadingstyletrees:
Okay so yes, that's what it's doing, and yes everything in that alias to pacman-key --refresh-keys.
Well you should consider yourself lucky if you haven't had a pacman database issue!

But really your questions are answered here:

https://wiki.archlinux.org/index.php/Pacman/Package_signing#Managing_the_keyring

ratijas:
> Okay so yes, that's what it's doing, and yes everything in that alias to pacman-key --refresh-keys.
Well you should consider yourself lucky if you haven't had a pacman database issue! Are you using Arch or another Arch derivative?

Well, guess I'm really lucky then.

I'm using vanilla Arch Linux with vanilla kernel, regularly upgrading, not every day though. Back in the days, I installed it with a little help from [aui](https://github.com/helmuthdu/aui), just because of how n00b I was at that point. I have gotten better since, but never reinstalled the system completely. Rather, I'm talking evolutionary approach, fixing and polishing system parts one by one as soon as I get into them. It is truly a miracle than I managed to survive this far, and even moved root file-system to btrfs with snapper backups. ;-)

My personal record was updating Arch on my old unused laptop which was busy collecting dust for a half-year or so. And even that went quite smoothly (except for KDE/Plasma crashing before reboot as f*cking usual).

Thanks for the link. I'll amend #9546 now.

ratijas:
Btw, Just upgraded my Arch. Watched pacman keyring update itself like this:

```
(118/435) upgrading archlinux-keyring                                          [---------------------------------------------] 100%
==> Appending keys from archlinux.gpg...
gpg: public key CF66D153D884358F is 16 seconds newer than the signature
==> Locally signing trusted keys in keyring...
  -> Locally signing key AB19265E5D7D20687D303246BA1DFB64FFF979E7...
  -> Locally signing key DDB867B92AA789C165EEFA799B729B06A680C281...
[...]
==> Importing owner trust values...
==> Disabling revoked keys in keyring...
  -> Disabling key 4A8B17E20B88ACA61860009B5CED81B7C2E5C0D2...
[...]
==> Updating trust database...
gpg: next trustdb check due at 2021-03-06
(119/435) upgrading kpty                                                       [---------------------------------------------] 100%
```

But I assume it is absolutely possible to screw up you keyring like nothing. Good thing it's working for me for years.

PR #9575, opened 2021-01-05
Tilde key switch plugin for Macbooks: tilde-switch plugin

mcanvar:
## Standards checklist:

## Changes:

- When enabling this plugin it switches back the `±` key to `~` without needing any third-party tool or service on Macbooks which has non-US native keyboards like Swedish or Turkish etc.

## Other comments:
I am using it in my local oh-my-zsh and it saves my time a lot. As you can see in this question on [stackexchange.com](https://apple.stackexchange.com/questions/329085/tilde-and-plus-minus-%c2%b1-in-wrong-place-on-keyboard) lots of people suffer. I would appreciate if my PR accepted. I think this is a serious need for people to have a non-US keyboard on their Macbooks.

Thank you for your amazing work!

### DIFF
--- a/plugins/tilde-switch/README.md
+++ b/plugins/tilde-switch/README.md
@@ -1,0 +1,9 @@
+# tilde-switch plugin
+
+By default, on some non-US keyboards of Macbooks like Swedish and Turkish, tilde key switched with a plus-minus key. So this plugin switches back the `±` key to `~` without needing any third-party tool or service.
+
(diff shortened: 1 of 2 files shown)


robbyrussell:

I'm closing it because the change isn't really a shell one: `hidutil` remaps the key for all of macOS until reboot, so starting a terminal would change the keyboard in every other app too. Karabiner or a login item is a better fit for that, and it also works as a custom plugin in `$ZSH_CUSTOM` if you'd rather keep it in zsh.

PR #9577, opened 2021-01-06
fix(lib): update Emacs terminal detection in `title` function

### LINKED ISSUE #6411: Emacs doesn't play nicely with oh-my-zsh anymore (termsupport.zsh)
I recently noticed my **zsh terminals in emacs** started printing junk/strange characters whenever I issue a command. See the screenshot below:

screen shot 2017-11-10 at 2 29 15 pm

After careful investigation, I noticed the problem comes from sourcing `ZSH/oh-my-zsh.sh`.

If I don't load `oh-my-zsh` everything works like a charm, but as soon as I source it (even without any plugins, options theme selection, etc), I get the problem again.

Moreover, I localized the problem to the loading of [`termsupport.zsh`](https://github.com/robbyrussell/oh-my-zsh/blob/master/lib/termsupport.zsh). If I force `oh-my-zsh `to *not* load this file (e.g. if I remove it), everything seems to works fine.

In case it helps I am using:
* `zsh 5.4.2 (x86_64-apple-darwin16.7.0)`
* The latest oh-my-zsh (I just run upgrade_oh_my_zsh)

What could be going on? If it helps I believe I started noticing the problem a few days/weeks ago.

felker:
Emacs replaced the environment variable that it sets in interactive subshells from `EMACS`  to `INSIDE_EMACS`. See https://www.gnu.org/software/emacs/manual/html_node/emacs/Interactive-Shell.html for details.

I couldn't find exactly when Emacs made this change, but [this post from 2015](https://lists.gnu.org/archive/html/bug-gnu-emacs/2015-03/msg01008.html) shows that `EMACS` was deprecated for 8 years at that point.

Closes #6411. 

## Standards checklist:

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -10,7 +10,7 @@ function title {
   emulate -L zsh
   setopt prompt_subst
 
-  [[ "$EMACS" == *term* ]] && return
+  [[ "$INSIDE_EMACS" == *term* ]] && return
 
   # if $2 is unset use $1 as default
   # if it is set and empty, leave it as is


mcornella:
Thanks!

It didn't cause any problems because `$TERM` is `eterm-color`, which ends up in the `else` branch of the function, and given that `tsl` and `fsl` aren't available for `eterm-color`, the function ends up doing nothing.

PR #9578, opened 2021-01-06
Add alias for kubectls dry-run flag

christophebeling:
## Changes:

I added two more aliases in order to support kubectls client and server-side dry run. 

## Other comments:
The try run flag is extremely useful to quickly generate template / boilerplate code for new Kubernetes resources.  https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands


robbyrussell:

Two things have changed since 2021: `kdrs` is now taken by `kubectl describe replicaset`, and the kubectl plugin has grown to well over a hundred aliases, so we've gotten stricter about adding more. `kubectl --dry-run=client` is also easy enough to type in full.

Closing for that reason. Thanks for suggesting it.

Issue #9165, opened 2020-08-09
add deno plugins


kalwabed:
i have been using <a href='https://deno.land'> Deno</a> for a long time, to make it easier for users like me in the present or future, it is better to bring support to Deno.

kapilpau:
Created PR for plugin: https://github.com/ohmyzsh/ohmyzsh/pull/9363

PR #9591, opened 2021-01-12
Added Deno plugin

carlosala:
## Standards checklist:

## Changes:

- It adds aliases for deno commands (issue #9165).
- It adds completion for deno binary.

## Other comments:

- It improves the PR #9363, adding more useful aliases and completion.


mcornella:
The completions were generated using `deno completions zsh`, can you call that instead and cache its output like the `npm` plugin does? Then it won't be necessary to update the script and it'll match what deno version the user has.

carlosala:
I'm trying to create a completion function similar to the npm plugin is, but I'm so newbie in shell scripting and I'm struggling a little bit with it. The output of `npm completion` and `deno completions zsh` are a bit different, and I don't know how to create the `_deno_completion()` function. Can someone help me?
Thanks, and sorry for my noob abilities :)

carlosala:
Improved in #9847

PR #9593, opened 2021-01-13
fix(colorize): bypass less aliases in `colorize_less`

KapJI:
## Standards checklist:

## Changes:

I want to set `alias less="colorize_less"` to have fancy `less`.
But in current state it will expand alias recursively basically in infinite loop causing the whole terminal to crash. Prevent this by prepending it with `command` which will use default `less` instead.


mcornella:
I think this change is valid, and I'll merge this, but you might consider instead using the `cless` alias. It's just one more character, and using `colorize_less` for everything might have unintended consequences.

KapJI:
Yeah, I understand this. But for now I'd like to try just `less` to make sure I won't forget to use `cless`.

PR #9600, opened 2021-01-15
Add foreman start shortcut to autoamte process

romanchsharma:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/rails/rails.plugin.zsh
+++ b/plugins/rails/rails.plugin.zsh
@@ -50,6 +50,7 @@ alias ru='rails runner'
 alias rs='rails server'
 alias rsd='rails server --debugger'
 alias rsp='rails server --port'
+alias rfs='foreman start'
 
 # Rake aliases
 alias rdm='rake db:migrate'


mcornella:
Please explain what this does and add it to the README.

romanchsharma:
Foreman is a system management tool, in ruby on rails, it helps to run automate the process through proc file. It's not compulsory to include it inside Gemfile, but it should be present in the ruby bundle, Then we can create proc file to just add multiple commands like:-
`web: bundle exec rails s`
`webpack: bin/webpack-dev-server`
`worker: bundle exec sidekiq`
then add a reference to .foreman file and done for one command, foreman start
this will auto compile Webpack, and can run many other commands with just `foreman start` and now `rfs`.
We can also use it with already existing commands `RED` or `REP` to change the environment. 
[reade more](http://ddollar.github.io/foreman/)

romanchsharma:
Sorry for language issues, I have a specific learning disability, I hope I write it correctly.

Issue #9603, opened 2021-01-16
Battery plugin shows incorrect values when you have two batteries in Linux


simcop2387:
**Describe the bug**
The battery plugin shows a value like
```
ryan@magellan [07:47:34] [~]  10098%
-> %
```

**To Reproduce**
Have a laptop with two or more batteries running linux, such as my ThinkPad T580.  Or emulate it with the following output from the acpi program.

```
-> % acpi
Battery 0: Full, 100%
Battery 1: Discharging, 97%, 03:28:57 remaining
ryan@magellan [07:49:34] [~]  10097%
```

**Expected behavior**
Something sane, but not reporting 10,000% battery level :)

**Desktop (please complete the following information):**

 - Debian Bullseye/Testing
 - ZSH Version: 5.8
 - Terminal emulator: Konsole (KDE), but it's obviously present in every other console including xterm

simcop2387:
I think the naive thing to do would be to just average the two values, even though it wouldn't give a true capacity since the batteries are of different capacities.  But that would then have it continually drop in % through the whole process since the laptop will discharge battery 1 until it's at 5% and then start discharging battery 0.

simcop2387:
From the looks of things, it might be better to use ```acpitool``` over ```acpi```, it's got a better output section for me at least

```
ryan@magellan [08:18:27] [~]  49%
-> % acpitool
  Battery #1     : Unknown, 99.55%
  Battery #2     : Discharging, 49.58%, 01:12:05
  All batteries  : 62.60%, 02:03:03
  AC adapter     : off-line 
  Thermal info   : <not available>
ryan@magellan [08:18:29] [~]  49%
-> % acpi 
Battery 0: Unknown, 99%
Battery 1: Discharging, 49%, 01:12:07 remaining
ryan@magellan [08:18:31] [~]  49%
-> % 
```

the All batteries: section gives a total capacity percentage (along with a combined expected time) so that no fancy math has to be done inside of zsh.

mcornella:
Fixed in https://github.com/ohmyzsh/ohmyzsh/pull/9609.

PR #9609, opened 2021-01-18
Fix issue #9603 by also being able to use apcitool which gives more d…

simcop2387:
…etails and complete information when multiple batteries are present

## Standards checklist:

## Changes:

- This changes to use the ```acpitool``` command when available on Linux over the ```acpi``` command as it provides more correct information when in a laptop with multiple batteries.

## Other comments:

I coded it to use either command so that it shouldn't remove any functionality from any existing users but will instead use acpitool to get more accurate information when it is available.  I haven't been able to test it on a laptop that has a single battery so I can't confirm that it works properly there yet but I believe it should based on my reading of the documentation.

This could be done by using the ```acpi``` command also, but you'd have to capture the total capacity and percentage of each battery, calculate the capacity available in watt hours, and then divide that by the total capacity of all batteries of the laptop.


simcop2387:
Had to do an amend real quick as i left out a ! on accident

plugins/battery/battery.plugin.zsh
```diff
@@ -110,13 +110,19 @@ elif [[ "$OSTYPE" = freebsd* ]]; then
 elif [[ "$OSTYPE" = linux*  ]]; then
 
   function battery_is_charging() {
-    ! acpi 2>/dev/null | command grep -v "rate information unavailable" | command grep -q '^Battery.*Discharging'
+    if (( $+commands[acpitool] )); then
+      ! apcitool 2>/dev/null | command grep -q '^\s+Battery.*Discharging'
+    else if (( $+commands[acpi] )); then
```
mcornella, line 115: The proper keyword is `elif`. This opens a new `if` and that's why you need the awkward `fi; fi` line below.

plugins/battery/battery.plugin.zsh
```diff
@@ -110,13 +110,19 @@ elif [[ "$OSTYPE" = freebsd* ]]; then
 elif [[ "$OSTYPE" = linux*  ]]; then
 
   function battery_is_charging() {
-    ! acpi 2>/dev/null | command grep -v "rate information unavailable" | command grep -q '^Battery.*Discharging'
+    if (( $+commands[acpitool] )); then
+      ! apcitool 2>/dev/null | command grep -q '^\s+Battery.*Discharging'
```
mcornella, line 114: typo: `apcitool`

plugins/battery/battery.plugin.zsh
```diff
@@ -110,13 +110,19 @@ elif [[ "$OSTYPE" = freebsd* ]]; then
...
+    else if (( $+commands[acpi] )); then
+      ! acpi 2>/dev/null | command grep -v "rate information unavailable" | command grep -q '^Battery.*Discharging'
+    fi; fi
   }
 
   function battery_pct() {
-    if (( $+commands[acpi] )); then
+    if (( $+commands[acpitool] )); then
+      acpitool 2>/dev/null | command grep -E '^\s+All batteries\s*:' | cut -f1 -d',' | tr -cd '[[:digit:].]'
```
mcornella, line 122: This doesn't work if there's only one battery.

plugins/battery/battery.plugin.zsh
mcornella: Also you need `grep -E` to use `\s+`, I believe. It's not working for me otherwise.

mcornella (changes requested):

plugins/battery/battery.plugin.zsh
simcop2387: Do you have an example of that output so I can have a bit to test with? I don't have easy access to one with only one battery right now.

plugins/battery/battery.plugin.zsh
simcop2387: Yep fixed

plugins/battery/battery.plugin.zsh
simcop2387: Thanks, i figured there had to be one but i couldn't find it on a quick google.

plugins/battery/battery.plugin.zsh
simcop2387: I've updated it to try to better handle a single battery situation, let me know how it works

PR #9620, opened 2021-01-25
Update VI_KEYMAP on zle-line-finish.  Fixes next prompt draw after running command from 'normal' mode.

CrispyConductor:
In some cases, the prompt was not always being redrawn properly after running a command from vi-mode.  If a command is executed with zle in 'normal' mode, after command completion, the next prompt drawn still indicates 'normal' mode, even though it is a new zle line and is back in 'insert' mode.  This was because the `VI_KEYMAP` var (on which the prompt selection is based) was being updated on zle init, which isn't called until after the prompt draw.  This PR fixes this by setting `VI_KEYMAP` back to 'main' on `zle-line-finish` to mirror what is already being done on `zle-line-init`.

## Standards checklist:

## Changes:

- In `vi-mode` plugin, reset `VI_KEYMAP` to 'main' after zle finishes editing line.

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -50,9 +50,10 @@ zle -N zle-keymap-select
 # These "echoti" statements were originally set in lib/key-bindings.zsh
 # Not sure the best way to extend without overriding.
 function zle-line-init() {
-  LAST_VI_KEYMAP=$VI_KEYMAP
+  local prev_vi_keymap
+  prev_vi_keymap="${VI_KEYMAP:-}"
   typeset -g VI_KEYMAP=main
-  [[ $LAST_VI_KEYMAP != 'main' ]] && [[ "${VI_MODE_RESET_PROMPT_ON_MODE_CHANGE:-}" = true ]] && zle reset-prompt
+  [[ "$prev_vi_keymap" != 'main' ]] && [[ "${VI_MODE_RESET_PROMPT_ON_MODE_CHANGE:-}" = true ]] && zle reset-prompt
   (( ! ${+terminfo[smkx]} )) || echoti smkx
   _vi-mode-set-cursor-shape-for-keymap "${VI_KEYMAP}"
 }
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -52,7 +52,7 @@ zle -N zle-keymap-select
 function zle-line-init() {
   LAST_VI_KEYMAP=$VI_KEYMAP
   typeset -g VI_KEYMAP=main
-  [[ $LAST_VI_KEYMAP != 'main' ]] && zle reset-prompt
+  [[ $LAST_VI_KEYMAP != 'main' ]] && [[ "${VI_MODE_RESET_PROMPT_ON_MODE_CHANGE:-}" = true ]] && zle reset-prompt
   (( ! ${+terminfo[smkx]} )) || echoti smkx
   _vi-mode-set-cursor-shape-for-keymap "${VI_KEYMAP}"
 }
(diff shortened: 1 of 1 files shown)


mcornella:
Have you tried `VI_MODE_RESET_PROMPT_ON_MODE_CHANGE=true`?

cb185222:
> Have you tried `VI_MODE_RESET_PROMPT_ON_MODE_CHANGE=true`?

Yes; that option must be set to redraw the prompt on mode change at all.  With that option set, redrawing the prompt on mode change does work.  The issue is that, even with the option set, it only redraws on mode change while editing a command.  If a command is executed or canceled (Ctrl-c) while in 'normal'/'main' mode, the prompt is not updated the next time it is drawn.

To reproduce:

- Configure the prompt to depend on the `VI_MODE` var and enable `VI_MODE_RESET_PROMPT_ON_MODE_CHANGE`.
- Start typing a command.
- After typing the command, enter vi-mode with Esc (the prompt should change to the 'normal mode' prompt).
- Hit enter (to run) or Ctrl-c (to cancel) the command.
- The next time the prompt is drawn (for the next command line), it is still drawn as if in 'normal mode', despite operating in insert mode.
- Hitting enter again (to draw another prompt) fixes it.

This PR fixes the issue in both the case of running the command from 'normal node' and canceling the command from 'normal mode'.

I did add an additional commit that disables this fix (which adds another prompt redraw in this edge case) if that option is not set for those that intentionally never want the vi-mode plugin to redraw the prompt.

mcornella:
@rwe can you review this?

plugins/vi-mode/vi-mode.plugin.zsh
```diff
@@ -50,13 +50,16 @@ zle -N zle-keymap-select
 # These "echoti" statements were originally set in lib/key-bindings.zsh
 # Not sure the best way to extend without overriding.
 function zle-line-init() {
+  LAST_VI_KEYMAP=$VI_KEYMAP
```
rwe, line 53:
```suggestion
  typeset -g LAST_VI_KEYMAP="${VI_KEYMAP}"
```
The `typeset` avoids warnings when running under `setopt warn_create_global`.

rwe, line 53:
```suggestion
  local prev_vi_keymap
  prev_vi_keymap="${VI_KEYMAP:-}"
```
This avoids warnings on `setopt warn_create_global` or failure on `setopt no_unset`.

crispy1989: Good to know, thanks!  (made a separate commit; one additional line had to be changed)

plugins/vi-mode/vi-mode.plugin.zsh
```diff
@@ -50,13 +50,16 @@ zle -N zle-keymap-select
...
   typeset -g VI_KEYMAP=main
+  [[ $LAST_VI_KEYMAP != 'main' ]] && [[ "${VI_MODE_RESET_PROMPT_ON_MODE_CHANGE:-}" = true ]] && zle reset-prompt
```
rwe:
```suggestion
  [[ "${prev_vi_keymap}" != 'main' ]] && [[ "${VI_MODE_RESET_PROMPT_ON_MODE_CHANGE:-}" = true ]] && zle reset-prompt
```
Just a quoting suggestion and using the local-var name from suggestion above

plugins/vi-mode/vi-mode.plugin.zsh
rwe: Eh, guess the `:-`  in my suggestion is unnecessary since the same var is used without that below.

plugins/vi-mode/vi-mode.plugin.zsh
crispy1989: Yep, included these in the commit that addresses the above.

rwe (approved):
This looks good to me.

mcornella:
Thanks!

PR #9623, opened 2021-01-30
Brew greedy upgrade aliases

xavicampa:
## Standards checklist:

## Changes:

- Added two extra aliases to support 'greedy' upgrades
- Added the descriptions of the two new aliases to README.md

## Other comments:

I had to expand the width of the columns in the markdown table in README.md to accommodate for the new aliases. Unsure if the max. of 80 characters per line applies here, since I believe the table was over 80 characters before my edit. They are at 130 characters width now.


gcamaraomarques:
👍 
@mcornella

Tozapid:
Good day! Can you also add --greedy to cask upgrades?

```
alias bcubog='brew update && brew outdated --cask --greedy'
alias bcugbc='brew upgrade --cask --greedy && brew cleanup'
```

PR #9637, opened 2021-02-04
feat(autojump): add new Homebrew default path on M1 Macs

jumaho:
On M1-Macs homebrew uses /opt/homebrew as default location for ARM
packages. This results in the autojump plugin not being able to find
autojump after a clean default installation.
This commit adds the new default location to the autojump plugin.

## Standards checklist:

## Changes:

- Add autojump path for new homebrew default location

### DIFF
--- a/plugins/autojump/autojump.plugin.zsh
+++ b/plugins/autojump/autojump.plugin.zsh
@@ -10,6 +10,7 @@ autojump_paths=(
   /usr/local/share/autojump/autojump.zsh             # FreeBSD installation
   /opt/local/etc/profile.d/autojump.sh               # macOS with MacPorts
   /usr/local/etc/profile.d/autojump.sh               # macOS with Homebrew (default)
+  /opt/homebrew/etc/profile.d/autojump.sh            # macOS with Homebrew (default on M1 macs)
 )
 
 for file in $autojump_paths; do


KevinBongart:
@jumaho Hey there, I found your PR while investigating why the auto jump plugin doesn't work on my M1 Mac. I believe your commit should fix it, however I don't understand why the Homebrew block below the list of hardcoded paths doesn't already work: https://github.com/ohmyzsh/ohmyzsh/pull/9637/files#diff-d0ad0eb3f09559d60f605005a2ccecdb7b28e8d13040d43fcfe8bd24d57bad9fR24-R31

On my M1 Mac, if I run this command, I get the correct path to the autojump alias:

```sh
ls -la $(brew --prefix)/etc/profile.d/autojump.sh
/opt/homebrew/etc/profile.d/autojump.sh -> ../../Cellar/autojump/22.5.3_2/etc/profile.d/autojump.sh
```

Yet, even with the code above which should find the correct file, the plugin doesn't find autojump:

```sh
[oh-my-zsh] autojump not found. Please install it first.
```

Ideally, the fix for the issue should focus on making the Homebrew-specific code work automatically rather than adding the hard-coded path to the list of potentials paths, though I'll take hardcoded working code over clever non-working code 😄

jumaho:
Hi @KevinBongart, thanks for your comment. 

To be honest, when creating the pull request, I didn't realize that there is a homebrew-specific block below the hardcoded paths. I did some testing and could reproduce the problem you had with `$+commands[brew]` returning `false` which makes the block useless. In contrast to your experience, adding Homebrew to my PATH before loading oh-my-zsh did fix this issue for me, so when using this order, the block now works for me, and the hardcoded paths are not necessary anymore for Homebrew installations. To summarize, the relevant parts of my zshrc looked like this for the successful test:
```
export PATH=/opt/homebrew/bin:$PATH
plugins=(autojump)
source $ZSH/oh-my-zsh.sh
[...]
```
One way to go forward would obviously be to remove the hardcoded lines for Homebrew. Since the dynamic detection below completely hinges on the order of zshrc and we both ran into the issue of having the Homebrew PATH setup "too late", I guess others would run into this as well, making this not a very good solution IMO. In some cases, with non-standard Homebrew prefixes, the block might also still be useful. Therefore, I would probably stick to having the standard locations hard-coded and also keeping the block, to make sure to have tried "everything".

KevinBongart:
Thanks for the detailed answer and reproducing my (our) issue with the PATH :) Your suggestion makes sense to me 👍

KevinBongart (approved):
Agree with @jumaho, adding the new Homebrew path to the list _and_ leaving the Homebrew-specific block below ensures this plugin would work for most people without having to worry about where their PATH is defined, and with no performance impact.

LinPoLen:
also add `/opt/homebrew/share/zsh/site-functions` ?

DIYgod:
Having the same problem, and I would like to know when this can be merged

PR #9642, opened 2021-02-07
kubectl - Add job mgmt, top and namespace creation

gilad-cohen:
## Changes:

**'kubectl' plugin additions:**
- k8s namespace creation
- k8s job management
- k8s top of node and pod, with sorting options
- k8s scale with specific number of replicas 0 to 10 for both deployment and statefulset
- k8s pod force deletion

## Other comments:

Corrected indentation in the 'kubectl' plugin readme, so view the README diff without whitespace changes, otherwise the entire file would look to you like it has been replaced.


ori-cohen:
@robbyrussell @mcornella @larson-carter Please help promote this pull-request

larson-carter:
@ori-cohen I just did a quick look over your PR. 

I do like how you setup the correct indention of the readme. 

Now as for the aliases and such, I have never actually used kubectl personally. So therefore I don't have much input on the usefulness of the aliases and such. 

What I would suggest is to join our Official Discord server https://discord.gg/ohmyzsh and then ask people to vote on this PR. That would give us the best judgement.

robbyrussell:

The job aliases you added here shipped separately in #9992 back in 2022, so that part is already done. For the rest, the `ksd0` through `ksd10` and `ksss0` through `ksss10` sets are the sticking point: that's 22 aliases for something `ksd 3 deploy/foo` already handles, and our guide asks us to prefer one generic alias over many specific ones.

PR #9680, opened 2021-02-22
[lib/git.zsh] Prefer `branch --show-current` for `git_current_branch`

EvyBongers:
## Standards checklist:

## Changes:

- Try to get current branch using `git branch --show-current`

## Other comments:

Git release 2.22.0[1] introduced `git branch --show-current` to return the current branch.[2] This change implements the current primary approach for determining the branch name natively.

[1] https://github.com/git/git/blob/master/Documentation/RelNotes/2.22.0.txt#L32
[2] https://github.com/git/git/commit/0ecb1fc7269e15be890870937b8b639c987abd08

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -94,8 +94,9 @@ function git_remote_status() {
 # Using '--quiet' with 'symbolic-ref' will not cause a fatal error (128) if
 # it's not a symbolic ref, but in a Git repo.
 function git_current_branch() {
-  local ref=$(__git_prompt_git branch --show-current 2> /dev/null)
-  local ret=$?
+  local ref ret
+  ref=$(__git_prompt_git branch --show-current 2> /dev/null)
+  ret=$?
   [[ $ret == 128 ]] && return  # no git repo.
   if [[ $ret != 0 ]]; then  # Git versions before 2.22.0
     ref=$(__git_prompt_git symbolic-ref --quiet HEAD 2> /dev/null)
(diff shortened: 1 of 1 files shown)


TheTechRobo (approved):
A quick look-over seems ok to me.

EvyBongers:
Any update on this?

lib/git.zsh
```diff
@@ -94,11 +94,13 @@ function git_remote_status() {
 # Using '--quiet' with 'symbolic-ref' will not cause a fatal error (128) if
 # it's not a symbolic ref, but in a Git repo.
 function git_current_branch() {
-  local ref
-  ref=$(__git_prompt_git symbolic-ref --quiet HEAD 2> /dev/null)
+  local ref=$(__git_prompt_git branch --show-current 2> /dev/null)
   local ret=$?
```
mcornella, line 98: This $? won't work. It'll get the return code of the `local` command instead of the git one. That's why it's a two-line statement.

mcornella (commented):
I don't know that this is worth the extra call on older versions. The end result is the same, the only difference is that it's more readable.

lib/git.zsh
EvyBongers: I've updated the branch accordingly.

mcornella: That's fine, but my point about it being counterproductive/unnecessary still stands.

EvyBongers: If you (and the other maintainers) decide that it's too much overhead, it's your prerogative to decide not to use this PR. I'll accept your decision either way.

PR #9681, opened 2021-02-23
Add Deno completions

### LINKED ISSUE #9165: add deno plugins

CallMeLaNN:
## Standards checklist:

## Changes:

- Add completions from `deno completions zsh `

## Other comments:

Elegant way of adding completions inspired by helm plugin, it will respect user Deno version.

There is an attempt to add this plugin before #9591 and #9363 but mostly for alias and I fix the script. I can combine those if there is a need for alias since using `zsh-autosuggestions` is good enough.

This should close #9165 too.


carlosala:
Would be interesting to add some aliases. Could you copy my aliases in #9591 and add to the deno.plugin.zsh? I revised that they don't create conflicts!

CallMeLaNN:
@carlosala I just check the `di` alias conflict with debian plugin, `dr` and `dt` conflict with dotnet and drush plugins.

I'm not sure how this alias being unique here, either this is ok or require conflict free with other plugins?
Probably maintainers can suggest and get this PR merged.

carlosala:
True, I just revised the compatibility with apps, nit with other plugins. I tried your completion funcion and it doesn't work... The command for Deno is `deno completionS zsh`, with the s. But it neither works.

CallMeLaNN:
My bad, just fix it.

I think Deno doesn't give much attention or maybe we need to wait for sometime.

carlosala:
This completion function doesn't work for me. It throws the error:
```_arguments:comparguments:325: can only be called from completion function```

mcornella:
Superseeded by #9847.

PR #9692, opened 2021-02-27
fix(cargo): complete flags for default cargo aliases

zethra:
By default, cargo provides alias for some commonly used functions. This patch
makes it that those alias will complete the same as their full commands.

Default alias:
b -> build
c -> check
t -> test
r -> run

Since cargo allows users to define custom alias it would be nice if
oh-my-zsh automatically detected those and preformed completions as such
but that would be a much more complex patch.

Reference documentation: https://doc.rust-lang.org/cargo/reference/config.html#alias

## Standards checklist:

## Changes:

- The case statement matching the cargo sub-command now also the commands default aliases as described above.

### DIFF
--- a/plugins/cargo/_cargo
+++ b/plugins/cargo/_cargo
@@ -77,7 +77,7 @@ _cargo() {
                         '*:args:_default'
                         ;;
 
-                build)
+                build|b)
                     _arguments -s -S $common $parallel $features $msgfmt $triple $target $manifest \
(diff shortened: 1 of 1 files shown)


TheTechRobo (approved):
Quick once-over looks good

zethra:
Any idea if/when this can be merged?

mcornella:
This will be overwritten when updating the completion version so we should remember the fix. Also I don't know if maybe they could be automatically generated on first use, like deno and gh plugins do.

PR #9697, opened 2021-03-01
feat: add alias for building before starting as daemon

alias-rahil:
Building and then starting containers as daemon comes very handy in re-deployments.

Alias: dcupbd
Command: docker-compose up --build -d


TheTechRobo:
Why did you close it? Inactivity doesn't mean rejected.

PR #9703, opened 2021-03-02
Allow waiting dots customization

mortezadadgar:
Signed-off-by: mortezadadgar <mortezadadgar97@gmail.com>

## Standards checklist:

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -60,7 +60,7 @@ zstyle '*' single-ignored show
 
 if [[ $COMPLETION_WAITING_DOTS = true ]]; then
   expand-or-complete-with-dots() {
-    print -Pn "%F{red}…%f"
+    print -Pn "${COMPLETION_WAITING_DOTS_FORMAT:-"%F{red}…%f"}"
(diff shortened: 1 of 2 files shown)


TheTechRobo (commented):
Does it default to the normal mode? This looks like if you don't uncomment the lines there will be no dots

mortezadadgar:
> Does it default to the normal mode? This looks like if you don't uncomment the lines there will be no dots

yes you're right, rebased

mortezadadgar:
@mcornella Is there any reason you don't merge this pull

mcornella:
This can be simplified to a single setting that includes the whole color + symbol sequence which also makes it more customizable.

TheTechRobo:
@mortezadadgar They get a lot of PRs, be patient. :)

mortezadadgar:
CC @mcornella

mcornella:
Pushed 16de51404773a5e849c25a95e7612e212e7256a7 which uses the same `COMPLETION_WAITING_DOTS` variable to change the default red dots string, while turning off line wrapping.

PR #9709, opened 2021-03-05
fix(terraform): use faster method to get workspace

vladistan:
## Standards checklist:

## Changes:

The purpose of this change is to speed up tf_prompt_info() function.   This function relies on 'terraform workspace show' which could be very slow operation.   To speed up I made the following changes

- Check for .terraform/environment directory.  If it doesn't exist we are probably not in the terraform stack or the stack doesn't have custom workspace
- To get the workspace one just needs to look into .terraform/environment file.  It contains the workspace name.  No need to launch the terraform command because it is very slow
- Also, added tf_prompt_info to the list of prompt_info functions in lib/prompt_info.  So that theme writers are aware of it

### DIFF
--- a/lib/prompt_info_functions.zsh
+++ b/lib/prompt_info_functions.zsh
@@ -18,6 +18,7 @@ function chruby_prompt_info \
   vi_mode_prompt_info \
   virtualenv_prompt_info \
   jenv_prompt_info \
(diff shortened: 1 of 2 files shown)


vladistan:
CC @mcornella :-)

PR #9713, opened 2021-03-06
Removing Yaourt and Pacaur from Archlinux Plugin

nowshed-imran:
## Standards checklist:

## Changes:
Both Yaourt and Pacaur are unmaintained AUR helper now. No one actually uses them anymore. They don't need to be in Archlinux Plugin anymore.

1. Yaourt: https://archlinux.fr/yaourt-en
2. Pacaur: https://github.com/E5ten/pacaur

### DIFF
--- a/plugins/archlinux/README.md
+++ b/plugins/archlinux/README.md
@@ -55,55 +55,6 @@ plugins=(... archlinux)
 | trsu    | trizen -Syua --no-confirm          | Same as `trupg`, but without confirmation                           |
 | upgrade | trizen -Syu                        | Sync with repositories before upgrading packages                    |
 
-#### YAOURT
-
-| Alias   | Command                            | Description                                                         |
-|---------|------------------------------------|---------------------------------------------------------------------|
-| yaconf  | yaourt -C                          | Fix all configuration files with vimdiff                            |
-| yain    | yaourt -S                          | Install packages from the repositories                              |
-| yains   | yaourt -U                          | Install a package from a local file                                 |
-| yainsd  | yaourt -S --asdeps                 | Install packages as dependencies of another package                 |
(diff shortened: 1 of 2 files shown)


mcornella:
@ratijas what should we do with `pacaur`? I haven't found it to be discontinued in my research (https://github.com/E5ten/pacaur doesn't show it, and it still appears in the [AUR helpers](https://wiki.archlinux.org/title/AUR_helpers#Comparison_tables) wiki).

nowshed-imran:
I think `pacaur` can be kept for now. It has been forked and maintenance is happening.

ratijas:
> @ratijas what should we do with `pacaur`? I haven't found it to be discontinued in my research (https://github.com/E5ten/pacaur doesn't show it, and it still appears in the [AUR helpers](https://wiki.archlinux.org/title/AUR_helpers#Comparison_tables) wiki).

Pacaur do be doing fine. They tried to burry it, but the demand and popularity is what brought it back.

nowshed-imran:
> Pacaur do be doing fine. They tried to burry it, but the demand and popularity is what brought it back.

Except no one tried to bury it. Its creator abandoned it. It was left in limbo for more than a year. `yay` and some other AUR helper came out which are excellent. Someone forked it and maintaining `pacman` compatibility. It is not literally dead, but better alternatives are here.

mcornella:
It it's still alive there is no reason to remove it. I'll consider doing that when it's no longer available as a package.

nowshed-imran:
> If it's still alive there is no reason to remove it. I'll consider doing that when it's no longer available as a package.

Yaourt should be removed at the least. In case of pacaur, I agree.

nowshed-imran:
I agree about `pacaur`.

PR #9745, opened 2021-03-17
add skip-worktree alias

3olkin:
Sometimes you want to modify files managed by Git locally but you do not want Git to manage that change.

## Changes:

- add git update-index --skip-worktree
- add git update-index --no-skip-worktree


fmalcher:
Could you please add the new aliases to the table in the README, too? 😊 Thanks!

Z4RX:
Sorry, I forgot about readme. Done now.

robbyrussell:
Thanks for this, and sorry for the long silence. Nicely done, by the way, with the undo and the listing alias alongside it.

The reason I'm closing it is the flag itself rather than the code. `git help update-index` warns about exactly the use case in the description: "Users often try to use the assume-unchanged and skip-worktree bits to tell Git to ignore changes to files that are tracked. This does not work as expected, since Git may still check working tree files against the index when performing certain operations." Making that one keystroke away tends to cost people more than it saves, and we'd rather not point more users at it.

Thanks again for taking the time.

PR #9749, opened 2021-03-19
mkcd plugin - create directory and change into it

KrzysztofHajdamowicz:
## Standards checklist:

## Changes:

- This plugin creates directory and changes into it, it's exactly `mkdir foo && cd foo`


TheTechRobo (commented):
Maybe add a bit more documentation in the readme (say the command and its usage) and maybe make the -p flag optional?

I like this. I think itd be very good for my workflow.

jeffmhubbard:
Why not just use `take` that's included in `lib/functions.zsh`?

```
function take() {
  mkdir -p $@ && cd ${@:$#}
}
```

rwmitchell:
mkcd is a logical combination of mkdir  and cd - what does take mean?

TheTechRobo:
Oh - the take command seems to have the exact same functionality with a different name. :thinking:

image

KrzysztofHajdamowicz:
> Oh - the take command seems to have the exact same functionality with a different name. 🤔
> 
> image

Maybe we should rename `take` into `mkcd` as new name will be more self-explanatory to users? This issue already contains 3 of them.

jeffmhubbard:
> > Oh - the take command seems to have the exact same functionality with a different name. thinking
> > image
> 
> Maybe we should rename `take` into `mkcd` as new name will be more self-explanatory to users? This issue already contains 3 of them.

Renaming it would be break existing workflows. `take()`  was added 12 years ago... Also it's documented on the [cheatsheet](https://github.com/ohmyzsh/ohmyzsh/wiki/Cheatsheet) pretty clearly.

rwmitchell:
maybe so, but it is still a poor name.  going from mkdir to mkcd makes a lot of sense.  'take' is meaningless.

rwmitchell:
i'm not 'take'ing anything.  A directory is being created and then cd'd into.

KrzysztofHajdamowicz:
Instead of renaming `take`, we can add an alias `mkcd` -> `take` or implement this as a plugin.

rwmitchell:
if anything, it should be the opposite with a note that take is being depricated.
clean things up instead of adding more junk.  'take' is junk.

jeffmhubbard:
> Instead of renaming `take`, we can add an alias `mkcd` -> `take` or implement this as a plugin.

An alias is preferable to adding a plugin for an existing feature.

rwmitchell:
it isn't a plugin, it's a 1 line function:
take is a shell function from /Users//.oh-my-zsh/lib/functions.zsh

take () {
	mkdir -p $@ && cd ${@:$#}
}

certainly no one is using it a script and anyone needing it can alias take to the new mkcd

jeffmhubbard:
> it isn't a plugin, it's a 1 line function:
> take is a shell function from /Users//.oh-my-zsh/lib/functions.zsh
> 
> take () {
> mkdir -p $@ && cd ${@:$#}
> }
> 
> certainly no one is using it a script and anyone needing it can alias take to the new mkcd

This PR is for a plugin.

rwmitchell:
I agree this should not be added as a plugin, but 'take' should be renamed.

`mk[TAB]` allows someone to easily discover mkcd when starting to type mkdir.  'take' is only discovered by hunting it down.

mcornella:
`take` was made into its own thing that supports different stuff (f68d65d32a4fce41f08ac229614b7e46baf5e916). I've added `mkcd` as an equivalent to `takedir`.

Thank you for your feedback!

PR #9755, opened 2021-03-22
add plugin move-cursor-in-buffer

gernotfeichter:
## Standards checklist:

## Changes:

- a plugin that exposes widgets to quickly jump to the beginning or end of multi-line commands (buffer)
- also works for single-line commands

## Other comments:
Unfortunately the builtin beginning-of-buffer-or-history behaves inconsistently: It jumps to the beginning for multi-line commands but retrieves the history when using single-line commands.


mcornella:
This already exists in zsh: `zle beginning-of-line` and `zle end-of-line`.

mcornella:
Correction: `zle -al | grep buffer`:
```
.beginning-of-buffer-or-history
.end-of-buffer-or-history
```
Both bound to <kbd>Alt+\<</kbd> and <kbd>Alt+\></kbd> respectively.

gernotfeichter:
zle beginning-of-buffer-or-history jumps to the beginning of the current line in a multiline string yes, but unfortunately this does not work for single line strings because there it retrieves the history, which I consider inconsistent behavior.

This plugin behaves consistently for both scenarios!

mcornella:
That's true. A couple of things:
- The plugin will be a part of OMZ, which is already licensed under MIT, so please remove that.
- I like `beginning-of-buffer` and `end-of-buffer` names better, it follows the common convention and it's less verbose.

gernotfeichter:
100% agreed, I pushed the updated changes!

gernotfeichter:
Since this was not merged for a long period, I am withdrawing this PR.

A couple of retrospective thoughts:
1. I think that said functions `beginning-of-buffer` better belong to zsh directly, rather than oh-my-zsh, since this is something zsh should offer.
2. While I personally dislike the "or-something-else" mentality of `.beginning-of-buffer-or-history` and `.end-of-buffer-or-history` I learned to accept it, for the following reason: As a user, I know whether I am editing a multi-line string, or a single-line string. In a multi-line string, `Ctrl+Pos1` gets me to the beginning of the buffer, while in a single line string, `Pos1` gets me to the beginning of the line. Therefore, if I make no mistake as a user, I get the behavior that I want.

PR #9761, opened 2021-03-25
fix(pass): update `pass` completion to ff5ac38

rmuit:
## Standards checklist:

## Changes:

Quote $@ in completion.

## Other comments:

I started using zsh yesterday, and my command completion broke. 

It happens at least with filenames containing colons. The details of the breakage are: "the completion suggestions are all weird so they are useless". (I haven't looked at further details.)

The completion command contains a $@, and I my first guess was maybe trying to quote it would help. Lo and behold, it did! So I stopped trying to understand much about zsh completion after that.

### DIFF
--- a/plugins/pass/_pass
+++ b/plugins/pass/_pass
@@ -133,7 +133,7 @@ _pass_complete_entries_helper () {
 	local IFS=$'\n'
 	local prefix
 	zstyle -s ":completion:${curcontext}:" prefix prefix || prefix="${PASSWORD_STORE_DIR:-$HOME/.password-store}"
-	_values -C 'passwords' ${$(find -L "$prefix" \( -name .git -o -name .gpg-id \) -prune -o $@ -print 2>/dev/null | sed -e "s#${prefix}/\{0,1\}##" -e 's#\.gpg##' -e 's#\\#\\\\#' | sort):-""}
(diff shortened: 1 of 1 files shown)


mcornella:
Can you send this patch upstream? This file is a bundle of the official pass autocompletion.

rmuit:
Upstream (https://www.passwordstore.org/) suggests sending an e-mail including a patch to a mailinglist. I did so, although I botched the subject line: https://lists.zx2c4.com/pipermail/password-store/2021-March/thread.html

I am not super sure who is going to respond there. I can understand you're not merging this PR but I'd like it to remain visible here for the while because visibility (or possibility for others to respond) isn't excellent with this upstream method.

rmuit:
Oh - also, it just re-broke silently after 1.5 days of working, and isn't fixed by me removing ~/.zcompdump* as suggested elsewhere. Mysterious.

mcornella:
Maybe it working was a false positive. I'd bet we need to quote the colon characters so that it's not treated as a separator.

Something like this maybe:
```
${variable//:/\\:}
```

Note the double backslash. I haven't looked at the code so I can't tell you where to do that, but maybe this helps you.

mcornella:
Or since it's using sed anyway maybe you can use quote colon characters with it, before it does the backslash duplication.

rmuit:
Thank you for updating and closing this with a second/proper commit -- and sorry for not getting back to this.

(There's never time... I've been slugging along with broken password completion until things suddenly worked because of this.)

PR #9762, opened 2021-03-27
fix essembeh theme to show ssh host ip correctly

pythonwood:
when I ssh to 192.168.1.35 from 192.168.1.3(my pc), SSH_CONNECTION env is:

$ echo $SSH_CONNECTION
192.168.1.3 49966 192.168.1.35 22
$ echo $SSH_CONNECTION | awk '{print $(NF-1)}'
192.168.1.35

I think, It is better prompt 192.168.1.35 , no 192.168.1.3. 

## Standards checklist:

## Changes:

- [...]

## Other comments:

OK because NR >= 2 when SSH_CONNECTION is set in env

### DIFF
--- a/themes/essembeh.zsh-theme
+++ b/themes/essembeh.zsh-theme
@@ -32,7 +32,7 @@ local ZSH_ESSEMBEH_COLOR="green"
 local ZSH_ESSEMBEH_PREFIX=""
 if [[ -n "$SSH_CONNECTION" ]]; then
 	# display the source address if connected via ssh
-	ZSH_ESSEMBEH_PREFIX="%{$fg[yellow]%}[$(echo $SSH_CONNECTION | awk '{print $1}')]%{$reset_color%} "
+	ZSH_ESSEMBEH_PREFIX="%{$fg[yellow]%}[$(echo $SSH_CONNECTION | awk '{print $(NF-1)}')]%{$reset_color%} "
 	# use red color to highlight a remote connection
 	ZSH_ESSEMBEH_COLOR="red"
 elif [[ -r /etc/debian_chroot ]]; then 


mcornella:
The comment says to display the source address, not the host, so this is by design. This constitutes a breaking change.

pythonwood:
I look into PR #9595 and merge commit [172b32c9](https://github.com/ohmyzsh/ohmyzsh/commit/172b32c90dadf09bcc1e9b562a68c1c282bf514e)

author said:

> - prefix with remote address for ssh shells

so, in my opinion,  **remote address**  is the remote target ip which is $(NF-1) in my case. 

                            ssh
    local(192.168.1.3)     -------->      remote(192.168.1.35)

---

before my pr, when I ssh to  192.168.1.35 and login a shell , shell prefix is [192.168.1.3],  is it just the correct point?

mcornella:
@essembeh can you clarify if you meant to display the host IP or the IP from where you're connected?

essembeh:
Hi

Thanks for reaching me up @mcornella, and sorry if the comments were not clear enough, English is not my native language.

>   - prefix with remote address for ssh shells

I wanted to display the address from where I initiated the SSH connection, because it fits my needs.
When in a shell, getting the address is easy, `ip a` does the job, plus you get all the addresses, not just the one openssh is listening to.
Having my local address (192.168.1.3 in your case @pythonwood) shows me that 1/ I'm connected to a remote server and 2/ from where I'm connected, if I used a VPN, a LAN, ipv6, or if my local address changed ...

Feel free to adapt this theme but I would appreciate if you create a new one, so that I do not end up having this strange (to me) behavior on all my servers on my next update ;)

Thanks

pythonwood:
ok, new theme is another way

Issue #9770, opened 2021-03-31
Dirhistory plugin: "no such widget" errors


stephane-caron:
**Describe the bug**
Using alt+left results in the following error written to the terminal:
```bash
No such widget `.dirhistory_zle_dirhistory_back'
```
Similar errors appear for alt+right and the other plugin shortcuts.

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Enable this plugin 'dirhistory'
2. Type alt+left
3. See error

**Expected behavior**
The plugin should go to the previous directory, as described in [its README](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/dirhistory).

**Screenshot**
![image](https://user-images.githubusercontent.com/1189580/113176835-31c40880-924d-11eb-848e-37d87b0aec15.png)

**Desktop (please complete the following information):**

 - OS / Distro: Ubuntu
 - Latest ohmyzsh update?: Yes
 - ZSH Version: 5.8
 - Terminal emulator: `gnome-terminal`
 - Plugins: ``plugins=(autojump dirhistory docker git python)``

PR #9771, opened 2021-03-31
fix(dirhistory): make sure to call built-in zle widgets

stephane-caron:
## Standards checklist:

## Changes:

- Make sure built-in widgets are used for `kill-buffer` and `accept-line`

## Other comments:

Fixes https://github.com/ohmyzsh/ohmyzsh/issues/9770

### DIFF
--- a/plugins/dirhistory/dirhistory.plugin.zsh
+++ b/plugins/dirhistory/dirhistory.plugin.zsh
@@ -108,16 +108,16 @@ function dirhistory_forward() {
 # Bind keys to history navigation
 function dirhistory_zle_dirhistory_back() {
   # Erase current line in buffer
-  zle kill-buffer
-  dirhistory_back 
-  zle accept-line
+  zle .kill-buffer
+  dirhistory_back
+  zle .accept-line
 }
 
 function dirhistory_zle_dirhistory_future() {
   # Erase current line in buffer
-  zle kill-buffer
+  zle .kill-buffer
   dirhistory_forward
-  zle accept-line
+  zle .accept-line
 }
 
 zle -N dirhistory_zle_dirhistory_back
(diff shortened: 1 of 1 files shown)


mcornella:
I can't reproduce #9770 so I can't verify this works. Are you sure that's the problem? It would look like there's a binding to `.dirhistory_zle_dirhistory_back` instead of `dirhistory_zle_dirhistory_back`.

stephane-caron:
> I can't reproduce #9770 so I can't verify this works. Are you sure that's the problem? It would look like there's a binding to `.dirhistory_zle_dirhistory_back` instead of `dirhistory_zle_dirhistory_back`.

I agree, but there's no mention of `.dirhistory_zle_dirhistory_back` in the code of the plugin :thinking: Although without the dot there definitely is:

```bash
~/.oh-my-zsh/plugins/dirhistory$ grep ".dirhistory_zle_dirhistory_back" dirhistory.plugin.zsh | wc -l
0
~/.oh-my-zsh/plugins/dirhistory$ grep "dirhistory_zle_dirhistory_back" dirhistory.plugin.zsh | wc -l
7
```

Sorry I'm not fluent enough with the zsh line editor to understand what's going on here.

stephane-caron:
Oh my bad! I found the guilty function in my `~/.oh-my-zsh/custom/redo.zsh` bindings (I totally forgot about that :sweat_smile:):

```zsh
#!/bin/zsh

# Redefine default Enter command so it inserts last command if the buffer is
# empty. Source: https://stackoverflow.com/a/46525918

function accept-line() {
    [ -z "$BUFFER" ] && zle up-history
    zle ".$WIDGET"
}

zle -N accept-line
```

Explains the dot :wink: 

This PR only makes the plugin robust against being messed up with like this :grin: I see that the `per-directory-history` plugin does it as well.

mcornella:
Thanks @stephane-caron, great explanation!

PR #9772, opened 2021-03-31
Rails Plugin: check for bundle if no other 'rails' context is found

faelin:
## Standards checklist:

## Changes:

- added check for `bundle` if no other rails context is found, and runs as `bundle exec rails $@`. This duplicates the format used by the `rake` section of this plugin.


mcornella:
According to https://github.com/ohmyzsh/ohmyzsh/pull/2923#issuecomment-1023243304 `rails` shouldn't be executed with bundler. So the current version (try for `rake`, don't for `rails`) seems to be the correct one.

The `-e script/server` branch seems to also be correct, since it was meant to check if rails scripts are found under the `script` folder, and it appears to be a really old construct.

PR #9782, opened 2021-04-03
Added "SSH Update Known Hosts" plugin

anatolykopyl:
## Standards checklist:

## Changes:
Added "SSH Update Known Hosts" plugin. It wraps ssh and in case the user has to update the known_hosts file it will prompt to do it automatically and continue connecting to the server.

### DIFF
--- a/plugins/sshukh/README.md
+++ b/plugins/sshukh/README.md
@@ -1,0 +1,42 @@
+# sshukh-zsh-plugin
+*SSH Update Known Hosts*
+
+Prompts to update `known_hosts` file for you if needed.
+
+To install, run the following command:
+
+```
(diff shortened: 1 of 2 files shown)


septs:
I am not a reviewer, but I suggest split to multi pull request.

robbyrussell:
Hi @anatolykopyl,

First off, thank you so much for the proposed contribution. This is a clever approach to dealing with those pesky unknown host issues that hit us when we're dealing with SSH. Having looked over the code, we've decided that while it is quite clever...we are uncomfortable adding a wrapper around the SSH command on people's computers.

There are some potential issues that would be difficult to debug--in particular, if the env for the subprocess is different. 

For anyone who would like do use something like this -- we'd encourage you to inject that into either a) custom plugins folder or b) in your .zshrc configuration.

Thanks again for contributing and we hope you'll be around with more ideas, too!

Cheers,
Robby (and Marc)

PR #9791, opened 2021-04-07
zshrc: Include `export GPG_TTY=$TTY`

alanmynah:
## Standards checklist:

- [x] The PR doesn't replicate another PR which is already open. As best as I could. Searched for different GPG related terms, but found nothing close. Sorry if I missed any. 

## Changes:
Include `export GPG_TTY=$TTY` to zshrc

When setting up GPG commit signing on Github ([example walkthrough article](https://sabbour.me/setting-up-signed-git-commits-on-macos/)), can get this error
`gpg: signing failed: Inappropriate ioctl for device`
https://github.com/keybase/keybase-issues/issues/2798

a fix is proposed to be to `export GPG_TTY=$(tty)` and put that into `~/.bash_profile`. 
but according to this post 
https://unix.stackexchange.com/a/608921
`export GPG_TTY=$TTY` is better. 

The change has been tested, here's the test commit
https://github.com/alanmynah/cheat-sheets/commit/8ff6913f4928e55cd6cfc6a511ae5d34684a9161

### DIFF
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -99,3 +99,7 @@ source $ZSH/oh-my-zsh.sh
 # Example aliases
 # alias zshconfig="mate ~/.zshrc"
 # alias ohmyzsh="mate ~/.oh-my-zsh"
+
+# To enable signed commits on GitHub
+# export GPG_TTY=$TTY
+


mcornella:
Hi Michael, given that this is git and gpg related I think this would be better suited for the git plugin. The [`gpg-agent` plugin already has this](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/gpg-agent/gpg-agent.plugin.zsh#L9). The zshrc template file will only apply that to new installs, and I don't think this should be set for everyone.

You can say why would people need to enable the whole git plugin for this fix, but if they don't want to they can just add this line to their zshrc file, which is what the guides recommend.

alanmynah:
@mcornella, thanks for your review. Fair enough, that makes sense. Closing.

PR #9798, opened 2021-04-08
fix(suse): avoid refresh attempts for Zypper queries

IlyaBizyaev:
## Standards checklist:

## Changes:

- On invocation, `zypper`'s query commands will attempt to refresh repositories. Since they lack root privileges, they will not be able to save new package information on the device, but can take lots of time if many repositories are enabled. For a series of consecutive query commands, a user has to wait for the pointless 'repository refresh' every time. This patch asks `zypper` to not refresh repositories. If a user intends to do that, they would need to use `sudo` anyway, which the `zref` alias already does.


mcornella:
Thanks!

Issue #8591, opened 2020-02-02
`abs` and `aur` should be removed from archlinux plugin


jeffmhubbard:
I'm having some trouble with the `pacupd` alias now that I'm using aurutils. The usual output is followed by aurutil's usage text.

```
jeff@pythos ~
➜ % pacupd                                                                                            130 ↵ 16:01:11
:: Synchronizing package databases...
 testing                                   25.7 KiB  1116 KiB/s 00:00 [#######################################] 100%
 core                                     133.5 KiB  3.03 MiB/s 00:00 [#######################################] 100%
 extra                                   1640.3 KiB  17.8 MiB/s 00:00 [#######################################] 100%
 community-testing                        596.5 KiB  12.7 MiB/s 00:00 [#######################################] 100%
 community                                  4.8 MiB  17.5 MiB/s 00:00 [#######################################] 100%
 aur-builds is up to date
usage: aur [command]

available commands:
build		fetch		pkglist		repo		search		sync
depends		graph		query		repo-filter	srcver		vercmp

available user commands:
jeff@pythos ~
➜ %                                                                                                     1 ↵ 16:06:16
```

I've looked at the plugin, and found pacman and the AUR helpers all seem to have a similiar "update" alias, calling `sudo aur` and `sudo abs` if those commands are available.
```sh
if (( $+commands[abs] && $+commands[aur] )); then
  alias pacupd='sudo pacman -Sy && sudo abs && sudo aur'
elif (( $+commands[abs] )); then
  alias pacupd='sudo pacman -Sy && sudo abs'
elif (( $+commands[aur] )); then
  alias pacupd='sudo pacman -Sy && sudo aur'
else
  alias pacupd='sudo pacman -Sy'
fi
```

I think these checks for `aur` and `abs` may be dated and broken. Aurutils is the only package I can find providing `aur` and Arch Build System (`abs`?) was deprecated in 2017. This is only a problem after installing aurutils because it provides the `aur` command, but `sudo aur` is not the correct way to use it. I think `alias pacupd='sudo pacman -Sy'` is the only valid bit of that if statement left.

mcornella:
Can you open a PR removing those so we can discuss it there? It should be easy now that the plugin was refactored.

mcornella:
Fixed in #9803.

PR #9803, opened 2021-04-09
feat(archlinux)!: remove deprecated commands `aur` and `abs`

jeffmhubbard:
## Standards checklist:

## Changes:

- Remove calls to aur and abs commands

## Other comments:

See #8591

### DIFF
--- a/plugins/archlinux/README.md
+++ b/plugins/archlinux/README.md
@@ -26,7 +26,7 @@ plugins=(... archlinux)
 | pacrep       | `pacman -Si`                           | Display information about a package in the repositories          |
 | pacreps      | `pacman -Ss`                           | Search for packages in the repositories                          |
 | pacrmorphans | `sudo pacman -Rs $(pacman -Qtdq)`      | Delete all orphaned packages                                     |
-| pacupd       | `sudo pacman -Sy && <abs/aur refresh>`[¹](#f1) | Update and refresh local package, ABS and AUR databases  |
+| pacupd       | `sudo pacman -Sy`                      | Update and refresh local package, ABS and AUR databases          |
 | pacupg       | `sudo pacman -Syu`                     | Sync with repositories before upgrading packages                 |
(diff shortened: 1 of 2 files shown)


ratijas:
haha, I was about to do it too. Actually, there's no need to remove them completely.  According to Arch Linux news, `abs` was superseded by `asp`: [Deprecation of ABS tool and rsync endpoint](https://archlinux.org/news/deprecation-of-abs/). Relevant [discussion](https://bbs.archlinux.org/viewtopic.php?id=226693) on forum.

ratijas:
> Other comments:
> See #859

That links to something completely irrelevant. Probably a typo or just a template leftover?

jeffmhubbard:
Even is `asp` is a dropin replacement for `abs`, the fact that it was broken for so long with no issues raised, indicates that no one was using it. I'm in favor dropping it, until someone who needs it comes along.

As for `aur`, I believe this was an older tool that is no longer maintained. `aurutils` provides a `aur` command, which is why I opened #8591. I asked Alad on IRC, but he was unaware of any other current tools also providing `aur`.

ratijas:
> it was broken for so long with no issues raised

Either no one gives a damn about raising an issue, or no one really uses such complicated aliases. The whole concept of conditional combination is strange enough.

But on the other hand, sets of aliases are a valuable source of educational material. If I were a n00bish Arch Linux wannabe, I'd probably be happy to learn about `asp` and its role in the whole ecosystem just by looking at the README of this plugin.

So, in the end, I'm voting to replace them with `asp update` or something. And to make them unconditional, to reduce confusion.

jeffmhubbard:
I think I'd like this PR to be just about removing what doesn't work. If you want to add support for other tools and support those changes, that would be great. I'm not using `asp`, so I don't want to take on that responsibility.

ratijas:
OK then.  Please, make sure to update ² and ³ references in README.

ratijas (approved):

mcornella:
Thank you both!

PR #9817, opened 2021-04-12
Fix(nvm): Disable lazy load when auto load nvmrc

AmirHosseinKarimi:
## Standards checklist:

## Changes:

- Disable lazy load when auto load nvmrc

## Other comments:

.nvmrc auto load will try to set specific version of node in .nvmrc file and according to code, if there is not any .nvmrc we will try to set default node version.
But if lazy load was enabled it going wrong and generate something like this:
```
Reverting to nvm default version
nvm is not compatible with the npm config "prefix" option: currently set to "Now using node v14.13.1 (npm v6.14.8)
/home/amir/.nvm/versions/node/v14.13.1"
Run `npm config delete prefix` or `nvm use --delete-prefix v14.13.1` to unset it.
```
To fix this problem I have disabled lazy load when .nvmrc auto load was enabled.

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -1,3 +1,8 @@
+# Disable lazy load when auto load nvmrc
+if (( $+NVM_AUTOLOAD )); then
+  unset NVM_LAZY
+fi
+
 # See https://github.com/nvm-sh/nvm#installation-and-update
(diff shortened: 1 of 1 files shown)


Zeragamba (approved):

carlosala:
This is a matter of user configuration. You should only have enabled one of the two, it doesn't make sense to have both as autoload is going to call `nvm` on startup.

PR #9830, opened 2021-04-15
added option to disable promotion

qpwo:
## Standards checklist:

  - Had to `export` flag to make it reach the update script

## Changes:

* Add a flag in config template with setting commented out
* Add a check for that flag in the update script

## Other comments:

I saw https://github.com/ohmyzsh/ohmyzsh/pull/5680 and I think having a setting is a fair compromise. For us impulsive shoppers who don't need more ideas of things to buy.

### DIFF
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -26,6 +26,9 @@ ZSH_THEME="robbyrussell"
 # Uncomment the following line to disable bi-weekly auto-update checks.
 # DISABLE_AUTO_UPDATE="true"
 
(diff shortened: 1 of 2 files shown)


mcornella:
With my utmost respect, this is not a problem, you can just ignore the message.

TheTechRobo:
Agreed. OhMyZsh isn't made of money, and even if it was I don't see a problem in showing a message about it. It's just like how youtubers with millions of subscribers still ask you to subscribe. Nothing wrong with it.

PR #9845, opened 2021-04-19
fix(wezm): fix dirty git repo glyph width

rotolof:
Some terminal emulators rendered this glyph as a wide character, causing a duplicate character glitch with tab autocompletion.

## Standards checklist:

## Changes:

- Added `%{%G<glyph>%}` around the gliph.

## Other comments:

I found the solution in this comment by @romkatv 
https://github.com/ohmyzsh/ohmyzsh/issues/7945#issuecomment-508724741


romkatv:
This will break all terminals that render this glyphs as wide.

rotolof:
@romkatv you are right, I didn't notice that in those terminals it's eating a space.
Is there a way to fix this once for all?

romkatv:
Fix the buggy terminals. This is a wide character. The Unicode spec says so.

rotolof:
Thanks for the tip! In my case it's CLion's terminal, I'll report the bug to them.

PR #9847, opened 2021-04-20
feat(plugins): add plugin for `deno`

### LINKED ISSUE #9165: add deno plugins

carlosala:
## Standards checklist:

## Changes:

I added completion and some useful aliases for `deno`. The completion function caches the user's version of `deno` and it copy the shell completions to `$ZSH/completions`, as @mcornella requested in #9591.
This closes #9363 #9591 #9681 PR because this one improves them. It also closes #9165 issue.
It would be great to have some feedback about the completion function, for me is working but it's the first time that I tried something like this. I got some code from #9557.

Thanks!


carlosala:
Hi!
I changed radically the code in this second version, it does the same with less code so, better!
Could someone revise the plugin?
Thanks @mcornella for all your work!

mcornella:
Hmm this uses `$ZSH/completions` which is a git tracked path and bound to be removed from the codebase, can you do something akin to the `npm` plugin which uses `$ZSH_CACHE_DIR` instead?

carlosala:
Hi @mcornella!
I'm very noob on this, but I have red the output of `npm completion` and it's quite different to the output of `deno completions zsh`. If you have time, could you revise how to create a proper completion function? How can I load the output of `deno completions zsh` without saving a `_deno` file in a directory that is in the `$fpath`?
Thanks!

mcornella:
Hmm you're right. Maybe put it in the same plugin folder and add a .gitignore, that seems more organized than using the phantom completions folder.

carlosala:
Hey @mcornella. It's done!

mcornella:
Thanks Carlo!

carlosala:
A tu! @mcornella

PR #9864, opened 2021-04-26
feat(plugins): add fnm plugin

sebenik:
## Standards checklist:

## Changes:

- Add plugin for [fnm](https://github.com/Schniz/fnm), Fast and simple Node.js version manager.

Similar plugin was attempted but never accepted. Also plugin at #8789 is outdated so I created a new PR.


Schniz:
Awesome!
fnm provides shell completions. So maybe the plugin could `eval $(fnm completions --shell=zsh)`?

sebenik:
@Schniz great idea, makes it compatible with current installed fnm version.
I'll change it.

sebenik:
Unfortunately output of `fnm completions --shell=zsh` isn't meant to be sourced and adding completions evaluating it is not working.

PR stays unchanged and is ready.

mcornella:
You can follow the same approach as in the [deno](https://github.com/ohmyzsh/ohmyzsh/blob/70ab2928/plugins/deno/deno.plugin.zsh) and [gh](https://github.com/ohmyzsh/ohmyzsh/blob/70ab2928/plugins/gh/gh.plugin.zsh plugins.

sebenik:
@mcornella thanks for the tip. Done.

mcornella:
This needs the `.gitignore` file as well to ignore the generated `_fnm` file.

sebenik:
Tnx, added.

PR #9868, opened 2021-04-28
fix(nvm): add check for .nvmrc to improve cd performance

jalavosus:
Going off of [this stackoverflow comment](https://stackoverflow.com/questions/23556330/run-nvm-use-automatically-every-time-theres-a-nvmrc-file-on-the-directory#comment104625451_39519460), adding `[[ -a .nvmrc ]] || return` improves the performance of `cd` when running this autoload function.

## Standards checklist:

## Changes:

- Add `[[ -a .nvmrc ]] || return` to `load-nvmrc` function

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -39,6 +39,7 @@ fi
 # Adapted from: https://github.com/nvm-sh/nvm#zsh
 if (( $+NVM_AUTOLOAD )); then
   load-nvmrc() {
+    [[ -a .nvmrc ]] || return
     local node_version="$(nvm version)"
     local nvmrc_path="$(nvm_find_nvmrc)"
 


Zeragamba:
Does this also handled checking for .nvmrc files in parent directories?

jalavosus:
@Zeragamba the autoload function takes care of that regardless of this change.

The purpose of this change is to avoid a very slow check for finding the file when possible, since the autoload function is meant to check the local directory and use that `.nvmrc` first if found.

Zeragamba:
Right, but from my understanding, if you `cd` into a subdirectory, the autoload won't trigger because there's no `.nvmrc` there.

jalavosus:
autoload loads parent directory `.nvmrc` files already -- all my change does is do a check for a `.nvmrc` in the _current_ directory, and use that if it's there. If there's no nvmrc in the changed-to directory, then the rest of autoload runs, which includes calling `nvm_find_nvmrc`, a function which will find the `.nvmrc` in the closest parent directory relative to the current directory. 

See this screenshot as an example:
Screen Shot 2022-06-26 at 13 23 29

PR #9876, opened 2021-05-01
fix(command-not-found): remove invalid argument for PackageKit

sevendials:
## Standards checklist:

## Changes:

- Removes `--` argument to PackageKit command-not-found that was inadvertently added

## Other comments:
`pk-command-not-found` doesn't accept `--` as an argument. This is what happens when `--` is the first argument:
```
foo@bar ~  gem 
zsh: --: command not found...
```
...


mcornella:
Thanks and sorry for the error, I assumed this was standard behavior.

PR #9879, opened 2021-05-03
fix(git plugin): check for origin head

AnatoleLucet:
## Standards checklist:

## Changes:

Adds a condition in `git_main_branch` to check for the origin remote's HEAD.
Not all repos use `master`, `main`, or `trunk` as their default branch. Some might choose `develop`, `release`, etc... This change makes this a bit more dynamic while still keeping the fallbacks for (e.g.) local repos with no origin remote.

## Other comments:

I'm not a pro in bash / zsh scripting, so there might be room for improvement here.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -33,6 +33,11 @@ function work_in_progress() {
 function git_main_branch() {
   command git rev-parse --git-dir &>/dev/null || return
   local branch
+  branch=$(command git symbolic-ref refs/remotes/origin/HEAD &>/dev/null | sed "s@^refs/remotes/origin/@@")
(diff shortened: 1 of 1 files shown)


mcornella:
A version of this was released some time ago but removed after complaints that now `gcd`  (`git checkout develop`) and `gcm` (`git checkout main`) did the same thing.

AnatoleLucet:
Then maybe origin's head could be the fallback if the default branch isn't master, main, or trunk?

mcornella:
I appreciate what you're trying to do, but we don't need to support all different configurations. The intent behind `git_main_branch` was to allow the transition from master to another branch name, and it's done that for the most common name alternatives. It doesn't need to do more.

PR #9880, opened 2021-05-03
feat(nodenv): add plugin for `nodenv`

slaytanic:
## Standards checklist:

## Changes:

- Adds the "nodenv" plugin, similar to the ones for rbenv and pyenv

### DIFF
--- a/plugins/nodenv/nodenv.plugin.zsh
+++ b/plugins/nodenv/nodenv.plugin.zsh
@@ -1,0 +1,38 @@
+# This plugin loads nodenv into the current shell and provides prompt info via
+# the 'nodenv_prompt_info' function.
+
+FOUND_NODENV=$+commands[nodenv]
+
+if [[ $FOUND_NODENV -ne 1 ]]; then
+    nodenvdirs=("$HOME/.nodenv" "/usr/local/nodenv" "/opt/nodenv" "/usr/local/opt/nodenv")
+    for dir in $nodenvdirs; do
+        if [[ -d $dir/bin ]]; then
+            export PATH="$PATH:$dir/bin"
+            FOUND_NODENV=1
+            break
+        fi
+    done
+fi
+
+if [[ $FOUND_NODENV -ne 1 ]]; then
+    if (( $+commands[brew] )) && dir=$(brew --prefix nodenv 2>/dev/null); then
+        if [[ -d $dir/bin ]]; then
+            export PATH="$PATH:$dir/bin"
+            FOUND_NODENV=1
+        fi
+    fi
+fi
+
+if [[ $FOUND_NODENV -eq 1 ]]; then
+    eval "$(nodenv init - zsh)"
+    function nodenv_prompt_info() {
+        echo "$(nodenv version-name)"
+    }
+else
+    # fallback to system node
+    function nodenv_prompt_info() {
+        echo "system: $(node -v 2>&1 | cut -c 2-)"
+    }
+fi
+
+unset FOUND_NODENV nodenvdirs dir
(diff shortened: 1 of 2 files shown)


designorant:
I've been using this successfully for a short while now so thanks @slaytanic! Always nice to have something baked in rather than rely on external plugins that vary in quality.

One thing I noticed was the `export PATH="$dir/bin:$PATH"` swap with `export PATH="$PATH:$dir/bin"` [compared to rbenv plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/rbenv/rbenv.plugin.zsh#L10-L20) but I'm not sure what to make of it.

plugins/nodenv/nodenv.plugin.zsh
```diff
@@ -0,0 +1,38 @@
+# This plugin loads nodenv into the current shell and provides prompt info via
+# the 'nodenv_prompt_info' function.
+
+FOUND_NODENV=$+commands[nodenv]
+
+if [[ $FOUND_NODENV -ne 1 ]]; then
+    nodenvdirs=("$HOME/.nodenv" "/usr/local/nodenv" "/opt/nodenv" "/usr/local/opt/nodenv")
+    for dir in $nodenvdirs; do
+        if [[ -d $dir/bin ]]; then
+            export PATH="$PATH:$dir/bin"
```
bostonaholic, line 10:
```suggestion
            export PATH="$dir/bin:$PATH"
```

plugins/nodenv/nodenv.plugin.zsh
```diff
@@ -0,0 +1,38 @@
...
+            FOUND_NODENV=1
+            break
+        fi
+    done
+fi
+
+if [[ $FOUND_NODENV -ne 1 ]]; then
+    if (( $+commands[brew] )) && dir=$(brew --prefix nodenv 2>/dev/null); then
+        if [[ -d $dir/bin ]]; then
+            export PATH="$PATH:$dir/bin"
```
bostonaholic, line 20:
```suggestion
            export PATH="$dir/bin:$PATH"
```

plugins/nodenv/nodenv.plugin.zsh
```diff
@@ -0,0 +1,38 @@
...
+            FOUND_NODENV=1
+        fi
+    fi
+fi
+
+if [[ $FOUND_NODENV -eq 1 ]]; then
+    eval "$(nodenv init - zsh)"
```
bostonaholic, line 27:
```suggestion
    eval "$(nodenv init --no-rehash - zsh)"
```

bostonaholic:
I've been using a version of this for quite a while and it works great. I've made a few recommended changes, also mentioned by @designorant. I would love to see this incorporated into the packaged plugins.

plugins/nodenv/nodenv.plugin.zsh
mcornella: Is there a specific reason why the nodenv folder must be first on `$PATH`?

According to [1] the required one for this is the shims folder, and I understand that this is taken care of by `nodenv init`.

[1]: https://github.com/nodenv/nodenv#understanding-shims

mcornella: If there is no other reason I have deferred to appending the directory then. There isn't much difference, but it is slightly more secure keeping the usual directories first in `$PATH`. Thanks for the review!

mcornella:
Thank you everyone who contributed!

plugins/nodenv/nodenv.plugin.zsh
bostonaholic: Sounds good @mcornella. You might want to look at those lines in the rbenv plugin and make the adjustment if there is a flaw in how the `PATH` modification is done.

PR #9881, opened 2021-05-03
feat(git): add `develop` branch name detection

max-ae:
## Standards checklist:

## Changes:

Some repositories use branch names other than `develop` for the development branch. This PR adds a `develop` branch name detection inspired by the `main` branch name detection.

Closes #10088

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -42,7 +42,7 @@ function git_main_branch() {
   echo master
 }
 
-# Check for develop and similar named branches
+# Check for develop and similarly named branches
 function git_develop_branch() {
   command git rev-parse --git-dir &>/dev/null || return
   local branch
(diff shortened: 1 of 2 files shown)


plugins/git/git.plugin.zsh
```diff
@@ -42,6 +42,19 @@ function git_main_branch() {
   echo master
 }
 
+# Check for develop and similar named branches
+function git_develop_branch() {
+  command git rev-parse --git-dir &>/dev/null || return
+  local branch
+  for branch in dev devel development; do
```
lodenrogue, line 49: What if you have all 3 in your repo?

max-ae: https://github.com/ohmyzsh/ohmyzsh/blob/f58e9a55d11477821f69eed57559f945b2773ce9/plugins/git/git.plugin.zsh#L36
The same that would happen if someone had both a `main` and `trunk` branch. The first existing one would get selected.
But I suppose that would be a very rare edge case.

PR #9883, opened 2021-05-03
Use global variable to disable git_prompt_status.

brunoborges:
Useful when working on heavy/large git repos. Allows me to disable temporarily in the shell, or globally.


lodenrogue (approved):

brunoborges:
hey @mcornella, wanna check on this one?

mcornella:
This has the same effect as a global `git config`:
```sh
git config --global oh-my-zsh.hide-status 1
```

PR #9908, opened 2021-05-12
feat(git): add alias `gstnu` for 'git status --untracked-files=no'

kunalg:
`gstnu` == `git status no untracked`

The list of untracked files can divert focus from main edited files.

Signed-off-by: Kunal Gangakhedkar <kunal.gangakhedkar@gmail.com>

## Standards checklist:

## Changes:

- Add new alias `gstnu`

## Other comments:

N/A


robbyrussell:

The idea is reasonable, but the name doesn't work in this plugin: apart from bare `gst`, every `gst*` alias is a stash alias, and `gstu` is already `git stash --include-untracked`. So `gstnu` would sit one keystroke away from an alias about untracked files that means something quite different.

Closing for that reason. Worth knowing that @fpaupier proposed the identical alias in #10538, so you weren't the only one who wanted it.

kunalg:
@robbyrussell : 

How about using `gsnu` as the alias?
* sticks to the `gs` == `git status` logic branch
* `nu` == `no untracked`

Alternatively, `gsnut`?
* `gs` - same logic as previous comment
* `nut` = short for `no un-tracked`

Let me know if either of these work - will raise a new PR 🙂 

I've cross-checked - neither of these aliases exist in the git plugin.
```shell
~ ❯ grep -E '.*gsnu.*' ~/.oh-my-zsh/plugins/git/git.plugin.zsh

~ ❯ echo $?
1
```

PR #9909, opened 2021-05-13
fix(pyenv): `pyenv init --path` before `pyenv init -`

danielpatrickdotdev:
## Standards checklist:

## Changes:

- In line with [recent changes](https://github.com/pyenv/pyenv/pull/1920) to pyenv, `pyenv init --path` must be run before `pyenv init -`.

## Other comments:

Without this, I get the following warning when using the latest pyenv version:
```bash
WARNING: `pyenv init -` no longer sets PATH.
Run `pyenv init` to see the necessary changes to make to your configuration.
```

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -29,6 +29,8 @@ if [[ $FOUND_PYENV -ne 1 ]]; then
 fi
 
 if [[ $FOUND_PYENV -eq 1 ]]; then
+    PYENV_VERSION="$(pyenv --version | cut -d ' ' -f 2)"
+    is-at-least 1.2.27-21 "$PYENV_VERSION" && eval "$(pyenv init --path)"
     eval "$(pyenv init - --no-rehash zsh)"
     if (( $+commands[pyenv-virtualenv-init] )); then
         eval "$(pyenv virtualenv-init - zsh)"


olemoign:
As mentioned [here](https://github.com/pyenv/pyenv/issues/1906#issuecomment-835027647) by one of the pyenv maintainers, this seems to break some pyenv functionalities, like `pyenv shell`.
I'm a big oh my zsh noob, but this feels like something that can't be solved here, isn't it? oh my zsh acts at ".zshrc level", where pyenv now needs both .zprofile and .zshrc.

hjpotter92:
> As mentioned [here](https://github.com/pyenv/pyenv/issues/1906#issuecomment-835027647) by one of the pyenv maintainers, this seems to break some pyenv functionalities, like `pyenv shell`.
> I'm a big oh my zsh noob, but this feels like something that can't be solved here, isn't it? oh my zsh acts at ".zshrc level", where pyenv now needs both .zprofile and .zshrc.

Not all display managers play nice with `~/.profile`/`~/.zprofile` either. GDM/XDM for eg. <https://unix.stackexchange.com/a/4628>

plugins/pyenv/pyenv.plugin.zsh
```diff
@@ -29,6 +29,8 @@ if [[ $FOUND_PYENV -ne 1 ]]; then
 fi
 
 if [[ $FOUND_PYENV -eq 1 ]]; then
+    PYENV_VERSION="$(pyenv --version | cut -d ' ' -f 2)"
```
miketheman, line 32: **suggestion:** ‏If a variable is set, unset it at the bottom of the plugin on this line:
https://github.com/ohmyzsh/ohmyzsh/pull/9909/files#diff-b6865657813739f9d07639d70642974ef35ff2b72529eefa76e1b9bc08852d84L46

With the current behavior, this breaks the value displayed in the prompt for some reason.

NeilGirdhar:
This fix is incorrect.  Please read the corresponding issue thread carefully.  If this fix is accepted, pyenv will only work for interactive shells.

danielpatrickdotdev:
Totally agreed @NeilGirdhar - this fix works for some use cases but clearly not for all. I've done a bit of reading since putting this together, but didn't get around to checking back here. Closing this now

PR #9913, opened 2021-05-15
Add plugin for gentoolinux

MattiaGaspa:
## Standards checklist:

## Changes:

- New gentoolinux plugin


robbyrussell:

The fork this came from has been deleted, so there's no longer a branch we could merge or update. Closing for that reason rather than because Gentoo support is unwelcome... if you or anyone else wants to open a fresh one, it'd get a look. One thing worth changing if you do: most of the aliases here wrap `sudo`, and we'd rather not hide that from people behind a short alias.

PR #9914, opened 2021-05-16
lib: Emit OSC 7 escape sequence (current host and working directory) for many Unix terminal emulators

ntninja:
## Standards checklist:

   * Don't know, but the code follows the style already present in the respective file.

## Changes:

- Report current host and working directory to many Unix terminal emulators in addition to iTerm/iTerm2

## Other comments:

Previously this was only emitted on on macOS with Apple's iTerm.app (and
compatible clones like iTerm2), but it is used by other terminal emulators
as well to obtain the actual current working directory wiht symlinks intact.
All non-supporting terminal emulators tested gracefully ignored this value,
so emit this on these as well in case they (or some other app masquarading
as them) add future support for this value.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -110,28 +110,52 @@ add-zsh-hook precmd omz_termsupport_precmd
 add-zsh-hook preexec omz_termsupport_preexec
 
 
-# Keep Apple Terminal.app's current working directory updated
-# Based on this answer: https://superuser.com/a/315029
+# Keep terminal emulator's current working directory correct,
+# even if the current working directory path contains symbolic links
+#
+# References:
+#   * Apple's Terminal.app: https://superuser.com/a/315029
+#   * iTerm2: https://iterm2.com/documentation-escape-codes.html (iTerm2 Extension / CurrentDir+RemoteHost)
+#   * Konsole: https://bugs.kde.org/show_bug.cgi?id=327720#c1
+#   * libvte (gnome-terminal, mate-terminal, …): https://bugzilla.gnome.org/show_bug.cgi?id=675987#c14
+#      - Apparently had bug before ~2012 were it would display the unknown OSC 7 code
+#
+# As of May 2021 mlterm, PuTTY, rxvt, screen, termux & xterm simply ignore the unknown OSC.
+#
 # With extra fixes to handle multibyte chars and non-UTF-8 locales
 
-if [[ "$TERM_PROGRAM" == "Apple_Terminal" ]] && [[ -z "$INSIDE_EMACS" ]]; then
-  # Emits the control sequence to notify Terminal.app of the cwd
-  # Identifies the directory using a file: URI scheme, including
-  # the host name to disambiguate local vs. remote paths.
-  function update_terminalapp_cwd() {
-    emulate -L zsh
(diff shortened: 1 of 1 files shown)


lib/termsupport.zsh
```diff
@@ -110,28 +110,52 @@ add-zsh-hook precmd omz_termsupport_precmd
...
+  # Percent-encode the host and path names.
+  local URL_HOST URL_PATH
+  URL_HOST="$(omz_urlencode -P $HOST)" || return 1
+  URL_PATH="$(omz_urlencode -P $PWD)" || return 1
+
+  # common control sequence (OSC 7) to set current host and path
+  printf "\e]7;%s\a" "file://${URL_HOST}${URL_PATH}"
+}
+
+#FIXME: Use a precmd hook instead of a chpwd hook to avoid contaminating output – why!?
```
ntninja, line 158: This one I'm unsure about:

The previous code had `precmd` with the above comment, but I sincerely do not understand what they mean by *avoid contaminating out* here?

ntninja:
Also is the commit supposed to be `feat` or `fix` in this case?

ngkz:
foot also supports OSC 7. Can you add that to the `${TERM}` case?

lib/termsupport.zsh
```diff
@@ -110,28 +110,52 @@ add-zsh-hook precmd omz_termsupport_precmd
...
+#
+# Identifies the directory using a file: URI scheme, including
+# the host name to disambiguate local vs. remote paths.
+function omz_termsupport_chpwd {
+  emulate -L zsh
+
+  [[ "$INSIDE_EMACS" == *term* ]] && return
+
+  case "${TERM}" in
+    xterm*|putty*|rxvt*|konsole*|mlterm*|alacritty|screen*|tmux*)
```
christianparpart, line 138:
```suggestion
    xterm*|putty*|rxvt*|konsole*|mlterm*|alacritty|screen*|tmux*|contour*)
```
While being at it (in #11599), you could extend it here accordingly as well. :)

lib/termsupport.zsh
```diff
@@ -110,28 +110,52 @@ add-zsh-hook precmd omz_termsupport_precmd
...
-  add-zsh-hook precmd update_terminalapp_cwd
-  # Run once to get initial cwd set
-  update_terminalapp_cwd
-fi
+# Emits the control sequence to notify many terminal emulators
+# of the cwd
+#
+# Identifies the directory using a file: URI scheme, including
+# the host name to disambiguate local vs. remote paths.
+function omz_termsupport_chpwd {
```
christianparpart, line 132: This name, `omz_termsupport_chpwd`, is badly named. `pwd` stands for `print working dir`. what would probably be a better fit is `omz_termsupport_set_cwd` (with `cwd` as in `current working directory`). That would be much more common sense.

christianparpart: ok, but then the hook is named very badly. I don't want to force you to switch it though. Thx for the heads-up ;)

lib/termsupport.zsh
christianparpart: Why was this comment ignored?

christianparpart: Never mind. I can't read. You redid the commit elsewhere. Thanks :)

PR #9926, opened 2021-05-23
feat(dnf)!: Rework of the aliases

signed-log:
## Standards checklist:

## Changes:

I reworked the aliases to use the official `zypper` aliases lists, but as it is totally different aliases I added the check for `old_dnf` variables to fallback to former aliases

## Other comments:

...

### DIFF
--- a/plugins/dnf/README.md
+++ b/plugins/dnf/README.md
@@ -36,3 +36,27 @@ plugins=(... dnf)
 | dcall          | `sudo dnf clean all`         | Clean cache              |
 | dshell         | `sudo dnf shell`             | Enters dnf shell         |
 | dsdl           | `sudo dnf download --source` | Download package source  |
+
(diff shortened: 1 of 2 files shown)


eclectic-coding:
I will totally `+1` this PR. I was just looking to see if `autoremove` was an available alias... this would be a great improvement

robbyrussell:

Renaming the whole set is the part we can't take. The existing `dnf*` aliases are in a lot of people's muscle memory, and putting them behind `OLD_DNF` means everyone's shell changes by default. We reverted a similar rename in #13485 last December for that reason. The plugin has also been reworked for dnf5 since you opened this, so the diff no longer applies.

One thing worth rescuing: @eclectic-coding's comment was really asking for an `autoremove` alias. If you'd like to open a small PR adding just that, it'd be an easy yes.

Closing this one. Thanks for taking the time.

PR #9935, opened 2021-05-27
fix(pyenv): fix pyenv PATH settings with a warning

NeilGirdhar:
## Standards checklist:

## Changes:

- PyEnv expects you to update your PATH in your .zshenv or .zprofile.

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -9,51 +9,62 @@ else
   FOUND_PYENV=0
 fi
 
+# Look for pyenv and try to load it (will only work on interactive shells)
 if [[ $FOUND_PYENV -ne 1 ]]; then
   pyenvdirs=("$HOME/.pyenv" "/usr/local/pyenv" "/opt/pyenv" "/usr/local/opt/pyenv")
   for dir in $pyenvdirs; do
-    if [[ -d $dir/bin ]]; then
+    if [[ -d "$dir/bin" ]]; then
       FOUND_PYENV=1
-      DISCOVERED_PYENV=1
       break
     fi
   done
-fi
 
-if [[ $FOUND_PYENV -ne 1 ]]; then
-  if (( $+commands[brew] )) && dir=$(brew --prefix pyenv 2>/dev/null); then
-    if [[ -d $dir/bin ]]; then
-      FOUND_PYENV=1
-      DISCOVERED_PYENV=1
+  if [[ $FOUND_PYENV -ne 1 ]]; then
+    if (( $+commands[brew] )) && dir=$(brew --prefix pyenv 2>/dev/null); then
+      if [[ -d "$dir/bin" ]]; then
+        FOUND_PYENV=1
+      fi
     fi
   fi
-fi
 
-if [[ $FOUND_PYENV -eq 1 ]]; then
-  if [[ $DISCOVERED_PYENV -eq 1 ]]; then
-    echo pyenv is badly configured in this shell.  Follow the instructions here:
-    echo https://github.com/pyenv/pyenv/blob/master/README.md#advanced-configuration
-    echo On linux, this means adding the following lines to your .profile and sourcing it in your .zprofile:
-    echo export PYENV_ROOT=$dir
-    echo export PATH="\$PYENV_ROOT/bin:\$PATH"
-    echo eval "\$(pyenv init --path)"
-    # Configuring in .zshrc only makes pyenv available for interactive shells.
+  # If we found pyenv, load it but show a caveat about non-interactive shells
+  if [[ $FOUND_PYENV -eq 1 ]]; then
+    cat <<EOF
+Found pyenv, but it is badly configured. pyenv might not work for
+non-interactive shells (for example, when run from a script).
+${bold_color}
+To fix this message, add these lines to the '.profile' and '.zprofile' files
+in your home directory:
+
+export PYENV_ROOT="${dir/#$HOME/\$HOME}"
+export PATH="\$PYENV_ROOT/bin:\$PATH"
+eval "\$(pyenv init --path)"
+${reset_color}
+For more info go to https://github.com/pyenv/pyenv/#installation.
+EOF
+
+    # Configuring in .zshrc only makes pyenv available for interactive shells
     export PYENV_ROOT=$dir
     export PATH="$PYENV_ROOT/bin:$PATH"
     eval "$(pyenv init --path)"
   fi
+fi
+
+if [[ $FOUND_PYENV -eq 1 ]]; then
   eval "$(pyenv init - --no-rehash zsh)"
-  if (( $+commands[pyenv-virtualenv-init] )); then
+
+  if (( ${+commands[pyenv-virtualenv-init]} )); then
     eval "$(pyenv virtualenv-init - zsh)"
   fi
+
   function pyenv_prompt_info() {
     echo "$(pyenv version-name)"
   }
 else
-  # Fall back to system python.
+  # Fall back to system python
   function pyenv_prompt_info() {
     echo "system: $(python -V 2>&1 | cut -f 2 -d ' ')"
   }
 fi
 
-unset FOUND_PYENV DISCOVERED_PYENV pyenvdirs dir
+unset FOUND_PYENV pyenvdirs dir
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -4,7 +4,6 @@
 # Load pyenv only if command not already available
 if command -v pyenv &> /dev/null && [[ "$(uname -r)" != *icrosoft* ]]; then
   FOUND_PYENV=1
-  DISCOVERED_PYENV=0
 else
   FOUND_PYENV=0
 fi
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -4,6 +4,7 @@
 # Load pyenv only if command not already available
 if command -v pyenv &> /dev/null && [[ "$(uname -r)" != *icrosoft* ]]; then
     FOUND_PYENV=1
+    DISCOVERED_PYENV=0
 else
     FOUND_PYENV=0
 fi
@@ -12,8 +13,8 @@ if [[ $FOUND_PYENV -ne 1 ]]; then
     pyenvdirs=("$HOME/.pyenv" "/usr/local/pyenv" "/opt/pyenv" "/usr/local/opt/pyenv")
     for dir in $pyenvdirs; do
         if [[ -d $dir/bin ]]; then
-            export PATH="$PATH:$dir/bin"
             FOUND_PYENV=1
+            DISCOVERED_PYENV=1
             break
         fi
     done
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
Fixes https://github.com/ohmyzsh/ohmyzsh/issues/9902

Kriechi:
IMO https://github.com/ohmyzsh/ohmyzsh/pull/9909 is a more correct fix.

NeilGirdhar:
@Kriechi No, because the PATH should be set in your .profile, which is sourced by your .zsh_profile.  This is explained in the thread.  That fix will wreck pynev for non-interactive shells.

Kriechi:
If I understand your PR correctly, you are just removing all logic for changing the PATH to find the pyenv binary itself - this is independent of the shims and installed python versions. That's what `pyenv init --path` is for.
The change you are proposing renders the pyenv plugin almost useless as there will be no init at all if the plugin can't find the pyenv binary.

NeilGirdhar:
> If I understand your PR correctly, you are just removing all logic for changing the PATH to find the pyenv binary itself - this is independent of the shims and installed python versions.

If that logic belongs anywhere, it's in your `.profile`.

> that's what `pyenv init --path` is for.

That line, which should also be in your `.profile`, makes the Python shims accessible.

> The change you are proposing renders the pyenv plugin almost useless as there will be no init at all if the plugin can't find the pyenv binary.

That's because most of what is being done in the pyenv plugin is being done in the wrong place.  If pyenv needs to work in non-interactive shells too, then most of its setup needs to happen in `.profile`.

Please read the corresponding issue thread.  I was as confused as you are until someone corrected my understanding.

Kriechi:
I don't think such a dogmatic approach is helping here. The pyenv plugin was working fine for lots of people.
The last pyenv release does not change anything in the way we should discover where pyenv is installed - only where pyenv itself installs Python versions and shims.

If you remove the path discovery for pyenv itself, as this PR currently tries to do, it will break e.g., for everybody using macOS and homebrew or custom install locations or any of the other options that are covered in the code you are removing. This means the omz pyenv plugin is useless and can be disabled/removed all together. This is the same mechanism that countless other omz plugins utilize to discover their own binaries as well - most of them with multiple options to account for different Linux distributions, macOS, custom and non-standard paths as fallback.

> If pyenv needs to work in non-interactive shells too, then most of its setup needs to happen in .profile.

Yes, ideally the path for the shims (as generated by `pyenv init --path`) would be placed into a file that is loaded for both interactive and non-interactive shells. I feel like this is out of scope of a omz plugin (and always has been). The current pyenv plugin works perfectly fine for interactive sessions, and even for any zsh-based scripts that you launch from your typical terminal. If you need full non-interactive sessions to work, then yes, go ahead and add the correct paths to your .profile - but why do you want to break a working system for everybody else?

NeilGirdhar:
I understand what you're saying, but I think we should be steering people towards properly setting up their system rather than encouraging a setup that only works some of the time, and inevitably leads to confusion when it stops working.

Maybe a reasonable compromise is to check whether pyenv is set up correctly in the plugin (maybe just check if `PYENV_ROOT` is set).  If it isn't, we can print a message explaining what needs to be added to your .profile.  If you really insist on not breaking anything, then we can do the setup for them under that guard.

This way, for the people that have correctly set up their systems, the pyenv plugin won't do anything.  For the people that were relying on the old behaviour, the pyenv plugin will continue to work and will print a message explaining how to improve their setup.

plugins/pyenv/pyenv.plugin.zsh
```diff
@@ -3,32 +3,6 @@
...
-    if (( $+commands[brew] )) && dir=$(brew --prefix pyenv 2>/dev/null); then
-        if [[ -d $dir/bin ]]; then
-            export PATH="$PATH:$dir/bin"
-            FOUND_PYENV=1
-        fi
-    fi
-fi
-
-if [[ $FOUND_PYENV -eq 1 ]]; then
     eval "$(pyenv init - --no-rehash zsh)"
```
januszm, line 6: I think after recent update to `pyenv` also this line should be changed to

```sh
eval "$(pyenv init --path --no-rehash zsh)"
```

NeilGirdhar: See my recent push.

NeilGirdhar:
@Kriechi Please see my recent push and let me know if it works on your system?

plugins/pyenv/pyenv.plugin.zsh
```diff
@@ -22,13 +23,23 @@ fi
...
-            export PATH="$PATH:$dir/bin"
             FOUND_PYENV=1
+            DISCOVERED_PYENV=1
         fi
     fi
 fi
 
 if [[ $FOUND_PYENV -eq 1 ]]; then
+    if [[ $DISCOVERED_PYENV -eq 1 ]]; then
+        echo pyenv is badly configured.  Add the following lines to your .profile and source it in your .zprofile:
```
Kriechi, line 34: nitpick: pyenv is not the one being "badly" configured - it's the user's shell.

NeilGirdhar: Updated, thanks.

Kriechi:
@NeilGirdhar I still disagree with your approach - but at this point it is a matter of principle of what this plugin should provide to the user. I can't speak for the other OMZ plugins, but I feel like they all try to make it as easy as possible for the users by trying to do the "sane" thing and doing as much as possible of the heavy lifting.

I believe this PR is a step backwards compared to the usefulness this plugin has in its current state on the master branch.
There are multiple other plugins that take care of setting the PATH so the user doesn't have to mess around with it. And at least for me, this has always worked - like magic. All my shells, interactive or not, behave as I expect them to. I do not have a `.profile` or `.zprofile` in my HOME. I do have a `/etc/zprofile` that calls `path_helper` on macOS. Maybe thats all we need to be happy.

I've been using this simple change to set up the shim paths, basically since the last pyenv release a few weeks ago - and it has worked flawlessly for me:
```
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -29,6 +29,7 @@ if [[ $FOUND_PYENV -ne 1 ]]; then
 fi

 if [[ $FOUND_PYENV -eq 1 ]]; then
+    eval "$(pyenv init --path)"
     eval "$(pyenv init - --no-rehash zsh)"
     if (( $+commands[pyenv-virtualenv-init] )); then
         eval "$(pyenv virtualenv-init - zsh)"
``` 

Let's see if others have opinions, preferences, or other comments either way.

NeilGirdhar:
> I feel like they all try to make it as easy as possible for the users by trying to do the "sane" thing and doing as much as possible of the heavy lifting.

Isn't that what we have in this pull request?  It does all the heavy lifting like you said.

> I believe this PR is a step backwards c

I don't  see how it's a "step backwards".   It continues to work for your use case, and it solves the error message that people are seeing.

It also guides users to properly configure their systems.

> I've been using this simple change to set up the shim paths, basically since the last pyenv release a few weeks ago - and it has worked flawlessly for me:

That's because you're not using pyenv in non-interactive shells.  If you did, then you would find that it isn't configured.  Please read the issue thread since it seems like you missed this important point.  I also added a comment explaining that in the pull request.

Kriechi:
@neilg please read my comment, since it seems like you missed my statement about me using interactive and non-interactive shells and not having surprises with it.

So instead of pyenv printing a message, now the plugin prints a message about an apparent misconfiguration? That's not a step forward in my eyes. 

I acknowledge though that it does indeed "do" it - I missed the part of the actual PATH setting after printing the messages. That makes me somewhat happier.

januszm:
I got to this thread because today I updated pyenv and got a message saying that pyenv init should now be run with the `--path` option if this is to work as intended. Whatever happens in this PR, let's make adding the `pyenv` plugin in oh-my-zsh enough to override any other manipulation of ENV vars and running `pyenv init`.

NeilGirdhar:
@Kriechi You may be intersted that pyenv [explicitly states here](https://github.com/pyenv/pyenv/blob/master/README.md#advanced-configuration) that the configuration should be done in `.zprofile`.

agoose77:
Precursory disclaimer that I'm NAE.

I think the wider issue here is that (interactive) shell configuration shouldn't really be setting up `PATH`s. I'm sure that we're all on the same page here with how shell initialisation works, but for posterity I'll restate it here :)

From what I remember (I no longer use OMZ directly, I use [zinit](https://github.com/zdharma/zinit)), OMZ does everything via `.zshrc`, which is sourced whenever a new interactive shell is launched. This is *fine*, and it works, but it means that anything placed here won't be available unless the consumer has been launched from an interactive shell. Shell initialisation is rather complicated, and every distribution has slightly different defaults, but I *think* ZSH is pretty predictable. The main issue there of course is whether you *actually* get a login shell at login (GDM :eyes:)
There's a diagram [here](https://blog.flowblok.id.au/2013-02/shell-startup-scripts.html) for anyone interested

So, *best* practice is to split the different configurations into different files. I prefer to put `PATH` in `.zprofile` (although, with GDM I place it in `.profile` and source that with compatibility mode to support ssh logins) but there's motivation to put it in `.zshenv` too. 

At the end of the day, power-users can always "do the right thing" and set their dotfiles appropriately. Your changes @NeilGirdhar look like they restore the old behaviour, which is probably all we need here.

NeilGirdhar:
@tilgovi Would you mind taking a look at this pull request?

tilgovi:
I'm guessing I got tagged in here for review because I've contributed changes before. I haven't updated pyenv yet, so I haven't run into these issues. Some time ago I did stop using the pyenv omz plugin because I wanted things set up for non-interactive shells. This is what I have in my .zprofile:

``` zsh
if command -v pyenv 1>/dev/null 2>&1; then
    eval "$(pyenv init -)"
fi
```

So, take from that what you will.

I think having the interactive shell print instructions and set things up is a reasonable idea.

NeilGirdhar:
@tilgovi I was hoping that you would be able to accept this pull request since you committed to this file before?  I wonder who I should ping?

TheTechRobo:
they can still "accept" the PR, if they want.

you want to ping @robbyrussel, @mcornella, and @larsen_carter

mcornella:
Ok, I restructured the code and reworded the message. I agree with the general approach here, so if there's no issue with my changes this has my +1.

NeilGirdhar:
@mcornella Thanks for your excellent edits.  Do I need to do anything further?

mcornella:
Just your approval was enough, thanks!

webknjaz:
Hi, so after I followed the instructions in the message, I was still getting the warning.
When I did some debugging, turned out that	`~/.zprofile` wasn't being loaded because zsh in a normal terminal emulator app does get executed as a login shell.
Of course, adding `source $HOME/.zprofile` right before loading the omz entrypoint "fixes" it, but it's a workaround that is incorrect.

I suggest prompting the users to add the snippet to the `~/.zshenv` script instead. It gets loaded in all the modes.

Ref: https://serverfault.com/a/901476/111608

NeilGirdhar:
> When I did some debugging, turned out that ~/.zprofile wasn't being loaded because zsh in a normal terminal emulator app does get executed as a login shell.

Did you source it in your `.profile` like the message says?

webknjaz:
> Did you source it in your `.profile` like the message says?

I tried but that didn't change anything. OTOH I didn't re-login. So maybe that's why it didn't pick up the vars... With `.zprofile`, though, it was there through the reboot.

> The instructions match those given by pyenv itself, which I linked above.

I see that now, I missed that it should go into both files. Will try to have it in both places and make a proper reboot.
I wonder if the message can be improved to emphasize that a reboot may be necessary.

NeilGirdhar:
> I tried but that didn't change anything. OTOH I didn't re-login. So maybe that's why it didn't pick up the vars... With .zprofile, though, it was there through the reboot.

Great.  That's how it's supposed to work, I think.

> I see that now, I missed that it should go into both files. Will try to have it in both places and make a proper reboot.
I wonder if the message can be improved to emphasize that a reboot may be necessary.

It's not a bad idea.  You could also link the pyenv instructions in the message.  😄

PR #9939, opened 2021-05-28
Added supercomputer-plugin

NormanTUD:
## Standards checklist:

I hope this way of contributing is correct. I've added a set of zsh-plugins that I regularly use on the supercomputer of my university that others may find useful.


robbyrussell:
Thanks for this, and sorry it sat unanswered for so long. There's clearly a lot of real work in here.

It isn't a fit for the bundled plugins, though. Several of the functions rely on `ws_list`, `ws_allocate` and `whypending`, which are your site's workspace tooling rather than Slurm itself, so they wouldn't work for most people who enabled it. There are also some general-purpose helpers mixed in, like `countdown` and `pretty_csv`, that would land in everyone's shell under unnamespaced names.

One thing worth knowing either way: the file is named `supercomputer.zsh`, and Oh My Zsh only loads `supercomputer.plugin.zsh`, so as it stands it wouldn't be picked up at all. Renaming it means this works as a custom plugin in `$ZSH_CUSTOM/plugins`, which is supported and survives updates.

Thanks again for offering it up.

PR #9956, opened 2021-06-09
fix(git): Makes gcs alias consistent with gcsm

gotmax23:
`gcsm` is an alias for `git commit -s -m`, while `gcs` is an alias for
`git commit -S`. One (-S) cryptographically signs commits, while the
other (-s) signs-off on commits

## Standards checklist:

## Changes:

- Changes `alias gcs='git commit -S'` to `alias gcs='git commit -s'`

## Other comments:
If we want to add other aliases with `-S` like we have for `git commit -s` (eg: `alias gcsm='git commit -s -m'`), we could capitalize the `s` (eg: `alias gcSm='git commit -S -m'`).

Related: #7460.

...

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -57,7 +57,7 @@ plugins=(... git)
 | gcp                  | git cherry-pick                                                                                                                  |
 | gcpa                 | git cherry-pick --abort                                                                                                          |
 | gcpc                 | git cherry-pick --continue                                                                                                       |
-| gcs                  | git commit -S                                                                                                                    |
+| gcs                  | git commit -s                                                                                                                    |
 | gd                   | git diff                                                                                                                         |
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry you never got a reply at the time. You're right that the naming is inconsistent... `s` means `-S` in `gcs`, `gcss` and `gcssm`, but `-s` in `gcsm`.

The trouble is which end to fix. Repointing `gcs` from `-S` to `-s` means anyone who types it out of habit quietly stops signing their commits, and they wouldn't get any warning that it happened. That's the kind of silent change we try hard to avoid. Marc closed the same swap back in 2018 in #6105 for the same reason.

`gcss` covers signing plus signoff together if that's the combination you were after.

Closing for that reason. Thanks for raising it.

PR #9960, opened 2021-06-09
fix(install.sh): chsh failed for user without password

yangxingwu:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...


mcornella:
`sudo` is not needed to change your own shell, you probably entered the password incorrectly or had another issue with input redirection (open an issue with details of your setup and how you ran the installer).

Merging this PR would mean users without administrator rights wouldn't be able to change their own shell.

PR #9972, opened 2021-06-13
$identities always an empty array

ayoubeddaouddi:
the value returned by zstyle is not set to $identities variable.
so the feature of adding keys automatically is not working.

I don't know how to fix it 😢.


TheTechRobo:
I think you should file an Issue, not a Pull Request for this.

PR #9973, opened 2021-06-14
Update git plugin to delete "gone" branches from local copy

### LINKED ISSUE #9710: add command for pruning local branches that have been merged
`git`

when I'm working on a lot of branches my local git copy contains a lot of already merged branches which I need to clean manually

i would like to propose a "gbp" command which removes all already merged branches

currently, I'm using this inside my global `.gitconfig`
```
[alias]
	branch-prune = "!git fetch -p && for b in $(git for-each-ref --format='%(if:equals=[gone])%(upstream:track)%(then)%(refname:short)%(end)' refs/heads); do git branch -d $b; done"
```

**Additional context**
the command removes all branches that are merged and do not contain additional changes locally

**Related Issues**
None

DocEmmetBrown:
## Standards checklist:

## Changes:

- update plugin/git/git.plugin.zsh 
- update plugin/git/README.md

## Other comments:

Fixes #9710   
the naming of the alias `gbdg` would refer to "git branch delete gone".


TheTechRobo:
The maintianers have a huge backlog of PRs; if you really want it merged, you could try asking on the Discord server.

plugins/git/git.plugin.zsh
```diff
@@ -74,6 +74,7 @@ alias gb='git branch'
 alias gba='git branch -a'
 alias gbd='git branch -d'
 alias gbda='git branch --no-color --merged | command grep -vE "^([+*]|\s*($(git_main_branch)|$(git_develop_branch))\s*$)" | command xargs git branch -d 2>/dev/null'
+alias gbdg='git fetch -p && git branch -vv | grep ": gone]" | awk "{print $1}" | xargs git branch -D'
```
antonu17, line 77: Since you're using `awk`, you can save on grep. Also, a good practice to use `command` built-in to make sure `awk` or `xargs` will be executed from `$PATH`, and not from a function or a built-in.
To be super safe `git branch --format="%(upstream:track)"` can be used to identify `gone` branches.
```suggestion
alias gbdg="git fetch --prune && git branch --format '%(if:equals=[gone])%(upstream:track)%(then)GONE %(end)%(refname:short)' | command awk '/GONE/{print $2}' | command xargs git branch -D"
```

erikhuizinga: I was happy to see you recent review on this PR, while I was searching for this exact feature!

IMHO, I find the proposed solution in the PR easier to understand than your suggestion, but there might be behavioural differences that I don't understand. If you mean by 'you can save on grep' that `grep` has a performance hit, then maybe that doesn't matter, because `gbdg` would probably be only invoked manually by human users, not automatically in performance-sensitive situations where every millisecond counts. Please explain if and why you think otherwise, as I might be misunderstanding you.

antonu17: Hello @erikhuizinga,

I don't think there are any significant performance penalties calling grep. It's just me being an optimizator. xD
Your code is overall very fine, just one small thing that might make it do a bad thing is `grep ': gone]'`. If on a rare occasion someone writes something something `: gone]` in the first line of their commit message it will be "detected" as a gone branch. And that's what I tried to address via `git branch --format '%(upstream:track)` template.

linkvt: Hi @antonu17, I just found this PR while searching for this feature in omz - would you mind opening a PR with your safer version?

guenhter:
Any update here? This would be really cool so have.

guenhter:
And is this MR actually already covered by https://github.com/ohmyzsh/ohmyzsh/pull/9250

PR #7851, opened 2019-05-15
Mercurial: add aliases for hg log

michalkielan:
Add aliases for mercurial hg log following by the same git convention
```
# git
glgg='git log --graph'
glgp='git log --stat -p'
# hg
hglg='hg log --stat -v'
hglgp='hg log --stat  -p -v'
```

PR #9979, opened 2021-06-16
feat(mercurial): add `hg log` aliases

TheTechRobo:
## Standards checklist:

- [X] I have read the contribution guide and followed all the instructions. - I think...
- [x] The code is stable and I have tested it myself, to the best of my abilities. - I have not but they were taken from #7851 so it should work.

## Changes:

- add a few Mercurial aliases

## Other comments:

It's #7851, but without the conflicts

Co-authored-by: Michał Kielan <kielanm@fb.com>


rwmitchell:
The alias for 'hg pull -u' being 'hgl' doesn't seem obvious, 'hgu' (update) would be better.  That also frees up 'hgl'  to be used for 'hg log' aliases.

TheTechRobo:
@rwmitchell If this is still a problem, open an issue; I don't use mercurial (nor did I make these aliases), and the PR has already been merged.

rwmitchell:
I don’t use this plug-in either because the aliases are whack and I can define my own.

PR #9986, opened 2021-06-17
add plugin to manage mocp player

victorhck:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/mocp/README.md
+++ b/plugins/mocp/README.md
@@ -1,0 +1,14 @@
+# MOCP player Plugin
+
+This plugin adds some functions you can use to manage [mocp player](https://github.com/jonsafari/mocp) in your command line.
+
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry it sat here so long without an answer.

A couple of things get in the way: `mp` and `ms` are already used by the meteor and multipass plugins, and the four volume aliases carry a literal `LEVEL` from the man page, so `m.` would run `mocp --volume=+5 LEVEL` and treat that as a filename. The fork this came from has also been deleted, so there's no branch left to fix it on.

moc itself hasn't had a release since 2016, and nobody else asked for this in the time it's been open, so it isn't one we'd take on to maintain. It works well as a custom plugin in `$ZSH_CUSTOM/plugins` if you still use it.

PR #9991, opened 2021-06-20
Fixed fail to register thefuck alias without $ZSH_CACHE_DIR

ZedG2:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...


mcornella:
`$ZSH_CACHE_DIR` is used throughout the rest of the project. If you're loading the plugin with a plugin manager it should define it automatically. Otherwise you can define it yourself with something like this:
```zsh
ZSH_CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/ohmyzsh"
```

PR #9996, opened 2021-06-21
Add check for $ZSH creation permission

RichardLuo0:
## Standards checklist:

## Changes:

- Check if you can really create $ZSH folder.

## Other comments:
The `if [ -d "$ZSH" ]` give false even if some folders in the middle are inaccessible.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -265,8 +265,18 @@ main() {
     exit 1
   fi
 
-  if [ -d "$ZSH" ]; then
+  is_zsh_valid=true
+  if [ ! -d "$ZSH" ]; then
+    if ! mkdir "$ZSH" ; then
+      echo "${YELLOW}Failed to create the \$ZSH folder ($ZSH).${RESET}"
+      is_zsh_valid=false
+    fi
+  else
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -254,6 +255,7 @@ main() {
       --unattended) RUNZSH=no; CHSH=no ;;
       --skip-chsh) CHSH=no ;;
       --keep-zshrc) KEEP_ZSHRC=yes ;;
+      --skip-git-clone) SKIP_GIT_CLONE=yes ;;
```
mcornella, line 258: What does this do?

RichardLuo0: If for some reason user clone the repo manually before running the script, he can skip git clone process. For example, user wants to share a same clone among multiple linux users.

larson-carter:
Circling back to this older PR, I do think that it is useful. However we would need more testing. I would also need more understanding as to the edge cases of this?

RichardLuo0:
It seems git will fail to clone anyway if the $ZSH is not writeable. I think this is not needed anymore unless you would like an early error.
It has been several years, so I don't quite remember why I did this PR at that time.

The second commit about adding skip-git-clone is an accident push. It is not related to this PR. I will revert it if you want to merge.

PR #9997, opened 2021-06-21
feat(mlh): let users configure the official MLH theme

bartekpacia:
## Standards checklist:

## Changes:

- let users configure MLH theme using environmental variables. User can set
  them conveniently, for example, in their `~/.zshrc` file, instead of being
  forced to override the whole theme.

- make the theme 2 line. Currently, when the user is working on a branch with a long name, it looks
  really awkward:

Screenshot 2021-06-21 at 13 09 08

Now, it looks just cool.
Screenshot 2021-06-21 at 22 09 10

### DIFF
--- a/themes/mlh.zsh-theme
+++ b/themes/mlh.zsh-theme
@@ -12,17 +12,38 @@
 # # # Feel free to customize! # # #
 # # # # # # # # # # # # # # # # # #
 
-# To easily discover colors and their codes, type `spectrum_ls` in the terminal
+# To customize symbols (e.g MLH_AT_SYMBOL), simply set them as environment variables
+# for example in your ~/.zshrc file, like this:
+# export MLH_AT_SYMBOL=" at "
 
-# enable or disable particular elements
-PRINT_EXIT_CODE=true
-PRINT_TIME=true
+# Hint: to easily discover colors and their codes, type `spectrum_ls` in the terminal
 
-# symbols
-AT_SYMBOL=" @ "
-IN_SYMBOL=" in "
-ON_SYMBOL=" on "
-SHELL_SYMBOL="$"
+# right prompt default settings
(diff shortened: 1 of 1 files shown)


themes/mlh.zsh-theme
```diff
@@ -12,17 +12,38 @@
 # # # Feel free to customize! # # #
 # # # # # # # # # # # # # # # # # #
 
-# To easily discover colors and their codes, type `spectrum_ls` in the terminal
+# To customize symbols (e.g MLH_AT_SYMBOL), simply set them as environment variables
+# for example in your ~/.zshrc file, like this:
+# export MLH_AT_SYMBOL=" at "
```
mcornella, line 17: No need to use `export`. Also you should document that the settings need to go before OMZ is sourced in the `.zshrc` file.

themes/mlh.zsh-theme
```diff
@@ -47,24 +68,28 @@ directory() {
 
 # Prints current time
 current_time() {
-  if [ "$PRINT_TIME" = true ]; then
-    echo " $TIME_COLOR%*%f"
+  if [ "$MLH_PRINT_TIME" = true ]; then
+    echo " $MLH_PRINT_TIME%*%f"
```
mcornella, line 72: Typo?

mcornella (changes requested):

themes/mlh.zsh-theme
bartekpacia: yes, thank you:)

themes/mlh.zsh-theme
bartekpacia: thanks:)

mcornella (approved):

mcornella:
Thanks Bartek!

PR #10007, opened 2021-06-26
Added pypi plugin which is used for uploading own python packages in pypi

ToastCoder:
## Standards checklist:

## Changes:

- [...]

## Other comments:

The plugin is used especially for python developers who work on uploading or testing their own python packages. This plugin is used to simplifying the commands which have to be executed every time.


plugins/pypi/pypi.plugin.zsh
```diff
@@ -0,0 +1,6 @@
+alias pypi-check='pip3 install twine setuptools'
```
TheTechRobo, line 1: maybe use `python3 -m pip`?

TheTechRobo (changes requested):
just have some thoughts

plugins/pypi/pypi.plugin.zsh
ToastCoder: I have updated, sir.

hjpotter92:
perhaps a few more aliases for build/test/install [from setuptools](https://pythonhosted.org/an_example_pypi_project/setuptools.html#using-setup-py).

the `pypi-ltest` is confusing, as my initial thought seeing it was this subcommand:

>   `test`              run unit tests after in-place build

ToastCoder:
`pypi-ltest` is the alias used for testing the python package in the local machine, so that you don't have to upload it to pypi if you haven't tested it yet.

ToastCoder:
Could someone at least have a look at the PR? In case if there's anything which I could improve, mentioning them would be great.

TheTechRobo:
as you can see at the top, there are a LOT of pending pull requests. this one will be reviewed eventually by the maintainers.

if you want to expedite it, you might want to ask on the discord server.

NeilGirdhar:
Should this be called twine or pypi?

robbyrussell:

A plugin that's four aliases around twine is more of a personal setup than something we'd want every user to enable, and the `setup.py sdist` calls it wraps have since been deprecated by the Python packaging folks. Two people also raised naming questions here in 2021 that never got resolved.

This works well as a custom plugin in `$ZSH_CUSTOM`. That's a supported setup and it'll survive updates.

PR #10010, opened 2021-06-28
fix(git-auto-fetch): cancel fetch if we don't have permission over git folder

XuehaiPan:
## Standards checklist:

## Changes:

Check directory and file permissions before fetching the Git remotes.

## Other comments:

The zsh prompt will display an error message if the user don't own the git directory.

Screenshot

...


mcornella:
Thanks!

PR #10016, opened 2021-07-01
feat(git): improve primary branch detection (`gcm`)

vladimyr:
## Standards checklist:

## Changes:

Support resolving the main branch from the remote reference.

## Other comments:

This is an improved version of #9879 in the sense that it is remote agnostic - doesn't assume the existence of `origin` remote. (closes #9879)

@mcornella wrote [there](https://github.com/ohmyzsh/ohmyzsh/pull/9879#issuecomment-831295623):
> A version of this was released some time ago but removed after complaints that now `gcd` (`git checkout develop`) and `gcm` (`git checkout main`) did the same thing.

That is true if the _primary_ branch is named dev/develop but as @AnatoleLucet said there are also more exotic _primary_ branch names out there. 🙃

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -29,16 +29,27 @@ function work_in_progress() {
   fi
 }
 
-# Check if main exists and use instead of master
+# Detect name of the primary branch
+# (Use master as a fallback.)
 function git_main_branch() {
+  # Abort if not inside git repository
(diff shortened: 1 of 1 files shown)


vladimyr:
I understand `gcm`/`gcd` symmetry perspective but there should be a shortcut for doing _primary_ branch checkout. I guess that would be more useful than `gcm` restricted to main/master/trunk? Maybe I should create an entirely new alias? 🤔

PR #10030, opened 2021-07-06
Fix virtualenvwrapper plugin's activate virtualenv from a git subtree working firectory

lieryan:
## Standards checklist:

## Changes:

In a `git subtree` working directory, `.git` is a file containing a path
to the parent git working direcotry. This change attempts to find a
virtualenv for a subtree.


robbyrussell (approved):
Thanks for this, and sorry it sat for five years without a reply. In a worktree or a submodule `.git` is a file rather than a directory, so the walk was stepping past the project root. `-e` also matches the `.venv` test on the line above. Merging.

PR #10038, opened 2021-07-10
refactor(lib): change tabs for spaces

ghost:


robbyrussell:
`lib/bzr.zsh` was rewritten in #12716 and is space-indented now, so this one has already landed by another route. Closing it out.

PR #10072, opened 2021-07-25
fix(avit): disable `log.showSignatures` in `_git_time_since_commit`

WeZZard:
Fixes avit's _git_time_since_commit does not work when git log was configured to show signature by default.


mcornella:
Thanks @WeZZard! I switched to using `git -c` because old git versions error'd on unrecognized `--no-show-signatures` argument.

mcornella:
Missed that the config name is `log.showSignature`, without the s. Fixed in 9c84c344d762b200de7acc794b9a0e7832144e7a.

PR #10086, opened 2021-08-04
Add KUBERNETES context to prompt for agnoster theme

alchazov:
Adds Kubernetes icon and  kubectl current context  to prompt the agnoster theme.

Setting the context name to anything containing `prod` will set the icon colour to red on black, anything containing `dev` or `stage` will set the icon colour to green on black,  otherwise it will be yellow icon on black.

Screenshot 2021-08-04 at 16 14 55

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -242,10 +242,45 @@ prompt_aws() {
   esac
 }
 
+#KUBERNETES icon:
+# - display KUBERNETES icon if context set
+# - displays red icon on black if context name contains 'prod'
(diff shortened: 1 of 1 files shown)


robbyrussell:

I'm closing this in favour of #11510, which is the same design with per-part toggles added on top and has drawn considerably more interest. Keeping both open just splits the conversation.

One thing worth carrying over either way: `prompt_kubernetes` is added to `build_prompt` unconditionally and the guard itself runs `kubectl config current-context`, so people with no Kubernetes set up still pay a process spawn on every prompt render. That's the main thing standing between this idea and agnoster, and it applies to #11510 too.

PR #10091, opened 2021-08-07
feat(archlinux): add aliases for cleaning package cache

harishnkr:
# Standards checklist:

## Changes:

- Added alias for pacman command to remove all files from cache that are not installed
- Added the same for yay and  yaourt

## Other comments:
- Did not do the same for other aur helpers but more than willing to do the same for them


ratijas:
Hi, thanks for the patch!
LGTM, although
1. Not sure we really need aur helpers for a pure pacman command (the cache it works with is entirely pacman's responsibility, right?), but maybe for consistency with other aliases it's OK.
2. we'll need to add it for the rest of helpers, of course.

As a contributor of archlinux plugin, you'd better install all the available helpers and check them.  When I was pushing my patch, I even had to dig out deprecated and already removed from everywhere Yaourt, and build it from sources :)

Haritronic909:
Thanks for replying so quickly :)
Actually I'm a complete noob and I manually entered everything via github.
I have never ever created any git repository via command line and
unfortunately I made enough mistakes that I have to reinstall everything
now. Luckily it was a virtual machine and there was nothing important to
lose.

 I'll check the working of other helpers and let you know, but since you
mentioned that we don't need the same command I'll check on that too and
give a proper response (I'm not sure if the pacman cache and the aur cache
is the same so I'm at a loss for now), although I may take some time since
I'm a student and exams are going on.

Thanks again for your reply and help.

harishnkr:
Hello there.

Sorry for giving a late reply, I know its quite unprofessional to reply after a month and a half. Things are going pretty downhill lately, and I'm just lucky to find the time for this pull request.
Anyway...

As you have mentioned in the earlier message, I tried the aliases on all the AUR helpers and it worked perfectly (even the ones that were deprecated and removed, like Yaourt :} ). All helpers work the same way in deleting cache.

Also it looks like someone kind of spam commited and definitely made a bunch of mistakes. Thay did the same thing as what I intended without adding the same alias to the other helpers. And the addition to yaourt is not even correct. 

https://github.com/ohmyzsh/ohmyzsh/commit/eba3be243f634a80ae8d8134bf88657afda71fbb

As you can see -Y is a yay specific flag, and it does not work with yaourt. I dont understand how he got the permission for the commit.

Since it kinds of makes thing difficult, I need your advice as to what to do next. Should I close this pull request and make another one, or should I add the earlier changes, and then change his additions in anotehr pull request? I dont mind using the same alias as him but its your call.

mcornella:
I've reverted that change as per your suggestion. `yaourt` could be removed so you don't need to add an alias for it (https://github.com/ohmyzsh/ohmyzsh/pull/9713, https://archlinux.fr/yaourt-en). There's also #6955 which adds some aliases for that, can we bundle them together?

harishnkr:
TL;DR is at the very bottom.

All of the changes have been made possible. However, there are a few things I could point out:

1.  https://github.com/ohmyzsh/ohmyzsh/pull/9713 has been done as mentioned, also the changes in the documentation have also been made.
2.  https://github.com/ohmyzsh/ohmyzsh/pull/6955 is also added, removing the one with pacaur, but as @ratijas mentioned earlier, I have added support for the rest of the helpers as well.
3.  Added one more option to remove the whole cache to free disk space ( -Scc flag) to pacman and all the other helpers.

Besides these changes, some other suggestions that I like to provide are:
 The option "yay -Sy" is not accepted (according to https://wiki.archlinux.org/title/AUR_helpers#Pacman_wrappers and https://github.com/Jguer/yay/commit/3bdb534).
 Also, the situation is the same for pacman and (probably, didn't check) the rest of the helpers. The { -Sy } flag has been closed in double-quotes instead of single quotes in the .zsh file, unlike the rest of the aliases, and I'm unsure as to why it is done, so I leave it to the maintainers to comment and decide.


TLDR: All the suggested changes, as well as extra changes, have been made. Other suggestions are also mentioned.

mcornella:
You didn't need to include #9713, it's good to keep a PR to a single change. Please rebase against master and only include the specific changes to this PR.

harishnkr:
Yeah, my bad. I reverted the change in #9713.

mcornella:
@ratijas can you review the current changes?

plugins/archlinux/README.md
```diff
@@ -17,8 +17,10 @@ plugins=(... archlinux)
 | pacin        | `sudo pacman -S`                       | Install packages from the repositories                           |
 | pacins       | `sudo pacman -U`                       | Install a package from a local file                              |
 | pacinsd      | `sudo pacman -S --asdeps`              | Install packages as dependencies of another package              |
+| paclean      | `sudo pacman -Sc`                      | Clean out old and unused caches and packages                     |
 | pacloc       | `pacman -Qi`                           | Display information about a package in the local database        |
 | paclocs      | `pacman -Qs`                           | Search for packages in the local database                        |
+| paclr        | `sudo pacman -Scc`                     | Remove all files from the cache                                  |
```
mcornella, line 23: `-Sc` (1 `c`) removes packages that are no longer installed
`-Scc` (2 `c`) removes all files from cache

[pacman man page](https://archlinux.org/pacman/pacman.8.html#_sync_options_apply_to_em_s_em_a_id_so_a):
> ## Sync Options (apply to -S)
> -c, --clean
> >	Remove packages that are no longer installed from the cache as well as currently unused sync databases to free up disk space. When pacman downloads packages, it saves them in a cache directory. In addition, databases are saved for every sync DB you download from and are not deleted even if they are removed from the configuration file pacman.conf(5). **Use one --clean switch to only remove packages that are no longer installed; use two to remove all files from the cache.** In both cases, you will have a yes or no option to remove packages and/or unused downloaded databases.

plugins/archlinux/README.md
```diff
@@ -52,6 +54,8 @@ upgrades were available. Use `pacman -Que` instead.
 
 | Alias   | Command                                         | Description                                                             |
 |---------|-------------------------------------------------|-------------------------------------------------------------------------|
+| auclean | `sudo aura -Sc`                                 | Clean out old and unused caches and packages                            |
+| auclr   | `sudo aura -Scc`                                | Remove all files from the cache                                         |
```
mcornella, line 58: [`aura` is fully compatible with `pacman`, they mean the same as the `pacman` aliases](https://fosskers.github.io/aura/pacman.html#clearing-your-package-cache).

plugins/archlinux/README.md
```diff
@@ -84,6 +88,8 @@ upgrades were available. Use `pacman -Que` instead.
 
 | Alias   | Command                           | Description                                                         |
 |---------|-----------------------------------|---------------------------------------------------------------------|
+| pacclean| `pacaur -Sc`                      | Clean out old and unused caches and packages                        |
+| pacclr  | `pacaur -Scc`                     | Remove all files from the cache                                     |
```
mcornella, line 92: > ## Pacman extension operations
> -S, -Ss, -Si, -Sw, -Su, -Qu, -Sc, -Scc
> > **Pacaur wraps all pacman operations** and by default extends respectively its sync, search, info, download only, system upgrade, upgrades check, clean and clean all functions to the AUR. This extension behavior is optional and can be disabled with the -r/--repo flag, and used with the variable set with the -a/--aur flag. Pacaur will also pass any other pacman operations and their related options to the pacman binary

[Same flags as `pacman`, same behavior](https://github.com/E5ten/pacaur#pacman-extension-operations).

plugins/archlinux/README.md
```diff
@@ -106,6 +112,8 @@ upgrades were available. Use `pacman -Que` instead.
 | Alias   | Command                           | Description                                                         |
 |---------|-----------------------------------|---------------------------------------------------------------------|
 | trconf  | `trizen -C`                       | Fix all configuration files with vimdiff                            |
+| trclean | `trizen -Sc`                      | Clean out old and unused caches and packages                        |
+| trclr   | `trizen -Scc`                     | Remove all files from the cache                                     |
```
mcornella, line 116: I'm not sure about these two. Documentation is harder to link to, but [here's the man source code](https://github.com/trizen/trizen/blob/fef9dec28b8d155f642c9e1a1b5b5454fb674548/TRIZEN.md#sync-options--s):
> ## SYNC OPTIONS (`-S`)
> usage: trizen {-S --sync} [options] [package(s)]
> 
> Main options:
> 
>   -c, --clean         : clean the cache directory of `trizen` and `pacman`

@harishnkr have you checked what `trizen -Scc` does?

plugins/archlinux/README.md
```diff
@@ -128,6 +136,8 @@ upgrades were available. Use `pacman -Que` instead.
 | Alias   | Command                        | Description                                                       |
 |---------|--------------------------------|-------------------------------------------------------------------|
 | yaconf  | `yay -Pg`                      | Print current configuration                                       |
+| yaclean | `yay -Sc`                      | Clean out old and unused caches and packages                      |
+| yaclr   | `yay -Scc`                     | Remove all files from the cache                                   |
```
mcornella, line 140: **Couldn't check what `-Scc` does**. Hard to link to docs, [here's a rendered man page](https://linuxcommandlibrary.com/man/yay#extended-pacman-operations) and [the source](https://github.com/Jguer/yay/blob/02d33f63c044906e014bfef7c96ed3e0c2586d76/doc/yay.8#L47-L50):
> ## EXTENDED PACMAN OPERATIONS
> -S, -Si, -Sl, -Ss, -Su, -Sc, -Qu
> > These operations are extended to support both AUR and repo packages.
> 
> -Sc
> > Yay will also clean cached AUR package and any untracked Files in the cache. Cleaning untracked files will wipe any downloaded sources or built packages but will keep already downloaded vcs sources.

plugins/archlinux/README.md
harishnkr: yes it works in the same way as the -Scc flag in pacman

plugins/archlinux/README.md
harishnkr: The documentation for the -Scc flag is given in the [arch wiki](https://wiki.archlinux.org/title/pacman)

> To remove all files from the cache, use the clean switch twice, this is the most aggressive approach and will leave nothing in the cache folder:

```
pacman -Scc
```

mcornella: But that's for pacman. Does `yay` support the same exact pacman flags?

harishnkr: yeah, I tried the same for all the other helpers including yay. If you scroll all the way to the start of this conversation @ratijas asked me specifically to try the same for all the helpers and add the same flag for all of them. Tried and tested on Yay, Trizen , Pacaur and Aura.

PR #10093, opened 2021-08-09
Add custom resource definition management aliases to kubectl plugin

3vanlock:
## Standards checklist:

## Changes:

- Add aliases for Kubernetes Custom Resource Definitions following the same get, edit, describe, delete pattern as other resources

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -168,6 +168,13 @@ alias kecj='kubectl edit cronjob'
 alias kdcj='kubectl describe cronjob'
 alias kdelcj='kubectl delete cronjob'
 
+# Custom resource definition management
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it sat so long without a reply. The names are all still free and `kgcrd` and `kdcrd` look genuinely useful.

Three things and I think we can take it:

- Drop `kgcrda`. CRDs are cluster-scoped, so `--all-namespaces` doesn't do anything there.
- Drop `kecrd` and `kdelcrd`. Editing and deleting CRDs is rare and cluster-wide, and we'd rather not put it behind six characters.
- Add the rows to `plugins/kubectl/README.md`, and rebase... a Job block landed just above where these go, so it conflicts now.

If you'd rather not pick this back up after this long, that's fine, just say so and I'll take it from here.

PR #10094, opened 2021-08-09
Add avd (Android Virtual Device) plugin

azabost:
## Standards checklist:

## Changes:

- Added avd plugin

## Other comments:

- I tested it locally in different conditions (`emulator` tool in path, not in path but available and not available at all) and it seemed to work well in all circumstances.
- I'm open to improvements and suggestions.
- Of course, there are more things a plugin for Android Virtual Devices could do so there are a lot of features that can be added here. The features I'm adding right now are simply the ones I wanted myself while developing Android apps because I found it really tedious to launch an emulator which, normally, requires copy-pasting the AVD name.
- I suggest squashing all my commits with the following commit message: "feat(avd): add avd plugin" (haven't done that myself to keep the history readable in case someone is in the middle of reading this)

### DIFF
--- a/plugins/avd/README.md
+++ b/plugins/avd/README.md
@@ -1,7 +1,8 @@
 # avd plugin
 
-The avd plugin provides aliases for Android Virtual Device / Android
-Emulator commands.
+The avd plugin provides a few functions for Android Virtual Device /
+Android Emulator commands to list all the available AVDs and launch them
+more easily.
 
 To use it, add avd to the plugins array of your zshrc file:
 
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry it sat here so long without an answer. The plugin is well put together... no hardcoded SDK paths, nothing running at shell startup.

Our test for bundling is really about how many people are asking for it, since every plugin becomes something our volunteers maintain for years. In the time this has been open it hasn't picked up any upvotes or comments, so it isn't one we'd take on. It works well as a custom plugin in `$ZSH_CUSTOM/plugins` for anyone who wants it, and that survives updates.

azabost:
Fair enough. Thanks for taking the time to review it.

PR #10100, opened 2021-08-12
feat(kubectl): add aliases for replicasets

esemeniuc:
## Standards checklist:

## Changes:

- Add `kdrs` alias for describing replicasets
- Update `kgrs` to allow grepping for `replicaset` more easily


esemeniuc:
Any status update?

carlosala:
I think both `replicaset` and `rs` are just the same thing, it's an alias, as far as I see in the `kubectl` docs. So the first alias renaming does not make so much sense, at the end it will do the same thing.

esemeniuc:
Yes, that is correct. The reason to replace line 105 is to grep the list of aliases more easily. Line 106 is the important addition, since there is no quick shortcut to describe a replicaset

esemeniuc:
Sorry for the delay, updated now!

PR #10105, opened 2021-08-14
feat(git): add most recently committed function and alias

nyoungstudios:
## Standards checklist:

## Changes:

- adds helper function to find the most recently committed branches
- adds alias `gmru` to print the 10 most recently committed branches (let me know if I should change the default number)
- adds alias `gswl` to switch to the last committed branch
- reformats the git plugin readme spacing


robbyrussell:

`plugins/git/git.plugin.zsh` has moved a lot since 2021 and this no longer merges. Beyond the conflicts, `gswl` switches to the branch with the newest commit, which is usually the branch you're already standing on, and `git switch -` already covers the "go back to where I was" case.

Issue #10110, opened 2021-08-15
Disable virtualenvwrapper_lazy.sh with environment variable


jmrohwer:
### If the feature request is for a plugin or theme, specify it here.

virtualenvwrapper plugin

### If the feature solves a problem you have, specify it here.

Currently the plugin sources `virtualenvwrapper_lazy.sh` by default if it is available. While this may lead to slightly faster loading of zsh for some people, I actually prefer the full version `virtualenvwrapper.sh`. I often switch to various Python environments and would like to be able to activate an environment with e.g.
```
workon pyd<TAB>
```
where `pyd` are the first unique letters of the environment. Then I can do so without having to hit TAB twice. With lazy loading the first TAB first actually loads all the Python environments to memory and two TAB hits are needed if `workon` is called for the first time.

While some people may prefer the lazy version, an environment variable to set this would be really great. In my experience the additional loading time of the full script is minimal and the other advantages outweigh this.

### Describe the proposed feature.

An environment variable (e.g. `VIRTUALENVWRAPPER_USE_LAZY` or similar), which when set to zero, disables the loading of the lazy version of the script but loads the full version.

### Describe alternatives you've considered

Short of forking the plugin and making my own copy, I cannot see any other options to force loading of the full script with the current version of the plugin. If there is, I am happy to be enlightened!

### Additional context

If there is any uptake for this, I am happy to submit a pull request that implements this.

### Related Issues

#6839

robbyrussell:
Hey @jmrohwer, sorry... you filed this and sent #10115 the next day, and neither got a single reply in five years. That's a bad look for us and I'm sorry for it.

Being honest about where we've come out: we're not going to add the switch. virtualenvwrapper itself has slowed right down — last PyPI release was October 2024, and upstream has had little but dependabot commits since — while most people have moved on to `venv`, `uv` or `poetry`. A new configuration knob here means carrying it for years for a shrinking number of users, and the lazy loader is the right default for everyone still on it: the full script is 1379 lines that end up spawning Python on every shell start.

If you want the non-lazy version, sourcing it yourself after Oh My Zsh loads does the job:

```zsh
source $ZSH/oh-my-zsh.sh
source /path/to/virtualenvwrapper.sh
```

Closing this and #10115 together. Thanks for the report and for the patch... both were good work, and you deserved an answer a lot sooner than this.

🤖 Co-authored with Claude Code

PR #10115, opened 2021-08-16
feat(virtualenvwrapper): disable lazy script and force full script wi…

jmrohwer:
…th environment variable

## Standards checklist:

## Changes:

- Allows the user to set an environment variable `VIRTUALENVWRAPPER_NO_LAZY` in their `.zshrc`. If set to one, this forces the loading of the full virtualenvwrapper script. If set to zero or the variable not defined, the lazy script is loaded by default.

## Other comments:

- @mcornella @lbolla  #6842 introduced loading `virtualenvwrapper_lazy.sh` by default if available. Some users (including myself) prefer the full script because the additional loading time is minimal, yet all the functionality is immediately available without first running `workon`. Further justification and explanation provided in feature request #10110.
- This PR allows configuring the behavior with a new environment variable. If this variable is not set, the behavior is exactly the same as the previous default behavior.

### DIFF
--- a/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
+++ b/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh
@@ -8,10 +8,14 @@ function {
       /usr/share/bash-completion/completions/virtualenvwrapper \
       $HOME/.local/bin/virtualenvwrapper.sh
     do
-        if [[ -f "$virtualenvwrapper" ]]; then
-            source "$virtualenvwrapper"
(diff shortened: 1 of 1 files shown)


robbyrussell:
Closing this alongside #10110, where I've written up the reasoning properly.

Short version: we're not adding the switch, mostly because virtualenvwrapper itself has gone quiet (last release October 2024) and the lazy loader is the right default for the people still using it.

Sorry this sat five years without a review. That shouldn't happen to a patch this small.

PR #10118, opened 2021-08-16
 feat(git) Add alias gswr => git switch --recurse-submodule

owngr:
## Standards checklist:

## Changes:

- added an alias alias gswr='git switch --recurse-submodules'  as git checkout --recurse-submodules already has an alias, this would help to switch branches containing submodules without having to use git checkout or deinit/init

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -167,6 +167,7 @@ plugins=(... git)
 | gsu                  | git submodule update                                                                                                             |
 | gsw                  | git switch                                                                                                                       |
 | gswc                 | git switch -c                                                                                                                    |
+| gswr                 | git switch --recurse-submodules                                                                                                  |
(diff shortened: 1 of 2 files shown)


fmalcher:
Can you please also add this new command to the readme? Thanks 😊

Olivierwenger:
@fmalcher 
Command has been added to the readme :)

esemeniuc:
Would love to see this merged!

adamchainz:
If you enable [`submodule.recurse`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse) in your Git configuration, all commands will act as if `--recurse-submodules` is provided. That sounds preferable to me over adding extra aliases that pass the option.

carlosala:
Good point Adam. Closing, modifying people's config rather than the whole shell framework makes sense.

Issue #10117, opened 2021-08-16
New alias "gmod" for git


Poyoman39:
### If the feature request is for a plugin or theme, specify it here.

Git plugin

### Describe the proposed feature.

Since we have gcd and other "develop" aliases,
And since we have gmom

What about the brand new gmod ? (i often launch this command for retrieval of develop devs on feature branches, or on master/main)

robbyrussell:
Hey @Poyoman39, sorry this sat for five years with no reply.

Being straight with you: this got answered three separate times, just never here. `gmod` came up in #7592, #11719 and #12742, and it was turned down each time for the same reason — it hardcodes one particular git flow, and the plugin's alias list is already very long. You shouldn't have had to find that out by watching other people's PRs get closed.

If it's part of your daily flow, it's a one-liner in your `.zshrc`:

```zsh
alias gmod='git merge origin/$(git_develop_branch)'
```

Closing this so it's not sitting in limbo.

🤖 Co-authored with Claude Code

PR #10119, opened 2021-08-16
New alias "gmod" for git

Poyoman39:
It Implements #10117

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -221,6 +221,7 @@ alias glp="_git_log_prettily"
 
 alias gm='git merge'
 alias gmom='git merge origin/$(git_main_branch)'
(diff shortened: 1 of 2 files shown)


plugins/git/git.plugin.zsh
```diff
@@ -221,6 +221,7 @@ alias glp="_git_log_prettily"
 
 alias gm='git merge'
 alias gmom='git merge origin/$(git_main_branch)'
+alias gmod='git merge origin/develop'
```
TonyArra, line 224: This would need to use the function `git_develop_branch` like the other aliases
```suggestion
alias gmod='git merge origin/$(git_develop_branch)'
```

Poyoman39:
Nice Thank you @TonyArra, i didn't saw this variable :D

PR #10124, opened 2021-08-19
Check for INSIDE_EMACS variable definition

pschorf:
In emacs shell mode, the existing code won't trigger
correctly due to the variable not containing term.

This change makes oh-my-zsh play nicer with emacs shell mode.

## Standards checklist:

## Changes:

- Check for INSIDE_EMACS variable definition regardless of presence of term.

## Other comments:

...

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -10,7 +10,7 @@ function title {
   emulate -L zsh
   setopt prompt_subst
 
-  [[ "$INSIDE_EMACS" == *term* ]] && return
+  [[ -v INSIDE_EMACS ]] && return
 
   # if $2 is unset use $1 as default
   # if it is set and empty, leave it as is


randomphrase:
Given that [vterm](https://github.com/akermu/emacs-libvterm) does support setting window title, perhaps the check should be ~INSIDE_EMACS != vterm~ ?

mcornella:
Can you try @randomphrase's suggestion?

randomphrase:
Hi, [this commit](https://github.com/randomphrase/ohmyzsh/commit/e46b63b6d6af876f79bc88ab043c8ba9205b4a1f) works for me. Happy to submit this as a PR if needed.

randomphrase:
Also - I would argue that this entire test is misplaced. Instead of exiting the title hook early if we are inside emacs, we should instead not add the title hook in the first place. But that's another PR :)

mcornella:
Added both of your suggestions in https://github.com/ohmyzsh/ohmyzsh/commit/54e3e8ef54765bb5d4548864ff2d6cff75a33976. Thanks for the help!

randomphrase:
Awesome - thanks!

PR #10127, opened 2021-08-20
feat(lib): add taketmp function

ekalinin:
## Standards checklist:

## Changes:

- This change adds a new function `taketmp` to create a temp directory and `cd` into it
- Besides that `take` function was also updated: add call `taketmp` if `take` was called without arguments

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -40,6 +40,10 @@ function mkcd takedir() {
   mkdir -p $@ && cd ${@:$#}
 }
 
+function taketmp() {
+  cd $(mktemp -d)
+}
+
 function takeurl() {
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry for the long wait.

The part that stops me isn't `taketmp` itself, it's that this also makes bare `take` change into a temp directory. Right now `take` with no arguments is a usage error, and turning that into a silent `cd` somewhere else is a surprise for anyone who mistypes.

Anything in `lib/` loads for every user in every shell, so I'd rather keep this one in `$ZSH_CUSTOM`, where it works exactly the same and survives updates. Closing for that reason, thanks again.

PR #10133, opened 2021-08-23
feat(pyenv): fix `pyenv init` call so it sets PATH properly

ChloeTigre:
## Standards checklist:

## Changes:

- When using pyenv without pyenv-virtualenv, make pyenv behave consistently with the other methods by passing the `--path` option to `pyenv init -`. Without this option, pyenv is initialized but does not get added to `$PATH` and this causes issues with the principle of least surprise.

### DIFF
--- a/plugins/pyenv/pyenv.plugin.zsh
+++ b/plugins/pyenv/pyenv.plugin.zsh
@@ -57,7 +57,7 @@ if [[ $FOUND_PYENV -eq 1 ]]; then
     export PYENV_ROOT="$(pyenv root)"
   fi
 
-  eval "$(pyenv init - --no-rehash zsh)"
+  eval "$(pyenv init - --no-rehash --path zsh)"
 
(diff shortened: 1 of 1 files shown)


abacigalup-eb:
This fixed it

mcornella:
This stops loading the `pyenv` function and completion, see:
```console
$ pyenv init - --no-rehash zsh
export PYENV_SHELL=zsh
source '/home/marc/.pyenv/libexec/../completions/pyenv.zsh'
pyenv() {
  local command
  command="${1:-}"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  activate|deactivate|rehash|shell)
    eval "$(pyenv "sh-$command" "$@")"
    ;;
  *)
    command pyenv "$command" "$@"
    ;;
  esac
}
$ pyenv init - --no-rehash --path zsh
export PATH="/home/marc/.pyenv/shims:${PATH}"
```
It should be made into separate calls.

PR #10134, opened 2021-08-24
fix(kubectx): show plain context if not mapped

yhrn:
## Standards checklist:

## Changes:

- If the associative array is declared but the desired key is not there the plugin will now fall back to the default value.
  Previously this situation would result in the empty string, meaning that if you want to remap some value you would have
  to remap all.

### DIFF
--- a/plugins/kubectx/kubectx.plugin.zsh
+++ b/plugins/kubectx/kubectx.plugin.zsh
@@ -3,12 +3,7 @@ typeset -A kubectx_mapping
 function kubectx_prompt_info() {
   if [ $commands[kubectl] ]; then
     local current_ctx=`kubectl config current-context`
-
-    #if associative array declared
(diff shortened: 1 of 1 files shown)


plugins/kubectx/kubectx.plugin.zsh
```diff
@@ -6,7 +6,7 @@ function kubectx_prompt_info() {
 
     #if associative array declared
     if [[ -n $kubectx_mapping ]]; then
-      echo "${kubectx_mapping[$current_ctx]}"
+      echo "${kubectx_mapping[$current_ctx]:-$current_ctx}"
     else
       echo $current_ctx
     fi
```
mcornella, line 12: This whole thing can now be simplified to
```sh
echo "${kubectx_mapping[$current_ctx]:-$current_ctx}"
```

PR #10138, opened 2021-08-25
Head branch

luisdavim:
I think this should be more robust since it doesn't rely on a fixed list of potential names.

## Standards checklist:

## Changes:

- Use `git branch -rl "*/HEAD" | rev | cut -d/ -f1 | rev` to get the `main` branch instead of trying to guess it.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -29,17 +29,10 @@ function work_in_progress() {
   fi
 }
 
-# Check if main exists and use instead of master
+# Return the head branch
 function git_main_branch() {
   command git rev-parse --git-dir &>/dev/null || return
-  local branch
-  for branch in main trunk; do
-    if command git show-ref -q --verify refs/heads/$branch; then
-      echo $branch
-      return
-    fi
-  done
-  echo master
+  git branch -rl "*/HEAD" | rev | cut -d/ -f1 | rev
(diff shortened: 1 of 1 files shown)


plugins/git/git.plugin.zsh
```diff
@@ -29,17 +29,10 @@ function work_in_progress() {
...
   command git rev-parse --git-dir &>/dev/null || return
-  local branch
-  for branch in main trunk; do
-    if command git show-ref -q --verify refs/heads/$branch; then
-      echo $branch
-      return
-    fi
-  done
-  echo master
+  git branch -rl "*/HEAD" | rev | cut -d/ -f1 | rev
```
luisdavim, line 35: `git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@'` should also work if you prefer to use `sed` instead of `rev` and `cut` but all of these are pretty standard tools that should be available in any system running `git` on `zsh`....

mcornella:
See https://github.com/ohmyzsh/ohmyzsh/pull/9879#issuecomment-896087681:

> I appreciate what you're trying to do, but we don't need to support all different configurations. The intent behind `git_main_branch` was to allow the transition from master to another branch name, and it's done that for the most common name alternatives. It doesn't need to do more.

PR #10154, opened 2021-09-01
Fix zero-length character that messes prompt when tab completion

thalesbarretto01:
Those two special characters "arrow" mess with zsh prompt when using tab completion.
The solution is to escape them properly.
This is and old issue, that has been reported and solved here: https://stackoverflow.com/a/58196769

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


mcornella:
This should be automatically handled by zsh if you have the proper settings. What's your `$ZSH_VERSION` and the output of `locale` and `locale -a`?

mcornella:
The `C` locale is the simplest locale and it doesn't support UTF-8 characters, you need to set it to the `en_US.utf8` one so that zsh recognizes the arrow symbol correctly. See [more for details here](https://github.com/ohmyzsh/ohmyzsh/wiki/FAQ#i-see-duplicate-typed-characters-after-i-complete-a-command).

PR #10158, opened 2021-09-02
feat(git): guess main branch name also from remotes

mvitaly:
## Standards checklist:

## Changes:

- When guessing the git main branch name consider remote branches from origin or upstream remotes that weren't checked out locally, this allows shortcuts like `gcm`, `glum`, and `gmom` to work even before the main branch is checked out locally.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -32,12 +32,15 @@ function work_in_progress() {
 # Check if main exists and use instead of master
 function git_main_branch() {
   command git rev-parse --git-dir &>/dev/null || return
+  local ref_prefix
   local branch
-  for branch in main trunk; do
-    if command git show-ref -q --verify refs/heads/$branch; then
-      echo $branch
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks for your contribution!

Cyberbeni:
Why guess the name of the default branch instead of actually getting it?
I always use origin, so I use this in my script:
```zsh
local defaultBranch=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
if [ -z "$defaultBranch" ]; then
	defaultBranch=$(git ls-remote --symref "$remoteUrl" HEAD | grep "^ref: " | sed 's@^ref: refs/heads/\([^[:space:]]*\).*@\1@')
fi
```

mcornella:
See https://github.com/ohmyzsh/ohmyzsh/pull/10138 https://github.com/ohmyzsh/ohmyzsh/pull/9879

PR #10164, opened 2021-09-07
Nwidart Module Commands

iamngoni:
## Standards checklist:

## Changes:

- created aliases for nwidart modules commands

## Other comments:


robbyrussell:
Thanks for this, and sorry for the very long silence.

I'm interested, but 28 aliases for one package is more than I want to add to the laravel plugin in one go. Could you cut it down to the handful you actually reach for daily, and name them to match the existing `pa*` convention in that file?

One thing to fix while you're in there: `pmseed` is defined twice, so one of them gets lost.

If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll close it out.

PR #10169, opened 2021-09-08
feat(git)!: ask for confirmation in `gpf` alias

### LINKED ISSUE #10168: About the dangerous gpf
### Describe the bug

I want to share this "user story". On french and english keyboards, the "g" and the "f" are closed letters. and there is a non null risk that when typing "gp" your finger also press the "f" key (happened to me more than one time).

This may lead to git push force on a branch you didn't want.

Git push force is that kind of command you can't revert and which is really critical. What do you guys think about making this alias "harder to launch" or about removing it ? Or make a gpf with confirmation ?

### Steps to reproduce

Type quickly "gp" on your keyboard :)

### Expected behavior

Safe gp

_No response_

### OS / Distro

Any

### Zsh version

last

### Terminal emulator

Terminal

_No response_

### Additional context

_No response_

mcornella:
## Standards checklist:

## Changes:

- Turn `gpf` alias into a function that shows a yes/no prompt
- Rename `git push --force-with-lease` to `gpff`

Fixes #10168

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -89,7 +89,7 @@ plugins=(... git)
 | `gdnolock`             | `git diff $@ ":(exclude)package-lock.json" ":(exclude)\*.lock"`                                                                 |
 | `gdt`                  | `git diff-tree --no-commit-id --name-only -r`                                                                                   |
 | `gf`                   | `git fetch`                                                                                                                     |
-| `gfa`                  | `git fetch --all --tags --prune`                                                                                                       |
+| `gfa`                  | `git fetch --all --tags --prune`                                                                                                |
 | `gfo`                  | `git fetch origin`                                                                                                              |
 | `gg`                   | `git gui citool`                                                                                                                |
 | `gga`                  | `git gui citool --amend`                                                                                                        |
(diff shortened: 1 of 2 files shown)


mcornella:
NOTE: `gpf!!` is not possible to run since `!!` expands to the last command:
Captura de pantalla de 2021-09-08 15-08-56

mcornella:
> NOTE: `gpf!!` is not possible to run since `!!` expands to the last command:
> Captura de pantalla de 2021-09-08 15-08-56

This is now fixed by making these changes:
- `gpf!` is kept to be `git push --force`.
- The previous `gpf` alias is renamed to `gpff`.

rwmitchell:
Is an alias for force really necessary?  Shouldn’t the user slow down and think before doing this?

mcornella:
I think having to type the exclamation is enough to slow down the user. This has been in the git plugin for years and it doesn't seem to cause any issues.

TonyArra:
these changes work as expected for me - one stupid question: after installing the plugin from this branch, I had to `unalias gpf` in my session before the `gpf()` function took over. Is that expected behavior?

mcornella:
How did you reload zsh? `source ~/.zshrc` is not the way to do it, you should run `exec zsh`.

TonyArra:
@mcornella ah my bad - works with `exec zsh`

VenomVendor:
@mcornella With rebase being part of daily git ceremony **at least** 50 times a day, confirming every time is annoying. I can override in my zsh config. But, the whole point of these alias is to avoid redundancy, save time/strokes.

`gpf` is safer than `gpf!`. Is there an use case where `gpf` can overwrite remote?

I believe this must be reverted.

mcornella:
I added `gpff` for the old behavior, which is less difficult to type than `gpf!`. Does this work for you? You're not the only person to complain so if it doesn't I will just put the safety aliases behind an opt-in setting.

VenomVendor:
`gpff/gpf` is different from `gpf!`. No changes in `gpf!` alias. 

Based on #10168, this confirmation should have been added to ~~`gpf`~~ `gpf!`, since `git push --force` is dangerous as it overwrites remote. But, `git push --force-with-lease` doesn't. It **DOES REJECT** if there are any changes to remote.

I'd still prefer `gpf`; if necessary, this can be added as opt-in.

mcornella:
Thanks fpr the feedback, I've reverted the commit for now.

PR #10170, opened 2021-09-08
fix(cp): add unique suffix to rsync backup directory for each user

XuehaiPan:
## Standards checklist:

## Changes:

Add a unique suffix (the user name) to `rsync` backup directory for each user.

## Other comments:

If the path to the backup directory is a fixed location, this may cause permission issues on a multi-user operating system (e.g., on a Linux server).

Suppose USERALICE uses the command `cpv` and that creates the backup folder `/tmp/rsync`. Then this backup folder is owned by USERALICE and anyone else on the system does not have permission to write this folder (to use `cpv` command).

### DIFF
--- a/plugins/cp/cp.plugin.zsh
+++ b/plugins/cp/cp.plugin.zsh
@@ -1,4 +1,4 @@
 cpv() {
-    rsync -pogbr -hhh --backup-dir="/tmp/rsync-${USER:-"$(whoami)"}" -e /dev/null --progress "$@"
+    rsync -pogbr -hhh --backup-dir="/tmp/rsync-${USERNAME}" -e /dev/null --progress "$@"
 }
 compdef _files cpv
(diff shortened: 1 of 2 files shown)


plugins/cp/cp.plugin.zsh
```diff
@@ -1,4 +1,4 @@
 cpv() {
-    rsync -pogbr -hhh --backup-dir=/tmp/rsync -e /dev/null --progress "$@"
+    rsync -pogbr -hhh --backup-dir="/tmp/rsync-${USER:-"$(whoami)"}" -e /dev/null --progress "$@"
```
mcornella, line 2: You can use `$USERNAME` which [is always guaranteed to be defined inside a Zsh session](https://zsh.sourceforge.io/Doc/Release/Parameters.html#index-USERNAME).

PR #10184, opened 2021-09-12
feat(archlinux): add cleanup aliases to `yaourt` and `yay`

mralaminahamed:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -170,9 +170,9 @@ if (( $+commands[yaourt] )); then
   alias yainsd='yaourt -S --asdeps'
   alias yamir='yaourt -Syy'
   alias yaupd="yaourt -Sy"
-  alias upgrade='yaourt -Syu'
   alias yacun='yaourt -Yc'
   alias yacf='yaourt -Scc'
+  alias upgrade='yaourt -Syu'
 fi
 
 if (( $+commands[yay] )); then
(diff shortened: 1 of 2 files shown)


mralaminahamed:
Hi friend, I miss two option of yay and yourt  for clean cache. I want to add archlinux package for everyone to use it.

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -171,6 +171,8 @@ if (( $+commands[yaourt] )); then
   alias yamir='yaourt -Syy'
   alias yaupd="yaourt -Sy"
   alias upgrade='yaourt -Syu'
+  alias yacun='yay -Yc'
+  alias yacf='yay -Scc'
```
mcornella, line 175: These two use `yay` even though they're in the `yaourt` block.

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -191,5 +193,7 @@ if (( $+commands[yay] )); then
   alias yamir='yay -Syy'
   alias yaupd="yay -Sy"
   alias upgrade='yay -Syu'
+  alias yacun='yay -Yc'
+  alias yacf='yay -Scc'
```
mcornella, line 197: You should document these in the plugin README.

mcornella (changes requested):

mcornella:
I assume you've made these changes via the GitHub website. To make more changes, you can [go to the `patch-1` branch](https://github.com/mralaminahamed/ohmyzsh/tree/patch-1) and edit the files there, then make a commit in the same branch.

Also use a more descriptive Pull Request title, it's not obvious what these aliases do. Thanks!

plugins/archlinux/archlinux.plugin.zsh
mralaminahamed: ok, i do this.

mcornella (approved):

Issue #10196, opened 2021-09-16
[Fluxcd Plugin] Create a FluxCD plugin for the autocompetion


arthtux:
### If the feature request is for a plugin or theme, specify it here.

Fluxdc plugins

### Describe the proposed feature.

The plugin make the autocompletion with the command Flux
Documentation is here -> https://fluxcd.io/docs/cmd/flux_completion_zsh/

mcornella:
If you have a PR ready why do you open an issue? You're not the first to do it, I'm just curious why people do that? Should I update some docs?

PR #10197, opened 2021-09-16
feat(flux): Create the plugin for FluxCD

arthtux:
## Standards checklist:

Create a new plugin for the Flux-cli (#10196)

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/flux/README.md
+++ b/plugins/flux/README.md
@@ -1,0 +1,9 @@
+# Flux plugin
+
+This plugin adds completion for the [Flux-cli](https://fluxcd.io/docs/cmd/),
+
+To use it, add `flux` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... flux)
+```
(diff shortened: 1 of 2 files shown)


serrovsky:
This PR seems interesting.... Why it's still open?

carlosala:
I can see [here](https://fluxcd.io/flux/cmd/flux_completion_zsh/) that this guys have some generation of completions in the CLI. Could you go something similar to what's done in [gh plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/gh/gh.plugin.zsh) to load the completion?
The code would be cleaner and also easier to maintain, as we wouldn't need to update the plugin each time they change the completion.
Thanks!

serrovsky:
@carlosala Thanks for your answer. 

Following your suggestion, I created a new PR with a cleaner approach
https://github.com/ohmyzsh/ohmyzsh/pull/11350

And since I notice that atlas cli is missing as well >
https://github.com/ohmyzsh/ohmyzsh/pull/11349

PR #10209, opened 2021-09-22
refactor(vim-interaction): clean up code and open gvim instance if none open

mcornella:
## Standards checklist:

## Changes:

- Simplified `$before` and `$after` command building.
- Open a gvim instance if none found.
- Clean up syntax to be more idiomatic zsh.


LunarWatcher:
Works fine for me - nothing is breaking substantially anyway (but I haven't been using this plugin that much before, so I might be missing something). All normal operations work fine at least

mcornella:
Great, thanks!

PR #10232, opened 2021-09-28
feat(lol): new function `qrcode`

tinogomes:
## Standards checklist:

## Changes:

- `qrcode` generates a QRCode image. Requires internet access

### DIFF
--- a/plugins/lol/lol.plugin.zsh
+++ b/plugins/lol/lol.plugin.zsh
@@ -49,3 +49,14 @@ alias yolo='git commit -m "$(curl -s http://whatthecommit.com/index.txt)"'
 alias letcat='git checkout'
 alias violenz='git rebase'
 
+qrcode() {
+  url=$1
+
+  if [[ "$url" != http* ]]; then
+    url=https://$1
+    echo
+    echo QRCode for: $url
+  fi
+
+  curl qrcode.show/$url
+}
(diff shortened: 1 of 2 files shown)


plugins/lol/lol.plugin.zsh
```diff
@@ -49,3 +49,14 @@ alias yolo='git commit -m "$(curl -s http://whatthecommit.com/index.txt)"'
...
+qrcode() {
+  url=$1
+
+  if [[ "$url" != http* ]]; then
+    url=https://$1
+    echo
+    echo QRCode for: $url
+  fi
+
+  curl qrcode.show/$url
```
JuniorJPDJ, line 61:
```suggestion
  curl qrcode.show -d "$url"
```

JuniorJPDJ:
You sometimes want to qrcode something which is not url, I'd not do protocol logic here.

mcornella:
Does this really fit with the intent of the `lol` plugin? Maybe this would go better as its own plugin, with the functions defined in the [SHELL FUNCTIONS section](https://qrcode.show/).

JuniorJPDJ:
I also suggest writing to qrcode.show admins about adding support to ohmyzsh. They could probably add info to their main page.

tinogomes:
> I also suggest writing to qrcode.show admins about adding support to ohmyzsh. They could probably add info to their main page.

I will try to do that too.

tinogomes:
extract to a new plugin #10239

Issue #10187, opened 2021-09-13
Update reminder instead of Update prompt


jjspace:
### If the feature solves a problem you have, specify it here.

I usually have the first command I want to run in my head when I open a terminal window. That means I usually start typing it before even looking too hard at the window. The problem is that when omz decides to prompt me to update it eats the first letter of said command meaning I 1) have to run it twice and 2) missed the chance to respond `Y` to update omz.

![2021-09-13_11-39](https://user-images.githubusercontent.com/8007967/133114557-2daee1da-8e40-4a7c-8cff-b2c450081b16.png)

### Describe the proposed feature.

I would love a way to set up omz so that it just gives me a reminder instead of a full prompt to update (maybe even in color instead of plain text but that's above and beyond). Something like:

```
[oh-my-zsh] It's time to update! You can do this by running `omz update`
```

Then when I see this I can run the command on my own time and it doesn't disrupt my flow. 

### Describe alternatives you've considered

I've read about the options under [Getting updates](https://github.com/ohmyzsh/ohmyzsh#getting-updates) but don't find them helpful for my problem. Just setting `DISABLE_UPDATE_PROMPT=true` would mean I don't get the prompt but it updates automatically and I would prefer it's an intentional thing I do (so I know when it happens and I can read the changes). If I also set `DISABLE_AUTO_UPDATE=true` then it becomes manual thing again but I no longer have a reminder to do it periodically. I'm a forgetful person with things like this so I can all but guarantee I would forget to do this for months which is not what I want.

PR #10244, opened 2021-10-01
Add option to have update reminder instead of update prompt

NoPreserveRoot:
## Standards checklist:

## Changes:

- Adds an option to have an update reminder rather than an update prompt upon startup
- Addresses #10187 

## Other comments:
- **I haven't been able to test it yet** because I haven't received any OMZ updates. I would hold off on merging until I or someone else can properly test it.
- My code technically doesn't follow the coding style because line 82 of `check_for_upgrade.sh` is about 100 characters long. I couldn't find a way to reasonably split the line. This may not actually be a violation of the style guidelines, but I thought it's better to be safe than sorry.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -78,17 +78,21 @@ function update_ohmyzsh() {
   if [[ "$DISABLE_UPDATE_PROMPT" = true ]]; then
     update_ohmyzsh
   else
-    # input sink to swallow all characters typed before the prompt
-    # and add a newline if there wasn't one after characters typed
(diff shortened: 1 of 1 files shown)


mcornella:
Hi there, thanks for your contribution. I've already done work to address this on https://github.com/mcornella/ohmyzsh/commits/feat/improved-auto-update (https://github.com/mcornella/ohmyzsh/commit/c80d795e8976165c1193b444fdb956ffe77dd911).

I can add you as co-author but we're moving on a different direction to simplify OMZ settings.

PR #10245, opened 2021-10-01
feat(prompt): add git short-hash option to git_prompt_info via env var

hossam-magdy:
Utilizing env:`ZSH_THEME_GIT_SHOW_COMMITHASH`, which potentially could greatly help developers, simply from the terminal prompt, to recover latest local HEAD commit after altering or removing commands (like: `commit --ammend` or `rebase`).

It is using the same existing strategy of env:`ZSH_THEME_GIT_SHOW_UPSTREAM`.

---

Example with `jonathan` theme (while the `cwd` is a git repo)

**Before** 

Screenshot 2021-10-10 at 20 14 47

**After**

Screenshot 2021-10-10 at 20 15 07

---

## Standards checklist:

## Changes:

- Updating the `git_prompt_info` function adding short-hash of the HEAD commit, if the env var `ZSH_THEME_GIT_SHOW_COMMITHASH` is set to 1

## Other comments:

N/A

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -29,7 +29,13 @@ function git_prompt_info() {
     && upstream=" -> ${upstream}"
   fi
 
-  echo "${ZSH_THEME_GIT_PROMPT_PREFIX}${ref}${upstream}$(parse_git_dirty)${ZSH_THEME_GIT_PROMPT_SUFFIX}"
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry for the delay.

This is the same feature as your earlier #9530, and that's where the conversation is, so I'm closing this one to keep it in a single place. `lib/git.zsh` has also moved since: `git_prompt_info` was renamed in #12257, so this no longer applies cleanly.

PR #10247, opened 2021-10-01
feat(qrcode): new plugin to generate qrcode via https://qrcode.show

tinogomes:
## Standards checklist:

## Changes:

- adds commands `qrcode` and `qrsvg`

## Other comments:

...

### DIFF
--- a/plugins/qrcode/qrcode.plugin.zsh
+++ b/plugins/qrcode/qrcode.plugin.zsh
@@ -1,0 +1,17 @@
+# Imported and improved from https://qrcode.show/, section SHELL FUNCTIONS
+
+_qrcode_show_message() {
+  echo "Type or paste your text, add a new blank line, and press ^d"
+}
+
+qrcode () {
+  local input="$*"
+  [ -z "$input" ] && _qrcode_show_message && local input="@/dev/stdin"
+  curl -d "$input" https://qrcode.show
+}
+
+qrsvg () {
+  local input="$*"
+  [ -z "$input" ] && _qrcode_show_message && local input="@/dev/stdin"
+  curl -d "${input}" https://qrcode.show -H "Accept: image/svg+xml"
+}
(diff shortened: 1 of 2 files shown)


plugins/qrcode/qrcode.plugin.zsh
```diff
@@ -0,0 +1,17 @@
...
+qrcode () {
+  local input="$*"
+  [ -z "$input" ] && _qrcode_show_message && local input="@/dev/stdin"
+  curl -d "$input" https://qrcode.show
+}
+
+qrsvg () {
+  local input="$*"
+  [ -z "$input" ] && _qrcode_show_message && local input="@/dev/stdin"
+  curl -d "${input}" https://qrcode.show -H "Accept: image/svg+xml"
```
JuniorJPDJ, line 16:
```suggestion
  curl -d "$input" https://qrcode.show -H "Accept: image/svg+xml"
```

robbyrussell:
@tinogomes thanks for the contribution!

madonius:
@tinogomes I would propose switching to a local qr-code generator. What do you think?

carlosala:
The cool thing about qrcode.show is that it does not require any installation and it has no tracking, it is OSS. I'm on the fence with this one.

madonius:
For the tracking, it's a trust me bro guarantee. The users would have to not only trust the developer but all parties involved. 

The biggest benefit, of the local solution, however, is that the user could create a qr code without the need for an internet connection. 

About qrencode, I would have proposed it, but the issue is that  it seems to be unmaintained (?). The qrrs tool seems promising though.

carlosala:
> For the tracking, it's a trust me bro guarantee. The users would have to not only trust the developer but all parties involved.

That's indeed true.

> The biggest benefit, of the local solution, however, is that the user could create a qr code without the need for an internet connection.

IMO, that would be a great benefit. We can check for `qrencode` (which seems to be the most common among all tools) and fallback to qrcode.show

madonius:
@carlosala As I wrote above qrencode seems to not have gotten tlc for a while now. Last comit I see was 4 years ago.
Ofcourse it _could_ be that there is no need for anything to be fixed.
I'd be happy to supply the patch with whatever tool we decide upon.

Issue #10241, opened 2021-09-30
[docker-compose plugin] Docker Compose 2.0 compatibility breakage


vijuarez:
### Describe the bug

With the recent release of [Docker Compose 2.0](https://github.com/docker/compose) the plugin no longer works as intended, with the most obvious change being commands changing from `docker-compose *` to `docker compose *`. 

### Steps to reproduce

1. Enable plugin in a system with the latest Docker Compose.
2. Run any command shortcut.
3. Watch it complain with `zsh: command not found: docker-compose`

### Expected behavior

The plugin should use the correct commands for the currently installed Docker Compose version. For now, I believe that a warning on the documentation about the plugin only being valid for Docker Compose v1 should be enough. Requiring [compose-switch](https://github.com/docker/compose-switch) would also solve the problem, but I personally had some trouble with it.

### OS / Distro

Arch Linux

### Zsh version

5.8

### Terminal emulator

Konsole

Frederick888:
Just noticed that starting from `docker-compose 2.0.1`, a symbolic link `usr/bin/docker-compose -> /usr/lib/docker/cli-plugins/docker-compose` is provided in Arch package. Not sure if official packages / other distros are going to follow suit.

Joniator:
There is a [compose-switch ](https://docs.docker.com/compose/cli-command/#compose-switch) mentioned in the official documentation, which should behave similiar to the way Arch does it. 

There was also some discussion in PR #10231 regarding an update to the plugin

mcornella:
Is there a way to test if `docker compose` is available?

mcornella:
Opened #10248 to support the new version, test it out with `omz pr test 10248`. Thanks!

vijuarez:
@mcornella I can confirm that the PR works for both Compose V2 vanilla and the Arch workaround that @Frederick888 mentioned.

mcornella:
Thanks!

PR #10248, opened 2021-10-01
fix(docker-compose): support Compose V2 `docker compose` command

mcornella:
## Standards checklist:

## Changes:

- Tests if `docker compose` runs successfully and uses that to decide whether to use `docker-compose` or `docker compose` in the aliases.

## Other comments:

Fixes #10231 #10241

### DIFF
--- a/plugins/docker-compose/docker-compose.plugin.zsh
+++ b/plugins/docker-compose/docker-compose.plugin.zsh
@@ -1,28 +1,22 @@
-# Authors:
-# https://github.com/tristola
-#
-# Docker-compose related zsh aliases
+# support Compose v2 as docker CLI plugin
+command docker compose &>/dev/null && dccmd='docker compose' || dccmd='docker-compose'
 
-# Aliases ###################################################################
+alias dco="$dccmd"
+alias dcb="$dccmd build"
+alias dce="$dccmd exec"
+alias dcps="$dccmd ps"
+alias dcrestart="$dccmd restart"
+alias dcrm="$dccmd rm"
+alias dcr="$dccmd run"
+alias dcstop="$dccmd stop"
+alias dcup="$dccmd up"
+alias dcupb="$dccmd up --build"
+alias dcupd="$dccmd up -d"
+alias dcdn="$dccmd down"
+alias dcl="$dccmd logs"
+alias dclf="$dccmd logs -f"
+alias dcpull="$dccmd pull"
+alias dcstart="$dccmd start"
+alias dck="$dccmd kill"
 
-# Use dco as alias for docker-compose, since dc on *nix is 'dc - an arbitrary precision calculator'
-# https://www.gnu.org/software/bc/manual/dc-1.05/html_mono/dc.html
-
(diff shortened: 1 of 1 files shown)


Frederick888:
Sorry I know I should've reported this earlier, but running `docker compose` to check if it's available can block the initialisation for a long time in case of using a remote Docker context.

I've configured a context to connect to a remote Docker host via SSH, and `docker compose` command blocks for ~20s if the Docker host is down. In fact even `docker --help` / `docker --version` block. If there's no good way to work around this, I'm wondering if it's easier to simply allow overriding the auto-detection via an environment variable?

mcornella:
I consider that a configuration error. I don't know how to do that, but [according to this page](https://blog.mikesir87.io/2019/08/using-ssh-connections-in-docker-contexts/) you could set up the context via an env variable after OMZ has loaded:
```sh
# OR use the DOCKER_CONTEXT env var
DOCKER_CONTEXT=ssh-box docker ps
```

Either way I don't see why a simple `docker compose` command should wait for an external host before displaying the help message 😐

Frederick888:
@mcaserta Just guessing, perhaps when there are new subcommands, it needs to check if it's supported by remote Docker host and then decides whether to show the subcommand or not? (But yeah, even this is pretty weird, printing an error message only when such subcommands get called is much simpler.) Anyway, I think this is a bit off-topic here.

And thanks for the info about `DOCKER_CONTEXT`. Instead of setting it after omz (it'll be tricky to apply this environment variable to non-interactive scripts if we do this), I can try applying it to omz initialisation only. I'll give it a shot tomorrow on my work machine. But tbh I don't think this is a very clean solution.

mcornella:
Oh, we could do that in the plugin. Does this work?
```
DOCKER_CONTEXT= docker compose
```

Frederick888:
@mcornella It should be `DOCKER_CONTEXT=default docker compose` I think. I used `docker context use ssh-box` make sure the context was applied globally, so here it needs to switch back to the default one temporarily. Not sure if the default context can be deleted or not though...

Frederick888:
@mcornella Something like
```sh
[[ -n "$DOCKER_CONTEXT" ]] && _DOCKER_CONTEXT="$DOCKER_CONTEXT" || _DOCKER_CONTEXT="default"
/usr/bin/env DOCKER_CONTEXT="$_DOCKER_CONTEXT" docker compose
```
?

mcornella:
You don't need all that, when you set the env variable in the same line you call the command it only applies to that command. So as you said this should work
```sh
DOCKER_CONTEXT=default docker compose
```
Let me know when you test it if it stops taking so long.

Frederick888:
@mcornella Yup that works! And I can confirm that the default context cannot be deleted (tested on mac):
```sh
$ docker context rm default
Error: context "default": forbidden
```

mcornella:
OK should be fixed in 49bc55f966d91b2a407c883cf1d0f21d0cdd9f4a. Try `omz update`. Thanks for the help!

PR #10249, opened 2021-10-02
feat(kubectl): add aliases for HPA management

### LINKED ISSUE #10029: Add kubectl get horizonal pod autoscaler alias to k8s plugin
kubectl plugin.

Please consider adding something like `kghpa` to perform a `kubectl get hpa` command.

Many thanks for this plugin, really like it!

jaydesl:
## Standards checklist:

## Changes:

- Adds a full set of aliases to manage HorizontalPodAutoscaler objects with kubectl

## Other comments:

- Fixes #10029

### DIFF
--- a/plugins/kubectl/README.md
+++ b/plugins/kubectl/README.md
@@ -120,6 +120,13 @@ plugins=(... kubectl)
 | kecj    | `kubectl edit cronjob`              | Edit CronJob from the default editor                                                             |
 | kdcj    | `kubectl describe cronjob`          | Describe a CronJob in details                                                                    |
 | kdelcj  | `kubectl delete cronjob`            | Delete the CronJob                                                                               |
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry it sat here for four years without a real answer. This is the best-built of the HPA attempts we've had... it follows the plugin's naming exactly, and I couldn't find a collision anywhere in the repo.

We're going to hold the line on the alias count though. The plugin already ships around 120 aliases, and HPA is something you reach for during capacity tuning or an incident rather than every day. I closed the original request (#10029) for the same reason earlier today, so it's the whole idea we're declining, not your version of it.

These work fine in your own `.zshrc` after the `source $ZSH/oh-my-zsh.sh` line. Thanks again for taking the time.

PR #10270, opened 2021-10-06
feat(cargo): merging rustup in cargo plugin

carlosala:
## Standards checklist:

## Changes:

As I said in #10087, I think that having two plugins for `rustup` and `cargo` instead of having just one is a bit redundant, because generating `cargo` completions needs `rustup`.
I merged both in `cargo`, which is much more daily used by users.

It is just an idea, but I think that would be very interesting to merge it on master!
Thanks!


mcornella:
Blocked by #10392.

mcornella:
Moved to the `rust` plugin.

PR #10271, opened 2021-10-06
fix(pip): use `pip3` in `pip3 uninstall` completion

### LINKED ISSUE #8977: pip plugin does't work for pip3 from Homebrew
**Describe the bug**
Let's suppose that you are using `pip3` and that there is no `pip` command:

```
$ which pip3
/usr/local/bin/pip3
$ pip
zsh: command not found: pip
```

In this case, the `pip` Oh-My-ZSH plugin will:

- Add a new alias that doesn't make sense here:
    - `alias pip="noglob pip"`
- Add a new PIP autocomplete to `pip3`:
    - `#compdef pip pip2 pip-2.7 pip3 pip-3.2 pip-3.3 pip-3.4`
- But the autocomplete will try to get the package list from `pip` instead of `pip3`:
    - ``installed_pkgs=(`pip freeze | cut -d '=' -f 1`)``
- And will try to get the subcommand list from `pip` instead of `pip3`:
    - `_describe -t commands "pip subcommand" _1st_arguments`

Because of this, you the autocomplete for `pip3` will not work.

There is another edge case which is when the user has both `pip` (symlink to `pip2`) and `pip3` installed, and they share the same cached list of installed packages, which is wrong (both of them have different installed packages).

**To Reproduce**

Bug 1:

1. Install python3 from Homebrew: `brew install python`
    - The `pip3` command will be installed at `/usr/local/bin/pip3` at this time
2. Put `/usr/local/bin` in your PATH
3. Enable the `pip` plugin
4. Run `pip3 ins<TAB><TAB>`
5. The autocomplete will work here
6. Run `pip3 uninstall <TAB><TAB>`
7. The autocomplete will not work and show the error: `"zsh: command not found: pip"`

Bug 2: (Didn't test, but this will very likely happen) 

1. Install python3 from Homebrew: `brew install python`
    - The `pip3` command will be installed at `/usr/local/bin/pip3` at this time
2. Install python2 and pip (for Python 2) from somewhere else: `/usr/bin/pip` or `/usr/bin/local/pip` will exist now
2. Put `/usr/local/bin` in your PATH
3. Enable the `pip` plugin
4. Run `pip install boto3`
5. Run `pip uninstall <TAB><TAB>`
6. Run `pip3 uninstall <TAB><TAB>`
7. The autocomplete will work, but it will show the packages for `pip` (not for `pip3`), including `boto3`.
8. Run `pip3 uninstall boto<TAB><TAB>` (will expand to `pip3 uninstall boto3`)
9. An error will occur: `WARNING: Skipping boto3 as it is not installed.`

**Expected behavior**

The autocomplete should work and provide the correct list of installed packages.

 - OS / Distro: macOS
 - Latest ohmyzsh Update?: Yes
 - ZSH Version: zsh 5.3 (x86_64-apple-darwin18.0)
 - Terminal emulator: iTerm2

**Additional context**

Maybe a possible solution could be:

- Depending on the called command (`pip` vs `pip2` vs `pip3` vs. `pip2.6` etc), change the lines to `installed_pkgs=(`$PIP_CMD freeze | cut -d '=' -f 1`)` and `_describe -t commands "$PIP_CMD subcommand" _1st_arguments`
    - `PIP_CMD` here means, the correct command that I'm trying to autocomplete
    - If I call pip2 or pip and if this is a symlink to pip2.6, the command pip2.6 should be used here
    - `$ZSH_PIP_CACHE_FILE` should change depending on the pip command used (each pip has a different list of installed packages)

But I have no idea how to do something like this.

Another solution could be: define `ZSH_PIP3_CACHE_FILE`, move as much of code to `pip.plugin.zsh` as possible, and define both `_pip` (autocomplete to pip2 family of commands) and `_pip3` (autocomplete to pip2 family of commands).

In both cases, I suspect that there would still be the edge case of the client loading a virtual env and running `pip uninstall <TAB><TAB>` (that would use the wrong cache file). A possible workaround would be ignoring the cache file when running the autocomplete inside a virtualenv.

osule:
## Standards checklist:

## Changes:

If only `pip3` is present in the shell environment,
pip uninstall tab returns an error like:

    _pip_installed:2: command not found: pip

This change makes it possible to alias `pip` to `pip3` and keep the completions
working as intended.

Whatever command that the `pip` command aliases is used for generating
the uninstall completions.

Fixes #8977

## Other comments:

n.a

### DIFF
--- a/plugins/pip/_pip
+++ b/plugins/pip/_pip
@@ -13,7 +13,7 @@ _pip_all() {
 }
 
 _pip_installed() {
-  installed_pkgs=(`pip freeze | cut -d '=' -f 1`)
+  installed_pkgs=(`$aliases[pip] freeze | cut -d '=' -f 1`)
 }
 
 local -a _1st_arguments
--- a/plugins/pip/_pip
+++ b/plugins/pip/_pip
@@ -13,7 +13,7 @@ _pip_all() {
 }
 
 _pip_installed() {
-  installed_pkgs=(`$aliases[pip] freeze | cut -d '=' -f 1`)
+  installed_pkgs=($($service freeze | cut -d '=' -f 1))
 }
 
 local -a _1st_arguments


mcornella:
You can use `$service` instead to know which pip command was completed.

PR #10277, opened 2021-10-09
feat(git): Update master & develop branch determination functions to be configurable

maverick1872:
This change became necessary for one of my workflows, as 'staging' is used as our default "develop" branch. This looks in the local git configuration for a defined override, then validates a ref exists for the defined branch name. If either of those checks fails; it then falls back to traditional branch names. 

Subsequently I removed the assumption that "master" or "develop" would be the appropriate fallback. So if no match can be made in the respective repo, I inform the user of the override they can leverage via stderr.

## Standards checklist:

## Changes:

- feat(git) - update branch helper functions

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -29,30 +29,46 @@ function work_in_progress() {
   fi
 }
 
-# Check if main exists and use instead of master
+# Attempt to determine main branch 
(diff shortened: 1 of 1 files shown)


Maverick1872:
@larson-carter @mcornella Any updates on whether or not there is any interest in accepting this proposed change?

PR #10292, opened 2021-10-12
Updated plugin shortcuts for Laravel Plug in as well as the Docs

lewis-munyi:
## Standards checklist:

## Changes:

- Added missing `artisan` Commands
- Updated doc to reflect this. 

## Other comments:
@zeshan77 
...

### DIFF
--- a/plugins/laravel/README.md
+++ b/plugins/laravel/README.md
@@ -6,52 +6,64 @@ This plugin adds aliases and autocompletion for Laravel [Artisan](https://larave
 plugins=(... laravel)
 ```
 
-| Alias | Description |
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry it waited so long for a reply. `pakg` and `pasl` are both worth having, and they match the naming in that file already.

The blocker is the README change: it reformats the whole table, which conflicts with master now and makes the diff much bigger than the two aliases deserve. Could you rebase on master and drop the reformatting, leaving just the two new rows?

If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll take it from here.

PR #10302, opened 2021-10-15
feat(git): Add difftool alias (gdtl)

### LINKED ISSUE #10298: Aliases for git mergetool and git difftool

oh-my-zsh git shortcuts

_No response_

I very often use graphical tools like meld and kdiff for my merges and especially diffs.
The git difftool and git mergetool are long commands.

I would suggest **gdt for git difftool and gmt for git mergetool.**
That's what I have set my own aliases to, but it would be nicer to have them alll ready in oh-my-zsh

I see that gdt conflicts as of now with:
**git diff-tree --no-commit-id --name-only -r**

But I would be surprised if that alias is more used than the ones I suggest.
In that case, rename git diff-tree to **gdtr**

And oh, I just love oh-my-zsh!

Skip it. I would love oh-my-zsh anyway.

### Additional context

_No response_

### Related Issues

_No response_

iccowan:
## Standards checklist:

## Changes:

- Adds the alias `gdtl` alias to call the `git difftool --no-prompt` command
- Documents the new alias in the git plugin documentation

## Other comments:

This command was removed a few years ago in favor of `git diff-tree --no-commit-id --name-only -r` because of the duplicate alias, `gdt`. This addition will add back the command `git difftool --no-prompt` without overwriting any already existing aliases.

Resolves #10298 given that the other alias in that feature request already exists.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -66,6 +66,7 @@ plugins=(... git)
 | gdct                 | git describe --tags $(git rev-list --tags --max-count=1)                                                                         |
 | gds                  | git diff --staged                                                                                                                |
 | gdt                  | git diff-tree --no-commit-id --name-only -r                                                                                      |
+| gdtl                 | git difftool --no-prompt                                                                                                         |
 | gdnolock             | git diff $@ ":(exclude)package-lock.json" ":(exclude)&ast;.lock"                                                                 |
 | gdup                 | git diff @{upstream}                                                                                                             |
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry for the long wait on it.

I closed #10298 a few days ago with the same answer: the mergetool half shipped as `gmtl`, but `difftool` isn't reached for often enough to earn a spot among the plugin's ~200 aliases. We removed `gdt` for the same reason a while back.

Closing this to match. Thanks for taking the time on it.

PR #10310, opened 2021-10-19
feat(git): add aliases for `git diff --ignore-all-space`

afonsocarlos:
## Standards checklist:

## Changes:

- Added alias for `git diff --ignore-all-space` - `gdi`
- Added alias for `git diff --staged --ignore-all-space` - `gdsi`

## Other comments:

I use the option `--ignore-all-space` on almost every diff I do and I thought a lot of people may do this too. So I decided to create these aliases


plugins/git/README.md
```diff
@@ -10,184 +10,186 @@ plugins=(... git)
...
-| gap                  | git apply                                                                                                                        |
-| gapt                 | git apply --3way                                                                                                                 |
-| gb                   | git branch                                                                                                                       |
-| gba                  | git branch -a                                                                                                                    |
-| gbd                  | git branch -d                                                                                                                    |
+| Alias                | Command                                                                                                                                     |
+|:---------------------|:--------------------------------------------------------------------------------------------------------------------------------------------|
```
fredden, line 14: It's not necessary to change this line at all; markdown will render the table properly as-is. (Because every line in this table has changed here, it's hard to see what the intended changes actually are.)
```suggestion
|:-|:-|
```

afonsocarlos: I see. Alright, I've updated it. It should be fixed now

afonsocarlos:
Any updates on this one?

adamchainz:
`--ignore-all-space` has a short option, [`-w`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--w). I don't think it's worth adding extra aliases when `-w` can be quickly appended to existing aliases, like `gd -w`, `gdca -w`, etc.

(Also, [the `-b` option](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--b) is preferable for most programming languages.)

carlosala:
Makes much more sense! Closing, thanks Adam for the triaging!

PR #10318, opened 2021-10-20
fix(bgnotify): avoid permission prompts by checking frontmost app ID

AuroraWright:
## Standards checklist:

## Changes:

- Replaced the applescript to get the pid of the frontmost app instead of the window id of the active window.

## Other comments:

On recent macOS versions, getting the window id of the active window of the frontmost app using `osascript` and `tell` requires granting permission for controlling said app. This means that if `bgnotify` is active and one puts the terminal in the background, after the terminal process completes it will spawn a permission request window relating to the app currently in the foreground. Obviously this gets annoying really quickly as each new foreground app has to be approved separately. Getting the pid of the frontmost app only requires this permission to be granted once, for the "System Events" app.

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -20,9 +20,15 @@ if ! (type bgnotify_formatted | grep -q 'function'); then ## allow custom functi
   }
 fi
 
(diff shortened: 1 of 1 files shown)


mcornella:
Let me know when you get a chance to look at this.

PR #10349, opened 2021-10-26
add npm pack alias

hellpirat:
Hi, there. I added npm pack alias for npm. I am not sure about alias npmpa - maybe do you have any other ideas about naming?


hellpirat:
Done.

mcornella:
Hey @jsoref, aliases are a heavily used feature by OMZ and most of them have weird naming. It should be possible to add a pattern for alias names, right? Just asking before I try it. If I want to try it out [these scripts](https://github.com/check-spelling/check-spelling/wiki/Looking-for-items-locally) would be the way to check locally right?

jsoref:
Yeah, I actually have a thing for it I was considering sending it. Lemme do that now.

mcornella:
I'm afraid this is not going to be needed by most of our users, and if they need it it'll be once or twice. Thanks for your contribution though!

PR #10374, opened 2021-11-02
fix(lib): fix comment typo

carlosala:
## Standards checklist:

## Changes:

I removed two quotes. To disable the compfix, the value of `ZSH_DISABLE_COMPFIX` must be `true`, not `"true"`, as you can see [here](https://github.com/ohmyzsh/ohmyzsh/blob/04c96e235ff522704a6f1482e7fd06a05467cbb7/oh-my-zsh.sh#L122).


mcornella:
Quotes are ignored:
```console
$ var="true"; [[ $var = true ]] && echo true || echo false
true
$ var=true; [[ $var = true ]] && echo true || echo false
true
```

carlosala:
Ups, sorry. I didn't knew that. Thanks!

PR #10381, opened 2021-11-03
fix(command-not-found): pass arguments correctly in NixOS

amnore:
## Standards checklist:

## Changes:

- Fix passing arguments to command-not-found in NixOS

## Other comments:

The [command-not-found script](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/command-not-found/command-not-found.pl) in NixOS uses the first argument to find programs.
Thus there should be no '--' when passing arguments.


mcornella:
Thanks!

PR #10392, opened 2021-11-06
fix(plugins): speed up rustup plugin

### LINKED ISSUE #10390: The new rustup and cargo plugin are incredibly slow
### Describe the bug

After I update my plugins, I found the start time of my ZSH becomes extremely slow. I located the problem to these two plugins.

Here is my `.zshrc` file https://github.com/QuarticCat/dotfiles/blob/a9662d9114747f1080a2fca115a077273febfe24/dot_zshrc

If uncomment the `OMZP::cargo` line and the `OMZP::rustup` line, the start time will be 3 times slower.

### Steps to reproduce

Copy the content of these two plugins into files and test their execution time.

```console
$ hyperfine './test.sh'
```

### Expected behavior

Shouldn't cost too much time.

The result of rustup plugin:

```test
Benchmark 1: ./test.sh
  Time (mean ± σ):      67.8 ms ±   6.4 ms    [User: 57.0 ms, System: 10.9 ms]
  Range (min … max):    62.3 ms …  84.6 ms    44 runs
```

The result of cargo plugin:

```
Benchmark 1: ./test.sh
  Time (mean ± σ):      66.7 ms ±   7.0 ms    [User: 56.6 ms, System: 10.0 ms]
  Range (min … max):    57.1 ms …  77.3 ms    49 runs
```

### OS / Distro

Arch Linux

### Zsh version

5.8

### Terminal emulator

yakuake

_No response_

### Additional context

_No response_

QuarticCat:
## Standards checklist:

## Changes:

- Change `rustup --version` to `rustup --help | head -n 1`, the result is the same but `--version` is much slower for some reason. See https://github.com/ohmyzsh/ohmyzsh/issues/10390

## Other comments:

`cargo --version` is also super slow, but I didn't find a faster alternative.

Fixes #10390

### DIFF
--- a/plugins/rustup/rustup.plugin.zsh
+++ b/plugins/rustup/rustup.plugin.zsh
@@ -2,7 +2,7 @@ if (( $+commands[rustup] )); then
   # remove old generated completion file
   command rm -f "${0:A:h}/_rustup"
 
-  ver="$(rustup --version 2>/dev/null)"
+  ver="$(rustup --help | head -n 1)" # `rustup --version` is slow
   ver_file="$ZSH_CACHE_DIR/rustup_version"
   comp_file="$ZSH_CACHE_DIR/completions/_rustup"
 


carlosala:
@mcornella should be merged fast, the with `rustup --version` uses `70ms` and with `rustup --help | head -n 1` just `2ms`, tested with hyperfine.

carlosala:
@QuarticCat you should do the same change to `cargo` plugin, it uses the same `rustup --version` command

carlosala:
I tried a little bit with hyperfine, and I saw that `rustup completions zsh` and `rustup completions zsh cargo` are really fast, just a couple of ms. Compared to --version option, there is a big improvement. Maybe we should shift to reloading the completion file on every init, as it's going to be faster than trying to compare the version. If @mcornella agree, I'm going to open a PR extending this one.

QuarticCat:
> Maybe we should shift to reloading the completion file on every init.

Vote for this.

carlosala:
Let's see if Marc agree, as he is the main maintainer ;)

> > Maybe we should shift to reloading the completion file on every init.
> 
> Vote for this.

mcornella:
I'm thinking whether we can generate teh completion script asynchronously, and still have the `autoload` + manual `compdef` before, skipping both cargo and rustup version checks. Something like:

- cargo.plugin.zsh:
  ```zsh
  if ! (( $+commands[rustup] && $+commands[cargo] )); then
    return
  fi

  # Remove old generated completion file
  command rm -f "${0:A:h}/_cargo"

  # Add completions/ folder in $ZSH_CACHE_DIR
  comp_file="$ZSH_CACHE_DIR/completions/_cargo"

  # If the completion file doesn't exist yet, we need to autoload it and
  # bind it to `cargo`. Otherwise, compinit will have already done that
  if [[ ! -f "$comp_file" ]]; then
    autoload -Uz _cargo
    declare -A _comps
    _comps[cargo]=_cargo
  fi

  # Generate completion file in the background
  rustup completions zsh cargo >| "$comp_file" &|
  unset comp_file
  ```

- rustup.plugin.zsh:
  ```zsh
  if (( ! $+commands[rustup] )); then
    return
  fi

  # Remove old generated completion file
  command rm -f "${0:A:h}/_rustup"

  # Add completions/ folder in $ZSH_CACHE_DIR
  comp_file="$ZSH_CACHE_DIR/completions/_rustup"

  # If the completion file doesn't exist yet, we need to autoload it and
  # bind it to `rustup`. Otherwise, compinit will have already done that
  if [[ ! -f "$comp_file" ]]; then
    autoload -Uz _rustup
    declare -A _comps
    _comps[rustup]=_rustup
  fi

  # Generate completion file in the background
  rustup completions zsh >| "$comp_file" &|
  unset comp_file
  ```

Also I'm considering whether we should merge both these plugins into a single `rust` plugin, since it seems like everyone using Rust needs both of these tools.

Apologies for the delay!

carlosala:
I think both ideas are great. I opened a PR (#10270) a few time ago merging both plugins in one, but I can also put this async changes into it.
If you think it's great, I go for it

mcornella:
Sounds good for me, but maybe calling it `rust` would be more simple? I don't know a lot about the Rust programming culture though, I may be wrong. 

Don't worry about the async code, let's fix this PR first and then we can merge them if everyone agrees. You'll have to rebase your PR but that's not going to be too difficult.

carlosala:
Great, when this will be done I will go for that rebate. Thanks Marc!

bjorn3:
> Sounds good for me, but maybe calling it rust would be more simple?

Makes sense to me. While it is possible to install rust without rustup, this is not very common I believe. If you install rustc, you will get cargo by default. Cargo is used most of the time as build system, so this is what you tend to interact with if you are using rust.

mcornella:
I applied the proposed changes and moved both `cargo` and `rustup` to the `rust` plugin, as part of #10270. You'll have to change your zshrc to use `rust` instead.

carlosala:
@mcornella maybe we should apply this same background code to many other plugins that use a similar system, to continue speeding up omz. What do you think about?

mcornella:
> @mcornella maybe we should apply this same background code to many other plugins that use a similar system, to continue speeding up omz. What do you think about?

Fixed in 49bae43f6202e7e1171d8911d8b243714dfe888a.

PR #10416, opened 2021-11-12
add toolbox plugin

mikebarkmin:
## Standards checklist:

## Description

[Toolbox](https://github.com/containers/toolbox) is a tool for containerized command line environments on Linux. It is shipped for example with Fedora Silverblue. Since toolbox should be used for isolated developer enviroments similiar to virtualenv but with containers, I wrote this plugin to ease the use of toolbox.

The plugin acts in similar fashin to the virtualenvwrapper plugin. If a git directory is found it will enter the default toolbox, or it will read a container name from a `.toolbox` file. If the git directory is exited, the toolbox is also exited. This behavior can be customized with env variables.

Additionally, some convenient aliases are set.

All of this is also documented in the README.

### DIFF
--- a/plugins/toolbox/toolbox.plugin.zsh
+++ b/plugins/toolbox/toolbox.plugin.zsh
@@ -1,5 +1,112 @@
-function toolbox_prompt_info() {
-  [[ -f /run/.toolboxenv ]] && echo "⬢"
+if [[ $? -eq 0 ]] && ! type toolbox &>/dev/null; then
+  print "[oh-my-zsh] toolbox plugin: shell function 'toolbox' not defined.\n" \
+    "Please check toolbox" >&2
+  return
+fi
+
+function is_in_toolbox() {
+  if [[ -f /run/.containerenv && -f /run/.toolboxenv ]]; then
+    return 0
+  else
+    return 1
+  fi
 }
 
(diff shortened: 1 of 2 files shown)


jkemp814:
Excellent!

mcornella:
We merged a very barebones version of a toolbox plugin, can you rebase and update with your changes if you're still interesting in this?

jbtrystram:
@mcornella this plugin is great ! How can help to get this merged ?

plugins/toolbox/toolbox.plugin.zsh
```diff
@@ -1,5 +1,110 @@
...
+        toolbox --container "$TOOLBOX_NAME" run sudo hostname "$TOOLBOX_NAME"
+        toolbox enter "$TOOLBOX_NAME"
+      fi
+    elif [[ "$(hostname)" != "toolbox" && $DISABLE_TOOLBOX_EXIT -ne 1 ]]; then
+      if is_in_toolbox; then
+        exit
+      fi
+    fi
+  fi
+}
```
mcornella, line 57: Could you describe what this does in detail, and where it is run? I doubt whether the exit will be reached at any point in time, and also I can't figure out when it is checked that we have cd-ed outside of the project root.

PR #10427, opened 2021-11-16
Fix iterm2 detection

frobnicaty:
https://github.com/ohmyzsh/ohmyzsh/blob/fb12e4135311b9c4189a7e4556be619922052f6b/plugins/iterm2/iterm2.plugin.zsh#L8

One might be running iterm2 locally, but in a remote shell (possibly nested) that's not on OSX, and possibly pretending to be a different terminal (eg xterm-256color).

A better test is to directly check for iterm2's zsh integration, which will create the function `iterm2_set_user_var`.

## Standards checklist:

### DIFF
--- a/plugins/iterm2/iterm2.plugin.zsh
+++ b/plugins/iterm2/iterm2.plugin.zsh
@@ -4,8 +4,8 @@
 #####################################################
 
 ###
-# This plugin is only relevant if the terminal is iTerm2 on OSX.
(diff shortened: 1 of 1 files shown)


mcornella:
I've tried looking for `iterm2_set_user_var` and it is not present in my iTerm2 session, running on v3.4.12.

mcornella:
Closing due to inactivity.

PR #10432, opened 2021-11-18
feat(fig): add plugin for Fig

marcofranssen:
Adds autocomplete for [Fig](https://fig.io)

## Standards checklist:

## Changes:

- Add [Fig](https://fig.io) autocompletion plugin

### DIFF
--- a/plugins/fig/README.md
+++ b/plugins/fig/README.md
@@ -1,0 +1,9 @@
+# Fig plugin
+
+This plugin sets up completion for [Fig](https://fig.io/).
+
+To use it, add `fig` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... fig)
+```
--- a/plugins/fig/fig.plugin.zsh
+++ b/plugins/fig/fig.plugin.zsh
@@ -1,0 +1,13 @@
+if ! (( $+commands[fig] )); then
+  return
+fi
+
+# If the completion file doesn't exist yet, we need to autoload it and
+# bind it to `fig`. Otherwise, compinit will have already done that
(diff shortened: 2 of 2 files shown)


robbyrussell:
Would be great to get some people to help us test this out.

mcornella:
This would be better if the output from `fig completion zsh` is run each time, kinda like the `deno`, `cargo` or `rustup` plugins (note that the logic on these is bound to change, so it's done properly).

marcofranssen:
> This would be better if the output from `fig completion zsh` is run each time, kinda like the `deno`, `cargo` or `rustup` plugins (note that the logic on these is bound to change, so it's done properly).

Much better indeed. Thanks for sharing. I have updated the plugin to be like that.

mcornella:
Looks great, thanks!

lomasz:
Is it typo in this line:
```bash
fig completions zsh >| "$ZSH_CACHE_DIR/completions/_fig" &|
``` 
?

Message after enabling plugin and rerun console:
```
Error: unknown command "completions" for "fig"

Did you mean this?
	completion

Run 'fig --help' for usage.
```

mcornella:
Fixed, thanks for catching that!

tomasp1189:
I have added the plugin but it doesn't seem to work. Other than adding `fig` to the plugins array in my `.zshrc` file, do I have to do anything else? BTW I installed fig before setting up Oh My Zsh, not sure if this is relevant.

mcornella:
The plugin only adds completion, you still need the initialization code that fig adds automatically to .zshrc.

tomasp1189:
Awesome! I rerun the install script option provided by fig) and it works! Thanks!

PR #10438, opened 2021-11-22
Add plugin for ExpressVPN

deb0ch:
ExpressVPN comes with a completion script for Bash, but none for Zsh. Fortunately, I was able to make it work for both shells pretty easily with only a small modification (see diff below).

This code is not publicly hosted by ExpressVPN, so I figured that here would be a place that makes sense to publish it.

I also submitted the fix via ExpressVPN's support, although I do not know how much faith I should have that it will make it into their codebase someday.

This fix makes the autocompletion script also work on zsh without changing how it works for Bash. The expression "${COMP_WORDS[@]:2:$COMP_CWORD-2}" does not behave the
same on Bash and Zsh when the command has only zero or one arguments,
e.g. typing "expressvpn conne<TAB>" or "expressvpn <TAB>" would print
the error "autocomplete:8: substring expression: 1 < 2". This fixes it
by handling the case of the short command separately in a simpler way
and handling the rest of the cases the same way as before.

With this fix, zsh users just have to use the command "source
/usr/share/bash-completion/completions/expressvpn" to get the same
auto-completion as bash users.

File : /usr/share/bash-completion/completions/expressvpn

Diff to the original file (using the `diff` command) :

```diff
9a10,12
>         opts=$( ${COMP_WORDS[0]} "$cmd" "${COMP_WORDS[@]:2:$COMP_CWORD-2}" --generate-bash-completion )
>     else
>         opts=$( ${COMP_WORDS[0]} --generate-bash-completion )
11d13
<     opts=$( ${COMP_WORDS[0]} "$cmd" "${COMP_WORDS[@]:2:$COMP_CWORD-2}" --generate-bash-completion )
```
**NB:** bash is regularly referenced throughout the file, it is on purpose as I wanted to keep the diff to the original file to a minimum.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/expressvpn/README.md
+++ b/plugins/expressvpn/README.md
@@ -1,0 +1,14 @@
+# ExpressVPN
+
+This plugin provides completion support for [`ExpressVPN`](https://www.expressvpn.com/vpn-software/vpn-linux)
+command line interface on Linux.
+
+To use it, add expressvpn to the plugins array in your zshrc file.
+
+```zsh
+plugins=(... expressvpn)
+```
+
+## INSTALLATION NOTES
+
+Besides oh-my-zsh, `expressvpn` needs to be installed by following these steps: https://www.expressvpn.com/support/vpn-setup/app-for-linux/.
--- a/plugins/expressvpn/_expressvpn
+++ b/plugins/expressvpn/_expressvpn
@@ -1,0 +1,49 @@
+#compdef expressvpn
+
+# bash completion for expressvpn                           -*- shell-script -*-
+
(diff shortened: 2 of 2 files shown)


robbyrussell:

Two things stop it. The file is ExpressVPN's own bash completion with a `#compdef` header added, so it runs through our bashcompinit bridge and only completes from the second TAB press, and its licensing isn't clear enough for us to ship. ExpressVPN also retired that CLI earlier this year in favour of the new app.

Closing for those reasons. Thanks for taking the time on it.

PR #10475, opened 2021-12-03
ci(spelling): automatically accept aliased commands

jsoref:
## Standards checklist:

## Changes:

- `expect.txt` shrinks a bit as items that are defined as aliases are automatically included in the dictionary

## Other comments:

Addresses https://github.com/ohmyzsh/ohmyzsh/pull/10349#issuecomment-985725019

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -22,6 +22,7 @@ acr
 acroread
 acs
 acsc
+acss
 acsp
 actionformats
 Adamantium
@@ -1459,6 +1460,7 @@ gpr
 gpristine
 gpsup
 gpu
+gpus
 gpv
 gradle
 gradlew
@@ -2396,6 +2398,7 @@ mirrorlist
 mixin
 mkcd
 mkdir
+mkdirs
 mktag
 mktemp
 mktree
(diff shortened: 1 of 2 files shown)


jsoref:
ok, the bot hates me, let me drop the expect changes, they can be addressed in the next PR after...

jsoref:
@mcornella, [these scripts](https://github.com/check-spelling/check-spelling/wiki/Looking-for-items-locally) are helpful for finding instances of items when you're trying to "fix" things, but they won't help _much_ here.

If you want to test how things will work, you can either use [act](https://github.com/nektos/act) or a fork where you've enabled actions (e.g. https://github.com/jsoref/ohmyzsh/actions).

The reason that a simple pattern won't help is that while an item is used as an alias in a regular form (`alias xyz=...`), it's also _documented_ (and this is a good thing, hurray for projects with documentation!) alone `xyz | ...`, which means that the naive pattern would hide the first and leave the second item stranded. And you certainly don't want to skip the documentation since it tends to be user-facing and can certainly have typos.

What this PR does is take advantage of the regular form (`alias xyz=...`) and adds that term to the local dictionary for the run, which means that when the spell checker sees `xyz` in the documentation, it won't complain about it.

PR #10477, opened 2021-12-04
Add support for aws-vault

m-zare:
AWS Vault is a tool to securely store and access AWS credentials in a development environment.
https://github.com/99designs/aws-vault

## Standards checklist:

## Changes:

- [Add support for aws-vault]

## Other comments:

...


maksyms:
@m-zare  - thanks for your contribution! I'm not using `aws-vault`, so I cannot check whether it works. Do you know anyone who does who could confirm whether it works or not?

@robbyrussell - agreed that testing is needed, although the change is innocent enough to at least not regress other functionality. I think it is a judgement call as to whether to wait for testers or move ahead on @m-zare 's confirmation.

Lets see whether we can find someone to test.

m-zare:
@shahinism, can you please test this PR?

mcornella:
I'm hesitant to merge this because it overrides the `$AWS_PROFILE` value. Was that the intention, or do you just want to display `$AWS_VAULT` if `$AWS_PROFILE` is empty?

maksyms:
> Was that the intention, or do you just want to display `$AWS_VAULT` if `$AWS_PROFILE` is empty?

I know the question isn't for me, but here is how I understood it: unless `$AWS_VAULT` is set, the software behaves as before. But a reasonable assumption could be that only the people who use `aws-vault` would set `$AWS_VAULT`, for whatever purpose - and what is supposed to happen then, I don't know.

I'm not the right person here, happy to wait on other testers confirmation.

m-zare:
@mcornella, both. Actually, `$AWS_VAULT`'s value is an AWS profile name. IMHO, there is no objection to duplicating its value in `$AWS_PROFILE`.

@maksyms is right, `$AWS_VALUT` is a variable that `aws-vault` set. You won't need this variable unless you don't use this app.

mcornella:
> IMHO, there is no objection to duplicating its value in `$AWS_PROFILE`.

Yeah, but `$AWS_PROFILE` is used by other stuff. We have entire functions to change the profile, with many checks. It seems a side effect to set `$AWS_PROFILE` in a prompt function, and something that might confuse users.

I'm not a user of aws-vault, but it seems like they'd change `$AWS_PROFILE` if that was what they intended. Instead, we can use a local variable and set that one, instead of the env var `$AWS_PROFILE`.

Something like
```sh
function aws_prompt_info() {
  # if $AWS_PROFILE is not set, display $AWS_VAULT
  local profile=${AWS_PROFILE:-$AWS_VAULT}

  # if neither $AWS_PROFILE nor $AWS_VAULT are set, do nothing
  [[ -n "$profile" ]] || return

  echo "${ZSH_THEME_AWS_PREFIX:=<aws:}${profile/%/%%}${ZSH_THEME_AWS_SUFFIX:=>}"
}
```

m-zare:
@mcornella, the snippet you provided didn't work but gave me the idea.
The current PR doesn't touch `$AWS_PROFILE`; it looks for the presence of `$AWS_VAULT`.

PR #10481, opened 2021-12-05
Fixing the path to homebrew installation of `asdf`

sshishov:
## Standards checklist:

## Changes:

As the. code has been moved to `libexec` it seems, the path should be updated accordingly

## Other comments:

...

### DIFF
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -4,7 +4,7 @@ ASDF_COMPLETIONS="$ASDF_DIR/completions"
 
 # If not found, check for Homebrew package
 if [[ ! -f "$ASDF_DIR/asdf.sh" || ! -f "$ASDF_COMPLETIONS/asdf.bash" ]] && (( $+commands[brew] )); then
-   ASDF_DIR="$(brew --prefix asdf)"
+   ASDF_DIR="$(brew --prefix asdf)/libexec"
    ASDF_COMPLETIONS="$ASDF_DIR/etc/bash_completion.d"
 fi
 
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -5,7 +5,7 @@ ASDF_COMPLETIONS="$ASDF_DIR/completions"
 # If not found, check for Homebrew package
 if [[ ! -f "$ASDF_DIR/asdf.sh" || ! -f "$ASDF_COMPLETIONS/asdf.bash" ]] && (( $+commands[brew] )); then
    ASDF_DIR="$(brew --prefix asdf)/libexec"
-   ASDF_COMPLETIONS="$ASDF_DIR/etc/bash_completion.d"
+   ASDF_COMPLETIONS="$(brew --prefix asdf)/etc/bash_completion.d"
 fi
 
 # Load command


jthegedus:
This change may be premature. Additionally, #9615 is conflicting change.

sshishov:
Yes, this change is just the possible solution without knowing the internals. We should not blindly merge it. Just need to understand why all of a sudden on release 0.9.0 of ASDF we start having issues

orien:
Looks related to https://github.com/Homebrew/homebrew-core/commit/cd3d21ba3a2c1041b64191b82cee3bffcbe2efdf

This fixes the problem for me.

emiliadecaudin:
I found the reason the plugin no longer works; it's because of a change in asdf's code, not the Homebrew formula: https://github.com/asdf-vm/asdf/commit/73efc9fa97744c49c5004ee8bb9b6064b6ce22f2. The change in this PR is not 100% consistent with how asdf previously changed $ASDF_DIR (result is /usr/local/opt/asdf/libexec instead of /usr/local/Cellar/asdf/0.9.0/libexec [the former is symlinked to the latter]), however, this is actually how Homebrew prefers formulae be accessed, and it causes no issues with asdf. (If anything, it makes the shims more stable across updates.

jthegedus:
I believe @ericbn is a Homebrew user and part of the change in https://github.com/asdf-vm/asdf/pull/1008 was an incremental change. I wasn't expecting it to break peoples setups in such a way though. He is a contributor on the diff shared by @orien [above](https://github.com/Homebrew/homebrew-core/commit/cd3d21ba3a2c1041b64191b82cee3bffcbe2efdf)

orien:
As shared on https://github.com/asdf-vm/asdf/issues/1103, I believe this is the correct fix.

jthegedus:
> As shared on [asdf-vm/asdf#1103](https://github.com/asdf-vm/asdf/issues/1103), I believe this is the correct fix.

For the OMZSH plugin yes, though there is still an Homebrew Formulae issue.

ericbn:
The fix LGTM. I just wonder: why is a Zsh plugin loading Bash completions? asdf comes with Zsh completions too.

FYI, this is how I'm initializing asdf: https://github.com/zimfw/asdf/blob/master/init.zsh

ericbn:
The reason the current Oh-My-Zsh plugin code no longer works starting with asdf v0.9.0 is because of this change in the release:

* don't override existing ASDF_DIR ([#1008](https://www.github.com/asdf-vm/asdf/issues/1008)) ([73efc9f](https://www.github.com/asdf-vm/asdf/commit/73efc9fa97744c49c5004ee8bb9b6064b6ce22f2))

Previously the ASDF_DIR set by the Oh-My-Zsh plugin would be overriden by asdf with the correct path. And the ASDF_DIR path is `$(brew --prefix asdf)/libexec` since the changes in https://github.com/Homebrew/homebrew-core/pull/81664.

plugins/asdf/asdf.plugin.zsh
```diff
@@ -4,7 +4,7 @@ ASDF_COMPLETIONS="$ASDF_DIR/completions"
 
 # If not found, check for Homebrew package
 if [[ ! -f "$ASDF_DIR/asdf.sh" || ! -f "$ASDF_COMPLETIONS/asdf.bash" ]] && (( $+commands[brew] )); then
-   ASDF_DIR="$(brew --prefix asdf)"
+   ASDF_DIR="$(brew --prefix asdf)/libexec"
    ASDF_COMPLETIONS="$ASDF_DIR/etc/bash_completion.d"
```
Snailslug, line 8: ASDF_COMPLETIONS does not evaluate to a valid path with this change. /etc/bash_completions.d lives under the root asdf directory rather than under libexec.

orien:
```suggestion
   ASDF_COMPLETIONS="$(brew --prefix asdf)/etc/bash_completion.d"
```

sshishov: Applied changes, @orien thanks!

emiliadecaudin:
Should this latest commit read `ASDF_COMPLETIONS="$ASDF_DIR/../etc/bash_completion.d"` to avoid running `git --prefix asdf` twice?

PR #10510, opened 2021-12-15
feat(init): add options to disable Oh My Zsh aliases

### LINKED ISSUE #6630: need contributor / plug-in guidelines on aliases and disabling setting of aliases
Some good discussion has already happened in #5721 but this is an issue not just limited to the plugin for git.

Some examples:
#4780
#5214
#5288
#5313
#5766
#5774
#5783
#6062
#6601

There needs to be some form of guidance on how to add aliases while being a good citizen, especially since some aliases can be very dangerous and destructive. It would be nice to adopt the added functionality that many plugins provide, but have the option to use, or not use the aliases provided by the plugin.

Similarly, the libraries also exhibit some of this. `lib/directories.zsh` creates 23 aliases on it's own, `lib/misc.zsh` creates 2 different aliases for `sudo`. In comparison, `lib/correction.zsh` also creates a number of aliases, but they are gated behind an option so I can easily enable or disable them.

I've avoided some plugins because of the sheer number of aliases I'd be "forced" to have from them. For the 23 aliases from `lib/directories.zsh` I resorted to creating an empty `custom/lib/directories.zsh` to disable them. I don't consider it to be realistic to have each user have to check every update to see if any new aliases were created, then add an `unalias yet_another_alias` for every new unwanted alias.

### LINKED ISSUE #7328: Load plugin completions without aliases
Oh-my-zsh is great! It makes configuring zsh very easy.

However, it does a bit too much for my taste. Many plugins tend to add autocompletion for a command and a bunch of aliases. Personally I love to have that autocompletion, but I don’t like most of the aliases.

Is there a nice way to load only plugin completions, but not the aliases?

### LINKED ISSUE #8015: Disable Aliases from plugins I didn't enable
It appears that a bunch of aliases have been enabled upon installing `oh-my-zsh`, even though the plugin files in which they are declared were never included in my `.zshrc`

How can I disable all those aliases?

P.S.: I noticed several similar issues but none actually discussing this specific issue.

### LINKED ISSUE #8344: cannot change to a directory named "..."
If there is a folder named as "..." , I cannot change into the folder using cd .../ , rather it would convert ... to ../../ and will execute cd ../../

**To Reproduce type the following commands**
bash(Going to bash for creating a folder)
mkdir ...(making a folder named "..." as zsh will be unable to make such a folder as it will convert ... to ../../)
exit(moving to zsh again from bash)
cd ...(it will change "..." to ../../ and will execute cd ../../ instead of moving to "..." directory)

**Expected behavior**
It should search for the "..." folder and if it does not exit then it should convert "..." to "../../" else it should execute ...

**System:**
 - x86_64-ubuntu-linux-gnu
 - zsh 5.5.1

**Additional context**
I know zsh has auto feature to convert ... to ../../ but it should first search for that file and if the file is not found then it should convert it making it much more convenient to use.

### LINKED ISSUE #8933: Oh My Zsh aliases override .profile aliases
**Describe the bug**

Oh My Zsh defines a bunch of aliases that I don't want, when I already define these in .profile for use with zsh, bash, sh, and ksh.

**To Reproduce**

1. Define `ll` in `$HOME/.profile`
2. Source `$HOME/.profile` from `$HOME/.zshenv`
3. Install Oh My Zsh.

**Expected behavior**

My aliases aren't messed with.

**Observed behavior**

Oh My Zsh aggressively overwrites many aliases, without bothering to check if these are already defined.

Please offer an opt-out environment variable to disable Oh My Zsh from defining these aliases. Really annoying.

### LINKED ISSUE #9414: Add option to disable ohmyzsh's aliases
Yes. I don't want aliases i never asked for. When I do `alias` i see that there are aliases like `egrep='egrep --color=auto --exclude-dir={.bzr,CVS,.git,.hg,.svn,.idea,.tox}'`, which I have no idea about.

It would be nice if I could set a variable like OMZ_ALIASES=false, so that i can disable aliases, or a even better solution would be to maintain the list of aliases in a separate file, so that I can comment out the ones if don't want//

I have tried to manually comment out the aliases i don't want by grep-ing, but that is tiresome

**Additional context**
I am not talking about the git aliases, which are rather helpful. I am talking about the aliases that are sourced even when `plugins=()`

**Related Issues**
Unwanted aliases can lead to people wondering what they are doing wrong when they aren't doing anything wrong, and its the alias they never knew they had causing the problem.

### LINKED ISSUE #9621: Activate completion features without adding any aliases
Sort of. Right now specifically the npm plugin, but it applies to others too.

Often I want the completion that a plugin provides but I never want any of the aliases. I have no interest in learning them, and if my shell is littered with aliases it massively increases the possibility that when I make a typo in a short command something totally unexpected happens.

Some way to specify whether or not I want each plugin's aliases. Perhaps it would be simplest to split completion and aliases to separate plugins?

I could edit the source to comment out all the aliases but that's not maintainable.
I could unalias everything after the fact but that's a mess and not maintainable.
I could install the completion separately but then why am I using ohmyzsh.

gabrielcosta42:
Closes #9414, resolves #8344, closes #7328, closes #6630, resolves #8933, resolves #8015, closes #9621

## Standards checklist:

## Changes:

- Added option to not enable any alias from Oh My Zsh: DISABLE_OMZ_ALIASES
- Added option to not enable Oh My Zsh plugins aliases: DISABLE_OMZ_PLUGIN_ALIASES

## Other comments:
Long time requested feature.
...

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -143,6 +143,12 @@ fi
 
 unset zcompdump_revision zcompdump_fpath zcompdump_refresh
 
+# Store current aliases if should not enable the default ones from Oh My Zsh 
+local aliases_backup="/tmp/aliases_backup.sh"
+
+if [[ "$DISABLE_OMZ_ALIASES" == "true" ]]; then
+  alias > $aliases_backup
+fi
 
 # Load all of the config files in ~/oh-my-zsh that end in .zsh
 # TIP: Add files you don't want in git to .gitignore
@@ -152,6 +158,11 @@ for config_file ($ZSH/lib/*.zsh); do
   source $config_file
 done
 
+# Store current aliases if should not enable the ones from plugins
+if [[ "$DISABLE_OMZ_PLUGIN_ALIASES" == "true" ]] && [[ "$DISABLE_OMZ_ALIASES" != "true" ]]; then
+  alias > $aliases_backup
+fi
+
 # Load all of the plugins that were defined in ~/.zshrc
 for plugin ($plugins); do
   if [ -f $ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh ]; then
@@ -160,6 +171,15 @@ for plugin ($plugins); do
     source $ZSH/plugins/$plugin/$plugin.plugin.zsh
   fi
 done
+
+# Restore saved aliases
+if [[ "$DISABLE_OMZ_ALIASES" == "true" ]] || [[ "$DISABLE_OMZ_PLUGIN_ALIASES" == "true" ]]; then
+  if [ -f "$aliases_backup" ]; then
+    unalias -a
+    sed -i -e 's/^/alias /' $aliases_backup
+    source $aliases_backup
+  fi
+fi
 
 # Load all of your custom configurations from custom/
 for config_file ($ZSH_CUSTOM/*.zsh(N)); do
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -43,6 +43,12 @@ ZSH_THEME="robbyrussell"
 # Uncomment the following line to enable command auto-correction.
 # ENABLE_CORRECTION="true"
 
+# Uncomment the following line to disable all Oh My Zsh aliases.
+# DISABLE_OMZ_ALIASES="true"
+
+# Uncomment the following line to disable only plugin related aliases.
+# DISABLE_OMZ_PLUGIN_ALIASES="true"
+
 # Uncomment the following line to display red dots whilst waiting for completion.
 # You can also set it to another string to have that shown instead of the default red dots.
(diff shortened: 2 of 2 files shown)


mcornella:
Hi @gabrielcosta42, thanks for the Pull Request and getting all the issues at the same time. I don't love the solution though. There's a proposal for a way to turn off aliases in https://github.com/ohmyzsh/ohmyzsh/issues/10644. Please have a look and add your feedback. Thanks!

gabrielcosta42:
@mcornella `zstyles` seems to be the way to go, I`m not familiar with it though, it will take some time to learn.

carlosala:
Yes, `zstyle` is absolutely the way of customizing things in `ZSH` nowadays. Adapting OMZ to it seems the most rational thing to do

azinsharaf:
Does this PR still in review?

carlosala:
We'll be implementing in the future a better approach to it. Closing this to avoid confusion.
Thanks for the contribution!

PR #10518, opened 2021-12-18
Handle dollar zero ($0) according to the zsh plugin standard

ctrueden:
## Problem

While setting up [zpm](https://github.com/zpm-zsh/zpm), I discovered that some oh-my-zsh plugins don't work correctly with it. In my case, `macos` and `z`:

```
(anon):source:261: no such file or directory: /home/curtis/code/ctrueden/dotfiles/music
(anon):source:264: no such file or directory: /home/curtis/code/ctrueden/dotfiles/spotify
(anon):source:2: no such file or directory: ./z.sh
```

## Diagnosis

The problem is caused by the usage of the `$0` environment variable. The author of zpm, @grigorii-horos, writes in zpm-zsh/zpm#39 that a declaration `0=${(%):-%N}` needs to be present for expressions involving `$0`, but the affected plugins do not have such a declaration. There is also another issue zpm-zsh/zpm#36 specifically discussing this issue with respect to the `macos` plugin.

## Workaround

In the case of `zpm`, you can work around the problem by adding a `,source:my.plugin.zsh` suffix to the `zpm load` command. For example, for `z`, change:
```zsh
zpm load @omz/z
```
to:
```
zpm load @omz/z,source:z.plugin.zsh
```
But if I understand correctly, this makes `zpm` re-source the plugin every time, rather than caching it for performance?

Edit: I have now tested on both macOS and Linux, and the above workaround only works on my Linux system, not my macOS system. So YMMV. But the solution implemented in this PR addresses the problem on both platforms.

## Proposed solution

I decided I would try my hand at a PR for the `macos` plugin. But while grep for `${0...}` usages within ohmyzsh codebase, I discovered quite a few. So this PR then became an attempt to fix all the instances of this issue that I found. I am a long-time *nix shell user, but not a zsh expert, so please consider it an attempt to start a dialogue about this issue, so that we can have better behavior between oh-my-zsh, zpm, and potentially other plugin managers as well.

## Points to consider

* As part of this work, I wanted to put a comment next to every `0=` declaration pointing to the Zsh Plugin Standard document, which is apparently where this whole thing got started. Unfortunately, the commonly shared URL for that standard, `http://zdharma.org/Zsh-100-Commits-Club/Zsh-Plugin-Standard.html`, is now defunct. (One plugin in this repository, `zsh-navigation-tools`, [references that broken link](https://github.com/ohmyzsh/ohmyzsh/blob/904f8685f75ff5dd3f544f8c6f2cabb8e5952e9a/plugins/zsh-navigation-tools/zsh-navigation-tools.plugin.zsh#L10-L11), so I updated it as part of this PR.) I found a newer functional URL for the standard at https://z-shell.github.io/zsh-plugin-assessor/Zsh-Plugin-Standard, although it appears that the `zsh-plugin-assessor` repository was recently renamed from `Zsh-100-Commits-Club`, so another URL `http://z-shell.github.io/Zsh-100-Commits-Club/Zsh-Plugin-Standard` is also floating around and unfortunately does not redirect to the current location.

* The Zsh Plugin Standard's recommended `$0` block is more complex than the one given by @grigorii-horos; I opted for the one given in the standard. If people think the more concise expression `0=${(%):-%N}` is better/sufficient, I can update the PR to use that expression instead—just let me know.

* I tried to put the `$0` handling code near where the `$0` gets referenced in each plugin, and nested inside the corresponding `if` statement where applicable. If people think it would be better to have these declarations always at the top of each file, I can update the PR to do that instead—just let me know.

* I don't like that the same 2-4 lines of code are repeated in many places across the codebase. I'm hopeful that the oh-my-zsh experts here can suggest a [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)er way of organizing this—perhaps some common function that can be used as a one-liner where needed? But regardless, I think this PR is a step forward compared to the currently broken behavior of various plugins.

* I confess that I did not test this change for every affected plugin, since I don't use them all. However, I did update my zpm installation to use this PR's branch, and the errors cited above go away, and everything appears to work correctly. If more thorough testing is needed, please just let me know what is needed and I will try to do it.

Finally, apologies if this topic has already been covered in a previous issue or PR. I tried searching for it, but could not find anything—as you can imagine, expressions like `0=${(%):-%N}` do not exactly make search engines happy. 😅

## Standards checklist:

- [x] The PR doesn't replicate another PR which is already open (as far as I can tell!).

### DIFF
--- a/plugins/cargo/cargo.plugin.zsh
+++ b/plugins/cargo/cargo.plugin.zsh
@@ -1,4 +1,9 @@
 if (( $+commands[rustup] && $+commands[cargo] )); then
+  # Handle $0 according to the standard:
+  # https://z-shell.github.io/zsh-plugin-assessor/Zsh-Plugin-Standard#zero-handling
+  0="${${ZERO:-${0:#$ZSH_ARGZERO}}:-${(%):-%N}}"
+  0="${${(M)0:#/*}:-$PWD/$0}"
+
   # remove old generated completion file
   command rm -f "${0:A:h}/_cargo"
 
--- a/plugins/colemak/colemak.plugin.zsh
+++ b/plugins/colemak/colemak.plugin.zsh
@@ -19,6 +19,11 @@ bindkey -a 'N' vi-join
 bindkey -a 'j' vi-forward-word-end
 bindkey -a 'J' vi-forward-blank-word-end
 
+# Handle $0 according to the standard:
+# https://z-shell.github.io/zsh-plugin-assessor/Zsh-Plugin-Standard#zero-handling
+0="${${ZERO:-${0:#$ZSH_ARGZERO}}:-${(%):-%N}}"
+0="${${(M)0:#/*}:-$PWD/$0}"
+
 # New less versions will read this file directly
 export LESSKEYIN="${0:h:A}/colemak-less"
 
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -16,6 +16,11 @@ less_termcap[se]="${reset_color}"
 less_termcap[us]="${fg_bold[green]}"
 less_termcap[ue]="${reset_color}"
 
+# Handle $0 according to the standard:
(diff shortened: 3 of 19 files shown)


mcornella:
Thanks for the contribution!

PR #10522, opened 2021-12-21
feat(adduser): add adduser plugin

ngaro:
A plugin with a function `adduser` that does an automatic "Oh My ZSH" installation after adding new users (which happens with the real`adduser`)

## Standards checklist:

## Other comments:

**Dependencies**:
- Obviously `adduser`itself... 
- Everything needed for a regular "Oh My ZSH" installation: `git`, `curl`, `zsh` and openssl with certificates.<br>_(But you obviously already have these because you managed to install "Oh My ZSH" for yourself ...)_
- `sudo` or `su` _<br>(Although it will fail really clean if you don't have these. When running as root the user is still created in a regular way)_
- Some basic tools that are included in even the most minimal systems:<br>`tail`, `cat`, `sed`, `mktemp`, `chmod`, `chown`
To be sure these dependencies are not issues, I've tested this plugin in a couple of minimal environments.<br>(Docker containers of multiple distributions)`

**Note:** 
This function assumes the last argument is the username.<br>The real program allows it to be called in such a way that the last name is a group.<br>But I haven't seen people using it like this.

### DIFF
--- a/plugins/adduser/adduser.plugin.zsh
+++ b/plugins/adduser/adduser.plugin.zsh
@@ -1,63 +1,70 @@
 # Copyright (c) 2021 Nikolas Garofil
 
-adduser() {
-	local returncode
-	local result_string="Installation of 'Oh My Zsh' "
-	local temp_installscript
-	local path_installscript="$ZSH/tools/install.sh"
+_adduser_result() {
+	local result_string="Installation of 'Oh My Zsh' for '$new_user'"
 
-
-	#Create user, errors will be reported by the 'real' adduser
-	#Use which/tail combination to call the binary instead of this function
-	command adduser $@ || return 1
-
-
-	echo "\nUser '${@[$#]}' has been created. I will now try to install 'Oh My Zsh'"
-	if [[ -f $path_installscript ]] ; then
-
-		#copy install.sh to a new file in temp that we can give the right owner to execute
-		#and also make sure that after the install script we are no longer the new user
-		temp_installscript=$(mktemp)
-		cp $path_installscript $temp_installscript
-		chown ${@[$#]} $temp_installscript && chmod +x $temp_installscript
-
-		#try installing with sudo or su when not available
-		if [[ -x "$commands[sudo]" ]] ; then
-			sudo -u ${@[$#]} sh -c "$temp_installscript --unattended"
-			returncode=$?
-			if [[ -x "$commands[chsh]" ]] ; then
-				sudo chsh -s $commands[zsh] ${@[$#]}
-			else
-				echo "'chsh' is not available, change the shell manually." > /dev/stderr
-			fi
-			returncode=$?
-		else
-			if [[ -x "$commands[su]" ]] ; then
-				su -l ${@[$#]} -c "$temp_installscript --unattended"
-				returncode=$?
-				if [[ -x "$commands[chsh]" ]] ; then
-					su -c "chsh -s $commands[zsh] ${@[$#]}"
-				else
-					echo "'chsh' is not available, change the shell manually." > /dev/stderr
-				fi
-			else
-				echo "You can't become ${@[$#]} (no 'sudo' or 'su' available)" > /dev/stderr;
-				returncode=1
-			fi
-		fi
-		#cleanup
+	#Erase the temporary copy of the installscript when necessary
+	if [[ ! -z $temp_installscript  ]] ; then
 		rm $temp_installscript
-	else
-		echo "Installationscript '$path_installscript' not available" > /dev/stderr;
-		returncode=1
 	fi
 
 	#mention the result
-	if [[ $returncode -eq 0 ]]; then
+	if [[ $1 -eq 0 ]]; then
 		echo "$result_string succeeded."
 	else
 		echo "$result_string failed." > /dev/stderr
 	fi
 
-	return $returncode
+}
+
+adduser() {
+	local path_installscript="$ZSH/tools/install.sh"
+	local unattended_options=" --unattended"
+
+	local new_user=${@[$#]}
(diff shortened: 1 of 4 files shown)


plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+adduser() {
+	local returncode
+	local result_string="Installation of 'Oh My Zsh' "
+	local temp_installscript
+	local path_installscript="$ZSH/tools/install.sh"
+
+
+	#Create user, errors will be reported by the 'real' adduser
+	#Use which/tail combination to call the binary instead of this function
+	$(which -a adduser | tail -1) $@ || return 1
```
mcornella, line 12: You can call `command adduser` if that's your intention.

plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+	else
+		echo "Installationscript '$path_installscript' not available" 2> /dev/stderr;
+		returncode=1
+	fi
+
+	#mention the result
+	if [[ $returncode -eq 0 ]]; then
+		echo "$result_string succeeded."
+	else
+		echo "$result_string failed." 2> /dev/stderr
```
mcornella, line 51: The proper way to redirect to stderr is `>&2`. Fix that in the rest of the redirects.

plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+				echo "You can't become ${@[$#]} (no 'sudo' or 'su' available)" > /dev/stderr;
+				returncode=1
+			fi
+		fi
+		#cleanup
+		rm $temp_installscript
+	else
+		echo "Installationscript '$path_installscript' not available" 2> /dev/stderr;
+		returncode=1
+	fi
```
mcornella, line 45: Instead of having `else` be the error, do it like this:
```sh
if <error>; then
  <show-error>
  <return>
fi

<code-if-no-error>
```
This clarifies the logic and removes to much indentation.

plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+
+	echo "\nUser '${@[$#]}' has been created. I will now try to install 'Oh My Zsh'"
+	if [[ -f $path_installscript ]] ; then
+
+		#copy install.sh to a new file in temp that we can give the right owner to execute
+		#and also make sure that after the install script we are no longer the new user
+		temp_installscript=$(mktemp)
+		cat $path_installscript | \
+		sed 's/exec zsh -l//' |
+	        sed 's/read -r opt/opt=y; echo "\n--- This time I am answering \\"yes\\" for you, but you will still have to type in the password of that user ---"/' \
```
mcornella, line 23: The install script allows bypassing certain steps. Look at the head section of the `install.sh` script for the options you want. I think you want to stop the final `exec`, and do the `chsh`.

The first can be skipped with `RUNZSH=no`, but the second can't unless you do what you did. In that case, I'd just run the script with `--unattended`, which skips both `exec zsh` and `chsh`, and do the `chsh` part in your call to `command adduser` (you can specify the shell with `adduser --shell <shell>`.

```sh
sh install.sh --unattended
```

plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+		> $temp_installscript
+		chown ${@[$#]} $temp_installscript && chmod +x $temp_installscript
```
mcornella, line 25: You don't need to do that if you run it like I said above, using `sh install.sh`

plugins/adduser/adduser.plugin.zsh
```diff
@@ -0,0 +1,55 @@
...
+
+		#try installing with sudo or su when not available
+		if [[ -x $(which sudo) ]] ; then
```
mcornella, line 28: `sudo` might be an alias. The proper way to check for the command under `zsh` is either
```sh
if (( $+commands[sudo] )); then
```
or
```sh
if [[ -x "$commands[sudo]" ]]; then
```

mcornella (changes requested):

ngaro:
- With the exception of the indentation-suggestion everything has been fixed
- I found a way to avoid `sed` everywhere
- Changing shell no longer needs the password

**Note:** As mentioned the current structure with indentation has not been changed yet.
I still have to find the _cleanest_ way to do it, your suggested method will not work everywhere.

ngaro:
The structure has now also been fixed/improved.
I've also included some extra local vars and an extra function to adjust it easier in the future

robbyrussell:
Thanks for this, and I'm sorry about the silence. You answered Marc's review back in 2021 and nobody came back to you, which is on us.

Two things I'd want changed before this could go in. The plugin defines a function named `adduser`, so enabling it shadows the real system command, and someone typing `adduser` expecting the system tool would get this instead. It also installs Oh My Zsh into the new account and runs `sudo chsh` on it, which is a lot of side effect for a command people run for one specific purpose.

Would you be up for renaming it to something that doesn't collide, and dropping the install and chsh steps so it just wraps account creation? If you'd rather not pick this back up after this long, that's completely fine. Just say so.

ngaro:
Renaming done, but not sure what you mean with dropping the installation/changing the shell. What is the added value compared to just running `adduser` in that case ?

robbyrussell:
Thanks for turning the rename around so quickly.

Thinking about it more though, I don't think this one gets used enough to carry in the core set of plugins, and it's limited to a few distros. So I'm going to pass on it.

If anyone is interested in this, it works well as a custom plugin in `$ZSH_CUSTOM`. That's supported and survives updates.

PR #10538, opened 2021-12-29
add: git alias - git status with no untracked files displayed

fpaupier:
## Standards checklist:

## Changes:

-  add an alias for git, to easily hide untracked files.

Name of the alias: `gstnu`:

- `gst` as basis for the  `git status` command
- `nu` as a short mnemo tips for 'no untracked' 

## Other comments:

Hiding untracked files is especially conveneinent when working in a repo with lots of autogenerated files you don't want to add in your .gitignore.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -279,6 +279,7 @@ alias gsps='git show --pretty=short --show-signature'
 alias gsr='git svn rebase'
 alias gss='git status -s'
 alias gst='git status'
+alias gstnu='git status --untracked-files=no'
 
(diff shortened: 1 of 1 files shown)


robbyrussell:

The idea is reasonable, but the name doesn't work in this plugin: apart from bare `gst`, every `gst*` alias is a stash alias, and `gstu` is already `git stash --include-untracked`. So `gstnu` would sit one keystroke away from an alias about untracked files that means something quite different.

Closing for that reason. Worth knowing that @kunalg proposed the identical alias in #9908, so you weren't the only one who wanted it.

PR #10542, opened 2021-12-30
feat(dapr): add plugin for dapr

jxlwqq:
## Standards checklist:


robbyrussell:
Thanks for this, and sorry for the long delay in replying.

The thing I'd want changed first: this sources the completions on every shell start, which every user pays for in load time. We've moved to caching them instead. `plugins/gh/gh.plugin.zsh` is the pattern to follow: it guards on the command existing, writes the completion file into `$ZSH_CACHE_DIR/completions` in the background, and lets compinit pick it up on later shells.

Would you be up for reworking it that way? If you'd rather not come back to this after this long, that's completely fine. Just say so.

PR #10543, opened 2021-12-30
fix(systemd): fix the restart command

scrouthtv:
`restart` is a unit-related command, not power-related, so it should appear as `sc-restart=sudo systemctl restart` and `scu-restart=systemctl --user restart`

## Standards checklist:

## Changes:

- In #9441 I introduced a bad commit which falsely deleted the `scu-restart=systemctl --user restart` command. I moved away from oh my zsh soon after that PR so never found the issue myself. However, after it was merged, someone else notified us about the mistake. This intends PR intends to fix the issue.


mcornella:
Thanks for the speed!

PR #10565, opened 2022-01-05
Add support for alternative completions directories

felipecrs:
## Standards checklist:

## Description

This will check if a `completions-dir.txt` is present in the root of the custom plugin before adding the plugin's root to `fpath`. If the file is there, the `fpath` will be filled with `$plugin_dir/<contents of completions-dir.txt>` instead of `$plugin_dir`.

This should not cause any breaking change, as no custom plugins would already have such a file.

It will be taken into account for custom plugins. This will decrease the amount of times we check for the file, and I think it's very reasonable that the plugins in this repo (not custom ones) needs to follow the rule of having completions in their root directories.

Also, regarding my concern at duplications in `fpath` (https://github.com/ohmyzsh/ohmyzsh/issues/10412#issuecomment-967214321), there isn't much we can do about it in Oh My ZSH itself. My suggestion is that this should be handled in the plugins, like how I did in https://github.com/zsh-users/zsh-completions/pull/835.

Fixes https://github.com/ohmyzsh/ohmyzsh/issues/10412

The PR which implements this at zsh-completions is https://github.com/zsh-users/zsh-completions/pull/835.

I'm testing this at https://github.com/felipecrs/dotfiles/pull/59.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -86,17 +86,17 @@ is_plugin() {
 # Add all defined plugins to fpath. This must be done
 # before running compinit.
 for plugin ($plugins); do
-  for dir ($ZSH_CUSTOM $ZSH); do
-    if is_plugin $dir $plugin; then
-      if [ -f $dir/plugins/$plugin/completions-dir.txt ]; then
-        fpath=($dir/plugins/$plugin/$(cat $dir/plugins/$plugin/completions-dir.txt) $fpath)
-      else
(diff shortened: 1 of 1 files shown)


felipecrs:
Marked PR as ready to review as it is working for me:

image

mcornella:
I will close this as not a proper fix, with reasoning as seen in https://github.com/zsh-users/zsh-completions/pull/835#issuecomment-1036620102.

PR #10568, opened 2022-01-07
feat(git-commit-template): add plugin for git commit template

ghasemdev:
## Standards checklist:

## Changes:

- add git commit template plugin
- add limit counter for short description 
- add sign option for commits
- fix description for non git directory 

## Other comments:

To better write git commit messages, we can use template to specify the 
desired description and type of message.

All you have to do is call the `gct` command and fill in the items that 
are not optional at each step to prepare the message format.

gct

result


robbyrussell:

We added conventional commit support in `plugins/git-commit` back in 2023, so this ground is covered now. Closing since the feature exists, not because the idea was wrong.

PR #10579, opened 2022-01-12
fix(yarn): update completion

### LINKED ISSUE #10578: [yarn plugin] Yarn completion is not working accordingly
### Describe the bug

Before #10516 I was able to run completion for `scripts` without having to add the `run` command.

### Steps to reproduce

Let's say I have a script called `start-dev`. Typing `yarn s` and pressing TAB was completing `start-dev` successfully. After #10516 it does not work anymore. I need to type `yarn run s` and press TAB to make it work.

Not sure why the two snippets below aren't using the same logic:
https://github.com/ohmyzsh/ohmyzsh/blob/3a1d83d778388b574bde70f41e9674092dafe6ef/plugins/yarn/_yarn#L119
https://github.com/ohmyzsh/ohmyzsh/blob/3a1d83d778388b574bde70f41e9674092dafe6ef/plugins/yarn/_yarn#L147

But I was able to fix it locally using the same logic of line 147 at line 119, that would be:
```bash
scripts=($(cat "${packageJson}" | perl -0777 -MJSON::PP -n -E '%r=%{decode_json($_)->{scripts}}; printf "$_:$r{$_}\n" for sort keys %r'))
```

After this change, I'm able to type `yarn s` and press TAB to make `yarn start-dev` completion.

### Expected behavior

Completion should work when typing `yarn <script>` without the `run` command. 

_No response_

### OS / Linux distribution

Garuda Linux

### Zsh version

5.8

### Oh My Zsh version

master (9c84c344)

### Terminal emulator

Konsole

_No response_

### Additional context

_No response_

carlosala:
## Standards checklist:

## Changes:

Updating `yarn` completion to latest version, found [here](https://github.com/zsh-users/zsh-completions/blob/master/src/_yarn).

## Other comments:

Closes #10578

### DIFF
--- a/plugins/yarn/_yarn
+++ b/plugins/yarn/_yarn
@@ -116,7 +116,7 @@ _yarn_commands_scripts() {
   fi
 
   if [[ -n $packageJson ]]; then
-    scripts=($(cat "$packageJson" | perl -0777 -MJSON::PP -n -E '%r=decode_json($_); say for sort keys %{$r->{scripts}}'))
+    scripts=($(cat "$packageJson" | perl -0777 -MJSON::PP -n -E '$r=decode_json($_); do{($k=$_)=~s/:/\\:/g;say $k}for sort keys %{$r->{scripts}}'))
   fi
 
   _describe 'command or script' _commands -- _global_commands -- scripts -- binaries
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks Carlo!

PR #10601, opened 2022-01-19
style(init): use consistent code style in init script

carlosala:
## Standards checklist:

## Changes:

Reviewing the `oh-my-zsh.sh` file I noticed that wasn't really consistent. Sometimes quotes for variables, sometimes not, sometimes `[ ]`, others `[[ ]]`, so I tried to unify a little bit the style. By the way, this could also be done for other parts of the repo.
I will open this as a draft PR as this is kinda _breaking_ so we can discuss a little bit about it before deciding whether or not merging it.

## Other changes:

I also modified a little bit the plugins loading and theme loading. Just a refactor to have a consistent code in every plugin or theme loading.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -152,9 +152,9 @@ done
 
 # Load all of the plugins that were defined in ~/.zshrc
 for plugin ($plugins); do
-  if is_plugin "$ZSH_CUSTOM" "$plugin"; then
+  if [[ -f "$ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh" ]]; then
     source "$ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh"
-  elif is_plugin "$ZSH" "$plugin"; then
+  elif [[ -f "$ZSH/plugins/$plugin/$plugin.plugin.zsh" ]]; then
     source "$ZSH/plugins/$plugin/$plugin.plugin.zsh"
   fi
 done
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -104,7 +104,7 @@ fi
 
 # Save the location of the current completion dump file.
 if [[ -z "$ZSH_COMPDUMP" ]]; then
-  ZSH_COMPDUMP="${ZDOTDIR:-$HOME}/.zcompdump-$SHORT_HOST-$ZSH_VERSION"
+  ZSH_COMPDUMP="${ZDOTDIR:-$HOME}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
 # Construct zcompdump OMZ metadata
@@ -139,7 +139,6 @@ $zcompdump_revision
 $zcompdump_fpath
 EOF
 fi
-
 unset zcompdump_revision zcompdump_fpath zcompdump_refresh
 
 # Load all of the config files in ~/oh-my-zsh that end in .zsh
@@ -149,6 +148,7 @@ for config_file ("$ZSH"/lib/*.zsh); do
   [[ -f "$custom_config_file" ]] && config_file="$custom_config_file"
   source "$config_file"
 done
+unset custom_config_file
 
 # Load all of the plugins that were defined in ~/.zshrc
 for plugin ($plugins); do
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -57,14 +57,14 @@ mkdir -p "$ZSH_CACHE_DIR/completions"
 (( ${fpath[(Ie)"$ZSH_CACHE_DIR/completions"]} )) || fpath=("$ZSH_CACHE_DIR/completions" $fpath)
 
 # Check for updates on initial load...
-if [ "$DISABLE_AUTO_UPDATE" != "true" ]; then
-  source $ZSH/tools/check_for_upgrade.sh
+if [[ "$DISABLE_AUTO_UPDATE" != true ]]; then
+  source "$ZSH/tools/check_for_upgrade.sh"
 fi
 
 # Initializes Oh My Zsh
 
 # add a function path
-fpath=($ZSH/functions $ZSH/completions $fpath)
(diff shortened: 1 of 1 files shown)


mcornella:
Hi Carlo, the general rule is that we should quote any variables that might contain spaces (it's a good practice to quote always when using variables). Therefore `$ZSH` will need quotes, but something like the literal string `true` doesn't.

Apart from that, +1 to using `[[ ]]` everywhere (except in the very first one which is supposed to look like that in case Oh My Zsh is run in a shell different than zsh.

+1 to using `is_plugin` when checking for the location of a plugin as well.

carlosala:
Ok, I am going to correct that tomorrow and try to enforce the style you said 👍🏻

carlosala:
Done!
I also added a function `is_theme`, similar to `is_plugin`. What do you think about?

oh-my-zsh.sh
```diff
@@ -143,37 +142,45 @@ fi
...
 # Load all of the plugins that were defined in ~/.zshrc
 for plugin ($plugins); do
-  if [ -f $ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh ]; then
-    source $ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh
-  elif [ -f $ZSH/plugins/$plugin/$plugin.plugin.zsh ]; then
-    source $ZSH/plugins/$plugin/$plugin.plugin.zsh
+  if is_plugin "$ZSH_CUSTOM" "$plugin"; then
+    source "$ZSH_CUSTOM/plugins/$plugin/$plugin.plugin.zsh"
+  elif is_plugin "$ZSH" "$plugin"; then
+    source "$ZSH/plugins/$plugin/$plugin.plugin.zsh"
```
mcornella, line 158: Now that I look more carefully, this is not totally correct because `is_plugin` will return correctly if a `_$plugin` file exists but not the `$plugin.plugin.zsh`, so this will break on plugins that only provide completion files.

It should only be used in the `fpath` change logic above.

carlosala: uff I didn't caught that. I'll revise it!

carlosala: I think that what you say is not correct Marc.
```zsh
is_plugin() {
  local base_dir=$1
  local name=$2
  builtin test -f $base_dir/plugins/$name/$name.plugin.zsh \
    || builtin test -f $base_dir/plugins/$name/_$name
}
```
This will return correctly in both cases. Maybe we should modify the logic here

mcornella: That would change how we add plugins to `$fpath`.

If we only return true if `$name.plugin.zsh` exists, plugins without these files (like most completion plugins) will not get added to `$fpath` and break.
If we only return true if `_$name` exists, plugins without `_$name` like the `genpass` plugin will not work correctly if they're not added to `$fpath`.

Also the proposed change does not decouple the path to be sourced from the check if the file exists, so it moves the logic to two places (`is_plugin` and the `source` line).

We could turn this into a `for` loop but given that we only need to check 2 locations that seems wasteful. It's better to leave it as is.

mcornella: Yeah that's the problem. We could also have a function check if the plugin file exists and loads it like you say, but we'd only be moving the logic somewhere else.

carlosala:
i think it's ready to merge!

oh-my-zsh.sh
```diff
@@ -98,14 +97,14 @@ done
...
 else
-  SHORT_HOST=${HOST/.*/}
+  SHORT_HOST="${HOST/.*/}"
 fi
 
 # Save the location of the current completion dump file.
-if [ -z "$ZSH_COMPDUMP" ]; then
-  ZSH_COMPDUMP="${ZDOTDIR:-${HOME}}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
+if [[ -z "$ZSH_COMPDUMP" ]]; then
+  ZSH_COMPDUMP="${ZDOTDIR:-$HOME}/.zcompdump-$SHORT_HOST-$ZSH_VERSION"
```
mcornella, line 107: nitpick: this doesn't fail, but in the future it'd be better to keep the braces (`{}`) when using - because it [has a special meaning when using braces](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion). This doesn't fail, again, but it makes the code clearer.

oh-my-zsh.sh
```diff
@@ -143,37 +142,45 @@ fi
...
 # TIP: Add files you don't want in git to .gitignore
-for config_file ($ZSH/lib/*.zsh); do
-  custom_config_file="${ZSH_CUSTOM}/lib/${config_file:t}"
-  [ -f "${custom_config_file}" ] && config_file=${custom_config_file}
-  source $config_file
+for config_file ("$ZSH"/lib/*.zsh); do
+  custom_config_file="$ZSH_CUSTOM/lib/${config_file:t}"
+  [[ -f "$custom_config_file" ]] && config_file="$custom_config_file"
+  source "$config_file"
 done
```
mcornella, line 152: Can we also `unset` `custom_config_file`?

mcornella (approved):
Generally perfect 👍🏻

oh-my-zsh.sh
carlosala: I pushed the changes. I understand that you are talking about just the braces from `${SHORT_HOST}` and `${ZSH_VERSION}`, or also from `$HOME`?

oh-my-zsh.sh
carlosala: sure, done!

oh-my-zsh.sh
mcornella: Yeah you got that right. In `$HOME` it's redundant.

carlosala: great, then it's resolved!
I'll have to look up that docs!

mcornella:
Thanks Carlo! 😄

carlosala:
> Thanks Carlo! smile

As always, you're welcome😉

PR #10602, opened 2022-01-20
fix slow problem with WSL2 ubuntu of git prompt

bang-eoh:
## Standards checklist:

## Changes:

- On Ubuntu run with WSL2, when go to a git folder that checkout from Window host machine, zsh prompt is running extremely slow
- Detect if running in WSL2 and execute git.exe instead

## Other comments:
- No

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -6,7 +6,11 @@
 # We wrap in a local function instead of exporting the variable directly in
 # order to avoid interfering with manually-run git commands by the user.
 function __git_prompt_git() {
-  GIT_OPTIONAL_LOCKS=0 command git "$@"
+  if $(pwd -P | grep -q "^\/mnt\/c\/*"); then
+    GIT_OPTIONAL_LOCKS=0 command git.exe "$@"
+  else
+    GIT_OPTIONAL_LOCKS=0 command git "$@"
+  fi
 }
 
 function git_prompt_info() {


bang-eoh:
@rwe please help review this

rwe:
Hi @bang-eoh — unfortunately I don't use WSL or WSL2 and so cannot speak much to the method of determining that environment. I have some comments, notes, and I hope they're helpful.

Couldn't this be solved by setting your own `PATH` to resolve to the correct git executable, instead? Or export `GIT_EXEC_PATH`?

  * The hard-coded `/mnt/c/…` path is a little alarming. Wouldn't this also trigger for people running Linux directly, with a mounted Windows partition? They could not run `git.exe`, but this would forcibly attempt that. And wouldn't it fail to trigger for people with, e.g. a `D:\` data drive in the same situation as you? This seems like such specific heuristic. There are many other `git` invocations in `ohmyzsh`, but they don't do anything similar.

  * In general, if you need to check a path, it would be better to compare with a glob, not `grep`. For example:
    ```zsh
      if [[ "${PWD:A}" == /mnt/c/* ]]; then
          …
     ````

  * Note that regex syntax is different than globs:
    * The backslashes in that regex are redundant, because forward-slashes don't require escaping.
    * The asterisk in the regex `/*` means "with zero or more forward slashes", _not_ "followed by anything".

PR #10604, opened 2022-01-20
perf(pipenv) - Add lazy-pipenv option

kajbaf:
## Standards checklist:

## Changes:

- Using `PIPENV_LAZY_LOAD` variable to enable/disable automatic loading of PIPENV environments. 

## Other comments:

Cases that such a behaviour is desired is:
- Those who rely on `direnv` for enabling venv
- The projects that have both `Pipfile` and `poetry.lock` for whatever reason
- In case you don't want slow down navigating folders, and `pipenv` aliases are enough for you.


mcornella:
Similar option introduced in 079dbff2.

PR #10619, opened 2022-01-25
feat(conda-env): add plugin for conda prompt display

cjeonguk:
## Standards checklist:

## Changes:

- Adds a `conda-env` plugin which provides `conda_propmt_info` function that we can use in our theme, which displays the basename of the current `$CONDA_DEFAULT_ENV`.

## Other comments:

- `conda-env` is a popular package manager for python.
- `conda-env` plugin works similarly to the `virtualenv` plugin, but for [`conda`](https://anaconda.com).


afeld:
Duplicate of https://github.com/ohmyzsh/ohmyzsh/pull/9691.

carlosala:
It's not really a duplicate I think, as they implement different things.

cjeonguk:
It's the same name, but #9691 is a plugin with aliases and mine is a plugin with prompt display function. I think another plugin name would be better. `conda` has a wide range to explain its function.

naturedamends:
Should all plugins use your function instead of https://github.com/ohmyzsh/ohmyzsh/pull/11785

mcornella:
Took a while, but thanks for the contribution!

skupr-anaconda:
If anyone asks how to use it then do the next:
Add `conda-env` to plugins in your `~/.zshrc`:
`plugins=(... conda-env)`
run `source ~/.zshrc.`
Then if you run :
`╰─❯ conda_prompt_info`
`[abs]`
It will say the name of your current environment.

PR #10637, opened 2022-01-29
feat(1password): add command to copy username

gheritarish:
## Standards checklist:

- [ xThe PR title is descriptive.

## Changes:

- This commit adds the `opuser` command, to copy the username of a
service into the clipboard
- I also renamed the `_opswd` function used for completion to `_opcomplete`
- The README was updated to take those changes into account

## Other comments:

This command might be less used than the existing `opswd` command but the script I had
on my computer before this plugin did both, so I thought it might help other people


mcornella:
@r-medina would you think this is the way to go or maybe we can add a pause after each copy to `opswd`. Something like
```console
$ opswd github.com
✔ username for github.com copied to clipboard. Press Enter to continue
✔ password for github.com copied to clipboard. Press Enter to continue
✔ TOTP for github.com copied to clipboard
```

cc-ing other users that participated in the original PR: @seanb4t @bartekpacia @ianchesal.

ianchesal:
@mcornella that's a super cool idea

gheritarish:
Love the idea!
I'll see if I can implement it when I have time (if someone else want to do it & has time, that's totally fine)

gheritarish:
So I updated it with my work account, but this works with y/N interaction from the user, which is pretty much the behavior we wanted!

PR #10639, opened 2022-01-29
feat(rbw): add `rbwpw` function to copy passwords to the clipboard

carlosala:
## Standards checklist:

## Changes:

Adding `rbwpw` function, very similar to `opwsd` (or something like that 😅) in `1password` plugin. It copies the password for a service and clears the clipboard after 20s.

### DIFF
--- a/plugins/rbw/README.md
+++ b/plugins/rbw/README.md
@@ -9,4 +9,11 @@ To use it, add `rbw` to the plugins array in your zshrc file:
 plugins=(... rbw)
 ```
 
+It also adds the `rbwpw` function, that copies the password in the clipboard
+for the service you ask for and clears the clipboard 20s later. Usage:
+
+```zsh
+rbwpw <service>
+```
+
 This plugin does not add any aliases.
--- a/plugins/rbw/README.md
+++ b/plugins/rbw/README.md
@@ -9,8 +9,11 @@ To use it, add `rbw` to the plugins array in your zshrc file:
 plugins=(... rbw)
 ```
 
-It also adds the `rbwpw` function, that copies the password in the clipboard
-for the service you ask for and clears the clipboard 20s later. Usage:
+## `rbwpw`
+
+The `rbwpw` function is a wrapper around `rbw`. It copies the password in the
+clipboard for the service you ask for and clears the clipboard 20s later.
+The usage is as follows:
 
 ```zsh
 rbwpw <service>
--- a/plugins/rbw/rbw.plugin.zsh
+++ b/plugins/rbw/rbw.plugin.zsh
@@ -17,3 +17,23 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_rbw" ]]; then
 fi
 
 rbw gen-completions zsh >| "$ZSH_CACHE_DIR/completions/_rbw" &|
+
+# rbwpw function copies the password of a service to the clipboard
+# and clears it after 20 seconds
+function rbwpw {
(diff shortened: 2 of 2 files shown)


plugins/rbw/rbw.plugin.zsh
```diff
@@ -17,3 +17,23 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_rbw" ]]; then
...
+  local pw=$(rbw get $service 2>/dev/null)
+  [[ -z $pw ]] && echo "$service not found" && return 1
+  echo -n $pw | clipcopy
+  echo "password for $service copied!"
+  (sleep 20 && clipcopy </dev/null 2>/dev/null) &|
+}
+
+function _rbwpw {
+  local -a services=("${(@f)$(rbw ls 2>/dev/null)}")
+  [[ -n services ]] && compadd -a -- services
```
mcornella, line 36:
```suggestion
  [[ -n "$services" ]] && compadd -a -- services
```

plugins/rbw/rbw.plugin.zsh
```diff
@@ -17,3 +17,23 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_rbw" ]]; then
...
+  rbw unlock || (echo "rbw is locked" && return 1)
+  local pw=$(rbw get $service 2>/dev/null)
+  [[ -z $pw ]] && echo "$service not found" && return 1
+  echo -n $pw | clipcopy
+  echo "password for $service copied!"
+  (sleep 20 && clipcopy </dev/null 2>/dev/null) &|
+}
+
+function _rbwpw {
+  local -a services=("${(@f)$(rbw ls 2>/dev/null)}")
```
mcornella, line 35: Inline declaration and assigning of an array breaks in older zsh versions (can't remember at which point it starts failing). You should do this in two lines:
```suggestion
  local -a services
  services=("${(@f)$(rbw ls 2>/dev/null)}")
```

plugins/rbw/rbw.plugin.zsh
```diff
@@ -17,3 +17,23 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_rbw" ]]; then
...
+function rbwpw {
+  [[ $# -ne 1 ]] && echo "usage: rbwpw <service>" && return 1
+  local service=$1
+  rbw unlock || (echo "rbw is locked" && return 1)
+  local pw=$(rbw get $service 2>/dev/null)
+  [[ -z $pw ]] && echo "$service not found" && return 1
+  echo -n $pw | clipcopy
+  echo "password for $service copied!"
+  (sleep 20 && clipcopy </dev/null 2>/dev/null) &|
+}
```
mcornella, line 32: This feels hacky, can you use full `if` sentences instead of chaining `&&`?

Also, to group two or more commands use `{ <command-list> }`. `(<command-list>)` creates a subshell.

plugins/rbw/README.md
```diff
@@ -9,4 +9,11 @@ To use it, add `rbw` to the plugins array in your zshrc file:
 plugins=(... rbw)
 ```
 
+It also adds the `rbwpw` function, that copies the password in the clipboard
+for the service you ask for and clears the clipboard 20s later. Usage:
+
+```zsh
+rbwpw <service>
+```
+
```
mcornella, line 18: This would go better as a separate section with a title.

mcornella (changes requested):

mcornella (approved):

PR #10668, opened 2022-02-08
Add git -C support to git_prompt_info.

obszczymucha:
## Standards checklist:

## Changes:

- Add git `-C` support to `git_prompt_info` function in `lib/git.zsh`.

## Other comments:
All git commands can be used with the `-C` option that allows working on a repository located in a directory different than the current one. This PR extends `git_prompt_info` function to accept an optional directory parameter.

One example use case for this would be a theme wanting to customise git prompt for a specific directory and display git prompt for a repository in a different location, e.g.:

```my-theme.zsh
function my_git_prompt_info() {
  if [[ $PWD == $HOME ]]; then
    echo $(git_prompt_info ".dotfiles")
  else
    echo $(git_prompt_info)
  fi
}

PROMPT=' %F{white}%2c%F{cyan} [%f '
RPROMPT='$(my_git_prompt_info) %F{cyan}] %F{lightblue}%D{%H:%M}'
```

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -10,34 +10,56 @@ function __git_prompt_git() {
 }
 
 function git_prompt_info() {
+  readonly dir="$1"
+
+  # Decorates git command if a directory is passed to us.
+  function __git_dir_decorator() {
+    if [[ -n $dir ]]; then
+      __git_prompt_git -C "$dir" "$@"
+    else
+      __git_prompt_git "$@"
+    fi
+  }
+
   # If we are on a folder not tracked by git, get out.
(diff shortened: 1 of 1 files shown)


robbyrussell:

The prompt has changed underneath it since 2022. `git_prompt_info` is now an async wrapper that echoes cached output and ignores its arguments, so a directory argument can't reach the code that would use it. Making it work would mean designing per-directory handling into the async layer, which is a bigger change than this one.

Closing for that reason rather than because the idea was wrong. For a dotfiles repo you're watching from `$HOME`, a small function in `$ZSH_CUSTOM` calling `git -C` directly is the simpler path.

PR #10672, opened 2022-02-09
fix(init): check for unsafe directories in `fpath`

carlosala:
## Standards checklist:

## Changes:

Actual `compinit`'s approach was loading completion functions from _insecure_ directories (those categorized as insecure by `compaudit`) even though not having `$ZSH_DISABLE_COMPFIX` set to `true`. This PR's approach should take into account those insecure directories.

Without the `-C` flag in `compinit` call, `.zcompdump` file could potentially change if there are changes in completion functions, so I also added a test to see if this happened and append `omz` metadata to `.zcompdump` file.

Now the function `handle_completion_insecurities` will run in background to speed up a little bit the init time.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -120,17 +120,18 @@ fi
 
 if [[ "$ZSH_DISABLE_COMPFIX" != true ]]; then
   source "$ZSH/lib/compfix.zsh"
-  # If completion insecurities exist, warn the user
-  handle_completion_insecurities
   # Load only from secure directories
-  compinit -i -C -d "$ZSH_COMPDUMP"
+  compinit -i -d "$ZSH_COMPDUMP"
+  # If completion insecurities exist, warn the user
+  handle_completion_insecurities &|
 else
   # If the user wants it, load from all found directories
-  compinit -u -C -d "$ZSH_COMPDUMP"
+  compinit -u -d "$ZSH_COMPDUMP"
 fi
 
 # Append zcompdump metadata if missing
-if (( $zcompdump_refresh )); then
+if (( $zcompdump_refresh )) \
+  || ! command grep -q -Fx "$zcompdump_revision" "$ZSH_COMPDUMP" 2>/dev/null; then
   # Use `tee` in case the $ZSH_COMPDUMP filename is invalid, to silence the error
   # See https://github.com/ohmyzsh/ohmyzsh/commit/dd1a7269#commitcomment-39003489
   tee -a "$ZSH_COMPDUMP" &>/dev/null <<EOF


oh-my-zsh.sh
```diff
@@ -120,17 +120,18 @@ fi
 
 if [[ "$ZSH_DISABLE_COMPFIX" != true ]]; then
   source "$ZSH/lib/compfix.zsh"
-  # If completion insecurities exist, warn the user
-  handle_completion_insecurities
   # Load only from secure directories
-  compinit -i -C -d "$ZSH_COMPDUMP"
+  compinit -i -d "$ZSH_COMPDUMP"
+  # If completion insecurities exist, warn the user
+  handle_completion_insecurities &|
```
BuonOmo, line 126: hey, do you know where I can find a list of those very useful keywords ? (tried some googling as well as `man zshall`, haven't found a decent list :/)

carlosala: Which keywords are you exactly talking about?

BuonOmo: the `&|`, and I've today also discovered `&!`

carlosala: I always find that information in `man zshall`!

BuonOmo: That's weird.. On my mac somehow it's not there (`man zshall | rg --fixed-strings '&!'`)

Found it under `man zshbuiltins` though!

Thanks for your time and that PR 🙂

fice-t:
This looks good, but is there a reason to remove the `-C` flag in the `compinit -u` case? If you remove it, then even in that case `compaudit` is called, which I'd think is unnecessary.

If not, then the `grep` call you added could be moved into the `compinit -i` section and used to set `zcompdump_refresh`.

Lastly, since this `handle_completion_insecurities` is running in the background, would it make sense to actually run `handle_completion_insecurities` in `ZSH_DISABLE_COMPFIX == True` case as well? Since AFAIU the reason to disable this is a runtime penalty, but background-processing largely eliminates that.

Perhaps a third option for `ZSH_DISABLE_COMPFIX` could be added and enabled by default, which does `compinit -C` _and_ a background `handle_completion_insecurities`.

carlosala:
Hi @fice-t!

As stated [here](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Use-of-compinit), using the `-u` flag in `compinit` will not run `compaudit` to check for potential security issues in completion files. The reason to remove `-C` flag is that, in case of having it, it will blindly follow `.zcompdump` file, without checking if there are new completion functions in present directories at `$fpath`.

Secondly, that `grep` call could not be moved to `compinit -i` part because in case there are new completion functions, also `compinit -u` will update the `.zcompdump` file and potentially remove the `#omz` part. I didn't knew that it would be faster to check if the whole line is present, but it makes sense as if the first character doesn't match it can pass to the next line. Thanks for the suggestion, i'll change that for sure!

The reason to not run always `handle_completion_insecurities` are mainly two. The first one is what you say, startup time, and this PR solves that basically because runs in background, without stopping the sourcing and execution of other parts of zsh startup. The second one is that many users reported that with MacOS it detects some problems related to `root` user, because there is an `admin` user that has the same function as `root` but its name is different, and a safe directory is detected as non-safe (I don't really know that, as I'm a Linux user).

And lastly, with this PR people is not going to have problems nor slow startup using the default approach (without enabling `ZSH_DISABLE_COMPFIX`). This alternative approach will only be useful for those MacOS users that have been reporting issues.

Thanks for your feedback, I really appreciate that! By the way, what do you think about what I'm saying, is everything right, do you think I can improve something?

PR #10675, opened 2022-02-09
fix(kubectx): don't error on missing k8s context

SamLR:
Previously if there isn't a current kubectl context every time the plugin runs (e.g. if in a prompt) it would display `error: current-context is not set`.

## Standards checklist:

## Changes:

- Add `$KUBECTX_UNSET` variable to configure what's displayed when no kubectl context is set
- Removes previous behaviour of displaying the error.

### DIFF
--- a/plugins/kubectx/kubectx.plugin.zsh
+++ b/plugins/kubectx/kubectx.plugin.zsh
@@ -3,7 +3,9 @@ typeset -g -A kubectx_mapping
 function kubectx_prompt_info() {
   (( $+commands[kubectl] )) || return
 
-  local current_ctx=$(kubectl config current-context)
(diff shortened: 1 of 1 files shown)


carlosala:
Nice idea! By the way, I think would be better to just print the info if any context is set and don't show anything if no context is set. I would remove `$KUBECTX_UNSET` and the `echo` part. What do you think about?

mcornella:
I agree with Carlo, on empty information the common behavior in prompt functions is to not show anything. You can exit the function with a check for empty content:
```zsh
[[ -n "$current_ctx" ]] || return
```

SamLR:
Does that look better @mcornella, @carlosala ?

carlosala (approved):
now it's great!

PR #10685, opened 2022-02-11
feat(toolbox): add plugin to show `toolbox` prompt info

### LINKED ISSUE #10446: Add prompt support for toolbx

_No response_

[toolbx](https://github.com/containers/toolbox) is appending icon indicator to bash shell to indicate when your shell is running in container. When using oh-my-zsh this indicator is lost on every theme so there is no easy way to distinguish if you're running inside toolbx container or not.

Make use of `/run/.toolboxenv` file to determine if you're running inside toolbx shell and append toolbx ⬢ indicator to the beginning of PS1 independently of theme.
Allow to opt-out using flag in .zshrc.

Modifying theme manually.

### Additional context

_No response_

### Related Issues

_No response_

carlosala:
## Standards checklist:

## Changes:

Adds `toolbox_prompt_info` function.
Closes #10446

### DIFF
--- a/plugins/toolbox/README.md
+++ b/plugins/toolbox/README.md
@@ -1,0 +1,19 @@
+# toolbox plugin
+
+Plugin for [toolbox](https://containertoolbx.org), a tool to use containerized CLI environments.
+
+To use it, add `toolbox` to your plugins array in your `.zshrc` file:
+
+```zsh
+plugins=(... toolbox)
+```
+
+## Prompt function
+
+This plugins adds `toolbox_prompt_info()` function. Using it in your prompt, it will show the toolbox indicator ⬢ (if you are running in a toolbox container), and nothing if not.
+
+You can use it by adding `$(toolbox_prompt_info)` to your `PROMPT` or `RPROMPT` variable:
+
+```zsh
+RPROMPT='$(toolbox_prompt_info)'
+```
--- a/plugins/toolbox/kubectx.plugin.zsh
+++ b/plugins/toolbox/kubectx.plugin.zsh
@@ -1,0 +1,3 @@
+function toolbox_prompt_info() {
+  [[ -f /run/.toolboxenv ]] && echo "⬢"
+}


plugins/toolbox/README.md
```diff
@@ -0,0 +1,9 @@
+# toolbox
+
+This plugins adds `toolbox_prompt_info()` function. It shows the toolbox indicator ⬢ if you are running in a container and nothing if not.
```
mcornella, line 3: The first description line should avoid going into specifics. A better tagline would be "a prompt function for toolbox (insert link to project here)".

plugins/toolbox/README.md
```diff
@@ -0,0 +1,9 @@
...
+
```
mcornella, line 4: Needs the usage standard info (to use it, add it ...).

plugins/toolbox/README.md
```diff
@@ -0,0 +1,9 @@
...
+It can be useful to customize themes. For example:
```
mcornella, line 5: Hmm I see you took the advice of using `kubectx` plugin as reference. That one's README should be fixed 😁 I like something more like the terraform plugin which is a little bit more verbose.

mcornella:
@JayDoubleu would something like this work for you?

JayDoubleu:
Hey @mcornella , apologies, I somehow managed to miss the notification. This works looks great!. I didn't know it would be so simple to implement ! 

I even prefer the indicator to be on the right side. Thank you.

Out of curiosity, is it possible to move it around, for example to the left side ?

mcornella:
You could put the function in `$PS1` instead, in the position you prefer between the different segments of the prompt. If you want it all the way to the left, granted that your prompt is not multiple lines, you could just use
```sh
PS1='$(toolbox_prompt_info)'"$PS1"
```

We could document these two cases as the probable ones to be used by users.

JayDoubleu:
I also recently found out tha there is `/run/.containerenv` available for toolbox users. Which returns bunch of metadata about the container. Perhaps an option to display the toolbox name alongside the icon would be a good idea `eval $(cat /run/.containerenv) && echo $name` .

I'll have a look at this tomorrow as it got me really excited ;)

carlosala:
Now it should be better. Sorry for the shitty PR Marc ;)

mcornella:
> I also recently found out tha there is `/run/.containerenv` available for toolbox users. Which returns bunch of metadata about the container. Perhaps an option to display the toolbox name alongside the icon would be a good idea `eval $(cat /run/.containerenv) && echo $name` .
> 
> I'll have a look at this tomorrow as it got me really excited ;)

Can you provide a sample .containerenv to see the format?

mcornella (approved):

JayDoubleu:
> > I also recently found out tha there is `/run/.containerenv` available for toolbox users. Which returns bunch of metadata about the container. Perhaps an option to display the toolbox name alongside the icon would be a good idea `eval $(cat /run/.containerenv) && echo $name` .
> > I'll have a look at this tomorrow as it got me really excited ;)
> 
> Can you provide a sample .containerenv to see the format?

Sure thing it's 

```
cat /run/.containerenv 
engine="podman-3.4.4"
name="fedora-toolbox-35"
id="362b87172756af906fd8e4bd144fb6311baa0093ebbed7bbe7afdd6a9a851c41"
image="registry.fedoraproject.org/fedora-toolbox:35"
imageid="40b181c70b7342032df0b13c16beaf18a3ea7f73617858923c3972a02139d6e8"
rootless=1
```

Where name is the toolbox container name.

carlosala:
I think we should merge this changes and open a new PR for more changes.

PR #10688, opened 2022-02-11
feat(fd): update `fd` completion to v8.3.2

carlosala:
## Standards checklist:

## Changes:

One more!


mcornella:
Thanks Carlo!

In the future we'd love a GitHub Action that automatically detects new versions of files and opens a PR so that we can review any changes.

A GitHub Action of this sort would need to be able to configure the source (repository + path) and destination files per plugin, as well as when to consider an update (on new release or on new commit).

carlosala:
I was also thinking that some days. Let's look further into it!

PR #10691, opened 2022-02-13
Add zig plugin

kassane:
## Standards checklist:

## Changes:

- add zig alias commands

### DIFF
--- a/plugins/zig/README.md
+++ b/plugins/zig/README.md
@@ -7,19 +7,19 @@ plugins=(... zig)
 ```
 # Alias
 
-| Alias                 | Command                                          | Description                |
-|:----------------------|:-------------------------------------------------|-----------------------------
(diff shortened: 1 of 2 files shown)


kassane:
Enhanced zsh-completion for zig:
https://github.com/ziglang/shell-completions

robbyrussell:
Thanks for this, and sorry it waited so long for a reply. Zig is well worth having a plugin for.

It needs a refresh before it can go in. `zbuildexe` and `zbuildlib` call `zig init-exe` and `zig init-lib`, which zig replaced with plain `zig init`, and there's nothing for `zig fetch` or the package manager. It would also want a `(( $+commands[zig] ))` guard at the top, since right now it exports `ZCC` and `ZCXX` even on machines without zig installed.

Would you be up for bringing it up to date against current zig? Completions from ziglang/shell-completions would be a nice addition too. If you'd rather not pick this back up after this long, that's completely fine. Just say so.

PR #10703, opened 2022-02-17
change to GPLV3 license

libererer:


carlosala:
Why should license change? I don't see your point.

PR #10707, opened 2022-02-18
feat(pre-commit) add aliases

hofbi:
## Standards checklist:

## Changes:

- Adding a new plugin with common aliases for [pre-commit](https://pre-commit.com/) 

## Other comments:

Maintainers: @hofbi @cbachhuber 

...

### DIFF
--- a/plugins/pre-commit/pre-commit.plugin.zsh
+++ b/plugins/pre-commit/pre-commit.plugin.zsh
@@ -1,0 +1,8 @@
+# Aliases for pre-commit
+alias prc='pre-commit'
+
+alias prcau='pre-commit autoupdate'
+
+alias prcr='pre-commit run'
+alias prcra='pre-commit run --all-files'
+alias prcrf='pre-commit run --files'
--- a/plugins/pre-commit/README.md
+++ b/plugins/pre-commit/README.md
@@ -1,0 +1,19 @@
+# Pre-commit plugin
+
+This plugin adds aliases for common commands of [pre-commit](https://pre-commit.com/).
+
+To use this plugin, add it to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... pre-commit)
+```
+
+## Aliases
+
+| Alias   | Command                                | Description                                            |
+| ------- | -------------------------------------- | ------------------------------------------------------ |
+| prc     | `pre-commit`                           | The `pre-commit` command                               |
+| prcau   | `pre-commit autoupdate`                | Update hooks automatically                             |
+| prcr    | `pre-commit run`                       | The `pre-commit run` command                           |
+| prcra   | `pre-commit run --all-files`           | Run pre-commit hooks on all files                      |
+| prcrf   | `pre-commit run --files`               | Run pre-commit hooks on a given list of files          |


cbachhuber (commented):
Nice, I want this on my machine 👍

plugins/pre-commit/pre-commit.plugin.zsh
```diff
@@ -0,0 +1,11 @@
+# Aliases for pre-commit
+alias pc='pre-commit'
```
LunarWatcher, line 2: Conflicts with https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/macports/README.md

LunarWatcher (commented):
There's a conflicting alias; might be worth renaming the whole range for consistency, but I have no suggestions on alternative aliases/prefixes.

LunarWatcher (changes requested):
See my previous review (re-submitting with request changes now that it's re-enabled): `pc` conflicts

plugins/pre-commit/pre-commit.plugin.zsh
hofbi: Thanks for the hint. I will think about an alternative base alias for pre-commit.

In general, are all plugin combinations collision free? And is there a CI job that validates that?

LunarWatcher: AFAIK, yes. And no, there aren't any automatic systems for avoiding collisions.

hofbi: @cbachhuber Any suggestion for a good alias here?

cbachhuber: Good question, how about

- `pct`
- `pcmt`
- `prcmt`

I don't see aliases in the repo for any of these options, and no similarly named Ubuntu package stands out to me. Do you like any of these?

hofbi: I also thought about `pct` because of brevity.

But should we use this base for all other aliases as well or stick to them as they are. Changing them would be a bit more consistent but also more to type. 

I would go for only changing `pc` to `pct`

cbachhuber: I'm torn - I would like to avoid this inconsistency in the aliases. Thus, I'd prefer using `pct` as base for all aliases. @LunarWatcher, what do you think?

LunarWatcher: `pct` sounds good to me. I agree with using it as a prefix for consistency; this should also let us avoid conflicts in the future. 

Alternatively, there's `prc` (**pr**e-**c**ommit), though I haven't properly checked it for package conflicts. The second advantage with `prc`, beyond using the early characters, is that there is 0 overlap with existing commands, beyond the `p`-based ones. There's nothing that starts on `pr` at the moment, though.

But I'm fine with either of these; pick whichever you feel is more natural

hofbi: I updated the aliases consistently to use the base `prc`.

cbachhuber (approved):
LGTM

LunarWatcher (approved):

hofbi:
@carlosala Is there anything that prevents this from beeing merged?

carlosala (changes requested):
Following the [contributing guidelines](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases) I'd remove some aliases, maybe letting `prc`, `prcau`, `prci` and `prcr`. What do you think about?

cbachhuber:
Sounds reasonable, however I personally use `prc`, `prcr`, `prcra`, `prcrf` much more frequently, so I vote for these.

carlosala:
Those are great! It's just to not pollute a lot aliases namespace with a bunch of stuff that you use time to time. @hofbi what do you think about?

hofbi:
I also use the autoupdate in addition to the 4 mentioned by @cbachhuber. If all of you agree I would then just drop the two install aliases. These can be expected to be rarely used.

hofbi:
Since we are talking about integrating aliases for pre-commit, what do you think about using pre-commit for this repo?

carlosala:
Thanks for your contribution! Even though, I don't think pre-commit hooks make sense in omz, we don't have much to check... Maybe commit messages, if they follow conventional commits, but that's actually not needed because everytime we merge we just squash everything and change the title.
We can think about that, thanks for the idea as well! 😉

hofbi:
I could already some good points for pre-commit. Can be just a few basic ones such as https://github.com/pre-commit/pre-commit-hooks that just clean trailing whitespaces or ensure newlines at the end of the file. But from there we could integrate more such as Python a formatter and checks, since the alias collision I am working on in #10859 is written in Python. I could also imagine a markdown formatter or a shellcheck. Adding some or all of them over time could improve the consistency of the codebase and support the maintenance. What do you think? We could also open a new issue to discuss this.

cbachhuber:
Addition: on @hofbi's recommendation, we also recently introduced the [typos](https://github.com/crate-ci/typos/blob/master/docs/pre-commit.md) hook in our work repo, and that's amazing!

PR #10714, opened 2022-02-22
feature plugin copyfilename

t-oster:
## Standards checklist:

## Changes:

- add plugin copyfilename similar to copyfile, but copy the absolute path of a file to the clipboard instead of its contents.

## Other comments:

- I will read the contribution guideline as soon as I have time, however, since I copied an existing one-liner plugin and just changed a bit, I guess the code follows the style


mcornella:
Closed in favor of #7569.

t-oster:
my bad. Thanks

PR #10721, opened 2022-02-24
fix(updater): various update check fixes

mcornella:
## Standards checklist:

## Changes:

- Add a timeout of 2 seconds on latest commit check. This helps when not having a stable connection, in which case the update check will assume no updates are available.

- Properly check whether the latest commit is present in the local repository. Previously, only the local HEAD was compared to the remote HEAD, and if it didn't match it was assumed that the remote HEAD was newer, so there were updates. This is not true if the local repository has newer commits on top of the remote HEAD, so this change checks for that using `git merge-base`.

## Other comments:

If merged, this PR should use the rebase strategy as both commits should appear separately in the changelog.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -65,11 +65,11 @@ function is_update_available() {
   local remote_head
   remote_head=$(
     if (( ${+commands[curl]} )); then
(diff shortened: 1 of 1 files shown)


carlosala (approved):
I like that!
I also tested it and works in the most common cases that I can imagin.
In my opinion, mergeable (or rebasable)!

PR #10733, opened 2022-02-25
fix permission error

lrn1380:
## Standards checklist:

## Changes:

- [...]

## Other comments:

problem


carlosala:
This change breaks a basic functionality in OMZ, which is be able to define the directory where you want to install it.

carlosala:
This change breaks a basic functionality in OMZ, which is be able to define the directory where you want to install it.

mcornella:
I'd like to know why you needed this. Does `$ZSH` point to `/root/.oh-my-zsh` when installing with a non-root user?

PR #10741, opened 2022-02-27
fix: this fix the installation in windows WSL to create ohmyzsh folder

omaryahir:
## Standards checklist:

## Changes:

- This is just fixing the creation of the folder in windows subsystem

## Other comments:

- I had the issue, and I just edit install.sh file with this change, and it solved the problem
...

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -272,7 +272,7 @@ setup_ohmyzsh() {
   fi
 
   # Manual clone with git config options to support git < v1.7.2
-  git init --quiet "$ZSH" && cd "$ZSH" \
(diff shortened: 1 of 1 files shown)


carlosala:
This should not be a problem for you, since `git init <folder>` has been on git for a long time.
Could you instead of this PR open an issue following the template to try to understand what's going on?
Thanks!

omaryahir:
> This should not be a problem for you, since `git init <folder>` has been on git for a long time. Could you instead of this PR open an issue following the template to try to understand what's going on? Thanks!

Thank you so much for taking a look, I figured out at the end the issue, it seems that the problem was the way I installed git in my wsl subsystem.

PR #10743, opened 2022-02-27
fix(rails): add bin/dev to _rails_command

kjellberg:
## Standards checklist:

## Changes:

- Since version 7.0.0, Rails now adds a `bin/dev` entrypoint for starting Rails and development dependencies with foreman.

## Other comments:
This also solves #9600.

### DIFF
--- a/plugins/rails/rails.plugin.zsh
+++ b/plugins/rails/rails.plugin.zsh
@@ -1,5 +1,7 @@
 function _rails_command () {
-  if [ -e "bin/stubs/rails" ]; then
+  if [ -e "bin/dev" ]; then
+    bin/dev $@
+  elif [ -e "bin/stubs/rails" ]; then
     bin/stubs/rails $@
   elif [ -e "bin/rails" ]; then
     bin/rails $@


carlosala (approved):

mcornella:
Can you document a little bit more about how `bin/dev` interfaces with other `rails` commands (i.e. if running `bin/dev console`, how will that be called). I haven't been able to find docs for `bin/dev` in rails 7.

Also if using a custom `bin/dev` (for example, to use `foreman` as in [this article](https://dev.to/software_writer/launching-multiple-processes-with-a-single-command-in-rails-l7)), will that mean that passed arguments get ignored?

kjellberg:
> Also if using a custom `bin/dev` (for example, to use `foreman` as in [this article](https://dev.to/software_writer/launching-multiple-processes-with-a-single-command-in-rails-l7)), will that mean that passed arguments get ignored?

No, which makes me realise that this was a bad idea. A better solution is to just alias `rdev` or something to `./bin/dev`. However I'm not sure that belongs in this repo/plugin, so I'm closing this PR.

Thanks for your time :)

PR #10747, opened 2022-03-01
feat(git): added git aliases for working with tags

gitaarik:
## Standards checklist:

## Changes:

- Added git aliases for working with tags


carlosala:
Hi Rik!
We have tons of PRs regarding the `git` plugin. In advance, sorry for the delay that could take this PRs to be reviewed ;)

gitaarik:
Ok, thanks for the update :)

adamchainz:
I like the idea of `gt` for `git tag`, but I’m not sold on the others.

`git push --tags` pushes *all* tags, which can lead to accidents. Using [`--follow-tags`](https://git-scm.com/docs/git-push#Documentation/git-push.txt---follow-tags) pushes only annotated tags associated with the current commit, which is preferable for many workflows. It can also be enabled by default with the `push.followTags` option, so there's no need for an alias change there.

`gshs` for `git show --stat` is not tag-related. Perhaps useful, but there are a bunch of other extra-detail-adding options on `git show` that seem similarly important...

carlosala:
What Adam says makes much sense!
Thanks for the triaging :)

gitaarik:
The `gt` alias for `git tag` was apparently already added by e2b855341bfd587590734da860d75ead7788810f and later again removed by 0e1e877ea9182f87b3ab34cbc2ea944b34f5345e because it supposedly clashes with [Graphite](https://graphite.dev/) CLI commands.

adamchainz:
Thanks. We have rediscovered that with comments on #11910.

PR #10753, opened 2022-03-03
fix(updater): prefix `cd` with `builtin` when it is aliased

SBado:
## Standards checklist:

## Changes:

- Modified check_for_upgrade.sh in order to use builtin cd.

## Other comments:

I'm using zoxide and I've aliased cd to z. This broke check_for_upgrade.sh because zoxide desn't understand the -q flag.
This PR force check_for_upgrade.sh to use builtin cd.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -36,11 +36,11 @@ function current_epoch() {
 
 function is_update_available() {
   local branch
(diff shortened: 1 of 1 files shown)


mcornella:
Aliasing `cd` is a bad idea and `z` is only one letter. In any case, this is also done in the CLI, so for consistence we can also do it here. There are a lot of calls to cd in the codebase though.

carlosala:
I have a small doubt. Is `builtin` in every POSIX-like shell?

PR #10771, opened 2022-03-13
feat(path): add plugin path(PATH variable operator)

kaivio:
## Standards checklist:

## Changes:
- add plugins/path/path.plugin.zsh

### DIFF
--- a/plugins/path/path.plugin.zsh
+++ b/plugins/path/path.plugin.zsh
@@ -1,0 +1,64 @@
+#!/bin/zsh
+
+# path - PATH variable operator
+#
+# varsion: 0.1
+# date: 2022-03-13 
+
+
+if [ -f ~/.zshrc.path ]; then
+  source ~/.zshrc.path
+fi
+
+function path() {
+  #local path=(${(s/:/)PATH})
+
+  if [ -z $1 ]; then
+    local i=0
+    for v in $path; do
+      echo -e "\e[3;36m$i\e[0m"  $v
+      (( i++ ))
+    done
+
+  elif [ $1 == 'add' ]; then
+    if [ -z $2 ]; then
+      path=("$PWD" "${path[@]}")
+    elif [ $2 == '-' ]; then
+      path=( "${path[@]}" "$PWD")
+    else
+      path=("${path[@]:0:$2}" "$PWD" "${path[@]:$2}")
+    fi 
+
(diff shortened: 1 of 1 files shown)


carlosala:
In my opinion this plugin is not really needed. Most times you only need to change the path in the `zshrc`, where you can easily add a line `export PATH="/path/to/dir:$PATH"`. Do you usually change the `PATH` variable interactively in the shell? By the way, this can easily be done executing the same line written before (but it's not recommended to do so, as many startup functions depend on `$PATH` and which directories are there).
What's your opinion about it?

kaivio:
> Do you usually change the PATH variable interactively in the shell? 

Yes, I often try alpha applications and temporary scripts, execute it in an external directory. Typing command shorter by added PATH, but long-term configuration is not necessary.

rwmitchell:
I manipulate my path as well.  Code I write goes into a project path that further has debug, release, and install components before going into the actual install area.  My tool for that does:

 pathtool -h

  addpath     : add current directory to path
  bugpath   N : use /debug/   in path
  rlspath   N : use /release/ in path
  gldpath   N : use /install/ in path
  cleanpath   : remove missing and duplicate path elements
  delpath   N : remove path
  ordpath O N : re-order path element
  dopath    N : renable path
  gop       N : go to path
  loadpath    : load saved paths
  nopath    N : disable path
  showpath    : show paths

the commands that "change" the path only outputs the new path to the terminal, using an alias wrapper to set it for the current shell.

Many of the options get the path to change by using a program name and finding with path element it exists in.

image

mcornella:
You can run these by using the full or relative path, you don't need to change `$PATH`. For example:
```console
$ wget https://example.com/some-command -O ~/Downloads
$ chmod +x ~/Downloads/some-command
$ ~/Downloads/some-command <arguments>
```

You could also add the current working directory to the path with
```sh
path+=(.)
```
and then you can run commands in the current directory without prefixing with `./` (note that if you do this, the recommended approach is to put the current working directory last, so as to not override system commands.

----

Regardless of that, while I appreciate the contribution this is not a thing that common users will need. You can still set it up as a custom plugin, see [adding a new plugin](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#adding-a-new-plugin). You can also create a repository for the plugin and add it to the list of [External plugins](https://github.com/ohmyzsh/ohmyzsh/wiki/External-plugins).

PR #10786, opened 2022-03-19
feat(burl): add plugin burl(comes with browser usual headers for curl)

kaivio:

## Changes:

- add plugins/burl/

### DIFF
--- a/plugins/burl/README.md
+++ b/plugins/burl/README.md
@@ -1,0 +1,3 @@
+# burl
+
+Comes with browser usual headers for curl.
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry for the very long delay.

The catch with this one is that its value is a block of hardcoded headers, and those go stale. It currently pins `Chrome/98.0.4758.87` from early 2022, and the user agent string mixes a Firefox `rv:85.0` token into a Chrome UA. Keeping that block current would be an ongoing job for us, and which headers to send is really a per-user choice.

Closing for that reason. It works well as a function in `$ZSH_CUSTOM`. Thanks for the contribution.

PR #10787, opened 2022-03-20
refactor(1password): update plugin and readme for 1Password CLI 2.0

### LINKED ISSUE #10789: 1Password plugin isn't working with 1Password CLI 2

1Password

The new CLI of 1Password is breaking the `1password` plugin – it introduces a new command structure.

See Docs: [https://developer.1password.com/docs/cli/upgrade](https://developer.1password.com/docs/cli/upgrade)
> 1Password CLI 2 introduces a noun-verb command structure that groups commands by topic rather than by operation.

Therefore the list command that's currently used by the plugin isn't working anymore.

Use the "Update your scripts" table, to support 1password cli 2: [https://developer.1password.com/docs/cli/upgrade#step-2-update-your-scripts](https://developer.1password.com/docs/cli/upgrade#step-2-update-your-scripts).

A workaround for now is not upgrading the 1Password CLI and stay with the predecessor for now. Though it would be nice to at least add a notice to the README.

### Additional context

_No response_

### Related Issues

_No response_

ag-adampike:
Minor changes to allow the plugin to work with 1Password CLI 2. Hi from 1Password! 👋🏻

## Standards checklist:

## Changes:

- refactored plugin for syntax changes in the latest version of 1Password CLI
- updated readme to reflect the updated name, change in sign-in behaviour, and documentation links

## Other comments:

1Password CLI 2.0 has recently been released with several added features and a new command schema (see [Upgrade to 1Password CLI 2](https://developer.1password.com/docs/cli/upgrade)). This PR is mainly a refactor, with a small change in behaviour of the underlying software used by this plugin (optional biometric login prompt), but no changes to the plugin's functionality.

Full disclosure: I work for 1Password.

Fixes #10789

### DIFF
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -27,7 +27,7 @@ function opswd() {
   echo -n "$password" | clipcopy
   echo "✔ password for $service copied to clipboard"
 
-  # If there's a one time password, copy it to the clipboard after 5 seconds
+  # If there's a one time password, copy it to the clipboard after 10 seconds
   local totp
   if totp=$(op item get --otp "$service" 2>/dev/null) && [[ -n "$totp" ]]; then
     sleep 10 && echo -n "$totp" | clipcopy
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -15,11 +15,11 @@ function opswd() {
   local service=$1
 
   # If not logged in, print error and return
-  op list users > /dev/null || return
+  op user list > /dev/null || return
 
   local password
   # Copy the password to the clipboard
-  if ! password=$(op get item "$service" --fields password 2>/dev/null); then
+  if ! password=$(op item get "$service" --fields password 2>/dev/null); then
     echo "error: could not obtain password for $service"
     return 1
   fi
@@ -29,7 +29,7 @@ function opswd() {
 
   # If there's a one time password, copy it to the clipboard after 5 seconds
   local totp
-  if totp=$(op get totp "$service" 2>/dev/null) && [[ -n "$totp" ]]; then
+  if totp=$(op item get --otp "$service" 2>/dev/null) && [[ -n "$totp" ]]; then
     sleep 10 && echo -n "$totp" | clipcopy
     echo "✔ TOTP for $service copied to clipboard"
   fi
@@ -39,7 +39,7 @@ function opswd() {
 
 function _opswd() {
   local -a services
-  services=("${(@f)$(op list items --categories Login 2>/dev/null | op get item - --fields title 2>/dev/null)}")
+  services=("${(@f)$(op item list --categories Login 2>/dev/null | op item get - --fields title 2>/dev/null)}")
   [[ -z "$services" ]] || compadd -a -- services
 }
 
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -1,46 +1,9 @@
-if (( ${+commands[op]} )); then
-  eval "$(op completion zsh)"
-  compdef _op op
-fi
+# Do nothing if op is not installed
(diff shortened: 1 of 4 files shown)


carlosala:
Hi Adam, thanks for your contribution! Maybe to maintain the compatibility with 1P 1.x we should have both codes. What do you think about, it's now 1.x CLI not working anymore or can be both used?

ag-adampike:
Thanks for the quick response!

The legacy CLI will continue to work and be supported for some time yet, but it _will_ eventually be deprecated. I certainly think it's fine to keep both around if that makes more sense.

Perhaps a version check (i.e. `op --version`) could be implemented to automatically use the correct syntax. 🤔

ag-adampike:
Apologies for the noise. Having some issues with signing off commits on this machine. :-/

mcornella:
Hi Adam! Thanks for the PR! Given that this is a breaking change we'll have to add a `BREAKING CHANGE` note so that users update to CLI 2. [These would be the upgrade docs](https://developer.1password.com/docs/cli/upgrade) that we link to, right? Something like:

```
refactor(1password)!: rewrite `opswd` to support 1password CLI 2 (#10787)

BREAKING CHANGE: this commit changes the `opswd` function to use the new
1password CLI 2 arguments. If you're using CLI 1 (check with `op --version`)
you're going to need to update to the new version.
See https://developer.1password.com/docs/cli/upgrade for instructions.
```

Otherwise, if we want to keep backward compatibility with v1 we could redefine the `opwsd` function depending on `op --version` (doing it differently will add complexity to the function which will be very difficult to maintain). We can leverage the `is-at-least` function for that:
```
# TODO (date): remove old code
autoload -Uz is-at-least
if is-at-least 2.0.0 "$(op --version)"; then
  function opswd {
    # new opswd function
  }
else
  function opswd {
    # new opswd function
  }
fi
```

This other approach will enable us to not add the breaking change message.

ag-adampike:
Hey Mark! That is the correct documentation, yes.

My _personal_ preference would be to encourage folks upgrade to the new version, but I am a bit biased here. I am also certainly sympathetic to anyone needing time to make a change.

ag-adampike:
There are no strict dependencies for any other app for 1Password CLI 2. The requirement for 1Password 8 for Mac is to enable biometric unlock.

ag-adampike:
And to clarify, biometric unlock is a new feature in 1Password CLI 2.0 that is not available in 1Password CLI 1.x.

moritzgvt:
Ah sorry, you're right. I was following this [signin guide](https://developer.1password.com/docs/cli/get-started#sign-in) and missed the part mentioning 1PW 7 support.

ag-adampike:
No apologies necessary! 😊 I would share the same concerns.

For the purposes of this plugin, there should be no difference in the behaviour at worst. At best, you also get biometric unlock!

mcornella:
OK, I'm ready to merge the PR as is while marking as breaking. Are there any other objections?

carlosala:
I'll be more for check which version the user has, but it's fine to merge it like that too👌🏻

mcornella:
I've extracted `opswd` into its own file and added a deprecation message that will only be displayed the first time `opswd` is run.

I've also fixed the completion for `opswd` in CLI 2: I'd get an error when trying to obtain only the title:
```console
$ op item list --categories Login --format json --cache | op item get - --fields title
[ERROR] 2022/03/24 16:35:08 unable to process element 1: "title" isn't a field in the "xxx" item
```

carlosala (approved):
Nice job Marc!

ag-adampike:
Ditto, thank you! 🤩

I had been working on this in my spare time, which is pretty scarce throughout the week. 😅

mcornella:
I recreated the commits and pushed the changes. Thanks Adam for your work on 1Password :)

ag-adampike:
Likewise, thanks for your work on this and all of Oh My Zsh. ❤️

PR #10797, opened 2022-03-24
Update laravel.plugin.zsh

geneowak:
## Standards checklist:

## Changes:

Added two new aliases
- `pat` runs tinker which is used a lot during development to interact with the app through the terminal
- `paoc` clears all the cache at once which saves the time that it would take to clear each cache individually.

## Other comments:

...

### DIFF
--- a/plugins/laravel/README.md
+++ b/plugins/laravel/README.md
@@ -10,6 +10,7 @@ plugins=(... laravel)
 |:-:|:-:|
 | `artisan`  | `php artisan`  |
 | `pas`  | `php artisan serve` |
+| `pat`  | `php artisan tinker` |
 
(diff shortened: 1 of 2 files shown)


abdessamadbettal (approved):

robbyrussell:
Thanks for this, and sorry it waited so long for a reply.

`pat` for `artisan tinker` is a good addition and we don't have it. `paoc` I'd leave out though: `optimize:clear` already has an alias, `paopc`, added a couple of years after you opened this.

Could you rebase on master and keep just `pat`, with its README row? If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll take it from here.

PR #10806, opened 2022-03-28
docs(bazel): improve plugin README

alealv:
## Standards checklist:

## Changes:

- Update bazel completion link

### DIFF
--- a/plugins/bazel/README.md
+++ b/plugins/bazel/README.md
@@ -1,5 +1,14 @@
-## Bazel autocomplete plugin
+# Bazel plugin
 
-A copy of the completion script from the
-[bazelbuild/bazel](https://bazel.build/install/completion#zsh)
-git repo.
+This plugin adds completion for [bazel](https://bazel.build), an open-source build and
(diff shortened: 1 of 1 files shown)


carlosala:
Hi Alejandro! I think it's better to let that line as is! Probably users would take more advantadge of understanding what basel is that just redirecting to a zsh file. Thanks for your contribution!

mcornella:
Link should be https://github.com/bazelbuild/bazel/blob/master/scripts/zsh_completion/_bazel though.

alealv:
Yes, link was broken, that's why I created the PR. Although, I should have pointed to the zsh file instead.

carlosala:
But the previous link was working. Why do you want to change it?

mcornella:
> But the previous link was working. Why do you want to change it?

It's missing `blob/` before `master`.

carlosala:
I was talking about the one bundled now with omz. The link to bazel website

carlosala:
Ough I messed up the `diff`. I thought that he was changing from bazel.build to github.com, so reversed to reality. Sorry guys 😅

PR #10812, opened 2022-03-30
feat(1password): add username copy to `opswd`

gheritarish:
## Standards checklist:

## Changes:

- This commit updates the `opswd` command to also copy the username of a
service into the clipboard
- The README was updated to take those changes into account

## Other comments:

This only works for v2 of the 1password CLI, and I closed my previous PR doing this for v1 of the CLI (#10637 )

### DIFF
--- a/plugins/1password/README.md
+++ b/plugins/1password/README.md
@@ -14,16 +14,18 @@ clipboard.
 ## `opswd`
 
 The `opswd` command is a wrapper around the `op` command. It takes a service
-name as an argument and copies the password for that service to the clipboard.
+name as an argument and copies the username, then the password for that service
+to the clipboard, after confirmation on the user part.
 
-If the service also contains a TOTP, it is copied to the clipboard after 10 seconds.
-Finally, after 20 seconds, the clipboard is cleared.
+If the service also contains a TOTP, it is copied to the clipboard after confirmation
+on the user part. Finally, after 20 seconds, the clipboard is cleared.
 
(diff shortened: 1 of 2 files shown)


mcornella:
Hi @gheritarish, thanks for reimplementing the functionality for CLI 2. I did prefer the UX of just asking to press Enter (https://github.com/ohmyzsh/ohmyzsh/pull/10637#issuecomment-1027021508). Now I think the user needs to do more work, i.e. pressing y is harder than pressing Enter, since you need to move your finger from the Enter key. 

It would be worse if it had to also press Enter, but you solved it by reading a single character, so kudos for that.

gheritarish:
Hi @mcornella and thanks for your answer!
If I recall correctly, I did it that way since I didn't found how to get the Enter key in the script. I'll try to see how I can update it to take the Enter key instead.

mcornella:
A simple read does that, and you can ignore the input.

gheritarish:
Thanks for letting me know!
I tested it on my machine, it works; maybe not the most elegant way to do it though.

mcornella:
Awesome! Can you post the changes?

gheritarish:
Just force-pushed it! (with the wrong account because I'm bad at managing several accounts from the same computer…)

gheritarish:
@mcornella just pinging you in case you missed it. Thanks!

mcornella:
I totally forgot about this, but thanks for your hard work!

PR #10818, opened 2022-04-03
Add additional alias to multipass

h2ogeek:
## Standards checklist:

- [ X ] The code is stable and I have tested it myself, to the best of my abilities.

## Changes:

- Add additional alias to multipass plugin

## Other comments:

...


robbyrussell:

`mpd`, `mpp`, `mpe` and `mpf` all look good to me. One thing to fix: `mpsh` duplicates `mps`, which already maps to `multipass shell`.

Could you drop that line and add rows for the other four to the README alias table? If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll take it from here.

PR #10819, opened 2022-04-03
use loop to create aliases for dir stack

rwmitchell:
If DIRSTACKSIZE is set, create alias for each stack entry.
if DIRSTACKSIZE is not set, create same as previous hard coded method.

## Standards checklist:

- [ X] The PR title is descriptive.
- [ X] I have read the contribution guide and followed all the instructions.
- [ X] The code follows the code style guide detailed in the wiki.

## Changes:

- rewrote hardcoded lines into a loop to match DIRSTACKSIZE

## Other comments:

...

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -9,16 +9,13 @@ alias -g .....='../../../..'
 alias -g ......='../../../../..'
 
 alias -- -='cd -'
-alias 1='cd -1'
-alias 2='cd -2'
-alias 3='cd -3'
-alias 4='cd -4'
-alias 5='cd -5'
-alias 6='cd -6'
-alias 7='cd -7'
-alias 8='cd -8'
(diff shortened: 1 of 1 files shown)


rwmitchell (commented):
Is there a better/safer way to do this that is keeping this from being approved ?

rwmitchell:
Any chance this can get reviewed?  Using a loop to create the exact numbers of aliases to match DIRSTACK is much better than hardcoding creating a set number that can as easily be too large as too small.

carlosala:
I like it! Could we have zstyle-like option? Maybe `':omz:lib:directories' cd-max-alias`?

carlosala:
Let's close this for the moment. We didn't see a huge need to limit the amount of aliases. Thanks for your contribution!

PR #10823, opened 2022-04-05
feat(golang): add aliases for  utility

180909:
## Standards checklist:

## Changes:

- Added aliases for go mod download, go mod graph, go mod init, go mod tidy and go mod vendor

## Other comments:

...


jmeridth (changes requested):
You need to update the README file too

jmeridth (approved):

180909:
ping?

carlosala:
Hi! We are developing a brand new alias policy where we are going to be more "strict" in how we accept aliases, due to the massive amount of them that we have in the project. One of the points is that we are going to accept one generic alias instead of several specific ones. In this case, `go mod` would be the generic one and `go mod download`, for example, the specific one. You can check that policy (is still being finished) [here](https://github.com/ohmyzsh/ohmyzsh/pull/11045).
Thanks for your contribution!

180909:
Thanks review. I will close this pr.

carlosala:
Thanks for your contribution 😉

PR #10830, opened 2022-04-06
feat(updater): check for typed input before automatically updating

mcornella:
## Standards checklist:

## Changes:

- Move check for `reminder` mode and typed user input together.
- Clear line and move cursor to the first column when displaying the reminder.
- Move check for `auto` mode after the check for typed user input. This aborts the automatic update if the user has started typing, meaning that they have stuff to do.


carlosala (approved):
Nice! I think this approach is really correct, it's what I would expect to happen!

PR #10832, opened 2022-04-07
fix: use `python3` for all `python` invocations

### LINKED ISSUE #10469: Support `python3` executable when `python` is not available
Many systems don't link `python` to `python3`. Debian even [maintains it's wrong for their packages to expect `python` to work](https://wiki.debian.org/Python):
> NOTE: Debian testing (bullseye) has removed the "python" package and the '/usr/bin/python' symlink due to the deprecation of Python 2. **No packaged scripts should depend on the existence of '/usr/bin/python'**: if they do, that is a bug that should be reported to Debian. You can use the 'python-is-python3' or 'python-is-python2' packages to restore an appropriate '/usr/bin/python' symlink for third-party or legacy scripts.

Therefore we should support both `python` and `python3`. This can be done from Zsh with the form
```zsh
${commands[python]:-python3}
```
but the use in file shebangs is harder to do.

It should be researched whether `python3` is expected to be available in all systems, so we can just rename all `python` uses to `python3`, or if either command can be available in different systems.

Reference: #10466

-----

Used in plugins:
- aliases
- [django](https://github.com/ohmyzsh/ohmyzsh/blob/3e4b520190f1e4aaced9267bd5dcde59b3d0211f/plugins/django/django.plugin.zsh#L377)
- git-prompt
- jsontools
- python
- shell-proxy (as `python3`)
- urltools (as `python2` or `python3`)

### LINKED ISSUE #10777: git-prompt plugin doesn't work on macOS 12.3
### Describe the bug

The git-prompt plugin depends on python. Apple had deprecated python2 ages ago, and with the latest release of macOS 12.3, python2 is is gone, and so is the binary `/usr/bin/python`. In order to keep using the plugin, I've had to resort to hackery, by aliasing `python` to `python3` in my .zshrc. Other hacks of using symlinks from python3 to python somewhere in PATH also work. However, the plugin should use python3, instead of python.

### Steps to reproduce

1. Upgrade to macOS 12.3
2. Enable the "git-prompt" plugin
3. cd to a git checkout

### Expected behavior

I expect to see git status information in the prompt, but I don't.

_No response_

### OS / Linux distribution

macOS 12.3

### Zsh version

5.8

### Terminal emulator

iTerm2

_No response_

### Additional context

_No response_

### LINKED ISSUE #10796: Aliases plugin not working
### Describe the bug

Hi. When I try to use aliases plugin I got `[error] No python executable detected` 

### Steps to reproduce

`acs git` -> `[error] No python executable detected`

### Expected behavior

Return list of available git aliases

_No response_

### OS / Linux distribution

macOS Monterey 12.3

### Zsh version

5.8

### Terminal emulator

macOS terminal

_No response_

### Additional context

_No response_

carlosala:
## Standards checklist:

## Changes:

The title is descriptive.

Closes #10469
Closes #10681 
Closes #10775
Closes #10777 
Closes #10796 
(and probably many others)

### DIFF
--- a/plugins/aliases/aliases.plugin.zsh
+++ b/plugins/aliases/aliases.plugin.zsh
@@ -3,9 +3,9 @@
 # - acs: alias cheatsheet
 #   group alias by command, pass addition argv to grep.
 function acs(){
-  (( $+commands[python] )) || {
+  (( $+commands[python3] )) || {
     echo "[error] No python executable detected"
     return
   }
-  alias | python ${functions_source[$0]:h}/cheatsheet.py $@
+  alias | python3 ${functions_source[$0]:h}/cheatsheet.py $@
 }
--- a/plugins/aliases/cheatsheet.py
+++ b/plugins/aliases/cheatsheet.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 import sys
 import itertools
 import termcolor
--- a/plugins/git-prompt/README.md
+++ b/plugins/git-prompt/README.md
@@ -13,7 +13,7 @@ See the [original repository](https://github.com/olivierverdier/zsh-git-prompt).
 
 ## Requirements
 
-This plugin uses `python`, so your host needs to have it installed
+This plugin uses `python3`, so your host needs to have it installed.
 
 ## Examples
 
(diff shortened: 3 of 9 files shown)


mcornella (approved):

mcornella:
LGTM. I'll reword it as a bugfix though. Thanks Carlo!

marshally:
FWIW, I was able to resolve this same issue on a OS X w/ M1 CPU by symlinking the homebrew `python3` to a path accessible `python`

```bash
ln -s /opt/homebrew/bin/python3 /opt/homebrew/bin/python
```

PR #10833, opened 2022-04-07
Added more terraform aliases

l13t:
## Standards checklist:

## Changes:

- Added aliases that simplify work with Terraform workspaces in CLI

## Other comments:

...


klyse:
did you consider adding another alias: `tfa!`for `terraform apply -auto-approve`.

l13t:
I'm not a fan of such approach :) . But it could be something like `tfaa`

carlosala:
Maybe those are too many aliases. Could we stick to just adding one or two? There's a new policy for that [here](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases).
Thanks!

nitrocode:
If it will speed things up, I'd be content with the subset of only workspace, list, and select aliases.

nitrocode:
cc @robbyrussell thoughts on this PR?

carlosala:
There are a lot of aliases in this PR, but it isn't getting much traction. Closing!

PR #10840, opened 2022-04-10
Export GIT_PROMPT in git-prompt plugin

njfamirm:
## Standards checklist:

## Changes:

- export GIT_PROMPT

## Other comments:

git prompt plugin is awesome, ‌but the user may need to use it in their PROMPT or elsewhere...
So instead of setting it as a RPROMPT in [this file](https://github.com/ohmyzsh/ohmyzsh/blob/846f417eb8ec76e8eee70000e289b8b81f19d480/plugins/git-prompt/git-prompt.plugin.zsh#L101), I think we should export `GIT_PROMPT` variable to the user to use wherever he wants!

### DIFF
--- a/plugins/git-prompt/README.md
+++ b/plugins/git-prompt/README.md
@@ -11,6 +11,14 @@ plugins=(... git-prompt)
 
 See the [original repository](https://github.com/olivierverdier/zsh-git-prompt).
 
+# Usage
+
+Use `$GIT_PROMPT` in PROMPT/RPROMPT variable in `.zshrc` file
+
+```zsh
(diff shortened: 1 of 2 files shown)


carlosala:
This is not necessary! You can use `$(git_super_status)` wherever in your PROMPT variables in order to use `git-prompt` plugin!

njfamirm:
> This is not necessary! You can use `$(git_super_status)` wherever in your PROMPT variables in order to use `git-prompt` plugin!

Yes, you are absolutely right

But there is no ability to disable the  `RPROMPT` and only use the `git_super_status`

It's right?

njfamirm:
True
thank you!

Issue #10857, opened 2022-04-15
Alias collision check for CI


hofbi:
### If the feature request is for a plugin or theme, specify it here.

GitHub Actions

### If the feature solves a problem you have, specify it here.

As part of #10707, we identified an alias collision, which we became aware of a manual review. I thought that this could be actually validated with a CI job.

### Describe the proposed feature.

One possible sulution for this could be a simple Python or Shell script that checks if there are any duplicate alias definitions accross all plugins. Then, an actions jobs checks this for any PR and fails if a collision is identified.

### Describe alternatives you've considered

The alternative would be to not include the feature, since it is only a check to avoid future problems.

hofbi:
If there is an interest in this feature, I could also come up with a draft for this feature.

carlosala:
I think this would be really interesting. We have been thinking about how to improve CI (for Zsh there are no linters nor prettifiers) and I think this is a nice way!
If you want to do a draft we'll really appreciate it!

hofbi:
Ok, I will go for it.

Regarding additional checks, I am a big fan of [pre-commit](https://pre-commit.com/), which is a great tool for managing all kinds of checks and linters. Thats why I submitted https://github.com/ohmyzsh/ohmyzsh/pull/10707, to add pre-commit aliases. Even if there are no specific checks for Zsh, there would be basic things such as trailling whitespace fixers or shell formatters. I also saw that you have some Python code, so there would be the opportunity to add Python formatters as well. So I am open to support here as well.

Issue #10851, opened 2022-04-13
'defining function based on alias `dls' error when using both docker and systemadmin plugins


DmitrySandalov:
`oh-my-zsh` version: `[master c6f0504]`

Plugins in `~/.zshrc`
```
plugins=(
  docker
  systemadmin
)
```

Try opening new shell.

Expected: no errors
Observed: 
```
/home/myuser/.oh-my-zsh/plugins/systemadmin/systemadmin.plugin.zsh:37: defining function based on alias `dls'
/home/myuser/.oh-my-zsh/plugins/systemadmin/systemadmin.plugin.zsh:37: parse error near `()'
```

```
$ alias | grep 'dls='
dls='docker container ls'
$ declare -f dls
< blank >
```

https://github.com/ohmyzsh/ohmyzsh/blob/c6f0504cf0f75bc17cc422e82fdb84f0c994c461/plugins/systemadmin/systemadmin.plugin.zsh#L37
https://github.com/ohmyzsh/ohmyzsh/blob/c6f0504cf0f75bc17cc422e82fdb84f0c994c461/plugins/docker/docker.plugin.zsh#L8

tunguyen9889:
The `dls` alias in docker plugin was introduced in this PR https://github.com/ohmyzsh/ohmyzsh/pull/6527, need to make another alias to avoid name conflict.

mcornella:
@abhinavmanchanda can we rename the `dls` and `dlsa` aliases to avoid a conflict?

abhinavmanchanda:
Sure, give me a couple of hours.

abhinavmanchanda:
@mcornella I've created the following pull request - 

https://github.com/ohmyzsh/ohmyzsh/pull/10852

I noticed that you've already pushed a fix to the conflict. If you feel the PR is not needed, please reject it.

abhinavmanchanda:
Got it.

PR #10859, opened 2022-04-15
Alias collision

hofbi:
## Standards checklist:

## Changes:

Relates #10857 

This is a first draft checking for alias collisions. I used Python to implement this feature to provide a solid set of tests and added a job to the GH actions

## Other comments:

Since this is a first draft there a some open points to discuss:

1. I added everything related to this check into the `tools` folder. Is this locations fine or should we move it somewhere else?
2. As a first draft, I kept the regex simple. I would first try to get the basic feature in and then improve over time.
3. Already with this basic feature, the script finds some alias collisions. That's why the CI fails right now. How should we proceed here? Disable the CI job for now and fix the collisions first or fix them as part of this PR?


mcornella:
I think we need to make this a zsh check to account for dynamic alias definitions and function conflicts (i.e. "aliases" that are defined as functions, like #10851). A way to do this could be to run zsh, then parse the output of `alias` and `print -l ${(k)functions}`, though this could potentially be super slow if we need to source all plugins.

The problem of side-effects and running extra code (things not related to aliases and functions) would need to be considered as well. I'll think carefully about this last part.

Good start though, love that it contains tests!

hofbi:
Is there an easy way to source all plugins? Then I would give it a try to see how the runtime will behave.

The current implementation already found a few duplicates. How should we proceed here? I could create a separate issue for every plugin to kickof a discussion for resolving the name conflict.

hofbi:
@mcornella How should we proceed here? Is this still relevant for you?

carlosala:
I'd push for a zsh check, so we can check also functions. Another problem might be that some aliases are defined only if some binary is present in `$PATH`. Maybe our best option is to go for regex checking all the files looking for `alias <word>`, `function <word> (or <words>)`, `<word> () {` and then if there are two with the same name.
What do you think? @hofbi @mcornella

hofbi:
Agreed.

My suggestion on how to proceed would be to go with this check for now and fix everything that was catched so far. Once this is done, we can extend the check for functions and everything else that comes to our mind.

TriMoon:
> @hofbi 
> 3. Already with this basic feature, the script finds some alias collisions. That's why the CI fails right now. How > should we proceed here? Disable the CI job for now and fix the collisions first or fix them as part of this PR?

I would vote for "fix them as part of this PR" :wink: 
Good job so far with this cleanup guys, im sure it's a lot of work... :+1:

hofbi:
@carlosala What is you opinion on how to proceed with the existing alias collisions?

mcornella:
Hey Markus, the technical work so far is spotless but I'm not sure how we proceed with the aspect of solving collisions. Perhaps we can proceed with a CI approach that is "this PR does not introduce new collisions", kinda like how an iterative approach is done with unit testing.

For this, the output of the tool would need to be refined and the job changed so it makes the comparison between master and the PR. I will work on refining the CI part and then we can begin a new issue discussing how to approach current alias collisions.

Thank you for your work!

carlosala:
I think we can move this forward! I'd go for the "this PR does not add new collisions" thing, and it'd be great to add an automated check as well for collisions with other cli binaries, leveraging something like https://command-not-found.com
@hofbi what do you think about?

carlosala:
Since this would be a CI-only check, I'm OK going for python. Previously I mentioned something like turning this PR into a shell script, but python is the way to go IMO. 👍🏻

hofbi:
Sounds good to me. I would leave the binary tool collision for a follow up PR then.

At some point, we could also consider making the part of a tool such as https://pre-commit.com. But I guess you don't want any user/developer to depend on Python.

carlosala:
Yep, it's probably better to have that only on CI 👍🏻

tools/alias_collision/check_alias_collision.py
```diff
@@ -0,0 +1,148 @@
...
+import itertools
+import re
+
+
+ERROR_MESSAGE_TEMPLATE = (
+    "Alias `%s` defined in `%s` already exists as alias `%s` in `%s`."
+)
+
+# TODO: We want that list to be empty
+KNOWN_COLLISIONS = [
```
hofbi, line 16: We can reduce this list in the future. For now, we prevent new duplicates from being introduced

hofbi:
@carlosala I slightly refactored this PR to use pytest over unittest, which is just nicer to use. With https://github.com/ohmyzsh/ohmyzsh/pull/10859#discussion_r1476081962, I added a filter that lists all know collisions. Execpt for those, the check will fail for new collisions. In the future we can work on reducing collisions and extend with further checks.

tools/alias_collision/check_alias_collision.py
carlosala: What about comparing the duplicates from `master` (in the target commit, or latest, we can think about it) and the ones from the PR? I think it's better than having a list.
Imagine if we add an alias called `bcubc` besides the two that are already in the project.

hofbi: How about explicitly listing the full `Alias` instances? IMO the benefit of having such a list is that you know exactly how far you are. Comparing against master can be a moving target.

carlosala: You could export them as JSON or something like that and then parse the answer.

hofbi: Done

hofbi:
I removed the known collision resolved in https://github.com/ohmyzsh/ohmyzsh/pull/12192

hofbi:
@carlosala quick reminder to move on with this

carlosala:
Yep, I have it on my TODO. Sorry for the delay, I'm a bit overwhelmed these days

carlosala (commented):
I kinda like it! I'd rather go to checking every time on the target branch (usually master) and on the source branch (the one you open the PR from).
Then, the workflow would be:
- Run the script in target ref (if it's not a PR and instead it's just a push, use `master`) and save the JSON
- Run the script in source ref and verify that all matches live as well in previous JSON file

hofbi:
If I understand the motivation for you approach correctly, you want to avoid storing the json file in VCS, right? On the other hand, having the reference from master in VCS, it is easier to track the status about how many collision you have right now.

Regardless of which way we go for, do you think this could become a follow up PR. This PR, as is provides already value and would prevent new collisions from being introduced.

carlosala:
Yep, exactly. I'd say it's better not to have it in the repo.
I would like to already merge a final thing, and avoid follow-up PRs.

hofbi:
@carlosala I updated the GH action. Now we create the baseline file on the source branch. I a second step, we compare the target branch collisions with this baseline.

carlosala:
Could you move all this code to `.github/workflows`? Something like https://github.com/ohmyzsh/ohmyzsh/tree/master/.github/workflows/dependencies

carlosala:
I'll verify the python code during this week, thanks for the hard work!

hofbi:
> Could you move all this code to `.github/workflows`? Something like https://github.com/ohmyzsh/ohmyzsh/tree/master/.github/workflows/dependencies

Should I moved it into the same folder and merge the requirements.txt or into a seperate one?

carlosala:
I'd do different ones actually

hofbi:
Moved it to `.github/workflows/alias_collisions`

carlosala:
Some things I spotted along the way:
- We're not really interested in the value itself of the alias, I'd only save the file where is declared and nothing else. My arch for it would be (automatically supporting >2 repeated aliases, which now is an issue:
```json
[
  {
    "alias": "<duplicated>"
	"locations": ["<file1>", "<file2>", "...", "<fileN>"]
  }
]
```
  Main issue with it: if we have to equal aliases in the same file, weird corner cases might appear.
- For the moment we're only matching `alias foo='bar'`, and aliases might be declared without quotes, etc.
- We only match aliases defined in the start of the line, let's remove the `^`.

carlosala:
I'll keep reviewing it during the next days, but really great job!

hofbi:
@carlosala a kind reminder for this PR. It's already some time ago, so let me know if I should rebase first or if you want to review as is.

robbyrussell (changes requested):
Thanks for your patience on this one, @hofbi. Four years and several rounds of redesign is a lot to ask of anyone, and the shape you landed on is right — the baseline-diff model means zero noise on day one, which is what makes this mergeable at all.

I ran your script against current master, unmodified. It reports 47 collisions. Then I ran it again with one line changed, and it reports 127.

The difference is the regex:

```python
matches = re.findall(r"^alias (.*)='(.*)'", file.read_text(), re.M)
```

It only matches alias definitions that start at column 0 and use single quotes. Across `plugins/` on master that is 1,895 of roughly 2,483 alias lines — it misses 484 double-quoted definitions and 198 indented ones.

The concrete cost: `brew` has `alias bi='brew install'` and `bundler` has `alias bi="bundle install"`. That collision is invisible to the check, because bundler uses double quotes. Of the two examples you gave in #10857, `ma` is caught and `db` is not.

Something along these lines closes most of the gap:

```python
re.findall(r"^[ \t]*alias (?:-[gs] )?([^ =]+)=['\"]?(.*?)['\"]?$", file.read_text(), re.M)
```

Two smaller things, both optional:

1. The workflow passes only `plugins`, so `lib/` is never scanned. `lib/directories.zsh` defines `l`, `la`, `ll` and `..`, which currently cannot collide with anything.
2. `is_new_collision` compares only the alias name against the baseline, so once `gcd` is a known collision, a third plugin adding `gcd` passes silently. Fine for a first version — worth a comment in the code so the next person knows.

And some tidying, take it or leave it: the 163-line Python `.gitignore` template could be just the one entry you actually need (`known_collisions.json`); `pytest-sugar` is cosmetic and does not need to be a CI dependency; and `return -1` exits with status 255, where `return 1` is the usual thing.

Fix the regex and I will merge this. The rest is optional.

hofbi:
@robbyrussell Addressed most of your comments. I leave the other ones for follow up PRs to at least get the first version in.

PR #10863, opened 2022-04-16
feat(sublime-merge): add windows path for sublime merge

mpythonite:
## Standards checklist:

- [ x The code follows the code style guide detailed in the wiki.
- [ x The code is efficient, to the best of my ability, and does not waste computer resources.

## Changes:

- Added known Windows path to Sublime Merge to `sublime-merge` plugin

### DIFF
--- a/plugins/sublime-merge/sublime-merge.plugin.zsh
+++ b/plugins/sublime-merge/sublime-merge.plugin.zsh
@@ -4,15 +4,21 @@
 
 	if [[ "$OSTYPE" == linux* ]]; then
 		local _sublime_linux_paths
-		_sublime_linux_paths=(
-			"$HOME/bin/sublime_merge"
-			"/opt/sublime_merge/sublime_merge"
-			"/usr/bin/sublime_merge"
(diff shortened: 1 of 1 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry it sat so long without a review. The gap is real: `plugins/sublime` has handled WSL since 2019 and this plugin never got the same treatment.

Two things I'd want changed before merging:

1. Under WSL this replaces the Linux path list rather than adding to it, so anyone running WSL with Sublime Merge installed natively inside Linux would lose the `sm` aliases they have today. Adding the `wslpath` entry to the existing list avoids that.
2. The `uname -r` call runs on every Linux shell start, including for people who never touch WSL. `plugins/sublime` only gets there after `(( $+commands[subl] ))` fails; something similar here would keep the cost off everyone else.

Separately, and not your problem: the cygwin branch has a variable name typo that predates this PR, so it never matches anything. I'll fix that on its own.

If you'd rather not pick this back up after this long, that's completely fine. Say so and I'll finish it and credit you.

PR #10864, opened 2022-04-17
ying yang symbol added

outlinepix:


carlosala:
As `agnoster` is a really famous and used theme, it would be better to edit the theme locally and use your customized theme. By the way, thanks for your contribution!

PR #10870, opened 2022-04-20
fix(kubectl): sanity checks to improve compatibility with non OMZ users

obvionaoe:
## Standards checklist:

## Changes:

- Check if ZSH_CACHE_DIR is set, if it isn't, ask user to set it
- Check if completions directory exists, if it doesn't, create it

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -1,7 +1,14 @@
 if (( $+commands[kubectl] )); then
   # If the completion file does not exist, generate it and then source it
   # Otherwise, source it and regenerate in the background
+  if ! (( ${+ZSH_CACHE_DIR} )); then
+    echo "The \$ZSH_CACHE_DIR variable is not set, please set it\!"
+  fi
+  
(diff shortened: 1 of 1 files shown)


mcornella:
Hi Luis, how are you using the kubectl plugin. I presume a plugin manager? If so, such plugin manager should take care of creating the necessary conditions that the Oh My Zsh script does.

obvionaoe:
> Hi Luís, how are you using the kubectl plugin. I presume a plugin manager? If so, such plugin manager should take care of creating the necessary conditions that the Oh My Zsh script does.

I kind of agree, but feel that adding this wouldn't harm OMZ and introduce some sort of compatibility with other plugin managers, since it would only execute these steps when used outside OMZ

carlosala:
I think that the main problem is that we should modify tons of plugins that generates completions. Probably is easier to just open a PR in your plugin manager of choice, or even just add that small script in your `zshrc`.
I think it's better to maintain our architecture as is, even though we should probably improve in that aspect (it's currently in our to-do ideas).
By the way, thanks for your contribution! 😊

PR #10874, opened 2022-04-22
feat(brew): add `brew shellenv` command at plugin load

### LINKED ISSUE #10389: Source homebrew shellenv to setup homebrew in new install location

brew

Homebrew seems to install in /opt/homebrew now and has a shell script that needs to be eval'd to set it up.

Could the following snippet be included in the brew plugin to configure homebrew correctly.

```bash
# Homebrew
if [[ -f /opt/homebrew/bin/brew ]]; then
  eval "$(/opt/homebrew/bin/brew shellenv)"
fi
```

I manually added the above snippet myself to my zshrc file

### Additional context

_No response_

### Related Issues

_No response_

carlosala:
## Standards checklist:

## Changes:

Title is quite descriptive 👍🏻 

Closes #10389

### DIFF
--- a/plugins/brew/brew.plugin.zsh
+++ b/plugins/brew/brew.plugin.zsh
@@ -1,11 +1,35 @@
+if (( ! $+commands[brew] )); then
+  if [[ -x /opt/homebrew/bin/brew ]]; then
+    BREW_LOCATION="/opt/homebrew/bin/brew"
+  elif [[ -x /usr/local/bin/brew ]]; then
+    BREW_LOCATION="/usr/local/bin/brew"
+  elif [[ -x /home/linuxbrew/.linuxbrew/bin/brew ]]; then
+    BREW_LOCATION="/home/linuxbrew/.linuxbrew/bin/brew"
+  elif [[ -x "$HOME/.linuxbrew/bin/brew" ]]; then
+    BREW_LOCATION="$HOME/.linuxbrew/bin/brew"
+  else
+    return
+  fi
+fi
+
+if [[ -z "$HOMEBREW_PREFIX" ]]; then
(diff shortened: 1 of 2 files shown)


plugins/brew/brew.plugin.zsh
```diff
@@ -1,11 +1,19 @@
+if (( $+commands[brew] )); then
```
mcornella, line 1: This check is backwards, it should be
```suggestion
if (( ! $+commands[brew] )); then
```

plugins/brew/brew.plugin.zsh
```diff
@@ -1,11 +1,19 @@
...
+  return
+fi
+
+if [[ -z "$HOMEBREW_PREFIX" ]]; then
+  eval "$(brew shellenv)"
```
mcornella, line 6: I think we should support running brew by checking the absolute path to the executable, if the command is not found. 

That means rewriting the initial check and only bailing out if neither command is found (`/opt/homebrew/bin/brew` or `/usr/local/bin/brew`).

I just realised that Homebrew can also be installed in Linux, maybe we should also check for that (see linuxbrew).

carlosala: Hmm looks nice. I'll update it tomorrow 👌🏻

plugins/brew/brew.plugin.zsh
carlosala: Ups sorry😅😅

carlosala:
@mcornella it should be done! `brew shellenv` is editing the PATH, so it's not necessary to edit it manually in case of `brew` not being in the PATH 👌🏻

carlosala:
@mcornella take a look on this asap! 😉

PR #10875, opened 2022-04-22
fix(nvm): `brew` prefix changed for Apple M1 users

### LINKED ISSUE #10784: The NVM plugin doesn't work when NVM installed via Homebrew on Apple Silicon Macs
### Describe the bug

The NVM plugin looks for nvm installed by Homebrew in the old location, which is not the location where it is installed on Homebrew 3 on Apple Silicon macs.

The following line is responsible:
https://github.com/hartbit/ohmyzsh/blob/master/plugins/nvm/nvm.plugin.zsh#L19

I attempted to fix the problem by using the `HOMEBREW_REPOSITORY` environment variable, but it's not present if Homebrew is loaded after zsh. The solution would probably be to check both locations then, but I don't feel proficient enough in Bash to fix the problem.

### Steps to reproduce

1. Install NVM via Homebrew on a Apple Silicon mac.
2. Enable the nvm plugin.
3. Notice that nvm is not loaded.

### Expected behavior

NVM should load.

_No response_

### OS / Linux distribution

macOS Big Sur with Apple Silicon

### Zsh version

5.8

### Oh My Zsh version

master (4a988c4)

### Terminal emulator

iTerm2

_No response_

### Additional context

_No response_

carlosala:
## Standards checklist:

## Changes:

Closes #10784

### DIFF
--- a/plugins/nvm/nvm.plugin.zsh
+++ b/plugins/nvm/nvm.plugin.zsh
@@ -16,7 +16,7 @@ if [[ -f "$NVM_DIR/nvm.sh" ]]; then
 else
   # Otherwise try to load nvm installed via Homebrew
   # User can set this if they have an unusual Homebrew setup
-  NVM_HOMEBREW="${NVM_HOMEBREW:-/usr/local/opt/nvm}"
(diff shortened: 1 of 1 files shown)


mcornella (approved):

jinwyp:
On centos or AlmaLinux , This cause a bug because brew is not found on AlmaLinux

PR #10881, opened 2022-04-23
feat(def): Add def Plugin

thevinter:
## Standards checklist:

## Changes:

- Added new `def` plugin to run default commands in a directory

### DIFF
--- a/plugins/def/README.md
+++ b/plugins/def/README.md
@@ -1,0 +1,26 @@
+### def
+This plugin is used to specify and run a default command in any directory of your choice
+
+To use it you can either run `def init` to create the default config file or create a `.def` file on a folder by folder basis
+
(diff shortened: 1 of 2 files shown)


thevinter:
Hi, I wonder if there's a chance of this being reviewed! Thanks :)

robbyrussell:
Thanks for this, and sorry it took so long to get you an answer, especially after you followed up.

The part I can't get past is that it sources a `.def` file from whatever directory you happen to be in. direnv and autoenv do something similar but gate it behind an explicit allow-list per directory, and without that it's a sharp edge for anyone who cds into a repo they didn't write.

Closing for that reason rather than because the idea was bad. It works well in `$ZSH_CUSTOM`.

PR #10896, opened 2022-04-28
fix(github): deprecate the `git.io` command

SukkaW:
## Standards checklist:

## Changes:

- Deprecating the `git.io` command.

## Other comments:

GitHub no longer accepts the new URL: https://github.blog/changelog/2022-01-11-git-io-no-longer-accepts-new-urls/

### DIFF
--- a/plugins/github/github.plugin.zsh
+++ b/plugins/github/github.plugin.zsh
@@ -68,8 +68,9 @@ exist_gh() { # [DIRECTORY]
 # documentation: https://github.com/blog/985-git-io-github-url-shortener
 #
 git.io() {
-  emulate -L zsh
-  curl -i -s https://git.io -F "url=$1" | grep "Location" | cut -f 2 -d " "
+  # emulate -L zsh
+  # curl -i -s https://git.io -F "url=$1" | grep "Location" | cut -f 2 -d " "
+  print -u2 ${(%):-"%F{yellow}%BThe \`git.io\` is deprecated.%b\nView the announcement made by GitHub: https://github.blog/changelog/2022-01-11-git-io-no-longer-accepts-new-urls/%f"}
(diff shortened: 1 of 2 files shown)


carlosala (approved):

plugins/github/github.plugin.zsh
```diff
@@ -68,8 +68,9 @@ exist_gh() { # [DIRECTORY]
 # documentation: https://github.com/blog/985-git-io-github-url-shortener
 #
 git.io() {
-  emulate -L zsh
-  curl -i -s https://git.io -F "url=$1" | grep "Location" | cut -f 2 -d " "
+  # emulate -L zsh
```
jmeridth, line 71: What do you think about deleting these lines instead of commenting them out?

jmeridth (commented):
LGTM but prefer deletion over commenting out. Wdyt?

plugins/github/github.plugin.zsh
SukkaW: I just commented out the code for archive purposes.

This way of submitting links is never officially documented by GitHub, so even if `git.io` stop accepting new link I'd like to keep it as a reference.

jmeridth: That is the purpose source control. If the comment doesn't have enough info maybe add more detail?

SukkaW: > That is the purpose of source control. If the comment doesn't have enough info maybe add more detail?

Sounds fair.

> To be clear. I'm talking about the code lines you commented out. Keeping the git.io comment is good.

Yeah, I know that.

mcornella:
Since this is going to be removed in a month or two it doesn't warrant a maintainer edit.

PR #10897, opened 2022-04-28
feat(git): fix logic in `parse_git_dirty()`: Do not output when `oh-my-zsh.hide-dirty`

bixycler:
Do not output when `oh-my-zsh.hide-dirty`

## Standards checklist:

## Changes:

- Previous behavior: output `$ZSH_THEME_GIT_PROMPT_CLEAN` when `oh-my-zsh.hide-dirty`
  Issue: in some big repo that we don't want waste CPU on `git status`, it is falsely shown as "clean", e.g. with an "🆗" mark
- New behavior: output nothing when `oh-my-zsh.hide-dirty`
  Rationale: the dirty state is unknown, so it is neither "clean" nor "dirty"

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -37,22 +37,23 @@ function parse_git_dirty() {
   local STATUS
   local -a FLAGS
   FLAGS=('--porcelain')
-  if [[ "$(__git_prompt_git config --get oh-my-zsh.hide-dirty)" != "1" ]]; then
-    if [[ "${DISABLE_UNTRACKED_FILES_DIRTY:-}" == "true" ]]; then
-      FLAGS+='--untracked-files=no'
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks!

mcornella:
> Note that this breaks various themes such as `robbyrussell`, so that the trailing paren gets cut off if the early return is hit.

Had a bit of time to check it out, turns out some themes are using the `ZSH_THEME_GIT_PROMPT_CLEAN` variable as the suffix because the dirty mark is put after the suffix, i.e. after the closing parenthesis (`awesomepanda`, `robbyrussell`, ...).

I'm reverting this PR and pausing the solution until we have a solution to this use case.

PR #10906, opened 2022-05-02
feat(autojump): add `$XDG_DATA_HOME` to config search path

flyingice:
A user may prefer to install autojump to ~/.local (following XDG base
directory specification) to keep $HOME clean.

## Standards checklist:

## Changes:

- Add $XDG_DATA_HOME to the list of search paths for autojump config files

## Other comments:

...

### DIFF
--- a/plugins/autojump/autojump.plugin.zsh
+++ b/plugins/autojump/autojump.plugin.zsh
@@ -1,5 +1,6 @@
 declare -a autojump_paths
 autojump_paths=(
+  ${XDG_DATA_HOME:-$HOME/.local/share}/autojump/autojump.zsh # XDG base directory specification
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it sat four years without a reply. This is a real bug: anyone who installed autojump with `./install.py -d ~/.local` gets "autojump not found" and `j` never works, and nothing in the plugin looks there today.

Two small things and I'll take it. It needs a rebase, since the search-path list has picked up several entries since 2022 and this conflicts now. And I'd put the new entry after the `$HOME` paths rather than first, so someone who has both a `~/.autojump` and an XDG copy keeps getting the one they get today.

If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll finish it and credit you.

PR #10918, opened 2022-05-08
Make the `toolbox` plugin more useful

overlisted:
## Standards checklist:

## Changes:

- Makes the plugin actually run when you add it to `plugins` (the original author forgot to change the filename?)
- Adds an alias (`tb`) for `toolbox enter`

## Other comments:
I made the alias so short because `toolbox enter` is a command that Toolbox users run very often, so it would be convenient if it was very short.


mcornella:
Apologies for the delay. Thanks @overlisted!

PR #10919, opened 2022-05-08
perf(archlinux): improve `paclist` and `pacdisowned`

WH-2099:
## Standards checklist:

### 1. Optimize the implementation of `pacdisowned`
- Use `mktemp` to create the temporary directory.
In the Archlinux environment, **`mktemp` is part of the base installation**, so there is no need to worry about its existence.

- Remove search path `/bin`, `/sbin` and `/lib`
In the Archlinux environment, `/bin`, `/sbin`, `/lib` and `/lib64` (which even not mentioned in the current version) are just **soft links** to the folders under /usr.
```bash
$ ls -ld /bin /sbin /lib /lib64
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /bin -> usr/bin
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /lib -> usr/lib
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /lib64 -> usr/lib
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /sbin -> usr/bin
```

- The command string passed to the `trap` command should use **double quotes** to ensure that parameter expansion inside is done immediately (`tmp_dir` is declared as `local`).

- Temporary files should be cleaned up **as soon as possible** after use, instead of waiting for the `trap` to take effect.

### 2. Use `pacman` rather than `expac` in `paclist`
The **`expac`** command is a community repository, **usually not installed** by users.
*Using only  `pacman`, the output information is more useful and more acceptable to native users.*

## Other comments:
None


ratijas:
Hi, sorry for the delay.

Changes look sane to me. I'd gladly merge them, thanks!

On a technical note: could you place reset your master branch (the one you used to send a PR from) to only include actual changes without the merge commit on top? Normally you'd create a new branch instead of reusing `master` in git. Also, I know this may sound annoying, but please modify commit messages so they follow local standards [outlined here](https://github.com/ohmyzsh/ohmyzsh/blob/4c82a2eedf0c43d47601ffa8b0303ed1326fab8f/CONTRIBUTING.md#format) — you can use [git rebase tool](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message) or any git GUI of your choice :)

PR #10939, opened 2022-05-17
Fix completion error for oc version 4.9 and 4.10

jvarela01:
## Standards checklist:

## Changes:

- Added sed command to replace the function definitions in source oc completion zsh (https://bugzilla.redhat.com/show_bug.cgi?id=2024427)

### DIFF
--- a/plugins/oc/README.md
+++ b/plugins/oc/README.md
@@ -11,3 +11,4 @@ plugins=(... oc)
 ## Contributors
 
 + [kevinkirkup](https://github.com/kevinkirkup) - Plugin Author
++ [jvarela01] (https://github.com/jvarela01) - Fix pluging completion for oc version 4.9 and 4.10
--- a/plugins/oc/oc.plugin.zsh
+++ b/plugins/oc/oc.plugin.zsh
@@ -1,7 +1,8 @@
 # Autocompletion for oc, the command line interface for OpenShift
 #
 # Author: https://github.com/kevinkirkup
+# Added correction to bug https://bugzilla.redhat.com/show_bug.cgi?id=2024427 for oc clients 4.9 and 4.10
 
 if [ $commands[oc] ]; then
-  source <(oc completion zsh)
+  source <(oc completion zsh | sed -e 's/kubectl/oc/g')
 fi


mcornella:
Can you post the output of `oc completion zsh` under these versions? I fear your `sed` command might be too general, so I'd like to hardcode the full line regex.

jvarela01:
Sure! This is the output from `oc completion zsh` command

```zsh
#compdef kubectl
compdef _kubectl kubectl

#compdef _oc oc

# zsh completion for oc                                   -*- shell-script -*-

__oc_debug()
{
    local file="$BASH_COMP_DEBUG_FILE"
    if [[ -n ${file} ]]; then
        echo "$*" >> "${file}"
    fi
}

_oc()
{
    local shellCompDirectiveError=1
    local shellCompDirectiveNoSpace=2
    local shellCompDirectiveNoFileComp=4
    local shellCompDirectiveFilterFileExt=8
    local shellCompDirectiveFilterDirs=16

    local lastParam lastChar flagPrefix requestComp out directive comp lastComp noSpace
    local -a completions

    __oc_debug "\n========= starting completion logic =========="
    __oc_debug "CURRENT: ${CURRENT}, words[*]: ${words[*]}"

    # The user could have moved the cursor backwards on the command-line.
    # We need to trigger completion from the $CURRENT location, so we need
    # to truncate the command-line ($words) up to the $CURRENT location.
    # (We cannot use $CURSOR as its value does not work when a command is an alias.)
    words=("${=words[1,CURRENT]}")
    __oc_debug "Truncated words[*]: ${words[*]},"

    lastParam=${words[-1]}
    lastChar=${lastParam[-1]}
    __oc_debug "lastParam: ${lastParam}, lastChar: ${lastChar}"

    # For zsh, when completing a flag with an = (e.g., oc -n=<TAB>)
    # completions must be prefixed with the flag
    setopt local_options BASH_REMATCH
    if [[ "${lastParam}" =~ '-.*=' ]]; then
        # We are dealing with a flag with an =
        flagPrefix="-P ${BASH_REMATCH}"
    fi

    # Prepare the command to obtain completions
    requestComp="${words[1]} __complete ${words[2,-1]}"
    if [ "${lastChar}" = "" ]; then
        # If the last parameter is complete (there is a space following it)
        # We add an extra empty parameter so we can indicate this to the go completion code.
        __oc_debug "Adding extra empty parameter"
        requestComp="${requestComp} \"\""
    fi

    __oc_debug "About to call: eval ${requestComp}"

    # Use eval to handle any environment variables and such
    out=$(eval ${requestComp} 2>/dev/null)
    __oc_debug "completion output: ${out}"

    # Extract the directive integer following a : from the last line
    local lastLine
    while IFS='\n' read -r line; do
        lastLine=${line}
    done < <(printf "%s\n" "${out[@]}")
    __oc_debug "last line: ${lastLine}"

    if [ "${lastLine[1]}" = : ]; then
        directive=${lastLine[2,-1]}
        # Remove the directive including the : and the newline
        local suffix
        (( suffix=${#lastLine}+2))
        out=${out[1,-$suffix]}
    else
        # There is no directive specified.  Leave $out as is.
        __oc_debug "No directive found.  Setting do default"
        directive=0
    fi

    __oc_debug "directive: ${directive}"
    __oc_debug "completions: ${out}"
    __oc_debug "flagPrefix: ${flagPrefix}"

    if [ $((directive & shellCompDirectiveError)) -ne 0 ]; then
        __oc_debug "Completion received error. Ignoring completions."
        return
    fi

    while IFS='\n' read -r comp; do
        if [ -n "$comp" ]; then
            # If requested, completions are returned with a description.
            # The description is preceded by a TAB character.
            # For zsh's _describe, we need to use a : instead of a TAB.
            # We first need to escape any : as part of the completion itself.
            comp=${comp//:/\\:}

            local tab=$(printf '\t')
            comp=${comp//$tab/:}

            __oc_debug "Adding completion: ${comp}"
            completions+=${comp}
            lastComp=$comp
        fi
    done < <(printf "%s\n" "${out[@]}")

    if [ $((directive & shellCompDirectiveNoSpace)) -ne 0 ]; then
        __oc_debug "Activating nospace."
        noSpace="-S ''"
    fi

    if [ $((directive & shellCompDirectiveFilterFileExt)) -ne 0 ]; then
        # File extension filtering
        local filteringCmd
        filteringCmd='_files'
        for filter in ${completions[@]}; do
            if [ ${filter[1]} != '*' ]; then
                # zsh requires a glob pattern to do file filtering
                filter="\*.$filter"
            fi
            filteringCmd+=" -g $filter"
        done
        filteringCmd+=" ${flagPrefix}"

        __oc_debug "File filtering command: $filteringCmd"
        _arguments '*:filename:'"$filteringCmd"
    elif [ $((directive & shellCompDirectiveFilterDirs)) -ne 0 ]; then
        # File completion for directories only
        local subDir
        subdir="${completions[1]}"
        if [ -n "$subdir" ]; then
            __oc_debug "Listing directories in $subdir"
            pushd "${subdir}" >/dev/null 2>&1
        else
            __oc_debug "Listing directories in ."
        fi

        local result
        _arguments '*:dirname:_files -/'" ${flagPrefix}"
        result=$?
        if [ -n "$subdir" ]; then
            popd >/dev/null 2>&1
        fi
        return $result
    else
        __oc_debug "Calling _describe"
        if eval _describe "completions" completions $flagPrefix $noSpace; then
            __oc_debug "_describe found some completions"

            # Return the success of having called _describe
            return 0
        else
            __oc_debug "_describe did not find completions."
            __oc_debug "Checking if we should do file completion."
            if [ $((directive & shellCompDirectiveNoFileComp)) -ne 0 ]; then
                __oc_debug "deactivating file completion"

                # We must return an error code here to let zsh know that there were no
                # completions found by _describe; this is what will trigger other
                # matching algorithms to attempt to find completions.
                # For example zsh can match letters in the middle of words.
                return 1
            else
                # Perform file completion
                __oc_debug "Activating file completion"

                # We must return the result of this command, so it must be the
                # last command, or else we must store its result to return it.
                _arguments '*:filename:_files'" ${flagPrefix}"
            fi
        fi
    fi
}

# don't run the completion function when being source-ed or eval-ed
if [ "$funcstack[1]" = "_oc" ]; then
	_oc
fi
```

faelin:
My above PR #10973 uses an alternate fix (described here: https://bytemeta.vip/repo/openshift/oc/issues/968) that fixes the `oc` completions without modifying the completions code at all.

mcornella:
Went with #10973 for being simpler and future-proof. Thanks both for the contribution!

PR #9063, opened 2020-06-25
[plugin: tmux] add default session name when autostart enabled

ShahinSorkh:
## Standards checklist:

- [x] The PR title is descriptive.
- [ ] The PR doesn't replicate another PR which is already open. (can be a subset of #7999)
- [x] I have read the contribution guide and followed all the instructions.
- [x] The code follows the code style guide detailed in the wiki.
- [x] The code is mine or it's from somewhere with an MIT-compatible license.
- [x] The code is efficient, to the best of my ability, and does not waste computer resources.
- [x] The code is stable and I have tested it myself, to the best of my abilities.

## Changes:

- Start tmux session with a default `session-name` if provided

## Other comments:

Similar behavior can be seen in [Prezto tmux module](https://github.com/sorin-ionescu/prezto/blob/master/modules/tmux/init.zsh)

PR #10946, opened 2022-05-22
fix(tmux): use `$...SESSION_NAME` while attaching

izumrudik:
## Standards checklist:

- [x] The PR doesn't replicate another PR which is already open. (Extends #9063)

## Changes:
```zsh
  if [[ -n "$ZSH_TMUX_DEFAULT_SESSION_NAME" ]]; then
    [[ "$ZSH_TMUX_AUTOCONNECT" == "true" ]] && $tmux_cmd attach -t $ZSH_TMUX_DEFAULT_SESSION_NAME
  else
    [[ "$ZSH_TMUX_AUTOCONNECT" == "true" ]] && $tmux_cmd attach
  fi
```
instead of 
```zsh
  [[ "$ZSH_TMUX_AUTOCONNECT" == "true" ]] && $tmux_cmd attach
```
Also fixed indentation, to be consistent inside the file.
## Other comments:
`$ZSH_TMUX_DEFAULT_SESSION_NAME` was not used when attaching to existing session. Now it is used.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -73,7 +73,11 @@ function _zsh_tmux_plugin_run() {
   [[ "$ZSH_TMUX_UNICODE" == "true" ]] && tmux_cmd+=(-u)
 
   # Try to connect to an existing session.
-  [[ "$ZSH_TMUX_AUTOCONNECT" == "true" ]] && $tmux_cmd attach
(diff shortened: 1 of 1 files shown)


bdwalton (commented):
WDYT about having ZSH_TMUX_DEFAULT_SESSION_NAME do something like default to "main" so that you don't need the additional branching logic everywhere?

izumrudik:
In that case user would be limited to always use some name wich could limit use cases for this extension and will be backwards incompatible. I don't think it's really worth for this small of an improvement in readability to sacrifice functionality

Issue #10925, opened 2022-05-11
`clipcopy` with `wl-clipboard` and `xclip` does not free a TTY


dluciv:
### Describe the bug

It is known fact that when copying with `wl-copy` in Wayland or `xclip` in X, the both tools go to the background to keep clipboard data owned by themselves, but do not detach from a TTY, so terminals do not close automatically even after you quit a shell.

The reason for OMZ `clipcopy` to suffer from it is [here for `wl-clipboard`](https://github.com/ohmyzsh/ohmyzsh/blob/8f56a8bdf39d7727ab0e220f0164f78c77f9c50e/lib/clipboard.zsh#L61) and [here for `xclip`](https://github.com/ohmyzsh/ohmyzsh/blob/8f56a8bdf39d7727ab0e220f0164f78c77f9c50e/lib/clipboard.zsh#L64). In my opinion, such a poor behavior is not an Oh My Zsh bug, but `xclip` and `wl-copy` do this for ages, so it looks better to workaround it.

In opposite, `xsel` detaches from a TTY completely after copying.

### Steps to reproduce

On any modern Linux distribution with `wl-clipboard` and `xclip` installed try:

To reproduce with `xclip`
1. Boot into X session or `unset WAYLAND_DISPLAY`
2. `echo 123 | clipcopy`
3. Quit shell
4. Terminal will keep running until you copy something else or kill `xclip` or close it explicitly

To reproduce with `wl-clipboard`
1. Boot into Wayland session or `unset DISPLAY`
2. `echo 123 | clipcopy`
3. Quit shell
4. Terminal will keep running until you copy something else or kill `wl-copy` or close it explicitly

### Expected behavior

`clipcopy` should detach from TTY

### OS / Linux distribution

Arch Linux 

### Zsh version

5.8.1

### Oh My Zsh version

master

### Terminal emulator

kitty

### Additional context

Suggested changes are in PR #10953.

For Wayland, change:

    wl-copy < "${1:-/dev/stdin}"

to:

    cat "${1:-/dev/stdin}" | nohup wl-copy &>/dev/null

or may be even bettter to [Zsh-specific disowning launch](https://unix.stackexchange.com/questions/243315/in-zsh-whats-the-difference-between-and/243316#243316):

    cat "${1:-/dev/stdin}" | { wl-copy &>/dev/null &| }

And the respectively for X with `xclip`:

    xclip -in -selection clipboard < "${1:-/dev/stdin}"

to:

[...]

PR #10953, opened 2022-05-26
Fix for #10925 — free TTY

dluciv:
Fixing #10925. Launching `xclip` and `wl-copy` detached to avoid waiting for them to finish (and keeping terminal open) after quitting the shell.

## Standards checklist:

## Changes:

- Launchng `wl-copy` in the way to free TTY after copying.

## Other comments:

- Wastes a bit of resources, keeping `wl-copy` or `xclip` running until something else is copied.
- Using separate `cat` and shell braces was the only option finally working for me, I will be glad to see the better way if any.

### DIFF
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -58,10 +58,10 @@ function detect-clipboard() {
     function clipcopy() { cat "${1:-/dev/stdin}" > /dev/clipboard; }
     function clippaste() { cat /dev/clipboard; }
   elif [ -n "${WAYLAND_DISPLAY:-}" ] && (( ${+commands[wl-copy]} )) && (( ${+commands[wl-paste]} )); then
-    function clipcopy() { wl-copy < "${1:-/dev/stdin}"; }
+    function clipcopy() { cat "${1:-/dev/stdin}" | { wl-copy &>/dev/null &| }; }
     function clippaste() { wl-paste; }
   elif [ -n "${DISPLAY:-}" ] && (( ${+commands[xclip]} )); then
-    function clipcopy() { xclip -in -selection clipboard < "${1:-/dev/stdin}"; }
+    function clipcopy() { cat "${1:-/dev/stdin}" | { xclip -selection clipboard -in &>/dev/null &| }; }
     function clippaste() { xclip -out -selection clipboard; }
   elif [ -n "${DISPLAY:-}" ] && (( ${+commands[xsel]} )); then
     function clipcopy() { xsel --clipboard --input < "${1:-/dev/stdin}"; }


carlosala:
I like that! It's an important fix in my opinion, let's ask @mcornella!

carlosala (approved):

mcornella:
I haven't found this to solve the issue with `xclip`. The terminal tab still hangs for a bit when closing it and then the xclip process keeps alive without an attached tty:

Captura de pantalla de 2022-06-12 17-23-54

_In the comment it should say pts/2_
Captura de pantalla de 2022-06-12 17-24-34

BTW, even though the screenshot shows the function using `nohup`, the same happens using `&|`.

I've even tried [running it with `setsid`](https://superuser.com/a/1430883/545959) so that it does not have a tty, but the problem is not that, it's that it is a child process of the terminal emulator (in my case, `gnome-terminal-server`):

Captura de pantalla de 2022-06-12 18-07-30

It seems like the best workaround is to prefer `xsel` over `xclip`: https://github.com/kovidgoyal/kitty/issues/828.

----

As per `wl-copy`, I haven't had the time to test it out yet.

mcornella:
For `wl-copy`, in my tests (Fedora 35 Silverblue) it looked like the only needed change was to silence output with `&>/dev/null`, but using `&|` as well doesn't cause any issues.

```sh
function clipcopy() { wl-copy < "${1:-/dev/stdin}" &>/dev/null &|; }
```

The extra cat was not needed in my tests, but please correct me and provide a reproduction case where that is crucial for the fix to work.

dluciv:
@mcornella Thank you for such a comprehansive reply!

As for `xclip`, it is really Kitty's oddity (or a feature) to have such a deafult setting. Gnome Terminal, XFCE Terminal and others do actually close after small pause. Not sure about what is better: to do some overkill as I suggested or to give up on such uncommon behavior like Kitty's (which can be [overrided via settings](https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.close_on_child_death) though).

As for `function clipcopy() { wl-copy < "${1:-/dev/stdin}" &>/dev/null &|; }` — tested it on the both Kitty and Gnome Terminal, and it does not even try to input anything from TTY in my case. But when I do `echo 123 | clipcopy`, it does its work nicely. That's why I used `cat` explicitly. Not sure why, but this helped...

So I'll be glad to improve my PR, but there is still no confidence if it will really benefit from the both changes =).

mcornella:
> As for `function clipcopy() { wl-copy < "${1:-/dev/stdin}" &>/dev/null &|; }` — tested it with the both Kitty and Gnome Terminal by just invoking it, and it does not even try to input anything from TTY in my case. But when I do `echo 123 | clipcopy`

Oh, that was supposed to be the only way to be used, but it does seem like an improvement to also allow waiting for input from the user to send to the clipboard. Many other clipboard methods do not allow this as well, so I think I'll separate both changes into different commits.

PR #10956, opened 2022-05-27
:pencil2: Change command 'lstheme' to 'themels', more semantic.

tonycody:
:sparkles: add the command `themes`,switch the theme and save.

Signed-off-by: XS <tonycody@qq.com>

## Changes:
- [*]Change command 'lstheme' to 'themels', more semantic.
- [+]add the command `themes`,switch the theme and save.


rwmitchell:
ls is a common command. lstheme is more easily discovered with ls[TAB] than having it on the end.

PR #10961, opened 2022-05-29
improvement(mercurial): modified `function hg_prompt_info() {...}` to be more consistent and similar in appearance with git-prompt plugin.

HimDek:
## Standards checklist:

## Changes:

- Added `function update_hg_vars() {...}` to [mercurial plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/mercurial/mercurial.plugin.zsh) which stores information about the mercurial repo into suitable variables.
- Modified `function hg_prompt_info() {...}` in [mercurial plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/mercurial/mercurial.plugin.zsh) which now shows the following info:
  - revision@branch
  - no. of commits ahead of origin (behind is disabled for now because it is too slow)
  - no. of known files modified
  - no. of new files added (by `hg add` command)
  - no. of known files removed (by `hg remove` command)
  - no. of known files deleted (by command other than hg)
  - no. of unknown files
  - whether local is clean
- Removed `function hg_dirty() {...}` from [mercurial plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/mercurial/mercurial.plugin.zsh) because it is no longer required and its features are handeled inside `function hg_prompt_info() {...}`.
- Added `# Default values for the appearance of the prompt.` to [mercurial plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/mercurial/mercurial.plugin.zsh)
- Updated [mercurial plugin README](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/mercurial/README.md) to show information about and the default values of the new `ZSH_THEME_HG_PROMPT_*` variables

## Other comments:

...


robbyrussell:
Thanks for this, and sorry it sat for four years without an answer. I'd like to take another look, but there are two things I'd want to sort out first.

The prompt cost is the big one. We moved `hg_prompt_info` to reading `.hg/branch` off disk on purpose (e52584c90, 4119f5300) because shelling out on every redraw was slow enough that people noticed. This version runs five `hg` commands per redraw. Is there a way to get the same output while caching, or at least keeping the common case off the subprocess path?

The other is `hg_dirty` and `ZSH_THEME_HG_PROMPT_DIRTY`. Four bundled themes still set that variable (bira, af-magic, minimal, afowler), so dropping it changes those prompts with nothing to explain why. Could the new output be opt-in, with today's behavior as the default?

PR #10965, opened 2022-05-31
✨ add the command themes,switch the theme and save.

tonycody:
Changes:
[+]add the command themes,switch the theme and save.


tonycody (commented):
fix

mcornella:
The `themes` plugin should be deprecated in favor of the `omz theme` command: https://github.com/ohmyzsh/ohmyzsh/blob/8168ec0174e7e3212be20ecc74810155772abff1/lib/cli.zsh#L686-L750

Thank you for the contribution though!

PR #10986, opened 2022-06-11
feat(tmux): add `$ZSH_TMUX_SKIP_CONFIG`

### LINKED ISSUE #10367: Optionally load the config defined by ZSH_TMUX_CONFIG for all tmux aliases

Tmux plugin

By default, the aliases to the tmux wrapper function do not append the `-f "$ZSH_TMUX_CONFIG"` parameter, meaning only a new unnamed session started with just the `tmux` command loads a user defined configuration file.

As a user, I like to have multiple tmux sessions, which respect my custom configuration, that I can switch between for projects or a different context.

Rather than interfering with current behaviour and enforcing an option to load a config file for all aliases, I would suggest making an additional boolean plugin parameter that enables the value of **ZSH_TMUX_CONFIG** to be appended to all or none of the aliases.

An alternative is to override or append the `-f "$ZSH_TMUX_CONFIG"` parameter in a self maintained aliases list (which in my opinion makes some of this plugin redundant)

### Additional context

_No response_

### Related Issues

_No response_

thanhnguyen2187:
Resolve #10367

## Standards checklist:

## Changes:

- Add `$ZSH_TMUX_SKIP_CONFIG` to skip `$ZSH_TMUX_CONFIG` usage

## Other comments:

...

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -27,6 +27,8 @@ fi
 : ${ZSH_TMUX_FIXTERM_WITH_256COLOR:=screen-256color}
 # Set the configuration path
 : ${ZSH_TMUX_CONFIG:=$HOME/.tmux.conf}
+# Skip $ZSH_TMUX_CONFIG usage
+: ${ZSH_TMUX_SKIP_CONFIG:=false}
 # Set -u option to support unicode
 : ${ZSH_TMUX_UNICODE:=false}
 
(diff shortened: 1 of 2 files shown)


plugins/tmux/tmux.plugin.zsh
```diff
@@ -77,9 +79,9 @@ function _zsh_tmux_plugin_run() {
 
   # If failed, just run tmux, fixing the TERM variable if requested.
   if [[ $? -ne 0 ]]; then
-    if [[ "$ZSH_TMUX_FIXTERM" == "true" ]]; then
+    if [[ "$ZSH_TMUX_FIXTERM" == "true" "$ZSH_TMUX_SKIP_CONFIG" == "false" ]]; then
```
bhushan, line 82:
```suggestion
    if [[ "$ZSH_TMUX_FIXTERM" == "true" && "$ZSH_TMUX_SKIP_CONFIG" == "false" ]]; then
```

bhushan (changes requested):

njfamirm (approved):

robbyrussell:
Thanks for this, and sorry it sat unanswered for so long. The config lookup was rewritten in #11103, so the plugin now finds `$XDG_CONFIG_HOME/tmux/tmux.conf` on its own, and #10367 was closed last week as not achievable.

Closing for that reason rather than because the idea was wrong.

PR #10992, opened 2022-06-14
fix(emacs): fix check for open frames w/ extra output

### LINKED ISSUE #10991: emacsclient.sh makes assumptions on emacs output that don't hold
### Describe the bug

Using the `emacs` plugin, which adds an `e` alias, which calls `oh-my-zsh/plugins/emacs/emacsclient.sh`.

Logic in that script: You call it with a file to be opened in emacs.

1. use the system's `emacsclient` to execute a bit of list that lists all active frames (i.e., what you'd call "window" in a modern GUI)
2. if there's empty string or `nil` returned, create a new frame, so show a new (X11/windows/…) window<br> if there's none, display in current frame

Sadly, the check fails: if you add a few debug `echos`, you'll get something funky happening:

```shell
  # Only create another X frame if there isn't one present
  if [ -z "${frames}" -o "${frames}" = 'nil' ]; then
    echo "frames ($frames) evaluated empty or nil"
    emacsclient --alternate-editor "" --create-frame "$@"
    return $?
  fi

  echo "frames ($frames) evaluated non-nil, no new frame"
  emacsclient --alternate-editor "" "$@"
```

which then happily proceeds to tell us, at least when run without an emacs daemon already running,

```text
frames (nil) evaluated non-nil, no new frame
```
Well, I was flabbergasted, `nil != nil`? And on the second run, suddenly `nil = nil`!

Turns out, when you hexdump the non-nil `nil`, i.e., the one that emacs spits out on the second (and subsequent) run, there's a bunch of invisible characters being printed. 

```text
00000000: 1b5d 3132 3b23 6666 6666 6666 6666 3030  .]12;#ffffffff00
00000010: 3030 076e 696c 0a                        00.nil.
```
So, this is an unlucky combination of emacs (with plugins, but who uses vanilla emacs?) in modern versions starting some kind of TUI without much ado, which already prints something, only to be then *followed* by something printable, namely the contents we instructed to get via lisp.

### Steps to reproduce

1. enable plugin `emacs`
2. (make sure no emacs daemon is running.) Run `e ~/.zshrc`
3. Instead of starting emacs, opening a new GUI window and opening .zshrc in that, you get a quickly popping up terminal-mode emacs (which you did not request), which instantly closes again

### Expected behavior

1. enable plugin `emacs`
2. (make sure no emacs daemon is running.) Run `e ~/.zshrc`
3. opens a new GUI window and .zshrc in that

_No response_

### OS / Linux distribution

Fedora 36 Linux, 

### Zsh version

5.8.1

### Oh My Zsh version

master (50550c70)

### Terminal emulator

alacritty

_No response_

### Additional context

A solution could either be

1. fix the lisp command passed to emacsclient to list frames: `(delete 't (mapcar 'framep (frame-list)))` to instead pipe the result into a (temporary) file. Read that file instead of stdout.
2. Strip all characters up to and including the 0x07 byte (ASCII Bell symbol) in the `$frames` string

If I knew enough lisp to make this not only work, but work *reliable* across platforms, I'd go for 1. Alas, I don't. So, `frames="$(emacsclient -a '' -n -e "$cmd" 2>/dev/null |sed 's/.*\x07//g' )"` it is for me for now, but it's a dirty hotfix.

marcusmueller:
Fixes #10991

This is only necessitated by some plugin or emacs itself writing to
stdout when it shouldn't. But that is out of our control - so being
tolerant becomes necessary.

Since the output in question always ends in \0x07 (ASCII BELL), we just
sed it away.

Signed-off-by: Marcus Müller <marcus@hostalia.de>

## Standards checklist:

## Changes:

- Strip the "noise" emacs starting can output, depending on emacs plugins out of our control

## Other comments:

See discussion at the end of the report in #10991 why this is necessary, and to my best ability.

### DIFF
--- a/plugins/emacs/emacsclient.sh
+++ b/plugins/emacs/emacsclient.sh
@@ -11,7 +11,7 @@ emacsfun() {
   esac
 
   # Check if there are suitable frames
-  frames="$(emacsclient -a '' -n -e "$cmd" 2>/dev/null)"
+  frames="$(emacsclient -a '' -n -e "$cmd" 2>/dev/null |sed 's/.*\x07//g' )"
 
   # Only create another X frame if there isn't one present
   if [ -z "$frames" -o "$frames" = nil ]; then


mcornella:
We could also use `[[ $(cmd) = *nil ]]` instead of `sed` if this was a zsh script. I can't reproduce the issue with default emacs, but I admittedly don't have any plugins.

I believe this solution is gonna be enough without having to refactor the whole thing, so LGTM. Thanks!

PR #10997, opened 2022-06-16
fix(nvm): trim non-printable chars from .nvmrc

Zeragamba:
## Standards checklist:

## Changes:

Ran into an issue where the nvm autoload would always try to install and use node. Turns out newline characters in my .nvmrc files caused issues with the `nvm version` check. This PR trims non-printable characters from the .nvmrc file before passing it to nvm.


carlosala:
Thanks for your contribution!

PR #11015, opened 2022-06-23
fix: fixed gl alias

bhushan:
## Standards checklist:

## Changes:

- fix: fixed wrong alias for gl

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -209,7 +209,7 @@ alias git-svn-dcommit-push='git svn dcommit && git push github $(git_main_branch
 alias gk='\gitk --all --branches &!'
 alias gke='\gitk --all $(git log -g --pretty=%h) &!'
 
-alias gl='git pull'
+alias gl='git log'
 alias glg='git log --stat'
 alias glgp='git log --stat -p'
 alias glgg='git log --graph'


jmeridth (approved):

carlosala:
`gl` has been `git pull` for a long time in the mindset of people, we shouldn't change that, it's so breaking.
But thanks for your contribution! 😉

rwmitchell:
Just to be clear, we’re opposed to fixing bad past decisions?

git has too many options to have aliases to be mnemonic for everyone, best to remove all git aliases and let each person assign the ones that works for them.

carlosala:
I don't know why should that be a bad decision. `git pull` is one of the most used commands every day, I feel like having an alias just composed by two letters to it is a great decision. Despite that, we can discuss about that if it's really a common feeling 👌🏻

rwmitchell:
Use:
gu for git pull
gp for git push

bhushan:
@carlosala do you agree to fixing it to `alias gl="git log"` ?

carlosala:
I kinda understand your point, but I think that, at the end, this may raise other issues related to changing one basic alias of `git`. Let's ask @mcornella, what do you think about?

rwmitchell:
CA,

The current 'gl' alias for 'git pull' is non-obvious, whereas 'gl' for 'git log' is.  Using 'gl' for pull was a bad decision from the beginning and not even consistent with the other aliases in the same plugin.

So it really comes down do this - do think you've already reached max omz/git plugin users and hence it is more important to cater to those (and assuming they also mentally unable to  adapt) or the new omz/git users to come is a larger group and it is more important to set things right now.

Is it time ti discard omz or does it have a long future?

PR #11017, opened 2022-06-26
chore(lib): only create `afind` alias if `ack` is installed

Hellzbellz123:
> idk if ack is supposed to exist on every distro if not then thats fine but this keeps the alias command a bit cleaner :/## Standards checklist:

## Changes:
added new if block to change alias ;/
https://patch-diff.githubusercontent.com/raw/ohmyzsh/ohmyzsh/pull/11017.diff
## Other comments:
this could probably be added withsome elses bigger merge because its really small

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -24,6 +24,7 @@ env_default 'LESS' '-R'
 ## super user alias
 alias _='sudo '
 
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
@mcornella LGTM! 😉

PR #11030, opened 2022-07-01
Update kubectl.plugin.zsh

sneakythr0ws:
If I use klf I read ALL logs of container by default. It could be a lot.
I suggest show 10 rows from tail by default

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


robbyrussell:
Thanks for this, and sorry it sat so long. Changing what an existing alias does is a different thing from adding one, and people have `klf` in muscle memory. We tried something similar in #13204 and reverted it in #13496 after the feedback, so I'm going to pass.

`alias klf='kubectl logs -f --tail=10'` in your `.zshrc` overrides ours cleanly.

PR #11034, opened 2022-07-05
fix(jonathan): set pwdsize to string display size

tangruize:
Some wide characters (e.g. CJK characters and emoji) have more than 1 size to display. If the PWD contains these characters, the prompt bar size is not correctly calculated and the first line will be broken into 2 lines.

Original style:
image

After fixing:
image

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


robbyrussell:
Thanks for this, and sorry it sat for four years. The bug still reproduces: `~/工作/项目文档/测试` measures 12 where it occupies 20 columns, and your line gives the right answer.

The branch has drifted since (the theme picked up `venvpromptsize` and `condapromptsize`), so it no longer applies. Any chance you could rebase it onto master? It should still be the same one line.

Worth noting for later, not for this PR: `rubypromptsize`, `venvpromptsize` and `condapromptsize` measure width the same way and have the same problem.

PR #11043, opened 2022-07-11
meta: removed ignored custom directory from index only.

janzenz:
## Standards checklist:

## Changes:

- Removes ignored files from custom directory so that git won't complain when these files are removed or overridden.

## Other comments:
There was request to remove example scripts from `custom/` directory and it was completed on https://github.com/ohmyzsh/ohmyzsh/commit/f5b2d9c7ce8ced5a4ec67a4ecfa877e6951124ef

This PR completes the process to remove these files from index so git won't complain if these files were removed.


carlosala:
Hi! I think these files are great for new users coming to `omz` to see the actual structure of the project. If they really annoy you, you can create a commit deleting those files and, when running `omz update`, this commit is going to be rebased and there are going to be no problems with it, as we don't change that files ofter (nearly never).
Thanks for your contribution!

PR #11045, opened 2022-07-12
docs: document new aliases policy

mcornella:
## Standards checklist:

## Changes:

- Document alias changes in CONTRIBUTING file
- Add checkbox in new PR template

## Other comments:

cc @carlosala @robbyrussell @larson-carter

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,6 +9,7 @@
 - [ ] The code is mine or it's from somewhere with an MIT-compatible license.
 - [ ] The code is efficient, to the best of my ability, and does not waste computer resources.
 - [ ] The code is stable and I have tested it myself, to the best of my abilities.
+- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below.
 
 ## Changes:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,6 +104,27 @@ maintainers) by mentioning their GitHub handle (starting with `@`) in your messa
 
 For any extensive change, such as a new plugin, you will have to find testers to +1 your PR.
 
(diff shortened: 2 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,6 +9,7 @@
 - [ ] The code is mine or it's from somewhere with an MIT-compatible license.
 - [ ] The code is efficient, to the best of my ability, and does not waste computer resources.
 - [ ] The code is stable and I have tested it myself, to the best of my abilities.
+- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below.
```
mcornella, line 12: @carlosala would this work?
```suggestion
- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users. See CONTRIBUTING file.
```

carlosala: Perfect for me! 👌🏻

larson-carter: > @carlosala would this work?
> 
> ```suggestion
> 
> - [ ] If the code introduces new aliases, I provide a valid use case for all plugin users. See CONTRIBUTING file.
> 
> ```

Just a note from me. Could we link to the file itself?

carlosala: The problem is that when you are writing the PR you have the raw display of markdown, and it's going to be a bit shitty to have that `[foo](https://bar.com)` syntax.

larson-carter: Ahh very true. Didn't take that into consideration I was thinking about it being rendered for the end user.

carlosala: looks like we can merge @mcornella

PR #11047, opened 2022-07-13
Create plugin with commands for developers

sercheo87:
## Standards checklist:

## Changes:

- Add custom aliases


carlosala:
Hi Sergio, thanks for your contribution!
We are not going to accept it because normally plugins are specific for one particular usage, we don't have almost any plugin that is for different scopes. Even more, we are now developing a new [alias policy](#11045) where we are not going to accept new aliases that are really specific, as that one of `fzf`.
Thanks for your contribution another time, and sorry for not being able to accept it!
Cheers :)

PR #11052, opened 2022-07-16
refactor(git-groh): Remove dependency for `git_current_branch`

liijunwei:
## Standards checklist:

## Changes:

- Remove dependency for `git_current_branch` in `groh` git alias

## Other comments:

...


liijunwei:
anyone have a moment to take a look? 👀

carlosala:
Hi @liijunwei!
Thanks for your contribution! Why should we remove `git_current_branch` dependency? It's actually really fast, and I don't see why should we do it. Thanks!

liijunwei:
I just realized that `git_current_branch` is in core lib of git, maybe I was wrong

But the main reason is that I find [current implementation for `git_current_branch`](https://github.com/ohmyzsh/ohmyzsh/blob/249c708ed3a4a7a63d16a6e911a46b6fb9623cbd/lib/git.zsh#L96) is too hard to understand, many indirection inside 😵 

Maybe there are some edge cases are the concern, but I'd like to suggest  this for git_current_branch
```
function git_current_branch() {
  git rev-parse --abbrev-ref HEAD
}
```

Another reason is that I prefer this tool have less dependencies(personal taste, coule be ignored 😄 )

liijunwei:
I like git plugin in omz, but I have to unalias it and redefine if myself
image

PR #11056, opened 2022-07-19
Update last updated time if there are no available updates

jnooree:
## Standards checklist:

## Changes:

- Call `update_last_updated_file` if there are no available updates.

## Other comments:

The `:omz:update frequency` configuration does not have to be exact "update" days; it is sufficient to check every `frequency` day.

This change is proposed because the `is_update_available` function accesses the network, so it consumes a significant amount of time on **every** startup until there are some upstream updates.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -177,6 +177,7 @@ function has_typed_input() {
 
   # Check if there are updates available before proceeding
   if ! is_update_available; then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry nobody replied at the time. Same one-line change landed in d889eca7 via #11617, about nine months after you filed this. You found it first. Closing this one out.

PR #11057, opened 2022-07-20
feat(jump): print the current path after jumping

bond-aspecta:
## Changes:

- Appended a `pwd` command in `jump` function.

## Other comments:

It helps to make sure I am in the right directory when I have multiple directories with the same name.


im-scared:
It would be nice to have an environment variable that lets you disable this. I definitely appreciate the value of this, but I would want this turned off.

PR #11060, opened 2022-07-21
feat(plugins/term_tab): MacOS support for completion of zsh sessions working directory

Simske:
## Standards checklist:

## Changes:

- Extended term_tab plugin to work on Mac
- Updated Readme of plugin term_tab

## Other comments:
The term_tab plugin only Linux and Solaris hosts, as the logic is OS specific.
This adds the same functionality for MacOS.

Tested on MacOS 12.2.1 ARM, zsh 5.8 (x86_64-apple-darwin21.0)

### DIFF
--- a/plugins/term_tab/README
+++ b/plugins/term_tab/README
@@ -11,6 +11,10 @@ to show you the current working directory of the other open zsh sessions.
 
 How it works:
 *************
-* It uses 'pidof zsh' to determine all zsh PIDs
-* It reads procfs to get the current working directory of this session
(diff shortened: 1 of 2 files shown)


robbyrussell:
Thanks for this, and sorry nobody replied at the time. The same `lsof` approach landed in #11391 in April 2023, so the code half of this is already in master. You got there first.

The README half never made it, so it still says the plugin only works on Solaris and Linux. I'm fixing that separately. Closing this one out.

PR #11068, opened 2022-07-24
fix(git-auto-fetch): don't override native `stat` command

Zocker1999NET:
## Standards checklist:

## Changes:

- Only load zstat builtin command from zsh/stat module

## Other comments:

According to man page `zshbuiltins` are the builtins `stat` and `zstat` synonymous, but because the builtin `stat` will shadow any `stat` executable in the path, it is recommended to only load `zstat`.

The plugin already used `zstat`.

~~I will fix the commit message ASAP to respect the Commit Guidelines.~~ **EDIT:** Changed commit message.


carlosala:
@mcornella LGTM!

mcornella:
Thanks!

PR #11070, opened 2022-07-25
feat(zoxide): Add support for zoxide init flags

dzfrias:
## Standards checklist:

## Changes:

- Add support for flags to be set for `zoxide init` in the `zoxide` plugin.

## Other comments:
These options can only be set during `zoxide init`, so they're important to include for users who want to use them.


robbyrussell:
Thanks for this, and sorry for the long silence. The main piece here, renaming the `z` command, landed in #12218 as `ZOXIDE_CMD_OVERRIDE`. That rewrote the same line this patches, so it no longer applies.

Closing it out. If you still want `--hook` or `--no-cmd` on a current version, a fresh PR against master would be small.

PR #11071, opened 2022-07-25
feat(plugins/rushjs) - Adds RushJS plugin 

Shockolate:
## Standards checklist:

## Changes:

- Adds a new plugin `rushjs` to ohmyzsh. The plugin consists of several aliases for developer convenience for those using the [RushJS](https://rushjs.io/) monorepo manager.

### DIFF
--- a/plugins/rushjs/README.md
+++ b/plugins/rushjs/README.md
@@ -1,0 +1,32 @@
+# RushJS
+
+This plugin adds aliases for frequent [rush](https://rushjs.io/) commands.
+
+## Commands
+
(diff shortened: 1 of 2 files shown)


z-hirschtritt:
+1

robbyrussell:

Before I take it further: would you be willing to keep this current if we merge it? Rush changes its command surface every so often, and a plugin nobody is watching goes quietly stale. That's the thing that holds me back on new plugins now, more than the code itself. The diff still applies cleanly and Rush is clearly alive, so this is the open question rather than anything in the patch.

If you'd rather not pick it back up after this long, that's completely fine. Just say so and I'll close it out.

Shockolate:
Hi @robbyrussell !

Thanks for reply and your explanation.

Due to the fact that this is no longer part of my toolchain, I do not intend to maintain it this plugin.

I'll close this out for you!

Thanks again,
Ted A.

PR #11073, opened 2022-07-27
fix(plugin/jump): fix text alignment when printing marks

dit-zy:
## Standards checklist:

## Changes:

the `marks` function right-justifies mark names, to make the output
easier to read. However, because of the escape sequences for colors in
the output, the printf command doesn't calculate the padding correctly.
This results in the mark names not actually being right-justified
correctly, with short names not being padded enough. E.g.:

     code -> /home/ditzy/code
     misc -> /home/ditzy/misc
    sys-conf -> /home/ditzy/.sys-conf

this commit fixes this problem by only applying padding to the actual
mark name itself, and not the name+colors. E.g.:

        code -> /home/ditzy/code
        misc -> /home/ditzy/misc
    sys-conf -> /home/ditzy/.sys-conf

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -35,12 +35,11 @@ marks() {
 			max=${#link:t}
 		fi
 	done
-	local printf_markname_template="$(printf -- "%%%us " "$max")"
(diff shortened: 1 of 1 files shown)


carlosala:
This has been solved in #11601. Thanks for your contribution!

PR #11074, opened 2022-07-27
fix(jump)!: only resolve mark symlink

dit-zy:
## Standards checklist:

## Changes:

BREAKING CHANGE

jumping previously took you to the absolute path of the marked
directory. If you create a mark to a symlinked path, then when you jump
you'll end up at the absolute version of that path:

```
$ mkdir /real-path
$ ln -s /real-path /link-path
$ cd /link-path
$ mark
Mark /link-path as link-path? (y/n) y
$ jump link-path
$ pwd
/real-path
```

this seems unintuitive to me, so this commit changes the jump function
to take you to the actual path of the mark (`/link-path` in the example
above) rather than the absolute path.

## Other comments:

although this is a breaking change, I felt conflicted about the value of adding an environment variable to preserve the old functionality, since this seems more intuitive and is such a small change. But I can totally imagine how this could inconvenience people, so if requested I'm happy to add an env var to keep the old functionality.

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -8,8 +8,10 @@
 #
 export MARKPATH=$HOME/.marks
 
+
 jump() {
-	builtin cd -P "$MARKPATH/$1" 2>/dev/null || {echo "No such mark: $1"; return 1}
+	local markpath="$(readlink $MARKPATH/$1)" || {echo "No such mark: $1"; return 1}
+	builtin cd "$markpath" 2>/dev/null || {echo "Destination does not exist for mark [$1]: $markpath"; return 2}
 }
 
 mark() {


jimweller:
I personally would appreciate this. I have symlink shortcuts to my work and personal spaces in my homedir for github that live on OneDrive which is a horrendous path. I use [includeif] in my .gitconfig to switch git profiles. Having the current `jump` go to the original path messes up my workflow (or makes the [includeifs] more numerous).

I was going to do a PR for this and found this one.

I was also going to do a PR to have env var toggles to not have mark/unmark prompt with (y/n).

carlosala:
Let's go for it, it seems to be a corner case that shouldn't break any user functionality.

carlosala (approved):
LGTM!

Issue #10004, opened 2021-06-24
add sso login to acp command on aws 


barakharyati:
currently for using SSO credentials is complicated,
```
 asp $profile
AWS sso login ##on the source profile
```
 
it would be awesome If we could run 
`acp --$SSOProfile`
 and then the plugin searches for the source profile and do SSO login on it.

robbyrussell:
Hey @barakharyati, sorry... five years and nobody ever replied to this. That's on us.

Some good news since you filed it: `asp` learned to do SSO logins a few months later, so `asp <profile> login` is the supported way to get an SSO session now (#9921, with `sso-session` block support added later in #12090). For most people that's the thing this issue was asking for.

`acp` itself still doesn't do any SSO handling — it goes straight to `aws sts get-session-token` — so if what you actually needed was for `acp` to log in on an SSO *source profile* before assuming a role, that part isn't covered. If that's your case, open a fresh issue saying so and we'll look at it as its own thing rather than keep this one going.

Closing it out so it isn't sitting in limbo. Thanks for reporting it, and sorry again for the long silence.

🤖 Co-authored with Claude Code

PR #9921, opened 2021-05-21
feat(aws): Adds the login option for AWS SSO via asp

hawkesnc:
## Standards checklist:

- [ ] The PR title is descriptive.
- [ ] The PR doesn't replicate another PR which is already open.
- [ ] I have read the contribution guide and followed all the instructions.
- [ ] The code follows the code style guide detailed in the wiki.
- [ ] The code is mine or it's from somewhere with an MIT-compatible license.
- [ ] The code is efficient, to the best of my ability, and does not waste computer resources.
- [ ] The code is stable and I have tested it myself, to the best of my abilities.

## Changes:

Allows users that have already setup their AWS_PROFILE with SSO to append `login` to the command. This subsequently runs `aws sso login` after their profile is set with `asp`.

Example: asp dev login -> Sets their profile and runs `aws sso login` immediately after

## Other comments:

...

PR #11075, opened 2022-07-28
feat(aws): add sso

hans-d:
## Standards checklist:

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -52,8 +52,25 @@ function acp() {
   local aws_secret_access_key="$(aws configure get aws_secret_access_key --profile $profile)"
   local aws_session_token="$(aws configure get aws_session_token --profile $profile)"
 
+  # Are we using SSO?
+  local sso_start_url="$(aws configure get sso_start_url --profile $profile)"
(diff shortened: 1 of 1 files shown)


ketiko:
+1

larson-carter:
I would like an explanation of the reasoning on how it would handle the actual authorization process? Is it 100% token based and will they ever expire?

hans-d:
Authorization is done completely via the sso flow:
- if token is not valid anymore (checked by doing a simple call), it will request a new token. Normally this involves displaying a URL you have to visit and enter some code, which in most cases can be opened automatically in a browser. As a user you still have to confirm the request
- if token is valid at the time of change, it will continue to use it

if the token becomes invalid/expired, one has to either do `aws sso login` or "switch" to the same profile. Expiration is tied to the sso settings within aws. (in my cases, sessions get expired)

robbyrussell:
Thanks for this, and sorry it took years to get you an answer. Most of what this does now ships through `asp`: SSO login landed in #9921 and `sso-session` support in #12090, and #10004 was closed last week.

`acp` with SSO is still a gap, but I think it needs a fresh implementation rather than this branch rebased. Closing this one out.

PR #11091, opened 2022-08-05
fix(lib): fix mismatch between ls colors and zsh tab completion colors

89iuv:
## Standards checklist:

## Changes:

- added default LS_COLORS (derived from LSCOLORS using https://geoff.greer.fm/lscolors/) if no LS_COLORS have been set
The logic for setting LS_COLORS is as follows:
1. if the user has set the value for LS_COLORS use that one unless it is not overwritten by a theme
2. If the user does not have a value for LS_COLORS then if the value can be determined from the system by using dircolors then use that one, if not then use the default LS_COLORS unless it is not overwritten by a theme

Not having a default LS_COLORS value on MacOS results in the following situation:
- ls will take the values from LSCOLORS which have a default color, or are overwritten by theams
- zsh tab completion will have no colors set because on mac there is no default value for LS_COLORS
this results in the mismatch of colors between ls and zsh tab completion

- move the setup for zsh tab completion from lib/theme-and-appearance.zsh to the end of oh-my-zsh.sh
while ls evaluates the LS_COLORS value every time it is called, it seems that zsh tab completion does not.
in this case we need to set zsh tab completion only after the theme is sourced in order to have the correct colors.

Testing was done on wsl and macos:
Wsl(theme: arrow)
no fix: wls_no_fix
with fix: wls_with_fix

MacOS(theme: arrow)
no fix: macos_no_fix
with fix: macos_with_fix

MacOS(theme: robbyrussell)
no fix: macos_no_fix_default
with fix: macos_with_fix_default

## Other comments:

This proposal is related to the following:
- https://github.com/ohmyzsh/ohmyzsh/pull/6353
- https://github.com/ohmyzsh/ohmyzsh/issues/6060

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -1,9 +1,6 @@
 # ls colors
 autoload -U colors && colors
 
-# Enable ls colors
-export LSCOLORS="Gxfxcxdxbxegedabagacad"
-
 # TODO organise this chaotic logic
 
 if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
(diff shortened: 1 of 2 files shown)


89iuv:
@mcornella any thoughts on this PR?

PR #11097, opened 2022-08-08
fix(aliases): Change `acs` to `alcs` to avoid collision with debian and ubuntu plugin aliases.

grimm26:


carlosala (approved):

mcornella:
This is a big breaking change. I need more time to think of alternatives.

PR #11102, opened 2022-08-08
feat(archlinux): order archlinux package managers by popularity

redxtech:
## Standards checklist:

## Changes:

- Change the order in which the package manager aliases are created in order to prioritize the more intentionally installed option.

## Other comments:
Currently, if you have multiple package managers installed, and you run the `upgrade` alias, it will default to `yay -Syu`, as that is the last set alias. This may seem like a sensible choice, as `yay` is the most popular AUR helper. However, because of this, it is installed by default on a lot more distributions of arch, and included by default in more install helpers. This means that if the user chooses to install one of the less popular options, it will be overwritten by `yay`'s alias. 

I have ordered the alias creation in order of popularity (from their page in the AUR), so that the least popular package manager will have priority for the `upgrade` alias. This may seem counterintuitive, but if something is less popular, it is more likely to have been installed intentionally by the user, as opposed to being included by default. This is why I believe the least popular option should, if installed, have priority for the `upgrade` alias.

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -87,6 +87,30 @@ fi
 #             AUR helpers             #
 #######################################
 
+# ordered by popularity (aur popularity score) 
+
+if (( $+commands[yay] )); then
+  alias yaconf='yay -Pg'
+  alias yaclean='yay -Sc'
+  alias yaclr='yay -Scc'
+  alias yaupg='yay -Syu'
+  alias yasu='yay -Syu --noconfirm'
+  alias yain='yay -S'
+  alias yains='yay -U'
+  alias yare='yay -R'
+  alias yarem='yay -Rns'
+  alias yarep='yay -Si'
+  alias yareps='yay -Ss'
+  alias yaloc='yay -Qi'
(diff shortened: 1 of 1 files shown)


redxtech:
Any way this can be merged? Or given feedback on what needs to be completed before merging, or rejected?

ratijas:
It's been discussed already in previous PRs that if someone's using multiple AUR helpers for some reason, then they'd be better invoking them directly or manually overriding aliases instead of relying on plugins.

IMHO ordering by popularity is not the right approach. Popularity is practically unmeasurable in the absence of consistent telemetry, and I don't believe that number of GitHub stars reflects anything at all. Besides popularity may change over time, and we don't want to shuffle the order of plugins in this list (and once again breaking the expected behavior for those who have multiple AUR helpers and still use aliases!). So, overall it's a "no" from me, sorry.

ratijas:
Ordering alphabetically would make at least some sense.  But again, it would
a) change expected priorities for existing installations;
b) screw up `git blame`, making it harder to diff and navigate history.

carlosala:
Thanks @ratijas for your review! Closing this👌🏻

PR #11106, opened 2022-08-09
feat(web-search): add support for Brave search engine

Bryan-netizen:
Added brave browser to the web search list, as well as my own alias.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/web-search/web-search.plugin.zsh
+++ b/plugins/web-search/web-search.plugin.zsh
@@ -7,23 +7,23 @@ function web_search() {
   typeset -A urls
   urls=(
     $ZSH_WEB_SEARCH_ENGINES
-    google      "https://www.google.com/search?q="
-    bing        "https://www.bing.com/search?q="
-    brave_search "https://search.brave.com/search?q="
-    yahoo       "https://search.yahoo.com/search?p="
-    duckduckgo  "https://www.duckduckgo.com/?q="
-    startpage   "https://www.startpage.com/do/search?q="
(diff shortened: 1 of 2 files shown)


plugins/web-search/web-search.plugin.zsh
```diff
@@ -47,6 +48,7 @@ function web_search() {
 
 
 alias bing='web_search bing'
+alias br='web_search brave'
```
carlosala, line 51: Could you change that to `brave`? `br` might exist in people's path, could be a tipical alias for whatever!

carlosala (changes requested):
I like it!

plugins/web-search/web-search.plugin.zsh
mcornella: `brave` might be the command to start the browser. An alternative might be `brave_search`, which is not likely to override anything, but I'm not satisfied with it either.

Bryan-netizen: Made the changes, thank you for the feedback.

larson-carter (commented):
I like this a lot as well. Good by me.

mcornella:
The `web_search` parameter can be `brave`, my point was that the alias could not be named the same way.

`br` is already a command: https://command-not-found.com/br

My advice is to name the alias something longer, like `braves` or something else like `brave` + an abbreviation of search.

Bryan-netizen:
Thanks for helping me with my 1st contribution, I really appreciate it!

PR #11107, opened 2022-08-11
feat(systemadmin): color `ip` if shell is interactive

benwaffle:
## Standards checklist:

## Changes:

- This makes `ip` use colored output, when not piping the output


benwaffle:
I guess we could use `ip -color=auto` but that's not available on Ubuntu 18.04

carlosala (approved):

carlosala:

PR #11116, opened 2022-08-14
feat(agnoster): add VRF name to prompt if available

Pelt10:
## Standards checklist:

## Changes:
- add VRF name to prompt if available("ip" command available and a vrf is actually in use)

Screenshot 2022-08-14 at 23 51 19

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -241,12 +241,25 @@ prompt_aws() {
   esac
 }
 
+
+# VRF Profile
+# Display current VRF name
+prompt_vrf() {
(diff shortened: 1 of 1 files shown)


bdwalton:
This looks technically fine (to me), but I wonder if it is something that everyone (even in a VRF environment) would want? Also, would it make sense to augment the context section with this instead? Maybe not, but asking.

robbyrussell:
Thanks for this, and sorry nobody followed up after @bdwalton's question in 2022.

Closing it. This runs `ip vrf identify` on every prompt, so the cost lands on every Linux agnoster user, and in four years nobody else asked for the segment.

PR #11119, opened 2022-08-16
chore: Add .dir-locals.el for emacs users

bdwalton:
Initially, we only enable sh-mode on a few additional .zsh-* files,
but this could be extended to incorporate the editor settings in
.editorconfig for people not using that package.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


mcornella:
We don't usually bundle editor settings, except for `.editorconfig`. This should be an editor-wide configuration in your system.

PR #11122, opened 2022-08-20
perf(fzf): speed up startup on debian

ppwwyyxx:
The code calls the external command `dpkg -s fzf` to check fzf installation and is therefore inefficient. 

I use `time zsh -i -c exit` to measure start up time. With this if condition, my startup time is 0.16-0.19s. Without this if condition, my startup time is 0.09-0.12s. This check almost doubles my start up time.

Replace it by checking the fzf binary exists at the right place.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/fzf/fzf.plugin.zsh
+++ b/plugins/fzf/fzf.plugin.zsh
@@ -59,8 +59,8 @@ function fzf_setup_using_base_dir() {
 
 
 function fzf_setup_using_debian() {
-  if (( ! $+commands[dpkg] )) || ! dpkg -s fzf &>/dev/null; then
-    # Either not a debian based distro, or no fzf installed
+  if (( ! $+commands[dpkg] )); then
+    # Not a debian based distro
     return 1
   fi
 
@@ -71,11 +71,19 @@ function fzf_setup_using_debian() {
 
   case $PREFIX in
     *com.termux*)
+      if [[ ! -f "${PREFIX}/bin/fzf" ]]; then
+        # fzf not installed
+        return 1
+      fi
       # Support Termux package
       completions="${PREFIX}/share/fzf/completion.zsh"
       key_bindings="${PREFIX}/share/fzf/key-bindings.zsh"
       ;;
     *)
+      if [[ ! -f /usr/bin/fzf ]]; then
+        # fzf not installed
+        return 1
+      fi
       # Determine completion file path: first bullseye/sid, then buster/stretch
       completions="/usr/share/doc/fzf/examples/completion.zsh"
       [[ -f "$completions" ]] || completions="/usr/share/zsh/vendor-completions/_fzf"


ppwwyyxx:
Any maintainers can look at this simple PR? thanks

ppwwyyxx:
Any maintainers can look at this simple PR? thanks

carlosala:
Thanks for your contribution!!! 😊

MarijnS95:
Haven't dissected what's happening in this script, but since this commit/PR I see:

```
fzf_setup_using_debian:source:40: no such file or directory: /usr/share/doc/fzf/examples/key-bindings.zsh
```

And consequently FZF doesn't work on <kbd>ctrl</kbd>+<kbd>r</kbd> (and other hotkeys).

Arch package has the following contents:

```console
$ pacman -Ql fzf
fzf /usr/
fzf /usr/bin/
fzf /usr/bin/fzf
fzf /usr/bin/fzf-tmux
fzf /usr/share/
fzf /usr/share/fish/
fzf /usr/share/fish/vendor_functions.d/
fzf /usr/share/fish/vendor_functions.d/fzf_key_bindings.fish
fzf /usr/share/fzf/
fzf /usr/share/fzf/completion.bash
fzf /usr/share/fzf/completion.zsh
fzf /usr/share/fzf/key-bindings.bash
fzf /usr/share/fzf/key-bindings.zsh
fzf /usr/share/licenses/
fzf /usr/share/licenses/fzf/
fzf /usr/share/licenses/fzf/LICENSE
fzf /usr/share/man/
fzf /usr/share/man/man1/
fzf /usr/share/man/man1/fzf-tmux.1.gz
fzf /usr/share/man/man1/fzf.1.gz
fzf /usr/share/vim/
fzf /usr/share/vim/vimfiles/
fzf /usr/share/vim/vimfiles/doc/
fzf /usr/share/vim/vimfiles/doc/fzf.txt
fzf /usr/share/vim/vimfiles/plugin/
fzf /usr/share/vim/vimfiles/plugin/fzf.vim
```

carlosala:
@MarijnS95 hmm, it's weird to have dpkg installed in pacman, though. Let's add a `zstyle` flag so we can opt out from it, even though dpkg is installed.

carlosala:
@MarijnS95 see a1c54e03f98b594a6fcc368c2c113d469ffaa368

MarijnS95:
@carlosala Sometimes you need to poke at `dpkg` files from an Arch-based machine :wink:

carlosala:
@MarijnS95 yeah, then I remembered it can be useful for creating some aur packages and that kind of stuff. Sorry for the regression 😉

MarijnS95:
No worries, things happen and I rather point them out than making a fuss or leaving things unsolved :+1:

ppwwyyxx:
Sorry about the regression. Shall we use a better way to test if user has a debian-based distribution? Now `commands[dpkg]` is clearly not right.

plugins/fzf/fzf.plugin.zsh
```diff
@@ -71,11 +71,19 @@ function fzf_setup_using_debian() {
...
+      if [[ ! -f "${PREFIX}/bin/fzf" ]]; then
+        # fzf not installed
+        return 1
+      fi
       # Support Termux package
       completions="${PREFIX}/share/fzf/completion.zsh"
       key_bindings="${PREFIX}/share/fzf/key-bindings.zsh"
       ;;
     *)
+      if [[ ! -f /usr/bin/fzf ]]; then
```
ppwwyyxx, line 83: maybe we should just add a check here that `/usr/share/doc/fzf/examples/` exists

MarijnS95: Hmm, something OOTB would be nice as well, instead of having to set a `skip-dpkg` option.

ppwwyyxx: Yup. Checking for `/usr/share/doc/fzf/examples/` is meant to solve this without an extra option.

wisp3rwind:
The problem also exists on Fedora 37, FWIW, so certainly not an Arch-only issue.

```
dnf repoquery --list fzf
...
/etc/bash_completion.d/fzf
/usr/bin/fzf
/usr/bin/fzf-tmux
/usr/lib/.build-id
/usr/lib/.build-id/cf
/usr/lib/.build-id/cf/779b437e390e8d3a854303a6d924e6c1c79e78
/usr/share/doc/fzf
/usr/share/doc/fzf/CHANGELOG.md
/usr/share/doc/fzf/README-VIM.md
/usr/share/doc/fzf/README.Fedora
/usr/share/doc/fzf/README.md
/usr/share/fzf
/usr/share/fzf/shell
/usr/share/fzf/shell/key-bindings.bash
/usr/share/fzf/shell/key-bindings.fish
/usr/share/fzf/shell/key-bindings.zsh
/usr/share/licenses/fzf
/usr/share/licenses/fzf/LICENSE
/usr/share/man/man1/fzf-tmux.1.gz
/usr/share/man/man1/fzf.1.gz
/usr/share/nvim/site/plugin
/usr/share/nvim/site/plugin/fzf.vim
/usr/share/vim/vimfiles/plugin
/usr/share/vim/vimfiles/plugin/fzf.vim
/usr/share/zsh/site-functions
/usr/share/zsh/site-functions/fzf
/etc/bash_completion.d/fzf
/usr/bin/fzf
/usr/bin/fzf-tmux
/usr/lib/.build-id
/usr/lib/.build-id/d0
/usr/lib/.build-id/d0/a408a08b95ad38dadfa6942b6f0f33af241354
/usr/share/doc/fzf
/usr/share/doc/fzf/CHANGELOG.md
/usr/share/doc/fzf/README-VIM.md
/usr/share/doc/fzf/README.Fedora
/usr/share/doc/fzf/README.md
/usr/share/fzf
/usr/share/fzf/shell
/usr/share/fzf/shell/key-bindings.bash
/usr/share/fzf/shell/key-bindings.fish
/usr/share/fzf/shell/key-bindings.zsh
/usr/share/licenses/fzf
/usr/share/licenses/fzf/LICENSE
/usr/share/man/man1/fzf-tmux.1.gz
/usr/share/man/man1/fzf.1.gz
/usr/share/nvim/site/plugin
/usr/share/nvim/site/plugin/fzf.vim
/usr/share/vim/vimfiles/plugin
/usr/share/vim/vimfiles/plugin/fzf.vim
/usr/share/zsh/site-functions
/usr/share/zsh/site-functions/fzf
```

mcornella:
Do any of those affected have `apt` or `apt-get` installed?

wisp3rwind:
> Do any of those affected have `apt` or `apt-get` installed?

I haven't.

mcornella:
The function also supports termux which uses `pkg`, but I see I have both `apt` and `apt-get` installed.

Some Debian-based systems have one or the other, so we should check for either of them:

```sh
(( ${+commands[apt]} || ${+commands[apt-get]} )) 
```

ppwwyyxx:
How about we do this?
```bash
  if (( ! $+commands[apt] )) || (( ! $+commands[apt-get] )) || (( ! $+commands[dpkg] )); then
    # Not a debian based distro
    return 1
  fi

# ...
  case $PREFIX in
    *com.termux*)
# ...
      ;;
    *)
      if [[ ! -f /usr/bin/fzf ]] || [[ ! -d /usr/share/doc/fzf/examples ]]; then
        # fzf or the completions are not installed
        return 1
      fi
```

The idea is just to be very conservative about checks: as long as it doesn't look like a debian (missing any command, or missing file at the right place) we `return 1`.

carlosala:
I'm concerned if that file will belong to every apt-like distro. Do you have apt installed? It seems like the definitive solution

mcornella:
```
if ((( ! $+commands[apt] )) && (( ! $+commands[apt-get] ))) || (( ! $+commands[dpkg] )); then
    # Not a debian based distro
    return 1
  fi
```

This logic will not match on a system that has dpkg installed but not any of the others, which is the current issue. This would still resort to checking the package files, which is very brittle.

mcornella:
Submitted PR #11460. Please test it out with `omz pr test 11460`.

PR #11123, opened 2022-08-20
Update example.zsh

sdondley:
Improve grammar, readability and make more concise.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


larson-carter:
Hmmm, I do believe that this somewhat clarifies the reasoning behind example. I do like some of the work, however I think we need to combine both versions. Thoughts? 

CC: @mcornella

sdondley:
Not sure what you mean by combine them. They say the same thing, but this rewrites it to be more clear.

PR #11129, opened 2022-08-23
Added plugin for tekton ci/cd CLI

linuxeroagrio:
## Standards checklist:

## Changes:

- Added Plugin for Tekton CI/CD CLI.

## Other comments:
- Added completion for tkn command


jhomer-hscl:
This would be helpful, I've written my own custom plugin and was going to submit but also found this that has been Open for over 2 years.

Any reason why it has not been reviewed, merged, etc. I'd be happy to help out with review, test, etc. if there is an indication where I can assist.

roeniss:
It's just my two cents passing by: There are a two styles to source zsh completion. I'm not sure which one is the most recommended way 🤔 

For example...

1. autoload & add to $ZSH_CACHE_DIR (some of this pattern use slightly different style)

https://github.com/ohmyzsh/ohmyzsh/blob/95579e3ca181036af31c0c9039fe3edc65c231b4/plugins/1password/1password.plugin.zsh#L1-L15

2. source and compdef (some of this pattern use slightly different style)

https://github.com/ohmyzsh/ohmyzsh/blob/72732a224e886933df6b64a49ec6f5e94c884612/plugins/istioctl/istioctl.plugin.zsh#L1-L4

jhomer-hscl:
Thanks @roeniss will look at sorting soemthing out.

jhomer-hscl:
Apologies for the delay. I've raised an issue to prompt a review of this PR. I'll keep a watch and look to get involved if there is anything I can help with.

roeniss (approved):

robbyrussell (changes requested):
Thanks for this, and I'm sorry it sat so long without a real review. Tekton is worth having a plugin for and people have asked for it (#13109), so I'd like to land it.

One change first. `source <(tkn completion zsh)` runs `tkn` on every shell start, synchronously. On a comparable Go CLI I measured 24-51 ms per shell for that, versus under 1 ms for reading a generated completion file. Every user who enables the plugin pays it in every terminal they open, so we've moved to caching the completion instead. 51 plugins use the newer pattern now.

Could you replace the body of `tkn.plugin.zsh` with this:

```zsh
# Autocompletion for tkn, the CLI for Tekton Cloud Native CI/CD (https://tekton.dev/)
if (( ! $+commands[tkn] )); then
  return
fi

# If the completion file doesn't exist yet, we need to autoload it and
# bind it to `tkn`. Otherwise, compinit will have already done that.
if [[ ! -f "$ZSH_CACHE_DIR/completions/_tkn" ]]; then
  typeset -g -A _comps
  autoload -Uz _tkn
  _comps[tkn]=_tkn
fi

zmodload -F zsh/files b:zf_mv
() {
  local TMPPREFIX="$ZSH_CACHE_DIR/completions/_tkn"
  zf_mv -f -- =( tkn completion zsh ) "$TMPPREFIX"
} &|
```

Three small things while you're in there: neither file ends with a newline, line 1 says "Natice" for "Native", and the README links to `github.com/jvarela01`, which 404s now.

If you'd rather not pick this back up after this long, say so and I'll apply the changes myself and keep your commit.

PR #11134, opened 2022-08-27
fancy-ctrl-z bindkey in vi mode

P4Cu:
## Standards checklist:

## Changes:

- just make the binding also for vi-mode

### DIFF
--- a/plugins/fancy-ctrl-z/fancy-ctrl-z.plugin.zsh
+++ b/plugins/fancy-ctrl-z/fancy-ctrl-z.plugin.zsh
@@ -9,4 +9,6 @@ fancy-ctrl-z () {
 }
 zle -N fancy-ctrl-z
 bindkey '^Z' fancy-ctrl-z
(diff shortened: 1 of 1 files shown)


P4Cu:
@mcornella Hey, I've noticed that this PR never got any attention. Maybe it could now ;) I've used this for a very long time and I've not noticed any problems.

mcornella (changes requested):
Hey! This is straightforward, apologies for the silence here. 

One small nitpick, to get a consistent style: could you also change the previous bindkey line to explicitly select the `emacs` keymap?

robbyrussell:
@P4Cu, are you still up for finishing this? It just needs Marc's suggestion from January: scope the existing line as `bindkey -M emacs '^Z' fancy-ctrl-z` so it matches the two you're adding.

The change itself looks right to me. Extra keymap entries are inert for people not using vi mode, so nothing changes for anyone else.

If you'd rather not, say so and I'll make that edit myself and keep your commit.

PR #11138, opened 2022-08-30
Better branch name parsing for Fossil plugin.

tangentsoft:
Using the automation-friendly "fossil branch current" feature added in
Fossil 2.7 instead of ad hoc parsing of human-readable "fossil branch"
output.  Not only does this fix a stray space in the Fossil trunk’s output, it's more
robust against changes in command output in general.

## Standards checklist:


tangentsoft:
I suck at Git (thus my contribution to the Fossil plugin) and so it is no surprise that I have screwed this PR up by making it against my fork's master, rather than a clean feature branch. This change is only the oldest commit, 7dfd60e320cc45d5da2d142bad7c72c0ec12b1a3. I hope you won't make me redo the PR; just cherry-pick that one commit, if you please.

mcornella:
Hi Warren, I'm about to merge this. If the rest of the changes are needed feel free to submit another PR and I will take a look. Thanks!

----

For future reference, fossil 2.7 or newer is in all the major distros except for Ubuntu 18.04 (see [pkgs.org](https://pkgs.org/search/?q=fossil)). As that is an LTS version from 4 years ago, it is expected that it won't be used by anybody using this plugin (i.e. the most common case will be for servers).

mcornella:
Note for the future: `fossil branch current` returns an error exit code which we can use to know whether we're in a fossil repo instead of `grep`-ing the error output.

PR #11139, opened 2022-08-30
Better placement of Fossil prompt info

tangentsoft:
Instead of blindly tacking `$fossil_prompt_info` onto the end of the
prompt string and making assumptions about whether `$PROMPT` or
`$RPROMPT` are in use here, assume the user is using a theme that puts
`$git_prompt_info` in there somewhere and inject our info immediately
afterward.  The theme creator decided that's a good place for Git prompt
info, so that's a good place for Fossil prompt info, too.

This also replaces some calls out to `grep` with internal Zsh string
manipulation for less overhead.

## Standards checklist:

### DIFF
--- a/plugins/fossil/fossil.plugin.zsh
+++ b/plugins/fossil/fossil.plugin.zsh
@@ -35,18 +35,23 @@ function fossil_prompt_info() {
 }
 
 function _fossil_prompt () {
-  local current=`echo $PROMPT $RPROMPT | grep fossil`
-
-  if [ "$_FOSSIL_PROMPT" = "" -o "$current" = "" ]; then
-    local _prompt=${PROMPT}
-    local _rprompt=${RPROMPT}
-
-    local is_prompt=`echo $PROMPT | grep git`
-
-    if [ "$is_prompt" = "" ]; then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for both of these, and sorry they sat so long.

I've read them together. They don't conflict: #11141 applies first and then #11139 goes on cleanly.

On #11139: replacing the grep on `$PROMPT` with substitutions is clearly better. The bit I want to check is the fallback. Today a theme with no git in `$PROMPT` gets fossil info appended to `$RPROMPT`, so it appears somewhere. After this, a theme with no literal `git_prompt_info` gets nothing until the user edits their theme by hand. Is that deliberate? I can see the argument for not guessing.

On #11141: moving the dirty marker inside the parens and dropping the padding matches what `lib/git.zsh` does, so no argument from me. The one thing I'd keep is the red branch name. 23 of our themes put red in the git prefix and fossil's matches today, so dropping it makes fossil the odd one out.

If you're up for it: #11141 with the red kept, then #11139 once the fallback question is settled.

tangentsoft:
At this late date, I'm not willing to quibble with _anything_. Get _something_ out there in shipping software, _then_ refine it.

PR #11141, opened 2022-08-30
Reduced unnecessary space in Fossil plugin prompt

tangentsoft:
The prior version was way too opinionated, adding space between each element of the prompt string, unlike common Git prompt strings.  If the user wants more space, they can override our public variables to suit.  The inverse argument doesn't hold: it's easy to add space to existing variables, but not as easy to remove it after the fact.

As part of this, changed the branch name from red to blue (matching the preceding part of the prompt) so the red "X" added to indicate a dirty branch doesn't get lost in the branch name.

## Standards checklist:

## Other comments:

I realize this is a purely cosmetic change based on personal taste, but as I explain in the commit comment, it’s easier to add space from too-little than to remove it from too-much.  I believe that makes this a better way to ship the plugin: end users who don’t like the change can add whatever space they like back in by overriding our public variables.

### DIFF
--- a/plugins/fossil/fossil.plugin.zsh
+++ b/plugins/fossil/fossil.plugin.zsh
@@ -1,16 +1,16 @@
 _FOSSIL_PROMPT=""
 
 # Prefix at the very beginning of the prompt, before the branch name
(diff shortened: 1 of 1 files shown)


robbyrussell:
See my note on #11139: no argument from me on the spacing, but I'd keep the red branch name. This one can land on its own, ahead of #11139.

PR #11150, opened 2022-09-05
feat(lib): support for tmux pane title

latipun7:
## Standards checklist:

## Changes:

- Add comment for better understanding
- Add escape sequence for setting tmux pane title

## Other comments:

As described in the [tmux man page](http://man.openbsd.org/OpenBSD-current/man1/tmux.1#NAMES_AND_TITLES), tmux has pane title. Currently, that pane title is only showing hostname for zsh, because zsh not editing the title with the current termsupport.

This PR adding escape sequence as described in the man page to set pane title.
This PR also add new comment for the current escape sequence for editing window name, that escape sequence would take effect/renaming tmux window name only if tmux configured with `set -g allow-rename on`.

### Before:
image
_all the tmux pane title is set to the default, which is only hostname, zsh not set the tmux pane title_

### After:
image
_all the tmux pane title now set by zsh_

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -22,7 +22,8 @@ function title {
       print -Pn "\e]1;${1:q}\a" # set tab name
       ;;
     screen*|tmux*)
-      print -Pn "\ek${1:q}\e\\" # set screen hardstatus
+      print -Pn "\e]2;${2:q}\e\\" # set tmux pane title
+      print -Pn "\ek${1:q}\e\\" # set screen hardstatus/tmux window name if `allow-rename` is on
       ;;
     *)
       if [[ "$TERM_PROGRAM" == "iTerm.app" ]]; then


latipun7:
I'm sorry to necrobump this PR. What is the status of this PR, @mcornella ? Thanks

robbyrussell (approved):
Thanks for this, and sorry it sat so long with no answer. This looks right to me.

I tested it before saying that: on master a tmux pane title stays at the hostname forever, and with this it tracks the theme value on every prompt. `allow-rename` stays off so window names don't change, and `DISABLE_AUTO_TITLE=true` still turns the whole thing off.

I did check whether screen would suffer, since it shares that case arm and has no panes. It ignores the sequence entirely, with either terminator, so screen behavior is unchanged.

The comment fix on the existing line is a good catch. That `allow-rename` behavior trips people up.

mcornella (approved):

robbyrussell:
@latipun7  -- we really appreciate the contribution!

PR #11152, opened 2022-09-05
feat(brew): autoconfigure brew zsh completions path

voronkovich:
## Standards checklist:

## Changes:

- Automatically configure completions path: https://docs.brew.sh/Shell-Completion#configuring-completions-in-zsh


robbyrussell:
Thanks for the contribution. We should get some folks to help test this one out for us

ogtega:
I was frustrated by the lack of auto completions too and came upon this solution which updates both the _comps and fpath variable.

```bash
BREW_COMPLETIONS="$(brew --prefix)/share/zsh/site-functions"
  if [[ -d $BREW_COMPLETIONS ]]; then
    fpath+=$BREW_COMPLETIONS

    typeset -g -A _comps

    for file in $BREW_COMPLETIONS/*(N)
    do
      comp=$file:t
      autoload -Uz $comp
      _comps[${comp:1}]=$comp
      unset comp
    done
  fi
```

HuaDeity:
Sorry, but in my case, there is no effect. 
I searched and think the problem may be because of #https://github.com/zsh-users/zsh-completions/issues/603.

gabriel-lima96:
> Thanks for the contribution. We should get some folks to help test this one out for us

Not working. I've added a few echo in my local installation and the change to the zpath is being done after the compinit is run. The echos:

brew.plugin.zsh:
image

oh-my-zsh.sh:
image

output:
image

Net-burst:
Can confirm the same that @gabriel-lima96 observed. I need to manually add brew completions to the FPATH so that they are referenced at the time of the compinit run or regenerate the completions file after OMZ initializes.

PR #11161, opened 2022-09-11
chore(lib): update deprecated grep aliases

### LINKED ISSUE #11160: egrep is obsolescent, aliases should be updated
### Discussed in https://github.com/ohmyzsh/ohmyzsh/discussions/11159

<div type='discussions-op-text'>

<sup>Originally posted by **Nadhum** September 10, 2022</sup>
I started getting warnings recently about these 
`egrep: warning: egrep is obsolescent; using grep -E`

https://github.com/ohmyzsh/ohmyzsh/blob/12669f29f0843b8b980dd137f150a74511f88842/lib/grep.zsh#L27-L28

Should these aliases be removed by default?</div>

egrep and fgrep issue a warning on shell startup. The suggestion is to keep the aliases but use grep -E instead of egrep and grep -F instead of fgrep.

Nadhum:
## Standards checklist:

## Changes:

Updating grep aliases:
- egrep to use grep -E instead of egrep
- fgrep to use grep -F instead of fgrep

## Other comments:

Fixes #11160

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -24,8 +24,8 @@ else
     if [[ -n "$GREP_OPTIONS" ]]; then
         # export grep, egrep and fgrep settings
         alias grep="grep $GREP_OPTIONS"
-        alias egrep="egrep $GREP_OPTIONS"
-        alias fgrep="fgrep $GREP_OPTIONS"
+        alias egrep="grep -E $GREP_OPTIONS"
+        alias fgrep="grep -F $GREP_OPTIONS"
 
         # write to cache file if cache directory is writable
         if [[ -w "$ZSH_CACHE_DIR" ]]; then


Nadhum:
Still getting the warnings on shell startup `egrep: warning: egrep is obsolescent; using grep -E` anywhere else this needs to be updated?

ouuan:
> Still getting the warnings on shell startup `egrep: warning: egrep is obsolescent; using grep -E` anywhere else this needs to be updated?

It seems that the alias itself doesn't cause the warning. The warning only appears when `egrep` is actually executed. In my case it's the `alias-tips` plugin that is using `egrep`, and the warning was appearing on every command. Maybe you can check your plugins.

Nadhum:
> It seems that the alias itself doesn't cause the warning. The warning only appears when `egrep` is actually executed. In my case it was the `alias-tips` plugin that was using `egrep`, and the warning was appearing on every command. Maybe you can check your plugins.

I only have the git plugin enabled, and it doesn't call egrep. Wondering where my warning is coming form? :thinking:

carlosala:
That's because it caches the aliases [here](https://github.com/ohmyzsh/ohmyzsh/blob/7dcabbe6826073ef6069c8a4b6f9a943f00d2df0/lib/grep.zsh#L32) and [here](https://github.com/ohmyzsh/ohmyzsh/blob/7dcabbe6826073ef6069c8a4b6f9a943f00d2df0/lib/grep.zsh#L5) tries to load them.
I would directly remove that caching of aliases, I think it's totally useless, do you @mcornella know why are them there?

ouuan:
> I would directly remove that caching of aliases, I think it's totally useless, do you @mcornella know why are them there?

#8444

carlosala:
Then looks like might be interesting to keep it 👌🏻

ouuan:
I don't think the warning @Nadhum is receiving is caused by the cache, nor the aliases modified in this PR, because I got rid of the warnings by upgrading the `alias-tips` plugin where the use of `egrep` is removed in recent commits. Simply defining the aliases without calling them doesn't cause the warnings, at least in my case.

However, this PR is still needed because it fixes the warning when a user uses `egrep` in the command line. The reason for @Nadhum's warning is another issue.

meliache:
I had the same warning since recently, grepepd my dotfiles  and `.oh-my-zsh` for `egrep` and also thought at first that my issue are the omz aliases, but turned out that for me it was the line
``` bash
source virtualenvwrapper.sh
```
which caused this error.

Nadhum:
> I had the same warning since recently, grepped my dotfiles and `.oh-my-zsh` for `egrep` and also thought at first that my issue are the omz aliases, but turned out that for me it was the line
> 
> ```shell
> source virtualenvwrapper.sh
> ```
> 
> in my `.zshenv` which caused this error.

Good catch, mine as well coming from virtualenvwrapper, there is an issue already opened there.

Nadhum:
@carlosala feel free to approve and merge the changes if they look good!

carlosala (approved):

mcornella:
The alias cache only lasts for a day, so it should be regenerated after that. I have searched the codebase and removed one use of `egrep` in the gradle plugin: b93b67b84490d45eb2544b1741e64dc16103207d.

In any case, thank you for the fixes @Nadhum. LGTM

PR #8204, opened 2019-10-03
Add Terragrunt Alias

Tensho:
https://github.com/gruntwork-io/terragrunt

PR #11164, opened 2022-09-13
feat(terragrunt): add tg alias & autocompletion

Tensho:
#8204


lancedikson (approved):
would be really nice to have this merged! 🙏

lancedikson:
@carlosala would you have time to look at it please? 🙏 I'm really keen on getting those tg-aliases support.

robbyrussell:
Thanks for this, and I'm sorry it waited so long for an answer.

I want a terragrunt plugin, and the aliases here are good. What I don't want to take on is the bundled `_terragrunt` completion. It's a copy marked current as of v0.19.16, and terragrunt is on 1.x now with a mostly different command set, so it would be wrong the day it merged and nobody would be watching it.

There's a newer take in #13261 built on our opentofu plugin, with no vendored completion. I'd rather land that shape and credit you for getting here first. Closing this one out.

PR #11167, opened 2022-09-15
fix(brew): do not source `brew shellenv` if already on path

### LINKED ISSUE #11166: brew plugin should not call brew shellenv if brew is already on the PATH.
### Describe the bug

This PR change https://github.com/ohmyzsh/ohmyzsh/pull/10874 has caused the `brew` plugin to always call `eval $(brew shellenv)` regardless if the environment is already configured correctly. This has unintended consequences, for instance the brew bin directory (`/usr/local/bin` on MacOS Intel) is exported in `PATH` twice.

I would suggest this plugin mimic the behaviour of the brew [installer.sh](https://github.com/Homebrew/install/blob/eb943a56afc66244ff216a511058a89570440610/install.sh#L1017) to only require `eval $(brew shellenv)` if `brew` is not found in the path.

One other thing to note is the brew installer post-install steps says to modify `~/.zprofile` and not `~/.zshrc`. Perhaps this plugin should do what `pyenv` plugin does and [warn if the user did not run the post-install steps](https://github.com/ohmyzsh/ohmyzsh/blob/7dcabbe6826073ef6069c8a4b6f9a943f00d2df0/plugins/pyenv/pyenv.plugin.zsh#L1-L20) so that it may be called from non interactive scripts.

Other notes regarding the other environment variables exported by `shellenv`, specifically for MacOS Intel:

- `MANPATH` and `INFOPATH` are not required to add `/usr/local/share/man` and `/usr/local/share/info` - they are available in the search path by default.

- Neither is `HOMEBREW_PREFIX`, `HOMEBREW_CELLAR` and `HOMEBREW_REPOSITORY` required by brew to be exported. The `brew` command will ignore these environment variables and recalculate their values. Users should generally prefer `$(brew --prefix)` and not rely on or require `HOMEBREW_PREFIX` to be exported.

### Steps to reproduce

1. Enable plugin `brew`

2. `echo $PATH | tr ':' '\n'`

3. See `/usr/local/bin` listed twice

### Expected behavior

`PATH` should have `/usr/local/bin` listed once

_No response_

### OS / Linux distribution

MacOS Intel 12.6 Monterey

### Zsh version

5.8.1

### Oh My Zsh version

7dcabbe6

### Terminal emulator

iTerm2

_No response_

### Additional context

_No response_

tmoschou:
## Standards checklist:

## Changes:

This changes the brew plugin to only source `brew shellenv` if brew cannot be found in `PATH`. This aligns with the brew installer.sh post install steps, and will prevent duplicate entries for `/usr/local/bin` in PATH on MacOS Intel.

The plugin will continue to export `HOMEBREW_PREFIX` for custom user config which previously relied on the behavior.

Fixes #11166

### DIFF
--- a/plugins/brew/brew.plugin.zsh
+++ b/plugins/brew/brew.plugin.zsh
@@ -10,17 +10,20 @@ if (( ! $+commands[brew] )); then
   else
     return
   fi
+
+  # Only add Homebrew installation to PATH, MANPATH, and INFOPATH if brew is
+  # not already on the path, to prevent duplicate entries. This aligns with
+  # the behavior of the brew installer.sh post-install steps.
+  eval "$("$BREW_LOCATION" shellenv)"
+  unset BREW_LOCATION
 fi
 
 if [[ -z "$HOMEBREW_PREFIX" ]]; then
-  if [[ -z $BREW_LOCATION ]]; then
-    eval "$(brew shellenv)"
-  else
-    eval "$("$BREW_LOCATION" shellenv)"
-  fi
+  # Maintain compatability with potential custom user profiles, where we had
+  # previously relied on always sourcing shellenv. OMZ plugins should not rely
+  # on this to be defined due to out of order processing.
+  export HOMEBREW_PREFIX="$(brew --prefix)"
 fi
-
-unset BREW_LOCATION
 
 alias bcubc='brew upgrade --cask && brew cleanup'
 alias bcubo='brew update && brew outdated --cask'
--- a/plugins/brew/README.md
+++ b/plugins/brew/README.md
@@ -10,7 +10,10 @@ plugins=(... brew)
 
 ## Shellenv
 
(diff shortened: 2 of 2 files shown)


plugins/brew/brew.plugin.zsh
```diff
@@ -1,4 +1,5 @@
 if (( ! $+commands[brew] )); then
+  local BREW_LOCATION
```
carlosala, line 2: As we are going to unset the variable, let's keep that not-local!

carlosala (approved):
I really like the approach, thanks for your contribution!

plugins/brew/brew.plugin.zsh
carlosala: There was something about old versions of zsh and local variables, now I don't exactly remember what was happening.

mcornella (approved):

PR #11176, opened 2022-09-21
fix(git): remove redundant escapes from WIP commit message

allanlewis:
## Standards checklist:

## Changes:

- Remove redundant escaping of WIP commit message for `gwip` and `gunwip`.

## Other comments:

When running `gunwip` I get the following error, three times:

    grep: warning: stray \ before -

This appears to be due to the second, third and fourth hyphens in `--wip--` being escaped when only the first needs to be. This PR addresses these warnings by removing the need to escape the hyphens entirely.


mcornella:
The README should keep the escapes, otherwise it's rendered with a strikethrough. The rest LGTM 👌🏻

carlosala:
Thanks for your contribution @allanlewis!

PR #11180, opened 2022-09-24
added zsh-toolbox

paxcoder:
This code is mine, I made zsh-toolbox, plugin which changes ls to show hidden files by default, and aliases mz to omz(for convenience.
You can add new stuff(too)
But it is good to ask me first.


carlosala:
HI! I would this plugin is quite personal, the aliases and the echoes are not really for everyone and I'm not sure if we should ship this with `omz`, and you could still use it in your `custom` plugins directory (or even create a repo for it so interested people could use it!).
What do you think about?

PR #11181, opened 2022-09-24
feat(bitwarden): completions and bw_unlock bw_login functions

casaper:
Plugin adding [Bitwarden CLI](https://bitwarden.com/help/cli/) completion 

`bw_unlock` helper function that unlocks and updates the BW_SESSION env export in zsh files, so other new shell can inherit the unlocked Bitwarden session automatically.
## Standards checklist:

## Changes:

- add slightly extended version of [Bitwarden CLI completion output](https://bitwarden.com/help/cli/#zsh-shell-completion)
- `bw_unlock` helper function that unlocks and updates the BW_SESSION env export in zsh files, so other new shell can inherit the unlocked Bitwarden session automatically.

## ToDo

- [x] implement completions as `deno` or `gh` plugins did

### DIFF
--- a/plugins/bitwarden/bitwarden.plugin.zsh
+++ b/plugins/bitwarden/bitwarden.plugin.zsh
@@ -1,0 +1,107 @@
+if ! (( $+commands[bw] )); then
+  return
+fi
+
+# bw_login - bitwaredn-cli authentication token helper
+#
+# - wraps the execution of `bw login`
+# - replaces the resulting token to zsh files
+#   - in `~/.zshrc` or `$ZSH_CUSTOM/**/*.zsh`
+#   - where a `export BW_SESSION=.*$` is found
+# - if none is found, it creates a new export BW_SESSION in `~/.zshrc`
+# - The arguments `bw login` takes are passed on
+# - the function offers the same completions that `bw login` has
+#
+# This way the BW_SESSION token is available for all new shells
+# as well, not just the current.
+function bw_login() {
+  _bw_session_env_update "$(bw --raw login $@)" "$ZSH_CUSTOM"
+}
+compdef _bw_login_completions bw_login
+
+
+# bw_unlock - bitwaredn-cli authentication token helper
+#
+# - wraps the execution of `bw unlock`
+# - replaces the resulting token to zsh files
+#   - in `~/.zshrc` or `$ZSH_CUSTOM/**/*.zsh`
+#   - where a `export BW_SESSION=.*$` is found
+# - if none is found, it appends a new `export BW_SESSION` in `~/.zshrc`
+# - The arguments `bw unlock` takes are passed on
+# - the function offers the same completions that `bw unlock` has
+#
+# This way the BW_SESSION token is available for all new shells
+# as well, not just the current.
+function bw_unlock() {
+  _bw_session_env_update "$(bw --raw unlock $@)" "$ZSH_CUSTOM"
+}
+compdef _bw_unlock_completions bw_unlock
+
+# Refresh all BW_SESSION token exports
+#
+# Find and replace all exports of BW_SESSION found in `~/.zshrc` or `$ZSH_CUSTOM/**/*.zsh`.
+# If none found, it appends it `~/.zshrc`
+function _bw_session_env_update() {
+  local session_unlock_token="$1"
+  local zsh_custom_dir="$2"
+  if [[ -z "${session_unlock_token}" ]]; then
+    return 1
+  fi
+  export BW_SESSION="${session_unlock_token}"
+  echo "export BW_SESSION: $BW_SESSION"
+  local files_with_bw_session="$(grep --files-with-matches 'export BW_SESSION' "${zsh_custom_dir}"/**/*.zsh ${HOME}/.zshrc)"
+  local files_count=$(wc -l <<<"$files_with_bw_session")
+  if (( files_count > 0 )); then
+    for env_export_file in ${files_with_bw_session}; do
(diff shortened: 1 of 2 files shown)


carlosala:
I see that bitwarden CLI allows you to get zsh completion function. We should do something as `deno` or `gh` plugin does to get that completion, could you check it?
Thanks!

casaper:
@carlosala 

> I see that bitwarden CLI allows you to get zsh completion function. We should do something as `deno` or `gh` plugin does to get that completion, could you check it? Thanks!

I wasn't aware about that. 
Will do so and then requeue this PR again when I've done that.

Sorry about that.

casaper:
@carlosala

Done that now. Thanks allot for pointing this out!

casaper:
> Could you maintain the same structure that `deno` or `gh` plugin have? Thanks!

As mentioned in https://github.com/ohmyzsh/ohmyzsh/pull/11229#issuecomment-1285461722, I will do that here, and keep this caching for myself (privately), because I am sure it helps, especially with bitwarden.

But no worries. I'll do as you request.

carlosala (approved):
LGETM! Let's ping @mcornella for a review too, as it's a new plugin adding a bunch of features

bt0r:
Hi @casaper !
I was working on a bitwarden plugin too and just see your PR today, main things of mine is an unlock and list helper. 
I added a `bitwarden list` command which provide an easier way to find credential by using `fzf` and by copying password directly on the clipboard.
What do you think about integrating this feature to your PR ?

You can check what i start to do here https://github.com/bt0r/ohmyzsh/tree/plugin/bitwarden/plugins/bitwarden

casaper:
Hi @bt0r 

> I added a bitwarden list command which provide an easier way to find credential by using fzf and by copying password directly on the clipboard.

Sounds interesting. Goes allot further than what I intended back then. My intentions where to store config files with secrets in them (e.g. the one for rclone) in bitwarden, and extract them from bitwarden when using rclone and remove these afterwards.
I kind of lost the Idea and did not go any further with it

> What do you think about integrating this feature to your PR ?

Totally fine for me. I have lost track of this PR completely, and I did not check if what I did back then still works etc. I guess it is stale, since no one reviewed it.

Take whatever you want/need from here for your plugin. I don't mind if you'd integrate any of it in your PR. No claims or anything from my side.

bt0r:
@casaper I'll test your PR and take a look at next steps 

Thanks for your kindness :)

carlosala:
if we have some testers approving the PR, it's ok for me to merge! I don't personally use `bw`, that's why I can't test myself. @bt0r could you test the PR? Thanks!

casaper:
> if we have some testers approving the PR, it's ok for me to merge! I don't personally use `bw`, that's why I can't test myself. @bt0r could you test the PR? Thanks!

@carlosala 
I lost track of this PR. But if you'd agree to it when someone with bw tests it is also fine by me.

Regarding the other stuff @bt0r did:
In case this one really gets merged, I think it would make most sense if you then extended this plugin with your much larger functionality.
I'd say it would not make sense to have more than one plugin for bw for oh-my-zsh.

As I said, if it is up to me, you'll have total freedom to adapt what is here to your needs.

carlosala:
yes, it makes more sense to extend than to create a new one.

casaper:
@carlosala I rebased this PR's branch with the target branch again now, just do make sure that this would not be in the way.

So you could agree with this one if @bt0r tested it and concluded with an approval?

carlosala:
yes, perfect! I'll do a detailed review this afternoon too👌🏻

plugins/bitwarden/bitwarden.plugin.zsh
```diff
@@ -0,0 +1,90 @@
+if ! (( $+commands[bw] )); then
+  return
+fi
+
+# bw_login - bitwaredn-cli authentication token helper
```
carlosala, line 5: there's a small typo here, but meh, not an issue actually

carlosala (commented):
The only thing that bothers me is to save the token (which I suppose give free access to all the vault) directly in plain text in a file, which can be read by whichever process running in your user.
It shouldn't be better to maybe use some keyring or something?
For bitwarden, I use [`rbw`](https://github.com/doy/rbw) which uses an agent (such as ssh-agent) to manage that.
What do you think about?

casaper:
> The only thing that bothers me is to save the token (which I suppose give free access to all the vault) directly in plain text in a file, which can be read by whichever process running in your user.
> It shouldn't be better to maybe use some keyring or something?
> For bitwarden, I use [`rbw`](https://github.com/doy/rbw) which uses an agent (such as ssh-agent) to manage that.
> What do you think about?
> If the recommended workflow is to save the token in your zshrc, why should we export also in ZSH_CUSTOM?
> Thanks!

Yeah, you are totally right.

The token expires, so that is why I thought for myself it would be tolerable.

But for a completely unwitting oh my zsh user it totally is not.

But in that case this PR is almost completely obsolete.
The only thing left would be the completions.
But if it is only that, @bt0r  can just copy & paste it to his PR.

carlosala:
thanks for your contribution! It was a really great PR☺️

xjunior:
What happened to this? Was it merged? I don't see it in main

casaper:
> What happened to this? Was it merged? I don't see it in main

@xjunior 

I closed it in favor of an older pr back then.  
It also had some notable flaws.

recently I creted a bunch of utility functions again, for personal use.

They are not suitable for OhMyZsh contributions, mainly because they make use of jq and ripgrep and more.

PR #11196, opened 2022-09-26
feat(aws): load profiles from credentials file

### LINKED ISSUE #8472: AWS plugin doesn't read profiles from credentials file
> The discussion above is incorrect because profiles are valid both defined in `~/.aws/config` and `~/.aws/credentials` files. In the credentials file they do not have a `[profile ]` prefix. The word "profile" is something AWS has been trying to get away from as it was too restrictive. Profile just means a set of credentials now and AWS CLI and SDK honour it. All the saml and mfa login tools don't bother maintaining a config file but just credentials file. The "improvement" that was done to check if profiles exist definitely broke the tool for many people. I had to reinstall machineand was surprised that it doesn't work any more for me. I guess I'm back to simple env variable setting...

_Originally posted by @StoyanIvanovI in https://github.com/ohmyzsh/ohmyzsh/issues/7890#issuecomment-559553829_

noamcohen97:
## Standards checklist:

## Changes:

- Use `aws configure list-profiles` to list the available profiles instead of manually parsing the config files.

## Other comments:

Using `aws configure list-profiles` makes sure we parse both config and credentials files.
Doing so will also make the function not prone to future changes of file format or creating / removing configurations files by AWS CLI.

Fixes #8472

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -146,7 +146,9 @@ function aws_change_access_key() {
 }
 
 function aws_profiles() {
-  aws --no-cli-pager configure list-profiles
+  aws --no-cli-pager configure list-profiles 2> /dev/null && return
+  [[ -r "${AWS_CONFIG_FILE:-$HOME/.aws/config}" ]] || return 1
+  grep --color=never -Eo '\[.*\]' "${AWS_CONFIG_FILE:-$HOME/.aws/config}" | sed -E 's/^[[:space:]]*\[(profile)?[[:space:]]*([^[:space:]]+)\][[:space:]]*$/\2/g'
 }
 
 function _aws_profiles() {
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -146,7 +146,7 @@ function aws_change_access_key() {
 }
 
 function aws_profiles() {
-  AWS_PAGER="" aws configure list-profiles
+  aws --no-cli-pager configure list-profiles
 }
 
 function _aws_profiles() {
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -146,8 +146,7 @@ function aws_change_access_key() {
 }
 
 function aws_profiles() {
-  [[ -r "${AWS_CONFIG_FILE:-$HOME/.aws/config}" ]] || return 1
-  grep --color=never -Eo '\[.*\]' "${AWS_CONFIG_FILE:-$HOME/.aws/config}" | sed -E 's/^[[:space:]]*\[(profile)?[[:space:]]*([^[:space:]]+)\][[:space:]]*$/\2/g'
+  AWS_PAGER="" aws configure list-profiles
 }
 
 function _aws_profiles() {


noamcohen97:
@maksyms ?

noamcohen97:
@maksyms ?

maksyms (commented):
I take it it is introducing a breaking change and from this point on, this plugin won't work with `aws-cli` earlier than v2, correct? If so, can we check the version first please?

Also, instead of setting `AWS_PAGER`, it is better to use `--no-cli-pager` command line.

noamcohen97:
> check the version first please?

@maksyms @mcornella 

What do you think the behaviour should be like in versions prior to v2? should we fall-back to the old way of extracting named profiles? or should we just fail with an error?

maksyms:
> What do you think the behaviour should be like in versions prior to v2? should we fall-back to the old way of extracting named profiles? or should we just fail with an error?

I personally would prefer for the plugin to continue to work - hence, fall-back to the old way of extracting named profiles.

But I'm no longer the codeowner here, so it'll be down to @mcornella .

carlosala:
@maksyms your opinion still counts a lot actually! Let's do what you say! @noamcohen97

noamcohen97:
> Let's do what you say!

@carlosala Sounds great!
WDYT?

elisiariocouto:
Hey guys, sorry to unearth this, but my `asp` autocomplete is considerably slow with this feature now (it takes about 1 sec) and I managed to pinpoint this MR. If I comment the `aws --no-cli-pager configure list-profiles 2> /dev/null && return` line, the autocomplete is almost instant. I think the Python bootstrap is what makes it slow. Should we add a variable to somehow disable this? Anyone with the same issue?

paulolieuthier:
I had the same experience as @elisiariocouto. Besides, in my understanding, this patch is completely unnecessary with regard to the original issue. The regex used to extract the profiles names had already been fixed to make the `profile` prefix optional.

noamcohen97:
> this patch is completely unnecessary

AWS CLI profiles can be defined in `AWS_SHARED_CREDENTIALS_FILE` (`~/.aws/credentials`) as well.
The previous code did not show profiles defined in this file.

Using AWS logic to process the profiles makes the code more forward-compatible to any change AWS will ever do to profiles.

With that said - I understand that Python bootstrap might be slow in some cases, so I suggest we'll change the old regex-based profile extraction to also look at the other file, and allow opting-out the new way (or maybe remove it completely).

Any thoughts?

paulolieuthier:
I agree with you that using AWS' own tooling would be ideal and future-proof. But speed-wise, when having multiple profiles, I think it defeats the purpose of using the plugin.

I believe the regex-based solution is good enough and should be improved to support the configuration environment variable. From what I read in the documentation, the file to be read is not the credentials file, but the config file. And so `AWS_CONFIG_FILE` should be used instead of `AWS_SHARED_CREDENTIALS_FILE`. Is there a scenario in which the config file is not available?

Another thing that would make the command faster is to get the profile region from the config file as well, instead of running `aws configure get region`.

noamcohen97:
> From what I read in the documentation, the file to be read is not the credentials file, but the config file

Both files are used when searching for a profile:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials

PR #11205, opened 2022-10-01
fix(nvm): source nvm script only when used

### LINKED ISSUE #11042: nvm plugin awfully slow

nvm plugin

Loading nvm plugin is awfully slow.  I set NVM_LAZY=1 but no difference.

My set up OSX Monteray and nvm 0.39.1 installed via brew 

solution for me was to disable it and go with a custom script found on:

https://www.reddit.com/r/node/comments/4tg5jg/lazy_load_nvm_for_faster_shell_start/d5ib9fs/?utm_term=28631259784&context=3&utm_medium=comment_embed&utm_source=embed&utm_name=null

```
declare -a NODE_GLOBALS=(`find ~/.nvm/versions/node -maxdepth 3 -type l -wholename '*/bin/*' | xargs -n1 basename | sort | uniq`)

NODE_GLOBALS+=("node")
NODE_GLOBALS+=("nvm")

load_nvm () {
    export NVM_DIR=~/.nvm
    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
}

for cmd in "${NODE_GLOBALS[@]}"; do
    eval "${cmd}(){ unset -f ${NODE_GLOBALS}; load_nvm; ${cmd} \$@ }"
done
```

maybe you could add that snippet behind an other variable that loads it and discards the rest

_No response_

### Additional context

_No response_

### Related Issues

_No response_

carlosala:
## Standards checklist:

## Changes:

Source `nvm` in an async way. This `nvm.sh` takes 50ms even if using `NVM_LAZY` so let's load it only when needs to be used the first time

closes #11042

### DIFF
--- a/plugins/nvm/README.md
+++ b/plugins/nvm/README.md
@@ -22,9 +22,9 @@ These settings should go in your zshrc file, before Oh My Zsh is sourced:
   nvm has been installed, regardless of chip architecture, use `NVM_HOMEBREW=$(brew --prefix nvm)`.
 
 - **`NVM_LAZY`**: if you want the plugin to defer the load of nvm to speed-up the start of your zsh session,
-  set `NVM_LAZY` to `1`. This will use the `--no-use` parameter when loading nvm, and will create a function
-  for `node`, `npm`, `yarn`, and the command(s) specified by `NVM_LAZY_CMD`, so when you call either of them,
-  nvm will load with `nvm use default`.
+  set `NVM_LAZY` to `1`. This will source nvm script only when using it, and will create a function for `node`,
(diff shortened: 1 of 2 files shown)


carlosala:
@mcornella let's also review this, what do you think about?

mcornella:
This won't work for globals, for which the snippet in the linked issue should be used.

However that version won't remove the lazy load bootstrapping functions once nvm has been loaded once.

To do that, all the functions should be removed much like the version here.

However this is a nice improvement and I think the other one can be added in the future if there is want for it.

LGTM

PR #11214, opened 2022-10-02
Add k3s auto-completion plugin

sgaist:
## Standards checklist:

## Changes:

- Add simple k3s auto completion plugin
- No alias added

## Other comments:

It complements #11207 to complete the Rancher toolbox.


mcornella:
Hello Samuel, it doesn't seem like there is much traction in getting this plugin in the official distribution of OMZ, and it requires a single line of code to add this to the `.zshrc` file.

Thank you for the contribution though!

PR #11218, opened 2022-10-03
fix(cli): change unrecognized `\s` in BSD awk (#11146)

### LINKED ISSUE #11146: `omz plugin disable` does not work
### Describe the bug

Command `omz plugin disable $plugin` does not work.
It just does not delete plugin from list plugins from `.zshrc` and from `$plugins` and plugins is continuing work.

### Steps to reproduce

For example:
1. `omz plugin enable github`
2. `echo $plugins`
github
3. `omz plugin disable github`
4. `echo $plugins`
github

### Expected behavior

For example:
1. `omz plugin enable github`
2. `echo $plugins`
github
3. `omz plugin disable github`
4. `echo $plugins`
empty line

изображение

### OS / Linux distribution

macOS 12.5.1 21G83 arm64

### Zsh version

5.8.1

### Oh My Zsh version

master (6d48309)

### Terminal emulator

macOS Treminal

_No response_

### Additional context

_No response_

mcornella:
## Standards checklist:

## Changes:

- Change `\s` with `[ \t]` in awk script for CLI zshrc change operations.

## Other comments:

In BSD awk, \s is not a valid sequence interchangeable with "space or tab characters"
as it is in GNU awk. This fix uses [ \t] instead, which is all the possibilities that
we need to contemplate when reading the .zshrc file.

Fixes #11146

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -241,21 +241,21 @@ function _omz::plugin::disable {
 
   # Remove plugins substitution awk script
   local awk_subst_plugins="\
-  gsub(/\s+(${(j:|:)dis_plugins})/, \"\") # with spaces before
-  gsub(/(${(j:|:)dis_plugins})\s+/, \"\") # with spaces after
+  gsub(/[ \t]+(${(j:|:)dis_plugins})/, \"\") # with spaces before
+  gsub(/(${(j:|:)dis_plugins})[ \t]+/, \"\") # with spaces after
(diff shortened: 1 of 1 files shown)


LunarWatcher (approved):
I assume this makes perfect sense as is, but if there are any other valid spaces that may occur in the wild (particularly thinking about the evil nbsp) _and_ that are valid shell (nbsp generally throws compiler/interpreter errors in other languages), it's probably worth adding even if it seems implausible to encounter them.

Obscure scripts are obscure, so if it's valid shell, there are probably scripts that use it :')

PR #11222, opened 2022-10-04
fix(npm): ignore `rm` alias

ZhongRuoyu:
## Standards checklist:

## Changes:

- npm plugin: disregard `rm` alias if set.

## Other comments:

See also: #11156.

### DIFF
--- a/plugins/npm/npm.plugin.zsh
+++ b/plugins/npm/npm.plugin.zsh
@@ -1,5 +1,5 @@
 (( $+commands[npm] )) && {
-  rm -f "${ZSH_CACHE_DIR:-$ZSH/cache}/npm_completion"
+  command rm -f "${ZSH_CACHE_DIR:-$ZSH/cache}/npm_completion"
 
   _npm_completion() {
     local si=$IFS


LunarWatcher (approved):
Looks reasonable to me (even though I don't understand why anyone would make an alias that shadows an existing base command; this has to cause problems beyond OMZ if it's aliased to the right thing:tm: -- aside the point, however)

rwmitchell:
People used to alias rm to rm -I
as a safety net, which is great up until the alias isn’t there.

ZhongRuoyu:
Thanks for reviewing!

Personally I've been aliasing `rm` to something like...

```zsh
alias rm="echo 'Think again and type: command rm'"
```

... ever since I had an unpleasant experience of executing `rm -rf $SOMETHING/` while forgetting to set variable `$SOMETHING`, and ended up executing `rm -rf /`. With the aliased `rm`, the above would instead print this out, with variables and wildcards properly evaluated:

```plaintext
Think again and type: command rm -rf /
```

So I can see clearly what's going on.

While this reason may not be justifiable enough, it has not brought me much trouble thus far, because aliases are not evaluated by shell scripts by default (but not if they are `source`d, like the npm plugin). There are other `rm` invocations in my OMZ, but they are unshadowed with `command rm`. So I believe it's justifiable to bring this to the npm plugin as well.

ZhongRuoyu:
Sorry for pinging, but I wonder if it would be possible to land this? @mcornella

mcornella (approved):

ZhongRuoyu:
Thanks!

mcornella:
Thank you! 😊

PR #11224, opened 2022-10-04
fix(git)!: Standardize `git pull --reabse` aliases

### LINKED ISSUE #11104: Duplicate Alias Command for Git
### Describe the bug

I wanted to know what the shorthand command for `git pull --rebase origin master` and typed `alias | grep 'git pull'` and noticed a duplicate command.

```
$ alias | grep 'git pull'
ggpull='git pull origin "$(git_current_branch)"'
gl='git pull'
glum='git pull upstream $(git_main_branch)'
gpr='git pull --rebase'
gup='git pull --rebase'
gupa='git pull --rebase --autostash'
gupav='git pull --rebase --autostash -v'
gupom='git pull --rebase origin $(git_main_branch)'
gupomi='git pull --rebase=interactive origin $(git_main_branch)'
gupv='git pull --rebase -v'
```
The command 'git pull --relbase' is duplicated.

### Steps to reproduce

1. Open terminal
2. type: alias | grep 'git pull --rebase'

### Expected behavior

Only one config for 'git pull --rebase'

_No response_

### OS / Linux distribution

macOS 12.5

### Zsh version

5.8.1

### Oh My Zsh version

master (d41ca84)

### Terminal emulator

iTerm2, Terminal.app, VSCode integrated terminal

_No response_

### Additional context

_No response_

zee-bit:
## Standards checklist:

## Changes:

- Revised aliases for `git pull --rebase` and it's derivates to the more standard form of `gpr*` from `gup*`.
- BREAKING CHANGE: The alias `gup` for `git pull --rebase` and its derivatives are replaced by `gpr` for standardization. This means the previous aliases would no longer be available. See git plugin README for list of deprecated aliases.

## Other comments:

Fixes #11104


zee-bit:
Sorry, the commit was incomplete earlier (forgot to update, because of dinner). I had to force-push with the new changes to the same commit. Let me know if there's any other inconsistency.

LunarWatcher (approved):
LGTM. But this is a breaking change, so the PR needs to indicate that. See https://www.conventionalcommits.org/en/v1.0.0/ if you're not familiar with how

zee-bit:
Hey @LunarWatcher, I have updated the PR desc as well as the commit to indicate a breaking change as per the link you provided. Please let me know if I am still missing anything. :)

esemeniuc:
would love to see this also

rafaelbernard (approved):

nicolasbarra:
Excited to see this merged after the conflicts have been resolved!

zee-bit:
Updated!

tjarbo (changes requested):
Reference to https://github.com/ohmyzsh/ohmyzsh/issues/11104#issuecomment-1336491627

esemeniuc:
Bump to see this merged!

rivenirvana:
+1 for merge

carlosala:
@mcornella @adamchainz I'd like to hear your comments about it. IMO it makes sense, but it's kinda of a super crazy breaking change.
Maybe we could add a message for it (something like that `gup` is now deprecated and you should use `gpr`)?

adamchainz:
Do we really need `--rebase` aliases anyway? One can configure `pull.rebase = true` and it will be used by default. I imagine most people using these aliases would be better off making it the default.

But if we are to keep them, I’m fine with a breaking change, perhaps with a warning for a period.

mcornella:
We're currently polling on whether to proceed with this change on [this Discord message](https://discord.com/channels/642496866407284746/829052424723365889/1158467038537334836), in case you want to get your voice heard.

> Do we really need --rebase aliases anyway? One can configure pull.rebase = true and it will be used by default. I imagine most people using these aliases would be better off making it the default.

I agree with you that a config option is best for your use case if you do this by default. These aliases solve the use case of people who do different things based on different repositories (think personal vs company ones), for which an alias is significantly less overhead than custom configs per repository / path.

mcornella:
Consensus is positive towards the change, reassigning to me to implement the deprecation notice functions for 1-2 months to accomodate alias users.

u3u:
```sh
[oh-my-zsh] 'gup' is a deprecated alias, using 'gpr' instead.
```

⚠️ Please do not deprecate it!!! We have already grown accustomed to it. We will be distressed if the command suddenly becomes unusable!

PR #11226, opened 2022-10-04
Update README.md

abhi99555:
Update README.md with proper syntax


mcornella:
As is it's correct grammar.

PR #11229, opened 2022-10-05
fix(gh): plugin only creates completion cache when needed

casaper:
Completions were generated and cached on every shell opened. This fix caches the version the cache was last generated for and only runs the completion generator in case there is a mismatch or no completion cache file available.

Also fixed the readme, to state correct path for completions cache and version cached file.

## Standards checklist:

## Changes:

- cache version of gh the completions were last cached for
- test if completion cache for cached version available
- only generate completions when version mismatch or no cache present

### DIFF
--- a/plugins/gh/README.md
+++ b/plugins/gh/README.md
@@ -17,7 +17,7 @@ plugin is loaded, which is usually when you start up a new terminal emulator.
 
 The cache is stored at:
 
-- `$ZSH/plugins/gh/_gh` completions script
+- `$ZSH_CACHE_DIR/completions/_gh` completions script
 
-- `$ZSH_CACHE_DIR/gh_version` version of GitHub CLI, used to invalidate
+- `$ZSH_CACHE_DIR/gh_cached_version` version of GitHub CLI, used to invalidate
   the cache.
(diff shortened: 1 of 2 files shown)


carlosala:
Hi! All this plugins (like `gh`, `deno`, and many others) are built in the same way that doesn't block execution of the shell and doesn't affect the performance.
Moreover, most of this calls are implemented by a famous rust library, and take less than 0.5ms.
Changing one would mean losing all generalisation and, in my opinion, is worst than removing some unneeded executions that doesn't affect performance.
Thanks for your contribution!

casaper:
@carlosala 

> Hi! All this plugins (like `gh`, `deno`, and many others) are built in the same way that doesn't block execution of the shell and doesn't affect the performance. Moreover, most of this calls are implemented by a famous rust library

Well, I guess you're right about this if it comes to rustup and even more so cargo (the file isn't even generated in that case).

> and take less than 0.5ms

Which is quite a bit, when cumulated with multiple plugins.

However:
I'm quite sure that what you say is not true for gh and especially not for bitwarden #11181 (it is a JavaScript application).
I had these caching changes active on my computers, and I claim I felt a difference in shell start up time, if I have it active.

Anyway:
I won't argue with you about this PR or even less about #11230 - I can and will totally accept you don't want this merged, and I appreciate your maintenance efforts (that includes denial of contributions in some cases as well). I know and respect that.

But the fact that I will keep the alternations on my personal setup, despite the effort it takes for me, to rebase with origin/master every now and then, might point out to you, that there could still be some relevance to this for some.

PR #11230, opened 2022-10-05
fix(rust): plugin only create completion cache when needed

casaper:
Completions were generated and cached on every shell opened. This fix caches the version the cache was last generated for and only runs the completion generator in case there is a mismatch or no completion cache file available.

## Standards checklist:

## Changes:

- cache version of rustc and cargo the completions were last cached for
- test if completion cache for cached version available
- only generate completions when version mismatch or no cache present


carlosala:
See https://github.com/ohmyzsh/ohmyzsh/pull/11229#issuecomment-1284143170

PR #11233, opened 2022-10-05
feat(tmux): Don't autostart `tmux` when using vscode's terminal

HeroCC:
Currently, we ignore editor-like terminal programs when checking tmux autolaunch. VSCode is not included in this check, resulting in tmux attaching when using the VSCode embedded terminal, preventing some autolaunches from working, and rendering the VSCode windows & panes unusable.

This PR will prevent tmux from autostarting if we are inside of vscode, fixing these usability issues.

## Standards checklist:

## Changes:

- Check if `$TERM_PROGRAM` is set to `vscode`, and if it is, do not autolaunch tmux

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -100,7 +100,7 @@ compdef _tmux _zsh_tmux_plugin_run
 alias tmux=_zsh_tmux_plugin_run
 
 # Autostart if not already in tmux and enabled.
-if [[ -z "$TMUX" && "$ZSH_TMUX_AUTOSTART" == "true" && -z "$INSIDE_EMACS" && -z "$EMACS" && -z "$VIM" ]]; then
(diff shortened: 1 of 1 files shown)


tjarbo (approved):
LGTM 👍

robbyrussell:
Thanks for this, and sorry it sat so long. We've since merged the same shape twice, for IntelliJ in #11927 and Zed in #13220, so there's no objection to the approach here. VS Code is just the gap nobody closed.

The line moved while this waited, so it doesn't apply any more. Could you rebase? On current master it's `plugins/tmux/tmux.plugin.zsh:195`, and the condition to add is the same one you already have:

```zsh
if [[ -z "$TMUX" && "$ZSH_TMUX_AUTOSTART" == "true" && -z "$INSIDE_EMACS" && -z "$EMACS" && -z "$VIM" && -z "$INTELLIJ_ENVIRONMENT_READER" && -z "$ZED_TERM" && "$TERM_PROGRAM" != "vscode" ]]; then
```

Worth mentioning in the README too: this covers Cursor and Windsurf, since they inherit the same `TERM_PROGRAM` value.

If you'd rather not pick it back up after this long, say so and I'll apply it myself and keep your commit.

PR #11234, opened 2022-10-05
fix(git-auto-fetch): avoid password prompt with `GIT_TERMINAL_PROMPT=0`

Syphdias:
If the remote is using HTTP authentication (and does not use a credential helper) it triggers writes to the terminal that cannot be responded to. This happens every time the repository is fetched by git-auto-fetch.

`GIT_TERMINAL_PROMPT=0` prevents prompts to the terminal. The man page explicitly names HTTP authentication as an example: https://git-scm.com/docs/git#Documentation/git.txt-codeGITTERMINALPROMPTcode

## Standards checklist:

## Changes:

Adds environment variable to `git fetch` command in the git-auto-fetch plugin

## Other comments:

Actually I am not 100% on if the indentation is correct. I believe it is though.
Also would be great if the PR would count for the Hacktoberfest even tough it is just one line :)


Syphdias:
Fixed commit message.

carlosala (approved):

mcornella:
Thanks for the contribution!

PR #11245, opened 2022-10-07
feat(git-prompt): show deleted files

lochel:
It seems that git-prompt isn't indicating deleted files. I couldn’t find a related issue or pull-request and have solved the issue following the code pattern used in `plugins/git-prompt`.

## How to reproduce the issue that is solved

Enable `git-prompt` and delete a file in any git repository. The prompt will not indicate it. This is solved by this pull request.

## Standards checklist:

## Changes:

- plugins/git-prompt/git-prompt.plugin.zsh:

Adding `ZSH_THEME_GIT_PROMPT_DELETED`

- plugins/git-prompt/gitstatus.py

Returning the number of deleted files

## Other comments:

Related but outdated pull request: #2054


carlosala (changes requested):
I like it! Could you put that at the end of each list, so we don't break anything?

carlosala:
Thanks for your contribution!

Issue #11257, opened 2022-10-14
Current Taskwarrior alias does not work?


manoelpqueiroz:
### Describe the bug

I would just like to make sure that the current implementation of Taskwarrior's [plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/taskwarrior/taskwarrior.plugin.zsh) should in fact not work due to the lack of ticks.

Currently, the plugin sets a general alias to reduce the command with `alias t=task`. However, my Zsh shell does not recognise `t` as a command after sourcing `zshrc`.

Should it be `alias t='task'` instead? Does the lack of ticks render the alias unusable? One thing that confuses me if that the plugin also implements an autocompletion command right after setting the alias, which is working fine with my installation.

### Steps to reproduce

1. Enable plugin `taskwarrior`
2. Source `.zshrc`
3. Run command `t`

### Expected behavior

By simply typing `t` at the terminal, a list of current tasks should show up at the terminal.

### OS / Linux distribution

Arch Linux

### Zsh version

5.9

### Oh My Zsh version

master (18e7e5d0)

### Terminal emulator

Konsole

rwmitchell:
Have you tried entering
```
alias t='task`
```
at the cmdline to see if it started working?  You can set aliases right in the current shell without going thru plugins or .zshrc files.

Have you installed taskwarrior ?  Is it installed in your PATH ?

manoelpqueiroz:
Yes, I have Taskwarrior installed and setting the alias for my session worked out. I'll open a pull request to correct the plugin's behaviour.

rwmitchell:
hmmm, your results were unexpected, at least on my system, the single quotes around task is not required.

manoelpqueiroz:
Now that's interesting. Any recommendations as to how to debug and try to reach the root-cause?

rwmitchell:
The single quotes are necessary when the thing being aliased includes multiple items, in this case there is only one.
You can test this on your system by repeating the previous test, but without the single quotes.

When I ```source taskwarrior.plugin.zsh``` t returns "command not found: task", which is correct as I don't have it installed, but it does show the full command.  Something in your .zshrc must be overriding  't'

What does ```which t``` report after sourcing .zshrc ?

manoelpqueiroz:
@rwmitchell, sorry for the late reply. But in my case, even by manually sourcing `.zshrc`, it does not recognise `t` as a command.

[...]

PR #11258, opened 2022-10-14
Update to Taskwarrior plugin with new aliases

manoelpqueiroz:
## Standards checklist:

## Changes:

- Corrected the `t` alias syntax to enable its usage;
- Added 49 aliases based on common Taskwarrior commands
- Added one alias to simplify a manual operation.

## Other comments:

This is my first PR ever opened on Github, thank you for allowing me to contribute to the project.

Acknowledging #6571, I have added aliases to the Taskwarrior plugin based on very common commands the program gives us and are listed on their official [documentation](https://taskwarrior.org/docs/commands/). Notable expceptions to this are:

- All standard reports receive a set of three additional aliases to easily filter tasks that are due in the day, week and month.
- Since Taskwarrior version 2.6, the user has the option to "purge" tasks instead of deleting them. This allows users to completely wipe out unwanted data from their program. However, in order to purge a task, one must get its UUID. I have made an alias to easily use the UUIDs of _all_ deleted tasks as the input to the command.

Duplicate aliases that conflict with other plugins are highlighted with comments, please instruct wether they should be renamed or not.

I do not know why the other PR was closed, the thread ended without a clear explanation. Should this implementation not be of the developers' interest, please let me know.

### DIFF
--- a/plugins/taskwarrior/README.md
+++ b/plugins/taskwarrior/README.md
@@ -1,8 +1,8 @@
 # taskwarrior
 
-This plugin adds smart tab completion for [TaskWarrior](https://taskwarrior.org/).
-It uses the zsh tab completion script (`_task`) shipped with TaskWarrior for the
-completion definitions.
+This plugin adds smart tab completion for [TaskWarrior][1]. It uses the zsh tab
+completion script (`_task`) shipped with TaskWarrior for the completion
+definitions.
 
 To use it, add `taskwarrior` to the plugins array in your zshrc file:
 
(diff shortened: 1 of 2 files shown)


manoelpqueiroz:
Waiting to check the situation with #11257 to advance.

carlosala:
Hi @manoelpqueiroz! I think we have too many aliases in this PR. We have been following a new [contributing guide](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases) that encourages to add not many aliases to omz. Because of this, I'd ask to to remove most of them.
Thanks!

mcornella:
Agree with Carlo, this is a lot of aliases when there is completion available and you probably don't use all of them on a daily basis. You should also check for not overriding existing commands with something like https://command-not-found.com/.

carlosala:
Closing for the moment as there are too many aliases. Feel free to re-open if you want to continue with this!

Issue #11244, opened 2022-10-07
Don't use abbreviations for git alias flags


jdevries3133:
### If the feature request is for a plugin or theme, specify it here.

Git

### If the feature solves a problem you have, specify it here.

I often lookup oh-my-zsh aliases as follows:

```zsh
alias | grep git
alias | grep "git commit"
```

Here are the results of the latter command:

```
gcas='git commit -a -s'
gcasm='git commit -a -s -m'
gcmsg='git commit -m'
'gcn!'='git commit -v --no-edit --amend'
```

I'm wondering: What is the `-a` flag? What is the `-s` flag? I have to separately go and lookup this information.

### Describe the proposed feature.

I'd like aliases to stay away from using abbreviated flags, and to use the verbose versions instead where available. I'd like to implement this feature for the git plugin, at least, if this feature is desired.

### Describe alternatives you've considered

I think it's possible that there's a blind spot in my workflow as stated above; please let me know if that's the case!

sevy2011:
man git
```
HIGH-LEVEL COMMANDS (PORCELAIN)
       We separate the porcelain commands into the main commands and some ancillary user utilities.

   Main porcelain commands
[...]
       git-commit(1)
           Record changes to the repository.
```

man git-commit
```
OPTIONS
       -a, --all
           Tell the command to automatically stage files that have been modified and deleted, but new files you have not told Git about are not affected.
[...]
       -s, --signoff, --no-signoff
           Add a Signed-off-by trailer by the committer at the end of the commit log message. The meaning of a signoff depends on the project to which you’re
           committing. For example, it may certify that the committer has the rights to submit the work under the project’s license or agrees to some contributor
           representation, such as a Developer Certificate of Origin. (See http://developercertificate.org for the one used by the Linux kernel and Git
           projects.) Consult the documentation or leadership of the project to which you’re contributing to understand how the signoffs are used in that
           project.

           The --no-signoff option can be used to countermand an earlier --signoff option on the command line.
```

jdevries3133:
Absolutely, just wanted to hear if the idea was supported first. I'll get started!

jdevries3133:
Absolutely, just wanted to hear if the idea was supported first. I'll get started!

mcornella:
The only downside to this is to have the aliases documented in the README as well, which has limited space for the full-style arguments.

[...]

PR #11259, opened 2022-10-15
feat(#11244): verbose git flags in aliases

jdevries3133:
## Standards checklist:

## Changes:

Use verbose flags in git aliases

## Other comments:

I figure these commits will be squashed and merged so I'm not sure if this nails the commit guidelines, but feel free to merge / reword as you see fit.


esemeniuc (approved):

esemeniuc:
Would love to see this merged!

carlosala:
Hey! Could you explain a bit why this would be great? Thanks!

rwmitchell:
It makes the aliases readable when expanding an alias to see what it is.

carlosala:
It makes total sense!

plugins/git/git.plugin.zsh
```diff
@@ -103,20 +103,20 @@ function gccd() {
...
+alias gclean='git clean --interactive -d'
 alias gpristine='git reset --hard && git clean -dffx'
 alias gcm='git checkout $(git_main_branch)'
 alias gcd='git checkout $(git_develop_branch)'
-alias gcmsg='git commit -m'
+alias gcmsg='git commit --message'
 alias gco='git checkout'
 alias gcor='git checkout --recurse-submodules'
-alias gcount='git shortlog -sn'
+alias gcount='git shortlog --summary -n'
```
tjarbo, line 113:
```suggestion
alias gcount='git shortlog --summary --numbered'
```
https://git-scm.com/docs/git-shortlog#Documentation/git-shortlog.txt---numbered

plugins/git/git.plugin.zsh
```diff
@@ -267,15 +267,15 @@ alias grst='git restore --staged'
 alias grt='cd "$(git rev-parse --show-toplevel || echo .)"'
 alias gru='git reset --'
 alias grup='git remote update'
-alias grv='git remote -v'
+alias grv='git remote --verbose'
 
-alias gsb='git status -sb'
+alias gsb='git status --short -b'
```
tjarbo, line 272:
```suggestion
alias gsb='git status --short --branch'
```
https://git-scm.com/docs/git-status#Documentation/git-status.txt---branch

plugins/git/git.plugin.zsh
jdevries3133: Oops, there's no reason that was just a mistake! I used some regexes to fix most flags, then came back for the grouped ones (i.e. `-sn`) by hand. Good catch! I'll fix these.

jdevries3133:
OK, I took another look and fixed the two that @tjarbo pointed out as well as a few more.

nicolasbarra:
Excited to see this merged after the conflicts have been resolved!

plugins/git/git.plugin.zsh
```diff
@@ -103,20 +103,20 @@ function gccd() {
...
-alias gcount='git shortlog -sn'
+alias gcount='git shortlog --summary --numbered'
 alias gcp='git cherry-pick'
 alias gcpa='git cherry-pick --abort'
 alias gcpc='git cherry-pick --continue'
 alias gcs='git commit -S'
-alias gcss='git commit -S -s'
-alias gcssm='git commit -S -s -m'
+alias gcss='git commit -S --signoff'
+alias gcssm='git commit -S --signoff --message'
```
tjarbo, line 119: `-S` should be replaceable with `--gpg-sign` according to the [git documentation](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt)

tjarbo (changes requested):
Hi @jdevries3133,
really great! Just found one missing parameter that is replaceable with a more descriptive one. An execption is the `gunwip` alias, where the `-n` could be replaced as well, but I would say that it does not make that much sense.

best regards
tjarbo

plugins/git/git.plugin.zsh
jdevries3133: OK I finally got the chance to fix this up; I am pretty sure there are no stragglers left but looking forward to your review. Thanks for your patience!

jdevries3133:
Also, merge conflicts have been resolved. Two notes while merging with upstream:

- I noticed that `gcn!` was removed; I was sure to carry that change forward
- I noticed that Carlos Sala had a patch (0145d744a9c4c11f00992f7f3ad9555bc8ac6177) to change `gunwip` message; I was careful to include that change as well

tjarbo (changes requested):
Hi @jdevries3133,
I just found a little mistake that sneaked in during your last commits. According to the logs, all your changes for `git commit` (line 85-95) got reverted. Please check again :) 

https://github.com/ohmyzsh/ohmyzsh/pull/11259/commits/7300030e1f8c71cb4910ed3bae9a505436099231#diff-bd72f3b7bb82839b6e8f34b96e7fc5ec3c923d24b79bec12427df62f07136678R85

best regards
tjarbo

jdevries3133:
> According to the logs, all your changes for `git commit` (line 85-95) got reverted.

Ah, nice catch; fixed in 2684f73db1f21c0b48e60b5c6470d479efd4150b

tjarbo (approved):
Perfect @jdevries3133! Could not find anything else left 👍

carlosala:
This PR is becoming really great! Did you checked that every alias is documented in the README, for example something like #10101? And maybe we could also bundle the idea of #9743 in this PR, what do you think @jdevries3133?

jdevries3133:
> Did you checked that every alias is documented in the README, for example something like #10101?

I'm pretty sure that every alias documentation is updated correctly, that's certainly my intention, but feel free to review the changes of course. 

> And maybe we could also bundle the idea of #9743 in this PR, what do you think @jdevries3133?

I already stupidly made an attempt at formatting the table which caused a bit of a gnarly merge conflict. I want to avoid sweeping changes to the readme, but I'm happy to do this in a follow up PR if we can turn it around quickly to minimize any conflict issues.

carlosala:
Thanks for your contribution @jdevries3133 and @tjarbo for the reviews. I've just squashed and pushed all this PR to 35454190f6aeae9b9f16acdbdd0cf72c32610ee8.
I'm really happy we managed to close this!

PR #11283, opened 2022-10-22
feat(aliases): Add git grep alias

trungqudinh:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -197,6 +197,7 @@ alias ggpur='ggu'
 alias ggpull='git pull origin "$(git_current_branch)"'
 alias ggpush='git push origin "$(git_current_branch)"'
 
+alias ggr='git grep -n --color=always'
 alias ggsup='git branch --set-upstream-to=origin/$(git_current_branch)'
 alias gpsup='git push --set-upstream origin $(git_current_branch)'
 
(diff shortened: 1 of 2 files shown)


ParthJadhav (approved):

plugins/git/git.plugin.zsh
```diff
@@ -197,6 +197,7 @@ alias ggpur='ggu'
 alias ggpull='git pull origin "$(git_current_branch)"'
 alias ggpush='git push origin "$(git_current_branch)"'
 
+alias ggr='git grep -n --color=always'
```
tjarbo, line 200: To keep it in sync with #11259, `-n` could be replaced with `--line-number`. Documentation [`git grep --line-number`](https://git-scm.com/docs/git-grep#Documentation/git-grep.txt--n)

adamchainz: Also, [the `--color` option](https://git-scm.com/docs/git-grep#Documentation/git-grep.txt---colorltwhengt) defaults to `always`, so we don't need to provide it.

mcornella:
This PR has been long open and there is no traction or user engagement, that's why I'm closing the PR. If this is still something you wish to have, consider setting the alias in your own `.zshrc` file.

PR #11302, opened 2022-10-29
fix(brew): suppress error message of brew cask on Linux

liblaf:
## Standards checklist:

## Changes:

- fix(brew): suppress error message of brew cask on Linux


liblaf:
Homebrew Cask is not supported on Linux yet. Every time I use the `brews` alias, it gives me a bunch of useless error messages.

mcornella:
Thanks!

PR #11307, opened 2022-10-30
fix(per-directory-history): prevent adding blank entries

alkssnz:
## Standards checklist:

## Changes:

- prevent adding blank entries 

## Other comments:

Default behavior doesn't allow adding blank entries.


carlosala:
Could you open the PR to https://github.com/jimhester/per-directory-history? Then we can just get the latest version from upstream, so we don't diverge.
Thanks for the contribution, it's really great though! 🎉

mcornella:
See https://github.com/ohmyzsh/ohmyzsh/pull/11225#issuecomment-1268394021. This might not be needed.

PR #11318, opened 2022-11-05
feat(git): display tag in git-prompt when possible

b4nst:
## Standards checklist:

- [] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- Add tag infromation to git prompt

## Other comments:

...

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -19,6 +19,7 @@ function git_prompt_info() {
 
   local ref
   ref=$(__git_prompt_git symbolic-ref --short HEAD 2> /dev/null) \
+  || ref=$(__git_prompt_git describe --exact-match HEAD 2> /dev/null) \
   || ref=$(__git_prompt_git rev-parse --short HEAD 2> /dev/null) \
   || return 0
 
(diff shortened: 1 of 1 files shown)


lib/git.zsh
```diff
@@ -19,6 +19,7 @@ function git_prompt_info() {
 
   local ref
   ref=$(__git_prompt_git symbolic-ref --short HEAD 2> /dev/null) \
+  || ref=$(__git_prompt_git describe --exact-match HEAD 2> /dev/null) \
```
carlosala, line 22: Should we use `command git describe --tags --exact-match HEAD`? According to git documentation, this would also match unannotated tags, I'd expect this kind of tags to match also. What do you think?
```
       --tags
           Instead of using only the annotated tags, use any tag found in refs/tags
           namespace. This option enables matching a lightweight (non-annotated) tag.
```

carlosala: For example, tags created by Github when doing a release are lightweight tags.
I found more information about this topic (that was totally unknown) here: https://stackoverflow.com/questions/4971746/why-should-i-care-about-lightweight-vs-annotated-tags

mcornella: Agreed, I'll make the changes.

mcornella:
```suggestion
  || ref=$(__git_prompt_git describe --tags --exact-match HEAD 2> /dev/null) \
```

PR #11319, opened 2022-11-06
Revert "fix(debian)!: remove ar alias" and add apr alias

tmiland:
This reverts commit 818f3de.## Standards checklist:

## Changes:

- Replaces `ar` command with `apr`, instead of removing it completely.

## Other comments:

...


carlosala:
Hi! We think that bundling aliases that use `sudo` it's not the best option nowadays, it's one of the reasons we removed it. This kind of aliases are better for the `zshrc`!
Thanks for your contribution 😊

tmiland:
> Hi! We think that bundling aliases that use `sudo` it's not the best option nowadays, it's one of the reasons we removed it. This kind of aliases are better for the `zshrc`!
> Thanks for your contribution 😊 

Howdy! 
That's understandable. 

Thanks for removing the alias anyways, now i know why the ar command didn't work a couple weeks back 😅

PR #11324, opened 2022-11-08
fix(macos): fix `man-preview` for macOS Ventura

cxy004:
## Standards checklist:

## Changes:

- Preview.app on macOS Ventura drops support for PostScript on which the current `man-preview` function with `man -t` (basically `mandoc -Tps`) relies. This commit switches to `mandoc -Tpdf` and keeps it simple.

### DIFF
--- a/plugins/macos/macos.plugin.zsh
+++ b/plugins/macos/macos.plugin.zsh
@@ -224,8 +224,9 @@ function quick-look() {
 }
 
 function man-preview() {
+  local location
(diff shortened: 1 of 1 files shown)


mcornella:
Is `mandoc` installed by default on macOS or would it need to be a separate dependency?

rwmitchell:
Default installed on Monterey

robbyrussell:
@mcornella by default comes included on Ventura

PR #6709, opened 2018-04-05
Renew terraform completions

ryosan-470:
See https://www.terraform.io/docs/commands/index.html#shell-tab-completion

I add generated `terraform -install-autocompletion` and remove old one.

PR #11330, opened 2022-11-10
Add Terraform auto-complete

### LINKED ISSUE #11328: Use native auto-completion feature provided by vendor on Terraform

terraform

Terraform could be update with newer options and auto-complete delivered with ohmyzsh it would be old. It wouldn't be nice to miss the new options because the auto-complete is outdated.

My idea is to use this command to auto-complete automatically, if the binary is available on path (like using packaging to install): `complete -o nospace -C /usr/bin/terraform terraform`.
`kubectl` plugin has a similar feature when is used `kubectl completion zsh` to populate the auto-completion with the most recent options available.

I'm using the command inside the `~/.zshrc` file directly, but it could be placed inside ohmyzsh so the feature could be use with minimum effort.

### Additional context

_No response_

### Related Issues

_No response_

mhalano:
## Standards checklist:

## Changes:

- Add automatic auto-complete using complete option instead of a static file that needs to be
updated from time to time

## Other comments:

It does the same as PR #6709, but the PR was closed without merge.

Fixes #11328 
...


carlosala:
Thanks for your contribution 😊

PR #11335, opened 2022-11-11
feat(nvm)!: settings now are zstyle-based

carlosala:
## Standards checklist:

## Changes:

Move plugin customization to `zstyle` and soft-deprecate old options.

### DIFF
--- a/plugins/nvm/README.md
+++ b/plugins/nvm/README.md
@@ -1,7 +1,7 @@
 # nvm plugin
 
-This plugin adds autocompletions for [nvm](https://github.com/nvm-sh/nvm) — a Node.js version manager.
-It also automatically sources nvm, so you don't need to do it manually in your `.zshrc`.
+This plugin adds autocompletions for [nvm](https://github.com/nvm-sh/nvm) — a Node.js version manager. It also
+automatically sources nvm, so you don't need to do it manually in your `.zshrc`.
 
 To use it, add `nvm` to the plugins array of your zshrc file:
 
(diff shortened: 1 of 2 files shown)


plugins/nvm/nvm.plugin.zsh
```diff
@@ -16,14 +16,37 @@ fi
...
+if (( $+NVM_LAZY_CMD )); then
+  echo "[oh-my-zsh] Instead of NVM_LAZY_CMD=$NVM_LAZY_CMD, please use"
+  echo "[oh-my-zsh] zstyle ':omz:plugins:nvm' lazy-cmd $NVM_LAZY_CMD"
+  zstyle ':omz:plugins:nvm' lazy-cmd $NVM_LAZY_CMD
+fi
+if (( $+NVM_AUTOLOAD )); then
+  echo "[oh-my-zsh] Instead of NVM_LAZY=1, please use"
+  echo "[oh-my-zsh] zstyle ':omz:plugins:nvm' lazy true"
+  zstyle ':omz:plugins:nvm' autoload true
+fi
```
mcornella, line 38: Not a fan of these being so verbose, I think it's going to scare some people and will not read the entire text. I think it's better to do it once, write the output to stderr, and separate the information text and the alternative `zstyle` commands, sort of like this:

```
# text in yellow
[nvm plugin] Variable-style settings are deprecated, instead of $<variable-names-used-by-the-user>, use:

[nvm plugin] zstyle ':omz:plugins:nvm' lazy
[nvm plugin] zstyle ':omz:plugins:nvm' lazy-cmd
[nvm plugin] zstyle ':omz:plugins:nvm' autoload
# ^ this list only shows the settings used by the user
```

Also good call in writing the exact `lazy-cmd` value used by the user in the warning.

plugins/nvm/README.md
```diff
@@ -21,14 +21,32 @@ These settings should go in your zshrc file, before Oh My Zsh is sourced:
...
-- **`NVM_AUTOLOAD`**: if `NVM_AUTOLOAD` is set to `1`, the plugin will automatically load a node version when
-  if finds a [`.nvmrc` file](https://github.com/nvm-sh/nvm#nvmrc) in the current working directory indicating
-  which node version to load.
+This option will help you to defer nvm's load until you use it to speed-up your zsh startup. This will source
+nvm script only when using it, and will create a function for `node`, `npm`, `pnpm`, `yarn`, and the
+command(s) specified by `lazy-cmd` option, so when you call either of them, nvm will be loaded and run with
+default version. To enable it, you can add this snippet to your zshrc, before Oh My Zsh is sourced:
+
+```zsh
+zstyle ':omz:plugins:nvm' lazy true
```
mcornella, line 34: We should standardize on what value we suggest for boolean options. I remember the `ssh-agent` plugin uses `lazy yes` instead of `true`, and we should look at how we're doing it on other plugins.

mcornella (commented):
Looks very nice and much more streamlined. It is also better that we're not polluting the variable namespace with variables that are not used by nvm, which could be expected seeing the `NVM_` prefix.

Left some notes, but this is almost ready!

plugins/nvm/README.md
carlosala: Other plugins are using `yes` as the boolean in the README, but `yarn` plugin is recommending `false`. We can move everything to `yes` and `no` (even though both are going to work the same way and are compatible).

mcornella:
Changed some things about how the variable settings were checked and refined the style of the deprecation notice:
image

I also moved the lazy function to be eval-ed so that we don't keep an extra variable around.

Let me know what you think!

mcornella (approved):

PR #11336, opened 2022-11-12
feat(git-prompt): add option to show upstream branch

lochel:
## Standards checklist:

## Changes:

The upstream branch will be displayed if `ZSH_THEME_GIT_SHOW_UPSTREAM` is defined.

The prompt may look like the following:

- `(prompt -> origin/prompt ↑3|✚1)`: on branch `prompt`, ahead of remote branch `origin/prompt` by 3 commits, 1 file changed but not staged
- `(prompt|●2)`: branch `prompt` isn't set up with an upstream branch


carlosala (commented):
In my opinion this is too much verbose. Almost always a branch will point to origin/branch. Why do you think this is important/valuable?

lochel:
In my experience the origin remote is usually read-only and mostly used to branch off feature branches. Such features branches are then usually connected to some kind of fork with write access. Also reviewing pull requests is easier when the upstream branch is visible in the terminal. 

And this new feature is optional and disabled by default, so it shouldn't bother people not interested in it.

carlosala:
Nice, then let's go for it! Could you update the readme?

plugins/git-prompt/git-prompt.plugin.zsh
```diff
@@ -101,6 +107,8 @@ ZSH_THEME_GIT_PROMPT_AHEAD="%{↑%G%}"
 ZSH_THEME_GIT_PROMPT_UNTRACKED="%{$fg[cyan]%}%{…%G%}"
 ZSH_THEME_GIT_PROMPT_STASHED="%{$fg_bold[blue]%}%{⚑%G%}"
 ZSH_THEME_GIT_PROMPT_CLEAN="%{$fg_bold[green]%}%{✔%G%}"
+#ZSH_THEME_GIT_SHOW_UPSTREAM=1
```
carlosala, line 110: Could you delete this line? Thanks!

carlosala (changes requested):

carlosala:
Thanks for your contribution!

PR #11345, opened 2022-11-18
fix(init): prevent multiple compdump compilations

### LINKED ISSUE #11341: zrecompile race condition when opening several terminals at once
### Describe the bug

I use the i3 window manager.  I have autologin configured and automatically start six terminals (alacritty) with zsh on the start of i3.  In several of those terminals I often get

``` zrecompile:zcompile:99: invalid zwc file: /home/w/.zcompdump-huette-5.9.zwc ```

I suspect that it is some kind of race condition when multiple processes try to update the zwc file at once.

### Steps to reproduce

It can be reproduced more or less reliably with (something like) the following:

``` touch ~/.zcompdump-${HOST}-5.9; alacritty&; alacritty&; alacritty& ```

When I run the previous command with my usual configuration the error appears almost every time in at least one of the terminals spawned.  However, when I create a new user on my machine with a minimal configuration, then it happens far less often.  But it does happen.

It may be harder to reproduce on more recent hardware.  I am working on an about 10 years old system with an "Intel(R) Celeron(R) CPU N3150 @ 1.60GHz".

### Expected behavior

No error message. Maybe some kind of locking.

_No response_

### OS / Linux distribution

Arch Linux

### Zsh version

5.9

### Terminal emulator

alacritty

_No response_

### Additional context

_No response_

carlosala:
## Standards checklist:

## Changes:

- Compile compdump only if there's no lockfile

Closes #11341

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -143,7 +143,10 @@ fi
 unset zcompdump_revision zcompdump_fpath zcompdump_refresh
 
 # zcompile the completion dump file if the .zwc is older or missing.
-zrecompile -q -p "$ZSH_COMPDUMP" && command rm -f "$ZSH_COMPDUMP.zwc.old"
+if command mkdir "${ZSH_COMPDUMP}.lock" 2>/dev/null; then
+  zrecompile -q -p "$ZSH_COMPDUMP"
+  command rm -rf "$ZSH_COMPDUMP.zwc.old" "${ZSH_COMPDUMP}.lock" 
+fi
 
 # Load all of the config files in ~/oh-my-zsh that end in .zsh
 # TIP: Add files you don't want in git to .gitignore


mcornella:
File creations (with touch) are not atomic, there could be two processes finding no lock file and enter the if at the same time before the `touch` runs.

See the `check_for_upgrade.sh` script for an example of a lock mechanism using `mkdir`.

There is also a lock functionality in a module, look for `zsystem flock` in https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html.

carlosala:
@mcornella yas, thanks for the hint! I tested launching multiple instances of `zsh` (and echoing stuff to be sure) and I everything works well from my side!

mcornella:
Looks good to me, good job!

@whxvd can you test if this solves your issue? You can check out this PR with `omz pr test 11345`. Thanks!

whxvd:
When I run `touch $ZSH_COMPDUMP; alacritty&; alacritty&; alacritty&` without the PR, the error appears at least half of the times. With the PR at least 30 tests in succession yielded no error. So it seems to work! Thanks a lot!

mcornella:
Thanks for testing!

carlosala:
Thanks for the contribution!!!

PR #11346, opened 2022-11-19
fix(last-working-dir): use builtin `pwd`

rwmitchell:
Use system 'pwd' instead of possible user's alias or function.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


carlosala:
`pwd` is also a `builtin` from `zsh`, I'd tend to use builtins over external binaries

plugins/last-working-dir/last-working-dir.plugin.zsh
```diff
@@ -9,7 +9,7 @@ chpwd_last_working_dir() {
   [[ "$ZSH_SUBSHELL" -eq 0 ]] || return 0
   # Add ".$SSH_USER" suffix to cache file if $SSH_USER is set and non-empty
   local cache_file="$ZSH_CACHE_DIR/last-working-dir${SSH_USER:+.$SSH_USER}"
-  pwd >| "$cache_file"
+  command pwd >| "$cache_file"
```
carlosala, line 12: Let's change to `builtin`

carlosala (changes requested):

carlosala:
Thanks for your contribution! 😉

PR #11347, opened 2022-11-20
feat: add modules plugin

ifurther:
## Standards checklist:

## Changes:

Add the Environment Modules plugin


carlosala:
I'd say this plugin is actually not needed. We have other options more standard to source and autoload `zsh` files. I see in Environment Modules documentation that is meant to change between versions from different binaries, etc, and also is "only" 500 stars on github, so it'd be a better option to go to custom plugins.
Even though, thanks for your contribution, and feel free to comment what do you think about! 😉

PR #11350, opened 2022-11-23
feat(fluxcd): add completion for flux

serrovsky:
## Standards checklist:

## Changes:

- Add fluxcd cli as a plugin to enable autocompletion

## Other comments:
Done, after this comment on this issue >
https://github.com/ohmyzsh/ohmyzsh/pull/10197#issuecomment-1324766214
...


plugins/flux/README.md
```diff
@@ -0,0 +1,9 @@
+# FluxCD plugin
+
+This plugin adds completion for [Flux](https://helm.sh/), an open and extensible continuous delivery solution for Kubernetes. Powered by GitOps Toolkit.
```
carlosala, line 3: Should we have https://fluxcd.io?

carlosala (changes requested):

plugins/flux/README.md
carlosala: I'd also name the plugin `fluxcd`, just to be more semantic in terms of not confusing with other projects named flux, such as https://github.com/influxdata/flux

serrovsky: Yes, make sense. I will change it right now

serrovsky (commented):
Plugon renamed and link updated

carlosala:
Thanks for your contribution!

PR #11356, opened 2022-11-27
fix(changelog): generate correct commit link in markdown

jzhang046:
## Standards checklist:

## Changes:

Currently `tools/changelog.sh` would generate same link of `https://github.com/ohmyzsh/ohmyzsh/commit/` for each commit in markdown mode, without the commit hash properly included in the link
* Currently: ``[`my-commit-hash`](https://github.com/ohmyzsh/ohmyzsh/commit/)``
* Expected: ``[`my-commit-hash`](https://github.com/ohmyzsh/ohmyzsh/commit/my-commit-hash)``


mcornella:
Good catch, thanks!

Issue #11369, opened 2022-12-05
Plugin: zsh-script-prefix


tmiland:
Hello,

just wanted to share a custom plugin which prefixes a local script file name. 

Say you have typed a script name and forgot to add `./` in front:

```bash
$ somescript.sh
```

By pressing the <kbd>esc</kbd> key once, you will have the same command with `./` prefixed without typing:

```bash
$ ./somescript.sh
```

It's a stripped down version of [plugins/sudo](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/sudo)

You'll find the plugin here: [tmiland/zsh-script-prefix](https://github.com/tmiland/zsh-script-prefix)

Hope someone will find it useful. 🙏‍

_Please disregard if posted in the wrong place._

carlosala:
It looks nice! Even though, closing this to prevent us to confuse this with an issue with the repo 😊 
Thanks for the contribution to the zsh community!

PR #11370, opened 2022-12-05
Plugin: zsh-script-prefix

tmiland:
## Standards checklist:

## Changes:

- [Added zsh-script-prefix plugin]

## Other comments:

...
I first shared this here: #11369
And thought I'd add this as a pr so more people can take advantage of this plugin.

Why did i create this? Because i was tired of; when using autocomplete to get a script name, i would have to manually go back and add the `./` prefix to the script name, which when working with lots of script, gets really tiresome. 

Now, all you have to do is hit [Esc] button **once**.

### DIFF
--- a/plugins/zsh-script-prefix/README.md
+++ b/plugins/zsh-script-prefix/README.md
@@ -1,0 +1,55 @@
+# zsh-script-prefix
+
+Easily prefix your current or previous commands with `./` by pressing <kbd>esc</kbd> once.
+
+_Modified version of ohmyzsh_ [plugins/sudo](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/sudo)
+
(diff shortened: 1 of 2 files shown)


mcornella:
Hi Tommy, thank you for the contribution.

When reviewing new plugins, we have to consider that there is a big number of plugins already included by default within OMZ.

In this case, I can see that it might be useful for your use case, however this might be less useful for the workflow of other people for the following reasons:
1. When writing a script filename, unless you use the prefix `./` you won't get autocompletion of the filename, which already will tell you that you're missing the prefix.
2. If you add the current directory to your `$PATH`, it is then not necessary to add the prefix. I consider that more useful, which is a single line of code in my `.zshrc` file:
   ```sh
   path+=(.)
   ```

That said, I'd like to thank you for the submission, and wish to see what you come up with next. Cheers!

PR #11373, opened 2022-12-06
refactor(kubectl): standarize completion generation

carlosala:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...


kadamanas93:
This change causes an issue where the `kubectl` aliases aren't loaded on shell startup but force loading aliases with `source ~/.zshrc` loads the aliases.

`$+commands[kubectl]` evaluates to 0 on startup but to 1 after

carlosala:
@kadamanas93 I can't reproduce it. Even though, you should never source `.zshrc`. Please make sure `kubectl` binary is in a directory in `$PATH` when sourcing `oh-my-zsh.sh`.

PR #11376, opened 2022-12-07
create poetry plugin dir if it doesn't exist

danielgafni:
Currently this script fails with errors if the `$ZSH_CACHE_DIR/completions` doesn't exist. This should fix it.

## Standards checklist:

## Changes:

- fix for the poetry plugin that removes the error when the plugin directory is not yet created

### DIFF
--- a/plugins/poetry/poetry.plugin.zsh
+++ b/plugins/poetry/poetry.plugin.zsh
@@ -11,4 +11,5 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_poetry" ]]; then
   _comps[poetry]=_poetry
 fi
 
+mkdir -p $ZSH_CACHE_DIR/completions
 poetry completions zsh >| "$ZSH_CACHE_DIR/completions/_poetry" &|


carlosala:
This is done [here](https://github.com/ohmyzsh/ohmyzsh/blob/master/oh-my-zsh.sh#L56) before the plugin load, it's not necessary!

danielgafni:
It really isn't if you are using a different plugin manager like sheldon. The plugin didn't work out of the box for this reason.

rwmitchell:
Then how is that an OMZ issue ?

danielgafni:
It surely isn't, but perhaps it would be nice to add this fix?

rwmitchell:
Because you're fixing it in the wrong place.  Anyone using this plugin via OMZ isn't going to have any issues, as Carl pointed out.  I don't see a .git subdirectory under poetry, so this particular plugin isn't being auto-pulled from somewhere else.  You need to go fix sheldon.

carlosala:
@danielgafni yeah, probably it would be great to have that, but then we'd need to replicate that in maybe 40 plugins that have this issue. Probably your best shot is to create that directory in your `zshrc` before loading the plugin(s) that require it!
Even though, thanks for your contribution! 😊

PR #11378, opened 2022-12-08
feat(aws)!: improve `aws_change_access_key`

grimm26:
Add improvements to aws_change_access_key to catch more error conditions
and give more useful feedback.
Offer to deactivate the old access key after the new one is
generated and in place.

### DIFF
--- a/plugins/aws/README.md
+++ b/plugins/aws/README.md
@@ -1,7 +1,8 @@
 # aws
 
-This plugin provides completion support for [awscli](https://docs.aws.amazon.com/cli/latest/reference/index.html)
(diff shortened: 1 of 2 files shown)


grimm26:
@maksyms can I get some eyes on this?

maksyms (approved):
I don't use this functionality, but the changes look good to me.
One thing to note: THIS BREAKS AWS CLI v1 COMPATIBILITY.

@mcornella good to merge IMHO.

grimm26:
@mcornella @maksyms any chance of this getting merged any time soon?

carlosala:
@grimm26 hi! As this change breaks the compatibility with aws cli v1, could we have that info in the readme, that this plugin only allows the usage with aws cli v2?
Thanks!

carlosala:
Thanks for your contribution!

PR #11385, opened 2022-12-11
docs(term_tab): plugin can only be used with linux and solaris

rwmitchell:
This plugin only works for OSTYPE solaris* and linux* - let users know in the readme

## Standards checklist:

- [X ] The PR doesn't replicate another PR which is already open.
- [X ] The code is mine or it's from somewhere with an MIT-compatible license.
- [X ] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- [X] Let non-solaris and non-linux users they need not bother with this.

## Other comments:

...

### DIFF
--- a/plugins/term_tab/README.md
+++ b/plugins/term_tab/README.md
@@ -1,5 +1,7 @@
 # term_tab plugin
 
+This plugin only works for Solaris and linux.
+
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #11391, opened 2022-12-15
feat(term_tab): add support for macOS

rwmitchell:
This can probably be improved with better lsof and zsh zen, but it works reasonably well on my system.

## Standards checklist:

## Changes:

- [x] use lsof, grep, and awk to generate list of CWDs of zsh sessions to support macOS that lacks /proc 

## Other comments:

...


grimm26 (changes requested):
Please follow the commit guidelines https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#commit-guidelines especially in making your commit message and PR title.  Thanks.

mcornella:
Using the man page for lsof I found the following command that will automatically select only zsh processes and print the directories on each line, which works for directories with spaces:
```
lsof -d cwd -c zsh -a -w -Fn | sed -n 's/^n//p'
```

Pending to check if it works for macOS lsof.

rwmitchell:
Your change works on my system:

darwin*)  dirs=( $( lsof -d cwd -c zsh -a -w -Fn | sed -n 's/^n//p' ) ) ;;

mcornella:
Sorry for forgetting about this, thanks for applying my suggestion!

PR #11392, opened 2022-12-15
feat(git): add `gcmbdp` and `gcdbdp` alias

mefengl:
## Standards checklist:

## Changes:

- \+ `alias gcmbdp='previous_branch=$(current_branch) && git checkout $(git_main_branch) && git pull && git branch -d $previous_branch'`
- \+ `alias gcdbdp='previous_branch=$(current_branch) && git checkout $(git_develop_branch) && git pull && git branch -d $previous_branch'`

## Other comments:

it means git checkout main/develop and then git delete previous_branch, I used it a lot, but I am not sure it should be an alias in the git plugin, but I make a pull request anyway

there is a git pull before deleting the previous branch because if you are still staying in a feature branch that just get merged, you probably don't have the newest main branch that has the previous branch merged, then the branch delete will fail


mefengl:
like in Github, I can delete the branch remotely after being merged, the local branch is useless, but I can't delete it right away, I have to first check out to another branch and then delete the previous branch.
I use this alias to simplify the above steps

PR #11398, opened 2022-12-18
feat(archlinux): optimize `paclist` and `pacdisowned`

WH-2099:
## Standards checklist:

### 1. Optimize the implementation of `pacdisowned`
- Use `mktemp` to create the temporary directory.
In the Archlinux environment, **`mktemp` is part of the base installation**, so there is no need to worry about its existence.

- Remove search path `/bin`, `/sbin` and `/lib`
In the Archlinux environment, `/bin`, `/sbin`, `/lib` and `/lib64` (which even not mentioned in the current version) are just **soft links** to the folders under /usr.
```bash
$ ls -ld /bin /sbin /lib /lib64
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /bin -> usr/bin
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /lib -> usr/lib
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /lib64 -> usr/lib
lrwxrwxrwx 1 root root 7 Dec  7 10:41 /sbin -> usr/bin
```

- The command string passed to the `trap` command should use **double quotes** to ensure that parameter expansion inside is done immediately (`tmp_dir` is declared as `local`).

- Temporary files should be cleaned up **as soon as possible** after use, instead of waiting for the `trap` to take effect.

### 2. Use `pacman` rather than `expac` in `paclist`
The **`expac`** command is a community repository, **usually not installed** by users.
*Using only  `pacman`, the output information is more useful and more acceptable to native users.*

## Other comments:
This PR is rebuilt from #10919 .

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -28,7 +28,7 @@ alias upgrade='sudo pacman -Syu'
 function paclist() {
   local pkgs=$(LC_ALL=C pacman -Qqe)
   for pkg in ${(f)pkgs}; do
-      pacman -Qs --color=auto "^${pkg}\$"
+      pacman -Qs --color=auto "^${pkg}\$" || break
   done
 }
 
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -33,20 +33,21 @@ function paclist() {
 }
 
 function pacdisowned() {
-  local tmp db fs
-  tmp=${TMPDIR-/tmp}/pacman-disowned-$UID-$$
-  db=$tmp/db
-  fs=$tmp/fs
+  local tmp_dir db fs
+  tmp_dir=$(mktemp --directory)
+  db=$tmp_dir/db
+  fs=$tmp_dir/fs
 
-  mkdir "$tmp"
-  trap 'rm -rf "$tmp"' EXIT
+  trap "rm -rf $tmp_dir" EXIT
 
   pacman -Qlq | sort -u > "$db"
 
-  find /bin /etc /lib /sbin /usr ! -name lost+found \
+  find /etc /usr ! -name lost+found \
     \( -type d -printf '%p/\n' -o -print \) | sort > "$fs"
 
   comm -23 "$fs" "$db"
+
+  rm -rf tmp_dir
 }
 
 alias pacmanallkeys='sudo pacman-key --refresh-keys'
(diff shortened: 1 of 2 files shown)


plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -26,27 +26,28 @@ alias pacupd="sudo pacman -Sy"
...
 
   pacman -Qlq | sort -u > "$db"
 
-  find /bin /etc /lib /sbin /usr ! -name lost+found \
+  find /etc /usr ! -name lost+found \
     \( -type d -printf '%p/\n' -o -print \) | sort > "$fs"
 
   comm -23 "$fs" "$db"
+
+  rm -rf tmp_dir
```
ratijas, line 50:
```suggestion
  rm -rf $tmp_dir
```

plugins/archlinux/archlinux.plugin.zsh
```diff
@@ -26,27 +26,28 @@ alias pacupd="sudo pacman -Sy"
 alias upgrade='sudo pacman -Syu'
 
 function paclist() {
-  # Based on https://bbs.archlinux.org/viewtopic.php?id=93683
-  pacman -Qqe | \
-    xargs -I '{}' \
-      expac "${bold_color}% 20n ${fg_no_bold[white]}%d${reset_color}" '{}'
+  local pkgs=$(LC_ALL=C pacman -Qqe)
+  for pkg in ${(f)pkgs}; do
+      pacman -Qs --color=auto "^${pkg}\$"
```
ratijas, line 31:
```suggestion
      pacman -Qs --color=auto "^${pkg}\$" || break
```

This would support interrupting the loop (Ctrl+C)

ratijas (approved):
Seems like I was already happy with the previous version of this PR (which got deleted and re-created for technical reasons). There is one little typo in `rm -rf`, but other than that it's fine. I don't think `LC_ALL=C` changes anything for `pacman -Qqe` combination of flags, except maybe if there are any error messages on stderr, in which case we probably want to present them in user-readable language anyway?

Also I have to note that paclist doesn't make much sense to me, especially the time it takes to print everything: the previous one (with expac) runs for 6s on my machine, your new one takes about 33 seconds, which is more than 5x loss. Of course, ideally we would stuff all those packages at once in one command invocation, but that's doesn't seem to be possible with the current pacman CLI.

Overall, fix those two suggestions, otherwise LGTM!

WH-2099:
Thank you for your careful inspection. I didn't find that there was such a detailed error. This submission should be re-initiated by me through the web page before, so it is not detailed enough. I will pay more attention to this aspect in the future.
`LC_ALL=C` is indeed for readability of messages on stderr, it is not required.
I'm following a performance issue with `paclist` and I'll be tinkering a bit more this weekend.

carlosala:
Thanks for your contribution to both!

PR #11400, opened 2022-12-19
fix(init): only check for upgrade if shell is interactive

### LINKED ISSUE #11390: auto-update interferes with VSCode shell environment detection
### Describe the bug

Visual Studio Code [launches a shell to detect your environment](https://code.visualstudio.com/docs/supporting/faq#_resolving-shell-environment-fails).  However, Oh My Zsh will occasionally block shell startup with the prompt `[oh-my-zsh] Would you like to update? [Y/n]` when enough time has passed since the last update.

I've seen this problem also occur when other GUIs run an embedded shell too.

### Steps to reproduce

1. Install Oh My Zsh with auto-update enabled
2. Wait
3. Start VS Code (or any other GUI that runs an embedded shell)

### Expected behavior

It would be nice if Oh My Zsh did not block for an auto-update confirmation unless the user was actually capable of answering it (i.e. when running from Terminal.app).

_No response_

### OS / Linux distribution

macOS 12

### Zsh version

5.8.1

### Oh My Zsh version

master (29cf199)

### Terminal emulator

Terminal.app

None

### Additional context

_No response_

carlosala:
Fixes #11390

## Standards checklist:

## Changes:

We are only using `check_for_upgrade.sh` on init file. I'd disable whole omz upgrade (no matter which mode is enabled by the user) to prevent auto updates that might introduce breaking changes and user would not notice about that.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "printWidth": 110,
+  "proseWrap": "always"
+}
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -57,9 +57,7 @@ mkdir -p "$ZSH_CACHE_DIR/completions"
 (( ${fpath[(Ie)"$ZSH_CACHE_DIR/completions"]} )) || fpath=("$ZSH_CACHE_DIR/completions" $fpath)
 
 # Check for updates on initial load...
-if [[ "$DISABLE_AUTO_UPDATE" != true ]]; then
-  source "$ZSH/tools/check_for_upgrade.sh"
-fi
+source "$ZSH/tools/check_for_upgrade.sh"
 
 # Initializes Oh My Zsh
 
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -21,9 +21,11 @@ zstyle -s ':omz:update' mode update_mode || {
 # Cancel update if:
 # - the automatic update is disabled.
 # - the current user doesn't have write permissions nor owns the $ZSH directory.
+# - is not run from a tty
 # - git is unavailable on the system.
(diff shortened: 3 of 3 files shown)


mcornella:
We're keeping the skip logic in `check_for_upgrade.sh` now, in fact this check should be removed I believe but I don't remeber why I kept it.

carlosala:
Actually true, we could just leave the interactive one and would be the same.

mcornella:
Wait, now that I remember it, zshrc is only read when interactive https://zsh.sourceforge.io/Guide/zshguide02.html#l9. We should troubleshoot why VS Code reads zshrc first.

carlosala:
That makes total sense. Maybe vscode open a interactive terminal for some reason, dunno. We should check

carlosala:
Let's convert to draft as the issue comes by other place 👌🏻

carlosala:
@mcornella this should be enough

carlosala:
remember to rebase instead of squash, there are two different commits

mcornella (changes requested):
Re: the init script check, this was added back due to https://github.com/ohmyzsh/ohmyzsh/issues/9039, but I can see that the Archlinux package no longer removes `tools/`, so it's fine to remove it: https://github.com/archlinuxcn/repo/blob/d95e3048281c615be37722deb270c6165a64e927/archlinuxcn/oh-my-zsh-git/PKGBUILD.

With regard to the check in `check_for_upgrade.sh`, we can put it right at the start of the file, lines 25 to 30.

Also let's check for `[[ ! -t 1 ]]` in order to cancel the update and have @alexchandel test the PR before using this specific VS Code check.

carlosala:
Nice points Marc, already fixed 👌🏻
Let's wait for Alex test 👌🏻

PR #11402, opened 2022-12-20
fix(doctl): standardize plugin completion

andrewsomething:
## Standards checklist:

## Changes:

- fix(doctl): add compdef line

## Other comments:

Reported to upstream doctl at: https://github.com/digitalocean/doctl/issues/1329

### DIFF
--- a/plugins/doctl/doctl.plugin.zsh
+++ b/plugins/doctl/doctl.plugin.zsh
@@ -4,7 +4,14 @@
 #
 # Author: https://github.com/HalisCz
 
-if [ $commands[doctl] ]; then
-  source <(doctl completion zsh)
-  compdef _doctl doctl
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution @andrewsomething! Maybe would be better to standarize plugin completion to what we are going in other plugins like [gh](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/gh/gh.plugin.zsh) one. Could you please handle that?
Thanks!

andrewsomething:
@carlosala Sounds good! Updated.

mcornella (approved):

mcornella:
Thank you Andrew! Really appreciate Digital Ocean's involvement in this :✨

PR #11403, opened 2022-12-22
Add `gdn` alias for `git diff --name-only`

gonzonieto:
## Standards checklist:

## Changes:

- Add `gdn` alias to git plugin for `git diff --name-only`
- Common use case is to quickly list all files that have changed against whatever branch is given as argument
- Doesn't conflict with any existing aliases as far as I can tell
- I'm not seeing any other issues/PRs that address this suggestion


robbyrussell:
Superseded by #13163, thanks for the idea!

PR #11415, opened 2022-12-29
add bd plugin

MrTpat:
## Standards checklist:

## Changes:

- Adds built-in plugin access to the BD plugin found here: https://github.com/Tarrasch/zsh-bd


mcornella:
We already have a lot of plugins that move around directories. Instead, you can set it up as a custom plugin, see [adding a new plugin](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#adding-a-new-plugin).

PR #11422, opened 2023-01-02
fix(autojump): fix autojump sourcing in nix (-darwin)

enniomara:
## Standards checklist:

## Changes:

This commit fixes an apparent bug when sourcing autojump when it is installed through nix-darwin in MacOS. Instead of sourcing a shell script (like the other entries), the changes line tries to source the binary (which is a python script), which fails as expected with the following message.
```
...
/etc/profiles/per-user/username/bin/autojump:21: command not found: from
...
```

I am not sure if/how the old PR (#11291) worked, as it should have failed then aswell. 

Let me know if you need anything changed.


mcornella:
Great PR, thanks!

PR #11432, opened 2023-01-06
Update CODEOWNERS to remove maksyms

maksyms:
@mcornella I'm afraid I no longer use `ohmyzsh` or even `zsh` in favour of `fish`. Hence, it doesn't make sense for me to continue being the code owner for the plugin. Good luck with supporting a fantastic tool!


mcornella:
Thank you for your hard work during this time!

PR #11438, opened 2023-01-12
Podman module

guenhter:
## Standards checklist:

## Changes:

Add a module for podman aligned to the docker commands.

There is a similar MR https://github.com/ohmyzsh/ohmyzsh/pull/9275 but that one doesn't provide any short commands.


carlosala:
Hi @guenhter, thanks for your contribution! As far as I know for `podman`, they simply recommend you to add `alias docker=podman` to your configuration and go on.
Maybe you could just use docker plugin adding that alias, what do you think?

guenhter:
Hi, 

you are in general right, but when you use docker and podman (as I unfortunately have to do), then also the commands have to be used via `podman` cli

laundry-96:
I unfortunately have to agree with @guenhter. I am looking for podman plugin since I often use it with toolbox at work, but most of our servers run docker so I have to use both.

One idea is that we rename "docker" to "container" and have aliases for both inside that plugin

tjarbo (commented):
@guenhter thank u for the PR! Expecially since #9275 was closed. I did not had the chance to review the aliases in detail yet, but I already have a first question: Your readme mentiones "adds auto-completion" and I see your link to https://github.com/containers/podman/blob/main/completions/zsh/_podman, but how is this autocompletion applied now? I was expecting some loading mechanism as it is in place for the argocd plugin (https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/argocd/argocd.plugin.zsh)

@mrcomoraes shared a nice workaround and example command how to generate the autocompletion with podman cli: https://github.com/ohmyzsh/ohmyzsh/issues/7489#issuecomment-814339789

guenhter:
@tjarbo :) I have no idea, but I'm not actively using this any more so if someone could have a look, I'd be glad, or I simply close this one.

tjarbo:
Hi @guenhter, no worries - I can take over this topic. I will create my own fork and replicate your changes and rise a new pull request.

carlosala:
closing this for the moment then :)

PR #11439, opened 2023-01-12
git: alphabetically sort documentation

jdevries3133:
@carlosala alphabetically sorted readme, as requested in https://github.com/ohmyzsh/ohmyzsh/pull/11259#issuecomment-1380138811


carlosala:
I was targeting something like #9743, sorting by command both the script and the readme. Could you go for it? Thanks! 😊

jdevries3133:
So, I tried. It was quite tricky to get the source code sorted alphabetically by command as I did in the readme. I'm trying to do this with vim-fu because doing this by hand is outside the oberton window of tedium for me, personally.

With the readme it was easy, because I could visual-select to reorder the whole columns in the table, then `:sort`, then flip the columns back into the right order.

I tried to do a similar thing with the source, but there are functions in there, there are distinct sections related to certain functionality where the ideal sort order is unclear, etc.

As an alternative, I'd suggest sorting each chunk of aliases in the source alphabetically, rather than try to have the source order match the documentation order exactly 1:1. What do you think?

carlosala:
hmm, I think we could sort everything following the same structure. It should be a simple script sorting all aliases by command, and then moving the functions manually to the desired place.

jdevries3133:
The source is somewhat grouped around topics now as far as I can tell. In particular, the aliases that use functions generally follow the function definition. 


What do you think the ideal sort order is?

carlosala:
In my opinion, the best would be grouping by command (so fetch, rebase, pull, clone...) and then, inside, grouping alphabetically by name of alias. What do you think?

jdevries3133:
@carlosala I added a draft commit. It's not 100% sorted, but it's a rough rearrangement into the order that makes sense to me:

1. All your functions and `compdef`s pretty much ordered
2. Aliases, in alphabetical order

Again, this is just a quick first pass to show my ideas in spirit – there are still quite a few things that are "out of order."

carlosala:
I kinda like this solution! I think we could move the functions to the end of each respective command block, same for aliases with `is-at-least`. Could you also update the branch to master? We added some things to git plugin in the meantime :)

ferdnyc:
Nothing against @carlosala or anyone else involved here, but IMHO it was a miscalculation to expand this PR into sorting of the documentation _and_ code. They're totally separate things, and it's not like either affects the other. The original documentation-sorting commit could've been merged forever ago, and the source sorting addressed in a different PR (either by @jdevries3133 or someone else). Instead, we still have this PR unmerged (and probably unmergeable, given the conflicts).

jmschp:
Hey

I agree with @ferdnyc.

I just updated my PR, and would love it to be merged. 🙏

PR #11444, opened 2023-01-17
Add the gfom alias to the git plugin

mbillard:
## Standards checklist:

## Changes:

- Adds the `gfom` alias to the git plugin

The alias is for `git fetch origin ${git_main_branch}` and follows a similar pattern to grbom and gmom.

It is useful when working on a branch and you want to update master before rebasing or merging without having to switch branches.

## Other comments:

...

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -76,6 +76,7 @@ plugins=(... git)
 | gfa                  | git fetch --all --prune                                                                                                                                                                  |
 | gfg                  | git ls-files \| grep                                                                                                                                                                     |
 | gfo                  | git fetch origin                                                                                                                                                                         |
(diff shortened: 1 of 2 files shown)


mcornella:
This doesn't make sense for most use cases. You could just use a `git fetch` alias, I feel that this is too specific to be useful and we already have many aliases #5721.

topizio:
@mcornella , I was about to open a similar PR. I think it it reasonable to have it as it makes the life much easier when you want to rebase to main/master

mcornella:
You can use `gfa` which is what I do, or `gfo` which does the same thing only for the `origin` remote. If you specifically want only one branch to be fetched (I presume, if you're on a repo with many branches), you can always autocomplete the branch name.

topizio:
@mcornella , thanks, that's not bad at all! Cheers

PR #11446, opened 2023-01-20
feat(functions/take): make `.tgz` behave as `.tar.gz`

jmigual:
## Standards checklist:

## Changes:

- The `take` now also accepts files ending with the extension `.tgz`.

## Other comments:

The `take` command only accepts files ending with the extensions `.tar.gz`, `.tar.bz2`, or `.tar.xz`. This also allows also for files ending in `.tgz`. This extension is quite used. An example of this usage is the python source code files (https://www.python.org/ftp/python/3.11.1/Python-3.11.1.tgz)

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -56,7 +56,7 @@ function takegit() {
 }
 
 function take() {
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution, great improvement! 😊

PR #11451, opened 2023-01-23
feat(mlh): add separate prompt symbol for root user

bartekpacia:
The `mlh` theme lacks this config option.

## Standards checklist:

## Changes:

- if `MLH_SHELL_SYMBOL_ROOT` env var is defined, `mlh` theme will use it. If it's not defined, it defaults to `# `.

  `MLH_SHELL_SYMBOL_ROOT` is used when `$UID` is `0` (root user).

### DIFF
--- a/themes/mlh.zsh-theme
+++ b/themes/mlh.zsh-theme
@@ -47,6 +47,10 @@ if [ -z "$MLH_SHELL_SYMBOL" ]; then
   MLH_SHELL_SYMBOL="$ "
 fi
 
+if [ -z "$MLH_SHELL_SYMBOL_ROOT" ]; then
+  MLH_SHELL_SYMBOL_ROOT="# "
+fi
+
 # colors
 USER_COLOR="%F{001}"
 DEVICE_COLOR="%F{033}"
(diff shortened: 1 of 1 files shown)


themes/mlh.zsh-theme
```diff
@@ -83,7 +87,11 @@ exit_code() {
 }
 
 prompt_end() {
-  printf "\n$MLH_SHELL_SYMBOL"
+  if [ "$UID" -eq "0" ]; then
```
carlosala, line 90: 0 doesn't need to be quoted

themes/mlh.zsh-theme
```diff
@@ -47,6 +47,10 @@ if [ -z "$MLH_SHELL_SYMBOL" ]; then
   MLH_SHELL_SYMBOL="$ "
 fi
 
+if [ -z "$MLH_SHELL_SYMBOL_ROOT" ]; then
+  MLH_SHELL_SYMBOL_ROOT="# "
```
carlosala, line 51: just to not be breaking, I'll make the default to also be `$`

carlosala (changes requested):

themes/mlh.zsh-theme
bartekpacia: It's not breaking - it's a fix. The Unix convention is to use `#` for superuser prompt. Please reconsider :)

themes/mlh.zsh-theme
bartekpacia: DONE

themes/mlh.zsh-theme
carlosala: it's a breaking fix actually, but let's go for it

carlosala:

PR #11460, opened 2023-01-27
fix(fzf): check for `apt` and `apt-get` in debian setup function

### LINKED ISSUE #11459: fzf: change check for dpkg for Debian-based distro function
The function also supports termux which uses `pkg`, but I see I have both `apt` and `apt-get` installed.

Some Debian-based systems have one or the other, so we should check for either of them:

```sh
(( ${+commands[apt]} || ${+commands[apt-get]} )) 
```

_Originally posted by @mcornella in https://github.com/ohmyzsh/ohmyzsh/pull/11122#issuecomment-1405596823_

mcornella:
## Standards checklist:

## Changes:

- Checks for `apt` and `apt-get` instead of `dpkg`, which might be installed in other non-Debian-based systems such as Arch Linux and Fedora. See https://github.com/ohmyzsh/ohmyzsh/pull/11122#issuecomment-1399607430

Fixes #11459

### DIFF
--- a/plugins/fzf/fzf.plugin.zsh
+++ b/plugins/fzf/fzf.plugin.zsh
@@ -81,7 +81,7 @@ function fzf_setup_using_debian() {
       key_bindings="${PREFIX}/share/fzf/key-bindings.zsh"
       ;;
     *)
-      if [[ ! -f /usr/bin/fzf ]]; then
+      if [[ ! -d /usr/share/doc/fzf/examples ]]; then
         # fzf not installed
         return 1
       fi


carlosala:
You could also update readme, there's the skip dpkg documentation.

ppwwyyxx (commented):
There is no harm to check `/usr/share/doc/fzf/examples` exists before sourcing files from it, right? Why not?

Given the issue with dpkg I won't be surprised to see people having a command called "apt" on their non-debian machine.

With a quick search, archlinux's `jdk7-openjdk` package has a file like this:
```
extra/jdk7-openjdk 7.u261_2.6.22-1	/usr/lib/jvm/java-7-openjdk/bin/apt
```
though I have no idea what it is.

mcornella:
I think that's unlikely, but it doesn't hurt to check for the mentioned directory as it is hardcoded in the function already. If it changes we'll have to change the function either way.

carlosala (approved):
lgtm

PR #11463, opened 2023-01-27
Add ggpush! to git.plugin.zsh

kubakostrzewski:
## Standards checklist:

- [x ] The PR title is descriptive.

## Changes:

- [Added new git.plugin alias ggpush! which push current branch with force flag]

## Other comments:

...


carlosala:
We are at this moment not adding any `--force` alias to `git` plugin. In this case, you could directly do `ggpush -f` and you'd get the same output, is not much shorter than `ggpush!`.
If you need this alias, it's better to add it to your `zshrc`. Thanks for your contribution!

PR #11469, opened 2023-01-31
feat(lib/directories): Add an escape hatch to the file

hazaelsan:
These aliases don't seem particularly high-value, and can interfere with other workflows given the global nature of some of the aliases.

This change makes it possible to bypass these aliases by setting `OMZ_DIRECTORY_ALIASES=false` before sourcing `oh-my-zsh.sh`. The default behavior is not changed to not break existing workflows.

## Standards checklist:

## Changes:

- Add `OMZ_DIRECTORY_ALIASES` environment variable

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -1,3 +1,5 @@
+[[ "${OMZ_DIRECTORY_ALIASES:-true}" == "false" ]] && return
+
 # Changing/making/removing directory
 setopt auto_pushd
 setopt pushd_ignore_dups


carlosala:
It's not a bad idea! Even though, could you refactor it to make it use of `zstyle`? And move the check below the `setopts`, that seems to be sensible for every user.

hazaelsan:
Done, thanks.

lib/directories.zsh
```diff
@@ -1,10 +1,11 @@
-[[ "${OMZ_DIRECTORY_ALIASES:-true}" == "false" ]] && return
-
 # Changing/making/removing directory
 setopt auto_pushd
 setopt pushd_ignore_dups
 setopt pushdminus
 
+zstyle -s ':omz:directories' aliases load_aliases || load_aliases=true
```
carlosala, line 6: Let's use the following line:
```zsh
zstyle -T ':omz:directories' aliases || return
```
It'll be configured in `zshrc` as
```zsh
zstyle ':omz:directories' aliases no
```

carlosala (changes requested):
We also need to document that somewhere

lib/directories.zsh
hazaelsan: Change done.  Should the doc changes go in the FAQ?

PR #11470, opened 2023-02-01
feat(dotnet): add `dwt` alias

stoyandimov:
## Standards checklist:

## Changes:

* Adds `dotnet watch test` alias (dwt).

## Other comments:

`dotnet watch test` to test projects is what `dotnet watch run` is to applications that need to be watched.
Since `dwr` is available, I believe that `dwt` would be a great addition for test-driven development.


carlosala:
Thanks for your contribution!

Issue #11388, opened 2022-12-14
plugins/git: Safer `git push --force`


larvanitis:
### If the feature request is for a plugin or theme, specify it here.

Git plugin

### If the feature solves a problem you have, specify it here.

The `git push --force-with-lease` command can become safer by adding the new (May 2022) `--force-if-includes` flag.

### Describe the proposed feature.

Change `alias gpf='git push --force-with-lease'` to `alias gpf='git push --force-with-lease --force-if-includes'` in
https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/git/git.plugin.zsh#L238

### Additional context

https://stackoverflow.com/questions/65837109/when-should-i-use-git-push-force-if-includes

mcornella:
This improves the case when a separate system automatically fetches all commits in a repository (as VS Code does on the background for example).

We are already getting the `git --version` in the plugin and checking it for flags, see https://github.com/ohmyzsh/ohmyzsh/blob/0e5f27481d9c4400b0d069ed884238e1d08e215c/plugins/git/git.plugin.zsh#L140-L142

It shouldn't be that difficult to do the same for git 2.30 and the `gpf` alias! Good first PR if you come across this one!

akorkos:
Hello, could I work on this issue?

mcornella:
Sure, go ahead!

UnnitMetaliya:
@mcornella 

I created a PR for this. Please review: https://github.com/ohmyzsh/ohmyzsh/pull/11481

PR #11481, opened 2023-02-05
feat(git)!: add `force-if-includes` flag

UnnitMetaliya:
## Standards checklist:

## Changes:

- As per the issue #11388 adds a new safer flag in git plugin `--force-if-includes`

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -89,7 +89,8 @@ plugins=(... git)
 | ggsup                | git branch --set-upstream-to=origin/$(git_current_branch)                                                                                                                                |
 | ggu                  | git pull --rebase origin $(current_branch)                                                                                                                                               |
 | gpsup                | git push --set-upstream origin $(git_current_branch)                                                                                                                                     |
-| gpsupf                | git push --set-upstream origin $(git_current_branch) --force-with-lease                                                                                                                                     |
+| gpsupf               | git push --set-upstream origin $(git_current_branch) --force-with-lease --force-if-includes (git version >= 2.30)                                                                        |
+| gpsupf               | git push --set-upstream origin $(git_current_branch) --force-with-lease (git version < 2.30)                                                                                             |
 | ghh                  | git help                                                                                                                                                                                 |
 | gignore              | git update-index --assume-unchanged                                                                                                                                                      |
 | gignored             | git ls-files -v \| grep "^[[:lower:]]"                                                                                                                                                   |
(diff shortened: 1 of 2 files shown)


carlosala:
Thanks for your contribution!

plugins/git/README.md
```diff
@@ -120,6 +121,8 @@ plugins=(... git)
 | gp                   | git push                                                                                                                                                                                 |
 | gpd                  | git push --dry-run                                                                                                                                                                       |
 | gpf                  | git push --force-with-lease                                                                                                                                                              |
```
tom-reinders, line 123: Should have replaced this line, not added to

carlosala: Fixed, thanks for the review!

aekt:
@UnnitMetaliya @carlosala thanks for contributions, but seems `--force-if-includes` having a different behaviour than the original `gpf`, can we consider keeping both? Thanks

carlosala:
@aekt yes, we decided on purpose to make the check stricter, it's flagged as a breaking change. What are you missing from previous behaviour?

aekt:
Thanks for quick reply @carlosala, probably because I don't a lot of branches locally...

I'm agree with the purpose, but do you think we can introduce a new one that is less strict? Say `gpfl` or `gpfrc`?

carlosala:
let's go for `gpfl`!

mcornella:
Can you describe the behavior it has on your case and why it's necessary to keep the old one?

aekt:
As a minimal example, I have a commit graph like this:

```
* e97c34b (HEAD -> main) Update amend
| * f229734 (origin/main-tmp) Update
|/  
* 7aaf892 (origin/main, origin/HEAD) Initial commit
```

Then when I `gpf origin main:main-tmp` or `gpf origin HEAD:main-tmp` (this is what I mostly do), it warns me as expected:

```
To github.com:aekt/gpf.git
 ! [rejected]        main -> main-tmp (remote ref updated since checkout)
error: failed to push some refs to 'github.com:aekt/gpf.git'
hint: Updates were rejected because the tip of the remote-tracking
hint: branch has been updated since the last checkout. You may want
hint: to integrate those changes locally (e.g., 'git pull ...')
hint: before forcing an update.
```

rwmitchell:
Everyone is certainly capable of creating their own aliases to make commands they use often, easier.

Is this a common workflow shared by most people?

mcornella:
That will be solved by `--force` which bypasses all checks and will do whatever you ask it. This is already in the alias `gpf!`.

rwmitchell:
How does adding an exclamation actually help?  If I mistype and accidentally hit ```gcn[TAB]``` the exclamation gets automatically added and if the tab is assumed to have done the right thing because it completely expanded something, it gets done.

Is there a way to block an exclamation from ever being expanded and required to be appended manually?

PR #11485, opened 2023-02-08
feat(gradle): update completion from upstream

bartekpacia:
## Standards checklist:

## Changes:

- I updates the plugin code from upstream at [gradle/gradle-completion](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/gradle/_gradle).

PS What's the recommended practice for keeping `ohymzsh`'s plugin in sync with upstreams?

### DIFF
--- a/plugins/gradle/_gradle
+++ b/plugins/gradle/_gradle
@@ -1,4 +1,3 @@
-#compdef gradle gradlew gw
 #
 # Taken from https://github.com/gradle/gradle-completion
 # Copyright (c) 2017 Eric Wendelin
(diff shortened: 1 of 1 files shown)


carlosala:
We do not have any particular practice for keeping updated with upstream. Normally there's a PR from a contributor. We prefer this way than automatically to prevent replicating bugs in omz. That happened already, for example, with Docker completion, that was adding a `*` to each command, what we fixed here.

bartekpacia:
Makes sense :)

Thank you for your work as `ohmyzsh` maintainer!

carlosala:
@bartekpacia thanks, it's really appreciated 😊

PR #11487, opened 2023-02-11
Remove aliases for non-standard commands

rwmitchell:
These aliases were for non-standard (installed by default) unix/linux commands and should not be included here.   They should be put into their appropriate plugin to be optionally added by users using those commands.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/lib/correction.zsh
+++ b/lib/correction.zsh
@@ -1,13 +1,8 @@
 if [[ "$ENABLE_CORRECTION" == "true" ]]; then
   alias cp='nocorrect cp'
-  alias ebuild='nocorrect ebuild'
-  alias gist='nocorrect gist'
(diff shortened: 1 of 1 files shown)


carlosala:
In this case it would be better to check if that command exists in path and then set the alias. Removing the alias now would be unnecessary, what do you think about?

rwmitchell:
Aliases should not be created in lib for utilities that are not routinely included in a default OS install.  If everything is open game, even if it is preceded by a test to see if the app is installed, where does it end?
Each test takes a bit of time, not universally necessary.  OMZ already supports plugins that allows people to optionally enable settings they want, why not use it?

carlosala:
Thanks for your contribution!

PR #11495, opened 2023-02-14
fix(af-magic): fix python venv checking logic

shelfofclub:
In some cases, such as command decorations in vscode's integrated terminal, PS1 doesn't start with `(` when venv is displayed. Therefore, it should check whether `($python_env)` exits in PS1 somewhere.

## Standards checklist:

## Changes:

- In some cases, such as command decorations in vscode's integrated terminal, PS1 doesn't start with `(` when venv is displayed. Therefore, it should check whether `($python_env)` exits in PS1 somewhere. `\(*`->`*\(${python_env}\)*`


carlosala:
Thanks for your contribution!

PR #11502, opened 2023-02-17
fix(gcloud): add location of latest snap install

tonio-ramirez:
Latest snap installs gcloud at /snap/google-cloud-cli, so add that to the list of possible locations.

## Standards checklist:

## Changes:

- Adds `/snap/google-cloud-cli/current` to the list of possible locations for gcloud installation

## Other comments:

...


mcornella:
Thanks for the contribution!

PR #11509, opened 2023-02-21
feat(iterm2): add shell integration script

reegnz:
Mirror the official script and source it with the iterm2 plugin.

The integration script brings features like automatic marks on each prompt, drag and drop scp support, or automatic profile switching.

## Standards checklist:

## Changes:

- adds script from https://iterm2.com/documentation-shell-integration.html
- sources that script

### DIFF
--- a/plugins/iterm2/README.md
+++ b/plugins/iterm2/README.md
@@ -13,7 +13,7 @@ You can enable the integration with zstyle. It's important to add this line
 before the line sourcing oh-my-zsh:
 
 ```
-zstyle :omz:plugins:iterm2 shell-integration true
+zstyle :omz:plugins:iterm2 shell-integration yes
 ```
 
 ## Plugin commands
--- a/plugins/iterm2/README.md
+++ b/plugins/iterm2/README.md
@@ -1,6 +1,7 @@
 # iTerm2 plugin
 
 This plugin adds a few functions that are useful when using [iTerm2](https://www.iterm2.com/).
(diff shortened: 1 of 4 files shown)


reegnz:
@avivrosenberg

carlosala:
in terms of license, it is ok. Even though, I'd make it opt-in, using `zstyle`. You can see how to do this in `nvm` plugin, for example.

reegnz:
I'm not sure about the opt-in, I think the shell-integration is quite essential when using iterm2. If the concern is that existing users that use iterm2 plugin don't want to be surprised by having it enabled, that's fair. I don't think it's backward-incompatible change though, so it won't break anything IMHO.

So sure, I can add a zstyle if it makes it easier to get merged.

plugins/iterm2/README.md
```diff
@@ -2,11 +2,20 @@
...
 ```
 plugins=(... iterm2)
 ```
 
+Optionally, the plugin also applies the [Shell Integration Script for iTerm2](https://iterm2.com/documentation-shell-integration.html).
+You can enable the integration with zstyle. It's important to add this line 
+before the line sourcing oh-my-zsh:
+
+```
+zstyle :omz:plugins:iterm2 shell-integration true
```
carlosala, line 16: Please change `true` with `yes`.

carlosala (changes requested):

plugins/iterm2/README.md
reegnz: Fixed

carlosala:
Thanks for your contribution!

PR #11518, opened 2023-02-23
fix(af-magic): fix number of dashes caused by python_venv checking logic

WaferJay:
## Standards checklist:

## Changes:

- fix number of dashes in af-magic theme. 

## Other comments:

The latest release of `af-magic` theme seems to have a problem on my zsh, I found #11495.

https://github.com/ohmyzsh/ohmyzsh/blob/8c50d1fa4a87b159080f8f3c70294669807fc7e2/themes/af-magic.zsh-theme#L9-L17

`python_env` is the **path** of Python venv like `/path/of/venv` and `PS1` contains the **name** of Python venv folder like `(venv)`. 

So line 13 is always false, therefore dashes will exceeds one line. Besides, line 14 minus too much and  dashes cannot fill one line.

image

### DIFF
--- a/themes/af-magic.zsh-theme
+++ b/themes/af-magic.zsh-theme
@@ -6,7 +6,8 @@
 # dashed separator size
 function afmagic_dashes {
   # check either virtualenv or condaenv variables
-  local python_env="${VIRTUAL_ENV:-$CONDA_DEFAULT_ENV}"
+  local python_env_dir="${VIRTUAL_ENV:-$CONDA_DEFAULT_ENV}"
+  local python_env="${python_env_dir##*/}"
 
   # if there is a python virtual environment and it is displayed in
   # the prompt, account for it when returning the number of dashes


mcornella:
Have you confirmed this working? What python env variable is getting read here, `CONDA_DEFAULT_ENV` or `VIRTUAL_ENV`? Also what version of venv are you using?

WaferJay:
Yes, I have confirmed it.

af-magic in the latest release:

image

af-magic before #11495:

image

My patch:

image

---

The version of Python and virtualenv (venv) is below:

image

`virtualenv` is a tool for creating isolated virtual python environments. You can execute `source venv/bin/activate` to enable the isolated python environments in `venv` folder, `VIRTUAL_ENV` indicate the full path of the `venv` folder.

mcornella:
I now remember that we used similar logic on other themes. Thank you for confirming and for your contribution!

PR #11520, opened 2023-02-23
feat(directories): re-assign duplicate alias

mrodus:
## Standards checklist:

## Changes:

- Re-assign the `lsa` alias

## Other comments:

The `lsa` alias pointed to the same command as the `l` alias.

After the changes, the two aliases would work like this:

image

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -39,7 +39,7 @@ function d () {
 compdef _dirs d
 
 # List directory contents
-alias lsa='ls -lah'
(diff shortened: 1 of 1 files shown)


mcornella:
These aliases are pretty much prehistoric and it's going to be an issue changing them because we don't know who's using the aliases and even if it doesn't seem like it it is quite a big breaking change because these are pretty much ingrained in people's workflows.

Thanks for the contribution!

PR #11521, opened 2023-02-23
fix(aliases): clarify how to pass in keywords to `acs`

yuxiang-zhang:
## Standards checklist:

## Changes:

The plugin `aliases` offers a `filter` argument to filter all available aliases, but it does not work. 

For example, `acs filter a` should only list all aliases containing `a`, but this command produces no output before this PR.

### DIFF
--- a/plugins/aliases/README.md
+++ b/plugins/aliases/README.md
@@ -19,7 +19,7 @@ Requirements: Python needs to be installed.
 
 - `acs -h/--help`: print help mesage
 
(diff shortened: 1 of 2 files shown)


yuxiang-zhang:
@mcornella In this case I've introduced a bug. I need to remove `[1:]` from https://github.com/yuxiang-zhang/ohmyzsh/blob/4969c2d67b9fbbd182c86f220835f6498b754b50/plugins/aliases/cheatsheet.py#L67

mcornella:
Sorry my mistake. Fixed now in master.

PR #11525, opened 2023-02-24
feat(dnf): install build dependencies of a package

acastanedam:
Standard check lists:
The PR title is descriptive.
The PR doesn't replicate another PR which is already open.
I have read the contribution guide and followed all the instructions.
The code follows the code style guide detailed in the wiki.
The code is mine or it's from somewhere with an MIT-compatible license.
The code is efficient, to the best of my ability, and does not waste computer resources.
The code is stable and I have tested it myself, to the best of my abilities.

Changes:
The proposed alias are useful for getting the build dependencies of a
package (e.g., for development and/or for *.rpm files).  Maybe, it could be included in #9926

### DIFF
--- a/plugins/dnf/README.md
+++ b/plugins/dnf/README.md
@@ -12,18 +12,20 @@ plugins=(... dnf)
 
 ## Aliases
 
-| Alias | Command                 | Description              |
-|-------|-------------------------|--------------------------|
-| dnfl  | `dnf list`              | List packages            |
-| dnfli | `dnf list installed`    | List installed packages  |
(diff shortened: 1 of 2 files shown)


robbyrussell:
Hey @acastanedam, sorry this one sat for so long. Still happy to take `dnfbd`... the plugin has moved to a `${dnfprog}` variable since (so it works with dnf5 too), so it'd need a rebase onto that. I'd drop `dnfbi` though, the builddep-then-install combo feels like a one-person workflow. Want to update it?

Issue #10972, opened 2022-06-03
Tab AutoCompletion not show directory


huyhoang8398:
### Describe the bug

Tab to autocomplete not show all directories

### Steps to reproduce

1. Enable ohmyzsh
2. Create two directories starting with adjacent letters:
   ```sh
   mkdir ccc
   mkdir ddd
   ```
3. Type `cd` then press TAB
4. Completion suggests D

The same happens with any two directories starting with adjacent letters: `aaa` and `bbb`, `hello` and `iris`.

### Expected behavior

suggest all files and folders

### Screenshots and recordings

https://user-images.githubusercontent.com/25379882/171906844-268c48b3-9252-4f39-8fae-9c74f87dbf3b.mp4

### OS / Linux distribution

Archlinux

### Zsh version

6.9

### Terminal emulator

kitty, alacritty

huyhoang8398:
Fixed with CASE_SENSITIVE="true"

huyhoang8398:
```
-)*.orig=00;90:(-default-)*.part=00;90:(-default-)*.rej=00;90:(-default-)*.swp=00;90:(-default-)*.tmp=00;90:(-default-)*.dpkg-dist=00;90:(-default-)*.dpkg-old=00;90:(-default-)*.ucf-dist=00;90:(-default-)*.ucf-new=00;90:(-default-)*.ucf-old=00;90:(-default-)*.rpmnew=00;90:(-default-)*.rpmorig=00;90:(-default-)*.rpmsave=00;90' 
  +_main_complete:405> funcs=( ) 
  +_main_complete:406> comppostfuncs=( ) 
  +_main_complete:411> _lastcomp=( list_lines 1 all_quotes '\' nmatches 2 restore auto context command vared '' unambiguous D list_max 100 unambiguous_cursor 2 list ambiguous insert_positions 4 unambiguous_positions 1 exact '' to_end match last_prompt yes pattern_insert menu ignored 0 insert automenu-unambiguous ) 
  +_main_complete:412> _lastcomp[nmatches]=2 
  +_main_complete:413> _lastcomp[completer]=complete 
  +_main_complete:414> _lastcomp[prefix]='' 
  +_main_complete:415> _lastcomp[suffix]='' 
  +_main_complete:416> _lastcomp[iprefix]='' 
  +_main_complete:417> _lastcomp[isuffix]='' 
  +_main_complete:418> _lastcomp[qiprefix]='' 
[...]

PR #11526, opened 2023-02-25
fix(lib): fix case-insensitive completion for zsh 5.9

mozzieongit:
## Standards checklist:

## Changes:

- Use character classes instead of ranges in zstyle matcher-list pattern to circumvent off-by-one error mentioned in https://www.zsh.org/mla/workers/2022/msg00652.html.

## Other comments:

Fixes: #10972


carlosala:
@mcornella could you take a look?

mcornella (approved):
Works well.

Tested with `docker run -it ohmyzsh/zsh:5.9 zsh`:

- [x] Case insensitive works
- [x] Hyphen/underscore works

Same is working for 5.8.1.

PR #11529, opened 2023-02-26
refactor(theme-and-appearance): reorganize and clean up logic

mcornella:
## Standards checklist:

## Changes:

- Reorganize sections
- Add and clarify comments
- Clean up logic for dealing with ls color flags
- Set both $LSCOLORS and $LS_COLORS defaults
- Set zsh completion coloring before first prompt (so that `$LS_COLORS` has been set to the final value)

Closes #3743
Closes #6353
Closes #8315
Closes #11091

Note: missing from #3743 is the logic to load and reset settings made by themes. The logic is pretty straightforward (capture state before and after theme setting), and it should be done on a separate PR.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -1,66 +1,93 @@
-# ls colors
+# Sets color variable such as $fg, $bg, $color and $reset_color
 autoload -U colors && colors
 
-# Enable ls colors
-export LSCOLORS="Gxfxcxdxbxegedabagacad"
-
-# TODO organise this chaotic logic
-
-if [[ "$DISABLE_LS_COLORS" != "true" ]]; then
-  if [[ -d "$ZSH" ]]; then
-    _test_dir="$ZSH"
-  else
-    _test_dir="."
-  fi
-  # Find the option for using colors in ls, depending on the version
-  if [[ "$OSTYPE" == netbsd* ]]; then
-    # On NetBSD, test if "gls" (GNU ls) is installed (this one supports colors);
-    # otherwise, leave ls as is, because NetBSD's ls doesn't support -G
-    gls --color -d "$_test_dir" &>/dev/null && alias ls='gls --color=tty'
-  elif [[ "$OSTYPE" == openbsd* ]]; then
-    # On OpenBSD, "gls" (ls from GNU coreutils) and "colorls" (ls from base,
-    # with color and multibyte support) are available from ports.  "colorls"
-    # will be installed on purpose and can't be pulled in by installing
-    # coreutils, so prefer it to "gls".
-    gls --color -d "$_test_dir" &>/dev/null && alias ls='gls --color=tty'
-    colorls -G -d "$_test_dir" &>/dev/null && alias ls='colorls -G'
-  elif [[ "$OSTYPE" == (darwin|freebsd)* ]]; then
-    # this is a good alias, it works by default just using $LSCOLORS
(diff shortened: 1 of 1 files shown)


lib/theme-and-appearance.zsh
```diff
@@ -1,66 +1,93 @@
...
-    [[ -n "$LS_COLORS" || -f "$HOME/.dircolors" ]] && gls --color -d "$_test_dir" &>/dev/null && alias ls='gls --color=tty'
-  else
-    # For GNU ls, we use the default ls color theme. They can later be overwritten by themes.
-    if [[ -z "$LS_COLORS" ]]; then
-      (( $+commands[dircolors] )) && eval "$(dircolors -b)"
+# Use diff --color if available
+if command diff --color /dev/null{,} &>/dev/null; then
+  function diff {
+    command diff --color "$@"
+  }
```
mcornella, line 20: This logic is cleaner than the previous one of using both a function and an alias, which is made possible by the fact that we're prefixing with `command` (otherwise the function would be recursively called until crash).

The only downside is that to bypass this default setting, `unalias diff` does not work and instead it should be done with `unfunction diff`.

lib/theme-and-appearance.zsh
```diff
@@ -1,66 +1,93 @@
...
+fi
+
+
+# Don't set ls coloring if disabled
+[[ "$DISABLE_LS_COLORS" != true ]] || return 0
+
+function test-ls-args {
+  local cmd="$1"          # ls, gls, colorls, ...
+  local args="${@[2,-1]}" # arguments except the first one
+  command "$cmd" "$args" /dev/null &>/dev/null
```
mcornella, line 30: `command` was missing from the previous logic tests leading to issues such as in #5552. Also we use `/dev/null` as an argument which will always exist and is also a single file so the command won't have to list a whole directory to finish.

lib/theme-and-appearance.zsh
```diff
@@ -1,66 +1,93 @@
...
 
-# git theming default: Variables for theming the git info prompt
-ZSH_THEME_GIT_PROMPT_PREFIX="git:("         # Prefix at the very beginning of the prompt, before the branch name
-ZSH_THEME_GIT_PROMPT_SUFFIX=")"             # At the very end of the prompt
-ZSH_THEME_GIT_PROMPT_DIRTY="*"              # Text to display if the branch is dirty
-ZSH_THEME_GIT_PROMPT_CLEAN=""               # Text to display if the branch is clean
-ZSH_THEME_RUBY_PROMPT_PREFIX="("
-ZSH_THEME_RUBY_PROMPT_SUFFIX=")"
+autoload -Uz add-zsh-hook
+add-zsh-hook precmd omz_set_completion_colors
```
mcornella, line 93: Using a precmd hook here because it will make the code run after all initialization has happened and we have the end `$LS_COLORS` value. Got the idea from #11529.

lib/theme-and-appearance.zsh
carlosala: LGTM 👌🏻

lib/theme-and-appearance.zsh
carlosala: It's a good idea, 100% agree

carlosala (approved):
Really great improvement, LGTM! 🎉

carlosala:
Change the title to fit in Github UI 👌🏻

carlosala:
@mcornella do you want to handle that missing feature, or should we find another one to do it?

PR #11531, opened 2023-02-28
add support to `take' for taking remote iso

yilas:
## Standards checklist:

## Changes:

- Add the support of remote `.iso` file.


mcornella:
Hi @yilas, thank you for the contribution. 

The purpose behind the take functions is to process the resource in some way. For example, for tgz resources in a URL they are downloaded and extracted.

Your proposed function would only download the iso file, for which using curl or wget directly should be preferred.

PR #11534, opened 2023-03-02
feat(macos): update spotify plugin to upstream

sobeckley:
## Standards checklist:

## Changes:

This script is a copy/paste from https://github.com/hnarayanan/shpotify (the plugin's acknowledgements list it as such). The upstream script has been updated since this plugin was created, this PR pulls in those changes to this plugin.

## Other comments:

I noticed the version bundled in this plugin does not work with the latest version of Spotify, however the upstream does. Hopefully this will resolve that discrepancy.

I can't tell who (if anyone) is an active maintainer of this plugin so I'm not sure who to `@`, apologies for that.

### DIFF
--- a/plugins/macos/spotify
+++ b/plugins/macos/spotify
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+function spotify() {
 # Copyright (c) 2012--2023 Harish Narayanan <mail@harishnarayanan.org>
 #
 # Contains numerous helpful contributions from Jorge Colindres, Thomas
@@ -150,11 +151,11 @@ if [ $# = 0 ]; then
 else
 	if [ ! -d /Applications/Spotify.app ] && [ ! -d $HOME/Applications/Spotify.app ]; then
 		echo "The Spotify application must be installed."
-		exit 1
+		return 1
 	fi
 
     if [ $(osascript -e 'application "Spotify" is running') = "false" ]; then
-        osascript -e 'tell application "Spotify" to activate' || exit 1
+        osascript -e 'tell application "Spotify" to activate' || return 1
         sleep 2
     fi
 fi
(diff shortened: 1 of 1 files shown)


plugins/macos/spotify
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
-function spotify() {
```
carlosala, line 3: This should not be removed, we are sourcing that on main plugin file.

plugins/macos/spotify
```diff
@@ -148,11 +150,11 @@ if [ $# = 0 ]; then
 else
 	if [ ! -d /Applications/Spotify.app ] && [ ! -d $HOME/Applications/Spotify.app ]; then
 		echo "The Spotify application must be installed."
-		return 1
+		exit 1
```
carlosala, line 153: We changed on purpose all exits to return, could you revert that?

plugins/macos/spotify
```diff
@@ -307,7 +308,7 @@ while [ $# -gt 0 ]; do
 
         "quit"    ) cecho "Quitting Spotify.";
             osascript -e 'tell application "Spotify" to quit';
-            break ;;
```
carlosala, line 310: Same here, it should be break.

plugins/macos/spotify
```diff
@@ -468,11 +469,9 @@ while [ $# -gt 0 ]; do
...
             break ;;
-
         * )
             showHelp;
-            return 1 ;;
+            exit 1;
 
     esac
 done
-}
```
carlosala, line 478: This should not be removed.

carlosala (changes requested):

sobeckley:
@carlosala Apologies for the extended delay. This should be ready for another look whenever you have a moment.

carlosala:
Thanks!

PR #11550, opened 2023-03-08
feat(init): allow turning off aliases for libs and plugins

### LINKED ISSUE #9414: Add option to disable ohmyzsh's aliases

mcornella:
## Standards checklist:

----

Proof of concept work for backing up and restoring aliases on libs and plugins that have them turned off by the user via ztyle:

```
zstyle ':omz:*' aliases no
zstyle ':omz:lib:*' aliases no
zstyle ':omz:plugins:*' aliases no
zstyle ':omz:plugins:git' aliases no
```

Work in collaboration with @carlosala.

Fixes #9414


carlosala:
We can see in #10510 what does this PR closes also 👌🏻

carlosala:
Just a quick reminder here, when merging this we should soft-deprecate https://github.com/ohmyzsh/ohmyzsh/blob/f7d903f3a31567f326d0f8ec2414722d0e3b992a/lib/directories.zsh#L12 as is not following the `':omz:lib:<file>` structure and document properly also that in README (removing also https://github.com/ohmyzsh/ohmyzsh/blob/f7d903f3a31567f326d0f8ec2414722d0e3b992a/README.md?plain=1#L279)

carlosala (approved):
LGTM! 😉

carlosala:
After testing with https://github.com/romkatv/zsh-bench, I cannot see a difference while removing all aliases and leaving all of them. It seems that the way we are keeping and restoring aliases has no performance issues.
I think we can merge this, and see how the community receives it.
I'm quite happy we finally sorted it :)

mcornella:
I'm currently debugging why setting `galiases` does not remove the global aliases, don't merge it yet.

carlosala:
At least from the documentation perspective, it should work exactly the same. Even though, if it's giving issues, I only see global aliases defined [here](https://github.com/ohmyzsh/ohmyzsh/blob/d47e1d65f66f9bb2e7a96ba58797b33f0e91a623/lib/directories.zsh#L14) and [here](https://github.com/ohmyzsh/ohmyzsh/blob/d47e1d65f66f9bb2e7a96ba58797b33f0e91a623/plugins/rails/rails.plugin.zsh#L41). We could have some workaround there 👍🏻

mcornella:
OK it looks like the same happens with `aliases` and the rest of special parameters when we're doing `array=()` (empty array).

When that happens, the resulting hash table does not actually get modified, I don't know exactly why by looking at the code:

https://github.com/zsh-users/zsh/blob/c006d7609703afcfb2162c36d4f745125df45879/Src/Modules/parameter.c#L1768-L1807

I have pushed a version that uses `unalias` directly in a "smarter" way to avoid many calls, but I don't know if it is actually slower. Could you try it out with `zsh-bench` again?

carlosala:
I think the code is a bit more difficult to read. What about adding always a garbage alias, maybe `alias __omz_tmp=<whatever>`. Then we are sure the hash table is never empty (and we could use `unalias` only for that).
What do you think?

mcornella:
That looks like a hack, I have instead pushed a change that doesn't resort to weird tricks, a simple `unalias` was enough. It should not be less fast, or even faster.

mcornella:
Shipped! 🚀  Thank you for the review!

PR #11565, opened 2023-03-16
✨ Add git checkout previous alias (gcl)

### LINKED ISSUE #11564: Git checkout previous alias

Git plugin

Being able to easily go back and forth between branches. We have `gcm` to easily go to main, but to make it easier to go back to a feature branch it would be nice to have a shortcut for `git checkout -`. This command goes to the previous checked out branch.

Add a new git alias like `gcw` that maps to `git checkout -`. 

If there is a better shortcut that's fine. I thought `gcp` would make sense but that's already taken so `gcw` (git checkout working) makes sense to me.

### Additional context

_No response_

### Related Issues

_No response_

flippidippi:
## Standards checklist:

## Changes:
Fixes #11564 
- Add git plugin alias `gcw` to run `git checkout -` (checks out the previous branch "working") 

## Other comments:
- I understand that we can currently do `gco -`, but I've talked to a few developers and most of them don't know the command. I think it would be nice to have it built in for discoverability and a little quicker to type. 
- If there is something better than `gcw` I'm open for suggestions


rwmitchell:
How does adding another alias solve a reading problem?  The existence of an alias doesn’t imply sudden knowledge of what it does.

As we’ve recently seen with the ‘aliases’ plugin, people would rather live in ignorance than learn. More aliases just eventually mean more conflicts.

carlosala:
It just skips a few types compared to `gco -`, seems not that change! Even though, seems a good alias to have in `zshrc`.
Thanks for your contribution!

flippidippi:
> As we’ve recently seen with the ‘aliases’ plugin, people would rather live in ignorance than learn. More aliases just eventually mean more conflicts.

Yeah I didn't mention it but I feel like adding a column for what each of the commands do in the README would help this.

PR #11577, opened 2023-03-23
feat(gswo): alias to return to old branch

rejmann:
## Standards checklist:

## Changes:

- [x] Added a new plugin with aliases and functions to switch with the last accessed branch.

## Other comments:

alias-to-return-to-old-branch

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -53,6 +53,19 @@ function git_develop_branch() {
   echo develop
 }
 
+function git_old_branch() {
+  command git rev-parse --git-dir &>/dev/null || return
+  local branches=($(git branch | cut -c 3-))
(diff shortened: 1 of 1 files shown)


victordev13 (approved):

gilbert-oliveira (approved):

esemeniuc:
What is the old branch? Is it like `git checkout -`?

rejmann:
Yes, it's like giving a git checkout on the last branch accessed, that is, the one before the current one!
Example: I'm on the master branch and I went to develop, instead of using git checkout master or git switch master, with the alias gswo we can just run the same and the script will understand which was the old branch accessed and give a git switch /checkout master/main.
If there is only 1 branch it will go straight to the default one

rejmann:
I did not know this command, I believe that this PR does not make more sense. Thanks for the feedback!

PR #11578, opened 2023-03-24
feat(changelog): provide links in changelog

### LINKED ISSUE #11545: Changelog improvements

_No response_

OMZ helpfully nudges me to update, and when I do, it lists a lot of things that were changed but it's a chore to get more details.

For instance, the latest update contained this line:

```
- 041c35f [amuse]                Add virtualenv support (#8987)
```

Sounds cool, but how does it work? It links the issue #, but now I have to open a browser, navigate to the project, then nav to its issues, and then lastly nav to #8987. Is it hard? No. Is it time-consuming? When there's 25 issues mentioned (like today), it can really add up!

That's part of the friction. The other part is updates like this:

```
 - cc73a92 [git]                  Add `gpod` alias
```

I understand that not all commits have an issue # to tie them to, but there's a git hash -- I can look at the code changes, but that requires roughly same amount of steps as above.

Add clickable links for each git hash & issue #. I'm making an assumption that most (if not all) of these refs are to the main project; if that's true, then this should be pretty straightforward. If not, then maybe alternatives would need to be considered.

>  - [041c35f](https://github.com/ohmyzsh/ohmyzsh/commit/041c35f) [amuse]                Add virtualenv support ([#8987](https://github.com/ohmyzsh/ohmyzsh/pull/8987))
>  - [cc73a92](https://github.com/ohmyzsh/ohmyzsh/commit/cc73a92) [git]                  Add `gpod` alias

An alternative to providing this in the terminal would be to publish such and link it from the terminal, but I think having the details available in your terminal would be the easiest to work with.

### Additional context

I apologize if this is rehashing old decisions, I searched the issues (and online) but it seems the only way I can find details on the updated items is to do a ton of manual navigating myself. This seems like such an obvious omission that I can't but think it's intentional for some reason?

### Related Issues

_No response_

GuySartorelli:
For terminals that support it, links are now provided for commit hash and issue references.

## Standards checklist:

## Changes:

- For terminals which support it, references to issues/prs and commit hashes are now hyperlinks.

## Other comments:

I started this work before #11545 was opened - there's a comment in there that implies that a change like this might already be in the works but I figured I may as well finish it up and offer it anyway.

This is my first contribution here, so it's possible I've missed some step - please let me know if there's anything I can do to improve both in terms of the process and code.

~~The logic for the condition used to set `supports_uris` is [taken from symfony/console](https://github.com/symfony/console/blob/75d4749d9620a8fa21a2d2847800a84b5c4e7682/Formatter/OutputFormatterStyle.php#L85-L90), which has to similarly determine whether a given terminal supports hyper links.~~
~~Personally I think this is a sufficiently small amount of code which had to be transformed enough to consider it now new code - but in case you disagree, you can [refer to their license](https://github.com/symfony/console/blob/6.2/LICENSE) and decide whether it is MIT compatible (I am not a lawyer so I cannot make that determination myself).~~
~~EDIT: Okay well github can just outright tell me it's an MIT license, that works too 😅~~"
EDIT2: See comments - we're not using that logic anymore.

@mcornella in https://github.com/ohmyzsh/ohmyzsh/issues/11545#issuecomment-1458236908 you said the following about hyperlink support:
> We are already testing for that feature on a variety of terminals in the installer and updater, so it isn't difficult to adapt the terminal changelog to add suport for that. If the terminal doesn't support this, it isn't feasible to add links as they'd be fully visible in the terminal output.

Given that, I suspect you'll have an opinion about the suitability of this PR. I'll be happy to amend it as required, especially if you have anything else you'd like in that conditional statement.

Fixes #11545

### DIFF
--- a/tools/changelog.sh
+++ b/tools/changelog.sh
@@ -157,6 +157,89 @@ function parse-commit {
   fi
 }
 
+################################
+# SUPPORTS HYPERLINKS FUNCTION #
+################################
+
+# The code for checking if a terminal supports hyperlinks is copied from install.sh
+
+# The [ -t 1 ] check only works when the function is not called from
+# a subshell (like in `$(...)` or `(...)`, so this hack redefines the
+# function at the top level to always return false when stdout is not
+# a tty.
+if [ -t 1 ]; then
+  is_tty() {
+    true
+  }
+else
+  is_tty() {
+    false
+  }
+fi
+
+# This function uses the logic from supports-hyperlinks[1][2], which is
+# made by Kat Marchán (@zkat) and licensed under the Apache License 2.0.
+# [1] https://github.com/zkat/supports-hyperlinks
+# [2] https://crates.io/crates/supports-hyperlinks
+#
+# Copyright (c) 2021 Kat Marchán
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+supports_hyperlinks() {
+  # $FORCE_HYPERLINK must be set and be non-zero (this acts as a logic bypass)
+  if [ -n "$FORCE_HYPERLINK" ]; then
+    [ "$FORCE_HYPERLINK" != 0 ]
+    return $?
+  fi
+
+  # If stdout is not a tty, it doesn't support hyperlinks
+  is_tty || return 1
+
+  # DomTerm terminal emulator (domterm.org)
+  if [ -n "$DOMTERM" ]; then
+    return 0
+  fi
+
+  # VTE-based terminals above v0.50 (Gnome Terminal, Guake, ROXTerm, etc)
+  if [ -n "$VTE_VERSION" ]; then
+    [ $VTE_VERSION -ge 5000 ]
+    return $?
+  fi
+
+  # If $TERM_PROGRAM is set, these terminals support hyperlinks
+  case "$TERM_PROGRAM" in
+  Hyper|iTerm.app|terminology|WezTerm) return 0 ;;
+  esac
+
+  # kitty supports hyperlinks
+  if [ "$TERM" = xterm-kitty ]; then
+    return 0
+  fi
+
+  # Windows Terminal also supports hyperlinks
+  if [ -n "$WT_SESSION" ]; then
+    return 0
+  fi
+
+  # Konsole supports hyperlinks, but it's an opt-in setting that can't be detected
+  # https://github.com/ohmyzsh/ohmyzsh/issues/10964
+  # if [ -n "$KONSOLE_VERSION" ]; then
+  #   return 0
+  # fi
+
+  return 1
+}
+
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @GuySartorelli thank you for taking care of this! I was not able to start working unfortunately so I really appreciate it. 

I can't test it right now on a supporting terminal, but I can point to our current code for detecting this: https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/install.sh#L140

I don't think we're testing for JetBrains so that should be a nice edition. It's not an externalized function so it should be copy-pasted, but at some point we should move it out because it's already used twice.

Let me know if you have any issues applying that change, the rest looks fine but I'll still want to test it. Cheers!

GuySartorelli:
Awesome, I'll make the change.
I'm happy to copy that code into this file if you like, but if there's somewhere that global functions can go I'd also be very happy to move it to the appropriate place - just let me know where would be the best spot for it.

mcornella:
It's tricky because these files (`tools/upgrade.sh` and `tools/install.sh` are supposed to be used separately, and especially the installer should be a standalone file. For now it's fine to copy the code and we'll think of refactoring that in a future PR.

GuySartorelli:
Sweet, no worries. I've copied that code over to the changelog file.

I ended up not adding a check for jetbrains because the check I had assumed everything supports it, and only marked otherwise if something explicitly doesn't support it - but the checks you have at the moment assume the opposite - so jetbrains should end up returning `1` anyway.

Let me know if you need any more changes once you've gotten a chance to do some testing.

carlosala:
There's https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/upgrade.sh#L113 to do the check and format the links!

GuySartorelli:
@carlosala are you recommending a cha ge or just pointing out more duplicate code?

If the former, can you please be very specific? Do you recommend I copy/paste the relevant code or is there a way to share it between the two?

I don't think I'll be able to use that existing function for the formatting in `sed`, and it seems like bad practice to copy a function to format the links if I'm only using it once in this file.

Also are you saying that the fallback for this text if hyperlinks aren't supported should be to use underlines? That would be a departure from the current formatting.

carlosala:
I was just pointing you to that function. It's nice to use an already tested function to prevent errors with those codes, etc. Even though, if you think it's going to be difficult, you can also adapt the function for this particular usage. It's just to avoid all `if then else fi` blocks.

GuySartorelli:
Ahh okie doke. I think the conditions I have here are simple enough, and like I say I don't think this would work with the `sed` lines so it would only really mean replacing one if statement with a more complicated one, I think.

I do notice the syntax for formatting the link is slightly different in that function though so when I'm on my pc again tomorrow I'll swap out so the same formatting code is used across the board.

carlosala:
Nice!

GuySartorelli:
Hmm okay so I tried putting the formatting (i.e. `printf '\033]8;;%s\033\\%s\033]8;;\033\\ "$url" "$text"'`) into this PR but it resulted in incorrect results, e.g:
> - [e\[33m95d0c4b](https://github.com/ohmyzsh/ohmyzsh/commit/95d0c4b) [theme-and-appearance]

So for whatever reason it looks like the formatting used in `upgrade.sh` isn't suitable here. If someone knows what's going on and can suggest improvements I'll be happy to make them - but I'll leave it as is for now, since this works well and I can't find a way to integrate the code from `upgrade.sh` without breaking things.

mcornella:
Looks awesome! 
image

I think it's fine that we don't use the same functions as the installer, especially given that we're already using sed for `md` output. At some point we'll have to thoroughly refactor that, but I'd rather push this big improvement and leave the rest for the future.

Thank you so much!

PR #11580, opened 2023-03-24
Add toolbox run (tbr) to Toolbox plugin

francoism90:
## Standards checklist:

## Changes:

- Add `tbr` to allow run `toolbox run`.
- Change `te` to `tbe` to have them in sync.

## Other comments:

I'm thinking of adding more commands, but I'm still new to Toolbox.

### DIFF
--- a/plugins/toolbox/README.md
+++ b/plugins/toolbox/README.md
@@ -22,4 +22,5 @@ RPROMPT='$(toolbox_prompt_info)'
 
 | Alias | Command              | Description                            |
 |-------|----------------------|----------------------------------------|
-| tb    | `toolbox enter`      | Enters the toolbox environment         |
+| tbe   | `toolbox enter`      | Enters the toolbox environment         |
+| tbr   | `toolbox run`        | Run a command in an existing toolbox   |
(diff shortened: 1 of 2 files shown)


carlosala:
could you update readme? I think adding one alias would be enough, but tell me if something essential is missing!

christianparpart:
Aliasing `tr` might accidentally break regular uses of the more advanced users not expecting this. Because `tr` is actually a [valid UNIX command](https://en.wikipedia.org/wiki/Tr_(Unix)#:~:text=tr%20is%20a%20command%20in,in%20its%20input%20data%20set.). :-)

carlosala:
we could move the other alias to `tbe` and this to `tbr`!

francoism90:
@carlosala Sorry for not being that active, if it's Ok, I'll change the alias for both. :)

francoism90:
Also is `tb` not used for Thunderbolt devices?

carlosala:
not sure, where?

francoism90:
@carlosala I have updated the PR, let me know, maybe more can be added. :)

PR #11581, opened 2023-03-26
Added the possibility of showing all preferred themes using theme_chooser.sh

maonat:
## Standards checklist:

## Changes:

- Created new function that allows to show the preferred themes written in .zsh_favlist file
- Added option to run the new function

## Other comments:

This function is a simple addition to the current script that allows to go through the .zsh_favlist file and show the previously selected themes set as favourites.
It's a simple addition to the already existing functionality to show all themes but instead will show just the ones liked before.

### DIFF
--- a/tools/theme_chooser.sh
+++ b/tools/theme_chooser.sh
@@ -29,6 +29,25 @@ function theme_preview() {
     print -P "$PROMPT                                                                                      $RPROMPT"
 }
 
+function fav_theme_preview() {
+    if [[ -s $FAVLIST ]]; then
+        for THEME_NAME in $(cat $FAVLIST); do
+            THEME="$THEME_NAME.zsh-theme"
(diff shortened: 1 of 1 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry it sat so long. The tool writes `.zsh_favlist` but never reads it back, so `-f` is a sensible thing to add. I ran it and it works as described.

Two things before I can merge:

1. The diff changes the file mode from 755 to 644, so `$ZSH/tools/theme_chooser.sh` stops being runnable directly after merge. `chmod +x tools/theme_chooser.sh` and commit that.
2. `fav_theme_preview` copies the body of `theme_preview`. It can just call it: `theme_preview "$THEME_NAME.zsh-theme"` inside the loop, and the duplicate lines go away.

Optional: there's trailing whitespace on a couple of the added lines, and "No ($FAVLIST) in the system" would read better as "No favourite list found at $FAVLIST".

If you'd rather not pick this back up after this long, that's completely fine. Just say so and I'll take it from here.

PR #11582, opened 2023-03-26
feat(docker): add more pruning aliases to docker plugin

augustfengd:
## Standards checklist:

## Changes:

- add some more pruning commands to docker plugin

## Other comments:

There is already an existing alias for pruning, and this commit introduces some more pruning commands that are consistent with the existing one.


JimMoen:
please update docs page [docker/README.md](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/docker/README.md)

augustfengd:
@JimMoen I added another alias (`dprune`) which combines all the prune commands together and updated the docs. Please let me know what you think, thanks!

plugins/docker/docker.plugin.zsh
```diff
@@ -28,7 +28,11 @@ alias dstp='docker container stop'
 alias dtop='docker top'
 alias dvi='docker volume inspect'
 alias dvls='docker volume ls'
+alias dcprune="docker container prune"
+alias diprune="docker image prune"
+alias dnprune="docker network prune"
 alias dvprune='docker volume prune'
+alias dprune="dcprune; diprune; dnprune; dvprune"
```
alvincrespo, line 35: @augustfengd Wouldn't `docker system prune -a` be more appropriate here?

alvincrespo: ref: https://docs.docker.com/reference/cli/docker/system/prune/

augustfengd: Yes! This makes sense. I've updated my PR to do that.

augustfengd: I also grouped the aliases by command so the diff is a bit noisy now.

alvincrespo (approved):
This looks awesome! We should definitely get this in 😄

alvincrespo:
@carlosala Would it be possible to get this one merged in?

baltic-tea:
Hi, @augustfengd! Create a pull request with your changes on my branch if you want: https://github.com/ohmyzsh/ohmyzsh/pull/11821

Maybe someday all of this will be merged 🙏

PR #11591, opened 2023-03-30
fix(installer): don't use `sudo` when user is in Termux

Webpage-gh:
Using "sudo -k chsh" in termux will cause an error "env: exec -k: No such file or directory", so it needs to judge

## Standards checklist:

## Changes:

- Add a link to determine whether sudo is tsu in install.sh

## Other comments:

...

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -84,6 +84,8 @@ command_exists() {
 user_can_sudo() {
   # Check if sudo is installed
   command_exists sudo || return 1
+  # Having tsu linked to sudo will cause errors.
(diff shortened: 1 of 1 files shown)


mcornella:
I modified the check so that it uses the current detection logic of checking `$PREFIX`. Thank you for the report and contribution!

Webpage-gh:
I found that the variable "termux" will be set to true during dynamic debugging, so you can use something like "[ $termux = false ]" to judge more easily.

mcornella:
Yes that was my first approach but ended changing this in case we end up using `user_can_sudo` in other places in the future.

PR #11596, opened 2023-04-02
feat(extract): always extract files into its own folder

### LINKED ISSUE #11537: extract Plugin: Can automatically create the decompression directory

extract plugin

If there is no package folder in the compressed file, a bunch of scattered files may appear after using extract plugin to extract.

For compressed files without folder package inside, can they be automatically put into a folder after decompression

_No response_

### Additional context

_No response_

### Related Issues

_No response_

mcornella:
## Standards checklist:

## Changes:

- Creates a folder based on file name and extracts the contents from within this folder
- Remove the folder if there was an error
- If the only content in the extraction folder is one folder, move the folder up and remove the created extraction folder

Fixes #11537

### DIFF
--- a/plugins/extract/extract.plugin.zsh
+++ b/plugins/extract/extract.plugin.zsh
@@ -29,61 +29,75 @@ EOF
     local success=0
     local extract_dir="${1:t:r}"
     local file="$1" full_path="${1:A}"
+
+    # Create an extraction directory based on the file name
+    command mkdir -p "$extract_dir"
+    builtin cd -q "$extract_dir"
+
     case "${file:l}" in
       (*.tar.gz|*.tgz)
-        (( $+commands[pigz] )) && { tar -I pigz -xvf "$file" } || tar zxvf "$file" ;;
+        (( $+commands[pigz] )) && { tar -I pigz -xvf "$full_path" } || tar zxvf "$full_path" ;;
       (*.tar.bz2|*.tbz|*.tbz2)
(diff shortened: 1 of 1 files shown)


plugins/extract/extract.plugin.zsh
```diff
@@ -29,61 +29,75 @@ EOF
...
 
-    # Go back to original working directory in case we ran cd previously
+    # Go back to original working directory
+    # and remove extraction directory if there was an error
     builtin cd -q "$pwd"
+    (( success > 0 )) && command rm -r "$extract_dir"
+
+    # If content of extract dir is a single directory, move its contents up
+    # Glob flags: N: no error if directory is empty, Y2: at most 2 files
+    local content=("${extract_dir}"/*(NY2))
```
carlosala, line 97: Being a local array, shouldn't be declared and then initialized, or that is just with associative arrays?

carlosala, line 97: Here `$content` does not take hidden files into account I'd say.

carlosala (commented):
Generally looks good to me! Just those two small things, really nice improvement Marc! 😊

plugins/extract/extract.plugin.zsh
mcornella: Added `D` ("glob dots") glob flag to account for that as well.

carlosala (approved):
Super fast fix! I learnt a new glob qualifier haha :)
LGTM😉

PR #11597, opened 2023-04-02
feat(archlinux): unify `upgrade` function

deimosian:
## Standards checklist:

## Changes:

- Arch Linux PLugin: Unifies the Upgrade aliases into a single function which is capable of cleanly handling an out of date Arch Linux PGP Keyring to prevent failure before handing off the system upgrade to pacman or an AUR helper.

## Other comments:

...

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -180,15 +180,17 @@ fi
 
 # Check Arch Linux PGP Keyring before System Upgrade to prevent failure.
 function upgrade() {
-  echo ":: Checking Arch Linux PGP Keyring..."
+  echo "[oh-my-zsh] Checking Arch Linux PGP Keyring"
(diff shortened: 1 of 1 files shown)


Froggo8311:
Maybe use `-Syuu` rather than `-Syu`? So that version fragmentation doesn't happen (like it did recently with OpenSSL :P)

deimosian:
Well, the second -u enables package downgrading, which is not normally needed.

I'd also consider that kinda out of scope for this, I'm only moving the -Syu lines around not actually changing them.

carlosala:
Thanks for your contribution!

deimosian:
You're welcome.

Unfortunately, your changes broke it. The keyring pre-upgrade needs to be done with `-Sy` _not_ `-Syu`, otherwise it tries to run the full upgrade with the out of date keyring and fails.

Additionally I had done the cosmetics of the echo lines to seamlessly integrate with Yay and Pacman, if that's not allowed then whatever but if it's not I'd rather it have the clean look.

PR #11599, opened 2023-04-04
Add support for detecting Contour terminal in title() function

christianparpart:
## Standards checklist:

## Changes:

Contour Terminal is is supporting OSC 1 and OSC 2 as well as host programmable statusline, which can be controlled via terminfo entries `tsl` / `fsl` / `dsl`. OMZ seems to use `tsl`/`fsl` to fallback-set the window title, and that is heavily confusing. So this change ensures OSC 1 + OSC 2 logic is used, to streamline this change with most TEs.

## Other comments:

I personally think that the host programmable statusline is a bad start to simply show that string is probably not the best choice to begin with. Despired having rare support for it. :)

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -17,7 +17,7 @@ function title {
   : ${2=$1}
 
   case "$TERM" in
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks!

PR #11601, opened 2023-04-05
fix(jump): even spacing between marks

ahrex:
Previously in 3c3766fdf5f64daa4a4e96bee56487314ebcd642, the applied patch did not account for the characters that would be consumed by colour coding.

Fix this by using `printf` on the inner mark name, which does not yet have colour coding.

## Standards checklist:


carlosala:
Thanks for your contribution!

PR #11607, opened 2023-04-09
feat(archlinux): fix keyring update function

deimosian:
## Standards checklist:

## Changes:

- Fixed Arch Linux upgrade's keyring update function.

## Other comments:

...

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -186,7 +186,7 @@ function upgrade() {
   if [ $installedver != $currentver ]; then
     echo "[oh-my-zsh] Arch Linux PGP Keyring is out of date."
     echo "[oh-my-zsh] Updating before full system upgrade."
-    sudo pacman -Syu --needed --noconfirm archlinux-keyring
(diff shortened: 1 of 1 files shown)


deimosian:
Fixes bug introduced in https://github.com/ohmyzsh/ohmyzsh/pull/11597

carlosala:
Sorry for that, you're right! I changed that to keep track what's omz and what's the CLI used to upgrade, I think it's great to have it like that.

deimosian:
I just dislike how it looks compared to making it blend in. I also don't think the Arch user who's calling an OMZ plugin function needs their hand held like that. 

For example, the two different versions with `Yay` installed.



Vs

carlosala:
Let's change that then! Maybe removing those `::` at the beginning, but let's go for it! It's not a big deal, could you open a PR?

deimosian:
I'd like to keep the `::`. They are used for section headers for Pacman and Pacman helpers. 

Without OMZ is starts with the `:: Synchronizing package databases...` line. 

The blue `::` are for Pacman while the green `::` are for Yay, so the `::` without color applied do subtly differentiate it.

I'll open another PR, do with it what you will.

PR #11611, opened 2023-04-11
FEAT(terraform-workspace-prompt): New plugins to support terraform workspace name in prompt.

ocontant:
## Standards checklist:

## Changes:

- Add support to display terraform workspace name in Agnoster theme prompt.
- This plugin support Agnoster prompt engine
- This plugin can be use standalone by adding a prompt variable in .zshrc

** This plugin doesn't integrate aliases, to avoid aliases pollution. **

## Other comments:
There is currently no plugin that support to easily and aesthetically display terraform workspace in zsh prompt, using agnoster theme prompt engine. 

I added a terraform prompt logic, to uniformely support terraform workspace name to be displayed in the agnoster prompt.

I'm currently using this code via /.oh-my-zsh/custom/... structure.

Dependency: 
- [Agnoster + Terraform prompt] (https://github.com/ocontant/terraform-agnoster)

If you find this contribution useful, feel free to merge. 

Best Regards,

### DIFF
--- a/plugins/terraform-workspace-prompt/README.md
+++ b/plugins/terraform-workspace-prompt/README.md
@@ -1,0 +1,90 @@
+# Terraform plugin to display workspace in prompt
+
+Plugin for Terraform from Hashicorp, an infrastructure as code tool that allows for multi-cloud support, modularity, 
+and the ability to plan and apply changes, making it easier to manage and provision infrastructure resources.
+
+## Features:
+- Completion for `terraform`
+- aliases
+- prompt function
+- Support for agnoster template
+
+## Structure:
+```sh
+.
+├── README.md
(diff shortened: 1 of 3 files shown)


robbyrussell:
Thanks for putting this together, and sorry it sat unanswered for so long.

Oh My Zsh already covers this in a different shape: the `terraform` plugin provides `tf_prompt_info`, and agnoster picked up a `prompt_terraform` segment that uses it in https://github.com/ohmyzsh/ohmyzsh/pull/7892. A second terraform plugin with its own copy of the completion file would be two things to keep in sync, so I'm closing this one.

If the existing segment is missing something you had here, an issue describing the gap would be the right next step.

PR #11616, opened 2023-04-14
Add update alias to pipenv plugin.

SnoopDraugg:
## Standards checklist:

## Changes:

- [add `pdt` alias to `pip update`]

## Other comments:

Updating all the packages in the env is something I do on a regular basis and I hope that others would as well.


carlosala:
`pup` is already a CLI tool. Could we get another alias instead?
https://command-not-found.com/pup

SnoopDraugg:
> `pup` is already a CLI tool. Could we get another alias instead?
> https://command-not-found.com/pup

We sure can. 

What about `pdt`? A preliminary search doesn't turn up any major conflicts.

Alternatively, `pupd` or even `pupdt`? I'd like to try to keep the alias in line with the majority of the other aliases; but it's not necessary.

carlosala:
pupd seems a better option, right?

SnoopDraugg:
> pupd seems a better option, right?

You're right. I'm overthinking it.

SnoopDraugg (commented):
Rename pup alias to pupd

carlosala:
Thanks for your contribution!

SnoopDraugg:
> Thanks for your contribution!

Thank you!

PR #11619, opened 2023-04-17
feat(aws): allow more customisation in prompt function

### LINKED ISSUE #11618: more detailed customization for `aws` plugin prompt

aws

_No response_

1. Empty variable should be treated as null variable, not unset variable.

if I `export ZSH_THEME_AWS_PROFILE_PREFIX=""`, it expands to`<aws:` because of the following code:
```
${ZSH_THEME_AWS_PROFILE_PREFIX:=<aws:}
```

([here](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/aws/aws.plugin.zsh#L202) and [here](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/aws/aws.plugin.zsh#L206))

2. Space between profile and region should be option, not mandatory.

```
_aws_to_show+=" "
```

I think it makes sense but it would be greater if I can remove or override this. I want something like `[admin/us-west-1]`.

([here](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/aws/aws.plugin.zsh#L205))

for feature 1 : use `-z` condition statement instead of `:=` expansion.

for feature 2 : make a space to optional (customizable) variable.

If moderators allow, I want to resolve this issue with my code.

### Additional context

_No response_

### Related Issues

_No response_

roeniss:
resolves #11618

tl;dr : `$(aws_prompt_info)` now can be `admin|us-east-1`

## Standards checklist:

## Changes

- feature 1: allow empty string for variables
  - e.g., `ZSH_THEME_AWS_PROFILE_PREFIX` can be an empty string (`""`)
- feature 2: add a divider variable instead of a single space

## Other comments:

Slight worry for the second feature: Before this commit, the divider (a single space) i visible only when AWS_PROFILE and AWS_REGION were set. But this PR, for straightforward usage, makes it always visible.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -203,9 +203,8 @@ function aws_prompt_info() {
     _aws_to_show+="${ZSH_THEME_AWS_PROFILE_PREFIX="<aws:"}${AWS_PROFILE}${ZSH_THEME_AWS_PROFILE_SUFFIX=">"}"
   fi
 
-  _aws_to_show+="${ZSH_THEME_AWS_DIVIDER=' '}"
-
   if [[ -n $AWS_REGION ]]; then
+    [[ -n $AWS_PROFILE ]] && _aws_to_show+="${ZSH_THEME_AWS_DIVIDER=' '}"
     _aws_to_show+="${ZSH_THEME_AWS_REGION_PREFIX="<region:"}${region}${ZSH_THEME_AWS_REGION_SUFFIX=">"}"
   fi
 
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -200,28 +200,13 @@ function aws_prompt_info() {
   local region="${AWS_REGION:-${AWS_DEFAULT_REGION:-$AWS_PROFILE_REGION}}"
 
   if [[ -n $AWS_PROFILE ]];then
-    if [[ -z ${ZSH_THEME_AWS_PROFILE_PREFIX+x} ]]; then # empty string is evaluated as false (ref : https://stackoverflow.com/a/13864829/8556340)
-      ZSH_THEME_AWS_PROFILE_PREFIX="<aws:"
-    fi
-    if [[ -z ${ZSH_THEME_AWS_PROFILE_SUFFIX+x} ]]; then
-      ZSH_THEME_AWS_PROFILE_SUFFIX=">"
-    fi
-    _aws_to_show+="${ZSH_THEME_AWS_PROFILE_PREFIX}${AWS_PROFILE}${ZSH_THEME_AWS_PROFILE_SUFFIX}"
+    _aws_to_show+="${ZSH_THEME_AWS_PROFILE_PREFIX="<aws:"}${AWS_PROFILE}${ZSH_THEME_AWS_PROFILE_SUFFIX=">"}"
   fi
 
-  if [[ -z ${ZSH_THEME_AWS_DIVIDER+x} ]]; then
-    ZSH_THEME_AWS_DIVIDER=" "
-  fi
-  _aws_to_show+="${ZSH_THEME_AWS_DIVIDER}"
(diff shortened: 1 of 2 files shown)


plugins/aws/aws.plugin.zsh
```diff
@@ -198,13 +198,32 @@ compctl -K _aws_profiles asp acp aws_change_access_key
 function aws_prompt_info() {
   local _aws_to_show
   local region="${AWS_REGION:-${AWS_DEFAULT_REGION:-$AWS_PROFILE_REGION}}"
+
   if [[ -n $AWS_PROFILE ]];then
-    _aws_to_show+="${ZSH_THEME_AWS_PROFILE_PREFIX:=<aws:}${AWS_PROFILE}${ZSH_THEME_AWS_PROFILE_SUFFIX:=>}"
```
carlosala, line 202: Just removing the `:` would do the trick. It should be `${ZSH_THEME_AWS_PROFILE_PREFIX=<aws:}`, for example.
There's no need of adding all those if cases.

plugins/aws/aws.plugin.zsh
```diff
@@ -198,13 +198,32 @@ compctl -K _aws_profiles asp acp aws_change_access_key
...
+  fi
+
+  if [[ -z ${ZSH_THEME_AWS_DIVIDER+x} ]]; then
+    ZSH_THEME_AWS_DIVIDER=" "
   fi
+  _aws_to_show+="${ZSH_THEME_AWS_DIVIDER}"
+
   if [[ -n $AWS_REGION ]]; then
-    [[ -n $AWS_PROFILE ]] && _aws_to_show+=" "
-    _aws_to_show+="${ZSH_THEME_AWS_REGION_PREFIX:=<region:}${region}${ZSH_THEME_AWS_REGION_SUFFIX:=>}"
```
carlosala, line 206: Same happens here.

plugins/aws/aws.plugin.zsh
```diff
@@ -198,13 +198,32 @@ compctl -K _aws_profiles asp acp aws_change_access_key
...
+    _aws_to_show+="${ZSH_THEME_AWS_PROFILE_PREFIX}${AWS_PROFILE}${ZSH_THEME_AWS_PROFILE_SUFFIX}"
+  fi
+
+  if [[ -z ${ZSH_THEME_AWS_DIVIDER+x} ]]; then
+    ZSH_THEME_AWS_DIVIDER=" "
   fi
+  _aws_to_show+="${ZSH_THEME_AWS_DIVIDER}"
+
   if [[ -n $AWS_REGION ]]; then
-    [[ -n $AWS_PROFILE ]] && _aws_to_show+=" "
```
carlosala, line 205: Let's also have this check. It'll prevent adding the divider if the region or profiles are not set.

plugins/aws/aws.plugin.zsh
roeniss: I didn't know this trick. Thank you! It was so hard to find [the reference](https://tldp.org/LDP/abs/html/parameter-substitution.html)...  I'll change it.

plugins/aws/aws.plugin.zsh
roeniss: https://github.com/ohmyzsh/ohmyzsh/pull/11619#discussion_r1171338084

plugins/aws/aws.plugin.zsh
roeniss: My basic idea was that most users would have both profile and region env. But it seems not great guess so I will change it like this: 

```
_aws_to_show+="${ZSH_THEME_AWS_DIVIDER}" # moves this in the previous commit to the...

  if [[ -n $AWS_REGION ]]; then
    [[ -n $AWS_PROFILE ]] && _aws_to_show+="${ZSH_THEME_AWS_DIVIDER=' ' }" # here
```

What do you think? Let me just push a commit first so you can see it easily...

carlosala (approved):

carlosala:
Thanks for your contribution!

PR #11627, opened 2023-04-19
feat(npm): add `npmrd` alias

KozmikNano:
## Standards checklist:

## Changes:

- Add an npmRD alias in the npm plugin to use `npm run dev`

## Other comments:

Many developers use `npm run dev` to test their code. If an alias for `npm test` exists, then `npm run dev` should also exist.


carlosala (commented):
Nice one! do you mind if we change it with `npmrd`?

KozmikNano:
> Nice one! do you mind if we change it with `npmrd`?

Not at all, I have made the change. 😃

carlosala:
Could you change that also in the readme?

carlosala:
Thanks for your contribution!

KozmikNano:
> Thanks for your contribution!

Thanks for merging!

PR #11630, opened 2023-04-19
fix(dirhistory): run properly if `ksh_arrays` is set

ntninja:
## Standards checklist:

## Changes:

- dirhistory plugin: Don’t fail with “parameter not set” errors on navigation in shells with -o ksh_arrays


carlosala:
Thanks for your contribution!

PR #11635, opened 2023-04-20
feat(dbt): Add plugin

msempere:
## Standards checklist:

## Changes:

- add aliases for [dbt](https://docs.getdbt.com/)


carlosala:
Hi! We [have a policy](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases) about which aliases to add, and I'd say this PR doesn't fit on it. Every alias needs to be explained, which is the workflow where'd be used, and not having a lot of them.
Could you please rewrite it, thanks!

msempere:
> Hi! We [have a policy](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases) about which aliases to add, and I'd say this PR doesn't fit on it. Every alias needs to be explained, which is the workflow where'd be used, and not having a lot of them.
> Could you please rewrite it, thanks!

Sure, np. Thanks for taking a look.

msempere:
@carlosala I have modified the aliases to focus more on workflows instead of simple commands based on your feedback. Please take another look and let me know what you think. Thanks

carlosala:
Thanks for your contribution!

PR #11636, opened 2023-04-21
Jadjay lxd plugin enhancement

jadjay:
## Standards checklist:

## Changes:

- Add elements listing
- Add instance listing
- Separate Flags, Global Flags, Commands
- Delete debug logs

## Other comments:

This avoid errors while using lxd autocompletion : 

Like this one : 

```shell
$ lxc image lis[TAB]
```
```shell
$ lxc image lisUsage:
  lxc image [flags]
  lxc image [command]

Available Commands:
  alias          Manage image aliases
  copy           Copy images between servers
  delete         Delete
(...)
```

### DIFF
--- a/plugins/lxd/lxd.plugin.zsh
+++ b/plugins/lxd/lxd.plugin.zsh
@@ -9,7 +9,7 @@ _lxc_get_global_flags () {
 }
 
 _lxc_get_command_list () {
-    echo "___command"
+    
     $_comp_command1 -h              | _lxc_get_commands
     $_comp_command1 -h              | _lxc_get_global_flags
 }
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and sorry it waited so long. Making `lxc <cmd><TAB>` actually complete instead of dumping usage was a reasonable thing to want.

Since this was opened, upstream `lxc` started shipping its own completion (`lxc completion zsh`, LXD 5.21+), including the live instance and image names you were pulling from `lxc list`. I don't think we should carry a parser for `-h` output when the tool now does this itself, so I'm closing this. The better follow-up is probably to have the plugin load `lxc completion zsh` when it's available. If you'd like to do that one, it's yours.

PR #11649, opened 2023-04-27
fix(aws): set the divider to an empty string when it is not provided

Idan-Fishman:
## Standards checklist:

## Changes:
- Last week's push changed the default divider view, and now everyone sees an annoying ' ' between the role and the region. To fix this issue, I set the default divider to an empty string character, which removes the unwanted character.

### before changes:
Screenshot 2023-04-27 at 12 18 06

### after changes:
Screenshot 2023-04-27 at 12 19 35


carlosala:
Thanks for the catch! I set that as an space, to maintain the expected behaviour.

PR #11658, opened 2023-05-01
Update directory aliases to use cd command

mattdodge:
## Standards checklist:

## Changes:

- Use the `cd` command inside the `...`, `....`, etc aliases

## Other comments:

Ever since #11550 has shipped, my `...` aliases (which I used to override) have broken.
```
[user:~]$ ...
zsh: permission denied: ../..
```
The new behavior of actually unaliasing the aliases, rather than just [bouncing out early](https://github.com/ohmyzsh/ohmyzsh/blob/277f38212aef31a6baba2cf1a0a355af611be5e0/lib/directories.zsh#L12) means that my overrides no longer work. There is likely somewhere downstream I could move my aliases in my network of dotfiles, but I actually do like a lot of the directory aliases and I see no reason we can't just use the `cd` command in them rather than relying on some other shell shortcut.

### DIFF
--- a/lib/directories.zsh
+++ b/lib/directories.zsh
@@ -5,10 +5,10 @@ setopt pushd_ignore_dups
 setopt pushdminus
 
 
-alias -g ...='../..'
-alias -g ....='../../..'
-alias -g .....='../../../..'
-alias -g ......='../../../../..'
+alias -g ...='cd ../..'
+alias -g ....='cd ../../..'
+alias -g .....='cd ../../../..'
+alias -g ......='cd ../../../../..'
 
 alias -- -='cd -'
 alias 1='cd -1'


carlosala:
Those are global aliases, that means that are not only commands. For example, `mv foo ...` would actually expand to `mv foo ../..`.
That's why we can't accept your proposal. In that case, I don't see why you have issues with the new system, you can just create the aliases and omz will keep them for you.

mattdodge:
Ah neat, I didn't realize global aliases worked that way. That's actually quite convenient.

Is it expected that `...`, which expands to `../..`, would actually traverse up the directory tree? How does `../..` do that without a `cd`? Is that a zsh thing, an OMZ thing, or just a system shortcut?

carlosala:
It's a zsh option, that is set up [here](https://github.com/ohmyzsh/ohmyzsh/blob/2e7a247cab663fbc0aafdefa15bf10eb4dd64563/lib/directories.zsh#L2) in ohmyzsh. It is disabled by default.

carlosala:
@mattdodge all your problems should be fixed in #11659. Could you try it with your existing config with `omz pr test 11659` and opening a new terminal? Then, when exiting the first terminal, make sure to answer yes to go back to previous branch.

mattdodge:
@carlosala I'm still having issues, even on that branch. Although it seems like my root problem is something with `auto_cd` getting permission denied errors. I get the error when I run `/`, `../..`, or even `~`. All of those directories have correct chmod/chown permissions too. Weird.

To at least test your PR, I've overridden the `...` alias and set `zstyle ':omz:lib:directories' aliases no` - and at least that is respecting my alias. So I think the work you've done there is good - thanks for that! Sounds like I just have something else wrong with my `auto_cd` setup and this recent change is just revealing it.

carlosala:
nice, thanks for the testing!

PR #11661, opened 2023-05-02
feat(theme-and-appearance): allow disabling gnu-ls in bsd

### LINKED ISSUE #11647: Inconsistent `ls` alias, new window vs. omz reload; overwrites `alias ls='gls --color=tty'` only on reload
### Describe the bug

relevant info:

- os: macOS `13.3.1`
- arch: `arm64`
- terminal application: iTerm2 3.5.0beta10
- shell: zsh 5.9, `/bin/zsh` (macOS built-in version)
- special config: using ["typewritten"](https://typewritten.dev) theme
- plugins: occurs with or without any plugins

---

Opening a new window vs. using `omz reload` (or entering the debug session with `zsh -xv 2> >(tee ~/omz-debug.log &>/dev/null)`, for that matter) creates inconsistent `ls` alias:

**new terminal window/tab**
```sh
> type -a ls
ls is an alias for ls -G
ls is /bin/ls
```

**after `omz reload`, `zsh`, `tmux` etc.**
```sh
> type -a ls
ls is an alias for gls --color=tty
ls is /bin/ls
```

this is problematic (at least on macOS) since there are several `ls` flags (such as `ls -@` which shows extended attributes) that result in error with `gls`.

I'm pretty sure the culprit is in [`lib/theme-and-appearance`,  lines 48-57](https://github.com/ohmyzsh/ohmyzsh/blob/5d3e86e2a48adf7a308773f8f1b725d187c7c5ef/lib/theme-and-appearance.zsh#L48-L57):
```sh
  (darwin|freebsd)*)
    # This alias works by default just using $LSCOLORS
    test-ls-args ls -G && alias ls='ls -G'
    # Only use GNU ls if installed and there are user defaults for $LS_COLORS,
    # as the default coloring scheme is not very pretty
    [[ -n "$LS_COLORS" || -f "$HOME/.dircolors" ]] \
      && test-ls-args gls --color \
      && alias ls='gls --color=tty'
    ;;
  *)
```

NOTE: this does **not** occur when `DISABLE_LS_COLORS="true"`

why does this seem to run after invoking `zsh` but _not_ when `zsh` is initialized for a new window/tab?
it would be cool if there were a flag to turn off overwriting of builtin commands with coreutils variants _but still be able to have nice LS\_COLORS_

### Steps to reproduce

1. open a new terminal
2. run `which ls` or `type -a ls` or whatever
3. note the result
4. run any of the following: `omz reload`, `exec zsh`, `zsh`, `tmux`, `asciinema`
5. note that now `ls is an alias for gls --color=tty`

### Expected behavior

either `type -a ls` should be consistent with new window / omz reload OR ideally `gls` wouldn't be used by default on macOS for the reason that `gls` is missing some features unique to the macOS implementation (e.g. `-@`)

unfortunately, it's impossible to reproduce this using `asciinema` because it runs a new `zsh` instance and so the `gls` alias is already used. here is a regular-ass video

https://user-images.githubusercontent.com/8978069/234725142-bef44c40-cf09-4a36-8a58-a14d8f778ed8.mp4

### workaround

manually set in `.zshrc` before `source ${OMZ}/oh-my-zsh.sh` (`LS_COLORS` might be different on your system)
```sh
export LS_COLORS="di=1;36:ln=35:so=32:pi=33:ex=31:bd=34;46:cd=34;43:su=30;41:sg=30;46:tw=30;42:ow=30;43"
alias ls='ls -G'
DISABLE_LS_COLORS="true"
```

### OS / Linux distribution

macOS 13.3.1

### Zsh version

5.9

### Oh My Zsh version

master (5d3e86e)

### Terminal emulator

iTerm2 3.5.0beta10

None

### Additional context

using theme `typewritten` but doesn't seem to interfere.

carlosala:
## Standards checklist:

## Changes:

Closes #11647 

To disable gnu-ls (`gls`) even if it's installed in freeBSD and macOS you can set it up with:
```zsh
zstyle ':omz:lib:theme-and-appearance' gnu-ls no
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ To learn more, visit [ohmyz.sh](https://ohmyz.sh), follow [@ohmyzsh](https://twi
     - [Manual Installation](#manual-installation)
   - [Installation Problems](#installation-problems)
   - [Custom Plugins and Themes](#custom-plugins-and-themes)
+  - [Disable GNU ls in macOS and freeBSD systems](#disable-gnu-ls)
   - [Skip aliases](#skip-aliases)
 - [Getting Updates](#getting-updates)
   - [Updates verbosity](#updates-verbosity)
@@ -277,6 +278,18 @@ If you want to override any of the default behaviors, just add a new file (endin
 If you have many functions that go well together, you can put them as a `XYZ.plugin.zsh` file in the `custom/plugins/` directory and then enable this plugin.
 
 If you would like to override the functionality of a plugin distributed with Oh My Zsh, create a plugin of the same name in the `custom/plugins/` directory and it will be loaded instead of the one in `plugins/`.
+
+### Disable GNU ls in macOS and freeBSD systems
+
+<a name="disable-gnu-ls"></a>
+
+The default behaviour in Oh My Zsh is to use GNU `ls` even in macOS and freeBSD systems if it's installed (as
+`gls` command) when enabling colorized `ls` in `lib/theme-and-appearance.zsh`. If you want to disable this
+behaviour you can use zstyle-based config before sourcing `oh-my-zsh.sh`:
+
+```zsh
+zstyle ':omz:lib:theme-and-appearance' gnu-ls no
+```
 
 ### Skip aliases
 
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -20,9 +20,24 @@ if command diff --color /dev/null{,} &>/dev/null; then
   }
 fi
 
-
 # Don't set ls coloring if disabled
 [[ "$DISABLE_LS_COLORS" != true ]] || return 0
+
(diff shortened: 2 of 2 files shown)


hwknsj:
thank you for working on this issue! this definitely solves the problem and i commend you for it because i wasn't sure the most elegant way to do so myself!

I wrote a detailed response in [some comments](https://github.com/ohmyzsh/ohmyzsh/issues/11647#issuecomment-1535532596) in the issue #11647 thread

but tl;dr – perhaps this should be an _opt in_ instead of _opt-out_ situation (`homebrew`'s [`coreutils` formula provides a good precedent](https://github.com/Homebrew/homebrew-core/blob/fe9eec52b85c2c9e3a103ff3e5bf37398421e528/Formula/coreutils.rb#L78-L100), i think)

PR #11668, opened 2023-05-04
move $ZSH_TMUX_CONFIG from $HOME to $XDG_CONFIG_HOME

TirtharajPramanik:
## Standards checklist:

## Changes:

use `XDG_CONFIG_HOME` variable with default location `~/.config/tmux.conf` for `ZSH_TMUX_CONFIG`
```
${ZSH_TMUX_CONFIG:=${XDG_CONFIG_HOME:=~/.config/tmux.conf}}
```

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -26,7 +26,7 @@ fi
 # systems without the proper terminfo
 : ${ZSH_TMUX_FIXTERM_WITH_256COLOR:=screen-256color}
 # Set the configuration path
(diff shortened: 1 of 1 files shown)


carlosala:
I'd recommend you setting `$ZSH_TMUX_CONFIG` variable to wherever you'll have your config. As far as I can see in [latest ubuntu lts tmux man](https://manpages.ubuntu.com/manpages/jammy/en/man1/tmux.1.html) having the config file in `.config/tmux/tmux.conf` it's still not an standard.
Breaking that a lot of users wouldn't make sense in my opinion.
Thanks for your contribution!

PR #11671, opened 2023-05-05
fix(aws): correct access key disable command

grimm26:


carlosala:
Thanks for your contribution!

PR #11673, opened 2023-05-06
Update vi-mode plugin doc to prevent overriden settings

vdedios:
## Standards checklist:

## Changes:

- Add basic comment to document how to configure settings for vi-mode plugin

## Other comments:

- I had some problems configuring settings for vi-mode plugin until I realized the plugin was overriding them. It may be due to me being a noob using oh-my-zsh but this quick note would have saved me some time 😃


carlosala:
The plugin should not overwrite the settings. I'll further check to see what's going on. Thanks for the catch :)

carlosala:
I was so slow on this one, sorry for that. Thanks for the report!

PR #11676, opened 2023-05-08
Add new aliases for rebasing interactive

guenhter:
## Standards checklist:

## Changes:

- api plugin

## Other comments:

...


carlosala:
This aliases seem great, even though we have already a lot of aliases in git plugin, and this ones seem a bit specific (even though used, in my case it's part of my regular workflow). I'd recommend in this case to go `grbi main` which is already really short also.
Thanks for your contribution!

PR #11681, opened 2023-05-09
fix(robbyrussell) multiline command editing

roodkcab:
## Changes:

- [] bugfix for multiline command mess up on editing for robbyrussell theme

### DIFF
--- a/themes/robbyrussell.zsh-theme
+++ b/themes/robbyrussell.zsh-theme
@@ -1,5 +1,5 @@
-PROMPT="%(?:%{$fg_bold[green]%}➜ :%{$fg_bold[red]%}➜ )"
-PROMPT+=' %{$fg[cyan]%}%c%{$reset_color%} $(git_prompt_info)'
+PROMPT="%(?:%{$fg_bold[green]%}➜ :%{$fg_bold[red]%}➜ ) %{$fg[cyan]%}%c%{$reset_color%}"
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

themes/robbyrussell.zsh-theme
```diff
@@ -1,5 +1,5 @@
-PROMPT="%(?:%{$fg_bold[green]%}➜ :%{$fg_bold[red]%}➜ )"
-PROMPT+=' %{$fg[cyan]%}%c%{$reset_color%} $(git_prompt_info)'
+PROMPT="%(?:%{$fg_bold[green]%}➜ :%{$fg_bold[red]%}➜ ) %{$fg[cyan]%}%c%{$reset_color%}"
+PROMPT+=' $(git_prompt_info)'
```
milesfrain, line 2: I'm very confused about how moving part of the prompt string to the previous line has any effect on the end result. It seems like we'd get the same `$PROMPT` variable in either case.

I'm working on my own local `robbyrussel` theme edits and want to make sure I'm not misunderstanding an important concept.

I'd be grateful if anyone could explain how this fix works.

mcornella: The previous line uses double quotes, which allow expansion of variables inside it. Single quotes show it literally as it looks.

PR #11690, opened 2023-05-11
Allow nvm plugin to use autoload and lazy together

hoylemd:
## Standards checklist:

- [x] The code follows the code style guide detailed in the wiki. (I think?)

## Changes:

Just added a simplified version of nvm's nvmrc detection function for the autoload case so lazy and autoload can be used concurrently.  This is needed because normally autoload needs to load nvm to find the nvmrc, so it won't work with lazy.  This allows it to detect an nvmrc in the *current directory only*, to lazily load nvm automatically.

## Other comments:

I'm submitting this as a draft for now mostly just to see if anyone else thinks it's a good idea.  It works very well for my workflow.  My terminals that need a rich prompt and nvm pretty much always open in a directory with an nvmrc.  But it might be confusing for others?  It was motivated initially by the fact that the readme doesn't specify that the two options are mutually exclusive, so I expected them to work together.  But this was throwing errors so I wanted to 'fix' it.  This is how I did it for myself.  If this isn't a good distributable solution, I'll probably do a second PR just to make it more clear in the readme that you have to pick one of those settings (and maybe print a warning or something if they're both turned on anyways?)

### DIFF
--- a/plugins/nvm/README.md
+++ b/plugins/nvm/README.md
@@ -58,3 +58,11 @@ zstyle ':omz:plugins:nvm' silent-autoload yes
 ```
 
 Note: _this will not remove regular `nvm` output_
+
+##### autoload with lazy
+
+Lazy and autoload mode can be used concurrently, but when doing so, only the current working directory will
+be searched for an `.nvmrc` file, unlike the usual nvm behaviour of searching all parent directories.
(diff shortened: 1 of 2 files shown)


carlosala:
Thanks for your contribution!

PR #11692, opened 2023-05-13
docs(pyenv): warn about pyenv init

MojitoMonday:
## Standards checklist:

## Changes:

- Add a warning/instruction to the Pyenv plugin README.md in order to help deal with the confusing issue when using the plugin. Solution based on: https://github.com/pyenv/pyenv/issues/2041#issuecomment-990253001

## Other comments:
When you add the Pyenv plugin to the `~/.zshrc` file and restart the terminal, you can get a message about pyenv being badlye configured. Even after following all the suggested advice, it can often still appear. This is due to the installation instructions being unclear, and often simply appending the needed commands to the end of the `~/.zshrc` file. However, this can be cleared up by running the necessary commands AFTER the plugin is initialized.

There is no mention of this requirement anywhere outside of that one comment thread, and adding a simple suggestion would go a long way to helping new installs. I would prefer that the requirement for it being before the plugins are initialized not be a requirement, but I would not know how to make that fix.

### DIFF
--- a/plugins/pyenv/README.md
+++ b/plugins/pyenv/README.md
@@ -10,6 +10,14 @@ To use it, add `pyenv` to the plugins array in your zshrc file:
 plugins=(... pyenv)
 ```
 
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #11707, opened 2023-05-22
feat(bureau): ignore hidden git repos

### LINKED ISSUE #11706: [bureau theme] Hidden git repo support

Bureau theme

I am currently using git-annex to manage files in $HOME.
The repo has grown very big and after every command in the shell, the prompt hangs for a few seconds before appearing.

After checking the code in lib/git.zsh, I realized that it would ignore repos with the config key oh-my-zsh-hidden-info set to 1.
As the bureau theme doesn't use the lib/git.zsh functions to parse git repos, I suggest adding the feature to the bureau theme.

I first tried to switch the theme to using the git_prompt_info() function in lib/git.zsh but then the theme loses its nice colors and characters.

### Additional context

_No response_

### Related Issues

_No response_

rogarb:
The git_prompt_info() function in lib/git.zsh ignores git repos which contains a specific config key, allowing to effectively "hide" them from the prompt. Unfortunately, the bureau theme doesn't use the library function to build its prompt.

This commit modifies the specific prompt generation function in the bureau theme in order to achieve the same behaviour.

Fixes #11706

## Standards checklist:

## Changes:

- Add support for hiding git repos with oh-my-zsh.hidden-info set to 1 as in lib/git.zsh

### DIFF
--- a/themes/bureau.zsh-theme
+++ b/themes/bureau.zsh-theme
@@ -67,6 +67,12 @@ bureau_git_status() {
 }
 
 bureau_git_prompt() {
+  # ignore non git folders and hidden repos (adapted from lib/git.zsh)
+  if ! command git rev-parse --git-dir &> /dev/null \
+     || [[ "$(command git config --get oh-my-zsh.hide-info 2>/dev/null)" == 1 ]]; then
+    return
+  fi
+
   # check git information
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #11709, opened 2023-05-23
fix(core): Use full path to env in functions.zsh

ferdnyc:
## Standards checklist:

## Changes:

- Replace bare 'env' with '/usr/bin/env', in `uninstall_oh_my_zsh()`

## Other comments:

I have a shell alias for `env`, in my interactive shell environment. So, with just a bare `env` called in the `uninstall_oh_my_zsh()` function, re-executions of my `.zshrc` would report a parse error:

```sh
$ . ~/.zshrc                       
/.../lib/functions.zsh:8: parse error near `ZSH="$ZSH"'
```

It's safer to use `/usr/bin/env` there, same as in the various shebang lines for the executable scripts.

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -5,7 +5,7 @@ function zsh_stats() {
 }
 
 function uninstall_oh_my_zsh() {
-  env ZSH="$ZSH" sh "$ZSH/tools/uninstall.sh"
+  /usr/bin/env ZSH="$ZSH" sh "$ZSH/tools/uninstall.sh"
 }
 
 function upgrade_oh_my_zsh() {


ferdnyc:
Although it wasn't my intention in making this change, this PR may also help to address the issues encountered in the discussion at #11602.

However, to _completely_ address that issue, it may also be advisable to **unset** the `PATH` envvar in the `env` command (`/usr/bin/env -u PATH ZSH="$ZSH"`), so that it will reset to the system default and allow `sh "$ZSH/tools/uninstall.sh"` to run successfully even if the `$PATH` was previously broken.

Thoughts on adding a `-u PATH` to the command?

carlosala:
Thanks for your contribution!

ferdnyc:
@carlosala 

`command env` is a better solution for my case of having `env` aliased, agreed. But IIUC, it still won't help at all with situations like #11602 where (seemingly) the `$PATH` has been utterly corrupted, correct?

I wonder if there's still more that can be  done to make the uninstall _bulletproof_ no matter **what** state the system is in? (Just to give people a parachute for when they've broken something and can't figure out what / how to fix it.)

PR #11714, opened 2023-05-24
perf(git): improve `gunwip` alias

anastygnome:
## Standards checklist:

## Changes:

In order to get better performance, we only query the subject (first line of the commit message), using a plumbing command, `git rev-list`.

## Other comments:

The `-c` flag of `grep` was useless, as `-q` disables output, so it was removed.


carlosala (changes requested):
I like it! But I don't see the benefit of having two options. I'd leave the second one, which works for every git version, there's no much improvement on the first one.
Could you also update the README?

anastygnome:
> I like it! But I don't see the benefit of having two options. I'd leave the second one, which works for every git version, there's no much improvement on the first one. Could you also update the README?

I have changed the commit as per your remarks.

carlosala:
Thanks for your contribution!

PR #11724, opened 2023-06-03
feat(autoenv): macOS compatibility, add more dirs

diegocostares:
## Standards checklist:

## Changes:

- Create a support for `venv`, `env`, `.venv` and `.env` folder.
- Support for `activate.sh` or `activate` files.

## Other comments:

I have come across repositories that document that to create virtual environments they use variations of folders like "env" or "venv", using for example:
```bash
python3 -m venv .venv
```
To support these, add some variations to `$install_locations`.

In mac M1 (I have not checked if it happens in other OS or if it happens by the venv library) the files that are generated do not end in .sh, so I added compatibility by doing `-e $d/activate || -e $d/activate.sh`


carlosala:
Thanks for your contribution!

PR #11739, opened 2023-06-07
feat(rails): add some aliases

joshRpowell:
## Standards checklist:

## Changes:

Adds 

- rails routes --controller
- rails routes --expanded
- rails routes --unused

Updates

- rails routes --grep

## Other comments:

see [Rails Guides - routing](https://guides.rubyonrails.org/routing.html#listing-existing-routes) for new alias use cases


plugins/rails/rails.plugin.zsh
```diff
@@ -68,7 +68,9 @@ alias rmd='rails middleware'
 alias rn='rails notes'
 alias rp='rails plugin'
 alias rr='rails routes'
-alias rrg='rails routes | grep'
+alias rre='rails routes --expanded'
+alias rrg='rails routes -g'
+alias rrc='rails routes -c'
```
tjarbo, line 73: @joshRpowell Is it possible to use the "extended version" of the flags `-g` and `-c`? Like you did in line 71 with `--expanded`. This would make the documentation and use more intuitive, in case a user is not so familiar with the flags of the `rails routes` command.

carlosala:
Thanks for your contribution!

PR #11754, opened 2023-06-15
fix(docker): only generate completion for docker >23

carlosala:
## Standards checklist:

### DIFF
--- a/plugins/docker/docker.plugin.zsh
+++ b/plugins/docker/docker.plugin.zsh
@@ -1,13 +1,3 @@
-# If the completion file doesn't exist yet, we need to autoload it and
-# bind it to `docker`. Otherwise, compinit will have already done that.
-if [[ ! -f "$ZSH_CACHE_DIR/completions/_docker" ]]; then
-  typeset -g -A _comps
-  autoload -Uz _docker
-  _comps[docker]=_docker
-fi
-
-docker completion zsh >| "$ZSH_CACHE_DIR/completions/_docker" &|
-
 alias dbl='docker build'
 alias dcin='docker container inspect'
 alias dcls='docker container ls'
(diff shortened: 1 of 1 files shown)


YHYJ:
Why only generate completion for docker > 23, `docker version --format '{{.Client.Version}}` causes docker.service to start with terminal, it should have been started by docker.socket when running docker

carlosala:
@YHYJ `docker completion zsh` cli command only exists for docker >23

sdondley:
I'm not sure this patch was a very good idea. Looks like docker engine version 23 is just a few months old and _docker file can get generated from sources available on GitHub. And the code completion file generated by `docker completion zsh` is very old and does not complete at all for many subcommands, at least on macOS running latest version of Ventura (13.4).

YHYJ:
I don't think `docker version --format '{{.Client.Version}}'` command should be used, this caused docker.service to start unexpectedly

carlosala:
@YHYJ does `docker --version` start also the service for you?

carlosala:
@sdondley docker folks are switching to cobra v2 completions, generated by `docker completion zsh`. As it's new, it'll keep improving more and more. For example, it allows to have completion for `docker compose`, `docker buildx`, and other CLI plugins.
If you want to have another completion, you can of course download any script you want and add it to a directory present in `$fpath`/

sdondley:
OK, but if I have an older version of docker on the machine, this new version of the plugin won't work at all and even if I have a newer version, it will not work at all for many subcommands. Seems to me you should wait until it's at least as good as the old version before using `docker completion zsh` don't you think?

carlosala:
@sdondley I see your point, but the new version has that CLI plugins completion that the old doesn't, and the old one hasn't been updated for 4 years and the repo has been archived.

sdondley:
@carlosala The new version is missing at least half the subcommands the "old" completion script and it also has no subcommand description.  And the "old" one is actually very up to date and was updated just last month: https://github.com/docker/cli/blob/7d723e2ba7b8607cf9ca5b3ea541f1ce60c700ec/contrib/completion/zsh/_docker#L6

At any rate, I will not be using your version of the docker plugin and will be overriding it with the older version. But at a minimum, I recommend adding something to the README that explains why the new version is crippled until the new completion script automation feature with Cobra is implemented properly and how people can use the old version of the docker plugin to get better completion until that happens.

YHYJ:
> @YHYJ`docker --version`还为您启动服务吗？

Yes, `docker` will make docker.socket start docker.service

carlosala:
Then we can't workaround that, we need to assess the version to see whether or not we can generate the completion script.

ondra-m:
Is there a way how to get completion for older docker using latest omz please?

Issue #11750, opened 2023-06-13
gunwipall cluttering the reflog 


anastygnome:
### Describe the bug

Because of the way guwipall has been inplemented in  #11725 , one git reset command gets executed for each wip commit. 

This isn't ideal if we need to go back, as the reflog may get filled with resets 

I'm unable to send a PR at this time, but the following should do the trick 

```shell
commit=$(git rev-list --grep='--wip--' --invert-grep --max-count=1 HEAD)

# Check if a commit without "--wip--" was found and it's not the same as HEAD
if [[ "$commit" != "$(git rev-parse HEAD)" ]]; then
  git reset $commit || return 1
fi
```

### Steps to reproduce

Make a git repository with a few wip commits
1. git init 
2. git commit --allow-empty -m "init"
3. gwip --allow-empty  a few times
4. gunwipall 
5. git reflog 
6. The reflog is filled with resets

### Expected behavior

Ideally, only one reset should be done 

### OS / Linux distribution

N/A

### Zsh version

5.9

### Oh My Zsh version

master (cb8b677)

### Terminal emulator

N/A

### If using WSL on Windows, which version of WSL

None

anastygnome:
@kopach @carlosala

kopach:
Hi @anastygnome, even though I see value in suggested by you improvement, IMO `gunwipall` does exactly how it was intended and how it's documented - it "executes" `gunwip` multiple times. That said, IMO, this issue is not a `Bug` but rather `Change request` or feature.
Anyway, personally, I agree, that proposed by you solution is better in most cases and still leaves room to make `gunwip` manually multiple times if needed, so I agree and would like to update `gunwipall` if that's fine with maintainers of the project

carlosala:
hi! Sure, I think it's a good change! It should not be noticeable for people that's not using reflog. Feel free to submit a PR!

PR #11758, opened 2023-06-16
fix(git): `gunwipall` now only resets once

kopach:
fix: #11750

## Standards checklist:

## Changes:

-  improve `gunwipall` function to produce 1 reflog record instead of many.
 

## Other comments:

Solution was provided by @anastygnome in #11750. I've tested it and it seems to work as expected


kopach:
@carlosala, could you, please, take a look?

carlosala:
Checking with several versions, it seems like `rev-list` does not have `--grep` argument. Let's use `git log`, that since almost 10 years ago.

carlosala:
Thansk for your contribution!

PR #11766, opened 2023-06-20
Fix (docker plugin): use `command docker` to bypass aliases

Haltarys:
Use `command docker` to avoid aliases and functions that may be overriding the `docker` command.

Since most docker commands require `sudo` privileges, I added an `alias docker='sudo docker'` to my `.zshrc`. I noticed that if I put the alias before `source $ZSH/oh-my-zsh.sh`, the [docker plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/docker) uses the user-defined alias instead of the actual `docker` command.

By invoking docker with `command docker` instead, aliases and functions are ignored, and the correct behaviour is always observed.

## Standards checklist:

## Changes:

- Replaces invocations of docker with `command docker` (not a breaking change) in the plugin

## Other comments:

...

### DIFF
--- a/plugins/docker/docker.plugin.zsh
+++ b/plugins/docker/docker.plugin.zsh
@@ -38,7 +38,7 @@ fi
 
 {
   # `docker completion` is only available from 23.0.0 on
-  local _docker_version=$(docker version --format '{{.Client.Version}}' 2>/dev/null)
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks!

Haltarys:
Always happy to contribute ! :)

PR #11771, opened 2023-06-23
comment out bindkey -e

rwmitchell:
vi vs emacs is a personal choice.  If the user sets bindkey -v before sourcing OMZ, it gets undone.  if the user sets bindkey -v after sourcing OMZ, then all of these other settings get undone, making this file useless.

bindkey -e is the default in zsh, it doesn't need to be set.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/lib/key-bindings.zsh
+++ b/lib/key-bindings.zsh
@@ -15,8 +15,8 @@ if (( ${+terminfo[smkx]} )) && (( ${+terminfo[rmkx]} )); then
   zle -N zle-line-finish
 fi
 
-# Use emacs key bindings
-bindkey -e
+# if you want emacs bindings, set this in YOUR .zshrc file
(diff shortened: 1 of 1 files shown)


rwmitchell:
Then you need to find a way to not set it thst still makes you happy. This is breaking my workflow. I don’t use emacs, there is no benefit to those key bindings. You are overriding the choice I have already set, that is wrong.

rwmitchell:
What this is doing is wrong. You’re saying you’re never fixing bugs because someone’s workflow may now depend on them?

carlosala:
We are not going to change this at this point Richard, I told you the reasons. There's a plugin to use `vi-mode`, you are free to choose between both, but default `omz` is emacs.

PR #11773, opened 2023-06-24
feat(alias-finder): add cheaper option and apply zstyle options

roeniss:
- resolves https://github.com/ohmyzsh/ohmyzsh/issues/11624

## Standards checklist:

- [x] The PR doesn't replicate another PR that is already open.

## Changes:

> Before starting, I want to clarify the norms I will use.
> On `alias a=abc`, 'abc' is the source of the alias, and 'a' is the destination of the alias.
> If this naming seems inappropriate, please let me know. I couldn't find better names... 
> (This is important because the same namings are also written in [readme](https://github.com/roeniss/ohmyzsh/commit/6eb647d5a5b9f8dad3c5ab5b8d8060c65af546a5#diff-7a6e14abfc821561312a2d3217f3d66c89bb4237460daa7fb0c03ceaaffe84a6R17)) 

### 1. --cheaper/-c

- Add `--cheaper/-c` option: Shows only the aliases where the destination is shorter than the current input (command). This option could filter out the following example:

```sh
> alias g=git
> alias longerCmd=git
> alias-finder -c git
g=git
# longerCmd would be filtered out
>
```
- You might wonder how such an odd alias could exist: When I made an issue I had a very specific plausible case but now I forgot. I'm sorry about such a poor example above.
- At the beginning, I tried with `shorter`, but it doesn't make sense with names on existing options (longer, exact) because those are about the source of aliases, and my new option is about the destination of aliases. You can see a more detailed explanation (not much different) on [readme](https://github.com/roeniss/ohmyzsh/commit/6eb647d5a5b9f8dad3c5ab5b8d8060c65af546a5#diff-7a6e14abfc821561312a2d3217f3d66c89bb4237460daa7fb0c03ceaaffe84a6R19-R21)

### 2. Reformat codes

The original codebase was pretty efficient but hard to follow (maybe because of my lack of competency). So I add some comments, refactor the codebase, and flatten the flow.

### 3. Support zstyle

As suggested in [the original issue](https://github.com/ohmyzsh/ohmyzsh/issues/11624#issuecomment-1517966744), I added some zstyle options but still support previous env variable option for backward compatibility (also added TODO comments for them). Because the env variable is not preferable, I removed the relevant section from readme. 

## Other comments:

I tried `zunit` test framework and it was quite helpful. But none of the plugins seems not having testcode like that Should I remove it from this pr?

One last thing: I haven't made pr for a change like this for any open source before. So if I was wrong somewhere, please give me some advice on that. :D

### DIFF
--- a/plugins/alias-finder/.zunit.yml
+++ b/plugins/alias-finder/.zunit.yml
@@ -1,0 +1,9 @@
+tap: false
+directories:
+  tests: tests
+  output: tests/_output
+  support: tests/_support
+time_limit: 0
+fail_fast: false
+allow_risky: false
+verbose: true
--- a/plugins/alias-finder/README.md
+++ b/plugins/alias-finder/README.md
@@ -2,22 +2,32 @@
 
 This plugin searches the defined aliases and outputs any that match the command inputted. This makes learning new aliases easier.
 
+## Usage
+
 To use it, add `alias-finder` to the `plugins` array of your zshrc file:
 ```
 plugins=(... alias-finder)
 ```
 
-## Usage
-To see if there is an alias defined for the command, pass it as an argument to `alias-finder`. This can also run automatically before each command you input - add `ZSH_ALIAS_FINDER_AUTOMATIC=true` to your zshrc if you want this.
+To enable it for every single command, set zstyle in your `~/.zshrc`.
+
+```zsh
+# ~/.zshrc
+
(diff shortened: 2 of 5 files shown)


carlosala:
Hi @roeniss! Amazing changes! I'll review it (asking for review also to @mcornella)	as soon as possible and give you some feedback. Thanks 😊

carlosala:
Thanks a lot! Sorry for the time I took to review it, super great contribution!

PR #11777, opened 2023-06-25
feat(lib): update nvm library

dgvai:
## Standards checklist:

## Changes:

- checks for node version only in directories that contains node application
- new ZSH env variable for show shorter node version

## Other comments:

...


carlosala:
I'm not actually sure if we should only focus on `package.json`, imagine if you are in a sub-directory of a nodejs project, it'd not appear.
Thanks for your contribution, but I still think as it's now it's great!

PR #11780, opened 2023-06-28
docker: avoid `docker version` subcommand.

anpol:
`docker version` subcommand is able to return both docker client and docker daemon information.  To get a daemon version, it connects to a possibly remote daemon.  If the remote daemon is not accessible, the client waits for some time, until it gets interrupted by timeout.

As a result we can have a docker client running in background. When zsh session is rather short, a terminal application (iTerm2) starts asking if that background docker process should be stopped.

On the other hand, to get a docker client version only, we can use `docker --version` instead. It does not connect to a daemon.

relates to #11754

## Standards checklist:

### DIFF
--- a/plugins/docker/docker.plugin.zsh
+++ b/plugins/docker/docker.plugin.zsh
@@ -38,7 +38,7 @@ fi
 
 {
   # `docker completion` is only available from 23.0.0 on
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution, good one!

PR #11782, opened 2023-06-29
fix(jira): support identifiers delimited with a hyphen

matthiasheyman:
Our git uses branch names like feature/PROJECT-1234-some-description-of-the-branch. Using "jira branch" does not work here, since it does not detect the PROJECT-1234 tag, but uses the full branch name.
This fix first checks if there is more than one part, then splits it with "-" if there is only one part (that contains a "-").

## Standards checklist:

## Changes:

When using "jira branch", the code splitted the branch name on underscore delimiter. The PR now adds support for branch names delimited with a hyphen

## Other comments:


carlosala:
Thanks for your contribution!

matthiasheyman:
No problem!

PR #11784, opened 2023-06-30
Update outdated scw plugin and add new commands

albinou:
## Standards checklist:

## Changes:

The current `scw` plugin seems outdated. Indeed, the Scaleway `scw` command now supports auto-completion natively and last commit has been done 5 years ago.

Then, this PR:
- removes the legacy auto-completion support and uses the new `scw autocomplete` command for auto-completion
- adds a set of new functions largely inspired by the `aws` plugin:
  - `ssp` allows to set the current profile
  - `sgp` gets the current profile
  - `scw_profiles` lists the available profiles
  - `scw_upgrade` upgrade the `scw` binary

Don't hesite to read the `README.md` file for more detail.

## Other comments:

@moul: since you are the original author of the plugin, please don't hesitate to comment. Thanks for your original work :smiley:


carlosala:
There are a lot of changes here, we'd need some testers.

philibea (approved):

PR #11794, opened 2023-07-12
feat(git): add alias gdh for `git diff HEAD`

crazyzh1984:
## Standards checklist:

## Changes:

- Add `gdh` alias to git plugin for `git diff HEAD`

## Other comments:

- Common use case is to quickly check all changes (staged & not staged) that have made of same file before add
- For example: check merge result of resolve conflicts before add


adamchainz:
Git has the `@` alias for `HEAD`. You can type `gd @` for this, which is already pretty short, so I'm not sure an extra alias is worth it.

crazyzh1984:
`gd @` is pretty short

PR #11797, opened 2023-07-15
Make it more clear that autoload option is required with silent-autol…

### LINKED ISSUE #11796: Doesn't autoload when with `silent-autoload`
When I have `zstyle ':omz:plugins:nvm' autoload yes` it works fine, but if I use `zstyle ':omz:plugins:nvm' silent-autoload yes` as shown in the documentation, it does not autoload. I'm using Powerlevel10k if that matters.

https://github.com/ohmyzsh/ohmyzsh/blob/fe4b5659863c388786986d70fa6d1bb66b00afb6/plugins/nvm/README.md?plain=1#L60

jclusso:
Closes #11796


carlosala:
Thanks!

PR #11820, opened 2023-07-30
Reordered #List directory contents

o0nd7ots:
## Standards checklist:

## Changes:

It is more readable now


carlosala:
I'd say this change doesn't make a big impact. We don't follow this kind of formatting, such as aligning dashes, words, and these kind of things.
Thanks for your contribution!

PR #11826, opened 2023-08-04
Add git rebase with origin develop

rklak:
Add missing git alias to git plugin:
 git rebase with origin develop


carlosala:
We already have an extreme amount of `git rebase` aliases.
In most cases, this can be achieved with `grbd`. In those that you want to rebase, for example, your own `develop` on top of `origin/develop` (which makes sense, but I wouldn't say it's the most common thing since you'd typically want to create your own branches from `develop`), you could add a personal alias if that's something that happens often.
Thanks for your contribution!

PR #11831, opened 2023-08-07
fix(colored-man-pages): add env GROFF_NO_SGR=1

ahmedwael216:
## Standards checklist:

## Changes:
The `colored-man-pages` plugin stopped working after this bug see
[this](https://unix.stackexchange.com/questions/746094/why-dont-these-less-termcap-variables-work-on-oracle-linux)
and [this](https://bugzilla.redhat.com/show_bug.cgi?id=1028764)
the fix was just to add
`export GROFF_NO_SGR=1` to the plugin

### DIFF
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -15,7 +15,6 @@ less_termcap[se]="${reset_color}"
 # underlining
 less_termcap[us]="${fg_bold[green]}"
 less_termcap[ue]="${reset_color}"
-export GROFF_NO_SGR=1
 
 # Handle $0 according to the standard:
 # https://zdharma-continuum.github.io/Zsh-100-Commits-Club/Zsh-Plugin-Standard.html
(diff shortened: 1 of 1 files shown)


plugins/colored-man-pages/colored-man-pages.plugin.zsh
```diff
@@ -15,6 +15,7 @@ less_termcap[se]="${reset_color}"
 # underlining
 less_termcap[us]="${fg_bold[green]}"
 less_termcap[ue]="${reset_color}"
+export GROFF_NO_SGR=1
```
bretello, line 18: Although this patch solves the issue, exporting this here will change the environment for the user, which is something that might be unexpected, a solution which doesn't change the environment follows:

```diff
diff --git a/plugins/colored-man-pages/colored-man-pages.plugin.zsh b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
index 981992d8..57facbb5 100644
--- a/plugins/colored-man-pages/colored-man-pages.plugin.zsh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.zsh
@@ -36,6 +36,7 @@ function colored() {
   # Prefer `less` whenever available, since we specifically configured
   # environment for it.
   environment+=( PAGER="${commands[less]:-$PAGER}" )
+  environment+=( GROFF_NO_SGR=1 )
 
   # See ./nroff script.
   if [[ "$OSTYPE" = solaris* ]]; then
```

bretello (changes requested):

carlosala:
Thanks for your contribution!

PR #11834, opened 2023-08-09
adding plugin zsh-autocomplete

compilerbyte:
## Standards checklist:

## Changes:

- adding zsh-autocomplete

## Other comments:

...


carlosala:
Hi! We do not add submodules to `oh-my-zsh`, and `zsh-autocomplete` is one of those plugins that's better to have in `custom/`.
Thanks for your contribution!

PR #11847, opened 2023-08-15
feat(systemd): add alias for --failed commands

bretello:
## Standards checklist:

## Changes:

Adds `alias sc-failed=systemctl --failed` to list failed systemd units

### DIFF
--- a/plugins/systemd/README.md
+++ b/plugins/systemd/README.md
@@ -12,6 +12,7 @@ plugins=(... systemd)
 
 | Alias                  | Command                            | Description                                                      |
 |:-----------------------|:-----------------------------------|:-----------------------------------------------------------------|
+| `sc-failed`            | `systemctl --failed`               | List failed systemd units                                        |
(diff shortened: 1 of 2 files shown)


robbyrussell:
@bretello  thanks for this, I see that you have two aliases added here. Could you update the README to include both of those? (only see one referenced at the moment)

birmehto:
cool

bretello:
> @bretello thanks for this, I see that you have two aliases added here. Could you update the README to include both of those? (only see one referenced at the moment)

Hi @robbyrussell, thanks for reviewing this.

Is this really required? Quoting the README:

https://github.com/ohmyzsh/ohmyzsh/blob/9ef881aa2829ae9e01a6426413cbb4d8901dfb2d/plugins/systemd/README.md?plain=1#L52-L56

bretello:
Anything else blocking this? @robbyrussell

carlosala:
Thanks!

PR #11851, opened 2023-08-20
refactor(init): rename variable

LuckyWindsck:
The variable `custom_config_file` is no longer exist, so `unset` it does nothing. Instead, this patch rename the variable from `config_file` to `lib_file`, and unset it immediately.

## Standards checklist:

## Changes:

- Rename the variable from `config_file` to `lib_file`
  - This increased readability by distinguish between the variable for Oh My Zsh lib file and the variable for custom configuration.
  - This solved the problem of `unset` a variable that no longer exist.

## Other comments:

...

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -187,12 +187,12 @@ _omz_source() {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #11858, opened 2023-08-25
feat(poetry): add aliases

baltic-tea:
## Standards checklist:

## Changes:

- Add aliases for the popular (and not so popular) poetry commands


jeroeness:
This PR is already quite old, why is this one not merged yet. I am actually waiting for this. 


While this PR is not merged yet. I would suggest using even shorter aliases (inspired by the [git aliases plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/git) which work like a charm for me).

Because I use poetry `add`, `rm`, `run`,  `update`, `install`, `lock`, `shell` and `build` the most frequent I would suggest to use shorter aliases them to `pa`, `prm`, `pr`, `pu`, `pi`, `pl`, `psh`, `pb` respectively. Based on these examples we could extrapolate the rest. What do you think about this idea @baltic-tea

robbyrussell:

Cheers,
Robby

PR #11865, opened 2023-08-30
feat(web-search): add `youtube` alias

t-aswath:
## Standards checklist:

## Changes:

- added `alias` for `youtube` instead of the duckduckgo `!bang` search

## Other comments:

!bang search for youtube using duckduckgo takes some time due to the `redirection` from duckduckgo to youtube also this makes  use of duckduckgo without the user concern which does not affect anything but still people might not like it.

with this change people can search youtube results using their primary search engine without any redirection or delay

Screenshot from 2023-08-30 14-57-28
Screenshot from 2023-08-30 14-58-52

also allowing user to use duckduckgo !bang search by using
Screenshot from 2023-08-30 19-19-38

### DIFF
--- a/plugins/web-search/web-search.plugin.zsh
+++ b/plugins/web-search/web-search.plugin.zsh
@@ -25,6 +25,7 @@ function web_search() {
     archive         "https://web.archive.org/web/*/"
     scholar         "https://scholar.google.com/scholar?q="
     ask             "https://www.ask.com/web?q="
+    youtube         "https://www.youtube.com/results?search_query="
   )
 
   # check whether the search engine is supported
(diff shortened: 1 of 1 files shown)


t-aswath:
Thanks for accepting my PR . This is my first ever open source contribution

carlosala:
Great to have you onboard! I encourage you to keep contributing to OSS💪🏻🎉

plugins/web-search/web-search.plugin.zsh
```diff
@@ -25,6 +25,7 @@ function web_search() {
     archive         "https://web.archive.org/web/*/"
     scholar         "https://scholar.google.com/scholar?q="
     ask             "https://www.ask.com/web?q="
+    youtube         "https://www.youtube.com/results?search_query="
```
mpagot, line 28: Does https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/web-search/README.md needs to be updated too?

carlosala: It should be, yes!

carlosala: Fixed in c91684bb1b723833a96415c2c724332a6e2e620f
Thanks for the report!

PR #11868, opened 2023-08-31
fix(termsupport): match foot* extra shipped terminfo files

trinitronx:
## Standards checklist:

- [N/A] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- fix(termsupport): match foot* extra shipped terminfo files

## Other comments:

- `foot` terminal ships `foot-extra` and/or `foot-extra-direct` terminfo files on many distros.
  - `TERM` is set accordingly (e.g. `TERM=foot-extra`),
    and would not otherwise match simply "`foot`",
    therefore disabling OMZ's features when in fact they are supported.
  - Reference:
    - https://codeberg.org/dnkl/foot/wiki\#foot-s-terminfo-vs-ncurses-terminfo

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -17,7 +17,7 @@ function title {
   : ${2=$1}
 
   case "$TERM" in
-    cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty|st*|foot|contour*)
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution, good catch!

trinitronx:
> Thanks for your contribution, good catch!

No problem!  Happy to help 😄

PR #11869, opened 2023-08-31
feat(git): add bisect aliases `gbsn` and `gbso`

adamchainz:
## Standards checklist:

## Changes:

Add two aliases for the [`git bisect` subcommands](https://git-scm.com/docs/git-bisect) `old` and `new`. These are alternative labels to the original `good` and `bad`, which are applicable in more situations. For example, when searching for a commit that accidentally *fixed* a bug, the meanings of “good” and “bad“ are confusingly reversed, whilst “old“ and “new“ remain clear.

## Other comments:

NOne.


carlosala:
I'm not quite sure that these aliases are quite relevant, actually. They are exactly the same as `good/bad`, and it's already kind of a standard to use `good/bad`, even though it makes complete sense what you say.
Taking it into account, it makes much more sense to me to have these kinds of aliases in your personal config, in case the semantics of `gbsg` and `gbsb` are not enough for you, or anyone in the same situation.
What do you kind about it?

adamchainz:
I think that if `git bisect` were created from scratch, there would only be `old` and `new`. There’d be no need for `good` and `bad`, since they’re specific only to finding new bugs.

I’m hoping to encourage using `old` and `new` by default by teaching them first in my new book. I also thought this PR could help “spread the word”.

carlosala:
Could you update the README accordingly?

carlosala:
Thanks for your contribution!

adamchainz:
Brilliant, thanks!

PR #11870, opened 2023-09-01
feat(termsupport): configurable terminal title patterns & cmdline truncate

trinitronx:
## Standards checklist:

- [N/A] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- Configurable terminal title patterns & cmdline truncation

## Other comments:

Allow user to set environment variables to override default terminal title patterns, and command line truncation char limit.  Fallback to original defaults if any are unset.

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -38,11 +38,15 @@ function title {
   esac
 }
 
-ZSH_THEME_TERM_TAB_TITLE_IDLE="%15<..<%~%<<" #15 char left truncated PWD
-ZSH_THEME_TERM_TITLE_IDLE="%n@%m:%~"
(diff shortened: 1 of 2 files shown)


mcornella:
Hello James, thank you for the contribution.

I've reviewed the PR and I think at this stage we are not looking at getting at such a fine-grained level of customization. Doing so requires a level of hacky-ness that I don't really feel worth it or stylish, and is very coupled with the title logic.

We'll be looking in the future to revise this decision to see if we can improve this somehow. In the meantime, you can redefine the functions yourself in your zshrc file, or disable our solution by setting `DISABLE_AUTO_TITLE=true` and fully customize your logic while keeping compatibility with the `title` function.

Thanks again for your contribution

PR #11872, opened 2023-09-02
fix(battery-plugin): add support for OpenBSD

### LINKED ISSUE #11857: Battery Plugin: support for OpenBSD

Battery plugin

Extends battery plugin support for OpenBSD

The battery plugin has support for FreeBSD through the sysctl hw.acpi* commands. But similar approach can be taken to extend support for OpenBSD using the [hw.sensors.acpibat*](https://man.openbsd.org/acpibat.4) commands.

But in OpenBSD the [apm](https://man.openbsd.org/apm.8) tool is more user friendly and works out of the box unlike the former which requires us to calculate percentage from power usage.

_No response_

### Additional context

_No response_

### Related Issues

_No response_

im-notpua:
Fixes #11857 

## Standards checklist:

## Changes:

- Uses the built-in `apm` ([OpenBSD manual](https://man.openbsd.org/apm.8)) command to fetch the battery status and details and adds OpenBSD support.

## Other comments:
I am a bit confused about the `battery_level_gauge` function. It seems like the main function doing most work but the [README](https://github.com/ohmyzsh/ohmyzsh/tree/ccce2e1cfdf5b9680f691a402e288d9cf6ce272a/plugins/battery) just recommends the usage of `battery_pct_prompt` function which only shows the percentage.

Other than that, this is a very useful plugin.

### DIFF
--- a/plugins/battery/battery.plugin.zsh
+++ b/plugins/battery/battery.plugin.zsh
@@ -13,6 +13,10 @@
 # Author: Avneet Singh (kalsi-avneet)     #
 # Modified to add support for Android     #
 ###########################################
+# Author: Not Pua (im-notpua)             #
+# Modified to add support for OpenBSD     #
+###########################################
+
 
 if [[ "$OSTYPE" = darwin* ]]; then
   function battery_is_charging() {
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #11874, opened 2023-09-04
fix(init): exit gracefully if on non-zsh emulation mode

### LINKED ISSUE #11686: Better handling of ZSH in emulation mode

_No response_

ZSH has the ability to emulate bash/ksh/sh either by the `emulate` command or by argv[0] starting with sh, ksh, or bash. This causes omz to error constantly because the modules are not automatically loaded, the syntax is different, and various other issues.

When `oh-my-zsh.sh` is loaded with a non-ZSH shell, it gives a nice error message, but when zsh is in emulation mode and the script is manually loaded (since in emulation mode it won't load .zshrc) you get
```
$ exec -a bash zsh -l
$ . .oh-my-zsh/oh-my-zsh.sh
.oh-my-zsh/tools/check_for_upgrade.sh:13: command not found: zstyle
(anon):40: command not found: zstyle
compinit:141: parse error: condition expected: $1
_omz_source:10: command not found: zstyle
_omz_source:15: bad set of key/value pairs for associative array
```

A simple check for `[ "$(emulate)" = zsh ]` could give a nicer error message.

If it is detected when the script is loaded, it could say
```
Error: Oh My Zsh can't be loaded in emulation mode.
This can happen if the emulate command is run, or zsh is renamed to sh, bash, or ksh.
```

Alternatively it could try to manually escape emulation mode, but that breaks other things.

### Additional context

_No response_

### Related Issues

_No response_

mcornella:
## Standards checklist:

## Changes:

- Skips loading Oh My Zsh if current zsh is in non-zsh emulation mode

## Other comments:

Fixes #11686

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -1,14 +1,14 @@
+# ANSI formatting function (\033[<code>m)
+# 0: reset, 1: bold, 4: underline, 22: no bold, 24: no underline, 31: red, 33: yellow
+omz_f() {
+  [ $# -gt 0 ] || return
+  IFS=";" printf "\033[%sm" $*
+}
+# If stdout is not a terminal ignore all formatting
+[ -t 1 ] || omz_f() { :; }
+
 # Protect against non-zsh execution of Oh My Zsh (use POSIX syntax here)
 [ -n "$ZSH_VERSION" ] || {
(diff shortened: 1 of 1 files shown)


mcornella:
Test for #11686: 
image

carlosala (approved):
LGTM! 💪🏻

PR #11883, opened 2023-09-09
fix(git-commit): fix compatibility of alias functions

### LINKED ISSUE #11882: Conventional commit aliases from `git-commit` plugin fail due to non POSIX syntax
### Describe the bug

[Git uses `sh` as its default shell, with no possibility to change it.](https://stackoverflow.com/questions/73162225/how-to-make-bash-as-default-shell-in-git-alias)

The recently created `git-commit` plugin for conventional commits adds git aliases which contain `[[` to check for conditions.
[This is non POSIX syntax](https://askubuntu.com/a/1277981) that `sh` does not recognise (but `bash` and `zsh` do.) One should use `[` instead when working with `sh`.

Offending line in the plugin:
https://github.com/ohmyzsh/ohmyzsh/blob/079dbff2c4f22935a71101c511e2285327d8ab68/plugins/git-commit/git-commit.plugin.zsh#L3

### Steps to reproduce

1. Add the plugin `plugins+=(git-commit)`
2. Reload with `omz reload`
3. `cd` into a git repository
4. Run an alias: `git feat --scope "zshrc/git-commit" "add plugin for conventional commits"`
5. Error: `error: pathspec 'something' did not match any file(s) known to git`

### Expected behavior

The alias should work properly.

Terminal output:
```zsh
$ git feat --scope "zshrc/git-commit" "add plugin for conventional commits"
a() { if [[ "$1" == "-s" || "$1" == "--scope" ]]; then git commit -m "feat(${2}): ${@:3}"; else git commit -m "feat: ${@}"; fi }; a: 1: [[: not found
a() { if [[ "$1" == "-s" || "$1" == "--scope" ]]; then git commit -m "feat(${2}): ${@:3}"; else git commit -m "feat: ${@}"; fi }; a: 1: --scope: not found
error: pathspec 'zshrc/git-commit' did not match any file(s) known to git
error: pathspec 'add plugin for conventional commits' did not match any file(s) known to git
```

### OS / Linux distribution

Ubuntu 20.04

### Zsh version

5.8

### Oh My Zsh version

master (079dbff)

### Terminal emulator

Windows Terminal (1.17.11461.0)

WSL2

### Additional context

I suspect that the creator of this plugin set up a symlink to point `sh` to `bash` on their machine and didn't realise that this was not the default for most people. Will open a PR to fix it.

Haltarys:
## Standards checklist:

## Changes:

- fix(git-commit): fix non-posix condition and array in alias

## Other comments:

Closes #11882

### DIFF
--- a/plugins/git-commit/git-commit.plugin.zsh
+++ b/plugins/git-commit/git-commit.plugin.zsh
@@ -1,6 +1,6 @@
 function _git_commit_register {
   if ! git config --global --get-all alias.$1 >/dev/null 2>&1; then
-    git config --global alias.$1 '!a() { if [[ "$1" == "-s" || "$1" == "--scope" ]]; then git commit -m "'$1'(${2}): ${@:3}"; else git commit -m "'$1': ${@}"; fi }; a'
+    git config --global alias.$1 '!a() { if [ "$1" = "-s" ] || [ "$1" = "--scope" ]; then local scope="$2"; shift 2; git commit -m "'$1'(${scope}): ${@}"; else git commit -m "'$1': ${@}"; fi }; a'
   fi
 }
 


plugins/git-commit/git-commit.plugin.zsh
```diff
@@ -1,6 +1,6 @@
 function _git_commit_register {
   if ! git config --global --get-all alias.$1 >/dev/null 2>&1; then
-    git config --global alias.$1 '!a() { if [[ "$1" == "-s" || "$1" == "--scope" ]]; then git commit -m "'$1'(${2}): ${@:3}"; else git commit -m "'$1': ${@}"; fi }; a'
+    git config --global alias.$1 '!a() { if [ "$1" = "-s" ] || [ "$1" = "--scope" ]; then local scope="$2"; shift 2; git commit -m "'$1'(${scope}): ${@}"; else git commit -m "'$1': ${@}"; fi }; a'
```
mcornella, line 3: The use of `$1` after `shift 2` is not going to work, it should also be captured first, like `$scope`. Can you fix that?

mcornella (changes requested):

plugins/git-commit/git-commit.plugin.zsh
mcornella: Nevermind, misread the code. `$1` is the one from `_git_commit_register`. LGTM

mcornella (approved):

PR #11884, opened 2023-09-11
feat(mortalscumbag): show toolbox info

strugee:
## Standards checklist:

## Changes:

- Show the Toolbox icon in mortalscumbag if relevant. This brings the theme in line with the default Fedora Silverblue bash prompt.

## Other comments:

I tested this with and without `toolbox` enabled.

/cc @carlosala

### DIFF
--- a/themes/mortalscumbag.zsh-theme
+++ b/themes/mortalscumbag.zsh-theme
@@ -51,8 +51,14 @@ function ssh_connection() {
   fi
 }
 
+function _toolbox_prompt_info() {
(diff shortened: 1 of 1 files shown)


strugee:
Hrm. Honestly, might be better for `toolbox` to just patch `PROMPT/RPROMPT`...

carlosala:
We prefer not to modify PROMPT or RPROMPT because some users might be using other themes and are only interested in the aliases.

PR #11887, opened 2023-09-11
feat(brew): support nonstandard install locations

curiousstranger:
Allows BREW_LOCATION to be passed in to the plugin. Checks that BREW_LOCATION is executable.

## Standards checklist:

## Changes:

- Adds a test to see if BREW_LOCATION is set and executable after checking the standard install locations.  This allows a user to set BREW_LOCATION before loading the plugin to support initializing the environment from a non-standard install location.

@voronkovich


curiousstranger:
Resubmitting with clean history

PR #11888, opened 2023-09-11
feat(brew): support nonstandard install locations

curiousstranger:
## Standards checklist:

## Changes:

- Allows BREW_LOCATION to be passed in to the plugin. Checks that BREW_LOCATION is executable.

## Other comments:

@voronkovich


carlosala:
I see, it makes more sense now (taking into account the MANPATH stuff). I pushed a commit, you should be able to set `BREW_LOCATION` now.
Thanks for your contribution!

PR #11895, opened 2023-09-14
docs(git): improve tables and language

adamchainz:
## Standards checklist:

- [-] The code is efficient, to the best of my ability, and does not waste computer resources.
- [-] The code is stable and I have tested it myself, to the best of my abilities.
- [-] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

1. Use backticks to place all aliases in monospaced font.
2. Change table syntax to use non-aligned `|` column markers, with just one space on either side of each cell’s contents. This avoids lines being wide except where needed. It also makes it easier to edit the table without introducing broken alignment, like I did before as fixed in #11878.
3. Fix the display of some aliases. For example, `gbgd` and `gbgD` weren't showing the whole aliased command due to unescaped `|` characters (interpreted by Markdown as a column marker).
4. Improve some wording.

## Other comments:

This will cause merge conflicts with open PRs, but they will be fairly easy to fix, editing the table contents to meet the new style.

I tested the change by scanning through each table row, checking it was rendering correctly.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -10,204 +10,204 @@ plugins=(... git)
 
 ## Aliases
 
-| Alias | Command |
-| :---- | :------ |
-| `grt` | `cd "$(git rev-parse --show-toplevel \|\| echo .)"` |
-| `ggpnp` | `ggl && ggp` |
-| `ggpur` | `ggu` |
-| `g` | `git` |
-| `ga` | `git add` |
-| `gaa` | `git add --all` |
-| `gapa` | `git add --patch` |
-| `gau` | `git add --update` |
-| `gav` | `git add --verbose` |
-| `gwip` | `git add -A; git rm $(git ls-files --deleted) 2> /dev/null; git commit --no-verify --no-gpg-sign --message "--wip-- [skip ci]"` |
-| `gam` | `git am` |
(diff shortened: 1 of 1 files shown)


adamchainz:
@carlosala , please can you review?

Also, I checked and found that 93 PRs are open against the Git plugin, some going back a way. How can I best help triage these?

carlosala:
Hi Adam! Thanks for all the work triaging `git` plugin stuff, it's really appreciated! :)
Normally, we use [`prettier`](https://github.com/prettier/prettier/) to format the code in markdown files. I like the other improvements though, could you please add a commit running that tool? Besides that, it's super great, thanks a lot!

adamchainz:
Done. That undoes my table change, but I guess if the alignment can be automated that’s fine.

Could we set up automatic execution of prettier? The [pre-commit](https://pre-commit.com/) tool is good for this, as it sets up tools for users as well as runs them on all files. The related CI service, [pre-commit.ci](https://pre-commit.ci/) is free for open source projects and automatically pushes commits with fixes. I covered pre-commit a lot in my book!

carlosala:
Thanks, Adam! You're boosting the `git` plugin so much. We really appreciate your contributions :)

PR #11896, opened 2023-09-14
feat(git): add reset aliases `grhk` and `grhs`

adamchainz:
⚠️ This PR is built on top of #11895, which needs reviewing first.

## Standards checklist:

## Changes:

Add two new aliases for `git reset`:

### 1. `grhk` = `git reset --keep`

`--keep` is similar to `--hard`, with an important data-preserving difference. Like `--hard`, it moves the current branch to point to a named commit, but it does not throw away uncommitted changes. I would say this is generally a better idea as uncommitted changes are precious—Git provides no way to retrieve them, since Git tracks files *within* commits.

`--keep` is not well known, perhaps due to the obtuse description [in the docs](https://git-scm.com/docs/git-reset#Documentation/git-reset.txt---keep). I’m trying to change that and promote it.

### 2. `grhs` = `git reset --soft`

`--soft` undoes commits, and preserves their changes. It’s similar to the default mode, `--mixed`, but keeps the changes staged:

```
$ git reset --soft @~

$ git status
On branch commentary
Changes to be committed:
    modified:   hamlet.txt
    modified:   macbeth.txt

Changes not staged for commit:
    modified:   othello.txt
```

This means you can tweak the changes before committing them again. With the default mode, you have to re-add changes before making another commit.

## Other comments:

Closes #9398


adamchainz:
After merging this, #9398 could be closed, which proposed `grhs` but with `HEAD~1` appended.

carlosala:
Thanks for your contribution!

adamchainz:
💪 yesss

PR #11897, opened 2023-09-15
fix(git): remove `--text` from `gsts`

adamchainz:
⚠️ This PR builds on top of #11895, which should be merged first

## Standards checklist:

## Changes:

Remove the [`--text` diff option](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--a) from the `gsts` alias for `git stash show`. `--text` forces treating binary files as text, causing their contents to be output and potentially breaking terminal rendering.

For example, here's what `git stash show` outputs with and without `--text` for a PNG file.

First, without:

```
$ git stash show
 donut.png | Bin 0 -> 238137 bytes
 1 file changed, 0 insertions(+), 0 deletions(-)
```

Plain stat details, and no attempt to represent the binary file. (Representation can be added by setting up a textconv filter.)

Second, with `--text`:

```
$ git stash show --text
diff --git donut.png donut.png
new file mode 100644
index 0000000..be8dd81
--- /dev/null
+++ donut.png
@@ -0,0 +1,938 @@
+<89>PNG^M
+^Z
...
```

The output continues with the remainder of the PNG's bytes. This is never useful behaviour since even if you could read the binary data, it isn't well displayed. On some terminal setups, random bytes can break rendering.

## Other comments:

The `--text` option was included with the creation of the alias, way back in #1864, without any discussion.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -183,7 +183,7 @@ plugins=(... git)
 | `gstp`                 | `git stash pop`                                                                                                                                                     |
 | `gsta`                 | On Git >= 2.13: `git stash push`                                                                                                                                    |
 | `gsta`                 | On Git < 2.13: `git stash save`                                                                                                                                     |
-| `gsts`                 | `git stash show --text`                                                                                                                                             |
+| `gsts`                 | `git stash show`                                                                                                                                                    |
 | `gst`                  | `git status`                                                                                                                                                        |
 | `gss`                  | `git status --short`                                                                                                                                                |
 | `gsb`                  | `git status --short -b`                                                                                                                                             |
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -331,7 +331,7 @@ alias gstp='git stash pop'
 is-at-least 2.13 "$git_version" \
   && alias gsta='git stash push' \
   || alias gsta='git stash save'
-alias gsts='git stash show --text'
+alias gsts='git stash show'
 alias gst='git status'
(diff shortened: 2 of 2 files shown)


carlosala:
I like it! Could you resolve the conflicts? Thanks :)

adamchainz:
Resolved!

carlosala:
Thanks for your contribution!

MarijnS95:
`--text` did however show patch differences for changes in _text_ files, which is missing now. This should perhaps use `-p` to show patch differences whenever there is a change in text files, without expanding binary differences, to more-or-less maintain compatibility with the original use of the `alias`. This also didn't appear to be considered/discussed in this PR.

MarijnS95:
@carlosala `git show` != `git stash show`.

From the `man git-stash` page under `show`:

```
By default, the command shows **the diffstat**, but it will accept any format known to git diff (**e.g., git stash
           show -p stash@{1}** to view the second most recent entry **in patch form**). If no <diff-option> is provided, the default behavior will be given by the stash.showStat, and stash.showPatch config variables. You can also use stash.showIncludeUntracked to set
           whether --include-untracked is enabled by default.
```

(Emphasis mine)

Guess I will either override the `alias` or set `stash.showPatch` to get back the old behaviour.

carlosala:
Oh, I misread `gsps` instead of `gsts`.
I see the point, then! Even though I think `gsts` (in terms of semantics and stuff) is more configurable and gives the user the option to choose what they want each moment.
`gsts -p` might make sense in some situations!

PR #11901, opened 2023-09-20
fix(gcloud): source `path.zsh.inc` always

marcantoinegodde:
## Standards checklist:

## Changes:

This PR removes the condition that checks if `gcloud` command is already available before sourcing the path file.

## Why:

If gcloud SDK was installed with brew, the binary is already symlinked in the folder `/opt/homebrew/bin` which is by default in the PATH. This means that the condition is never verified and therefore, the file never sourced. This causes an issue when you need to connect on GKE clusters for instance. You would need to use the `gke-gcloud-auth-plugin` binary which won't be available because not in the PATH.


carlosala:
Thanks for your contribution!

PR #11902, opened 2023-09-20
feat(help) Add help plugin

a-b:
## Standards checklist:

## Changes:

- add 'help` command

## Other comments:
This feature matches bash `help echo` command that prints help for the internal shell commands.
Let me know if you want me to move this from the plugin to the stl.

### DIFF
--- a/plugins/help/README.md
+++ b/plugins/help/README.md
@@ -1,0 +1,14 @@
+# help plugin
+
+This plugin load zsh `run-help`and link it to `help`.
+
+To enable, add `help` to your `plugins` array in your zshrc file:
+
+```zsh
+plugins=(... help)
+```
+
+## Description
+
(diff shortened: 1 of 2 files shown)


carlosala:
Hi! I'd say this alias is interesting, but `run-help` is most of the times aliased to `man` (according to zsh documentation).
Executing `man <cmd>` seems much shorter!
Thanks for your contribution!

a-b:
Can you please provide reference to the documentation?

carlosala:
The idea for `run-help` is to be used by the ZLE (keybindings and stuff).
You can, for example, type `git` in your terminal (no click enter), and then click <Esc> <h>.
You'll enter help. The documentation says literally `run-help is normally aliased to man.`

PR #11904, opened 2023-09-21
feat(dnf): use `dnf5` if available

marcusmueller:
## Standards checklist:

## Changes:

- Fedora 39 or 40 moves from classic dnf to dnf5, which makes dependency
solving, package cache updates etc. faster. It's desirable to use it now
if available.

## Other comments:

No new aliases were added.


carlosala:
Thanks for the contribution!

PR #11909, opened 2023-09-25
feat(git): add branch rename alias `gbm`

adamchainz:
## Standards checklist:

## Changes:

Re-do of #10748 without merge conflicts. Renaming branches is a pretty common operation, and `gbm` is an alias I have also added myself, so I think this one would be useful.
Closes #10748

On the other hand, one *can* already use `gb -m`, which isn’t much more typing. So I’d understand deciding against this one.

## Other comments:

Attributed the original author @otilor.


carlosala:
This is an alias that makes sense, I'd say. It's one of the core stuff of git.
Let's merge this one, but it's true that we can do it easily with a kinda short command.
Thanks for your contribution!

otilor:
Thanks guys. @adamchainz @carlosala 
This was finally merged lol.

PR #11910, opened 2023-09-25
feat(git): add tag aliases `gt` and `gta`

adamchainz:
## Standards checklist:

## Changes:

Add two aliases:

1. `gt` for `git tag` (suggested in #10747).

   This is useful for listing tags and using other `git tag` options.

   I searched Ubuntu manpages for `gt` and found [this tool](https://manpages.ubuntu.com/manpages/lunar/en/man1/gt.1.html), “gt - The GenomeTools genome analysis system.“ So there will be a clash between this alias and that tool. It’s unlikely *many* users will be doing genome analyses, but perhaps the existence in a major package manager is enough reason to avoid the alias?

2. `gta` for `git tag --annotate` (suggested in #10854).

   This makes an “annotated tag”, the modern tag format that includes a message, the author details, and timestamp. Annotated tags should nearly always be used, so we can help this with an alias.

   It’s also a bit weird we already have `gts` for `git tag --sign`, which is an extended form of annotated tag with GPG-signing. I would think GPG-signing is much less common than unsigned tags.
  
## Other comments:

Closes #10747 
Closes #10854


adamchainz:
@carlosala I extracted these two from existing PRs, what do you think? Is the alias clash an acceptable risk? If not we could find an alternative.

carlosala:
It makes much sense to have something for tags, we literally have no alias.
Thanks for your contribution and grouping up all the commits!

goldjacobe:
Hi, @carlosala — is the `git tag` change set in stone? I'm an engineer at [Graphite](https://graphite.dev) and `gt` is the name of our CLI's executable — we've already had a few users bitten by this change.  If we opened a PR to remove the alias, would you be open to considering it?

carlosala:
@goldjacobe yep, let's go for it. `gtg` and `gtga` would do the trick

carlosala:
@goldjacobe @adamchainz I removed the alias temporarily in 0e1e877ea9182f87b3ab34cbc2ea944b34f5345e. We'll see if we find another alias for it 👍🏻

adamchainz:
Totally fair. Let’s avoid a `git tag` alias for now, it is a short command and the `gt*` space sounds crowded with tools.

PR #11912, opened 2023-09-26
feat(git): add submodule alias `gsur`

lukas-holzner:
## Standards checklist:

## Changes:

`gsur` for `git submodule update --remote`

This pulls the remote HEAD for the submodules in the repository. When working with remote submodules, this feature is often used. I researched if there is any other way to update the remote submodules with the existing aliases, but I didn't find any.

## Other comments:

...


carlosala:
Hi! Updating a submodule to the latest remote seems a bit niche or not a common thing. Doing `gsu --remote` is not super long, considering that it is not a daily thing.
I have that alias in my config because I do that often, but it should be rare to the common user.
Similar to https://github.com/ohmyzsh/ohmyzsh/pull/10071#issuecomment-1734997161.
Thanks for the contribution!

PR #11917, opened 2023-09-28
fix(gradle): exclude lookup for settings.gradle* files for `gradle` function

### LINKED ISSUE #11916: Gradle plugin should not lookup `settings.gradle` and `settings.gradle.kts`
Consider the following case:

```
gradle/
  wrapper/
plugin/
  build.gradle
  settings.gradle
build.gradle
gradlew
gradlew.bat
settings.gradle
```

```
# settings.gradle
pluginManagement {
    includeBuild('plugin/')
}
```

In the above case, it is not possible to call the `gradle` alias when I'm in the `plugin` directory, for the reason that the script breaks the search for the `gradlew` script because it has found the `settings.gradle` file. In the following condition, the script should only check for the existence of the `gradlew` file and possibly whether it has the executable bit set, but nothing more.

https://github.com/ohmyzsh/ohmyzsh/blob/0e1e877ea9182f87b3ab34cbc2ea944b34f5345e/plugins/gradle/gradle.plugin.zsh#L9

Greetings
Piotr Minkina

piotrminkina:
Now `gradle` alias works properly with composite builds inside project

## Standards checklist:

## Changes:

- removed lookup of `settings.gradle` and `settings.gradle.kts` files to work with composite builds inside project

## Other comments:

Fixes #11916

### DIFF
--- a/plugins/gradle/gradle.plugin.zsh
+++ b/plugins/gradle/gradle.plugin.zsh
@@ -6,7 +6,7 @@ function gradle-or-gradlew() {
   # taken from https://github.com/gradle/gradle-completion
   local dir="$PWD" project_root="$PWD"
   while [[ "$dir" != / ]]; do
-    if [[ -f "$dir/settings.gradle" || -f "$dir/settings.gradle.kts" || -f "$dir/gradlew" ]]; then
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks!

PR #11918, opened 2023-09-29
docs(gas): update gas plugin URL

Conway:
## Standards checklist:

## Changes:

- Changed the link for the `gas` plugin to the current URL

## Other comments:

The previous url was a GitHub page hosted at http://walle.github.com/gas. The error message that appears says that github.com links no longer work for GitHub pages and to use github.io instead. After changing the URL to github.io, it redirected to this new link.


carlosala:
Thanks!

PR #11922, opened 2023-10-01
Add the node-bin plugin

### LINKED ISSUE #10924: Add node_modules/.bin to path

_No response_

When working with Node.js, it can be useful to run commands from the local `node_modules` folder. So for example instead of `yarn eslint .` or `npx eslint .`, one could simply run `eslint .`

Add `node_modules/.bin` to path. For monorepo support, this should be done for each `node_modules/.bin` folder found up the parent tree of the current working directory. It should also be cleaned up.

_No response_

### Additional context

I have been using this feature for a while now and I like it. I think it might be useful for others too.

```zsh
function precmd() {
  # Add node_modules/.bin to path
  path=( ${path[@]:#*node_modules*} )
  local p="$(pwd)"
  while [[ "$p" != '/' ]]; do
    if [[ -d "$p/node_modules/.bin" ]]; then
      path+=("$p/node_modules/.bin")
    fi
    p="$(dirname "$p")"
  done
  typeset -U path
}
```

It’s important the local paths are appended, not prepended, so these commands don’t override existing commands. I.e. a malicious package could register the binary `ls`.

This works with both npm and yarn, and any other package managers that store executables in `node_modules/.bin`.

### Related Issues

_No response_

remcohaszing:
## Standards checklist:

## Changes:

This plugin adds `node_modules/.bin` to your path. It does so recursively up the ancestor tree, meaning it also  works for subdirectories and monorepos.

## Other comments:

The `node_modules/.bin` paths are appended, not prepended. This means system executables take precedence over local node modules.

Fixes #10924

### DIFF
--- a/plugins/node-bin/README.md
+++ b/plugins/node-bin/README.md
@@ -1,0 +1,10 @@
+# nodenv plugin
+
+This plugin adds `node_modules/.bin` to your path. It does so recursively up the ancestor tree,
+meaning it also  works for subdirectories and monorepos.
+
+To use it, add `node-bin` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... node-bin)
+```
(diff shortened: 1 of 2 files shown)


carlosala:
See https://github.com/ohmyzsh/ohmyzsh/issues/10924#issuecomment-1762041759.
Thanks for the contribution!

PR #11927, opened 2023-10-02
fix(tmux): don't autostart inside background intelliJ

HeroCC:
## Standards checklist:

## Changes:

- Check for presence of `$INTELLIJ_ENVIRONMENT_READER` and if it exists, skip tmux autolaunch

## Other comments:

Jetbrains IDEs automatically launch an interactive terminal in the background to evaluate the system environment. However, using the `tmux` plugin with autostart breaks this behavior. This results in a degraded IDE experience (missing tools from $PATH for example), and a warning dialog.

This PR will use the suggested check provided by Jetbrains to disable the autostart feature in such an environment.

See more background on https://youtrack.jetbrains.com/articles/IDEA-A-19/Shell-Environment-Loading


carlosala:
Thanks for your contribution!

Issue #11829, opened 2023-08-05
ssh-agent plugin should fail gracefully and better handle edge cases.


EvanCarroll:
### Describe the bug

There are currently two failures modes which should be better addresses in the [`ssh-agent` plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/ssh-agent).

1. [In the event you do **NOT** have ssh-agent and you have the plugin enabled](https://unix.stackexchange.com/q/753314/3285), a more desirable error would read "You have enabled the `ssh-agent` plugin for ohmyzsh which requires `ssh-agent` which is not installed."

2. [In the event you do **NOT** have a `~/.ssh` directory](https://unix.stackexchange.com/q/753316/3285), a more desirable would read _"You have enabled the `ssh-agent` which requires a directory be created called `~/.ssh` to store a cache file. This directory does not exist, please create it."_ Alternatively, I suppose we could just create the `~/.ssh` directory with perms 700.

### Steps to reproduce

1. Enable plugin
2. Log in without `ssh-agent` for the first obscure error, or without an `~/.ssh` directory for the second obscure error.

### Expected behavior

Better error messages.

### OS / Linux distribution

Alpine/Debian Testing

### Zsh version

5.9

### Terminal emulator

kitty

### If using WSL on Windows, which version of WSL

None

PR #11929, opened 2023-10-03
fix(ssh-agent): add error message if `~/.ssh` is not found

vishal-sharma-369:
solves:  #11829

Added the code for ssh-agent plugin to fail gracefully and better handle edge cases.
Before this PR: If the ssh-agent plugin was enabled and the required ~/.ssh directory was not found , the plugin used to respond with an error message which was hard to understand.

Now if the ~/.ssh directory doesn't exist, plugin will respond with understandable error and will automatically create ~/.ssh directory as mentioned in the issue.

Message will be:

You have enabled the ssh-agent which requires ~/.ssh directory to store a cache file. 
Created ~/.ssh directory to store cache file.

This will only be displayed for first time if the directory doesn't existed before.

## Standards checklist:

## Changes:
Included Code after line 22 in ssh-agent.plugin.zsh file

    if [[ $? -ne 0 ]]; then
    mkdir ~/.ssh
    chmod 700 ~/.ssh
    echo "You have enabled the ssh-agent which requires ~/.ssh directory to store a cache file."
    echo "Created ~/.ssh directory to store cache file."
    return
    fi

Added the code for ssh-agent plugin to fail gracefully and better handle edge cases.
Before this PR: If the ssh-agent plugin was enabled and the required ~/.ssh directory was not found , the plugin used to respond with an error message which was hard to understand.

Now if the ~/.ssh directory doesn't exist, plugin will respond with understandable error and will automatically create ~/.ssh directory as mentioned in the issue.

Message will be:

You have enabled the ssh-agent which requires ~/.ssh directory to store a cache file.
Created ~/.ssh directory to store cache file.

This will only be displayed for first time if the directory doesn't existed before.

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -20,6 +20,13 @@ function _start_agent() {
   # start ssh-agent and setup environment
   zstyle -t :omz:plugins:ssh-agent quiet || echo >&2 "Starting ssh-agent ..."
   ssh-agent -s ${lifetime:+-t} ${lifetime} | sed '/^echo/d' >! "$ssh_env_cache"
+  if [[ $? -ne 0 ]]; then
+  mkdir ~/.ssh
+  chmod 700 ~/.ssh
+  echo "You have enabled the ssh-agent which requires ~/.ssh directory to store a cache file."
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution! I removed the part creating the directory since I'd say it's better not to modify any kind of sensitive directories like `~/.ssh`.

Issue #11700, opened 2023-05-19
rtx plugin


tenjaa:
### Describe the proposed feature.

I would love to see an official plugin for [rtx](https://github.com/jdxcode/rtx).

It is an alternative asdf written in rust.

mcornella:
Looks great! Would you like to submit a PR?

geckguy:
@mcornella can i work on this? Might need a little guidance but ill get the work done

carlosala:
sure, feel free to submit a PR, we'll be happy to review it!

sujalcodes3:
Is someone working on this or can I work on this issue ?
@carlosala

Cioraz:
Hello, is this issue open? I would like to work on it 
@carlosala  @tenjaa

tenjaa:
Go for it :)

alienishi:
Hello @carlosala @tenjaa , is this issue still open ? I would like to work on it 
please assign it to me

himanshuParashar0101:
okay we got it

demonkillerr:
Submitted a PR, kindly have a look: #11932

AlecRust:
Nice one for the plugin!

For me, [the new `rtx` plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/rtx) doesn't behave the same as the `eval "$(rtx activate zsh)"` I had in my `.zshrc` before, is this expected?

My installed Node version no longer displays in my Starship prompt, and commands like `npm` and `yarn` are no longer working. All the aforementioned things are loaded from OMZ:

https://github.com/AlecRust/dotfiles/blob/b3939525ea54dc397cb0c7223ba58979b7ccbc8e/dotfiles/.zsh_plugins.txt#L13-L14

**Before:**
![image](https://github.com/ohmyzsh/ohmyzsh/assets/247634/9a2ee421-3c52-4410-9146-538612b208b1)

**After:**
![image](https://github.com/ohmyzsh/ohmyzsh/assets/247634/2669c5fc-f437-4be3-91f3-526e4f9afd73)

carlosala:
There's a PR fixing all the stuff #11949.
Closing this, it should be done!

PR #11932, opened 2023-10-03
Added rtx plugin (asdf replacement)

demonkillerr:
## Standards checklist:

## Changes:
Added a plugin for rtx, which is faster asdf as it is written in rust. Solves #11700.
- [...]

## Other comments:
I would like to add aliases to both rtx and asdf based on use cases of users. An example would be
```alias install = "rtx install"```
...

### DIFF
--- a/plugins/rtx/README.md
+++ b/plugins/rtx/README.md
@@ -25,6 +25,6 @@ See the [rtx readme](https://github.com/jdx/rtx#table-of-contents) for informati
 ```
 rtx install node         Install the current version specified in .tool-versions/.rtx.toml
 rtx use -g node@system   Use system node as global default
-rtx install node@20.0.0  //Install a specific version number
+rtx install node@20.0.0  Install a specific version number
 rtx use -g node@20       Use node-20.x as global default
 ```
--- a/plugins/rtx/README.md
+++ b/plugins/rtx/README.md
@@ -8,8 +8,9 @@ Adds integration with [rtx](https://github.com/jdx/rtx), a runtime executor comp
 1. [Download & install rtx](https://github.com/jdx/rtx#installation) by running the following:
 
   ```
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by robbyrussell** [comment]

Cheers,
Robby

**Comment by tilgovi** [comment]
@demonkillerr I'm testing this out on two systems and seeing some weirdness.

On mac OS with rtx installed by brew, the `libexec` directory that's being referenced in the code doesn't exist. Is it possibly an arm/x86 difference? I'm on x86.

On Ubuntu with rtx installed via apt, the binary is `/usr/bin/rtx` and there are no other files installed that I can find. There is no completion file, for instance. I have to run `eval "$(rtx completion zsh)"`.

Any thoughts on either of these? I'm happy to try a PR, unless this all makes immediate sense to you and you want to make a quick fix.

**Comment by mcornella** [comment]
@tilgovi can you test #11949? `omz pr test 11949` should work for you

**Comment by tilgovi** [comment]
That is better insofar as it finds the rtx command. I think I prefer an option to set up shims instead of using activate, though, as it's more compatible with IDEs and such. But if this is how the community would prefer it, what you've got is fine!

**Comment by tilgovi** [comment]
Alternatively, it probably does no harm to add the shims first, then activate, and let the activation take precedence in the PATH? I defer to folks who are better acquainted with rtx.

**Comment by tilgovi** [comment]
For now I have this in `.oh-my-zsh/custom/rtx.zsh`, along with the plugin activated:

```
if [[ -z "$RTX_DATA_DIR" ]]; then
	export RTX_DATA_DIR="${XDG_DATA_HOME:-${HOME}/.local/share}/rtx"
fi

path=("$RTX_DATA_DIR/shims" $path)
typeset -U path
```

mcornella:
> Alternatively, it probably does no harm to add the shims first, then activate, and let the activation take precedence in the PATH?

This seems like a sensible option, could you open a PR for this?

PR #11935, opened 2023-10-06
feat(autojump): add path for nix per-user

puffnfresh:
NixOS configuration can have user-scoped packages, e.g.:

    {
      users.users.example.packages = [ pkgs.autojump ];
    }

There was an entry to try and support this but I think autojump must have changed in nixpkgs to break it.

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


carlosala:
Hi Brian! Looks good! Could we still maintain also the old entry, and maintain the previous order (with per-user at the bottom).
Thanks!

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #11952, opened 2023-10-10
feat(git): add new alias for searching branches

MrRoiz:
## Standards checklist:

## Changes:
- Add a new alias `gbf` to look for a pattern in local branches


mcornella:
Hi Oscar, thank you for the contribution. However, we will not move forward with this suggestion, for the following reasons:

- We have already a ton of aliases for git, and we need to be more careful about which aliases we include.
- This alias can already be used via `gb | grep`, or using git branch completion in the command you'll be running (i.e. `git checkout <TAB>`).
- We don't see this alias to be run very frequently, so it is not worth the addition.
- As `| grep` is also very common, you can also define a global alias to more quickly add it to any commands, such as
```sh
alias -g :G='| grep'
```

If you still have a requirement for this alias, I'd suggest adding it to your own `.zshrc` file. Thanks!

PR #11966, opened 2023-10-13
fix(git): fix fallback to develop branch if `master` not found

### LINKED ISSUE #11964: `gbda` no longer works when the default branch is called something other than main, trunk, mainline, default
### Describe the bug

When running `gbda`, I now get:

```
client git:(develop) gbda            
fatal: Not a valid object name master
fatal: not a valid object name -m
fatal: unknown commit master
...more of the same in a loop
```

This is probably caused by https://github.com/ohmyzsh/ohmyzsh/pull/11948

### Steps to reproduce

1. Start a repo with a default branch named, say, "develop"
1. Enable plugin 'git'
1. Run `gbda`
1. 💥 

### Expected behavior

It should delete all merged and squash-merged branches.

_No response_

### OS / Linux distribution

macOs 13.6

### Zsh version

5.9

### Terminal emulator

iTerm2

None

### Additional context

_No response_

mcornella:
## Standards checklist:

## Changes:

- Return error code if no main or develop branch were found
- Fall back to `git_develop_branch` if `git_main_branch` was not found in `gbda` function

## Other comments:

Fixes #11964

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -20,26 +20,31 @@ function current_branch() {
 function git_develop_branch() {
   command git rev-parse --git-dir &>/dev/null || return
   local branch
-  for branch in dev devel development; do
+  for branch in dev devel develop development; do
     if command git show-ref -q --verify refs/heads/$branch; then
       echo $branch
-      return
+      return 0
     fi
   done
+
   echo develop
+  return 1
 }
 
 # Check if main exists and use instead of master
 function git_main_branch() {
   command git rev-parse --git-dir &>/dev/null || return
   local ref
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!
It indeed fixes the issue for me.
It might be nice maybe to default to `main` instead of `master`, seems to be more common nowadays.

LunarWatcher:
I wouldn't recommend hard switching from master to main. Both are still widely used. Consider checking both by default instead

carlosala:
@LunarWatcher I was talking about the branch name we output if no one is found.

mcornella:
I'd rather not over-rely on the default here and not introduce a breaking change if it's possible, so I'll merge it directly.

Thanks for the review!

PR #11993, opened 2023-10-19
Check if cmd is available in current lando dir.

rtvenge:
## Standards checklist:

## Changes:

- Added a check before executing the command within lando to see if the command exists in lando.

## Other comments:

The initial issue was the command would always execute inside the Lando container(s), even if it didn't exist. Example: running `npm` would execute `lando npm` when I didn't have Node installed in that lando container(s).

I'm not super familiar with shell scripting, so if there is a better way for this conditional, please let me know. 😄

### DIFF
--- a/plugins/lando/lando.plugin.zsh
+++ b/plugins/lando/lando.plugin.zsh
@@ -11,7 +11,7 @@ function artisan \
          php \
          wp \
          yarn {
-  if checkForLandoFile; then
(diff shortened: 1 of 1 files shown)


rtvenge:
@mcornella thanks for putting that fix in so quickly!

…and clearly you know more about shell scripting than I. 😆 I saw you rewrote it for perf. 🎉

PR #11998, opened 2023-10-21
fix(firewalld): remove ` (default)` from the end of zone string

Z4HD:
## Standards checklist:

## Changes:

- fix `Error: INVALID_ZONE: public (default)` when using `fwl` with `firewall-cmd` version `2.0.1`

## Other comments:

In `firewall-cmd` version `2.0.1`, command `firewall-cmd --get-active-zones` output like this.

```
public (default)
  interfaces: wlp0s20f3
trusted
  sources: 127.0.0.1
```

Just remove ` (default)` from the end of zone string. This fix won't affect any zone string without ` (default)`.


carlosala (approved):

carlosala:
Thanks for your contribution!

PR #12018, opened 2023-10-31
feat(aws): implement permanent state

0rxa:
## Standards checklist:

- [?] The code is efficient, to the best of my ability, and does not waste computer resources.
- [n/a] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- Introduced a mechanism to store the AWS Profile in a file in order to survive shell sessions 

## Other comments:
~This implementation will affect the shell startup time noticeably because of the file read operation and call to asp/asr on each new shell. If you clear the profile (meaning state file is empty) it will have no measurable impact.~

EDIT: We fixed the performance issue by avoiding validation when loading profile from state.


carlosala (approved):
LGTM!

carlosala:
I like the idea, let's go for it!

carlosala:
Thanks for your contribution!

PR #12026, opened 2023-11-05
feat(plugins): Add c plugin

Myzel394:
## Standards checklist:

## Changes:

- Add new `c` command

## Other comments:

Not stable yet, gonna have to figure out how to properly find out the amount of newlines each prompt creates. Just wanted to open this, in case anyone else has the same idea and can then work on it :D

### DIFF
--- a/plugins/c/README.md
+++ b/plugins/c/README.md
@@ -1,0 +1,20 @@
+# c
+
+Navigate through your files in the current prompt.
+
+No more thousands of lines of output to go to the correct directory.
+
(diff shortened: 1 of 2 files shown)


carlosala:
Not sure which is the actual benefit against using directly `cd`. It seems like a really corner case to prefer not displaying a new prompt.

carlosala:
Hi! Closing here. As I said before it seems like a weird corner case. Thanks for your contribution!

PR #12029, opened 2023-11-07
Avoid shell injection

soliton-:


soliton-:
Oops, was imagining a -c there. Nevermind.

PR #12030, opened 2023-11-07
Update git.plugin.zsh

fsa:
## Standards checklist:

## Changes:

- gpa - git push in all repo
- gpat -  git push --tags in all repo


carlosala:
Both aliases seem to have a kind of destructive behaviour. One could mistype `gpa` when typing `gp` just to push to one repo (for example, a private fork) and instead push to all repos (for example, a public upstream).
I'd say these functions fit better in each interested user `zshrc`. Thanks for your contribution!

PR #12038, opened 2023-11-15
fix(direnv): unbreak and simplify direnv support

luzat:
To install direnv in zsh, one should `eval "$(direnv hook zsh)"`. Previously, ohmyzsh included the code generated by direnv hook, which included a `direnv export zsh` call that PR #12000 erroneously replaced: `direnv hook` at that place was correct. Now, direnv is only being initialized after the first command executed and doesn't work for the first command issued after the shell has started.

This commit simplifies and fixes the code by just using the recommended official way to install `direnv` in zsh:

https://github.com/direnv/direnv/blob/master/docs/hook.md#zsh

The result is about the same as with the working version before, but with less code and automatically receiving updates to direnv's zsh integration.

## Standards checklist:

### DIFF
--- a/plugins/direnv/direnv.plugin.zsh
+++ b/plugins/direnv/direnv.plugin.zsh
@@ -1,16 +1,5 @@
 # Don't continue if direnv is not found
 command -v direnv &>/dev/null || return
 
-_direnv_hook() {
-  trap -- '' SIGINT;
-  eval "$(direnv hook zsh)";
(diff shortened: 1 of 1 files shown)


carlosala:
I tested it and it seems that `$(direnv hook zsh)` takes ~10ms and it has always the same output. It seems to be better to embed it in ohmyzsh.
Thanks for your contribution!

PR #12045, opened 2023-11-17
feat(bgnotify): add support to wayland

bretello:
`bgnotify_appid` only checked for the presence of `xprop`, which might be available when using wayland, but will not work. Checking for `DISPLAY` is enough to prevent calling `xprop` and failing.

The other case is for wayland + sway and requires jq

## Standards checklist:

## Changes:

- now works properly on wayland
- explicit support for sway
- explicit support for Alacritty (adds alacritty icon to notification)


carlosala:
Thanks for your contribution!

PR #12046, opened 2023-11-17
fix(juju): add `public-address` fallback to `jaddr`

babakks:
## Standards checklist:

## Changes:

- The helper function `jaddr` is not working correctly. This might be because of the newer Juju versions (I'm using 3.3, for example). As I checked, the `address` field is now missing in the JSON data returned by `juju status` command. I've added a second fallback that reads the `public-address` field, as well.

## Other comments:

- In jq, `//` is called *Alternative* operator. The jq's man page explains:
  > A filter of the form a // b produces the same results as a, if a produces results other than false and null. Otherwise, a // b produces the same results as b.

### DIFF
--- a/plugins/juju/juju.plugin.zsh
+++ b/plugins/juju/juju.plugin.zsh
@@ -98,7 +98,7 @@ jaddr() {
   elif [[ $# -eq 2 ]]; then
     # Get unit address
     juju status "$1/$2" --format=json \
-      | jq -r ".applications.\"$1\".units.\"$1/$2\".address"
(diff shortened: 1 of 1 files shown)


babakks:
@sed-i @jnsgruk @mcornella @Abuelodelanada I'd appreciate it if you review this PR.

carlosala (approved):
I didn't test it, but it has no incompatibilities with the previous version. LGTM! 🚀

carlosala:
Thanks for your contribution!

PR #12055, opened 2023-11-22
feat(bgnotify): add icon parameter

sjoblomj:
## Standards checklist:

## Changes:

- Minor cleanup by using named parameter variables instead of numbered parameter variables
- If using Wayland but not Sway, it will no longer fetch app_id from [xprop](https://man.archlinux.org/man/xprop.1.en), which gives the wrong result.
- Added an icon parameter to the `bgnotify` function. While this is unused currently, it allows the user to put something like this in their ~/.zshrc to set the icon depending on the command's exit status:

```
function bgnotify_formatted {
  ## $1=exit_status, $2=command, $3=elapsed_time
  [ $1 -eq 0 ] && title="Great success" || title="Command failed"
  [ $1 -eq 0 ] && icon="$HOME/icons/success.png" || icon="$HOME/icons/fail.png"
  bgnotify "$title - took ${3}s" "$2" "$icon"
}
plugins=(git bgnotify sudo)  # for example
```

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -52,10 +52,10 @@ function bgnotify_formatted {
   (( $3 < 60 )) || elapsed="$((( $3 % 3600) / 60 ))m $elapsed"
   (( $3 < 3600 )) || elapsed="$(( $3 / 3600 ))h $elapsed"
 
-  if [[ $1 -eq 0 ]]; then
-    bgnotify "#win (took $elapsed)" "$2"
+  if [[ $exit_status -eq 0 ]]; then
+    bgnotify "#win (took $elapsed)" "$cmd"
   else
-    bgnotify "#fail (took $elapsed)" "$2"
+    bgnotify "#fail (took $elapsed)" "$cmd"
   fi
 }
 
@@ -66,7 +66,7 @@ function bgnotify_appid {
   elif [[ -n $WAYLAND_DISPLAY ]] && (( ${+commands[swaymsg]} )) && (( ${+commands[jq]} )); then # wayland+sway
     # output is "app_id, container id" (Alacritty, 1694)
     swaymsg -t get_tree | jq '.. | select(.type?) | select(.focused==true) | {app_id, id} | join(", ")'
-  elif [[ -n $DISPLAY ]] && (( ${+commands[xprop]} )); then
+  elif [[ -z $WAYLAND_DISPLAY ]] && [[ -n $DISPLAY ]] && (( ${+commands[xprop]} )); then
     xprop -root _NET_ACTIVE_WINDOW 2>/dev/null | cut -d' ' -f5
   else
     echo $EPOCHSECONDS
@@ -76,6 +76,7 @@ function bgnotify_appid {
 function bgnotify {
   local title="$1"
   local message="$2"
+  local icon="$3"
   if (( ${+commands[terminal-notifier]} )); then # macOS
     local term_id="${bgnotify_termid%%,*}" # remove window id
     if [[ -z "$term_id" ]]; then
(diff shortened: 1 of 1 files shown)


plugins/bgnotify/bgnotify.plugin.zsh
```diff
@@ -85,23 +86,15 @@ function bgnotify {
...
-      terminal-notifier -message "$message" -title "$title" &>/dev/null
-    else
-      terminal-notifier -message "$message" -title "$title" -activate "$term_id" -sender "$term_id" &>/dev/null
-    fi
+    terminal-notifier -message "$message" -title "$title" ${=icon:+-appIcon "$icon"} ${=term_id:+-activate "$term_id" -sender "$term_id"} &>/dev/null
   elif (( ${+commands[growlnotify]} )); then # macOS growl
     growlnotify -m "$title" "$message"
   elif (( ${+commands[notify-send]} )); then
-    if [[ -n $ALACRITTY_WINDOW_ID ]]; then
-        notify-send -i Alacritty "$title" "$message"
```
sjoblomj, line 97: This was added recently in #12045. This PR will no longer show the Alacritty icon, but the user can easily re-add this functionality by adding this in their ~/.zshrc file:

```
function bgnotify_formatted {
  ## $1=exit_status, $2=command, $3=elapsed_time
  [ $1 -eq 0 ] && title="Great success" || title="Command failed"
  bgnotify "$title - took ${3}s" "$2" "Alacritty"
}
plugins=(git bgnotify sudo)
```

There are several reasons why I'm skeptical to hard-code Alacritty here:
* There are [dozens of different terminal emulators](https://en.wikipedia.org/wiki/List_of_terminal_emulators). How many are a reasonable number to hard-code here?
* Can we ensure icons are somewhat reliably installed in accessible places across different distributions and OSes?

plugins/bgnotify/bgnotify.plugin.zsh
```diff
@@ -85,23 +86,15 @@ function bgnotify {
...
     fi
 
-    if [[ -z "$term_id" ]]; then
-      terminal-notifier -message "$message" -title "$title" &>/dev/null
-    else
-      terminal-notifier -message "$message" -title "$title" -activate "$term_id" -sender "$term_id" &>/dev/null
-    fi
+    terminal-notifier -message "$message" -title "$title" ${=icon:+-appIcon "$icon"} ${=term_id:+-activate "$term_id" -sender "$term_id"} &>/dev/null
   elif (( ${+commands[growlnotify]} )); then # macOS growl
     growlnotify -m "$title" "$message"
```
sjoblomj, line 91: Not part of this PR, but is this software used anymore? A somewhat quick search suggests that growlnotify was abandoned some years ago.

carlosala: Let's keep it. Maybe someone is still using it. We support zsh down to 5.0.8 (released in mid-2015).

plugins/bgnotify/bgnotify.plugin.zsh
carlosala: I agree, let's go for it!

carlosala (approved):
LGTM! 🎉

plugins/bgnotify/bgnotify.plugin.zsh
bretello: Regarding linux: `notify-send` can take a file path or an app name as icon, and we can expect those icons to be in `/usr/share/icons` or `/usr/share/pixmaps`. The issue I see is that there's no reliable way (that I know of) of getting the appropriate icon name for each terminal, and we'd have to maintain a mapping of env vars -> icon names.

Another approach that could work, is to add an icon for bgnotify in this repo, and have that be the default if `icon` is not explicitly set by the user.

plugins/bgnotify/bgnotify.plugin.zsh
```diff
@@ -76,6 +76,7 @@ function bgnotify_appid {
 function bgnotify {
   local title="$1"
   local message="$2"
+  local icon="$3"
```
bretello, line 79: How about something like this?

```diff
-    local icon="$3"
+    local icon="${3=$ZSH/plugins/bgnotify.png}"
```

sjoblomj: Bundling icons is maybe the neatest, yeah. I'd be interested to hear what the maintainers think about that approach.

Rather than changing the line you suggest, I'd argue that the icon should be put inside the function `bgnotify_formatted`.  And I'd argue that it is more helpful to have one icon for success and a different for failure. Attached is a screenshot of my own setup, but maybe everyone is not as fond of wolves as I am? :sweat_smile:

swappy-20231124_201728

bretello: > swappy-20231124_201728

:rofl: that's a neat one.

Yeah, it's a neat idea. I just found about [zsh-notify](https://github.com/marzocchi/zsh-notify) where it looks like this is already implemented

sjoblomj: Looks nice, thanks for the heads up. And I see that others also like wolves and dogs. 😎 🐺  Much amaze.
https://gist.github.com/marzocchi/14c47a49643389029a2026b4d4fec7ae

PR #12060, opened 2023-11-24
fix(misc): only set PAGER if `less` or `more` are available

### LINKED ISSUE #12049: Oh my zsh uses inconsistent default environment variable (PAGER)
### Describe the bug

In the lib/misc.zsh file, the environment variable PAGER is placed as the default less, however, in some environments it does not have the less command. I know that it is possible to change it in .zshrc, but it would be interesting to check if you have less/more installed before adding it. If not, it is preferable not to have the PAGER.

### Steps to reproduce

To reproduce the error, simply print the default environment variable without command less.
```
echo $PAGER
```

### Expected behavior

To reproduce the error, simply print the default environment variable without command less.

_No response_

### OS / Linux distribution

Arch Linux

### Zsh version

5.9

### Oh My Zsh version

master (c8e600f3)

### Terminal emulator

Kitty

None

### Additional context

_No response_

wonko:
Simple detection of `less` or `more`, and set PAGER to whatever is available.

Fixes #12049 

## Standards checklist:

## Other comments:

I did not touch the line in https://github.com/ohmyzsh/ohmyzsh/blob/master/lib/cli.zsh#L392 as I can't be certain that the pager is already set when hitting the code. Seemed over-engineering to have the same detection there again.

### DIFF
--- a/lib/misc.zsh
+++ b/lib/misc.zsh
@@ -19,8 +19,13 @@ setopt multios              # enable redirect to multiple streams: echo >file1 >
 setopt long_list_jobs       # show long list format job notifications
 setopt interactivecomments  # recognize comments
 
-env_default 'PAGER' 'less'
-env_default 'LESS' '-R'
+# define pager dependant on what is available (less or more)
+if (( ${+commands[less]} )); then
+  env_default 'PAGER' 'less'
+  env_default 'LESS' '-R'
+elif (( ${+commands[more]} )); then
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM! 🚀

carlosala:
I like the idea, and does not break the current behaviour if `less` is installed. Thanks for the contribution!

PR #12062, opened 2023-11-25
refactor(bgnotify): Getting appid using awk instead of jq

sjoblomj:
## Standards checklist:

- [x] ~~If the code introduces new aliases, I provide a valid use case for all plugin users down below.~~ (No new aliases)

## Changes:

The idea of bgnotify is that a user can start a long running command, switch window and do something else, and get a system notification when the command is done. If the user has not switched windows, the user should not get the system notification - the plugin does not want to spam when it is already obvious that the command finished. To know if a notification should be displayed, bgnotify gets the appid of the currently focused window.

In the excellent work by @bretello in #12045, the `appid, id` of the currently focused window is retrieved from `swaymsg` for users of Wayland+Sway. The relevant parts from `swaymsg` are then extracted using `jq`, which is definitely the right tool for the job. The issue however, is that `jq` is not part of the standard tools on Linux or Mac, and I believe we could cater to a wider audience by not using `jq`.

This PR is my attempt at extracting the same data using `awk` instead (`awk` being [installed by default on every modern full-fledged Unix](https://unix.stackexchange.com/a/82411)). This is not nearly as neat and pretty as the original solution using `jq` (again, `jq` is the right tool for the job), but as stated, it can be expected to work for far more users.

**I am fine with this PR being rejected.** In my mind, we're choosing between a neater and more maintainable solution by keeping `jq`, versus catering to more users by using `awk`.

Attached is output from `swaymsg -t get_tree`, if anyone wants to poke around with my awk-script: [output.json](https://github.com/ohmyzsh/ohmyzsh/files/13463939/output.json)

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -59,13 +59,39 @@ function bgnotify_formatted {
   fi
 }
 
+function find_sway_appid {
+  # output is "app_id,container id", for instance "Alacritty,1694"
+  swaymsg -t get_tree | awk '
+    BEGIN { Id = ""; Appid = ""; FocusNesting = -1; Nesting = 0 }
+    {
+      # Enter a block
+      if ($0 ~ /.*{$/) Nesting++
+
+      # Exit a block. If Nesting is now less than FocusNesting, we have the data we are looking for
+      if ($0 ~ /^[[:blank:]]*}.*/) { Nesting--; if (FocusNesting > 0 && Nesting < FocusNesting) exit 0 }
+
+      # Save the Id, it is potentially what we are looking for
+      if ($0 ~ /^[[:blank:]]*"id": [0-9]*,$/)    { sub(/^[[:blank:]]*"id": /, "");      sub(/,$/,  ""); Id = $0 }
+
+      # Save the Appid, it is potentially what we are looking for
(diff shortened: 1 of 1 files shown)


mcornella:
We're already using awk / sed for JSON parsing in other plugins such as yarn, I agree with this change. Assigning to me for testing.

bretello:
>The issue however, is that jq is not part of the standard tools on Linux or Mac, and I believe we could cater to a wider audience by not using jq.

Even though this is true:
- We do not care about MacOS: this path is only taken on wayland when `swaymsg` is present
- Even though `jq` is not installed by default, I'd argue that it's highly likely that if `sway` is being used, `jq` is installed, since it's both a convenient and very common tool used to process `swaymsg`'s json output.
- I'd rather maintain a `jq` oneliner rather than 20+ lines `awk`: I really don't think it's the right tool for the job.

carlosala:
> * Even though `jq` is not installed by default, I'd argue that it's highly likely that if `sway` is being used, `jq` is installed, since it's both a convenient and very common tool used to process `swaymsg`'s json output.
> 

Doing a quick search in Archlinux's `sway` package, I don't see `jq` as a dependency (not even an optional dependency). I agree though that it may be a bit too much to have that `awk` script to just parse a JSON. On the other hand, it's not that long and would be really convenient for people that didn't installed `jq`.

bretello:
@carlosala 
Yeah, it's not a dependency, although it's widely used, just see the wiki page as an example:  https://wiki.archlinux.org/title/Sway

mcornella:
Included the changes while making `jq` the first option. The `awk` script works well with the current output from `swaymsg`, but future changes may break it, so it's still better to rely on `jq`. Thank you for the contribution!

PR #12066, opened 2023-11-27
Adding in nextjs search functionality. And add in DDG lucky feature.

jhubbardsf:
Adding in nextjs search functionality. Also added in lucky feature the top result from DDO automatically added for things tha the fallback search provider..

## Standards checklist:

## Changes:

- Add nextjs search option for frontend-search plugin.
- Add FRONTEND_SEARCH_FALLBACK_LUCKY env variable which allows the plugin to automatically open the top DuckDuckGo search result for context's that use the fallback search url.

## Other comments:
I can break this up into two separate PRs if that makes more sense. But since the lucky ("ducky") feature is completely opt-in, and doesn't change the default behavior of the plugin, I thought it might be okay to do it in one.

### DIFF
--- a/plugins/frontend-search/README.md
+++ b/plugins/frontend-search/README.md
@@ -66,7 +66,16 @@ If you want to have another context, open an Issue and tell us!
 
 ## Fallback search behaviour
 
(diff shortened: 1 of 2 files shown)


carlosala (approved):
I like it, thanks! LGTM 🚀

PR #12072, opened 2023-11-29
ci(external): add automation for updating external dependencies

mcornella:
## Standards checklist:

## Changes:

- Add `dependencies.yml` which tracks which dependencies we have in the repository and the path they are included in.
- Add `updater.py` script which:
  - Checks upstream repositories for updates (currently only tracks commit SHAs, next version should also allow tracking tags)
  - Includes changes into the right path, with ability to run `precopy` and `postcopy` scripts (`precopy` runs on cloned upstream repository, `postcopy` runs on the dependency path within our repo)
  - Creates a Pull Request if everything went well, otherwise opens an issue notifying of the available update

## Other comments

Currently there is a workflow included that is only manually runnable. In the future, a cron script should be included (currently commented out), and also:

- The GitHub App "ohmyzsh" should be used to make these PRs as [referred here](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation)
- PR creation has not been tested, I believe we should also push the branch (for which deploy Keys would be needed)

We have many other dependencies that are not included in the `dependencies.yml`. Once this is verified to work, we can include these as well.

We should also do something about tracking whether an Issue/PR has already been created by the automation.

### DIFF
--- a/.github/workflows/external.yml
+++ b/.github/workflows/external.yml
@@ -1,0 +1,27 @@
+name: Update dependencies
+on:
+  workflow_dispatch: {}
+  # schedule:
+  #   - cron: '34 3 * * */8'
+
+jobs:
+  check:
+    name: Check for updates
+    runs-on: ubuntu-latest
+    container: alpine:latest
+    steps:
+      - name: Checkout
+        if: github.repository == 'ohmyzsh/ohmyzsh'
+        uses: actions/checkout@v4
+      - name: Authenticate as @ohmyzsh
+        uses: ohmyzsh/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92
+        with:
+          app-id: ${{ secrets.OHMYZSH_APP_ID }}
+          private-key: ${{ secrets.OHMYZSH_APP_PRIVATE_KEY }}
+      - name: Process dependencies
+        env:
+          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+          TMP_DIR: ${{ env.RUNNER_TEMP }}
+        run: |
+          gh auth login --with-token <<< "${GITHUB_TOKEN}"
+          python3 .github/workflows/dependencies/updater.py
--- a/.github/dependencies.yml
+++ b/.github/dependencies.yml
@@ -1,0 +1,20 @@
+dependencies:
+  plugins/gitfast:
+    repo: felipec/git-completion
+    branch: master
+    version: d700a27137bafeacd8a7734a47ae577614f6c32a
+    postcopy: |
+      set -e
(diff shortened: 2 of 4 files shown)


.github/workflows/external.yml
```diff
@@ -0,0 +1,27 @@
...
+  check:
+    name: Check for updates
+    runs-on: ubuntu-latest
+    container: alpine:latest
+    steps:
+      - name: Checkout
+        if: github.repository == 'ohmyzsh/ohmyzsh'
+        uses: actions/checkout@v4
+      - name: Authenticate as @ohmyzsh
+        uses: ohmyzsh/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92
```
carlosala, line 17: Maybe we can push some tag, it looks really bad to have the commit here.

.github/workflows/external.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  workflow_dispatch: {}
+  # schedule:
+  #   - cron: '34 3 * * */8'
+
+jobs:
+  check:
+    name: Check for updates
+    runs-on: ubuntu-latest
+    container: alpine:latest
```
carlosala, line 11: It seems like we don't need to use a container.

.github/workflows/external.yml
carlosala, line 17: Maybe we can push some tag, it looks really bad to have the commit here.

.github/workflows/external.yml
```diff
@@ -0,0 +1,27 @@
+name: Update dependencies
+on:
+  workflow_dispatch: {}
+  # schedule:
+  #   - cron: '34 3 * * */8'
+
+jobs:
+  check:
```
carlosala, line 8: I'd use https://github.com/actions/setup-python and have some `requirements.txt`. Not sure if all packages used in the script are expected to be installed.

carlosala (approved):
I like it! Let's see how it works and we can modify it afterwards accordingly.
LGTM 🚀

.github/workflows/external.yml
mcornella: All packages imported are included in python3 by default, and the GitHub Actions runner already has python3 set up I believe.

mcornella:
Before merging, I'd like to set up support for tag updating.

Checking tags can be made with
```
# set up local repo with remote
# check remote tags
git ls-remote --tags
```
And checking out a specific tag with `--branch`, e.g.
```
git clone --depth=1 --branch v2.0 https://github.com/felipec/git-completion
```

carlosala:
Sounds great! We could use https://github.com/python-semver/python-semver to check if it's indeed a newer version.

PR #12076, opened 2023-11-30
fix(lib): patch `omz_urlencode` to not encode UTF-8 chars in Termux

### LINKED ISSUE #12061: web-search baidu 你好
Hello Developer, as the title suggests, when I use "web-search baidu 你好" command, it shows y81o2BY29. I suspect this might be due to the lack of support for Chinese characters. I hope this issue could be resolved as soon as possible. This problem was observed while using Termux terminal. Thank you.

mcornella:
## Standards checklist:

## Changes:

- Termux only has a single UTF-8 locale, so byte-by-byte encoding is not possible. This PR effectively bypasses that so that the resulting URL has the UTF-8 characters directly, which is [already possible in URLs](https://stackoverflow.com/a/913653).

Fixes #12061

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -182,6 +182,8 @@ function omz_urlencode() {
   fi
 
   # Use LC_CTYPE=C to process text byte-by-byte
+  # Note that this doesn't work in Termux, as it only has UTF-8 locale.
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM, we can merge! 🚀

PR #12078, opened 2023-12-01
Add more configuration  robbyrussell.zsh-theme 

chiddekel:
DIFF code add more configuration for theme by alias the git_prompt_info  using custom function  git_prompt_info_color

### DIFF
--- a/themes/robbyrussell.zsh-theme
+++ b/themes/robbyrussell.zsh-theme
@@ -1,3 +1,34 @@
+function git_prompt_info_color() {
+  local ref_color=""
+  local git_status="$(parse_git_dirty)"
+ if [[ -d  .git || $(git rev-parse --is-inside-work-tree 2>/dev/null) = true ]]; then
+  if [[ -n $(git status -s) ]]; then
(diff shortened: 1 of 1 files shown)


M21H (approved):

carlosala:
Hi! Thanks for the contribution!
We're not interested at this point in adding particular customization to specific plugins since that git prompt is standard for all oh-my-zsh(defined in [lib/git.zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/lib/git.zsh)).
It's a great idea, though! I recommend you add that customization in your custom directory.
Thanks for your contribution!

chiddekel:
> Hi! Thanks for the contribution! We're not interested at this point in adding particular customization to specific plugins since that git prompt is standard for all oh-my-zsh(defined in [lib/git.zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/lib/git.zsh)). It's a great idea, though! I recommend you add that customization in your custom directory. Thanks for your contribution!

Where move code to add This customization with switch on of in oh my zsh and I will make another PR.??

PR #12081, opened 2023-12-03
chore(amuse): add a space at the end of virtualenv prefix

khanhnt2:
## Standards checklist:

## Changes:
Before
Screenshot 2023-12-03 at 08 46 22
After
Screenshot 2023-12-03 at 08 46 58


mcornella:
Hi Nguyen, thank you for the contribution. It looks like the emoji is not being rendered correctly with your terminal emulator, hence why you need the spacing.

That said, these variables are designed so that you can change them yourself. In this case, it would suffice redefinining the variable in your .zshrc file, after OMZ has been sourced.

Thanks!

PR #12084, opened 2023-12-05
feat(plugins): Add nala plugin

james-horrocks:
## Standards checklist:

## Changes:

- Added new plugin under `plugins/nala/nala.plugin.zsh`

## Use case

Nala provides a user friendly wrapper around `apt` on Debian based distros.

I couldn't say exactly how many people use Nala but it has featured on [itsfoss.com](https://itsfoss.com/nala) and [omgubuntu.co.uk](https://www.omgubuntu.co.uk/2023/01/install-nala-on-ubuntu). 
It is also installable from official repos in Debian and Ubuntu.

## Other comments:

I haven't tested the `nar` function as I didn't have a repository to test it on, but it's almost an exact copy of the `aar` function from the **[Ubuntu plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/ubuntu)** so it *should* work

### DIFF
--- a/plugins/nala/README.md
+++ b/plugins/nala/README.md
@@ -1,0 +1,53 @@
+# Nala plugin
+
+This plugin adds aliases for [Nala](https://gitlab.com/volian/nala).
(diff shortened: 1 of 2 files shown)


carlosala:
Hi! It doesn't seem very used (<1k stars on Github). Even though, the main issue is the amount of sudo aliases. We try to avoid as much as possible sudo aliases (most of them were created a long time ago).
Thanks for your contribution!

PR #12108, opened 2023-12-19
fix(rake-fast): make `.rake_tasks` write atomic

mehanoid:
## Standards checklist:

## Changes:

- The sequence of steps when generating `.rake_tasks` file has been changed
  - a variable with the contents of this file is created
  - the variable is written to a temporary file
  - the temporary file is moved to the location of the final `.rake_tasks` file.

## Other comments:

This ensures that the `.rake_tasks` file is not corrupted if a process terminates prematurely or if multiple processes run in parallel. This is useful in particular when used in conjunction with the [marlonrichert/zsh-autocomplete](https://github.com/marlonrichert/zsh-autocomplete) plugin, which, while typing, runs an autocomplete command every time a key is pressed, causing the file to become corrupted.

### DIFF
--- a/plugins/rake-fast/rake-fast.plugin.zsh
+++ b/plugins/rake-fast/rake-fast.plugin.zsh
@@ -43,14 +43,17 @@ _tasks_changed () {
 }
 
 _rake_generate () {
-  echo "version:$_rake_tasks_version" > .rake_tasks
-
-  rake --silent --tasks --all \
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
`mktemp` seems to be widely spread among all UNIX-like systems. Let's merge it!

carlosala:
Thanks for your contribution!

PR #12109, opened 2023-12-20
ci(dependencies): add automation for updating external dependencies

mcornella:
## Standards checklist:

## Changes:

- Add `dependencies.yml` which tracks which dependencies we have in the repository and the path they are included in.
- Add `updater.py` script which:
  - Checks upstream repositories for updates, tracking newer SHAs and tags (algorithm for tags is quite simple, might have issues in the future)
  - Includes changes into the right path, with ability to run `precopy` and `postcopy` scripts (`precopy` runs on cloned upstream repository, `postcopy` runs on the dependency path within our repo)
  - Creates a Pull Request if everything went well, otherwise opens an issue notifying of the available update

## Other comments

Currently there is a workflow included that is only manually runnable. In the future, a cron script should be included (currently commented out), and also:

- The GitHub App "ohmyzsh" should be used to make these PRs as [referred here](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation)
- PR creation needs to happen in GitHub Actions runner, for which a deploy SSH key needs to be also added in secrets.

We have many other dependencies that are not included in the `dependencies.yml`. Once this is verified to work, we can include these as well.

We should also do something about tracking whether an Issue/PR has already been created by the automation. This would be already possible to check for PRs by looking at whether the branch already exists, otherwise we can check if there is an issue with the same title.

Republish of #12072 which was broken.

### DIFF
--- a/.github/dependencies.yml
+++ b/.github/dependencies.yml
@@ -2,7 +2,7 @@ dependencies:
   plugins/gitfast:
     repo: felipec/git-completion
     branch: master
-    version: d700a27137bafeacd8a7734a47ae577614f6c32a
+    version: tag:v1.3.7
     postcopy: |
       set -e
       rm -rf git-completion.plugin.zsh Makefile README.adoc t tools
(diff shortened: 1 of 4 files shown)


carlosala (approved):
LGTM!

PR #12110, opened 2023-12-20
update to replace current dir in prompt with git repo name, when in a git dir.

WesleyRen:
## Standards checklist:

## Changes:

The existing prompt has this format:
`user_id@hostname ~<current_dir_from_user_root> > main >`
Current dir can get long. Since we mostly care about what repo I'm working on for current shell, I replaced the current dir with the repo name instead when I'm in a git repo dir.
I found that's been working well for me.
`user_id@hostname ~<repo_name> > main >`

This PR doesn't change the prompt behavior when in a non-git dir.

## Other comments:

...

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -217,7 +217,13 @@ prompt_hg() {
 
 # Dir: current working directory
 prompt_dir() {
-  prompt_segment blue $CURRENT_FG '%~'
+  if $(git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
+    gn=`git rev-parse --show-toplevel`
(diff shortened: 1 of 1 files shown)


gautamkrishnar:
As a user of Agonster, I think it would be pretty invasive to force this on all the users of this theme IMO. Providing this as an opt-in feature is fine. It's up to the maintainer though.

carlosala:
Hey @WesleyRen! Thanks for your contribution! I agree with Gautam, agnoster is a widely used theme, and changing that would be a bit breaking in terms of UI perspective. I encourage you to modify it and keep in your `custom` directory.
Thanks!

Issue #12097, opened 2023-12-18
When running with bgnotify plugin, `set -e` will terminate zsh immediately


ipChrisLee:
### Describe the bug

When `bgnotify` is in usage, running `set -e` will cause zsh terminate.

### Steps to reproduce

1. Add `bgnotify` to plugin list in .zshrc
2. Run `set -e` on interactive zsh, or source a shell script that contains `set -e` statement.

### Expected behavior

I think `set -e` should not cause shell down.

### OS / Linux distribution

Kubuntu 22.04.3 LTS x86_64

### Zsh version

5.8.1

### Terminal emulator

konsole

### If using WSL on Windows, which version of WSL

None

### Additional context

This issue may be solved by :
```Diff
diff --git a/plugins/bgnotify/bgnotify.plugin.zsh b/plugins/bgnotify/bgnotify.plugin.zsh
index 3c07661..c28a29e 100644
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -21,11 +21,11 @@ function bgnotify_end {
     local elapsed=$(( EPOCHSECONDS - bgnotify_timestamp ))
 
     # check time elapsed
-    [[ $bgnotify_timestamp -gt 0 ]] || return
-    [[ $elapsed -ge $bgnotify_threshold ]] || return
+    [[ $bgnotify_timestamp -gt 0 ]] || return 0
+    [[ $elapsed -ge $bgnotify_threshold ]] || return 0
 
     # check if Terminal app is not active
-    [[ $(bgnotify_appid) != "$bgnotify_termid" ]] || return
+    [[ $(bgnotify_appid) != "$bgnotify_termid" ]] || return 0
 
     [[ $bgnotify_bell = true ]] && printf '\a' # beep sound
     bgnotify_formatted "$exit_status" "$bgnotify_lastcmd" "$elapsed"

```

But I'm not an expert of zsh, and I'm not sure that return 0 is expected. So this is just a information for you.

mcornella:
This will fix it yeah, `return` without argument sends the previous return code which in this case will be non-zero. Please submit a PR if you're able!

gautamkrishnar:
@mcornella opened a PR, given attribution to the original author as well

PR #12111, opened 2023-12-21
fix(bgnotify): make it work with `set -e`

gautamkrishnar:
## Standards checklist:

## Changes:
fixes: #12097 When running with bgnotify plugin, set -e will terminate zsh immediately 

- Added return 0 instead of return

## Other comments:
- Attributed the code change to the original author @ipChrisLee on the commit


carlosala:
Thanks!

PR #12113, opened 2023-12-22
feat(aws): add sso logout capabilities

krakenbite:
The aws plugin gives the user the possibility to login into the sso session when using the AWS profile selection. Unfortunately, it is not possible to also logout this way. Naturally, one would think that when one is typing "asp <aws_profile> login" it is also possible to type "asp <aws_profile> logout" but this is currently not the case. This commit adds this possibility and therefore enables the user to login and logout with the plugin without having to use awscli directly.

## Standards checklist:

- [] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- Add aws sso logout possibility for aws plugin

## Other comments:

...


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

krakenbite:
@carlosala You are welcome, thanks for the quick response! I wish you a wonderful rest of the year and a happy new year!

PR #12139, opened 2023-12-30
feat(git): add aliases to pull from upstream with `--rebase` and `--autostash`

hasecilu:
## Standards checklist:

## Changes:

- Add new alias analogous to the currently `gluc` and `glum`

## Other comments:

Currently we have alias to pull from upstream to our current/main branch
but if you have changes in the branch you may need to stash
first those changes and then pull.

Having alias with `--rebase` and `--autostash` flags let you pull code with a single command.

When you use `gluc`/`glum` you need to have a clean branch in order to succesfully.

Replacing `gluc` with `glucra`/`glucrav` and  `glum` with `glumra`/`glumrav` lets you pull remote branches even when you have a minor change in your local branch.

Instead of doing `gsta && glum && gstp` you just need to do `glumra`.

Inpired by `gupav` now `gprav`.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -284,7 +284,11 @@ function ggl() {
 compdef _git ggl=git-checkout
 
 alias gluc='git pull upstream $(git_current_branch)'
+alias glucra='git pull upstream $(git_current_branch) --rebase --autostash'
+alias glucrav='git pull upstream $(git_current_branch) --rebase --autostash -v'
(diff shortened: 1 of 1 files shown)


thesayfulla (approved):
I think not bad feature

carlosala:
Hi, thanks for your contribution!
Similar to https://github.com/ohmyzsh/ohmyzsh/pull/11871#issuecomment-1703824705, it's probably better to enforce it at the user configuration level!
Thanks!

PR #12143, opened 2024-01-03
fix(rtx)!: rename `rtx` plugin to `mise`

jaredallard:
## Standards checklist:

## Changes:

- Renames the `rtx` plugin to `mise` to reflect the new project name.

## Other comments:

This is a BREAKING CHANGE because the `rtx` plugin will no longer exist. I decided to not implement any transition logic because the author hasn't provided any wrappers for `rtx` to continue working once `mise` is installed. I feel that ohmyzsh shouldn't provide that.


jaredallard:
Sorry for the plethora of post-open changes. Wanted to update the title and desc to be better :D

nelsonmestevao (approved):

carlosala:
Hi! For sure we need to do so! I'll think about the best option, but removing 100% the old `rtx` plugin is not the best solution. Let's have a transition period, I'll push it here! Thanks for your contribution!

carlosala:
Thanks for your contribution!

nelsonmestevao:
I am getting at zsh start "] command not found" because of this plugin. Any idea why?

carlosala:
I'm not able to reproduce. Could you send the session initialization logs (check how to [here](https://github.com/ohmyzsh/ohmyzsh/wiki/Troubleshooting#other-problems)) and send them to logs@ohmyz.sh?
Thanks!

PR #12147, opened 2024-01-05
fix(macos): print usage for `man-preview` with no args

jkaving:
## Standards checklist:

## Changes:

Instead of just running `man` with no argument, which leads to it printing the question `What manual page do you want?`, we now print a usage statement if no arguments are provided.

**Before this change**
```
> man-preview
What manual page do you want?
```
The command then hangs and you need to use Ctrl-C to abort it.

**After this change**
```
> man-preview
Usage: man-preview command1 [command2 ...]
>
```

### DIFF
--- a/plugins/macos/macos.plugin.zsh
+++ b/plugins/macos/macos.plugin.zsh
@@ -224,6 +224,8 @@ function quick-look() {
 }
 
 function man-preview() {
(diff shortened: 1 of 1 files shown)


carlosala (commented):
LGTM 🚀

carlosala (approved):

carlosala:
Thanks!

PR #12148, opened 2024-01-06
fix(rkj-repos): check if `~/.hgrc` exists

ChaosInventor:
If the `hg` command is available but `~/.hgrc` does not exist, the theme would print `grep: ~/.hgrc: No such file or directory` before every prompt line.

## Standards checklist:

## Other comments:

Could be also achieved by redirecting grep's stderr to `/dev/null` or adding the `-s` flag. Not sure which approach is best, please comment.


carlosala:
Thanks for your contribution!

PR #12150, opened 2024-01-07
Tweak exit code to avoid exiting scripts using "omz update"

CortezJEL:
Tweak exit code to avoid exiting scripts using "omz update" on successful exit

## Standards checklist:

- [ X] The PR title is descriptive.
- [ X] I have read the contribution guide and followed all the instructions.
- [ X] The code follows the code style guide detailed in the wiki.
- [ X] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- add an if statement to check exit code prior to running "exit" and skip if exit code is 0

## Other comments:

This is helpful as if someone runs `omz update ; reboot` on successful update the exit commands stops the continuation of the command

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -281,5 +281,8 @@ case "$resetAutoStash" in
   *) git config rebase.autoStash "$resetAutoStash" ;;
 esac
 
-# Exit with `1` if the update failed
-exit $ret
+# Let script exit without exit command if update succeeds to allow ease of external automated updates
+if [ "$ret" -ne 0 ]; then
+    # Exit with exit code if the update failed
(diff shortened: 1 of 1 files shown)


carlosala:
I can't reproduce. That specific `exit` only applies to the subshell created [here](https://github.com/ohmyzsh/ohmyzsh/blob/428866e28def25593e7dd73f1c3bd909db47635b/lib/cli.zsh#L783).
In case you want to do some kind of automated update, make sure you pass the `omz update --unattended` flag to avoid reloading the shell in case there were updates. Otherwise I don't fully understand which is the issue here.
Thanks for the report!

Issue #12153, opened 2024-01-09
Aliasing ipython overrides command line arguments


stefanv:
### Describe the bug

[Here](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/python/python.plugin.zsh#L47C16-L47C89), oh-my-zsh overrides the command for IPython.

When IPython is installed in the current virtual environment, this is not necessary, since the ipython command is already correctly installed.

But adding this aliases changes behavior. It would be better not to create the alias, since it does not add any real advantage.

### Steps to reproduce

 To reproduce, run:

```
ipython --ignore-cwd
```

with, and without, the alias, and inside IPython print `sys.path`.

The flag is ignored when the alias is defined.

### Expected behavior

No alias should be created that impacts the behavior of the `ipython` command.

### OS / Linux distribution

Fedora 36

### Zsh version

5.9

### Terminal emulator

Alacritty

### If using WSL on Windows, which version of WSL

None

gautamkrishnar:
I will take a look at this

carlosala:
Thanks for the report, and thanks Gautam as well! It should be fixed now!

PR #12159, opened 2024-01-13
fix(python): add support for cli args

gautamkrishnar:
## Standards checklist:

## Changes:
Added command line args support to ipython alias, fixes: #12153

## Other comments:
n/a

### DIFF
--- a/plugins/python/python.plugin.zsh
+++ b/plugins/python/python.plugin.zsh
@@ -44,7 +44,7 @@ function pyuserpaths() {
 alias pygrep='grep -nr --include="*.py"'
 
 # Run proper IPython regarding current virtualenv (if any)
-alias ipython='python3 -c "import IPython, sys; IPython.start_ipython(argv=sys.argv[1:])"'
+alias ipython='python3 -c "import IPython, sys; sys.exit(IPython.start_ipython())"'
 
 # Share local directory as a HTTP server
 alias pyserver="python3 -m http.server"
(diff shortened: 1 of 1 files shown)


plugins/python/python.plugin.zsh
```diff
@@ -44,7 +44,7 @@ function pyuserpaths() {
 alias pygrep='grep -nr --include="*.py"'
 
 # Run proper IPython regarding current virtualenv (if any)
-alias ipython="python3 -c 'import IPython; IPython.terminal.ipapp.launch_new_instance()'"
+alias ipython='python3 -c "import IPython, sys; IPython.start_ipython(argv=sys.argv[1:])"'
```
carlosala, line 47: It seems that it's not needed to pass `argv`, it should already parse them: https://ipython.readthedocs.io/en/stable/api/generated/IPython.html#IPython.start_ipython

carlosala:
Even more, this is the script that launches `ipython` in my system (`v8.2.0`).
```python
#!/usr/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from IPython import start_ipython
if __name__ == "__main__":
    sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])
    sys.exit(start_ipython())
```
We could simplify it to:
```zsh
alias ipython='python3 -c "import IPython, sys; sys.exit(IPython.start_ipython())"'
```

plugins/python/python.plugin.zsh
gautamkrishnar: Interesting 😄

gautamkrishnar:
@carlosala updated the alias to your suggestion. Please squash the pr before merging.

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

gautamkrishnar:
Thanks for reviewing.

PR #12165, opened 2024-01-16
feat(web-search): add `deepL`

KaitoMuraoka:
## Standards checklist:

## Changes:

- added alias for DeepL URL: `https://www.deepl.com/translator#auto/auto/`

## Other comments:

I discovered that when I input a specific string into the URLs of DeepL, they automatically translate it for me.
I often use web-search.plugin and frequently use DeepL as well. Therefore, thinking it would be convenient to be able to translate from web-search.plugin to DeepL, I sent a request.

DeepL automatically translates text when you add a string after `https://www.deepl.com/translator#auto/auto/`.
 For example, like this:
[https://www.deepl.com/translator#auto/auto/こんにちは世界](https://www.deepl.com/translator#auto/auto/こんにちは世界)

### DIFF
--- a/plugins/web-search/README.md
+++ b/plugins/web-search/README.md
@@ -45,6 +45,7 @@ Available search contexts are:
 | `scholar`             | `https://scholar.google.com/scholar?q=`         |
 | `ask`                 | `https://www.ask.com/web?q=`                    |
 | `youtube`             | `https://www.youtube.com/results?search_query=` |
+| `deepl`               | `https://www.deepl.com/translator#auto/auto/`   |
 
 Also there are aliases for bang-searching DuckDuckGo:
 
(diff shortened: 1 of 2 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

loicraux:
I get + symbols between words in deepl.com URL and text area, is that normal ?

If I try `deepl how are you` in terminal, https://www.deepl.com/translator#en/fr/how%2Bare%2Byou gets opened, and DeepL attempts to translate `how+are+you`

Same happens if I try `deepl "how are you"` or `deepl 'how are you'` in terminal...

carlosala:
@loicraux fixed in latest version. Thanks for your contribution!

PR #12175, opened 2024-01-22
fix(aws): pass default region to fetch regions

gdlx:
## Standards checklist:

## Changes:

- Sets a default region for the `aws ec2 describe-regions` call in `aws_regions` to avoid this situation:
```
# asr eu-west-1

You must specify a region. You can also configure your region by running "aws configure".

You must specify a region. You can also configure your region by running "aws configure".
Available regions:
```

## Other comments:

I've used `$AWS_DEFAULT_REGION` and `$AWS_REGION` as regions for the call in case they are set.
I think we could always use any region like `us-west-1` but I can't test if the list is different when using an "exotic" source region.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -202,8 +202,17 @@ function aws_change_access_key() {
 }
 
 function aws_regions() {
+  local -a region
+  if [[ $AWS_DEFAULT_REGION ]];then
+      region="$AWS_DEFAULT_REGION"
+  elif [[ $AWS_REGION ]];then
+      region="$AWS_REGION"
+  else
+      region="us-west-1"
+  fi
+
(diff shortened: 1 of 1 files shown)


plugins/aws/aws.plugin.zsh
```diff
@@ -202,8 +202,17 @@ function aws_change_access_key() {
 }
 
 function aws_regions() {
+  local -a region
```
carlosala, line 205: Why is this an array? Shouldn't be a regular variable?

carlosala (changes requested):

plugins/aws/aws.plugin.zsh
gdlx: You're right, sorry.
It's fixed.

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12186, opened 2024-01-29
feat(ssh): add plugin

ferdnyc:
## Standards checklist:

## Changes:

- This is an updated version of #2707, applying a commit which addresses review feedback on top of the commits from the original author @elias5000

## Other comments:

- Regarding the second and third checkboxes, obviously this does duplicate an open PR, and while I've READ the contributors guide, this PR doesn't follow all instructions. The commits lack Conventional Commits style titles.
- Regarding the seventh checkbox, I haven't tested all of this code myself, as I don't use most of these features.

cc: @carlosala

### DIFF
--- a/plugins/ssh/ssh.plugin.zsh
+++ b/plugins/ssh/ssh.plugin.zsh
@@ -4,7 +4,7 @@
 # Filter out wildcard host sections.
 local hosts
 if [[ -f $HOME/.ssh/config ]]; then
-  hosts=($(cat $HOME/.ssh/config | egrep '^Host.*' | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//' | grep '\.'))
+  hosts=($(egrep '^Host.*' $HOME/.ssh/config | awk '{print $2}' | grep -v '^*' | sed -e 's/\.*\*$//'))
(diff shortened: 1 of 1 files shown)


plugins/ssh/ssh.plugin.zsh
```diff
@@ -0,0 +1,46 @@
...
+
+############################################################
+# Remove SSH key from agent
+function ssh_unload_key {
+  local key="$1"
+  if [[ -z "$key" ]]; then return; fi
+  local keyfile="$HOME/.ssh/$key"
+  local keysig=$(ssh-keygen -k -l "$keyfile")
+  if ( ssh-add -l | grep -q "$keysig" ); then
+    ssh-add -d "$keyfile"
```
ferdnyc, line 44: Note that this may fail, as not all SSH Agent implementations support removing identities. When the agent is gnome-keyring, for example, any delete requests will be met with:

```console
$ssh-add -v -d $file
Could not remove identity "$file": agent refused operation
```

...I don't really know what to do about that, just silently ignoring it seems the only option.

carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for the contribution and adapt of PR!

PR #12194, opened 2024-02-03
fix(sublime): pass user's env to `sst`

Rayzeq:
## Standards checklist:

## Changes:

- Changed the alias `sst` from `sudo subl` to `sudo -EH subl`

## Other comments:

On some (all ?) wayland environnements, like Hyprland,  `sudo subl` is not able to open a new window (although it can open files in an already open window). This fix allow it to work.

### DIFF
--- a/plugins/sublime/sublime.plugin.zsh
+++ b/plugins/sublime/sublime.plugin.zsh
@@ -4,7 +4,7 @@ alias st=subl
 alias stt='subl .'
 
 # Define sst only if sudo exists
-(( $+commands[sudo] )) && alias sst='sudo subl'
+(( $+commands[sudo] )) && alias sst='sudo -EH subl'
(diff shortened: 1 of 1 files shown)


carlosala (changes requested):
I generally like it, I don't think that passing the env would be an issue. I'd say `-H` is not needed, does it work just with `-E`?

Rayzeq:
The `-H` is necessary. I have no idea why, but sublime doesn't launch without it (maybe because it can't find `~/.config/sublime-text/`, but I'm not sure).

carlosala:
Let's go for it, shouldn't be an issue!

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12203, opened 2024-02-06
feat(envu) add envu command that updates env vars

GNUSheep:
## Standards checklist:

- [X ] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

This plugin adds `envu` command that updates environment variables.

## Other comments:

...


carlosala:
Hi! I like the idea! Even though I think having an extra plugin to update an env variable does not pay off, and it has a lot of corner cases, I encourage you to keep it in your custom directory.
Thanks for your contribution!

PR #12208, opened 2024-02-08
feat(kube-ps1): add symlink support

hedgieinsocks:
## Standards checklist:

## Changes:

Add `KUBE_PS1_KUBECONFIG_SYMLINK` variable that defaults to `false`. This varialble will control the behaviour of `stat` and `zstat` commands. Users will be able to set the var to `true` and thus force the plugin to evaluate the mtime of the symlink instead of the real file. This will be useful when `~/.kube/config` is actually a symlink to a real kubeconfig, and the user regularly updates the symlink to point to another kubeconfig. 

## Other comments:

There was a pr https://github.com/ohmyzsh/ohmyzsh/issues/8786 that actually broke kube-ps1 for me. And this pr is an attempt to find a compromise without breaking anything.


hedgieinsocks:
@jonmosco hi, tagging you as the plugin author

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12209, opened 2024-02-10
feat(git): add `grget` alias

NoamPrag:
## Standards checklist:

## Changes:

- Alias `grget` added for `git remote get-url`.

## Other comments:

- This doesn't collide with any other defined aliases/functions.
- `grset` is already aliased to `git remote set-url`, so it'll be  convenient to have `grget` as well.


carlosala:
Hi! I'd say this alias does not add a lot of value. You can still use `grv` to retrieve all the remotes (typically, you wouldn't have more than 2).
Thanks for your contribution!

PR #12227, opened 2024-02-20
fix(ssh): allow multiple definitions per host

Y3K:
With the small changes in this commit we split all the host values in lines, sort them out and make a unique array of them.

## Standards checklist:

## Changes:

- Update SSH plugin.

## Other comments:

The current completion doesn't work correctly when there are multiple hosts per line, e.g.:

```
Host foo bar baz
```

SSH config also allows to have repeated hosts, e.g.:

```
Host foo bar
  ...

Host foo
  ...
```


carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for your contribution, really appreciated!

PR #12232, opened 2024-02-21
fix(tmux): do not pass empty flags to aliases

### LINKED ISSUE #12230: tmux attach-session -t expects an argument
### Describe the bug

The tmux plugin defines the alias `ta` as `tmux attach-session -t`. With Tmux 3.4, I get the error `command attach-session: -t expects an argument`. Could be due to a recent upgrade from 3.3a -> 3.4 or just something I've done on my end.

I would suggest just omitting the `-t` flag and aliasing "ta" to `tmux attach-session`, but I imagine there's a reason that's been included?

### Steps to reproduce

1. Running tmux 3.4.
2. Enable plugin tmux.
3. Run `tmux attach-session -t` or use the alias `ta`.

### Expected behavior

Tmux should attach to the last opened session.

_No response_

### OS / Linux distribution

Nixos 24.05 (Unstable)

### Zsh version

5.9

### Terminal emulator

Kitty

None

### Additional context

_No response_

detroyejr:
## Standards checklist:

## Changes:

- Removes the -t flag from the aliases. (Fixes #12230)

## Other comments:

If passing -t does something important, this may not be ideal. AFAICT `tmux attach-session` and `tmux attach-session -t` are identical.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -47,7 +47,7 @@ fi
 : ${ZSH_TMUX_UNICODE:=false}
 
 # ALIASES
-function _build_alias {
+function _build_tmux_alias {
   eval "function $1 {
     if [[ \${1::1} == '-' ]]; then
       tmux $2 \"\$@\"
@@ -61,10 +61,12 @@ alias tksv='tmux kill-server'
 alias tl='tmux list-sessions'
 alias tmuxconf='$EDITOR $ZSH_TMUX_CONFIG'
 
-_build_alias "ta" "attach" "-t"
-_build_alias "tad" "attach -d" "-t"
-_build_alias "ts" "new-session" "-s"
-_build_alias "tkss" "kill-session" "-t"
+_build_tmux_alias "ta" "attach" "-t"
+_build_tmux_alias "tad" "attach -d" "-t"
+_build_tmux_alias "ts" "new-session" "-s"
+_build_tmux_alias "tkss" "kill-session" "-t"
+
+unfunction _build_tmux_alias
 
 # Determine if the terminal supports 256 colors
 if [[ $terminfo[colors] == 256 ]]; then
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -49,7 +49,7 @@ fi
 # ALIASES
 function _build_tmux_alias {
   eval "function $1 {
-    if [[ \${1::1} == '-' ]]; then
+    if [[ -z \$1 ]] || [[ \${1::1} == '-' ]]; then
       tmux $2 \"\$@\"
     else
       tmux $2 $3 \"\$@\"
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -37,13 +37,42 @@ fi
 : ${ZSH_TMUX_UNICODE:=false}
 
 # ALIASES
-alias ta='tmux attach -t'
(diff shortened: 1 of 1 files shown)


carlosala:
I'm not 100% sure how to proceed here. I don't use `tmux,` so I'm unsure if this will break the user's workflows. Does in `tmux` 3.3a the commands work the same with `-t` and without it? I kinda understand the purpose of it; if you want a specific session, you just add it, and if not, you leave it there. It's unfortunate that `tmux` decided to make that argument strict.

detroyejr:
Ah, yes I think that's correct. As-is this change would break some functionality. 

I've asked [here](https://github.com/tmux/tmux/issues/3836#issuecomment-1959596650) to see if anyone might have a good idea.

pepoluan:
> Ah, yes I think that's correct. As-is this change would break some functionality.
> 
> I've asked [here](https://github.com/tmux/tmux/issues/3836#issuecomment-1959596650) to see if anyone might have a good idea.

I responded in that issue, that I think the alias needs to be turned into a function so it can leverage the `${VAR:+sub}` pattern.

For example, the `ta` alias will need to be converted to:

```zsh
function ta() {
    tmux attach-sessions ${1:+-t }$1
}
``

**Comment by detroyejr** [comment]
Thank you for the suggestion!

4 of the 7 aliases would need to be replaced with functions in order to work with 4.4. I've swapped those aliases out with something that I believe is the correct use of that pattern (let me know if you think otherwise). Looks like it works with both 3.3a and 3.4.

**Comment by carlosala** [comment]
Does it happen the same with `-s`?

**Comment by detroyejr** [comment]
It does. `tmux new-session -s` without any arguments throws `command new-session: -s expects an argument`. I replaced that as well.

**Review by carlosala — CHANGES_REQUESTED** [review]
I like it! Could you add a check to verify that `$1` does not start with `-`? Then we can avoid situations like `ts -d` being converted to `tmux new-session -s -d` and crashing in `3.4`.

**Comment by detroyejr** [comment]
Running `tmux new-session -s -d` creates a session named "-d". Others like `ta -d` attempts to attach to "-d" which doesn't feel exactly like a crash but I can add this in if we think this result is confusing.

Could you clarify what you think the result should be if a flag is passed?

If someone tries to run `ts -d`, should the response be some sort of message and early exit. For example,

```
function ts() {
  if [[ $1 == -* ]]; then
    print "Error: argument should not start with -" >&2
    return 1
  else
    tmux new-session ${1:+-s }$1
  fi
}
```

Or maybe just run `tmux <command> $1` and not pass the `-s` or `-t` flags?

carlosala:
Yep, I was thinking about not passing the `-s` flag and just pass `$1` if it's a flag. Actually, we should bind it to `$@`

detroyejr:
Gave it a go. Let me know if that doesn't accomplish what you asked for and I can try something else.

carlosala:
Thanks for your contribution!

robbyrussell:
Much appreciated, @detroyejr! ✌️

MoetaYuko:
This appears to break tab completion. For instance, "ta\<tab\>" provided completion for open sessions prior to this change. Is there a workaround for it?

detroyejr:
Thanks for pointing this out @MoetaYuko. I've submitted the PR mentioned above that should fix tab completion.

MoetaYuko:
Thanks for the fix.

MoetaYuko:
@detroyejr @matthew-nm I proposed https://github.com/ohmyzsh/ohmyzsh/pull/12468 which wraps zsh's built-in completion function, please give it a try.

matthew-nm:
@moetayuko Works great! Thanks!

PR #12234, opened 2024-02-22
feat(git-commit): added new flags

rejmann:
## Standards checklist:

## Changes:
- [Commit message with ! to draw attention to breaking change](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change)
```zsh
git <type> [(-a, --attention)] "<message>"
git <type> [(-s, --scope) "<scope>"] [(-a, --attention)] "<message>"
git <type> [(-sa, -as)] "<scope>" "<message>"
```
- Update of documentation related to types

### DIFF
--- a/plugins/git-commit/git-commit.plugin.zsh
+++ b/plugins/git-commit/git-commit.plugin.zsh
@@ -24,19 +24,19 @@ _git_commit_flags=(
   '-as'
 )
 
-remove_flags() {
-  local regex=$(
-    IFS="|"
-    echo "${_git_commit_flags[*]}"
-  )
-
-  echo $(sed -E "s/\s($regex)\s|\s($regex)$//g" <<<"$1")
-}
-
 build_message() {
   local argv=("$@")
   local argc=${#@}
   local i=1
+
+  remove_flags() {
+    local regex=$(
+      IFS="|"
+      echo "${_git_commit_flags[*]}"
+    )
+
+    echo $(sed -E "s/\s($regex)\s|\s($regex)$//g" <<<"$1")
+  }
 
   while [[ $i -lt $argc ]]; do
     local flag="${argv[$i]}"
--- a/plugins/git-commit/git-commit.plugin.zsh
+++ b/plugins/git-commit/git-commit.plugin.zsh
@@ -14,7 +14,6 @@ _git_commit_aliases=(
   'wip'
 )
 
-local -a _git_commit_flags
 _git_commit_flags=(
   '-s'
   '--scope'
@@ -33,7 +32,7 @@ remove_flags() {
   message=$(sed -E "s/\s($regex)\s|\s($regex)$//g" <<< "$message")
 
   echo "$message"
-}
+};
 
 build_message () {
   if [[ "$*" =~ (-sa|-as) ]]; then
@@ -77,4 +76,4 @@ for type in "${_git_commit_aliases[@]}"; do
   fi
 done
 
-unset _git_commit_aliases _flags alias type
+unset _git_commit_aliases _git_commit_flags alias type
--- a/plugins/git-commit/git-commit.plugin.zsh
+++ b/plugins/git-commit/git-commit.plugin.zsh
@@ -29,12 +29,6 @@ build_message() {
   local argc=${#@}
   local i=1
 
-  remove_flags() {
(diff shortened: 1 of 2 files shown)


plugins/git-commit/git-commit.plugin.zsh
```diff
@@ -14,6 +14,56 @@ _git_commit_aliases=(
   'wip'
 )
 
+_git_commit_flags=(
```
carlosala, line 17: Could you `typeset -a _git_commit_flags` before? It shouldn't be necessary, but it's always nice to do it!

plugins/git-commit/git-commit.plugin.zsh
```diff
@@ -14,6 +14,56 @@ _git_commit_aliases=(
...
+  '-s'
+  '--scope'
+  '-a'
+  '--attention'
+  '-sa'
+  '-as'
+)
+
+remove_flags() {
+  local regex=$(IFS="|"; echo "${_git_commit_flags[*]}")
```
carlosala, line 27: There's a better way to do it without spawning a subshell!
```suggestion
  local regex="${(j:|:)_git_commit_flags}"
```

plugins/git-commit/git-commit.plugin.zsh
```diff
@@ -14,6 +14,56 @@ _git_commit_aliases=(
...
+
+  echo $(sed -E "s/\s($regex)\s|\s($regex)$//g" <<< "$1")
```
carlosala, line 29: It should be enough to use `*` (matches between zero and unlimited times!). Let's add `-n` as well!
```suggestion
  echo -n $(sed -E "s/\s*($regex)\s*//g" <<< "$1")
```

plugins/git-commit/git-commit.plugin.zsh
```diff
@@ -14,6 +14,56 @@ _git_commit_aliases=(
...
+
+    if [[ "$flag" =~ (-s|--scope) && -n "$value" ]]; then
+      local scope="($value)"
+      i=$next
+    fi
+  i=$next
+  done
+
+  local template="'$type'${scope}$attention: ${@}";
+  local message=$(remove_flags "$template" "${_git_commit_flags[@]}")
```
carlosala, line 62: Since `remove_flags` is only used once, could you embed it here?

carlosala (changes requested):
I like the idea to extend it! Could you address this issues?
Thanks!

plugins/git-commit/git-commit.plugin.zsh
rejmann: Very good, it makes sense!

plugins/git-commit/git-commit.plugin.zsh
rejmann: Interesting, I liked that!

plugins/git-commit/git-commit.plugin.zsh
rejmann: Yes I can for sure!

plugins/git-commit/git-commit.plugin.zsh
carlosala: Indeed, it's a `zsh` syntax thing

plugins/git-commit/git-commit.plugin.zsh
carlosala: Adding `noglob` to avoid `*` being expanded!

plugins/git-commit/git-commit.plugin.zsh
rejmann: Got it, cool!

plugins/git-commit/git-commit.plugin.zsh
rejmann: Thank you, I understood differently!
> What I mean here is to do the following directly:

carlosala:
I was actually testing this changes and they don't work the way they are coded, since you need to pass the actual code of the function to git, and not the function itself.
I like the idea though, I'll try to make the code easier to maintain.

carlosala:
Thanks for your idea, I gave you the authorship of the commit! I added a bit of stability and refactor the function as well, thanks!

PR #12237, opened 2024-02-23
feat(eza): create plugin

pepoluan:
The `eza` plugin provides some configurable aliases to simplify usage of the [`eza`](https://github.com/eza-community/eza) tool.

## Standards checklist:

## Changes:

* Creates 5 aliases that are very common to invoke when using the `ls` command, but implemented using `eza` instead
* Allows limited configuration for `eza` defaults through usage of `zstyle` (see `README.md` for info)

### DIFF
--- a/plugins/eza/README.md
+++ b/plugins/eza/README.md
@@ -86,13 +86,18 @@ Default: Not set, which means the default behavior of `eza` will take place.
 
 ## Aliases
 
-Note that aliases may be modified by Configuration.
+**Notes:**
+
+* Aliases may be modified by Configuration
+* The term "files" without "only" qualifier means both files & directories
 
 
 | Alias   | Command           | Description                                                                 |
 | ------- | ----------------- | --------------------------------------------------------------------------- |
 | `la`    | `eza -la`         | List all files (except . and ..) as a long list                             |
-| `ldot`  | `eza -ld .*`      | List all dotfiles (directories shown as entries instead of recursed into)   |
+| `ldot`  | `eza -ld .*`      | List dotfiles only (directories shown as entries instead of recursed into)  |
+| `lD`    | `eza -lD`         | List only directories (excluding dotdirs) as a long list                    |
+| `lDD`   | `eza -laD`        | List only directories (including dotdirs) as a long list                    |
 | `ll`    | `eza -l`          | List files as a long list                                                   |
 | `ls`    | `eza`             | Plain eza call                                                              |
 | `lS`    | `eza -l -ssize`   | List files as a long list, sorted by size                                   |
--- a/plugins/eza/README.md
+++ b/plugins/eza/README.md
@@ -20,9 +20,9 @@ All configurations are done using the `zstyle` command in the `:omz:plugins:eza`
 zstyle ':omz:plugins:eza' 'dirs-first' $BOOL
 ```
 
-If `1`, directories will be grouped first.
+If `true`, directories will be grouped first.
 
-Default: `0`
+Default: `false`
 
 
 ### `showgroup`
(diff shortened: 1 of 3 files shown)


plugins/eza/eza.plugin.zsh
```diff
@@ -0,0 +1,46 @@
...
+  # Get the tail long-options
+  zstyle -s ':omz:plugins:eza' 'dirs-first' _val
+  if [[ $_val == 1 ]]; then
+    _EZA_TAIL+=("--group-directories-first")
+  fi
+  zstyle -s ':omz:plugins:eza' 'time-style' _val
+  if [[ $_val ]]; then
+    _EZA_TAIL+=("--time-style='$_val'")
+  fi
+}
```
carlosala, line 25: I'd instead use the same way of `zstyle` as in https://github.com/ohmyzsh/ohmyzsh/blob/c891b7e6ada5427f4c8cca05c8ed1ae33cb49943/plugins/yarn/yarn.plugin.zsh#L1-L11.
With `-T` is truthy by default, and `-t` is falsy by default.

carlosala (changes requested):
I think it'd be better to keep arguments separated, instead of trying to join `$2` with the head. For example, if `head` and `$2` are empty, we end up with `eza - <tail>`

pepoluan:
> I think it'd be better to keep arguments separated, instead of trying to join `$2` with the head. For example, if `head` and `$2` are empty, we end up with `eza - <tail>`

If it's within a function then I agree. But I'd like to keep it compact because it will become an alias, and for me it's a bit cluttery if the head (single-char options) part are separated.

But you do highlight an edge case that I had overlooked; my next commit will fix this.

plugins/eza/eza.plugin.zsh
pepoluan: Ahh good idea! Never knew about the `-T`|`-t` flag before!

pepoluan: Implemented in commit [c7c7dd8](https://github.com/ohmyzsh/ohmyzsh/pull/12237/commits/c7c7dd81ac8cb7366c34946d980d104474adc397)

pepoluan (commented):
All suggested changes have been considered and adapted.

plugins/eza/README.md
carlosala, line 1: Could you use `yes/no` instead of `true/false`? That's the design that we are using across the project!

carlosala (changes requested):

carlosala:
Do you want to be the CODEOWNER for the plugin? Then you'll be notified every time someone modifies it! Feel free to modify https://github.com/ohmyzsh/ohmyzsh/blob/master/.github/CODEOWNERS and add you!

pepoluan:
> Do you want to be the CODEOWNER for the plugin? Then you'll be notified every time someone modifies it! Feel free to modify https://github.com/ohmyzsh/ohmyzsh/blob/master/.github/CODEOWNERS and add yourself!

Ah okay! Going to add myself there. Thanks!

plugins/eza/README.md
pepoluan: Ah okay, I can change the doc. Using `yes/no` will also be picked up by `-T/-t` right?

pepoluan: I will miss `vim` highlighting the `true|false` value though...

image
image

But I'll defer to the current convention.

Implemented in commit [a8e7178](https://github.com/ohmyzsh/ohmyzsh/pull/12237/commits/a8e7178fe7411a7f31bfb7aecf0f5f6da6b19327)

carlosala (approved):
I changed the `showgroup` to `show-group` for consistency! LGTM! 🚀

PR #12244, opened 2024-02-28
Adds plugin for git interactive checkout

adamsir:
## Standards checklist:

## Changes:

- plugin which allows to do a quick switch between git branches

## Other comments:

...

### DIFF
--- a/plugins/git-checkout-interactive/README.md
+++ b/plugins/git-checkout-interactive/README.md
@@ -1,0 +1,15 @@
+# lol
+
+This plugin adds quick switch between your branches.
+
+To use it, add `git-checkout-interactive` to the plugins array in your `.zshrc` file:
+
+```zsh
+plugins=(... git-checkout-interactive)
+```
+
+## Usage Examples
+
+```sh
+gci
+```
--- a/plugins/git-checkout-interactive/git-checkout-interactive.plugin.zsh
+++ b/plugins/git-checkout-interactive/git-checkout-interactive.plugin.zsh
@@ -1,0 +1,56 @@
+#######################################
+#       git checkout interactive      #
+#######################################
+
+function git-checkout-interactive() {
+  # Get all branches sorted by committer date, along with their last commit hash
+  local branches
+  branches=$(git for-each-ref --sort=-committerdate --format='%(refname:short) %(objectname:short)' refs/heads/)
+
+  # Parse branches
+  local branch_list=()
+  local current_branch
+  current_branch=$(git rev-parse --abbrev-ref HEAD)
+  if [[ "$current_branch" == "" ]]; then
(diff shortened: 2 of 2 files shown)


plugins/git-checkout-interactive/README.md
```diff
@@ -0,0 +1,15 @@
+# lol
```
LunarWatcher, line 1: This needs to be replaced with a proper title

plugins/git-checkout-interactive/git-checkout-interactive.plugin.zsh
```diff
@@ -0,0 +1,57 @@
...
+    local choice
+    read -r choice
+    if (( choice > 0 && choice <= ${#branch_list[@]} )); then
+      local selected_branch="${branch_list[$((choice))]}"
+      local target_branch="${selected_branch//->}"
+      target_branch="${target_branch%% *}"
+      git checkout "$target_branch"
+      break
+    else
+      break
```
LunarWatcher, line 52: This should arguably return a non-zero code, as it's an error state. Even then, why is this a loop at all? Both branches break, so there's no looping in any scenario

plugins/git-checkout-interactive/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This plugin adds quick switch between your branches.
```
LunarWatcher, line 3:
```suggestion
This plugin adds a command that lets you interactively switch between branches.
```

plugins/git-checkout-interactive/git-checkout-interactive.plugin.zsh
```diff
@@ -0,0 +1,57 @@
+#######################################
+#       git checkout interactive      #
+#######################################
+
+function git-checkout-interactive() {
+  local ITEMS_TO_SHOW=10
```
LunarWatcher, line 6: This doesn't necessarily need to be changed (it's not a problem, nor something I consider a significant problem), but why have a function-local constant as a variable if it can't be configured in some way?

plugins/git-checkout-interactive/git-checkout-interactive.plugin.zsh
LunarWatcher: If this was supposed to be a `continue` and not a `break`, the `echo -n` on line 39 should be moved into the loop as well

LunarWatcher (changes requested):
On top of the other comments, it's probably also a good idea to error check the git commands. Outside a git repo, at least two git commands are executed before the function stops:
```
❯ gci
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
```

PR #12248, opened 2024-03-01
feat(python): add auto venv activation

rcrews:
## Standards checklist:

## Changes:

- Adds and runs a function that listens for directory changes. When cd'ing into a directory
  containing `venv/bin/activate`, it automatically activates the virtual environment.
  When cd'ing out of the directory, it automatically deactivates the virtual environment.
- It supports two global environment variables,
  - `VENV_NAME`  where a name different from `venv` can be specified. (For example,
    `.venv` is also suggested by the Python docs.)
  - Also, if anyone doesn't want this, they can simply set `DISABLE_AUTO_VRUN` and the listener
    will always exit without making any changes.

## Other comments:

Auto-activating the environment is a feature of the Ruby Version Manager (rvm). I really like that feature,
and researched till I could add it to my Python environments. I've been running this on my computer for
about six months without issues.

### DIFF
--- a/plugins/python/README.md
+++ b/plugins/python/README.md
@@ -22,33 +22,18 @@ plugins=(... python)
 
 ## Virtual environments
 
-The plugin provides three utilities to manage Python 3.3+
-[venv](https://docs.python.org/3/library/venv.html) virtual
-environments:
+The plugin provides three utilities to manage Python 3.3+ [venv](https://docs.python.org/3/library/venv.html)
+virtual environments:
 
-- `mkv [name]`: Make a new virtual environment called *name* (else
-  `$PYTHON_VENV_NAME`, else "venv") in the current directory.
+- `mkv [name]`: make a new virtual environment called `name` (default: if set `$PYTHON_VENV_NAME`, else
+  `venv`) in the current directory.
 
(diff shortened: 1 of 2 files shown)


rcrews:
Providing at-mentions to people who have contributed to the `python` plug-in in the last year: @gautamkrishnar @MikuHatsune39 @carlosala

carlosala (changes requested):
I like it! Could you make it opt-in instead of out-out? Thanks!

rcrews:
Thanks for your kind reaction, @carlosala! I added two commits. One does as you requested and makes the new feature opt-in, and improves the description of the new feature in the README.
Since the new feature introduces a couple new configuration environment variables, I updated the two existing venv functions -- `vrun` and `mkv` -- to also use `PYTHON_VENV_NAME`, if it is set.
Let me know if you want me to squash these into a single commit, or if you want me to add "feat(python):" to the commit messages for all of the commits.

carlosala (approved):
LGTM! 🚀

carlosala:
I made some changes to simplify the code and align it with our usual design.
Thanks for your contribution!

robbyrussell:
Hey @rcrews  -- we really appreciate the contribution! ✌️

PR #12249, opened 2024-03-02
fix(tmux):  use the offset parameter when extracting the first letter

zhangtianze1997:
## Standards checklist:

## Changes:

- Use the offset as 0 to extract the first letter when checking whether the parameter is a flag like `-t`.

## Other comments:

My zsh version is v5.8.1 and tmux version is 3.2a. After the change in [fix(tmux): do not pass empty flags to aliases](https://github.com/ohmyzsh/ohmyzsh/pull/12232), I found the alias command `ta server` can not work to attach the session. The error is `ta:1: closing brace expected`. This is caused by the omitted offset parameter. Add 0 as offset can fix it.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -49,7 +49,7 @@ fi
 # ALIASES
 function _build_tmux_alias {
   eval "function $1 {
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
Weird, it should work! Even though, it does not hurt, so let's go for it!
Thanks for your contribution!

PR #12257, opened 2024-03-04
feat(async)!: implement async prompt API and apply to git prompt

### LINKED ISSUE #10869: Add async prompt features
Motivation: async prompt functions have been a thing in Zsh for a while, having this would allows us to fix a lot of performance issues especially in the area of git prompt information.

Steps:

- [x] Research how it works behind the scenes
- [x] Research already implemented solutions
- [x] Design how to architect it for OMZ
- [x] Create a draft Pull Request
- [x] Test
- [ ] Deploy

----

Some implementations to get started:

- https://github.com/woefe/git-prompt.zsh
- https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/async.zsh
- https://github.com/mafredri/zsh-async
- https://github.com/romkatv/zsh-defer
- kinda related: https://github.com/romkatv/gitstatus

Performance testing is adjacent to the problem:

- https://github.com/romkatv/zsh-bench

mcornella:
## Standards checklist:

## Changes:

- Implement generic async prompt feature
- Apply to git prompt by default

## Other comments:

BREAKING CHANGE: the `git_prompt_info` prompt function has been reworked by default to use the new async prompt feature. If you're experiencing issues, you can set the following zstyle setting in your `.zshrc` file before Oh My Zsh is sourced:

```sh
zstyle ':omz:alpha:lib:git' async-prompt no
```

Fixes #10869

### DIFF
--- a/lib/async_prompt.zsh
+++ b/lib/async_prompt.zsh
@@ -1,0 +1,127 @@
+# The async code is taken from
+# https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/async.zsh
+# https://github.com/woefe/git-prompt.zsh/blob/master/git-prompt.zsh
+
+zmodload zsh/system
+
+# For now, async prompt function handlers are set up like so:
+# First, define the async function handler and add the function name
+# to the _omz_async_functions array:
+#
+#  function _git_prompt_status_async {
+#    # Do some expensive operation that outputs to stdout
+#  }
+#  _omz_async_functions+=(_git_prompt_status_async)
+#
+# Then add a stub prompt function in `$PROMPT` or similar prompt variables,
+# which will show the output of "$_OMZ_ASYNC_OUTPUT[handler_name]":
+#
+#  function git_prompt_status {
+#    echo $_OMZ_ASYNC_OUTPUT[_git_prompt_status]
+#  }
+#
+#  RPROMPT='$(git_prompt_status)'
+#
+# This API is subject to change and optimization. Rely on it at your own risk.
+typeset -ga _omz_async_functions
+
+# Set up async handlers and callbacks
+function _omz_async_request {
+  typeset -gA _OMZ_ASYNC_FDS _OMZ_ASYNC_PIDS _OMZ_ASYNC_OUTPUT
+
+  # executor runs a subshell for all async requests based on key
+  local handler
+  for handler in ${_omz_async_functions}; do
+    (( ${+functions[$handler]} )) || continue
+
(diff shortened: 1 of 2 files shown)


lib/git.zsh
carlosala, line 1: We could define the `_omz_git_prompt_status` function outside and de-duplicate the code then! 👍🏻

carlosala (changes requested):
I just realized that if you don't have a `git_prompt_info`-theme you'll still be executing the `git status` thing everytime.
IMO the best way to overtake this issue is by moving the `_omz_async_functions` into its own function that enables the precmd as soon as there's any function appended to it, and the `git_prompt_info` could call there the first time that is executed.
What do you think?

carlosala (approved):
LGTM!🚀

lib/async_prompt.zsh
```diff
@@ -0,0 +1,139 @@
...
+function _omz_register_handler {
+  typeset -ga _omz_async_functions
+  # we want to do nothing if there's no $1 function or we already set it up
+  if [[ -z "$1" ]] || (( ! ${+functions[$1]} )) \
+    || (( ${+_omz_async_functions[$1]} )); then
+    return
+  fi
+  _omz_async_functions+=("$1")
+  # let's add the hook to async_request if it's not there yet
+  if (( ! ${+precmd_functions[_omz_async_request]} )) \
```
mcornella, line 36: This check will always fail because `precmd_functions` is a simple array. Check should be like https://unix.stackexchange.com/a/411331

lib/async_prompt.zsh
```diff
@@ -0,0 +1,139 @@
...
+#
+#  RPROMPT='$(git_prompt_status)'
+#
+# This API is subject to change and optimization. Rely on it at your own risk.
+
+function _omz_register_handler {
+  typeset -ga _omz_async_functions
+  # we want to do nothing if there's no $1 function or we already set it up
+  if [[ -z "$1" ]] || (( ! ${+functions[$1]} )) \
+    || (( ${+_omz_async_functions[$1]} )); then
```
mcornella, line 31: The same here

mcornella (commented):
Last changes required

lib/async_prompt.zsh
carlosala: Yep, I implemented it like in SO but then I realized that we were doing this style in the whole file, for example https://github.com/ohmyzsh/ohmyzsh/pull/12257/files#diff-b16a54a737cad41cb811a6448a1c9064484c98835726f7552cf8f6ab6785e1f9R50

PR #12261, opened 2024-03-06
feat: Implement auto-detection of subexecutor

### LINKED ISSUE #12245: Provide OMZ-wide option to specify subexecutor

_No response_

I am using **Gentoo Linux**, and I don't use `sudo` for subexecutor. Rather, I use [`doas`](https://github.com/slicer69/doas)

Unfortunately, all plugins that use a subexecutor to do their work, are hardcoded using `sudo`.

I propose having an OMZ-wide knob that allows plugins to automatically adjust to whatever subexecutor is being used.

There are several alternatives:

1. Define `ZSH_SUBEXECUTOR` that, by default, is set to `sudo`. Defining this will allow simple search-and-replace for most plugins to replace `sudo` with `${ZSH_SUBEXECUTOR}`
   For example:
   ```zsh
   alias nmap_slow="sudo nmap -sS -v -T1"
   ```
   becomes
   ```zsh
   alias nmap_slow="${ZSH_SUBEXECUTOR} nmap -sS -v -T1"
   ```

2. Define an OMZ-level `zstyle` to contain the name of the subexecutor being used, defaults to `sudo`.
   This will need some code in the plugins. So, to use the above example, will become:
   ```zsh
   typeset _SUBEX
   zstyle -s ':omz' 'subexecutor' _SUBEX
   alias nmap_slow="${_SUBEX} nmap -sS -v -T1"
   unset _SUBEX
   ```

3. Define a global OMZ function that adjusts the call to subexecutor automatically. Let's call it `subex`.
   So the above example becomes:
   ```zsh
   alias nmap_slow="subex nmap -sS -v -T1"
   ```
   The function may internally be referring to an env var and/or a zstyle knob to provide manual override.

An alternative is to add the following code for all plugins:

```zsh
# If in the future a new subexecuter is created, we only need to edit this array
typeset _KNOWN_SUBEXES=( "doas" "sudo" )
typeset _SUBEX
for _i in "${_KNOWN_SUBEXES[@]}"; do
  if command -v "$_i" > /dev/null; then
    _subex="$_i"
     break
   fi
done
if [[ -z $_subex ]]; then
  _subex="sudo"
fi
unset _KNOWN_SUBEXES
```

However, if we add this code to all plugins that define `sudo` as part of their sets of aliases, that is highly inefficient since the code will be duplicated over all impacted plugins.

Furthermore this code does not provide a way to temporarily override the subexecutor being used.

### Additional context

_No response_

### Related Issues

This is tangentially related to this discussion: https://github.com/ohmyzsh/ohmyzsh/discussions/11947 , specifically [this comment](https://github.com/ohmyzsh/ohmyzsh/discussions/11947#discussioncomment-7222913) which mentions the `doas` command.

So if this issue gets solved, the solution will provide partial solution to the challenge discussed, specifically for *BSD users that use the `doas` command.

pepoluan:
## Standards checklist:

## Changes:

- Adds a new library, named `00subexecutor.zsh`. The `00` part is to ensure that it runs first, because it will set a certain `zstyle` that is needed by other (patched) libraries.
- Patches several other libraries formerly hard-coded to use `sudo`, to adapt to whatever subexecuter has been detected in `00subexecutor.zsh`
- Adds a new 'global' function called `subex` that can be used in place of (parameterless) `sudo` in all plugins, if we ever want to do the heavy-lifting of making all plugins have transparent support for non-sudo subexecutor

## Other comments:

This should close #12245 

This PR modifies the core parts of OMZ, it is DIFFERENT from #11395, which has been closed anyways.


lib/00subexecutor.zsh
```diff
@@ -0,0 +1,44 @@
...
+  fi
+  zstyle ':omz' 'subexecutor' "$_SUBEX"
+}
+
+_SetupSubexecutor
+unfunction _SetupSubexecutor
+unset _KNOWN_SUBEXES
+
+# The alias provides a 'hardcoded', invariant subexecutor to use throughout the shell session
+alias _="$_SUBEX "
```
pepoluan, line 36: This is taken out from `lib/misc.zsh` because the definition of the alias now relies completely on what subexecutor is being used.

lib/termsupport.zsh
```diff
@@ -95,8 +95,10 @@ function omz_termsupport_preexec {
     fi
   fi
 
-  # cmd name only, or if this is sudo or ssh, the next cmd
-  local CMD="${1[(wr)^(*=*|sudo|ssh|mosh|rake|-*)]:gs/%/%%}"
+  # cmd name only, or if this is doas/sudo or ssh, the next cmd
+  local _subex
+  zstyle -s ':omz' 'subexecutor' _subex
+  local CMD="${1[(wr)^(*=*|${_subex}|_|subex|mosh|rake|-*)]:gs/%/%%}"
```
pepoluan, line 101: I replaced `sudo` in the filter with `${_subex}|_|subex` for the following reasons:

* The subexecutor is not necessarily `sudo` now (can be `doas`, for instance)
* There's an underscore `_` alias (previously defined in `lib/misc.zsh`) that aliases the subexecutor
* There's a new global function `subex` that also invokes the subexecutor

lib/misc.zsh
```diff
@@ -27,9 +27,6 @@ elif (( ${+commands[more]} )); then
   env_default 'PAGER' 'more'
 fi
 
-## super user alias
-alias _='sudo '
-
```
pepoluan, line 32: Moved to `00subexecutor.zsh`

pepoluan:
Didn't edit `CODEOWNERS` file because I'm not sure if that file only applies to plugins, or can also cover a singular file such as `lib/00subexecutor.zsh`

pepoluan:
One thing I'm not sure is where to document the `zstyle :omz subexecutor` knob.

(Setting that knob before starting up OMZ will disable the auto-detection mechanism and 'locks' the subexecutor for the session.)

pepoluan:
Just synced with latest OMZ master. No conflicts.

README.md
pepoluan, line 1: I have decided to document the subexecutor auto-detection (and its related zstyle-based configuration knob) in OMZ's README.md, especially since there are mentions of other possible zstyle-based configuration in this file.

pepoluan:
1 month since last changes from my part (commit 7ddee97), and so far merging from ohmyzsh:master is no problem, and no conflicts.

So, any blocking issues, @carlosala ?

carlosala:
Hi @pepoluan! It's on our backlog, and we want to rethink it before merging. We're shaping a new architecture for omz that will arrive at some point, and we really want to make this part of it.
I'll get back to you as soon as we decide something!
Thanks for your work!

pepoluan:
> Hi @pepoluan! It's on our backlog, and we want to rethink it before merging. We're shaping a new architecture for omz that will arrive at some point, and we really want to make this part of it. I'll get back to you as soon as we decide something! Thanks for your work!

Ah okay, fair enough.

In the meantime I will keep updating the branch to sync with master.

carlosala:
Hi @pepoluan! I'll close this one for the moment. As I told you some months ago, we still don't want to make such a big change before rethinking the project's architecture.
Thank you so much for all your contributions, let's reopen it if we need it at some point!

PR #12262, opened 2024-03-06
chore(gitfast): remove update script

felipec:
## Standards checklist:

## Changes:

- General update

## Other comments:

I've removed the `update` script since it seems a GitHub action is already taking care of the updating. I also updated the plug-in standard code to match upstream. And I removed the aliases section in the readme, since it only applies to ancient versions of oh-my-zsh.


carlosala (approved):
LGTM!

carlosala:
Thanks!

PR #12264, opened 2024-03-06
enabled showing number of git stashes for themes and applied the change to af-magic theme

jacobmoroni:
## Standards checklist:

## Changes:

- in git.sh I check the stash list and if the theme variables are set, it reports the number of stashed entries in the git status
- in af-magic.sh I set the necessary variables to enable visualizing the stash
- if those are not set, this logic is ignored
## Other comments:

...


jacobmoroni:
I think it would make sense to keep those lines in af-magic theme commented out or just not in there to not mess up other people using it. It would just be nice to add the option to view this in any theme

PR #12270, opened 2024-03-08
fix(poetry-env): `poetry info --path` `cannot find directory '/bin/activate'` because...

claytonjroberts:
... 'poetry info --path' does not return anything.

## Standards checklist:

## Changes:

- Poetry shell now invoked through it's own `poetry shell` instead of using `source ...`

## Other comments:

- This has not been tested extensively with other poetry versions.


carlosala:
Hi! It seems like `poetry env info --path` still exist in recent versions of poetry, and spawning a new shell is not desirable for simplicity and ensuring we're in the right shell.
Maybe it's a different bug, but this is changing too much the way it works.
Thanks for your contribution though!

PR #12278, opened 2024-03-11
fix(tmux): Add completion for alias functions

detroyejr:
fix(tmux): Add completion for alias functions

Fixes https://github.com/ohmyzsh/ohmyzsh/issues/12282

## Standards checklist:

## Changes:

- Fixes completion for aliased functions (`ta`, `tad`, and `tkss`) introduced in https://github.com/ohmyzsh/ohmyzsh/pull/12232.
## Other comments:

I barely understand the ZSH completion system, but this seems to get the job done. `ta<tab>` and `tmux attach -t<tab>` seem to provide identical results. Any feedback is welcome.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -59,11 +59,11 @@ function _build_tmux_alias {
 
 # ALIAS COMPLETION (Modified from https://github.com/zsh-users/zsh)
 _tmux-attach-session() {
-  local sessions
-  sessions=(${(f)"$(tmux list-session)"})
   [[ -n ${tmux_describe} ]] && print "attach or switch to a session" && return
 
-  # If no flag is given, complete with session names.
+  local sessions
+  sessions=(${(f)"$(tmux list-session)"})
+
   if (( CURRENT == 2 )); then
     _alternative \
       'commands:: _describe -t sessions "tmux session" sessions'
@@ -72,19 +72,17 @@ _tmux-attach-session() {
 
   if [[ $words[1] =~ ta ]]; then 
     _arguments -s \
-      '1:sessions' \
+      '1:session' \
       '-c+[specify working directory for the session]:directory:_directories' \
       '-d[detach other clients attached to target session]' \
       '-f+[set client flags]: :_tmux_client_flags' \
       '-r[put the client into read-only mode]' \
-      '-t+[specify target session]:target session: __tmux-sessions-separately' \
       "-E[don't apply update-environment option]" \
       '-x[with -d, send SIGHUP to the parent of the attached client]'
     elif [[ $words[1] = "tkss" ]]; then
       _arguments -s \
-        '1:sessions' \
+        '1:session' \
         '-a[kill all session except the one specified by -t]' \
-        '-t+[specify target session]:session:__tmux-sessions' \
         '-C[clear alerts (bell, activity, silence) in all windows linked to the session]'
     fi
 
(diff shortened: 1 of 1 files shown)


patrick96:
For what it's worth, if you do things like `ta 0 -f`, tab completion doesn't work for that flag. But it's still better than the current state, where completion isn't working at all

detroyejr:
Ah, yes I see what you mean. Alternatively, [the ZSH user repository](https://githubcom/zsh-users/zsh/blob/ea5a5d6ec430cad1a73f3179fcc018c90a315c35/Completion/Unix/Command/_tmux#L167) provides a lot of good functionality for completion. I see other plugins have borrowed from them before, so maybe we can just import the relevant parts rather than re-inventing the wheel. I'll see if that's doable.

detroyejr:
@patrick96 thanks for the feedback. I've addressed your comment and now the following should happen:

`ta<tab>` without arguments completes on session names as before.
`ta <session name><tab>` will now start completing any of the additional arguments that you can pass to tmux attach.

`tmux kill-session` takes different arguments so I've added those as well, but it should act like the above.

detroyejr:
Cleaned this up a bit so that completion for `ta <session> -<tab>` looks the same as `tmux attach -t <session> -<tab>`. Same goes for `tkss`.

Barring any further feedback/cleanup, I think this good to go.

matthew-nm:
Is my comment here valid?: https://github.com/ohmyzsh/ohmyzsh/issues/12282#issuecomment-2021346588

Would changing the approach to setting an alias (as in the linked comment) vs defining a function be better? I don't know anything about this codebase but a quick test on my end seems to show that doing so would avoid having to reimplement any autocompletion functionality now or on future changes.

detroyejr:
Does the snippet you wrote provide more than the alias  `alias ta="tmux attach -t"`? I think the only thing it might be missing is being able to run `ta` without any arguments. That was the main reason that prompted me to open a PR in the first place.

If the maintainers want to support better completion over my workflow preference to attach to the last session with`ta`, that's fine by me. It makes a lot of sense to say this will just no longer work if it makes maintaining this too complex.

matthew-nm:
Yes, the snippet replaces the current '_build_tmux_alias' function and can provide the same 4 aliases (ta, tad, tkss, ts) and ta, tad, and ts seem to work without a specified session as well as with. But I don't know if `ta` is necessarily attaching to the last session or what it's doing exactly (seems kinda random to me). 

I don't follow what use case the 'if' statement is designed for so feel free to add the functionality you're looking for and I can always create my own aliases if I feel the need.

I appreciate you figuring out a way to add the autocomplete back in to the new methodology.

detroyejr:
For some reason, I still get the error I had previously. Just to be sure, does your `tmux -V` return 3.4?

Otherwise, I'm testing as follows:
```
# Reset
unfunction ta

function _build_tmux_alias {
  eval "if [[ -z \$1 ]] || [[ \${1:0:1} == '-' ]]; then
    alias \$1=\"tmux $2\"
  else
    alias \$1=\"tmux $2 $3\"
  fi"
}

_build_tmux_alias "ta" "attach" "-t"
```

Everything else works, but I still get `command attach-session: -t expects an argument` when trying to use `ta` without a session name.

matthew-nm:
Yes, my `tmux -V` returns `tmux 3.4`.

And if I enter either of the commands `tmux attach-session` or `tmux attach-session -t` (no target specified for `-t`) it takes me to one of my running sessions. I'm on macOS Sonoma 14.2.1.

However, my Ubuntu workstation on tmux 3.2a doesn't like it when I don't have an argument for `-t`, but using your version of `ta` gets around that. I understand what your function is getting at now and it does seem to work properly for lower versions of tmux.

matthew-nm:
If you want more work ;) you could perhaps check for the tmux version and set the aliases to `tmux attach -t`, `tmux attach -d -t`, etc. if the version is >=3.4, and use the function approach otherwise. That way future autocomplete changes to tmux will come through directly via the aliases (which support empty targets), whereas lower versions get the benefit of empty target commands as well as the current implementation of autocompletes that you've duplicated.

Thanks for making this functional for a wide range of tmux's.

detroyejr:
Thanks for confirming. That's weird because tmux didn't complain when I was on 3.2a (Linux/Nixos) and only started requiring an argument for `-t` after moving to 3.4. Maybe others can let us know which behavior they're seeing.

PR #12281, opened 2024-03-12
gradle plugin: introduce ZSH_GRADLE_PLUGIN_ALIAS_ENABLED

bartekpacia:
## Standards checklist:

## Changes:

The `gradle` provides two things:
- gradle zsh completion, which I want
- [`gradle-or-gradlew`](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/gradle#usage) function, which I don't want

This PR disables the `gradle-or-gradlew` function by default, and updates the plugin's README to explain this behavior.

### DIFF
--- a/plugins/gradle/README.md
+++ b/plugins/gradle/README.md
@@ -1,6 +1,6 @@
 # Gradle plugin
 
-This plugin adds completions and aliases for [Gradle](https://gradle.org/).
+This plugin adds completions and aliases for [Gradle](https://gradle.org).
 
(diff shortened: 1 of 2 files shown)


carlosala:
Hi! Thanks for your contribution! Actually, all this logic is not required. If you set the following style before sourcing `oh-my-zsh.sh` it'll not alias `gradle` to `gradlew`, but you'll still have access to `gradle-or-gradlew`, which I think is beneficial enough.

```zsh
zstyle ':omz:plugins:gradle' aliases no
```

PR #12285, opened 2024-03-15
feat: Add plugin for OpenTofu 

chenzejun:
## Standards checklist:

## Changes:
Add plugin for OpenTofu which is a fork of Terraform that is open-source, community-driven

Migrate completion ability form Terraform plugin with some update for Opentofu, such as descrptions
Use different alias from Terraform plugin to avoid conflict(`alias tt='tofu'`  vs. `alias tf='terraform'`)
Add three useful alias for refresh(`alias ttr='tofu refresh'`), test(`alias ttt='tofu test'`) and workspace(`alias ttws='tofu workspace'`) command

## Other comments:

...


miketheman (approved):

Jitsusama:
I'm definitely interested to see this happen!

mcornella:
Hi @chenzejun, this looks very nice and almost ready to merge, and there is strong support for it.

I have one question though in regard to the completion, where is this from? I haven't seen it anywhere else, and I can't see opentofu bundling it either. I'm hesitant to merge it this way over just doing

```sh
complete -C tofu tofu
```

which is the bash method and works for zsh as well\*, since the zsh way has to be kept up to date with the changes to `tofu`.

Is this just a rework of the completion in the terraform plugin? I've tested the `complete` method and it seems to do the same, minus the descriptions, so for maintainability I would rather go with the `complete` version.

\* After doing `bashcompinit`, which we do by default in Oh My Zsh.

mcornella:
Alright, made several improvements based on https://github.com/ohmyzsh/ohmyzsh/pull/12285#issuecomment-2226978526.

carlosala (approved):
LGTM!🚀

PR #12294, opened 2024-03-21
fix(mise): update environment when loading the plugin

Sacro:
Now mise installs can be found by any following OMZ plugins.

## Standards checklist:

## Changes:

- Run `mise hook-env` automatically

## Other comments:

...


mcornella (approved):
LGTM

This will update the `$PATH` if needed when the plugin runs, not only before the prompt is drawn, so any tools installed via mise will be available for that period.

Seems that [`mise activate`](https://mise.jdx.dev/faq.html#what-does-mise-activate-do) doesn't already do it.

PR #12302, opened 2024-03-25
Introduce `gwipe` alias

MatheusRich:
## Problem

You're working on a code problem for a while. You tried a bunch of things, nothing works. You're tired, confused by all the files you modified and created. You need a clean slate. "I'm sure ohmyzsh has a handy helper method to help me get that", you think.

You remember `gpristine` and runs it. It gets you that clean slate. But also deletes all of your .env files and encryption keys. You dev environment won't work now. You need to ask IT for those files again 😓.

## Solution

We need `gpristine` to be a little less aggressive. I propose `gwipe`, a way to wipe out "trash" while keeping the important things.

## Standards checklist:

## Changes:

Introduces `gwipe` as a way to get a "clean slate" but still have your dev environment working.
It runs `git reset --hard` and `git clean`, but doesn't delete untracked files


plugins/git/git.plugin.zsh
```diff
@@ -349,6 +349,7 @@ alias grhh='git reset --hard'
 alias grhk='git reset --keep'
 alias grhs='git reset --soft'
 alias gpristine='git reset --hard && git clean --force -dfx'
+alias gwipe='git reset --hard && git clean --force -df'
```
MatheusRich, line 352: I wasn't sure if the alias were supposed to be sorted alphabetically or not, so I added it next to gpristine, since they're similar.

robbyrussell:
Hey @MatheusRich, it was so nice meeting you in 🇧🇷 at #TropicalRB. Appreciate the addition! ✌️

Cheers,
Robby

PR #12304, opened 2024-03-25
fix(async): avoid blocking the shell while waiting

### LINKED ISSUE #12290: Freezes when using more than one async prompt handler
### Describe the bug

When trying to implement async prompt handling for `git_prompt_status`, I found that when I register two functions with `_omz_register_handler`, pressing enter quickly enough (using key repeats for example) :
 - the terminal freezes
 - after about 20s, it handles a few key presses, but freezes again
 - after almost 20s, it it responsive again
 
Sometimes, even other shells seem to be blocked (but for a shorter duration), so something seems to be shared.

### Steps to reproduce

In `~/.zshrc`, after `source $ZSH/oh-my-zsh.sh`, add :
```
function foo() {
}
function bar() {
}
function _defer_async_register() {
    _omz_register_handler foo
    _omz_register_handler bar
}
precmd_functions=(_defer_async_register $precmd_functions)
```

Then start a shell, press enter and leave the key pressed : after a few seconds the shell will freeze.

### Expected behavior

Async prompt handler should support several handlers without any issue.

_No response_

### OS / Linux distribution

Fedora 39

### Zsh version

5.9

### Oh My Zsh version

master (0fed3668)

### Terminal emulator

Konsole (same on a VT)

None

### Additional context

_No response_

hwti:
## Standards checklist:

## Changes:

- Make sure `_omz_async_callback` is only called when the handler writes data (at the end), instead of just after the fork : it blocked the shell until the end of the async prompt handler
- Do not use a subshell to set the status
- Fixes #12290 (cc @mcornella)

## Other comments:
This is draft for now, as I cannot explain why the subshell could sometimes block for several seconds (which blocked the shell in `_omz_async_callback`).
Even when it didn't block, the subshell was 10-20ms slower anyway.

### DIFF
--- a/lib/async_prompt.zsh
+++ b/lib/async_prompt.zsh
@@ -44,7 +44,7 @@ function _omz_register_handler {
 # Set up async handlers and callbacks
 function _omz_async_request {
   local -i ret=$?
-  typeset -gA _OMZ_ASYNC_FDS _OMZ_ASYNC_PIDS _OMZ_ASYNC_OUTPUT _OMZ_ASYNC_HANDLERS
+  typeset -gA _OMZ_ASYNC_FDS _OMZ_ASYNC_PIDS _OMZ_ASYNC_OUTPUT
 
   # executor runs a subshell for all async requests based on key
   local handler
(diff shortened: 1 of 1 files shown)


mcornella (approved):
Everything looks fast and snappy, and works well with 5.0.8. LGTM!

mcornella:
Thanks Loïc!

PR #12309, opened 2024-03-30
fix(starship)!: keep `ZSH_THEME` when Starship is not found

masaru-iritani:
## Standards checklist:

## Changes:

BREAKING CHANGE: starship plugin will keep ZSH_THEME as a fallback if Starship is not found. You can put ZSH_THEME="" before running oh-my-zsh.sh to keep the previous behavior where starship plugin always unset the Zsh theme.

## Background:

While I usually use Starship to configure the prompt style, some of my environments don't have Starship installed (by default). I want to use one of Oh My Zsh themes as the fallback instead of the default Zsh prompt if Starship is not (or cannot be) installed. 

## Tests:

- [x] `ZSH_THEME` is used when starship plugin is enabled and Starship is not installed.
- [x] The Starship prompt is used when starship plugin is enabled and Starship is installed.

## Other comments:

CC @axieax as the author of the great plugin

### DIFF
--- a/plugins/starship/starship.plugin.zsh
+++ b/plugins/starship/starship.plugin.zsh
@@ -1,7 +1,7 @@
-# ignore oh-my-zsh theme
-unset ZSH_THEME
-
 if (( $+commands[starship] )); then
+  # ignore oh-my-zsh theme
+  unset ZSH_THEME
+
   eval "$(starship init zsh)"
 else
   echo '[oh-my-zsh] starship not found, please install it from https://starship.rs'


axieax:
Good catch! 😎

carlosala (approved):

carlosala:
Thanks!

axieax:
@masaru-iritani would you like to create a PR in the main plugin [repo](https://github.com/axieax/zsh-starship) as well?

PR #12318, opened 2024-04-03
perf(async): avoid executing true

hwti:
## Standards checklist:

## Changes:

- Do not run `command true` (fork + exec) in `_omz_async_request` with zsh >= 5.8

### DIFF
--- a/lib/async_prompt.zsh
+++ b/lib/async_prompt.zsh
@@ -3,6 +3,7 @@
 # https://github.com/woefe/git-prompt.zsh/blob/master/git-prompt.zsh
 
 zmodload zsh/system
+autoload -Uz is-at-least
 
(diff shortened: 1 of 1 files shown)


lib/async_prompt.zsh
```diff
@@ -93,7 +93,9 @@ function _omz_async_request {
 
     # There's a weird bug here where ^C stops working unless we force a fork
     # See https://github.com/zsh-users/zsh-autosuggestions/issues/364
-    command true
+    # and https://github.com/zsh-users/zsh-autosuggestions/pull/612
+    autoload -Uz is-at-least
```
mcornella, line 97: Can you extract this out of the function?

mcornella (changes requested):

PR #12319, opened 2024-04-03
feat(git): implement async completion git_prompt_status

hwti:
## Standards checklist:

## Changes:

- `git_prompt_status` is now asynchronous like `git_prompt_info`, it's important since there are themes using both

## Other comments:
 - the handler for git_prompt_info was incorrectly named _omz_git_prompt_status
 - _defer_async_git_register was kept in precmd, there is no need to call it on each prompt


hwti:
cc @mcornella

mcornella (approved):
LGTM

PR #12323, opened 2024-04-04
feat(tasko): new tasko plugin

knid:
## Standards checklist:

## Changes:

- Created tasko function
- Created README.md file

## Other comments:

...

### DIFF
--- a/plugins/tasko/README.md
+++ b/plugins/tasko/README.md
@@ -1,0 +1,11 @@
+# TaskWarrior Plugin
+
+Open annotation content with native open commands (xdg-open or open)
+
+To use it add `tasko` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... tasko)
+```
+
+**Note:** you have to [install taskwarrior](https://github.com/GothenburgBitFactory/taskwarrior) first.
--- a/plugins/tasko/tasko.plugin.zsh
+++ b/plugins/tasko/tasko.plugin.zsh
@@ -7,7 +7,7 @@ function _tasko_print_help() {
 function tasko () {
   if [[ $# -ne 2 ]]; then
     echo 'Error: Too many/few arguments, expecting two.' >&2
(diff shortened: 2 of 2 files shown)


robbyrussell:
Hi @knid, what are you offering here that isn't currently available in the [existing TaskWarrior plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/taskwarrior)? 

Might this be something we can blend into the existing plugin?

knid:
Hi @robbyrussell, it's different. The plugin you shared enables tab completion. Mine is opens annotation content with native open commands (xdg-open or open). Like URL or file.

robbyrussell:
@knid I'm worried that 'tasko' might not be descriptive enough of a name and if people are using TaskWarrior, they'll see the other. Could we explore merging these two concepts into one?

knid:
@robbyrussell Taskwarrior uses `task` cli command. I thought `task-open` or `tasko` would be nice. But that's okay for me. We can merge two repo into one.

knid:
Or i can change repo name to `taskwarrior-open`

mcornella:
Can you show a recording of what your plugin does, for example using `asciinema`?

It's not clear from reading TaskWarrior docs why `xdg-open` or `open` is needed, as according to [[1]] descriptions are just text-based.

BTW we have a utility function called `open_command` that uses the right command based on platform/OS.

[1]: https://taskwarrior.org/docs/dom/

knid:
[asciinema record](https://asciinema.org/a/vKZYyLYMVQ6qK5u7CmSXDlPqQ)

Its just open annotation content quickly.

mcornella:
I think this makes the most sense as an addition to Taskwarrior itself, if it's a common use case to add URLs as task annotations. If not, that is too specific of a use case and won't be useful to enough people to be worth the expensive maintenance cost and manual testing.

Thanks for the contribution!

PR #12337, opened 2024-04-08
feat(macos): `ofd` now opens any directory

bartekpacia:
## Standards checklist:

I also tested the code locally (see all scenarios above) and it works as expected (`zsh 5.9 (x86_64-apple-darwin23.0)`)

## Changes:

This PR fixes the `ofd` function to not open the `$PWD` in Finder if there are other arguments passed.

I often use `ofd`, but almost equally often I use `ofd <path>`, but in that case, two Finder windows are opened: one with `$PWD`, and the second one with the path I passed. I think this behavior is a small bug, so I changed it.

### Without these changes

- `ofd` opens `$PWD` in Finder
- `ofd .` opens `$PWD` in Finder
- `ofd path1` opens `$PWD` and `path1` in Finder
- `ofd path1 path2` opens `$PWD`, `path1` and `path2` in Finder

### With these changes

- `ofd` opens `$PWD` in Finder
- `ofd .` opens `$PWD` in Finder
- `ofd path1` opens `path1` in Finder
- `ofd path1 path2` opens `path1` and `path2` in Finder

### DIFF
--- a/plugins/macos/README.md
+++ b/plugins/macos/README.md
@@ -17,7 +17,7 @@ Original author: [Sorin Ionescu](https://github.com/sorin-ionescu)
 | `tab`         | Open the current directory in a new tab                  |
 | `split_tab`   | Split the current terminal tab horizontally              |
 | `vsplit_tab`  | Split the current terminal tab vertically                |
-| `ofd`         | Open the current directory in a Finder window            |
+| `ofd`         | Open passed directories (or $PWD by default) in Finder   |
 | `pfd`         | Return the path of the frontmost Finder window           |
 | `pfs`         | Return the current Finder selection                      |
 | `cdf`         | `cd` to the current Finder directory                     |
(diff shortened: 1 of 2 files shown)


bartekpacia:
(This is a draft)

carlosala (changes requested):
Hi! I like the idea! I think it's better just to modify `macos` alias. I'd rather define something like
```zsh
function ofd {
  if (( ! $# )); then
    open_command $PWD
  else
    open_command $@
  fi
}
```
You can modify as well the comment in the code and the README to reflect the new behaviour of the function.

bartekpacia:
I like it! Thanks!

carlosala (commented):
Could you update the comment on top of the function and the plugin README to match the behaviour of the function as well?
Thanks!

bartekpacia:
Sure, done!

carlosala (approved):

carlosala:
Thanks!

PR #12338, opened 2024-04-08
fix(history)!: fix arg passing to `fc` for `history` command

### LINKED ISSUE #12334: omz_history -p doesn't work as expected
### Describe the bug

- In bash, in order to not save the history for the current session you use `history -c`. To do the same in zsh you use `history -p` instead.
- Using ohmyzsh, `history -c` clears the history file completely instead of just not saving the history for the current session. 
- The history file path is stored in the `HISTFILE` environment variable. By default, it's the `$HOME/.zsh_history` file.
- When using oh-my-zsh, `history` is aliased to the `omz_history` function. which is defined as follows:
https://github.com/ohmyzsh/ohmyzsh/blob/bf713e2c112ee1f0daf10deffa1215c982513f9b/lib/history.zsh#L2-L18

- When attempting to clear the history completely, you run `history -c` or its equivalent `omz_history -c`. This clears the history file completely as intended.
- When attempting to **not** save history for the _current_ session but not tampering with the history file, you would run `history -p`.
- However, when attempting to run `history -p` with ohmyzsh this results in running `builtin fc -l "$@" 1`. Here `"$@"` is set to `-p` which results in `builtin fc -l -p 1`. This indeed does not save the current session history to the original history file. However:
- What this causes is that the `HISTFILE` environment variable is overwritten with the value `1`. A file named `1` is automatically created in the current working directory and the session history is stored in it. Any attempt to remove the file within the session will result in it being recreated whenever any interactive command is executed until the session is exited.
- When the session is exited, the file still exists.

References:
- https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html (zsh builtins)

### Steps to reproduce

1. Start a new session 
2. Navigate to a directory with no files called `1`.
3. `echo $HISTFILE` should result in `$HOME/.zsh_history` by default.
4. run `history -p`
5. `echo $HISTFILE` results in `1`.
6. `ls` you find a file called `1`.

### Expected behavior

1. `history -p` unsets the `HISTFILE` environment variable.
2. No file is created.
3. Current session history is not preserved beyond the session.

[asciicast](https://asciinema.org/a/6HQvRYH9WHoWIN0H5tYJa1CIC)

### OS / Linux distribution

macOS  Sonoma 14.4.1

### Zsh version

5.9

### Oh My Zsh version

master (bf713e2)

### Terminal emulator

iTerm2

None

### Additional context

_No response_

MoAlkhateeb:
fixes #12334 by not forcing the `1` argument and giving the user more control.

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -12,8 +12,8 @@ function omz_history {
     # if -l provided, run as if calling `fc' directly
     builtin fc "$@"
   else
-    # unless a number is provided, show all history events (starting from 1)
-    [[ ${@[-1]-} = *[0-9]* ]] && builtin fc -l "$@" || builtin fc -l "$@" 1
+    # otherwise, run `fc -l` with a custom format
+    builtin fc -l "$@"
   fi
 }
 


mhagnumdw:
The side effect of this change is that the user expects to see the full history when typing `history`, but will now only see the last 16 records. You even think that the history has been deleted!!

filipeacacio:
I believe this is not a good solution, as the default behavior of a widely used command has changed.

MoAlkhateeb:
We could fix that by explicitly handling the case when the user doesn't use any arguments to show all the history starting from 1. This would keep the behaviour the same while still fixing the issue. What do you think? @filipeacacio @mhagnumdw @mcornella @carlosala

filipeacacio:
> We could fix that by explicitly handling the case when the user doesn't use any arguments to show all the history starting from 1. This would keep the behaviour the same while still fixing the issue. What do you think? @filipeacacio @mhagnumdw @mcornella @carlosala

Sounds good for me.

mhagnumdw:
> We could fix that by explicitly handling the case when the user doesn't use any arguments to show all the history starting from 1. This would keep the behaviour the same while still fixing the issue. What do you think? @filipeacacio @mhagnumdw @mcornella @carlosala

Great

PR #12340, opened 2024-04-09
fix(shell-proxy): support ssh-proxy using http proxy in cygwin

tangruize:
The ssh-proxy.py generates `nc` commands to connect the proxy. However, in cygwin, it does not set the protocol type, leading to `nc` using wrong default protocol (socks5) if we actually use an http proxy.

## Standards checklist:

## Changes:

- add a check ("cygwin") for sys.platform

### DIFF
--- a/plugins/shell-proxy/ssh-proxy.py
+++ b/plugins/shell-proxy/ssh-proxy.py
@@ -22,7 +22,8 @@ if parsed.scheme not in proxy_protocols:
 
 def make_argv():
     yield "nc"
-    if sys.platform == 'linux':
+    if sys.platform in {'linux', 'cygwin'}:
+        # caveats: the built-in netcat of most linux distributions and cygwin support proxy type
         # caveats: macOS built-in netcat command not supported proxy-type
         yield "-X" # --proxy-type
         # Supported protocols are 4 (SOCKS v4), 5 (SOCKS v5) and connect (HTTP proxy).


septs (approved):
very good, plz provide tested gif (video)?

tangruize:
Sorry, I don't know how to record a gif. I screenshotted the process (I added a print for the nc commands below):

The proxy setting:

image

After apply the patch this pr ( `if sys.platform in {'linux', 'cygwin'}:`):

image

Before apply the patch (`if sys.platform == 'linux':`):
image

plugins/shell-proxy/ssh-proxy.py
```diff
@@ -22,7 +22,7 @@
 
 def make_argv():
     yield "nc"
-    if sys.platform == 'linux':
+    if sys.platform in {'linux', 'cygwin'}:
         # caveats: macOS built-in netcat command not supported proxy-type
```
septs, line 26: plz update the caveats

carlosala (commented):
As soon as we get an approval by @septs I'm happy to merge!

plugins/shell-proxy/ssh-proxy.py
tangruize: I will update the comments tomorrow (because I found that there exists several netcat versions on Linux, such as `netcat-openbsd` (default, works), `netcat-traditional` and `ncat` (nmap netcat)).

tangruize: I've examined the preinstalled versions of netcat on some Linux distributions, MacOS, and Cygwin. It seems that all of them now support the '-X' option. For a more general check, I think the following modification is better. I have tested it.

```diff
diff --git a/plugins/shell-proxy/ssh-proxy.py b/plugins/shell-proxy/ssh-proxy.py
index 5d239177..e68d109e 100755
--- a/plugins/shell-proxy/ssh-proxy.py
+++ b/plugins/shell-proxy/ssh-proxy.py
@@ -22,7 +22,7 @@ if parsed.scheme not in proxy_protocols:
 
 def make_argv():
     yield "nc"
-    if sys.platform in {'linux', 'cygwin'}:
+    if b'-X ' in subprocess.check_output(['nc', '-h'], stderr=subprocess.STDOUT):
         # caveats: macOS built-in netcat command not supported proxy-type
         yield "-X" # --proxy-type
         # Supported protocols are 4 (SOCKS v4), 5 (SOCKS v5) and connect (HTTP proxy).
```

tangruize: I've added a caveat.

tangruize: I also investigated the netcat versions on linux, cygwin and macos, and found that they all support the '-X proto' option. I tried to make it more general by using `if b'-X ' in subprocess.run(['nc', '-h'], check=False, stderr=subprocess.STDOUT, stdout=subprocess.PIPE).stdout:`. However, I failed to clone my repo on MacOS.

septs (approved):

carlosala (approved):

carlosala:
Thanks!

PR #12341, opened 2024-04-09
Reverting to old behaviour when running just `history`

MoAlkhateeb:
Handles:
https://github.com/ohmyzsh/ohmyzsh/pull/12338#issuecomment-2045281096
https://github.com/ohmyzsh/ohmyzsh/pull/12338#issuecomment-2045304370

Running `history` prints the whole history starting from 1.

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -12,8 +12,8 @@ function omz_history {
     # if -l provided, run as if calling `fc' directly
     builtin fc "$@"
   else
-    # unless a number is provided, show all history events (starting from 1)
-    [[ ${@[-1]-} = *[0-9]* ]] && builtin fc -l "$@" || builtin fc -l "$@" 1
+    # otherwise, run `fc -l` with a custom format
+    builtin fc -l "$@"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


brettmiller:
At least in my case it still only shows the last 16 records. It looks like possibly because of the alias that's set when $HIST_STAMPS is set.

carlosala:
> At least in my case it still only shows the last 16 records. It looks like possibly because of the alias that's set when $HIST_STAMPS is set.

Are you using `history` directly, or with any parameter?

brettmiller:
Directly with no parameters. 
```
$ history | wc -l
16
$ which history
history: aliased to omz_history -i
```

brettmiller:
I'm not setting the `history` alias to `omz_history -i` that's coming from lib/history.zsh. So, to me it makes since to explicitly  handle the aliases it sets as well. Testing locally adding 
```
  elif [[ $# -eq 1 ]]; then
    if [[ $1 -eq "-f" || $1 -eq "-E" || $1 -eq "-i" ]]; then
      builtin fc -l $1 1
    fi
```
between the `if [[ $# -eq 0 ]]; then` and `elif [[ -n "$clear" ]]; then` that was added in this PR seems to handle those cases correctly. I'd be happy to open a PR with that change.

mcornella:
This is now fixed in #12357 thanks to @artnim. Cheers!

PR #12345, opened 2024-04-10
feat(git) Add alias for signed git commit amend

nic-lan:
## Changes:

- Add `gcsa` alias to allow sign amend of existing commits.

#### Notes

#7460 has a similar goal, but it looks abandoned to me. Also, it shows conflicts with the base.


carlosala:
Hi! In general, one wants to set up gpgsign in `.gitconfig` for all commits.
I'd say this alias does not add a lot of value for a big chunk of users. We're trying to limit the amount of aliases that reach git plugin since we already have an outrageous amount. I encourage you to set it up in your zshrc if it's critical for you! You can use as well `gc! -S` that would have the same effect.
Thanks for your contribution!

PR #12347, opened 2024-04-11
fix(yarn): use `builtin cd` in completion

hwknsj:
This simply replaces 3 instances of `cd` in the `_yarn` completion script with `builtin cd` to avoid issues in the case that `cd` has been aliased or redefined. While this may sound like an unusual situation, consider the example below.

Example: `cd` is 'enhanced' to list contents of the new directory, e.g.
```shell
function cd () {
    builtin cd "$1"
    ls -laFhGT
}
```
Chaos ensues on invoking completion, `yarn ⇥` <kbd>tab</kbd>

<a href="https://asciinema.org/a/ejagW9I1NnnSpciUanxshhx0r" target="_blank">asciinema of yarn completion bug</a>

Here is a video screen recording if asciinema link goes away:
<video controls muted src="https://github.com/ohmyzsh/ohmyzsh/assets/8978069/82cd8f18-ba01-4ab8-8d5b-147d8500a547">
  <source src="https://github.com/ohmyzsh/ohmyzsh/assets/8978069/82cd8f18-ba01-4ab8-8d5b-147d8500a547" type="video/mp4" />
<a href="https://github.com/ohmyzsh/ohmyzsh/assets/8978069/82cd8f18-ba01-4ab8-8d5b-147d8500a547" target="_blank">video link</a>
</video>

## Standards checklist:

- ~~[ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below.~~

## Changes:

- replaced 3 instances of `cd` with `builtin cd` in `_yarn` completion definition script

## Other comments:

I can't really think of a better way to do this. I considered `/bin/cd`, but we can pretty safely assume that the user's shell is `zsh` and `builtin cd` always works.

### DIFF
--- a/plugins/yarn/_yarn
+++ b/plugins/yarn/_yarn
@@ -86,7 +86,7 @@ _global_commands=(
 )
 
 _yarn_find_package_json() {
-  local dir=$(cd "$1" && pwd)
+  local dir=$(builtin cd "$1" && pwd)
 
   while true
   do
@@ -109,7 +109,7 @@ _yarn_commands_scripts() {
 
   if [[ -n $opt_args[--cwd] ]]; then
     packageJson=$(_yarn_find_package_json $opt_args[--cwd])
-    binaries=($(cd $opt_args[--cwd] && echo node_modules/.bin/*(x:t)))
+    binaries=($(builtin cd $opt_args[--cwd] && echo node_modules/.bin/*(x:t)))
   else
     packageJson=$(_yarn_find_package_json $pwd)
     binaries=($(echo node_modules/.bin/*(x:t)))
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution! Could you open a PR as well in https://github.com/zsh-users/zsh-completions/blob/master/src/_yarn?
That's the upstream!

hwknsj:
> Thanks for your contribution! Could you open a PR as well in https://github.com/zsh-users/zsh-completions/blob/master/src/_yarn? That's the upstream!

Oh yeah, sure thing!

@carlosala thanks for merging my PR!

PR #12349, opened 2024-04-12
Add auto install feature for pre-commit plugin

marcodisarno:
## Standards checklist:
This was inspired by ohmyzsh's dotenv plugin, which I've grown to like a lot. I re-used the code from it's `source_env` function, so thanks to whoever wrote that!

## Changes:

- adds a zsh chpwd hook to automatically prompt for installing pre-commit's git hooks when .pre-commit-config.yaml exists in the current working dir.

## Other comments:

it's quite configurable, but switched to "Prompt" by default.


robbyrussell:
Thanks for this, and sorry it went unanswered for so long. Going to close it... the plugin got rewritten around `prek` earlier this year so this no longer applies, and in the two years it sat here nobody else asked for an auto-install hook. If the idea still appeals to you, a fresh take on the current plugin would be the way to go.

PR #12351, opened 2024-04-13
feat(tmux): sessions named as current dir

jfpedroza:
## Standards checklist:

## Changes:

When `$ZSH_TMUX_AUTONAME_SESSION=true`, name the tmux session after the working directory.

If you are in `~/foo/bar`, the name will be `bar`. If you are in `~/`, the name will be `home`.


carlosala (changes requested):
I like it! I'm ok with merging it! I'd name `HOME` in caps; it's less common to have a directory called `HOME` than `home`.
The only thing that's missing is a special case for `/`. Maybe `ROOT`?

jfpedroza:
Never considered opening a session at `/`. I agree with `ROOT`.

carlosala (approved):

carlosala:
Thanks for your contribution!

Issue #12328, opened 2024-04-06
git async prompt does not appear


carlosala:
Same problems as OP:

Before:

![image](https://github.com/ohmyzsh/ohmyzsh/assets/13176356/0092068d-c639-45bd-bbea-3fbffcbabef9)

After:

![image](https://github.com/ohmyzsh/ohmyzsh/assets/13176356/8a8f5ec2-5f54-467f-a7a7-c1cd9a06cd03)

theme: https://github.com/sobolevn/sobole-zsh-theme
zsh version: 5.9 (x86_64-debian-linux-gnu)
git version: 2.39.2
Debian GNU/Linux 12

_Originally posted by @hwalinga in https://github.com/ohmyzsh/ohmyzsh/issues/12267#issuecomment-2040655456_

carlosala:
I see, your theme uses a [custom function](https://github.com/sobolevn/sobole-zsh-theme/blob/40ef40c1fb5eb5c1e7bbab913a7dbfbf1f449aa9/sobole.zsh-theme#L49-L51) print git information. We didn't take into account this situation. It's reproducible as well with `minimal` theme bundled in `oh-my-zsh`.

carlosala:
@mcornella we need to find a new way of discovering that it's indeed being used. Maybe with a fd or something like that?

carlosala:
#12327 reports exactly the same regression

134130:
FYI: The third-party plugin 'bullet-train' is also in the same problem.

LockonS:
> FYI: The third-party plugin 'bullet-train' is also in the same problem.

Encounter the same issue for bullet-train theme. There is a temporary work around for this.

```shell
# replace this line
BULLETTRAIN_GIT_PROMPT_CMD="\$(git_prompt_info)"
# to 
BULLETTRAIN_GIT_PROMPT_CMD="\$(_omz_git_prompt_info)"
```

dintz:
> FYI: The third-party plugin 'bullet-train' is also in the same problem.

Yes, same problem here (also with bullet-train)!

dintz:
> > FYI: The third-party plugin 'bullet-train' is also in the same problem.
> 
> Encounter the same issue for bullet-train theme. There is a temporary work around for this.
> 
> ```shell
> # replace this line
> BULLETTRAIN_GIT_PROMPT_CMD="\$(git_prompt_info)"
> # to 
[...]

PR #12356, opened 2024-04-16
[F] fix pygmolin theme - not have git promt

artinu777:
[[F] fix pygmolin theme - not have git promt]


carlosala:
This is due to a side effect of #12328.
We're working on a fix!

PR #12357, opened 2024-04-16
fix(history): fix edge cases introduced in #12341

artnim:
Introduce local stamp and catch options -f, -E and -i to it.

## Standards checklist:

## Changes:

- Add a local stamp variable and use `zparesopts` to catch option `-f`, `-E` or `-i'

## Other comments:

...

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -1,19 +1,19 @@
 ## History wrapper
 function omz_history {
-  local clear list
-  zparseopts -E c=clear l=list
+  local clear list stamp
+  zparseopts -E -D c=clear l=list f=stamp E=stamp i=stamp
 
   if [[ $# -eq 0 ]]; then
     # if no arguments provided, show full history starting from 1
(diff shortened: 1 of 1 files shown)


mcornella:
Thanks for the PR! Tested everything is now back to normal. Cheers!

thomaslienbacher:
There still seems to be a problem with the old default behaviour when `$HIST_STAMPS` is set. In this case `history` does not by default display the full history. Thus the default behaviour is still broken, this was also observed by @brettmiller in https://github.com/ohmyzsh/ohmyzsh/issues/12341#issuecomment-2050918069.

artnim:
@thomaslienbacher I confirm that there still just 16 lines if $HIST_STAMPS set to a value other than `"mm/dd/yyy|dd.mm.yyyy|yyyy-mm-dd"` -- which is fully valid as long as it meets the `strftime` format specification.

Issue #12331, opened 2024-04-07
VS CODE terminal cannot start, after udpate oh-my-zsh


yingzi17:
### Describe the bug

when i update oh-my-zsh, i cannot start a terminal window in vscode
> 17:14:11 段错误(吐核)

### Steps to reproduce

1.update oh-my-zsh,
2. i cannot start a terminal window in vscode

### Expected behavior

how can i fix it

### OS / Linux distribution

linux

### Zsh version

zsh 5.0.2 (x86_64-redhat-linux-gnu)

### Terminal emulator

vscode

### If using WSL on Windows, which version of WSL

None

taekinkim:
You can avoid the problem temporarily.

```sh
cd ~/.oh-my-zsh
git checkout 6dfa9507ce0eb0f4d386bd03268e33943ea55c0f
```
and then, try again.

carlosala:
Hi! I can't reproduce any issue. What's exactly happening? Could you share a screen recording or further describe the issue? @taekinkim does it happen to you as well?

taekinkim:
@carlosala 
Yes, it happens to me as well.

I have worked in vscode with remote-ssh connection.
I could open my repository in vscode, but I couldn't open the terminal window in vscode.
When I try to open a terminal window, the terminal window appears and then it disappears immediately.
`core.98912` file is created, and I tried to run `gdb`.
```
$ gdb /bin/zsh core.98912
GNU gdb (GDB) Red Hat Enterprise Linux 10.2-5.el7
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /bin/zsh...
Reading symbols from .gnu_debugdata for /usr/bin/zsh...
(No debugging symbols found in .gnu_debugdata for /usr/bin/zsh)
[New LWP 98912]
[Thread debugging using libthread_db enabled]
[...]

PR #12358, opened 2024-04-16
fix(async): bypass zsh 5.0.2 crash on `zle reset-prompt`

mcornella:
## Standards checklist:

## Changes:

- Use `zle .reset-prompt` in `_omz_async_callback` to bypass crash reported in #12331 (test)

## Other comments:

~~This is only a test based on a hunch related to [[1]] and [[2]] fixed in 5.0.6.~~ Confirmed to solve the issue in [[3]].

[1]: https://www.zsh.org/mla/workers/2014/msg00237.html
[2]: https://github.com/zsh-users/zsh/commit/97115e0e7f33378390aabd3c15d0cd69bf8d191c
[3]: https://github.com/ohmyzsh/ohmyzsh/issues/12331#issuecomment-2065941645

### DIFF
--- a/lib/async_prompt.zsh
+++ b/lib/async_prompt.zsh
@@ -124,7 +124,7 @@ function _omz_async_callback() {
 
     # Repaint prompt if output has changed
     if [[ "$old_output" != "${_OMZ_ASYNC_OUTPUT[$handler]}" ]]; then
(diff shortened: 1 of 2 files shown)


carlosala (approved):
LGTM!🚀🚀

PR #12365, opened 2024-04-18
Add `t` option to omz_history function

### LINKED ISSUE #12364: history not by default display the full history when `$HIST_STAMPS` set to valid strftime format.
There still seems to be a problem with the old default behaviour when `$HIST_STAMPS` is set. In this case `history` does not by default display the full history. Thus the default behaviour is still broken, this was also observed by @brettmiller in https://github.com/ohmyzsh/ohmyzsh/issues/12341#issuecomment-2050918069.

_Originally posted by @thomaslienbacher in https://github.com/ohmyzsh/ohmyzsh/issues/12357#issuecomment-2063233501_
            
This happens, when $HIST_STAMP is set to any value not in `"mm/dd/yyyy|dd.mm.yyyy|yyyy-mm-dd"`

artnim:
The `t` option takes a strftime format specification as argument. This fixes cases in which $HIST_STAMPS is set to a value outrange `"mm/dd/yyyy|dd.mm.yyyy|yyyy-mm-dd"`.

## Standards checklist:

## Changes:

- Add `t` option to the zparseoptions to fix #12364 
## Other comments:

...

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -2,7 +2,7 @@
 function omz_history {
   # parse arguments and remove from $@
   local clear list stamp
-  zparseopts -E -D c=clear l=list f=stamp E=stamp i=stamp
+  zparseopts -E -D c=clear l=list f=stamp E=stamp i=stamp t:=stamp
 
   if [[ -n "$clear" ]]; then
     # if -c provided, clobber the history file


mcornella (approved):

thomaslienbacher:
I have `HIST_STAMPS="%c"` set and I just tested it and it fixed the problem. `history` now prints the complete history again.
Thanks for the quick response and fix! :)

PR #12366, opened 2024-04-19
Fix issue with homebrew-installed nc taking precedence in $PATH

vangie:
This commit addresses a problem where the nc command installed via homebrew was being used preferentially over the intended version due to its higher precedence in the $PATH environment variable. Adjustments have been made to ensure the script selects the correct nc executable, avoiding conflicts and ensuring consistent behavior across different setups.

```bash
$ which nc
/opt/homebrew/bin/nc
```

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/shell-proxy/ssh-proxy.py
+++ b/plugins/shell-proxy/ssh-proxy.py
@@ -21,8 +21,12 @@ if parsed.scheme not in proxy_protocols:
     raise TypeError('unsupported proxy protocol: "{}"'.format(parsed.scheme))
 
 def make_argv():
-    yield "nc"
-    if sys.platform in {'linux', 'cygwin'}:
+    if sys.platform == 'darwin':
+        # 'nc' in $PATH may be installed by homebrew, if without path
+        yield "/usr/bin/nc"
+    else:
+        yield "nc"
+    if sys.platform in {'linux', 'cygwin', 'darwin'}:
         # caveats: the built-in netcat of most linux distributions and cygwin support proxy type
         # caveats: macOS built-in netcat command not supported proxy-type
         yield "-X" # --proxy-type


plugins/shell-proxy/ssh-proxy.py
```diff
@@ -21,8 +21,12 @@
     raise TypeError('unsupported proxy protocol: "{}"'.format(parsed.scheme))
 
 def make_argv():
-    yield "nc"
-    if sys.platform in {'linux', 'cygwin'}:
+    if sys.platform == 'darwin':
+        # 'nc' in $PATH may be installed by homebrew, if without path
+        yield "/usr/bin/nc"
+    else:
+        yield "nc"
```
septs, line 28: hardcoded path is bad code.

i think it is needed by environ.

```suggestion
    yield os.environ.get("SHELLPROXY_NETCAT_PROGRAM", "nc")
```

plugins/shell-proxy/ssh-proxy.py
```diff
@@ -21,8 +21,12 @@
...
+    if sys.platform in {'linux', 'cygwin', 'darwin'}:
```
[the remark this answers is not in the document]
vangie: /usr/bin/nc is the BSD version pre-installed in macos, which need need use the same config as linux and cygwin. Therefore, it's all about addressing the issue of macOS not being ready to use out of the box.

plugins/shell-proxy/ssh-proxy.py
vangie: /usr/bin/nc is the BSD version pre-installed in MacOS，If users need to configure the system using SHELLPROXY_NETCAT_PROGRAM, then it might require debugging to identify the issue.

carlosala:
@septs what do you think about it?

septs:
> @septs what do you think about it?

The author wants to use it out of the box and hard-coded for macOS platform

carlosala:
Wouldn't it be a better solution to properly set `$PATH` to take the `nc` you prefer? @vangie

vangie:
@carlosala @septs Let me explain, I am a homebrew and oh-my-zsh user. Initially, my machine only had the BSD version of the nc command that comes with the Mac system.

```shell
which -a nc
/usr/bin/nc
```

shell-proxy is one of my favorite plugins, and I use it almost every day. One day, I installed a Homebrew version of nc.Because the Homebrew command in the PATH has a higher priority than the system default path, most of the time, it meets expectations to use the GNU version of commands.

```shell
which -a nc
/opt/homebrew/bin/nc
/usr/bin/nc
```

After some time, I discovered that shell-proxy was no longer working, but I did not realize that it was due to the previous Homebrew installation. with no choice, I became someone who manually exports the HTTP_PROXY setting. Until I could no longer tolerate it, I began to read the source code of shell-proxy, and after painstakingly debugging step by step, I discovered the aforementioned issue.

Setting the environment variable indeed solves my problem, but users similar to me might not infer from the error messages that the issue stems from having two versions of nc , specifically the differences between the BSD and GNU versions, causing the problem.

Of course, if you still believe that setting an environment variable is more appropriate, I can revise this PR (Pull Request) according to the previous suggestions. At the very least, the next time the plugin updates, I won't have to resolve conflicts locally.

septs:
I still can only accept

```python
yield os.environ.get("SHELLPROXY_NETCAT_PROGRAM", "nc")
```

carlosala:
I agree with @septs, we shouldn't pollute with particular OS cases.

carlosala:
Closing for the moment, we didn't get to a point of agreement. Feel free to reopen with the changes by septs.
Thanks for your contribution!

PR #12369, opened 2024-04-20
feat(fishy): add color to username

cidrmill:
Modifies the prompt in fishy.zsh-theme to display the hostname in a different color (yellow) when the user is connected via SSH. Additionally, the username is now displayed in a separate color block, improving readability and making it easier to distinguish different components of the prompt.

## Standards checklist:

## Changes:

- Modified the fishy theme prompt to add color to the username. This is consistent with the current version of fish-shell.
- Modified the fishy theme prompt to display the hostname in a different color (yellow) when the user is connected via SSH. This is consistent with the current version of fish-shell.

### DIFF
--- a/themes/fishy.zsh-theme
+++ b/themes/fishy.zsh-theme
@@ -16,7 +16,8 @@ _fishy_collapsed_wd() {
 }
 
 local user_color='green'; [ $UID -eq 0 ] && user_color='red'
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12370, opened 2024-04-20
fix(timer): skip timer after running `clear`

Nbelles:
## Standards checklist:

## Changes:

- fix(timer): Add check to see if last command that was run was `clear` and if so, don’t print the output of the timer. 

## Other comments:
The current behavior of the timer plugin is that it outputs the result of the timer even when the last command is the `clear` command. This poses a problem because once the screen has been cleared, it adds one more line to print the timer at the top of the screen before then printing out the terminal prompt. To fix this, I have added a simple check to see if the last command run is the `clear` command and if so, do not print the result of the timer.

### DIFF
--- a/plugins/timer/timer.plugin.zsh
+++ b/plugins/timer/timer.plugin.zsh
@@ -23,9 +23,12 @@ __timer_display_timer_precmd() {
     local tdiff=$((cmd_end_time - __timer_cmd_start_time))
     unset __timer_cmd_start_time
     if [[ -z "${TIMER_THRESHOLD}" || ${tdiff} -ge "${TIMER_THRESHOLD}" ]]; then
+      local last_cmd=$(fc -ln -1 | awk '{print $1}')
+      if [[ "$last_cmd" != clear ]]; then
         local tdiffstr=$(__timer_format_duration ${tdiff})
         local cols=$((COLUMNS - ${#tdiffstr} - 1))
         echo -e "\033[1A\033[${cols}C ${tdiffstr}"
+      fi 
     fi
   fi
 }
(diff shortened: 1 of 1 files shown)


Mkeays501:
zmodload zsh/datetime

__timer_current_time() {
  echo $EPOCHREALTIME
}

__timer_format_duration() {
  local total_secs=$1
  local mins=$(printf '%.0f' $(echo "$total_secs / 60" | bc -l))
  local secs=$(printf "%.${TIMER_PRECISION:-1}f" $(echo "$total_secs - 60 * $mins" | bc -l))
  local duration_str="${mins}m${secs}s"
  local format="${TIMER_FORMAT:-/%d}"
  echo "${format//\%d/${duration_str#0m}}"
}

__timer_save_time_preexec() {
  __timer_cmd_start_time=$(__timer_current_time)
  __timer_last_command=$1
}

__timer_display_timer_precmd() {
  if [ -n "${__timer_cmd_start_time}" ] && [ "${__timer_last_command}" != "clear" ]; then
    local cmd_end_time=$(__timer_current_time)
    local tdiff=$(echo "$cmd_end_time - $__timer_cmd_start_time" | bc -l)
    unset __timer_cmd_start_time
    if [[ -z "${TIMER_THRESHOLD}" || $(echo "$tdiff >= $TIMER_THRESHOLD" | bc -l) -eq 1 ]]; then
      local tdiffstr=$(__timer_format_duration $tdiff)
      local cols=$((COLUMNS - ${#tdiffstr} - 1))
      printf "\033[1A\033[${cols}C %s\n" "$tdiffstr"
    fi
  fi
}

autoload -U add-zsh-hook
add-zsh-hook preexec __timer_save_time_preexec
add-zsh-hook precmd __timer_display_timer_precmd

mcornella (approved):
I tested it and it works well. I only changed the method to get the last command via the special `$history` array. LGTM!

PR #12371, opened 2024-04-20
fix(ssh-agent): add identity only if identity exists

zikoengxi:
## Standards checklist:

## Changes:

Check if the identity exists.


mcornella:
Does this fix an existing issue? If so, can you provide a link or reproduction steps?

zikoengxi:
> Does this fix an existing issue? If so, can you provide a link or reproduction steps?

if the identity

zikoengxi:
> Does this fix an existing issue? If so, can you provide a link or reproduction steps?

If the identities are set up via zstyle, the non-existing identities are not checked. When loading the non-existing identity, `ssh-keygen -lf "$file"` will raise an exception: `ssh-keygen: /root/.ssh/$file: No such file or directory`.

mcornella:
Thanks, that makes sense. It's been a while since I wrote the plugin so the refresher was nice. Thanks for the contribution!

PR #12374, opened 2024-04-22
feat(yarn)!: add optional support for yarn berry

favna:
## Standards checklist:

## Changes:

- This updates the yarn plugin to better support Yarn berry (aka Yarn v2 and beyond) because some commands are no longer available or changed in modern Yarn.
- A new pair of aliases `yii` and `yifl` are added. These are to install dependencies while locking versions down to the `yarn.lock` file. The reason `yifl` is an alias for `yii` is to respect that for v1 it's **y**arn **i**nstall --**f**rozen-**l**ockfile, whereas for berry it's **y**arn **i**nstall --**i**mmutable, so `yii`.
- The aliases `ydlx` and `yn` are added to cover [`yarn dlx`](https://yarnpkg.com/cli/dlx) and [`yarn node`](https://yarnpkg.com/cli/node)

## Other comments:

Version checking inspired by the git plugin

### DIFF
--- a/plugins/yarn/yarn.plugin.zsh
+++ b/plugins/yarn/yarn.plugin.zsh
@@ -1,6 +1,5 @@
-# Get the configured version of yarn
+# Check if the user has configured Yarn Berry
 is_yarn_berry="false"
-
 if zstyle -t ':omz:plugins:yarn' berry; then
   is_yarn_berry="true"
 fi
--- a/plugins/yarn/yarn.plugin.zsh
+++ b/plugins/yarn/yarn.plugin.zsh
@@ -1,9 +1,3 @@
-# Check if the user has configured Yarn Berry
-is_yarn_berry="false"
-if zstyle -t ':omz:plugins:yarn' berry; then
-  is_yarn_berry="true"
-fi
-
 if zstyle -T ':omz:plugins:yarn' global-path; then
   # Skip yarn call if default global bin dir exists
   [[ -d "$HOME/.yarn/bin" ]] && bindir="$HOME/.yarn/bin" || bindir="$(yarn global bin 2>/dev/null)"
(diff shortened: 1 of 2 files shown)


carlosala (changes requested):
`yarn --version` is a really slow command. In my tests it was around 150ms for v1 and 300ms for v4.
We can't add 300ms in startup, it'd make it really slow.
I like the idea though, and I propose using a `zstyle` config for it, like `global-path`

favna:
> I like the idea though, and I propose using a zstyle config for it, like global-path

I'm not familiar with this. Can you provide a resource? Also, do you mean that through a zstyle the user would configure if they want the Yarn v1 or Yarn Berry set?

carlosala:
Yep, I mean exactly that. You can see an example in `yarn` plugin precisely, the config is called `global-path`. Here you have the zsh manual! https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html#The-zsh_002fzutil-Module

carlosala (approved):
LGTM!

plugins/yarn/yarn.plugin.zsh
```diff
@@ -18,29 +18,49 @@ alias yb="yarn build"
...
+  alias yga="yarn global add"
+  alias ygls="yarn global list"
+  alias ygrm="yarn global remove"
+  alias ygu="yarn global upgrade"
+  alias yls="yarn list"
+  alias yout="yarn outdated"
+  alias yuca="yarn global upgrade && yarn cache clean"
+fi
+
+unset is_yarn_berry
```
favna, line 66: @carlosala you forgot to remove this line in your "nit" commit

carlosala: Right, thanks!

PR #12383, opened 2024-04-29
improvement: Using `/usr/bin/env` as shebang.

alps2006:
## Standards checklist:

## Changes:

- Using `/usr/bin/env` as shebang, it is more flexible.

## Other comments:

...

### DIFF
--- a/plugins/coffee/coffee.plugin.zsh
+++ b/plugins/coffee/coffee.plugin.zsh
@@ -1,4 +1,4 @@
-#!/bin/zsh
+#!/usr/bin/env zsh
 
 # compile a string of coffeescript and print to output
 cf () {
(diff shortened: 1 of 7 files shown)


mcornella:
This doesn't solve anything, as none of these files are actually run directly but are instead sourced. Thanks for the contribution though.

alps2006:
The fixed `/bin/zsh` can lead to problems in some cases.

If I use `brew zsh` to support `pcre` and set `setopt rematchpcre`, then enter a multi-byte directory, the error will be shown below:

```
omz_urlencode:45: pcre_exec() error [-10]
# or
failed to load module `zsh/pcre': dlopen(/usr/lib/zsh/5.9/zsh/pcre.so
```

The reason for this is the fixed shebang of `plugins/zsh-navigation-tools/doc/generate_single_file`.

PR #12396, opened 2024-05-04
feat(python): autovenv keeps activated on subdirs

Feasuro:
## Standards checklist:

## Changes:

1. `auto_vrun` function uses `PYTHON_VENV_NAME` as pattern rather than full name.
2. `auto_vrun` function searches for venv not only in current directory but also in parent directories.

## Other comments:

1. Allows auto-activation of  venvs with descriptive names that follow naming convention: (e.g. venv_name, venv-description, venvsomething).
2. Does not deactivate venv when entering a subdirectory of a project but only when cd'ing up the directory tree. That is more natural behavior in most use cases.


carlosala (approved):
LGTM!

carlosala:
Thanks!

PR #12397, opened 2024-05-06
fix(git): add checked-out branch support to `gbg*`

filariow:
## Standards checklist:

If a branch is checked-out, `git branch` will show `*` (or `+` if it's checked-out in a worktree) as first character. This change will skip the first 2 characters that can be `*\s`, `+\s`, or `\s\s`.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -41,8 +41,8 @@ plugins=(... git)
 | `gba`                  | `git branch --all`                                                                                                              |
 | `gbd`                  | `git branch --delete`                                                                                                           |
 | `gbD`                  | `git branch --delete --force`                                                                                                   |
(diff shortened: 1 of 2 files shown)


carlosala:
Are we interested in removing the exact branch where we are? Shouldn't it be better to skip that branch?

filariow:
It will not be removed. `git` will print an error saying that the branch is checked out and it can't be deleted

carlosala:
Yep, I meant if we want that error to be there, or just skip the branch.

filariow:
Oh, sorry I didn't get that.

I'd say having the error is a good feedback so the user knows why it's gone but it wasn't deleted.

carlosala:
Thanks!

filariow:
Thank you!

PR #12408, opened 2024-05-09
feat(dependencies): improve and enable cron-based checks

### LINKED ISSUE #12178: Improve and smarten dependency updater

_No response_

The Ad-Hoc Dependency Updater™️  works, but is missing some things.

- To update an existing PR if it exists, for example after updating `.github/dependencies.yml` and re-running the tool
- To check whether an issue is existing an open, and not do anything else
- Other necessary improvements to be defined

_No response_

### Additional context

_No response_

### Related Issues

_No response_

carlosala:
## Standards checklist:

Closes #12178

## Changes:

- Use semver-based tag verification
- Enable cron-based action run

## Fixes:

- Avoid creating PR if it was created previously.

Other stuff:

- Update requirements
- Use `actions/setup-python` to ensure Python version consistency
- Run `ruff` formatter

### DIFF
--- a/.github/workflows/dependencies/updater.py
+++ b/.github/workflows/dependencies/updater.py
@@ -4,7 +4,7 @@ import subprocess
 import sys
 import timeit
 from copy import deepcopy
-from typing import Optional, TypedDict
+from typing import Literal, NotRequired, TypedDict
 
 import requests
 import yaml
@@ -49,20 +49,24 @@ class DependencyDict(TypedDict):
     repo: str
     branch: str
     version: str
-    precopy: Optional[str]
-    postcopy: Optional[str]
+    precopy: NotRequired[str]
+    postcopy: NotRequired[str]
 
 
 class DependencyYAML(TypedDict):
     dependencies: dict[str, DependencyDict]
 
 
-class UpdateStatus(TypedDict):
-    has_updates: bool
-    version: Optional[str]
-    compare_url: Optional[str]
-    head_ref: Optional[str]
-    head_url: Optional[str]
+class UpdateStatusFalse(TypedDict):
+    has_updates: Literal[False]
+
+
+class UpdateStatusTrue(TypedDict):
+    has_updates: Literal[True]
+    version: str
+    compare_url: str
+    head_ref: str
+    head_url: str
 
 
 class CommandRunner:
@@ -105,7 +109,9 @@ class DependencyStore:
         with CodeTimer(f"store deepcopy: {path}"):
             store_copy = deepcopy(DependencyStore.store)
 
-        dependency = store_copy["dependencies"].get(path, {})
+        dependency = store_copy["dependencies"].get(path)
+        if dependency is None:
+            raise ValueError(f"Dependency {path} {version} not found")
         dependency["version"] = version
         store_copy["dependencies"][path] = dependency
 
@@ -171,7 +177,7 @@ class Dependency:
                 else:
                     status = GitHub.check_updates(repo, remote_branch, version)
 
-            if status["has_updates"]:
+            if status["has_updates"] is True:
                 short_sha = status["head_ref"][:8]
                 new_version = status["version"] if is_tag else short_sha
 
@@ -212,10 +218,10 @@ Check out the [list of changes]({status['compare_url']}).
                         case CommandRunner.Exception:
                             # Print error message
                             print(
-                                f"Error running {e.stage} command: {e.returncode}",
+                                f"Error running {e.stage} command: {e.returncode}",  # pyright: ignore[reportAttributeAccessIssue]
                                 file=sys.stderr,
                             )
-                            print(e.stderr, file=sys.stderr)
+                            print(e.stderr, file=sys.stderr)  # pyright: ignore[reportAttributeAccessIssue]
                         case shutil.Error:
                             print(f"Error copying files: {e}", file=sys.stderr)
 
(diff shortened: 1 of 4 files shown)


.github/workflows/dependencies/updater.py
```diff
@@ -14,437 +19,551 @@
...
+    (?P<major>0|[1-9]\d*)
+    (\.
+    (?P<minor>0|[1-9]\d*)
+    (\.
+    (?P<patch>0|[1-9]\d*)
+    )?
+    )?
+    """,
+    re.VERBOSE,
+)
```
mcornella, line 34: This regex here fails when the version numbers begin with `0`, in which case the result are 0s:
```py
>>> coerce("v2.03.4")
Version(major=2, minor=0, patch=0, prerelease=None, build=None)
```

The `0|[1-9]` part should go inside an additional parenthesis:
```suggestion
BASEVERSION = re.compile(
    r"""[vV]?
    (?P<major>(0|[1-9])\d*)
    (\.
    (?P<minor>(0|[1-9])\d*)
    (\.
    (?P<patch>(0|[1-9])\d*)
    )?
    )?
    """,
    re.VERBOSE,
)
```

With this change the `coerce` function works again:
```py
>>> coerce("v2.3.4")
Version(major=2, minor=3, patch=4, prerelease=None, build=None)
>>> coerce("v2.03.4")
Version(major=2, minor=3, patch=4, prerelease=None, build=None)
>>> coerce("zsh2.03.4")
Version(major=2, minor=3, patch=4, prerelease=None, build=None)
>>> coerce("zsh-2.03.4")
Version(major=2, minor=3, patch=4, prerelease=None, build=None)
```

I see that the `semver` package [has a parse function](https://pypi.org/project/semver/), but I have tested it and I understand you're doing it this way because it does not work well with prefixed strings on the versions, so we'd have to remove that from tags either way. Am I right?
```py
>>> Version.parse("v2.3.4")
  File "<stdin>", line 1, in <module>
  File "/home/.../ohmyzsh/.github/workflows/dependencies/venv/lib/python3.12/site-packages/semver/version.py", line 646, in parse
    raise ValueError(f"{version} is not valid SemVer string")
ValueError: v2.3.4 is not valid SemVer string
```

.github/workflows/dependencies/updater.py
```diff
@@ -14,437 +19,551 @@
...
+
+
+def coerce(version: str) -> Optional[Version]:
+    match = BASEVERSION.search(version)
+    if not match:
+        return None
+
+    ver = {
```
mcornella, line 42: I'd add a comment here about what this is doing, including when using this variable on the `Version(**ver)` line. Looks complex enough to warrant it.

.github/workflows/dependencies/updater.py
```diff
@@ -14,437 +19,551 @@
...
+                if tag_version is None:
+                    continue
+                if latest_version is None or tag_version.compare(latest_version) > 0:
+                    latest_version = tag_version
+                    latest_ref = ref
+
+            if latest_ref is None:
+                raise ValueError(f"No tags following semver found in {repo}")
+
+            latest_tag = latest_ref["ref"].replace("refs/tags/", "")
```
mcornella, line 454: I'd add comments throughout these lines as well.

mcornella (changes requested):
Looks very nice! Only some missing edge case, but the PR is almost ready to go! :shipit:

.github/workflows/dependencies/updater.py
carlosala: It does not work with prefixed strings, and it doesn't work either with `2.0`, which lacks a `patch` section. Theoretically, in `semver`, `M.m.p` are compulsory.

carlosala: I just took that `coerce` function from [`python-semver` docs](https://python-semver.readthedocs.io/en/latest/advanced/deal-with-invalid-versions.html).

.github/workflows/dependencies/updater.py
```diff
@@ -14,437 +19,559 @@
...
+                    # we skip every tag that is not semver-complaint
+                    continue
+                if latest_version is None or tag_version.compare(latest_version) > 0:
+                    # if we have a "greater" semver version, set it as latest
+                    latest_version = tag_version
+                    latest_ref = ref
+
+            # raise if no valid semver tag is found
+            if latest_ref is None:
+                raise ValueError(f"No tags following semver found in {repo}")
```
mcornella, line 459: Wouldn't this get triggered also if no semver tag is found that is greater than `tag_version`? `latest_ref` seems to be non-None if one is found, as per the previous if.

mcornella (commented):
Found some more thingies

.github/workflows/dependencies/updater.py
mcornella: OK as we discussed offline, this is prevented by the first if condition. +1

.github/workflows/dependencies/updater.py
```diff
@@ -14,437 +19,559 @@
...
+
+            # we get the tag since GitHub returns it as plain git ref
+            latest_tag = latest_ref["ref"].replace("refs/tags/", "")
+
+            if latest_tag == current_tag:
```
mcornella, line 464: Needs checking if the found version is actually semver-greater than the current one we have (e.g. if the version tag we have is removed and the latest one is now one version less).

mcornella (approved):
LGTM!

PR #12410, opened 2024-05-10
fix(nvm): use `command cat` to avoid alias

koiralakiran1:
## Standards checklist:

## Changes:
- When `cat` is aliased to something like `bat --style=number` ([bat](https://github.com/sharkdp/bat)), `$(cat "$nvmrc_path" | tr -dc '[:print:]')` fails because of additional information that these helpful tools display. This change fixes this behavior by using non-aliased version of `cat` command, which I think is a safer option to use within zsh plugins.

## Other comments:
**Reasoning / Steps to reproduce**

image


carlosala (approved):
LGTM!

carlosala:
Thanks!

Kboxoxx (approved):

PR #12417, opened 2024-05-14
chore(template): propose add `.local/bin` to PATH

lifeisfoo:
## Standards checklist:

## Changes:

- fix(zshrc): added `~/.local/bin/` to PATH, as required by systemd, so user specific binaries can be found

## Other comments:

The vast majority of modern Linux distributions are systemd based. User specific binaries can be now placed inside `~/.local/bin/` ([SPEC](https://www.freedesktop.org/software/systemd/man/latest/file-hierarchy.html#~/.local/bin/)).

For this reason, `~/.bashrc` already adds that location to PATH. E.g. (code from a Fedora 40 brand new installation)
```
# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]; then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH
```

### DIFF
--- a/templates/zshrc.zsh-template
+++ b/templates/zshrc.zsh-template
@@ -1,5 +1,5 @@
 # If you come from bash you might have to change your $PATH.
-# export PATH=$HOME/bin:/usr/local/bin:$PATH
+# export PATH=$HOME/bin:$HOME/.local/bin:/usr/local/bin:$PATH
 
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

PR #12422, opened 2024-05-15
feat: change apple logo color based on previous return code

Ian729:
## Standards checklist:

## Changes:

1. if return code is 0, print purple apple logo
2. if return code is not 0, print return code then white apple logo

image

## Other comments:
if you think user might not want to see the exact return code, I can change to 
- purple if previous command exits successfully(return code==0)
- white if otherwise

### DIFF
--- a/themes/apple.zsh-theme
+++ b/themes/apple.zsh-theme
@@ -1,5 +1,10 @@
 function toon {
-  echo -n ""
+  r=$?
+  if [ $r -eq 0 ]; then
+    echo -n "%{$fg[magenta]%}"
+  else
+    echo -n "[$r]"
+  fi
 }
 
 autoload -Uz vcs_info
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @Ian729, thank you for the submission. Unfortunately this constitutes a breaking change, and the apple theme is pretty widely used. You can follow [this guide](https://github.com/ohmyzsh/ohmyzsh/wiki/Customization#overriding-and-adding-themes) to copy the theme to the [`$ZSH_CUSTOM` folder](https://github.com/ohmyzsh/ohmyzsh/wiki/Settings#zsh_custom) and changing its code.

A quick tip: what you need can be better achieved by using prompt sequences [[1]]:

```patch
diff --git a/themes/apple.zsh-theme b/themes/apple.zsh-theme
index 0c183258..fe292451 100644
--- a/themes/apple.zsh-theme
+++ b/themes/apple.zsh-theme
@@ -18,7 +18,7 @@ theme_precmd () {
 }
 
 setopt prompt_subst
-PROMPT='%{$fg[magenta]%}$(toon)%{$reset_color%} %~/ %{$reset_color%}${vcs_info_msg_0_}%{$reset_col>
+PROMPT='%(?.%{${fg[magenta]}%}.[%?]%{${fg[white]}%})$(toon)%{$reset_color%} %~/ %{$reset_color%}${>
 
 autoload -U add-zsh-hook
 add-zsh-hook precmd theme_precmd
```

Which would still give you the ability to override the `toon` function to put another symbol of your choosing.

[1]: https://zsh.sourceforge.io/Doc/Release/Prompt-Expansion.html

PR #12434, opened 2024-05-21
feat(agnoster): add `anaconda` env information

Ochoja:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -227,6 +227,14 @@ prompt_virtualenv() {
   fi
 }
 
+# Conda Virtualenv
+promp_conda_virtualenv() {
+  if [[ -n $CONDA_PROMPT_MODIFIER ]]; then
+    prompt_segment black default ${CONDA_PROMPT_MODIFIER:1:-2}
+  fi
+}
+
+
 # Status:
 # - was there an error
 # - am I root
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
Thanks, great addition!

mhart:
Hmmm. This has suddenly made "base" appear in *all* of my prompts. Was it intended to be shown by default, even when on the "base" environment?

themes/agnoster.zsh-theme
```diff
@@ -227,6 +227,14 @@ prompt_virtualenv() {
   fi
 }
 
+# Conda Virtualenv
+promp_conda_virtualenv() {
```
mhart, line 231: Suggestion: `prompt_conda_virtualenv`

carlosala:
Maybe we could skip base, sounds great

carlosala:
Could you make a PR?

H-Cao:
This change makes my shell prompt behave weirdly (the conda environment name appears twice). I tried to configure conda not to change PS1 in the hope that it would drop  ```(base)``` at the front of the shell prompt. But the other one disappeared as well.

Is there any way to get around this? Or should I keep using the old theme config? Thanks!
Screenshot 2024-05-21 at 23 18 27

carlosala:
I reverted the changes 👍🏻

PR #12442, opened 2024-05-22
Prepend --namespace=xyz when OPENSHIFT_NAMESPACE is set

rmonico:
Prepend --namespace=xyz when OPENSHIFT_NAMESPACE is set


robbyrussell:
Thanks for this, and sorry it sat so long without an answer. Going to close it... aliasing `oc` itself and writing straight into `$PROMPT` is more than a plugin should do here. `oc project <ns>` already handles the namespace switch, and if you want it in the prompt, the pattern we use is a `*_prompt_info` function that themes can call (see the kube-ps1 plugin).

PR #12464, opened 2024-05-30
fix(pass): completion for multiple repositories

ombre8:
## Standards checklist:

## Changes:

- changed pass plugin according to https://lists.zx2c4.com/pipermail/password-store/2019-January/003502.html I don't know why this wasn't merged upstream, but this fixes the completion if you use multiple repositories

## Other comments:

without the fix `workpass secre[tab]` works but `workpass edit secre[tab]` would search the default pass directory for that secret

### DIFF
--- a/plugins/pass/_pass
+++ b/plugins/pass/_pass
@@ -20,6 +20,8 @@
 
 _pass () {
 	local cmd
+  local rootcontext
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

PR #12467, opened 2024-06-02
fix(bgnotify): use terminal-notifier args properly

davidlj95:
## Standards checklist:

- N/A If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:
`bgnotify` notifications don't work in macOS when using latest version of `terminal-notifier`

After trying out the command calling `terminal-notifier` used to send notifications manually, could reproduce the issue. Removing `-sender` arg seems fixed the issue.

After reading [`terminal-notifier`'s README](https://github.com/julienXX/terminal-notifier/blob/2.0.0/README.markdown) seems this is actually expected:

> _`-sender ID`_
> [...]
> _Because of this it is important to note that **you cannot combine this with options like -execute and -activate** which depend on the sender of the notification to be ‘terminal-notifier’ to perform its work._

Using therefore only `-activate` argument so that the terminal is opened when tapping the notification

- Remove `-sender` flag. Use `-activate` only as instructed per `terminal-notifier` docs

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -117,7 +117,7 @@ function bgnotify {
   local icon="$3"
   if (( ${+commands[terminal-notifier]} )); then # macOS
     local term_id=$(bgnotify_programid)
-    terminal-notifier -message "$message" -title "$title" ${=icon:+-appIcon "$icon"} ${=term_id:+-activate "$term_id" -sender "$term_id"} &>/dev/null
+    terminal-notifier -message "$message" -title "$title" ${=icon:+-appIcon "$icon"} ${=term_id:+-activate "$term_id"} &>/dev/null
   elif (( ${+commands[growlnotify]} )); then # macOS growl
     growlnotify -m "$title" "$message"
   elif (( ${+commands[notify-send]} )); then


carlosala:
Hey! Is there any changelog or something? It'd be great to keep compatibility with older versions.

davidlj95:
> Hey! Is there any changelog or something, or you know which is the first version of terminal-notifier that requires it? It'd be great to keep compatibility with older versions.

After `git blame` on the README, seems that [the docs about not using `-sender` and `-activate` together dates back to 2013](https://github.com/julienXX/terminal-notifier/commit/2d3e817fbb769ba20591d085ed64e025bed19a96). Which appears to be v1.5.2 . Latest is v2.0.0 but anyway more than ten years since this was documented as not correct.

Checked when did the usage of `-activate` and `-sender` started appearing all together in this plugin. [It dates back to 2015](https://github.com/ohmyzsh/ohmyzsh/commit/7c31c7a96391e80bdd6ebdb6579751f4d44d6ac4).

Seems the issue [also exists](https://github.com/t413/zsh-background-notify/blob/d5f0430cb052f82c433c17707816910da87e201e/bgnotify.plugin.zsh#L39) in the [`t413/zsh-background-notify` repo](https://github.com/t413/zsh-background-notify) where this plugin comes from. With [an analogous PR offering same fix](https://github.com/t413/zsh-background-notify/pull/12)

My guess is it has been working until now despite not being the correct way of doing it. And maybe some macOS update dropped support for that. Just guessing though

davidlj95:
PD: The wording in the PR description about _when using latest version_ was because I'm just trying this plugin in macOS after not using this plugin for a long time (and last time I used it in Arch Linux). And installed latest version of `terminal-notifier`. But don't know if installing previous version 1.8.0 of 2017 would work. 

> I'm guessing it wouldn't anyway as looks like some kind of macOS API's issue. But again just a guess

carlosala (approved):

PR #12468, opened 2024-06-02
fix(tmux): add completion for alias functions

### LINKED ISSUE #12282: Recent update breaks autocomplete for tmux plugin aliases
### Describe the bug

The latest `_build_tmux_alias` causes the tmux `ta` alias to be defined as the following function,
```
ta () {
  if [[ -z $1 ]] || [[ ${1:0:1} == '-' ]]
  then
    tmux attach "$@"
  else
    tmux attach -t "$@"
  fi
}
```

instead of the previous command,
```
ta: aliased to tmux attach -t
```

which makes autocompletion of session names not work. When using TAB to autocomplete, instead of providing the list of sessions to choose, it provides the list of files/folders in the current directory, presumably because it doesn't know what command it's trying to autocomplete for.

### Steps to reproduce

1. run `omz update`
2. try to use `ta` TAB autocompletion as before

### Expected behavior

Ideally, pressing TAB after entering `ta` would provide a list of current sessions, as before.

_No response_

### OS / Linux distribution

macOS

### Zsh version

5.9

### Terminal emulator

iTerm2

None

### Additional context

_No response_

MoetaYuko:
Fixes #12282
Closes #12278 

## Standards checklist:

## Changes:

- Fixes completion for aliased functions (`ta`, `tad`, and `tkss`) introduced in [fix(tmux): do not pass empty flags to aliases #12232](https://github.com/ohmyzsh/ohmyzsh/pull/12232).

## Other comments:

Inspired by https://github.com/ohmyzsh/ohmyzsh/blob/e0c6cb147030350c8e27dbdeda6e8a4d367d1e66/plugins/debian/debian.plugin.zsh#L120-L133

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -57,6 +57,19 @@ function _build_tmux_alias {
       tmux $2 $3 \"\$@\"
     fi
   }"
+
+  local f s
+  f="_omz_tmux_alias_${1}"
+  s=(${(z)2})
+
+  eval "function ${f}() {
+    shift words;
+    words=(tmux ${@:2} \$words);
+    ((CURRENT+=${#s[@]}+1))
+    _tmux
+  }"
+
+  compdef "$f" "$1"
 }
 
 alias tksv='tmux kill-server'


carlosala (approved):
LGTM!

PR #12477, opened 2024-06-04
feat(nanotimer): New precise command timer plugin

2art:
## Standards checklist:

## Changes:

- Add plugin `nanotimer` for command timer prompt hook
- Add initial documentation and preview image

## Other comments:

This is my first contribution. Excuse me if I have done something in the wrong order, and please notify me in such case. This plugin times commands and displays it nicely in the right prompt, down to nanoseconds.


robbyrussell:
Would it be possible to get a screenshot that doesn't have transparency? We aren't sure we know what we're looking at. ;-)

carlosala:
Hi! We already have the `timer` plugin, which kinda does the trick!
Thanks for your contribution!

PR #12483, opened 2024-06-07
fix(eza): icons tab completion

martinbjeldbak:
## Standards checklist:

## Changes:

Explicitly pass the default `automatic` argument to `--icons` flag, if icons are configured to display. Without this argument, it causes tab-completion not to work.

See issue on Eza's tracker here: https://github.com/eza-community/eza/issues/964#issuecomment-2143417856

Also occurred to me on zsh 5.9 in Arch Linux.

## Other comments:

Follow up to PR #12469 merged 3 days ago from this PR, adding support for this option


carlosala (approved):
LGTM!

PR #12495, opened 2024-06-11
refactor: poetry improved aliases 

jeroeness:
## Standards checklist:

## Changes:

- This PR made slight changes to existing aliases. These changes optimize for efficiency: more frequent aliases are shorter and follow the same structure as the [git aliases plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/git). The aliases for `add`, `rm`, `run`,  `update`, `install`, `lock`, `shell` and `build` are now  `pa`, `prm`, `pr`, `pu`, `pi`, `pl`, `psh`, `pb` respectively

## Other comments:

One new alias introduced for adding development dependencies:
```
alias pad='poetry add -G dev' 
```


carlosala:
Hi! We'll rather keep everything as is, making such breaking changes in aliases without any particular reason is something that we try to avoid.
Thanks for your contribution!

jeroeness:
@carlosala Please mind that the aliases to poetry were first introduced to main one day prior to this MR. Please the relating MR: https://github.com/ohmyzsh/ohmyzsh/pull/11858 . In the MR I proposed changes before it would get merged. Robbyrussel probably did not notice my comment. There is still time to discuss this and make it better fit the best practices used elsewhere in ohmyzsh

PR #12500, opened 2024-06-12
fix(history): fix logic error in prompt for `history -c` (ohmyzsh#12472)

Sargates:
## Standards checklist:

## Changes:

- Fixes logic error from 35a6725 when using `history -c`
- Prompting for confirmation causes history to be deleted when typing anything but explicitly `n`, `N`, or sending `\n`. 
- New logic no longer proceeds with deletion when pressing wrong key and only deletes history when sending `y` or `Y`.

## Other comments:
Copying @mcornella as he was assigned to the original issue

### DIFF
--- a/lib/history.zsh
+++ b/lib/history.zsh
@@ -11,7 +11,7 @@ function omz_history {
     print -nu2 "This action will irreversibly delete your command history. Are you sure? [y/N] "
     builtin read -k1
     [[ "$REPLY" = $'\n' ]] || print -u2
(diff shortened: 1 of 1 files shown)


mcornella (approved):

robots4life:
How can I send a `y` to `history -c` in this alias `alias hit='history -c && exit'` with this change please ?

With this change so far now I have to confirm every time I run the alias, however I would like to not have to do that manually, thank you.

Issue #11466, opened 2023-01-29
Respect `$ZDOTDIR` in uninstaller


mcornella:
### Describe the bug

Uninstaller does not respect the actual install directory (#10537) and it won't respect the .zshrc location if installed on a different directory using `$ZDOTDIR`.

### Steps to reproduce

1. Create ZDOTDIR directory if not yet created: `mkdir -p ~/.config/zsh`
2. Install OMZ in custom `ZDOTDIR`: `ZDOTDIR=~/.config/zsh sh install.sh`
3. Run `uninstall_oh_my_zsh`
4. .zshrc and `$ZSH` folder are not removed

### Expected behavior

- Correct OMZ folder is removed
- .zshrc file is looked for in `$ZDOTDIR`

### Screenshots and recordings

```console
$ uninstall_oh_my_zsh 
Are you sure you want to remove Oh My Zsh? [y/N] y
Removing ~/.oh-my-zsh
Looking for original zsh config...
No original zsh config found
Thanks for trying out Oh My Zsh. It's been uninstalled.
Don't forget to restart your terminal!
```

### OS / Linux distribution

Any

### Zsh version

Any

### Oh My Zsh version

master (b2313ec)

### Terminal emulator

Any

### If using WSL on Windows, which version of WSL

None

PR #12503, opened 2024-06-13
Fix uninstall script

chyeh:
## Changes:
- Make the uninstall script remove the oh-my-zsh directly according to environmental variable `$ZSH`.

## Other comments:
- The original implementation was hardcoded with `~/.oh-my-zsh`, which can't remove custom installed paths. By removing where the `$ZSH` variable points to, it fixes the issue.


carlosala:
Related to #11466

PR #12508, opened 2024-06-16
feat(git): add `gclf` alias

ttytm:
## Standards checklist:

## Changes:

- This PR adds an alias that results in git clone that can be significantly less in size. While doing so, it still preserves ability to easily check out the the cloned repository history, since it is not a shallow clone (its `depth` is not being reduced). 

  The alias has become my default for cloning repositories over the recent years.
  
  For example:
  ```sh
  # gclf
  git clone --recursive --shallow-submodules --filter=blob:none --also-filter-submodules \
  https://github.com/webui-dev/webui.git webui-filtered # ~10,2MB
  ```
  ```sh
  # glc
  git clone --recurse-submodules https://github.com/webui-dev/webui.git # ~34,4MB
  ```
  
  The style of the added alias matches with other git alias like:
  https://github.com/ohmyzsh/ohmyzsh/blob/677f5010daf226509cc19d7244a689df82820c82/plugins/git/git.plugin.zsh#L111-L112
  https://github.com/ohmyzsh/ohmyzsh/blob/677f5010daf226509cc19d7244a689df82820c82/plugins/git/git.plugin.zsh#L118-L119
 

## Other comments:

...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -165,6 +165,7 @@ alias gcpa='git cherry-pick --abort'
 alias gcpc='git cherry-pick --continue'
 alias gclean='git clean --interactive -d'
 alias gcl='git clone --recurse-submodules'
+alias gclf='git clone --recursive --shallow-submodules --filter=blob:none --also-filter-submodules'
 
 function gccd() {
   setopt localoptions extendedglob


carlosala:
I think that's a great addition. Partial clones make complete sense in the modern era of monorepos and, in general, incredibly huge repos. It makes more sense to me to just remove the `--also-filter-submodules` since we're already cloning them in a shallow way.

ttytm:
> I think that's a great addition. Partial clones make complete sense in the modern era of monorepos and, in general, incredibly huge repos. It makes more sense to me to just remove the `--also-filter-submodules` since we're already cloning them in a shallow way.

Thanks for the reply, @carlosala. I added the `--also-filter-submodules` to the alias when it was introduced with Git 2.36 while intentionally keeping `--shallow-submodules`. It serves the purpose of still getting the benefits of the filtered clone when `unshallowing` a submodule. For me personally, unshallowing submodules and working in them is a rare use case, but it has happened a few times. Maybe that's a missing piece of info to keep it as it is, else please let me know if it should be reduced.

carlosala:
Sounds good, let's keep it as is if there's a rationale behind it. Could you add the alias to the readme as well?

ttytm:
Thanks! I missed the doc update. Added it.

I kept a formatter fix for one line in the readme for another alias.

carlosala (approved):
LGTM!

carlosala:
Thanks!

PR #12525, opened 2024-06-25
fix(agnoster): print white text over black

Kekun:
This makes them readable in light terminals too and not only dark ones.

## Standards checklist:

## Changes:

- Prints white text rather than default-colored text on black segments (context and status)

### Before

Capture d’écran du 2024-06-25 14-16-20

Capture d’écran du 2024-06-25 14-16-47

### After (no change for dark)

Capture d’écran du 2024-06-25 14-16-04

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -48,7 +48,7 @@ esac
 : ${AGNOSTER_DIR_BG:=blue}
 
 # user@host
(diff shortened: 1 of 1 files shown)


carlosala:
Great one! I have in my pipeline to look closely to agnoster theme, I'll reach it soon!
Thanks!

carlosala:
Thanks for your contribution! Sorry for the delay!

PR #12530, opened 2024-06-30
docs(command-not-found): mention gentoo support

BluCobalt:
SUSE's /usr/bin/command-not-found works for Gentoo, too

## Standards checklist:

## Changes:
Just a small comment that mentions that this works out of the box on gentoo.

## Small comments:
I'm not sure whether to link to the github page of the package or to the actual package on https://packages.gentoo.org. I think users would find the github page more useful because it also mentions that pfl should be installed.


carlosala (approved):
LGTM!

PR #12534, opened 2024-07-02
fix(kubectx): fix kubectx_prompt_info for contexts without spaces.

kenhkelly:
## Standards checklist:

## Changes:

- Fix the problem: When Kubernetes context has no spaces, custom names are not working. This bug was introduced in #12191 
- This expands the logic to support context names with spaces, without spaces, and non-defined.

### DIFF
--- a/plugins/kubectx/kubectx.plugin.zsh
+++ b/plugins/kubectx/kubectx.plugin.zsh
@@ -9,5 +9,14 @@ function kubectx_prompt_info() {
 
   # use value in associative array if it exists
   # otherwise fall back to the context name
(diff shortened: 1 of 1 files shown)


jlcrow (approved):
Looks great

mcornella:
The double quotes were superfluous and shouldn't have been introduced in #12191. I fixed the code and documented exactly how to define the `kubectx_mapping` array, and also added you as contributor.

Thanks for the contribution!

kenhkelly:
Sounds good! Thanks @mcornella

PR #12542, opened 2024-07-06
fix: Use `-L` to check for symlinks in conditional expressions (#12457)

### LINKED ISSUE #12457: parse error near `>&' in diagnostics.zsh
### Describe the bug

Loading `oh-my-zsh` is generating error as below

```bash
/Users/moulick.aggarwal/.oh-my-zsh/lib/diagnostics.zsh:134: parse error near `>&'
/Users/moulick.aggarwal/.oh-my-zsh/plugins/macos/music:103: parse error near `>&'
```

### Steps to reproduce

1. Run omz update to get the latest HEAD
2. run `source ~/.zshrc` 
3. See error below

```bash
/Users/moulick.aggarwal/.oh-my-zsh/lib/diagnostics.zsh:134: parse error near `>&'
/Users/moulick.aggarwal/.oh-my-zsh/plugins/macos/music:103: parse error near `>&'
```

### Expected behavior

No error

_No response_

### OS / Linux distribution

MacOS 14.5 (23F79)

### Zsh version

5.9

### Oh My Zsh version

master (5e59d21e)

### Terminal emulator

iTerm2

None

### Additional context

_No response_

ccaviness:
## Standards checklist:

## Changes:

- In conditional expressions (`[[ ... ]]`), use `-L` instead of `-h` as this avoids conflicting with a somewhat common global `-h` alias recommended by [`bat`](https://github.com/sharkdp/bat?tab=readme-ov-file#highlighting---help-messages)

Fixes #12457

## Other comments:

Tested locally. With a global `-h` alias, executing `source ~/.zshrc` in an existing session triggers:
```
❯ source ~/.zshrc
/Users/clay/.oh-my-zsh/lib/diagnostics.zsh:134: parse error near `>&'
```
With this change, no errors are emitted:
```
❯ source ~/.zshrc
```

### DIFF
--- a/lib/diagnostics.zsh
+++ b/lib/diagnostics.zsh
@@ -131,7 +131,7 @@ function _omz_diag_dump_one_big_text() {
           sha_str=$sha_str[1]
           extra_str+=" SHA $sha_str"
         fi
-        if [[ -h "$progfile" ]]; then
+        if [[ -L "$progfile" ]]; then
           extra_str+=" ( -> ${progfile:A} )"
         fi
       fi
@@ -199,7 +199,7 @@ function _omz_diag_dump_one_big_text() {
   builtin echo
   if [[ -e $ZSH_CUSTOM ]]; then
     local custom_dir=$ZSH_CUSTOM
-    if [[ -h $custom_dir ]]; then
+    if [[ -L $custom_dir ]]; then
       custom_dir=$(builtin cd $custom_dir && pwd -P)
     fi
     builtin echo "oh-my-zsh custom dir:"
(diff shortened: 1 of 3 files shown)


mcornella:
Hi Clay, appreciate your contribution and for letting us know about this, but unfortunately this does not make sense.

The global alias suggested is bad for interoperability, and you will find other issues having it set (for example, in some commands `-h` means "human" display, or sort).

It's better to fix the global alias, for example changing it to the `:h` or `:help` or any other non-common symbol. `-h` condition is the POSIX one.

Thanks for the contribution though.

ccaviness:
Oh, I'm aware of the possibilities of conflicts and accept there may be unwanted side-effects. I've encountered very few actual issues, just a couple small things like this.

This PR was just switching from one conditional (`-h`) to another (`-L`) with the exact same effect to work around a minor side-effect that only occurs when loading ohmyzsh a second time in the same session. I'm happy to just keep my own fork -- it's a tiny diff -- but thought it could help future users from seeing/reporting an issue with ohmyzsh that could be easily avoided.

Thanks for your work maintaining ohmyzsh, by the way! It's very useful.

PR #12546, opened 2024-07-08
Changed how conda enviroments interact with the bira theme

remsforian:
## Standards checklist:

## Changes:

introduces conda environments to the possible list of prompts, and integrates it into the bira theme. 

## Other comments:
This is a modification of PR #10274. I modified this code so that it does not break git integration. This version also does not slow down the terminal as much as the original PR. Thank you @elfisworking for doing most of the hard work.
...


mcornella:
Merged in be10a91. Please note that for backwards compatibility this requires using the `conda-env` plugin which sets up the prompt function. Thanks!

PR #12550, opened 2024-07-11
Simplify Pygmalion Theme

mnkhouri:
## Standards checklist:

## Changes:

The Pygmalion theme previously had some logic for counting line length and taking action based on line length. The action was removed in c52441b624c0b1fa075bc6110032bd4e75311909, but the counting still needlessly happens. 

This commit removes the code to count line length, and simplifies the Pygmalion theme by moving the "precmd" into the setup. 

## Other comments:

As a bonus unintended side effect, this makes the Pygmalion theme compatible with the current async prompting logic (i.e. after this change, Pygmalion is no longer affected by https://github.com/ohmyzsh/ohmyzsh/issues/12328)

### DIFF
--- a/themes/pygmalion.zsh-theme
+++ b/themes/pygmalion.zsh-theme
@@ -1,7 +1,7 @@
 # Yay! High voltage and arrows!
 
 prompt_setup_pygmalion(){
-  setopt localoptions extendedglob
+  setopt localoptions nopromptsubst extendedglob
 
   ZSH_THEME_GIT_PROMPT_PREFIX="%{$reset_color%}%{$fg[green]%}"
   ZSH_THEME_GIT_PROMPT_SUFFIX="%{$reset_color%} "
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @mnkhouri, thanks for the PR! This was indeed an oversight on my part. I've refactored this further to keep only the necessary bits. Thanks!

mnkhouri:
Hi @vstepko, you're welcome! Unfortunately I'm not sure how to handle `pygmalion-virtualenv`. 

In this PR, I could remove the `precmd` from `pygmalion` because it wasn't doing anything anymore. On the other hand,  `pygmalion-virtualenv` does still counts the line length ([code reference](https://github.com/ohmyzsh/ohmyzsh/blob/b1e3a82bb5a2545b495c667f8bf838c3e6033ce4/themes/pygmalion-virtualenv.zsh-theme#L45-L49)). I believe this means that the `precmd` is still necessary, and therefore I can't do the same simplification as in this PR. 

I'm not familiar enough with OMZ to understand how the precmd works, so I'm not able to help with `pygmalion-virtualenv`. Sorry :-/

PR #12559, opened 2024-07-17
feat(git): add support for tag deletion

JoranVanBelle:
## Standards checklist:

## Changes:

- Created an alias `gtd` which performs `git tag -d`. This will delete a tag only locally
- Created an alias`gtdr` which performs `git push --delete origin`. This will delete a tag on remote

## Other comments:

Since there is already some support for tags, I would love to have some more useful aliases for tags.
Some basic things would be deleting tags local and on remote

@carlosala what do you think about this proposal?


carlosala:
Hi! I don't think these aliases add considerable value. Deleting a tag seems to be a non-daily task, and besides, we're always trying to avoid adding destructive aliases to the git plugin, including the `--delete` one.
If anyone needs it often, they can, of course, be added to their `zshrc`.
Thanks for your contribution!

PR #12560, opened 2024-07-17
feat(git): add support for tag creation

JoranVanBelle:
## Standards checklist:

## Changes:

- Create a new alias (`gtc`) which can be used to create new tags. Adding a tagname is mandatory, the message is optional and will default to the tagname.

## Other comments:

While working as a software developer where continous releasing is a must, creating tags in an easy way will save us some trouble. Since `gta` is already in use, I choose to create the alias `gtc` which stands for `git tag create`.
By using this alias you will be able to quickly create tags manually and push them towards origin in one go. This could in theory replace the current `gta` alias as it does the same, but push it towards origin as well.
Another option would be to expand `gta` so that it also pushes the tag towards origin.

@mcornella / @carlosala what are your thoughts about this?

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -205,7 +205,7 @@ plugins=(... git)
 | `gta`                  | `git tag --annotate`                                                                                                            |
 | `gts`                  | `git tag -s`                                                                                                                    |
 | `gtv`                  | `git tag \| sort -V`                                                                                                            |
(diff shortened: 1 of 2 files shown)


carlosala:
Hi! In general I think `gta` is good enough for tag creation, I wouldn't actually add a new alias for something that is not very common (in general tags are used for releases, versions, etc; it is not a daily thing). I think the `zshrc` is a nice place for this alias if any other user finds it useful.
Thanks for your contribution!

PR #12561, opened 2024-07-17
docs(eza): fix variable name to enable hyperlink flag

kriths:
## Standards checklist:

## Changes:

The docs to enable the `hyperlink`-flag described setting the wrong variable. Probably just copy-pasted from the other part of the docs.

In the script where the variable is used, it's checking for the correct name. So it only seems to be a doc issue.

https://github.com/ohmyzsh/ohmyzsh/blob/be10a9127731e7c6a9a57f8c8c1789e233117ae9/plugins/eza/eza.plugin.zsh#L41-L43

## Other comments:

-

### DIFF
--- a/plugins/eza/README.md
+++ b/plugins/eza/README.md
@@ -93,7 +93,7 @@ Default: Not set, which means the default behavior of `eza` will take place.
 ### `hyperlink`
 
 ```zsh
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
Thanks!

PR #12565, opened 2024-07-18
Support version 5.10

adamhotep:
Just taking the first three characters of `$ZSH_VERSION` would result in `5.10` becoming `5.1` and then failing this check. I've updated your logic to strip off the second dot and anything that follows it, thus ensuring all of these work: `5.10` `5.10.1` `5.10.1.2.3` `5.100`.

## Standards checklist:

## Changes:

- Simple bugfix for zsh 5.10 (for whenever that comes out)


mcornella:
Thanks for the contribution! I changed it to use `is-at-least`, which is the proper zsh way.

PR #12576, opened 2024-07-23
Remove outdated completions

### LINKED ISSUE #10401: In some cases, ohmyzsh installs worse or older completion plugins
### Describe the bug

Zsh itself comes with completions for gem and adb that are newer and significantly better than those that come with oh-my-zsh. Both have been in zsh long enough. I'd suggest you simply remove them.

Many others appear to be copies from completions that upstream projects include but which distribution packagers may not always include properly in `/usr/share/zsh/site-functions`. Examples include, _docker, _fd, _ripgrep, _ag. This is all very well until the one in ohmyzsh gets out-of-date. In a few of these cases, the one upstream has been updated and is significantly better.

You've also got some that are unique to ohmyzsh. In these cases, you're not really doing a favour to the wider zsh community by not encouraging these to be contributed either to the upstream project or zsh where that isn't easy.

### Steps to reproduce

N/A

### Expected behavior

N/A

_No response_

### OS / Distro

N/A

### Zsh version

N/A

### Terminal emulator

N/A

_No response_

### Additional context

_No response_

mcornella:
## Standards checklist:

## Changes:

This PR removes 4 plugins which bundle outdated completions which are already included in Zsh or in the own package from package managers. This results in less burden to update and completion versions are always up-to-date.

Removed plugins:

- `adb`: completion already included since Zsh 4.3.14.
- `ag`: completion included within the package from package manager.
- `fd`: completion included within the package from package manager.
- `ripgrep`: completion included within the package from package manager.

Partial removals:

- `gem`: the completion was included in Zsh 5.5. With this change, we still ship and use the completion for older versions.

## Other comments:

Closes #10401

> NOTE: there may be other plugins with outdated completions.

**To merge, use Rebase strategy**

----

@carlosala I believe we should enforce the policy of rejecting completions with these possible alternatives:

- Advise contributor to push the provided completion upstream (to specific project), to Zsh, or to https://github.com/zsh-users/zsh-completions to benefit the whole Zsh ecosystem.
- Include only completion initialization in the plugin, e.g. for tools that require running a generation command, as we already do for other ones.


carlosala:
I strongly agree with bundling completions directly (besides the async generation on startup). Great initiative! 😉

carlosala (approved):
LGTM!

PR #12584, opened 2024-07-28
Removed unhelpful debug mode

Jxtopher:
## Standards checklist:

## Changes:

The return value of the 'wd' command is always '0' by default so it cannot be processed by another program.
The return value is given as degguage mode associated with an exit which does not allow it to be used with '$?'.

- Removed debug mode
- Returns a relevant value to the shell

## Example:

Before the change:
image

Now with this PR:
image

### DIFF
--- a/plugins/wd/wd.sh
+++ b/plugins/wd/wd.sh
@@ -90,7 +90,6 @@ Commands:
     clean                Remove points warping to nonexistent directories (will prompt unless --force is used)
 
     -v | --version  Print version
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @Jxtopher, we ship the official plugin from https://github.com/mfaerevaag/wd, and don't want to diverge from it.

So please submit your PR first to upstream repository. Also I believe this change should be made to the function: https://github.com/mfaerevaag/wd/blob/759cd29828108a70d58b784059552da8333087c8/wd.plugin.zsh#L15

Issue #12222, opened 2024-02-19
Load sdk-init when adding the sdk plugin


csifinos:
### If the feature request is for a plugin or theme, specify it here.

SDK plugin

### Describe the proposed feature.

When enabling the SDK plugin, the relative script (sdk.plugin.zsh) can load the sdk-init.sh script if the env SDKMAN_DIR is missing.

### Describe alternatives you've considered

SDK can be loaded automatically by adding the following in the sdk.plugin.zsh

```bash
if [[ -z "$SDKMAN_DIR" ]]; then
  if [[ -d "$HOME/.sdkman" ]]; then
    [[ -s "$HOME/.sdkman/bin/sdkman-init.sh" ]] && source "$HOME/.sdkman/bin/sdkman-init.sh"
  fi
fi
```

robbyrussell:
Update on this one: #12585 implements it, and I've asked for a small rework there (a fix to the Homebrew fallback, and moving it into the existing `sdk` plugin rather than adding a second one). Leaving this open until that lands... the PR is the place to follow along.

PR #12585, opened 2024-07-28
feat(sdkman): add basic initialization for sdkman

davidfmatheson:
## Standards checklist:

## Changes:

Address Issue #12222 to provide basic initialization for `sdkman`.

## Other comments:

The [previous sdk plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/sdk) has not been updated in 4 years and does not work with current versions of sdkman. Activating with sdkman's new init provides autocomplete already, so I thought it best to go with a new plugin with a more descriptive name.

### DIFF
--- a/plugins/sdkman/README.md
+++ b/plugins/sdkman/README.md
@@ -1,0 +1,14 @@
+# sdkman
+
+Plugin for SDKMAN, a tool for managing parallel versions of multiple Software Development Kits on most Unix based systems.
+Provides initialization for different scenarios where users may have installed.
+
+To use it, add `sdkman` to your plugins array in your zshrc file:
+
+```zsh
+plugins=(... sdkman)
+```
+
+## Requirements
+
+* [SDKMAN](http://sdkman.io/)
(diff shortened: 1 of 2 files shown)


GOODBOY008:
feature wanted! +1

robbyrussell (changes requested):
Thanks for this, and sorry it sat for over a year without a review. SDKMAN is popular enough that I'd like to get init support in.

Two things before it can land:

1. The brew fallback misfires. `export` always returns 0, so the `|| echo` never runs, and when `brew --prefix sdkman-cli` fails you end up with `SDKMAN_DIR=/libexec` exported. On Linux (no `brew`) that means everyone with the plugin enabled but no `~/.sdkman` sees `command not found: brew` on every new shell, and on macOS without the sdkman tap it's a brew error instead. I checked this with `zsh -f`. The nvm plugin has the shape I'd copy: guard on `(( $+commands[brew] ))`, check the directory exists before exporting, and stay quiet if SDKMAN isn't there.

2. I'd rather this go into the existing `sdk` plugin than add a second one. Two plugins for one tool means people pick the wrong name, and everyone who already has `sdk` enabled would get the init for free. You're right that sdkman-init now loads its own completion (through bashcompinit), so if the old `_sdk` completer is broken on current SDKMAN, dropping it in the same PR is fine by me... just say so in the description so it's a deliberate change.

Are you up for reworking it that way? I'll keep an eye out for the update.

PR #12606, opened 2024-08-08
Add .dotnet tools to PATH

Crisfole:
When using `dotnet` with installed global tools this enables users to use any installed tools (`dotnet-trace` for instance) without having to manage the PATH.

I don't have time/energy/interest in altering my commits to meet the commit style. Consider this a "fancy" Issue. If someone wants to pull it in correctly: be my guest.

## Standards checklist:

## Changes:

- Adds dotnet tools to the path. That's it.

## Other comments:

There _MAY_ be a `DOTNET_DIR` we could use instead? I don't know. But I *do* know that this works.

### DIFF
--- a/plugins/dotnet/dotnet.plugin.zsh
+++ b/plugins/dotnet/dotnet.plugin.zsh
@@ -1,4 +1,7 @@
-# This scripts is copied from (MIT License):
+# Enable usage of global dotnet tools.
+export PATH="$PATH:$HOME/.dotnet/tools/"
+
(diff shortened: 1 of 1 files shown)


carlosala:
Hi @Crisfole! Thanks for your contribution!
Nevertheless, I think I tend to recommend modifying the `PATH` manually on your `zshrc`. Other users might have different locations for their tooling.
Thanks again for your contribution, and have a nice day!

PR #6636, opened 2018-02-21
Update kubernetes kubectl aliases for port-forward and get pod by label

DrMegavolt:
Adding `k port-forward` alias 
and a shortcut to search pods by labels `kgp -l "app:hello" ...`

PR #12612, opened 2024-08-13
feat(kubectl): show labels in aliases `kgpsl` (get pod) and `kgnosl` (get node)

PixiBixi:
## Standards checklist:

## Changes:

Add 2 new kubectl aliases

## Other comments:

...

### DIFF
--- a/plugins/kubectl/README.md
+++ b/plugins/kubectl/README.md
@@ -11,124 +11,124 @@ plugins=(... kubectl)
 
 ## Aliases
 
-| Alias   | Command                             | Description                                                                                      |
-|:--------|:------------------------------------|:-------------------------------------------------------------------------------------------------|
-| k       | `kubectl`                           | The kubectl command                                                                              |
-| kca     | `kubectl --all-namespaces`          | The kubectl command targeting all namespaces                                                     |
-| kaf     | `kubectl apply -f`                  | Apply a YML file                                                                                 |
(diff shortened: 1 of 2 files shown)


mcornella:
Hi, what's the use case for this, instead of completing the argument whenever you need it?

PixiBixi:
To be honest I'm just lazy, but we can say the same for all alias? :D

mcornella:
What I meant to say is how often do you use this command, and more generally how often would you say it's needed for k8s users?

PixiBixi:
Labels are very useful on K8S to get taint... and its usually often used

mcornella:
Thanks for providing reasoning, I agree these are useful. Please add them to the plugin README as well, then I'll merge the PR. Thanks!

PixiBixi:
Fixed.

It was an already kgpl alias (which was in fact not useful), i've deleted it.

mcornella (changes requested):
The `kgpl` alias was introduced in #6636 to filter out pods by label. That is something that's very common in k8s, and removing an alias is considered a breaking change and will have many people affected.

If you want to keep yours, it'd be better to rename it to something else, such as `kgpsl`, `kgplb` or something similar.

PixiBixi:
you right! i've re-introduced the old kgpl and add mine as kgpsl

Issue #12627, opened 2024-08-19
The git-commit plugin's implementation should use git-<command> scripts in $PATH rather than adding aliases


mattmc3:
### If the feature solves a problem you have, specify it here.

I tried out the git-commit plugin, and it cluttered my `.gitconfig` with its aliases. When I stopped using the plugin, it had no way to clean up after itself. This is unexpected side-effect behavior for a plugin, and was frustrating. There's a better way to handle aliases like this.

### Describe the proposed feature.

Git supports running any script named `git-<command> "$@"` in you `$PATH` when you run `git <command> arg1 arg2...`. That means that the OMZ `git-commit` plugin could be simplified to something like:

```zsh
# git-commit.plugin.zsh
path=(${0:h}/commands $path)
```

And then add `git-foo` scripts into `$ZSH/plugins/git-commit/commands`. This would have a few benefits:

1. It's easier to maintain a real script [instead of a script string](https://github.com/ohmyzsh/ohmyzsh/blob/2056aeeeaddd977eb205619c6f236b94dac896be/plugins/git-commit/git-commit.plugin.zsh#L32-L55), or inside of `.gitconfig`.
2. It's more canonical - rather than adding big shell scripted commands as large stringy aliases in `.gitconfig`, it's a commonly used git feature to add subcommands this way.
3. Because it's more canonical, it serves better to teach users how to add their own subcommands themselves. Many view OMZ as a teaching tool for their own config, and this hacky implementation isn't good for teaching best practices.
4. There's no cleanup needed. If you stop using the plugin, those commands stop being part of your `$PATH`.

Downsides:
1. Bash/Fish/other shells don't get the benefit of the side effects of having added `git-commit` in your Zsh config
2. The scripts need to be executable (`chmod u+x git-foo`), but that's not too difficult.

I am more than willing to submit a PR with a re-write that simplifies this plugin, but I don't want to do that without opening this issue first for feedback.

PR #12628, opened 2024-08-19
feat(git-commit): Use git-dash commands

mattmc3:
Issue: ohmyzsh/ohmyzsh#12627

This commit refactors the prior implementation which used git aliases and modified .gitconfig. This implementation uses proper git-dash commands, which are all just symlinks to one handler - omz_git_commit.

## Standards checklist:

## Changes:

See #12627

### DIFF
--- a/plugins/git-commit/README.md
+++ b/plugins/git-commit/README.md
@@ -1,7 +1,6 @@
 # git-commit plugin
 
-The git-commit plugin adds several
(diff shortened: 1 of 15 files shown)


mattmc3:
If using symlinks for the git commit commands gives anyone heartburn, we could also make them a stub that calls the `omz_git_commit` command with a new "type" parameter. I liked the symlink method, but that is a viable alternative. Other feedback welcome.

PR #12632, opened 2024-08-22
Create setup.bat, batch script to automate setup and running on Windows.

Thepathakarpit:
## Standards checklist:

## Changes:

- This is a batch script which automates whole setup on windows it does the following:
- Installs Chocolatey if it's not already installed.
- Uses Chocolatey to install Git if it's not already installed.
- Installs Windows Subsystem for Linux (WSL) if it's not already installed.
- Installs Ubuntu on WSL (you can change this to another distribution if preferred).
- Inside the WSL Ubuntu environment, it updates the package list, installs Zsh and curl, and then installs Oh My Zsh.
- Sets Zsh as the default shell in the WSL environment.
- You can just run setup.bat and it will install it all. 

## Other comments:

...

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -1,0 +1,40 @@
+@echo off
+
+:: Check if Chocolatey is installed
+where choco >nul 2>nul
+if %errorlevel% neq 0 (
+    echo Installing Chocolatey...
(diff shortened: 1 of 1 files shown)


carlosala:
Hi! We're not really looking forward to adding external install scripts. These tend to depend massively on the user environment.
Thanks for your contribution!

PR #12633, opened 2024-08-22
Create setup.sh, bash script to automate the setup and run in Linux based sys...

Thepathakarpit:
…rating systems

## Standards checklist:

## Changes:
- It is a bash script to automate the setup of Oh my Zsh:
- This script does the following:
- Defines a function to check if a command exists.
- Installs prerequisites (Zsh, Git, and curl) if they're not already installed. It supports apt-get, yum, and brew package managers.
- Installs Oh My Zsh using the official install script.
- Sets Zsh as the default shell.

## Other comments:

...

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# Function to check if a command exists
+command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
+
+# Check and install prerequisites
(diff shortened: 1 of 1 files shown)


carlosala:
Hi! We're not really looking forward to adding external install scripts. These tend to depend massively on the user environment.
Thanks for your contribution!

PR #12634, opened 2024-08-22
fix(ys): change # to % for root

xPaw:
## Standards checklist:

## Changes:

I remember 'ys' theme at some point showing '%' when using root (but maybe I am misremembering), and the prompt format comment in this file also indicates "PRIVILEGES USER", but it always shows blue # character.

This PR changes it to a red % character under root.

image


rwmitchell:
I don't know what 'ys' used to do, but having '#' as the prompt for root has been a long honored tradition since unix escaped bell labs.  $ = sh, % = csh, # = root in either.

PR #12641, opened 2024-08-28
fix(scd): don't use hardcoded path to sort executable

bjsowa:
## Standards checklist:

## Changes:

- Replaces `/usr/bin/sort` with `sort`

## Other comments:

The scd plugin assumes the `sort` executable is available under `/usr/bin` which is not the case for every distribution (NixOS here)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12644, opened 2024-09-01
Add clear alias to common-aliases.plugin.zsh

JuanVqz:
## Standards checklist:

## Changes:

- [x] Add `alias c='clear'` alias

## Other comments:

...


carlosala:
Hi! The term can be cleared using `Ctrl+l` or `Cmd+l`. Adding an alias to meet the same purpose seems like overkill.
Nevertheless, I personally use this alias that you just added, but I have it in my `zshrc`. Let's not pollute alias namespace for everyone.
Thanks for your contribution!

PR #12660, opened 2024-09-08
feat(web-search): add chatgpt

heytulsiprasad:
## Standards checklist:

## Changes:

- Add search with chatgpt --> "https://chat.openai.com?q=";
- Add new aliases in README.md ohmyzsh/plugins/web-search/README.md

## Other comments:

ChatGPT like LLMs are a great developer resource while building stuff and I've been using it since a long time to debug stuff, would be great to have an util in the CLI as well.

We can do other LLMs as well like perplexity but not sure if they support search through query. 

Also it'll work with `!!` operator to search about last query using chatgpt (which works for all the other alias as well)
...

### DIFF
--- a/plugins/web-search/README.md
+++ b/plugins/web-search/README.md
@@ -50,6 +50,7 @@ Available search contexts are:
 | `npmpkg`              | `https://www.npmjs.com/search?q=`               |
 | `packagist`           | `https://packagist.org/?query=`                 |
 | `gopkg`               | `https://pkg.go.dev/search?m=package&q=`        |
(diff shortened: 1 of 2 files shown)


carlosala (changes requested):
It seems to work as well with `https://chatgpt.com/?q=`. I'd go for this one since it seems to be the recommended URL now.

carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for your contribution!

heytulsiprasad:
Wow that was fast! ⚡ 

Keep up the good work @carlosala 💜

PR #12670, opened 2024-09-13
feat(functions): add `takezip`

### LINKED ISSUE #12669: Add zip file support to take function

take

I use the `take` function pretty often to pull down archives or git repositories from remote servers. An issue I frequently encounter is that `take` doesn't support zip archives. So if I try to do `take https://site.com/path/to/file.zip`, `take` treats it as a directory and passes it to `takedir`, and then I'm left with a mess of directories in my current path, like `~/dev/http:/site.com/path/to/file.zip`. 

I'd like to add a `takezip` function to `lib/functions.zsh` that can handle zip files. After checking the `$1` for the https protocol and a tar/gz/bz2/xz etc extension, we'd add a similar `elif` that checks for a zip extension. If found, it calls `takezip` which then fetches the archive to a temp directory, unzips it, and moves the files to the working directory, just as `takeurl` works.

_No response_

### Additional context

_No response_

### Related Issues

_No response_

eriktdesign:
Fixes #12669

## Standards checklist:

## Changes:

- Adds a `takezip` function to the `functions.zsh` file to handle ZIP downloads with `take`
- Adds a conditional to the `take` function to call `takezip` if the URL is a ZIP file.

## Other comments:

...

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -57,6 +57,16 @@ function takeurl() {
   cd "$thedir"
 }
 
+function takezip() {
+  local data thedir
+  data="$(mktemp)"
+  curl -L "$1" > "$data"
+  unzip "$data" -d "./"
+  thedir="$(unzip -l "$data" | awk 'NR==4 {print $4}' | sed 's/\/.*//')"
+  rm "$data"
+  cd "$thedir"
+}
+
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #12674, opened 2024-09-14
fix(git): update prefix map for modified files

saravanaj:
## Standards checklist:

## Changes:

- Files with short status `M` were incorrectly being considered as `ADDED` instead of `MODIFIED`. This updates the mapping to `MODIFIED`

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -227,7 +227,7 @@ function _omz_git_prompt_status() {
   prefix_constant_map=(
     '\?\? '     'UNTRACKED'
     'A  '       'ADDED'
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
Thanks for your contribution!

CyJimmy264:
> Files with short status M were incorrectly being considered as ADDED instead of MODIFIED. This updates the mapping to MODIFIED

This breaks my mind, rollbacked it for myself

saravanaj:
@CyJimmy264 - what's so complicated?

PR #12679, opened 2024-09-15
feat(bgnotify): add config to pass extra args

Cryolitia:
## Standards checklist:

## Changes:

Add a new option `bgnotify_extraargs` for `bgnotify` to pass user defined args to notifier

It is useful for some notifier. e.g. `notify-send` will by default send a persistent notification keeping in notification bar. We can add an arg `-e` to `notify-send` so that the notification will be transient.

Other useful extra args for `notify-send` could be `--urgency` to specifies the urgency level

I believe in that it is also useful for other notifier than `notify-send`


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12685, opened 2024-09-20
feat(grep): exclude Python virtualenv from `grep`

holmboe:
The convention to use `.venv/` directories for Python virtualenv's are widespreasd. This directory is huge and gives very little value when grepping in a source code directory.

## Standards checklist:

## Changes:

- exclude Python virtualenv (.venv) dir from the grep alias

## Other comments:

nope


mcornella:
I've seen some convention to put it in `venv` as well, without being hidden. Could you also include that?

PR #12691, opened 2024-09-23
feat(k9s): add completion plugin for `k9s`

shubhamrajvanshi:
This adds k9s command completions as plugin


mcornella:
Thanks for the contribution! ✨

PR #12697, opened 2024-09-25
feat(systemadmin): add `ping6` alias with count limit

aschaber1:
## Standards checklist:

## Changes:

- feat(systemadmin): Add alias for `ping6` adding a count limit

## Other comments:

- Reason for new alias: `ping` already has the limit set, ipv6 becomes more and more widespread, hence `ping6` should receive the same treatment.
- Can't reopen #4997 so this is the new one.


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12705, opened 2024-09-30
feat(jonathan): add virtualenv support

kutukvpavel:
## Standards checklist:

## Changes:

- Add support for virtualenv prompt mangling in Jonathan theme. Relies on `virtualenv` plugin.

## Other comments:

Venv prompt mangling is a problem for themes with a horizontal bar over the prompt. It causes the first line of the prompt to overflow and wrap, turning into an unreadable mess. This PR fixes it for Jonathan, following the existing approach used to support Ruby prompts.


carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for your contribution!

PR #12708, opened 2024-10-01
fix(fastfile): use idiomatic expressions

gsspdev:
'default' is not a standard command in zsh causing this plug-in to break in vanilla zsh. By substituting 'default' for zsh's built-in parameter expansion we are able to assign default values to unset variables while remaining compatible with vanilla zsh. Without this change, user's configs are liable to break.

## Standards checklist:

## Changes:

- FIX: Two 'command not found' that broke the plug-in in vanilla zsh environments.
- REFACTOR: Replaced external command invocations with built-in zsh parameter expansions.

## Other comments:

...

### DIFF
--- a/plugins/fastfile/fastfile.plugin.zsh
+++ b/plugins/fastfile/fastfile.plugin.zsh
@@ -5,8 +5,8 @@
 # If they are not set yet, they will be
 # overwritten with their default values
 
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12711, opened 2024-10-01
fix: changed minimal theme to use the default git prompt

eliOcs:
# Context
I made this change locally and it fix the git prompt for me. I noticed that there are many other themes using the `git_prompt_info` function so probably my approach isn't the correct way to fix the root cause of the issue but I'm willing to make the proper change if I get some guidance.

# Changes
- Use the default git prompt function in the minimal theme.

### DIFF
--- a/themes/minimal.zsh-theme
+++ b/themes/minimal.zsh-theme
@@ -19,7 +19,7 @@ vcs_status() {
   elif (( ${+functions[in_hg]} )) && in_hg; then
     hg_prompt_info
   else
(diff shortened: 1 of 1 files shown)


carlosala:
Hi! Please use [this suggestion](https://github.com/ohmyzsh/ohmyzsh/issues/12328#issuecomment-2400576507) to ensure that git prompt gets loaded.
We're working on a better solution.
Thanks for your contribution!

PR #12717, opened 2024-10-02
Add VirtualBox plugin providing status prompt and aliases

Zedeldi:
## Standards checklist:

## Changes:

- Add new plugin `virtualbox.plugin.zsh` to provide a status prompt and aliases for frequently used commands to manage a [VirtualBox](https://www.virtualbox.org) instance.

## Other comments:

Additional aliases assist with frequent administration and management of virtual machines, allowing actions such as starting, creating or deleting a virtual machine to be much simpler.

For example: `vbox-poweroff "Arch Linux"` => `VBoxManage controlvm "Arch Linux" poweroff`

Additionally, a configurable status prompt `virtualbox_prompt_info` can be used to monitor the health of all and/or individual virtual machines.

### DIFF
--- a/plugins/virtualbox/README.md
+++ b/plugins/virtualbox/README.md
@@ -1,0 +1,81 @@
+# VirtualBox plugin
+
+The `virtualbox` plugin provides many useful aliases for VirtualBox.
+
+To use it, add `virtualbox` to the plugins array of your `.zshrc` file:
+
+```zsh
+plugins=(... virtualbox)
+```
+
+## Aliases
+
+| Alias                  | Command                               | Description                                                        |
+|:-----------------------|:--------------------------------------|:-------------------------------------------------------------------|
+| `vbox-start`           | `VBoxManage startvm`                  | Start a virtual machine                                            |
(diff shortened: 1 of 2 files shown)


robbyrussell (changes requested):
Sorry this sat for two years without a reply. The aliases look fine. The prompt function is what I'd want changed before merging:

1. It runs `VBoxManage` 2+N times per prompt draw (once for the count, once for the total, once per VM). With `VBoxManage list` taking a few hundred ms each, that's a second or more on every Enter. One `list vms` and one `list runningvms` per call, parsed once, would do it.
2. `wc -l` pads with spaces on macOS, so the count shows as `[       1 /        2]`. Counting lines in zsh instead (`${#${(f)"$(VBoxManage list vms)"}}`) avoids both that and one of the spawns.

Smaller: a `(( $+commands[VBoxManage] )) || return` at the top of the prompt function so it stays quiet when VirtualBox isn't installed.

I'll be honest that I'm a bit unsure about demand for this one... nobody else has asked in the time it's been open. But if you're still using it and want to make those changes, I'm happy to take another look.

PR #12718, opened 2024-10-02
refactor(ohmyzsh): simplify and improve code

chikoYEAT:
## Standards checklist:

## Changes:

- Simplified the ANSI formatting function and its conditional definition.
- Improved grouping of related operations with clearer comments.


carlosala:
Hi! These changes don't really add much value, and we're quite happy with the current state of the initial script. I think the code is readable enough.
Thanks for your contribution!

mcornella:
Let me also add that, the majority of the changed code is not `sh`-compatible, which was one of the decisions made to detect any misconfiguration.

PR #12720, opened 2024-10-03
fix(theme-chooser): use `env` in shebang

matifali:
When the zsh shell is installed at a non-default location like  `~/.nix-profile/bin/zsh` for installation via nix. I was getting

```console
zsh: ./.oh-my-zsh/tools/theme_chooser.sh: bad interpreter: /bin/zsh: no such file or directory
```

This change fixes that issue.

## Standards checklist:

## Changes:

- Replaces the shebang to make it more portable

```suggestion
- #!/bin/zsh
+ #!/usr/bin/env zsh
```

### DIFF
--- a/tools/theme_chooser.sh
+++ b/tools/theme_chooser.sh
@@ -1,4 +1,4 @@
-#!/bin/zsh
+#!/usr/bin/env zsh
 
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #12726, opened 2024-10-05
fix(tailscale): load completion when `tailscale` is an alias

Hyffer:
## Standards checklist:

## Changes:

- only changes on tailscale plugin, no new aliases introduced.

## Other comments:

according to https://tailscale.com/kb/1080/cli?tab=macos, if a user is using App Store version of tailscale, it would be an alias tailscale="/Applications/Tailscale.app/Contents/MacOS/Tailscale". Current tailscale plugin does not work in that case.

cc @lukeab


mcornella (approved):

lukeab:
Thanks for the note on aliases, good to know. Delighted the PR got merged.

PR #12727, opened 2024-10-05
Typo and defaults fix README.md

JOJ0:
Defaults stated in the docs seem to be incorrect: 

https://github.com/ohmyzsh/ohmyzsh/blob/62cf1201b031399e7251abeee859e895ee825a48/plugins/alias-finder/alias-finder.plugin.zsh#L17-L19

## Standards checklist:

## Changes:

- [...]

## Other comments:

...


mcornella:
`zstyle -t` returns false by default, so the condition will evaluate to false and the settings will not be enabled:

https://github.com/ohmyzsh/ohmyzsh/blob/6a10a4d6a3004b39e0bbcff530f97c95cad5127d/plugins/alias-finder/alias-finder.plugin.zsh#L17-L19

PR #12750, opened 2024-10-13
feat(helmfile): introduce helmfile plugin

nervo:
## Standards checklist:

## Changes:

Introduce [Helmfile](https://helmfile.readthedocs.io) plugin.

## Other comments:

Largely inspired by the helm plugin ;)


robbyrussell:

I'm going to close it. It's put together the right way, but nobody else has asked for a helmfile plugin in the two years this has been open, and I don't think we should carry another plugin in the main tree without that.

It works exactly the same from `$ZSH_CUSTOM/plugins/helmfile`, and it survives updates there. Thanks again for contributing.

PR #12752, opened 2024-10-13
Add buf cli plugin

LinPr:
## Standards checklist:

## Changes:
Introduce [buf](https://github.com/bufbuild/buf) plugin.

## Other comments:

...

### DIFF
--- a/plugins/buf/README.md
+++ b/plugins/buf/README.md
@@ -1,0 +1,11 @@
+# Buf plugin
+
+This plugin adds completion for [Buf CLI](https://github.com/bufbuild/buf), a tool working with Protocol Buffers.
+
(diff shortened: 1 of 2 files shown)


carlosala (changes requested):
I like it! Could you maintain the same structure as in `deno` or `gh` plugins, instead of inlining the completion script? I see it can be generated with `buf completion zsh`.
Thanks!

LinPr:
> I like it! Could you maintain the same structure as in `deno` or `gh` plugins, instead of inlining the completion script? I see it can be generated with `buf completion zsh`. Thanks!

Sure，I will close this PR and open up a new one that will match the same structure as in `deno` or `gh` plugins! thanks for your suggestion!

PR #12767, opened 2024-10-20
feat(plugins/kubectl): add kdpa alias

PixiBixi:
## Standards checklist:

## Changes:

- Add alias to describe all pods at once

## Other comments:

...


PixiBixi:
@mcornella are you here? :)

mcornella:
Hi there, I don't see a justification or use case for why this alias is needed.

- Why do you use it?
- How often do you use it?
- What value will it provide other plugin users?

PixiBixi:
* I use it to retrieve informations on my deployments
* I often use it
* It will provide the same value as other alias?

PR #12772, opened 2024-10-22
Podman Compsoe plugins added. To-Do: README

rttss-sahil:
## Standards checklist:

## Changes:

- Added `podman-compose` plugin for aliases.

## Other comments:

Use-case for Plugin User:

This plugin introduces aliases for podman-compose[https://docs.podman.io/en/latest/markdown/podman-compose.1.html]

...


carlosala:
Hi! This PR seems to add many things at the same time, and too many aliases. The completion seems like a good idea though! Could you please open a new PR if you're still interested just with the completion, and we can iterate from there on?
Thanks!

PR #12773, opened 2024-10-23
add new alias to clear cache of pacman and yay together

shosseinib:
## Standards checklist:

## Changes:

**Here's the rephrased pull request in Markdown format:**
- Add new alias `allcl` which runs the following commands:

```bash
sudo pacman -Scc; yay -Scc
```

**The semicolon (`;`) is used instead of the ampersand (`&&`) to ensure that the first command runs correctly before the second command is executed.**
- Updated `README.md` for this alias


ratijas:
> The semicolon (;) is used instead of the ampersand (&&) to ensure that the first command runs correctly before the second command is executed.

That sounds backwards

shosseinib:
> That sounds backwards
What is you suggestion?

carlosala:
Doesn't `yay` already does `pacman -Scc` behind the scenes? I never used `yay`, but at least in `paru` I guess it works that way.

ratijas:
> > That sounds backwards
> 
> What is you suggestion?

The operators you described work the other way than how you presented it. My suggestion would be to use ` && ` ampersand operator because that is the one that actually checks the return value and short-circuits if the left hand side command fails.

shosseinib:
> Doesn't `yay` already does `pacman -Scc` behind the scenes? I never used `yay`, but at least in `paru` I guess it works that way.

You're right I close this PR.

PR #12784, opened 2024-10-28
feat(mise): add ZSH_MISE_AUTOEXPORT_VERSIONS setting to export MISE_T…

shtrom:
…OOL_* variables on directory change

## Standards checklist:

- [?] The code is efficient, to the best of my ability, and does not waste computer resources.
- [-] If the code introduces new aliases, I provide a valid use case for all plugin users down below.

## Changes:

- add ZSH_MISE_AUTOEXPORT_VERSIONS setting to export MISE_TOOL_{toolname}

## Other comments:

- This is useful for integrating the relevant version number in shell prompts
- This parses the output of `mise ls` with `sed` and outputs a script string containing a series of `export` that is, you guessed it, `eval`. Happy for other suggestions
- Performance seems OK, but I have a beefy machine (which was slow with Pyenv nonetheless). I have gated the behaviour with a `$ZSH_MISE_AUTOEXPORT_VERSIONS`, which needs to be set to `true ` for this to be enabled.
- Tested with GNU `sed`. It would be worth testing on a Mac / BSD.

### DIFF
--- a/plugins/mise/README.md
+++ b/plugins/mise/README.md
@@ -30,3 +30,7 @@ mise use -g node@system   Use system node as global default
 mise install node@20.0.0  Install a specific version number
 mise use -g node@20       Use node-20.x as global default
 ```
+
+## Setting
+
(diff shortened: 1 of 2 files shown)


robbyrussell:

I'm going to pass on it. Two things drive that: nothing in the tree reads these variables and nobody else has asked for them in two years, and the sed is GNU-only (as you noted), so on macOS the hook would run on every directory change and produce nothing. mise can also do this itself now, with a tools-aware entry in its `[env]` config, for example `NODE_VERSION = { value = "{{ tools.node.version }}", tools = true }`, which its own activation hook keeps current: https://mise.jdx.dev/templates.html

Closing for that reason. If that template route doesn't cover what you needed, say so and I'll take another look.

PR #12797, opened 2024-11-07
feat(bazel): add bzlmod integrity gen function

hofbi:
## Standards checklist:

For https://registry.bazel.build/ files and archives use an integrety value to proof that you get the correct artifact. However, getting this value is not that trivial and so far not yet documented. https://github.com/bazelbuild/bazel/issues/17124 is the issue for adding the documentation but is open for about 2 years.

This PR adds the generation as function to the bazel plugin.

## Changes:

- Add function for generating bzlmod integrety value
- Fix indention of the markdown table for the aliases

## Other comments:

...

### DIFF
--- a/plugins/bazel/README.md
+++ b/plugins/bazel/README.md
@@ -1,6 +1,7 @@
 # Bazel plugin
 
-This plugin adds completion and aliases for [bazel](https://bazel.build), an open-source build and test tool that scalably supports multi-language and multi-platform projects.
+This plugin adds completion and aliases for [bazel](https://bazel.build), an open-source build and test tool
+that scalably supports multi-language and multi-platform projects.
 
 To use it, add `bazel` to the plugins array in your zshrc file:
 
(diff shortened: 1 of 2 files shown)


plugins/bazel/bazel.plugin.zsh
```diff
@@ -3,3 +3,7 @@ alias bzb='bazel build'
 alias bzt='bazel test'
 alias bzr='bazel run'
 alias bzq='bazel query'
+
+bzlmod_integrety() {
```
hofbi, line 7:
```suggestion
sri-hash() {
```

plugins/bazel/README.md
```diff
@@ -16,7 +16,13 @@ The plugin has a copy of [the completion script from the git repository][1].
...
+| bzb     | `bazel build`                          | The `bazel build` command                              |
+| bzt     | `bazel test`                           | The `bazel test` command                               |
+| bzr     | `bazel run`                            | The `bazel run` command                                |
+| bzq     | `bazel query`                          | The `bazel query` command                              |
+
+## Functions
+
+| Function | Description                              |
+| -------- | ---------------------------------------- |
+| bzlmod_integrety  | Generate bzlmod integrity value |
```
hofbi, line 28:
```suggestion
| sri-hash  | Generate SRI hash used by bzlmod |
```

hofbi:
Applied suggestions from @fmeum

lalten:
Consider https://github.com/bazelbuild/bazel/issues/17124#issuecomment-2462877192 which doesn't need openssl

hofbi:
> Consider [bazelbuild/bazel#17124 (comment)](https://github.com/bazelbuild/bazel/issues/17124#issuecomment-2462877192) which doesn't need openssl

I think you approach does not work as convenient for local files.

carlosala:
Thanks for your contribution!

PR #12805, opened 2024-11-14
feat(adoc): introduce asciidoctor function

t-h2o:
## Standards checklist:

## Changes:

- create asciidoctor function with useful parameters to generate an `index.html` into a `docs` directory.

## Other comments:

...


carlosala:
Hi! This plugin seems to fulfill your particular needs when using `adoc`. I think the best solution in this case is to create a script per project you want to use `adoc`, to customize it to the project needs.
Thanks for your contribution!

PR #12806, opened 2024-11-15
New yarn alias: ydd (yarn dedupe)

EinfachAleks:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...


carlosala:
Hi Aleks, thanks for your contribution!
I find `yarn dedupe` helpful, and I use it often in the projects I have with yarn. Nevertheless, I don't think it is a command that is generally executed very often, and maybe adding a three-letter alias to all users is not the best idea. It is probably better suited for your `zshrc`.
Thanks for your contribution!

PR #12812, opened 2024-11-19
Make the perl plugin activate perlbrew

jamesrtnz:
## Standards checklist:

## Changes:

- The current perl plugin defines a number of aliases for 'perlbrew' but doesn't in itself actually activate perlbrew; this PR adds (optional) perlbrew activation to the plugin.  It will check if the 'perlbrew' function is defined; and if not, attempt to source the activation code included in perlbrew.  The perlbrew root directory can be defined by the standard PERLBREW_ROOT variable; and then entire activation can be disabled by setting PERLBREW_NO_AUTO_ADD to some value before sourcing oh-my-zsh.

## Other comments:

Yes, the perlbrew code being sourced is actually named "bashrc"; but it appears to work correctly in zsh to the best of my knowledge and testing.

### DIFF
--- a/plugins/perl/perl.plugin.zsh
+++ b/plugins/perl/perl.plugin.zsh
@@ -54,3 +54,29 @@ pgs() { # [find] [replace] [filename]
 prep() { # [pattern] [filename unless STDOUT]
     perl -nle 'print if /'"$1"'/;' $2
 }
+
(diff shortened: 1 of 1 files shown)


jamesrtnz:
Closed to resubmit via a branch from my fork

PR #12813, opened 2024-11-19
fix(pipenv): fix auto-shell functionality when cd-ing out

### LINKED ISSUE #12184: Need to "cd .." twice to exit a Pipenv folder when pipenv plugin is enabled
When the "pipenv" plugin is enabled, we need to "cd .." twice instead of once to move back to the parent folder.

Under the hood, the auto-activation feature of the pipenv plugin (which I could no longer work without) spawns a subshell on entry and exits it on leave. Therefore the first "cd .." is actually an exit to the parent shell which is still in the previously newly entered folder.

https://github.com/ohmyzsh/ohmyzsh/blob/80c114cb3a64044ea50b623f96a35bc022db5e8d/plugins/pipenv/pipenv.plugin.zsh#L15-L38

Actually, this has some more drawbacks:
* Any non-exported variable in the parent shell is masked when entering the new directory
* Any variable set or exported in the subshell will be lost when exiting the env
* A double "Ctrl+D" would also be needed to exit

and possibly others. Of course, non of them are deal breakers, rather simple day-to-day paper cuts / intuitive behaviors.

**Workaround**:

```shell
# Insert before the plugins declaration
zstyle ':omz:plugins:pipenv' auto-shell no

# Insert after the plugin declaration
# --> Replace 'pipenv shell' by 'source "$(pipenv --venv)/bin/activate"'
# --> Replace 'exit' by 'deactivate'
_togglePipenvShell() {
  # deactivate shell if Pipfile doesn't exist and not in a subdir
  if [[ ! -f "$PWD/Pipfile" ]]; then
    if [[ "$PIPENV_ACTIVE" == 1 ]]; then
      if [[ "$PWD" != "$pipfile_dir"* ]]; then
        unset PIPENV_ACTIVE
        deactivate
      fi
    fi
  fi

  # activate the shell if Pipfile exists
  if [[ "$PIPENV_ACTIVE" != 1 ]]; then
    if [[ -f "$PWD/Pipfile" ]]; then
      export pipfile_dir="$PWD"
      source "$(pipenv --venv)/bin/activate"
      export PIPENV_ACTIVE=1
    fi
  fi
}
autoload -U add-zsh-hook
add-zsh-hook chpwd _togglePipenvShell
_togglePipenvShell
```

If this is acceptable to you, I'd gladly open a PR with the modified code :slightly_smiling_face:

mcornella:
## Standards checklist:

## Changes:

This change uses traditional `activate` instead of `pipenv shell` so that transitions between pipenv projects happen in the same shell and it's possible to properly deactivate the shell.

Fixes #12184

### DIFF
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -19,7 +19,8 @@ if zstyle -T ':omz:plugins:pipenv' auto-shell; then
     if [[ ! -f "$PWD/Pipfile" ]]; then
       if [[ "$PIPENV_ACTIVE" == 1 ]]; then
         if [[ "$PWD" != "$pipfile_dir"* ]]; then
-          exit
+          unset PIPENV_ACTIVE pipfile_dir
+          deactivate
         fi
       fi
     fi
@@ -28,7 +29,8 @@ if zstyle -T ':omz:plugins:pipenv' auto-shell; then
     if [[ "$PIPENV_ACTIVE" != 1 ]]; then
       if [[ -f "$PWD/Pipfile" ]]; then
         export pipfile_dir="$PWD"
-        pipenv shell
+        source "$(pipenv --venv)/bin/activate"
+        export PIPENV_ACTIVE=1
       fi
     fi
   }


carlosala (approved):
LGTM! 🚀

MHM5000:
the twice `cd ..` was annoying, but now after upgrading, if I create a new tab, in macos , iterm, it doesn't source properly.
I have to do a cd .., then create a new tab, then move back to the project
Does anybody else have the same issue?
soundless video: https://app.usebubbles.com/rPnndb9JZE9UxeS5PY1xXS/omz-issue
@carlosala @mcornella

mcornella:
@MHM5000 I can't reproduce this with macOS and iTerm2. Please raise a new issue, create a log file with command below and send log file to logs@ohmyz.sh.

```sh
# run this from folder with Pipenv file
zsh -xvic exit &>~/omz-debug.log
```

PR #12814, opened 2024-11-19
feat(perl): add `perlbrew` auto activation

jamesrtnz:
## Standards checklist:

## Changes:

- Add code to perl plugin to activate perlbrew if 'perlbrew' function not defined already 

## Other comments:

Currently, the PERL plugin defines some aliases for perlbrew, but doesn't check or activate perlbrew.  This PR adds (optional) perlbrew activation to the plugin. It will check if the 'perlbrew' function is defined; and if not, attempt to source the activation code included in perlbrew. The perlbrew root directory can be defined by the standard PERLBREW_ROOT variable; and then entire activation can be disabled by setting PERLBREW_NO_AUTO_ADD to some value before sourcing oh-my-zsh.

### DIFF
--- a/plugins/perl/README.md
+++ b/plugins/perl/README.md
@@ -8,30 +8,36 @@ To use it, add `perl` to the plugins array in your zshrc file:
 plugins=(... perl)
 ```
 
+## Perlbrew activation
+
+If the plugin detects that `perlbrew` hasn't been activated, yet there is an installation of it in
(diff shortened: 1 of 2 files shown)


carlosala (approved):
LGTM 🚀

carlosala:
Thanks for your contribution! I refactored a bit the code and documented before merging 😊

PR #12822, opened 2024-11-24
For chucknorris plugin: add note about possibly needing to generate fortune data

keithrbennett:
## Standards checklist:

## Changes:

When I installed this plugin and tried to run it, I got an error indicating it could not find the fortune data. The fix was to run `strfile $ZSH/plugins/chucknorris/fortunes/chucknorris\n`.

I would like this to be included in the README for this plugin for the sake of future users.

This PR adds a note:

"In addition to installing `fortune`, it may be necessary to run `strfile $ZSH/plugins/chucknorris/fortunes/chucknorris\n` to write the fortune data to the proper directory."


carlosala:
Thanks for your contribution!

PR #12823, opened 2024-11-25
fix(mercurial): use "hg identify --template", avoid "hg status" when possible

rdpate:
Use "hg identify" with a template parameter to generate revision identifiers (branch, bookmarks, tags) because this is faster, more complete, and more customizable.

When DISABLE_UNTRACKED_FILES_DIRTY=true, avoid "hg status" entirely; otherwise, stop once the first untracked file is found.  This should be much faster on really large repositories.

## Standards checklist:

### DIFF
--- a/plugins/mercurial/README.md
+++ b/plugins/mercurial/README.md
@@ -49,13 +49,25 @@ plugins=(... mercurial)
   PROMPT='${ret_status}%{$fg_bold[green]%}%p %{$fg[cyan]%}%c %{$fg_bold[blue]%}$(git_prompt_info)$(hg_prompt_info)%{$fg_bold[blue]%} % %{$reset_color%}'
   ```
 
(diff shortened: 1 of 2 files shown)


carlosala:
I'm not sure how to review this at all. I have never used mercurial in my whole life.
Does this change the visual behaviour of the mercurial prompt?
@mcornella, any ideas?

rdpate:
There are two purposes: (1) efficiency, being considerably faster to display dirty-or-clean, depending on repo size and modifications, with "hg identify" than storing the full "hg status" for grep, and (2) showing the current branch, bookmark, and tags -- blanking out each if not used, and customizable with the IDENTIFY variable.

> Does this change the visual behaviour of the mercurial prompt?

In addition to the second above, and after reviewing it fresh now: yes, a tiny visual change (unbolded magenta and bolded yellow).  That is customizable with the same variables as previously used.

PR #12837, opened 2024-12-06
fix(updater): allow non-tty usage when `update_mode=auto`

### LINKED ISSUE #12781: check_for_upgrade fails on MacOS terminal launch
### Describe the bug

Automatic upgrade fails to run on MacOS when launching a terminal -- both the native terminal and wezterm have been tried.

The offending test appears to be ```if [[ ! -t 1 ]]``` which for some reason evaluates to true when either terminal is first launched.

If ```source .oh-my-zsh/tools/check_for_upgrade.sh``` is invoked manually after launch, this test evaluates to false and the upgrade runs as expected.

### Steps to reproduce

Launch a terminal on MacOS.

### Expected behavior

Automatic updates should run when using MacOS terminals.

_No response_

### OS / Linux distribution

MacOS 15.0.1

### Zsh version

5.9

### Terminal emulator

Terminal.app, Wezterm

None

### Additional context

_No response_

Omoeba:
## Standards checklist:

## Changes:

-  Non-interactive use is permitted when `update_mode=auto`

## Other comments:

Fixes #12781

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -27,7 +27,7 @@ zstyle -s ':omz:update' mode update_mode || {
 # - $ZSH is not a git repository
 if [[ "$update_mode" = disabled ]] \
    || [[ ! -w "$ZSH" || ! -O "$ZSH" ]] \
-   || [[ ! -t 1 ]] \
+   || ([[ "$update_mode" != auto ]] && [[ ! -t 1 ]]) \
    || ! command git --version 2>&1 >/dev/null \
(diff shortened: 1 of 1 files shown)


Omoeba:
@mcornella can you check this out?

Omoeba:
@carlosala Can you take a look instead since Marc seems to be unavailable?

mcornella:
Please see https://github.com/ohmyzsh/ohmyzsh/issues/12781#issuecomment-2620189056. Thanks!

mcornella:
Superseeded by #12971.

PR #12839, opened 2024-12-07
Add list of aliases for magento2

AbdulrahmanMasoud:
## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/magento2/README.md
+++ b/plugins/magento2/README.md
@@ -1,0 +1,73 @@
+```markdown
+# Magento 2
+
+This plugin adds aliases for Magento2 commands.
+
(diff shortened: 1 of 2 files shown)


robbyrussell:

A newer magento2 plugin PR, #13107, covers the same ground with a fuller set, and I'd rather make the "should Oh My Zsh carry a magento2 plugin" call once, over there. Closing this one for that reason. If you'd like to weigh in on that PR, please do.

One thing worth fixing in your own config either way: several aliases here use `bin\magento`, and zsh reads that as `binmagento`.

PR #12840, opened 2024-12-10
Warn user if direnv not found

thethomasboyer:
## Standards checklist:

## Changes:

Instead of silently not installing the shell hooks, issue a warning (then keep doing nothing).

## Misc:

Follow-up of https://github.com/ohmyzsh/ohmyzsh/pull/8809#issuecomment-612344196 and https://github.com/ohmyzsh/ohmyzsh/issues/12798#issuecomment-2464749859.

### DIFF
--- a/plugins/direnv/direnv.plugin.zsh
+++ b/plugins/direnv/direnv.plugin.zsh
@@ -1,5 +1,8 @@
-# Don't continue if direnv is not found
-command -v direnv &>/dev/null || return
+# If direnv is not found, don't continue and print a warning
+if ! command -v direnv &>/dev/null; then
+  echo "Warning: direnv not found. Please install direnv and ensure it's in your PATH before using this plugin."
+  return
+fi
 
 _direnv_hook() {
   trap -- '' SIGINT;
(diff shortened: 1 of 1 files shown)


plugins/direnv/direnv.plugin.zsh
```diff
@@ -1,5 +1,8 @@
-# Don't continue if direnv is not found
-command -v direnv &>/dev/null || return
+# If direnv is not found, don't continue and print a warning
+if ! command -v direnv &>/dev/null; then
```
carlosala, line 2: idiomatic nit :)
```suggestion
if (( ! $+commands[direnv] )); then
```

carlosala (approved):
LGTM 🚀

carlosala:
Thanks for your contribution!

ajilty:
This seems like an anti-pattern. If all plug-ins did this, what a mess it would be.

carlosala:
> This seems like an anti-pattern. If all plug-ins did this, what a mess it would be.

Many of them do. Nevertheless, I agree that we could discuss removing all of those warnings.

Issue #12852, opened 2024-12-18
plugin: virtualenvwrapper doesn't cleanup properly after being disabled/reconfigured


mvk:
### Describe the bug

hello.
`virtualenvwrapper` plugin has this feature where it installs a zsh `chpwd` hook function named `workon_cwd`.
It has a configurable env variable for people like me: `DISABLE_VENV_CD`.
However if you define `export DISABLE_VENV_CD=1` and then source the plugin, it will leave previously defined function intact. :)

basically, when being sourced the plugin needs to check whether it needs the function and whether it is already present.

### Steps to reproduce

Assuming you have oh-my-zsh installed at default location.
 

## I. Enable the plugin `virtualenvwrapper` and see its operation:
```!zsh
$ omz plugin enable virtualenvwrapper
$ source ~/.zshrc
$ pushd "${PWD}"
$ cd /tmp
$ mkdir -p kuku
$ cd kuku
$ python3 -m venv .venv
$ cd ../kuku
(.venv) $ cd ../
$ popd
```
so far as expected.

## II. Disable cd handling:
```
$ export DISABLE_VENV_CD=1
$ source ~/.zshrc
$ cd /tmp/kuku
(.venv) $ 
```
i.e. the hook is still working :)

### Expected behavior

I expected at II last line I would not get any `chpw` hooks.

### Screenshots and recordings

don't forget to clean up:
```
rm -fr /tmp/kuku
```

### OS / Linux distribution

Mac OS 15.2

### Zsh version

5.9

### Terminal emulator

iTerm2

### If using WSL on Windows, which version of WSL

None

## workarounds: 
1. use `omz reload`
2. manually clean up

## FIX:
add a test for DISABLE_VENV_CD undef or 0 and ensure that the hook ` is removed from `chpwd_functions`:
```
if [[ "${DISABLE_VENV_CD}" -eq 1 ]]; then
  autoload -U add-zsh-hook
  add-zsh-hook -d chpwd workon_cwd
fi
```

I know this is a rare use case and not a critical one.

mcornella:
Hi @mvk, thanks for the report.

The expectation we have in terms of plugin settings is that they are only read on plugin load, except in very few plugins.

To properly load a plugin, we need to reload the zsh session. Sourcing the zshrc file is not the way to do it, as there are many side effects.

Therefore, to do what you want, you need to:

1. Set the variable setting in your .zshrc file (before OMZ is sourced), to make it persistent.
[...]

PR #12853, opened 2024-12-18
virtualenvwrapper plugin fix #112852

mvk:
when we're not in enabled mode and `DISABLE_VENV_CD=1`:
  - safely delete the chpwd hook `workon_cwd` if it were previously set up

## Standards checklist:

## Changes:

- see the diff :)

## Other comments:

- tested only manually

...


mvk:
closing since the corresponding issue #12852 is a non-issue

PR #12866, opened 2024-12-25
feat(uv): add uv='noglob uv' alias

bretello:
## Standards checklist:

## Changes:

Adds `alias uv="noglob pip"` for the uv plugin , [following the `pip` plugin](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/pip/pip.plugin.zsh#L85-L89). This is especially useful when installing modules with extras: e.g. `pip install module[extra]`

### DIFF
--- a/plugins/uv/uv.plugin.zsh
+++ b/plugins/uv/uv.plugin.zsh
@@ -3,6 +3,8 @@ if (( ! ${+commands[uv]} )); then
   return
 fi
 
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks, and sorry for the delay!

sebastian-correa:
Ahh, good call! Thanks for the contribution!

PR #12873, opened 2024-12-29
feat(clipboard): add simple osc52 support

zwyyy456:
## Standards checklist:

## Changes:

- add osc52 support for `lib/clipboard`, so when we copy content on server, there will be the same content on the clipboard

### DIFF
--- a/lib/clipboard.zsh
+++ b/lib/clipboard.zsh
@@ -84,6 +84,17 @@ function detect-clipboard() {
   elif [ -n "${TMUX:-}" ] && (( ${+commands[tmux]} )); then
     function clipcopy() { tmux load-buffer "${1:--}"; }
     function clippaste() { tmux save-buffer -; }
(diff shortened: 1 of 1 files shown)


juru1234:
There is actual OSC52 support for [pasting](https://github.com/milhnl/dotfiles/blob/9d694cc936345217b3fbcb4d91332ddfded62ddd/XDG_BIN_HOME/vis-clipboard#L22). But I'm not sure if we want to implement that because many terminal emulators have disabled it as default for security reasons.

Also, I would rather do nothing than print a text if pasting is triggered.

In my personal config I ended up with `printf "\033]52;c;%s\a" "$(head -c 838860 "${1:-/dev/stdin}" | base64 -w 0)"` which also ensures that the copied content isn't larger than 8MB which seems to be a limit for OSC52 (see [here](https://sw.kovidgoyal.net/kitty/changelog/)). 

If I ssh to a system is there always   `"$TERM" == "xterm"* || "$TERM" == "screen"* || "$TERM" == "tmux"*` set? Maybe we could get rid of the existing `else` and just use OSC52 if none of the other cases matched....

PR #12884, opened 2025-01-02
'resetlaunchpad' alias to reset Launchpad layout

zumbiepig:
## Standards checklist:

## Changes:

- Added a `resetlaunchpad` alias to reset Launchpad layout to the `macos` plugin
- Included the new method to reset Launchpad layout on MacOS 15.2+

## Added code:
```bash
# Reset Launchpad layout (defaults method does not work in MacOS 15.2+)
alias resetlaunchpad='defaults write com.apple.dock ResetLaunchPad -bool true && rm -rf /private/$(getconf DARWIN_USER_DIR)/com.apple.dock.launchpad && killall Dock'
```

### DIFF
--- a/plugins/macos/macos.plugin.zsh
+++ b/plugins/macos/macos.plugin.zsh
@@ -17,6 +17,9 @@ function ofd {
 alias showfiles="defaults write com.apple.finder AppleShowAllFiles -bool true && killall Finder"
 alias hidefiles="defaults write com.apple.finder AppleShowAllFiles -bool false && killall Finder"
 
+# Reset Launchpad layout (defaults method does not work in MacOS 15.2+)
+alias resetlaunchpad='defaults write com.apple.dock ResetLaunchPad -bool true && rm -rf /private/$(getconf DARWIN_USER_DIR)/com.apple.dock.launchpad && killall Dock'
+
 # Bluetooth restart
 function btrestart() {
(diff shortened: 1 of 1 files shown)


plugins/macos/macos.plugin.zsh
```diff
@@ -17,6 +17,13 @@ function ofd {
 alias showfiles="defaults write com.apple.finder AppleShowAllFiles -bool true && killall Finder"
 alias hidefiles="defaults write com.apple.finder AppleShowAllFiles -bool false && killall Finder"
 
+# Reset Launchpad layout (defaults method does not work in MacOS Sequoia and onward)
+if [[ "$(sw_vers --productVersion)" -ge 15.0 ]]; then
+  alias resetlaunchpad='rm -rf /private/$(getconf DARWIN_USER_DIR)/com.apple.dock.launchpad && killall Dock'
```
mcornella, line 22: This rm command haunts me, what would happen on broken `getconf` output?

I think the whole thing should be a function with proper checks and validations of expected output.

mcornella: Whole thing meaning the 2 aliases converted into a single function.

zumbiepig:
ok, sure
--
Sincerely,
Orion Leon

plugins/macos/macos.plugin.zsh
zumbiepig: fixed!

PR #12891, opened 2025-01-06
feat(rbenv): provide completion for `rbenv` command

ybiquitous:
## Standards checklist:

## Changes:

This PR adds the completion for the `rbenv` command, which is officially supported.
I believe the support could be pretty helpful for most `rbenv` users.

For example, when you type <kbd>TAB</kbd> during the input, then some sub commands are suggested:

```sh-session
$ rbenv r
rehash  -- Regenerate rbenv shims
root    -- Display the root directory where versions and shims are kept
```

Ref https://github.com/rbenv/rbenv/#shell-completions

## Other comments:

None.


robbyrussell (changes requested):
Thanks for this, and sorry it sat so long without a reply. You're right that `rbenv init - zsh` only wires up bash completion, so a git-clone install gets nothing on Tab. I'd like to land it, but two things need to change first.

1. Oh My Zsh runs `compinit` before it sources plugins, so adding a directory to `fpath` here doesn't register `_rbenv` for the `rbenv` command. I tested this against a stub install with Homebrew's copy of `_rbenv` excluded: after the PR's `fpath+=`, `$_comps[rbenv]` is still empty. I suspect your test worked because Homebrew already ships `_rbenv` in `site-functions`. The fix is to autoload and register it explicitly after extending fpath.

2. `RBENV_ROOT=''` overwrites the value for anyone who has exported their own `RBENV_ROOT`, and rbenv then silently falls back to `~/.rbenv`. Please use a differently named variable.

While you're in there, deriving the directory from the binary rather than from the search loop also covers people who put `~/.rbenv/bin` on PATH themselves. Something like:

```zsh
rbenv_completions="${commands[rbenv]:A:h:h}/completions"
if [[ -f "$rbenv_completions/_rbenv" ]]; then
  fpath+=("$rbenv_completions")
  autoload -Uz _rbenv
  compdef _rbenv rbenv
fi
```

(and add `rbenv_completions` to the `unset` line at the bottom). I tested that block and it registers the completion in both the clone and Homebrew cases. Happy to push it onto your branch if you'd rather.

ybiquitous:
Hi @robbyrussell, thank you for the suggestion! I agree with your idea, so can you look at 029d3ea I pushed?

PR #12899, opened 2025-01-10
Add alias for git stash untracked with message

sathishmanohar:
## Standards checklist:

## Changes:

- It is often helpful to have git stashes with messages. Currently there isn't an alias for git stash with untracked files and a message. The `gstum` alias adds that.

## Other comments:
Usage:

`gstum "Descriptive Commit Name"` or `gstum Descriptive-Commit-Name`
...


carlosala:
Hi! I don't see a massive improvement here. `gtsum "hello world"` is almost the same keystrokes than `gstu -m "hello world"`. I wouldn't pollute with a new alias all user's env. Nevertheless, it seems like a great alias for anyone that uses it often, I think it is best suited for a personal zshrc config.
Thanks for your contribution!

PR #12908, opened 2025-01-14
feat(jira-auto-worklog): Add jira-auto-worklog plugin

DylanEtris:
## Standards checklist:

## Changes:

- Adds a plugin for automatically logging time to a JIRA instance


robbyrussell:
@DylanEtris This is pretty cool! Having said that, I think this is probably a good example of a custom plugin that might be worth promoting to people elsewhere vs being baked into Oh My Zsh, itself. 

We'd need to consider the requirements that Ruby is necessary on the machine for this to work, too. Am going to pass on this for now -- but once you share this somewhere, please add a comment so people know where to find it!

PR #12935, opened 2025-01-23
fix(cli)!: remove harmful `--unattended` flag for `omz update`

### LINKED ISSUE #9808: Remove --unattended flag from `omz update`
> Are there any drawbacks to using  instead of  (e.g.,  is not updated)?

There are no drawbacks as far as I'm aware, it is generally preferable to use the script directly. [I've documented this in the FAQ](https://github.com/ohmyzsh/ohmyzsh/wiki/FAQ#how-do-i-manually-update-oh-my-zsh-from-a-script).

I'm reconsidering removing the flag though. What I don't like is that you need to run the full  session in order to call , perhaps it could be reconverted into a binary.

_Originally posted by @mcornella in https://github.com/ohmyzsh/ohmyzsh/issues/9187#issuecomment-754190223_

mcornella:
## Standards checklist:

## Changes:

- Removes the `omz update --unattended` flag, which should not be used.
- Adds the documentation to the README.
- Shows error if used.
- Shows breaking change notice.

Fixes #9808
Related: https://github.com/ohmyzsh/ohmyzsh/issues/12168#issuecomment-2313224758

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,7 +488,7 @@ omz update
 ```
 
 > [!NOTE]
-> If you want to automate this process in a script, you should call directly the `update` script, like this:
+> If you want to automate this process in a script, you should call directly the `upgrade` script, like this:
 >
 > ```sh
 > $ZSH/tools/upgrade.sh
--- a/README.md
+++ b/README.md
@@ -487,6 +487,17 @@ wait a week?) you just need to run:
 omz update
 ```
 
+> [!NOTE]
+> If you want to automate this process in a script, you should call directly the `update` script, like this:
+>
+> ```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -487,6 +487,17 @@ wait a week?) you just need to run:
 omz update
 ```
 
+> [!NOTE]
+> If you want to automate this process in a script, you should call directly the `update` script, like this:
```
carlosala, line 491:
```suggestion
> If you want to automate this process in a script, you should call directly the `upgrade` script, like this:
```

carlosala (approved):
LGTM! 🚀

PR #12940, opened 2025-01-26
Add gdss alias to git plugin

armoucar-neon:
## Standards checklist:

## Changes:

- add `gdss` for checking quick stats of the current branch against the main branch


robbyrussell:

I'm going to pass on it. New aliases in the git plugin need a case that a lot of people run this often, and a diff summary against the main branch is more of a personal shortcut than a shared one. It fits well in your own `~/.zshrc`.

If you keep it there, use single quotes: with double quotes, `$(git_current_branch)` is expanded once when the alias is defined, so it sticks to whichever branch you were on when the shell started.

PR #12941, opened 2025-01-27
fix(chruby): update brew sourcing for Apple Silicon

dskecse:
## Standards checklist:

## Changes:

- Fixes (and hopefully speeds up) the `chruby` plugin loading process on Apple Silicon macs where the default Homebrew prefix is `/opt/homebrew`

cc @mcornella @carlosala

### DIFF
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -16,14 +16,15 @@ _source-from-omz-settings() {
 }
 
 _source-from-homebrew() {
-  (( $+commands[brew] )) || return 1
-
   local _brew_prefix
   # check default brew prefix
-  if [[ -h /usr/local/opt/chruby ]];then
-    _brew_prefix="/usr/local/opt/chruby"
+  if [[ -h /opt/homebrew/opt/chruby ]]; then
+    _brew_prefix="/opt/homebrew/opt/chruby"
   else
     # ok , it is not default prefix
+    # check to see if brew command is available
+    (( $+commands[brew] )) || return 1
+
     # this call to brew is expensive ( about 400 ms ), so at least let's make it only once
     _brew_prefix=$(brew --prefix chruby)
   fi
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -16,15 +16,16 @@ _source-from-omz-settings() {
 }
 
 _source-from-homebrew() {
+  (( $+commands[brew] )) || return 1
+
   local _brew_prefix
   # check default brew prefix
-  if [[ -h /opt/homebrew/opt/chruby ]]; then
+  if [[ -h /usr/local/opt/chruby ]];then
+    _brew_prefix="/usr/local/opt/chruby"
+  elif [[ -h /opt/homebrew/opt/chruby ]]; then
     _brew_prefix="/opt/homebrew/opt/chruby"
   else
     # ok , it is not default prefix
-    # check to see if brew command is available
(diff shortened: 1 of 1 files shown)


plugins/chruby/chruby.plugin.zsh
```diff
@@ -16,14 +16,15 @@ _source-from-omz-settings() {
 }
 
 _source-from-homebrew() {
-  (( $+commands[brew] )) || return 1
```
dskecse, line 19: On Apple Silicon macs `brew` is now in `/opt/homebrew/bin` instead of `/usr/local/bin` and it's not in the `PATH` by default, so this command fails and `command not found: chruby` is shown, while `chruby` could be loaded from the new default brew prefix. And it makes more sense to check to see if the `brew` command is available only before shelling out to the command.

plugins/chruby/chruby.plugin.zsh
```diff
@@ -16,14 +16,15 @@ _source-from-omz-settings() {
...
-
   local _brew_prefix
   # check default brew prefix
-  if [[ -h /usr/local/opt/chruby ]];then
```
dskecse, line 23: This will continue to work on older Intel-based macs, as this Homebrew prefix is covered by checking the `/usr/local/share/chruby/chruby.sh` path [below](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/chruby/chruby.plugin.zsh#L49), and the following symlinks exist on those macs (I've double-checked it on mine):

```
/usr/local/share/chruby -> /usr/local/Cellar/chruby/0.3.9/share/chruby
/usr/local/opt/chruby -> /usr/local/Cellar/chruby/0.3.9
```

plugins/chruby/chruby.plugin.zsh
```diff
@@ -16,14 +16,15 @@ _source-from-omz-settings() {
...
-    _brew_prefix="/usr/local/opt/chruby"
+  if [[ -h /opt/homebrew/opt/chruby ]]; then
```
dskecse, line 21: This references the [new default Homebrew prefix](https://docs.brew.sh/Installation) on Apple Silicon, which is `/opt/homebrew`

dskecse:
Hey @mcornella @carlosala, I'd appreciate if you could have a look at this fix for the `chruby` plugin, thank you!

carlosala:
Hey! I think you should ensure that `brew` is in your `$PATH` before sourcing this plugin. You can do it by using `brew` plugin (before this one, of course) or by changing the path manually before sourcing `oh-my-zsh.sh`.

carlosala:
I changed it to only add the default apple silicon. Ensure you have brew on the path before hitting this plugin.
Thanks!

dskecse:
Hey @carlosala, first of all thanks for taking a look!
When it comes to ensuring that `brew` is in the `$PATH` before loading the plugin: I could, but omz and this plugin just used to work on my older Intel-based mac with the same dotfiles, and it stopped working on my newer Apple Silicon mac.
Don't you think we should keep the forward compatibility?

carlosala:
There is a `brew` plugin that ensures that.
You can add it the first in the list to have it ready!

dskecse:
@carlosala I know I can, but I wouldn't like to add yet another plugin just for that, as the more plugins are added the more it slows down the shell startup, so my point was that, perhaps, we could focus a bit more on the forward compatibility and performance, wdyt?

carlosala:
If you want to avoid adding brew plugin, add 
```zsh
path=(/opt/homebrew/bin $path)
```
before `source oh-my-zsh.sh`.

PR #12957, opened 2025-02-03
fix(mvn): Handle directories in <relativePath>

mensinda:
## Standards checklist:

- [ ] ~If the code introduces new aliases, I provide a valid use case for all plugin users down below.~

## Changes:

Maven also supports putting a directory into `<relativePath>`, however, the current version of the plugin only supports references to files and produce the following error with directories:

```
sed: read error on XXXXXXXX: Is a directory
```

This PR fixes that.

## Other comments:

Not sure who maintains the mvn plugin so I guess cc @mcornella from the git history?

### DIFF
--- a/plugins/mvn/mvn.plugin.zsh
+++ b/plugins/mvn/mvn.plugin.zsh
@@ -101,8 +101,14 @@ function listMavenCompletions {
       new_file="../pom.xml"
     fi
 
-    # if file doesn't exist break
     file="${file:h}/${new_file}"
+
+    # if the file points to a directory, assume it is a pom.xml in that directory
+    if [[ -d "$file" ]]; then
+      file="${file}/pom.xml"
+    fi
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @mensinda, the patch looks good, but I've tried to look for supporting docs and only found this that says it might not be supported: https://youtrack.jetbrains.com/issue/IDEA-318500/Maven-parent-relative-path-without-explicit-pointing-to-pom.xml

Can you post docs or code that shows a directory is valid?

mensinda:
I don't have docs, but I have source code:

https://github.com/apache/maven/blob/maven-3.0.x/maven-model-builder/src/main/java/org/apache/maven/model/building/DefaultModelBuilder.java#L761-L787

mcornella (approved):
LGTM

PR #12975, opened 2025-02-14
feat(cli): add `plugin cat` command

emiliadecaudin:
## Standards checklist:

## Changes:

- This pull request adds a `plugin cat` command which shows the source of a particular plugin, just like `plugin info` shows the README of a particular plugin.

## Other comments:

I thought of this change after often finding myself wanting to check to see what a particular plugin actually does, esp. in the context of breaking changes on the part of a plugin's subject (most recently, asdf 0.16.0 breaking the omz asdf plugin).

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -41,6 +41,7 @@ function _omz {
         refs=("${(@f)$(builtin cd -q "$ZSH"; command git for-each-ref --format="%(refname:short):%(subject)" refs/heads refs/tags)}")
         _describe 'command' refs ;;
       plugin) subcmds=(
+        'cat:Show plugin source'
         'disable:Disable plugin(s)'
         'enable:Enable plugin(s)'
         'info:Get plugin information'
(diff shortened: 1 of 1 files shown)


Maliksb11:
Yes

mcornella:
This is a really good idea!

What do you think about renaming the command to `view` and having a `-w` argument opening the plugin folder in GitHub, as other CLIs do? 

Also, we have to take into consideration plugins with multiple files, I'm not sure what we'd do then.

emiliadecaudin:
I can definitely switch to `view` from `cat` if that's what you would prefer.

RE the `-w` argument; do you have a recommendation on how to programmatically/consistently refer to the plugin's folder on Github?

Lastly, can you send an example of a plugin that has multiple files?

Thank you!

emiliadecaudin:
Sorry for the delay here. @mcornella I added code to display `_{plugin}` if `{plugin}.plugin.zsh` doesn't exist, but I'm not sure what the best way to handle multiple files would be. Open to suggestions if you feel it should be addressed before merging.

PR #12980, opened 2025-02-16
fix(chruby): fix the `rubies` command output

dskecse:
## Standards checklist:

## Changes:

- Removes redundant `_load-chruby-dirs` function
- Extracts a `_source-from-default-location` function
- Fixes the output of `chruby` and its `rubies` alias

## Context

After updating oh-my-zsh to the latest version, the `chruby` command (and its `rubies` alias) now adds `.rubies` to the output:

```sh
~ λ rubies
 * ruby-3.4.1
   .rubies
```

It used to only display the actual Ruby versions installed (in my case in the default `chruby` `$HOME/.rubies` directory):

```sh
~ λ rubies
 * ruby-3.4.1
```

It was revealed using `git bisect` that the output was broken by [this commit](https://github.com/ohmyzsh/ohmyzsh/commit/5c14474eb2a252ad61e55cf084c5bbe6c1c934b9).
I also noticed that `chruby`'s `share/chruby/chruby.sh` [script](https://github.com/postmodern/chruby/blob/master/share/chruby/chruby.sh), which is `source`d by the plugin ([1](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/chruby/chruby.plugin.zsh#L10), [2](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/chruby/chruby.plugin.zsh#L35), [3](https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/chruby/chruby.plugin.zsh#L52)), does [the same thing](https://github.com/postmodern/chruby/blob/master/share/chruby/chruby.sh#L5) of adding directories with Ruby versions installed to the `RUBIES` array:

```sh
for dir in "$PREFIX/opt/rubies" "$HOME/.rubies"; do
  [[ -d "$dir" && -n "$(command ls -A "$dir")" ]] && RUBIES+=("$dir"/*)
done
```

so this plugin's `_load-chruby-dirs` function is essentially doing the same thing the `chruby.sh` script already does for us, but the function is doing it wrong, i.e. `RUBIES+=("$dir")` instead of `RUBIES+=("$dir"/*)`.

This PR fixes the process of loading `chruby` dirs and fixes the output of `chruby` (and its `rubies` alias):

```sh
~ λ rubies
 * ruby-3.4.1
```

### DIFF
--- a/plugins/chruby/chruby.plugin.zsh
+++ b/plugins/chruby/chruby.plugin.zsh
@@ -15,6 +15,13 @@ _source-from-omz-settings() {
   fi
 }
 
+_source-from-default-location() {
+  [[ -r /usr/local/share/chruby/chruby.sh ]] || return 1
+
+  source /usr/local/share/chruby/chruby.sh
+  source /usr/local/share/chruby/auto.sh
+}
+
 _source-from-homebrew() {
   (( $+commands[brew] )) || return 1
 
@@ -36,27 +43,14 @@ _source-from-homebrew() {
   source $_brew_prefix/share/chruby/auto.sh
 }
 
-_load-chruby-dirs() {
-  local dir
-  for dir in "$HOME/.rubies" "$PREFIX/opt/rubies"; do
-    if [[ -d "$dir" ]]; then
-      RUBIES+=("$dir")
-    fi
-  done
-}
-
 # Load chruby
-if _source-from-omz-settings; then
-  _load-chruby-dirs
-elif [[ -r "/usr/local/share/chruby/chruby.sh" ]] ; then
-  source /usr/local/share/chruby/chruby.sh
-  source /usr/local/share/chruby/auto.sh
-  _load-chruby-dirs
-elif _source-from-homebrew; then
-  _load-chruby-dirs
-fi
+_source-from-omz-settings || \
+  _source-from-default-location || \
+  _source-from-homebrew
 
-unfunction _source-from-homebrew _source-from-omz-settings _load-chruby-dirs
+unfunction _source-from-homebrew \
+           _source-from-default-location \
+           _source-from-omz-settings
 
 
 ## chruby utility functions and aliases


dskecse:
Hey @mcornella @carlosala, I would appreciate you taking a look at this one, thanks!

plugins/chruby/chruby.plugin.zsh
```diff
@@ -36,27 +43,14 @@ _source-from-homebrew() {
   source $_brew_prefix/share/chruby/auto.sh
 }
 
-_load-chruby-dirs() {
```
dskecse, line 39: FYI: This plugin's `_load-chruby-dirs` function is essentially doing the same thing `chruby`'s `share/chruby/chruby.sh` script [already does](https://github.com/postmodern/chruby/blob/a543a35790e5528b5a67de20e78a7390f5f7606e/share/chruby/chruby.sh#L5) for us, but the function is doing it wrong, i.e. `RUBIES+=("$dir")` instead of `RUBIES+=("$dir"/*)`, and that's the reason the output of `chruby` is broken, hence the removal of the function.

plugins/chruby/chruby.plugin.zsh
```diff
@@ -15,6 +15,13 @@ _source-from-omz-settings() {
   fi
 }
 
+_source-from-default-location() {
+  [[ -r /usr/local/share/chruby/chruby.sh ]] || return 1
+
+  source /usr/local/share/chruby/chruby.sh
+  source /usr/local/share/chruby/auto.sh
+}
```
dskecse, line 23: Extracted from lines 51 through 54

dskecse:
Hey all, could the fix be prioritized please?

carlosala (approved):
LGTM! 🚀

carlosala:
Good catch! Thanks for your contribution!

PR #12983, opened 2025-02-19
fix(agnoster): print white text over black for light theme only

tuzi3040:
## Standards checklist:

## Changes:

- Print white text over black only when `SOLARIZED_THEME` is set to `light`

## Other comments:

Changes introduced by #12525 changes `AGNOSTER_CONTEXT_FG` and `AGNOSTER_STATUS_FG` text color from default to white under default configuration, which makes a difference when terminal default text color is not white.
This should be achieved by setting `SOLARIZED_THEME` to `light` or modifying `AGNOSTER_CONTEXT_FG` and `AGNOSTER_STATUS_FG` manually in .zshrc.
cc @Kekun @carlosala

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -35,8 +35,14 @@
 CURRENT_BG='NONE'
 
 case ${SOLARIZED_THEME:-dark} in
-    light) CURRENT_FG='white';;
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM! 🚀 
Thanks!

PR #12990, opened 2025-02-23
feat(git): add alias `gcsm` for `git commit --gpg-sign --message`

roeniss:
## Standards checklist:

## Changes:

- Add new `git` alias `gcsm`, which stands for `git commit --gpg-sign --message`

## Background:

- I do gpg-sign regularly but not sign-off. Because my normal projects don't need [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin). What's more, it makes the commit a bit more verbose than needed in GitHub UI: 

Screenshot 2025-02-23 at 17 05 15

As we can see above, the sign-off itself creates new lines in the commit, so when the user clicks the three-dots button, they see nothing but the sign-off, which seems meaningless in normal projects.

## Other comments:

- I had a little concern about whether `-s` should be `--sign-off` or `--gpg-sign` because `ss` is `--gpg-sign --signoff` currently. But since there's `alias gcs='git commit --gpg-sign'` in [the code](https://github.com/ohmyzsh/ohmyzsh/blob/90cd0d16161382a51f41dee3d281e4c682f06468/plugins/git/git.plugin.zsh#L188) already, I'll go with it.
- I didn't split my commit into `feat` and `docs` for simplicity. If the maintainer insists, I'll split it happily.


roeniss:
@carlosala Hi. Can you review my code?

roeniss:
I didn't notice there's `gcsm` already. I'll close this PR for now.

Issue #12986, opened 2025-02-21
Add git commit fixup alias in Git plugin


Thibuko:
### If the feature request is for a plugin or theme, specify it here.

Git plugin

### If the feature solves a problem you have, specify it here.

fixup feature of [git commit](https://git-scm.com/docs/git-commit/2.32.0#Documentation/git-commit.txt---fixupamendrewordltcommitgt) allows to easily organize correction of existing commits during reviews / validation

### Describe the proposed feature.

Would be nice to have an alias for it, something like `gcf` that will take a commit id as parameter and launch the command `git commit --fixup=<commit>`

roeniss:
**tl;dr: Can we change the existing `gcl`?** (git config --list)

---

Looks like `gcf` is already taken.

https://github.com/ohmyzsh/ohmyzsh/blob/90cd0d16161382a51f41dee3d281e4c682f06468/plugins/git/git.plugin.zsh#L203

... which means that we need to introduce a new alias such as `gcfx` or `gcfu`.

However, I anticipate that`git-commit` will be more widely and frequently used in most situations.

(please note that `git-commit` belongs to the porcelain command while `git-config` doesn't. (https://github.com/git/git/blob/master/command-list.txt#L82-L85) This is not critical rationale though)

**So I want to suggest replacing `gcl` with `gcls` and using `gcf` for `git config --fixup`.** 

Unfortunately, I am unaware of the correct person to speak to about this matter.

roeniss:
I made two PRs that will be needed if my suggestion is accepted: #12991 #12992

Thibuko:
Thanks guys, love it.

roeniss:
I'm also happy. Since this PR I use a lot of fixup lol

PR #12992, opened 2025-02-23
feat(git)!: add alias `gcf` for `git commit --fixup` (#12986)

roeniss:
## Standards checklist:

## Changes:

- Related to https://github.com/ohmyzsh/ohmyzsh/issues/12986
- BREAKING CHANGE: Existing `gcf` becomes alias for `git config --fixup`
- This PR is intended to be merge after https://github.com/ohmyzsh/ohmyzsh/pull/12991, which replaces the original `gcf` to other alias

## Other comments:

- This PR is not discussed yet.


Pascalmh:
@roeniss what do you think of using `gcfu` which is still available and wouldn't cause a breaking change and doesn't have any dependencies on other changes?

roeniss:
@Pascalmh That's a fair point. I'll close this and another PR (https://github.com/ohmyzsh/ohmyzsh/pull/12991) then make a new one.

PR #12998, opened 2025-02-28
feat(ssh): add ssh_fingerprint function and ssh_fp alias

pepoluan:
Adds a new function `ssh_fingerprint` (and its alias `ssh_fp`) to the `ssh` plugin.

`README.md` has also been updated to reflect the new function.

## Standards checklist:

## Changes:

- New function: `ssh_fingerprint`
- New alias: `ssh_fp`

### DIFF
--- a/plugins/ssh/README.md
+++ b/plugins/ssh/README.md
@@ -14,3 +14,10 @@ plugins=(... ssh)
 - `ssh_rmhkey`: remove host key from known hosts based on a host section name from `.ssh/config`.
 - `ssh_load_key`: load SSH key into agent.
 - `ssh_unload_key`: remove SSH key from agent.
+- `ssh_fingerprint` : calculate fingerprint of specifed key files. It has some options:
+  - `-md5` : Use MD5 fingerprint instead of SHA256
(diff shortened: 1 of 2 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry it sat unanswered for so long. A fingerprint helper fits next to the other key functions in this plugin, so I'd like to take it, with a few changes:

1. `if ! ssh-keygen ...; then return $?; fi` returns 0, because `$?` inside the `then` branch is the result of the negated test, not of ssh-keygen. A missing key file currently reports success. Capture the status before the `if`, or use `ssh-keygen ... || return $?`.

2. `printf "${ansi[1]}$f:${ansi[2]}\n"` uses the filename as the format string, so a path with `%` in it breaks. Use `printf '%s%s:%s\n' "${ansi[1]}" "$f" "${ansi[2]}"`.

3. Please drop the `ssh_fp` alias and the `{` to `() {` changes on the two existing functions. New aliases need a case that a lot of people would use them, and the style edits are unrelated churn that makes the diff harder to read.

If you'd rather not pick this back up, that's fine, just say so and I'll close it out.

PR #13008, opened 2025-03-07
fix(systemadmin): don't hardcode http20 network interface

mjgardner:
## Standards checklist:

## Changes:

- Fix: Dynamically determine appropriate network interface for the `http20` function rather than hardcoding it to `eth0`, which is nonexistent on macOS.
- Style: One line per pipeline command in the `http20` function per ohmyzsh [Code Style Guide](https://github.com/ohmyzsh/ohmyzsh/wiki/Code-Style-Guide)

## Other comments:

This change uses the `ip route` command which is commonly installed on Linux and provided by the [iproute2](https://wiki.linuxfoundation.org/networking/iproute2) project. It checks what interface would be used to route to Cloudflare’s 1.1.1.1 DNS service, which although also hardcoded, at least is common across the Internet.

The macOS port of `ip route` is available from [Homebrew](https://brew.sh) by running `brew install iproute2mac`. Although an external dependency, it does at last make the function work on macOS with a command also common on Linux.

For example, on my macOS system, it returns either `en0` (Wi-Fi) or `en7` (the Ethernet slot on my Thunderbolt dock).

### DIFF
--- a/plugins/systemadmin/systemadmin.plugin.zsh
+++ b/plugins/systemadmin/systemadmin.plugin.zsh
@@ -94,7 +94,13 @@ function req20() {
 
 # top20 of Using tcpdump port 80 access to view
 function http20() {
-  sudo tcpdump -i eth0 -tnn dst port 80 -c 1000 | awk -F"." '{print $1"."$2"."$3"."$4}' | sort | uniq -c | sort -nr | head -n 20
+  sudo tcpdump -i $(ip route get 1.1.1.1 \
+    | awk '/^1.1.1.1 via / {print $5}') -tnn dst port 80 -c 1000 \
+    | awk -F"." '{print $1"."$2"."$3"."$4}' \
+    | sort \
+    | uniq -c \
+    | sort -nr \
+    | head -n 20
 }
 
 # top20 of Find time_wait connection


roeniss:
tl;dr: We don't have any workaround for zero additional package?

---

I'm macOS user and I found that my macbook uses en1 instead of en0, as you said. Your patch makes sense to some extent.

But is it the only solution for us to require another package? As a normal omz user, I'm a bit worried about this solution might demand excessive to some users, including me. It would be a little frustrated if I enable **`systemadmin`** then computer said **"ip:2: command not found: ip"**.

mjgardner:
> tl;dr: We don't have any workaround for zero additional package?

It's an improvement over the previous situation which didn't work at all on macOS. I can either work on an additional PR that uses `route get` on macOS, or if you prefer I'll revise this one.

I thought it best to get the simplest portable solution out first so as not to get bogged down over the mechanism to detect what platform we’re running on.

roeniss:
I agree with you, but still imagine what it would be like to inform a user that they need `iproute2mac`. The package name is too hard to reason about! 😂 

reference: 
https://github.com/ohmyzsh/ohmyzsh/blob/1c2127727af0ac452292f844ee32306f12906d03/plugins/kubectl/kubectl.plugin.zsh#L1-L3

I'm not a maintainer so please take my word as a small opinion.

mjgardner:
> I agree with you, but still imagine what it would be like to inform a user that they need `iproute2mac`. The package name is too hard to reason about! 😂

OK, you have convinced me and pointed to a nice example of failing more gracefully. Stay tuned for a revised PR.

carlosala:
Hey! Closing for the moment, feel free to open a new one if you want!

roeniss:
@mjgardner If you want, I can make PR instead of you

PR #13030, opened 2025-03-23
fix(alias-finder): Return early on very short command (1-length)

roeniss:
…ble part## Standards checklist:

## Changes:

- Fix bug in alias-finder: If the input command's length is 1 or less, return early instead of making an invalid regex format.

## Other comments:

- fixes https://github.com/ohmyzsh/ohmyzsh/issues/13016


carlosala (approved):
LGTM! 🚀

PR #13035, opened 2025-03-26
Fix async status registration by declaring it prior to registration

mrmikeface:
## Standards checklist:

## Changes:

- Move async conditional block to bottom of file to avoid undefined functions

## Other comments:

> If anyone finds that its git prompt does not appear they can set before `source oh-my-zsh.sh` the following snippet:
>
> ```sh
> zstyle ':omz:alpha:lib:git' async-prompt force
> ```
>
> This was introduced in `2a109d30afe4ab164a946c307abc3d2a444a42ad`

 _Originally posted by @carlosala in [#12328](https://github.com/ohmyzsh/ohmyzsh/issues/12328#issuecomment-2400576507)_

This fix does not work because `git_prompt_status` is defined after `_omz_register_handler` is called thereby skipping registration. The simplest fix seemed like moving the conditional to bottom but I am open to alternatives

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -39,73 +39,6 @@ function _omz_git_prompt_info() {
   echo "${ZSH_THEME_GIT_PROMPT_PREFIX}${ref:gs/%/%%}${upstream:gs/%/%%}$(parse_git_dirty)${ZSH_THEME_GIT_PROMPT_SUFFIX}"
 }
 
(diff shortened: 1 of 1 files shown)


mrmikeface:
ffs the conflict is caused by this slow review process

carlosala:
Actually, this was fixed in #13134. Thanks for your contribution, I'm sorry this PR fell deep in the backlog.

PR #13042, opened 2025-03-29
fix(cli): support `noexec` environments

### LINKED ISSUE #13032: Support $ZSH on a noexec-mounted partition

_No response_

If the OMZ's main directory is on a partition with noexec set, `omz update` partially fails - I think the main update part passes, it only fails to print out the changelog.

I'm hitting this specifically on a TrueNAS box, as since upgrading to 24.10.1+ the noexec flag was fixed in ZFS - ref discussion on truenas forums: https://forums.truenas.com/t/shell-script-permission-denied-with-24-10-1/27941

Calling all sub-scripts via a `zsh <script>` instead of directly should solve the issue.

In particular changing the `$ZSH/tools/changelog.sh` calls in `tools/upgrade.sh:257` and `lib/cli.zsh:196` should mitigate the observed `omz update` partially failing.

I've additionally grepped the OMZ directory and found no other suspicious direct script calls, but my grep-fu might be weak here.

_No response_

### Additional context

_No response_

### Related Issues

Similar issue was described here:
* https://github.com/ohmyzsh/ohmyzsh/issues/6973
* https://github.com/ohmyzsh/ohmyzsh/issues/1245

carlosala:
Closes #13032

## Standards checklist:

## Changes:

Ensure that `omz update` can be called in `noexec` environments.
Closes #13032

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -193,7 +193,7 @@ EOF
     return 1
   fi
 
-  "$ZSH/tools/changelog.sh" "$version" "${2:-}" "$format"
+  ZSH="$ZSH" command zsh -f "$ZSH/tools/changelog.sh" "$version" "${2:-}" "$format"
 }
 
 function _omz::plugin {
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -254,7 +254,7 @@ if LANG= git pull --quiet --rebase $remote $branch; then
 
     # Print changelog to the terminal
     if [[ $interactive == true && $verbose_mode == default ]]; then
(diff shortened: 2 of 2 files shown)


mcornella (approved):
Essentially does the same thing, so this won't have any side effects. I think we should do the same in https://github.com/ohmyzsh/ohmyzsh/blob/a78d006faa022a5104b542d7dec5a3439e72046e/tools/upgrade.sh#L257

LGTM!

carlosala:
Yep, it is done as well in the same commit 👍🏻

PR #13045, opened 2025-03-31
chore(asdf): Add warning for incompatible version users

roeniss:
## Standards checklist:

## Changes:

- Ease the surprise for asdf older version users — show warning before completion registration finished.

## Other comments:

- `asdf --version` shows quite different results between 0.15.0 and 0.16.0, `v0.15.0-31e8c93` and `asdf version 0.16.5` —  so I thought this would be adequate implementation.
- I'm not sure whether I can use `[omz]` prefix for warning echo message.


plugins/asdf/asdf.plugin.zsh
```diff
@@ -1,5 +1,10 @@
 (( ! $+commands[asdf] )) && return
 
+# TODO(roeniss): 2025-03-31 - remove this warning and related README section after enough grace period.
+if [[ -d "$HOME/.asdf" ]]; then
```
bsouty-imb, line 4: Unfortunately, `$HOME/.asdf` still exists with v0.16+ (c.f. https://asdf-vm.com/guide/upgrading-to-v0-16.html#_2-set-asdf-data-dir) so it's not a reliable way of checking.  Maybe you could parse the output of `asdf version` to determine if it's older than v0.16

plugins/asdf/README.md
```diff
@@ -34,6 +34,11 @@ asdf set nodejs latest
 asdf set -u nodejs 16.5.0
 ```
 
+## Backward Compatibility
+
+`asdf` has significantly changed since 0.16.0. The current implementation doesn't along with older versions. **If you're using 0.15.0 or older version, please follow upgrade guide in official page**: https://asdf-vm.com/guide/upgrading-to-v0-16.html.
```
bsouty-imb, line 39: > `asdf` has significantly changed since 0.16.0. The current implementation doesn't along with older versions.

You're missing a word there

plugins/asdf/asdf.plugin.zsh
roeniss: thank you for the advice. 6ed551d

plugins/asdf/README.md
roeniss: Did I understand you well? 12a8e48

PR #13047, opened 2025-04-01
feat(fancy-ctrl-z): Flip-flop between the last two suspended jobs

whisperity:
Added a feature to `fancy-ctrl-z` which, if and only if there are precisely two backgrounded jobs, switches between them (using `fg %-`) when bringing back something to the foreground.

## Standards checklist:

## Changes:

- Flip-flop between the most recent two background, suspended jobs

## Other comments:

Multiple times I've happened to find myself needing to quickly change between the two most recent background jobs, such as two `vim`s. The most recent use case is one editor for the work and another for the input of the work (which is another project), but in a prior job, I had similar situation needing one editor for the back-end (implemented in Python or C++) and one for the front-end (implemented in Vue or Angular) of a product. I often need two separate editors because, due to some of the Vim plugins being dependent on the working directory I can't have one editor run both projects. But similar situation can be envisioned, e.g., by wanting to check a process monitor (`htop`) after starting a compilation.

Previously, I used `fg %1` and `fg %2` for this, always meticulously reminding myself of the job numbers, which is clunky. Instead, `fg %-` does the trick automatically, as the `jobs` subsystem remembers which one was your most recent, and second most recent background jobs, e.g.,

```
❯ jobs
[1]    suspended  /opt/homebrew/bin/nvim
[2]  - suspended  htop
[3]  + suspended  /opt/homebrew/bin/nvim
[4]    suspended  man
```

The `+` and `-` are the two I'm talking about. So I added to the plugin to use the `%-` shorthand for the "other" job, essentially making it toggle between the two most recently backgrounded jobs in the session. 

My inspiration came from TV remotes which had a dedicated button to switch always to the previously watched channel. If you tuned to channel 52 and then to 17 with the numerical buttons, pressing this dedicated switch button let you go 52,17,52,17 in a loop.

### DIFF
--- a/plugins/fancy-ctrl-z/README.md
+++ b/plugins/fancy-ctrl-z/README.md
@@ -1,8 +1,8 @@
 # fancy-ctrl-z
 
-Allows pressing Ctrl-Z again to switch back to the most recently backgrounded
-job. If there are precisely two backgrounded jobs, Ctrl-Z will toggle between
-them, bringing the other one back.
+Allows pressing <kbd>Ctrl</kbd>-<kbd>Z</kbd> is an empty prompt to bring to the
+foreground the only suspended job, or, if there are more than one such jobs, to
+switch between the two most recently suspended ones.
 
 To use it, add `fancy-ctrl-z` to the `plugins` array in your `.zshrc` file:
 
@@ -12,15 +12,29 @@ plugins=(... fancy-ctrl-z)
 
 ## Motivation
 
-I frequently need to execute random commands in my shell. To achieve it I pause
-Vim by pressing Ctrl-Z, type command and press fg<Enter> to switch back to Vim.
-The fg part really hurts me. I just wanted to hit Ctrl-Z once again to get back
-to Vim. I could not find a solution, so I developed one on my own that
-works wonderfully with ZSH.
+I frequently need to execute random commands in my shell.
+To achieve it, I often pause Vim by pressing <kbd>Ctrl</kbd>-<kbd>Z</kbd>, type
+a command and then would use `fg`<kbd>↵ Enter</kbd> to switch back to Vim.
+Having to type in the `fg` part really hurt me.
+I just wanted to hit <kbd>Ctrl</kbd>-<kbd>Z</kbd> once again to get back to Vim.
+I could not find a solution, so I developed one on my own that works wonderfully
+with Zsh.
 
-Source: http://sheerun.net/2014/03/21/how-to-boost-your-vim-productivity/
+Switching between the last two suspended jobs is motivated by both TV remotes
+that had such feature, and tools like `cd -` and `git checkout -` that switch
+between the current and the second most recent state (directory, branch, etc.).
+Sometimes, you have your Vim where code is changed, and another longer-running
+process (e.g., a `tail` on some logs, or a Python interpreter) where you want
+to test or observe your changes.
+There is no point in time where you would "have the editor open" and "have the
+program open" together, and the workflow clearly mandates always switching
+back and forth between the two.
+That's why the original version of _fancy-ctrl-z_ was extended with this "even
+fancier" behaviour, because the original version would've opened Vim back again
+and again.
 
-Credits:
-- original idea by [@sheerun](https://github.com/sheerun)
-- added to OMZ by [@mbologna](https://github.com/mbologna)
-- two-job switching added by [@Whisperity](https://github.com/Whisperity)
+## Credits
+
+- Original idea by [@sheerun](https://github.com/sheerun), http://sheerun.net/2014/03/21/how-to-boost-your-vim-productivity
+- Added to OMZ by [@mbologna](https://github.com/mbologna)
+- Two-job switching added by [@Whisperity](https://github.com/Whisperity)
--- a/plugins/fancy-ctrl-z/fancy-ctrl-z.plugin.zsh
+++ b/plugins/fancy-ctrl-z/fancy-ctrl-z.plugin.zsh
@@ -1,16 +1,38 @@
 fancy-ctrl-z () {
-  if [[ $#BUFFER -eq 0 ]]; then
-    BUFFER="fg"
+  local -i hascmd=$(( ${#BUFFER} > 0 ))
+  local -a sjobs=("${(@f)$(jobs -s)}")
+  local -a opts=("${(@f)$(setopt)}")
+  local -i isverbose=$opts[(Ie)verbose]
+  # In Zsh, arrays are 1-indexed, and an empty array has size 1, not 0.
+  # So we must check the first element's length to see whether it describes a
+  # suspended job.
+  local -i nsjobs="${#${(@)sjobs}}"
+  local -i hassjob=$(( ${#sjobs[1]} > 0 ))
+  local -i hassjobs=$(( nsjobs >= 2 ))
+  # Whether the ^Z action will result in a side effect to the terminal.
+  local -i sideeffect=$(( hassjob || hassjobs || isverbose ))
 
-    IFS=$'\n' local num_jobs=($(jobs))
-    if [[ "${#num_jobs[@]}" -eq 2 ]]; then
-      BUFFER="fg %-"
-    fi
-
-    zle accept-line -w
-  else
+  if (( hascmd && sideeffect )); then
+    # Save the current command.
+    # It will be restored after the side-effect is over, e.g., if the
+    # foregrounded job is put to the background again.
     zle push-input -w
-    zle clear-screen -w
+  fi
+
+  if (( hassjobs )); then
+    # Multiple suspended jobs: foreground the second-to-last suspended job.
+    BUFFER="fg %-"
+    zle accept-line -w
+  elif (( hassjob )); then
+    # Single suspended job: foreground it.
+    # "fg %-" would result in an error as the only suspended job is the
+    # last one, which is referenced by "fg" or "fg %+".
+    BUFFER="fg %+"
+    zle accept-line -w
+  elif (( isverbose )); then
+    # No suspended jobs, but verbose mode, let show the "fg: no current job".
+    BUFFER="fg"
+    zle accept-line -w
   fi
 }
 


roeniss:
I have no idea what name would be the best.... flip-flop? fancy-ctrl-z? back-and-forth-fg? Haha. anyway the idea is so cool. 

I always use 2-split pane strategy (w/ iTerm2) for this situation, because I don't want my two job to be suspended (stopped) (e.g., one pane for curl request, one pane for docker logs -f). Your use case, on the other hand, have no problem with suspended jobs, right?

whisperity:
> I have no idea what name would be the best.... flip-flop? fancy-ctrl-z? back-and-forth-fg? Haha. anyway the idea is so cool.

`fancy-ctrl-z` is the name of a plugin already in the repository. I do not think adding a whole new plug-in for this would be a better strategy. 🙂 

> I always use 2-split pane strategy (w/ iTerm2) for this situation, because I don't want my two job to be suspended (stopped) (e.g., one pane for curl request, one pane for docker logs -f).

I could not use split panes for the following reasons.

1. The terminal I'm using does not directly implement pane-based viewports. (Tabs are a recent invention there, but it opened up a can of worms for which I had to roll back to an older version.) I could use `tmux` but because I have very fancy things in my Vim configuration, `tmux`'s inability to understand a few custom VT sequences (that my terminal **does** understand) would lessen the experience.
2. When I do use parallel workflows I open two terminals and use one for this, one for that. Usually, I do use that, one terminal runs the editor(s), the other runs the build / execution. So I don't background my editor to run a build or observe logs.

Sometimes the "main" screen (I call it main as that is the one directly in front of me) has one editor and one `htop` (usually the htop is backgrounded), because sometimes I need to kill the build. But often it's 2 editors for two separate projects I have to wrangle at the same time. If I have to read or change both projects, it's always a sequence of "this uses that, let's go over to see how that one does it".

> Your use case, on the other hand, have no problem with suspended jobs, right?

No. I've found Vim to be very good at handling SIGTSTP, and generally for plugins in Vim which run subprocesses (LSP services, debugger, etc.) are usually cascade-suspended there. Also, for `fancy-ctrl-z` to activate and work, it must have been the user's prior conscious action to have suspended a process.

roeniss:
I understood all your saying. Thank you for the sincere answer. I still have one question.

I think the "exactly two" condition is quite cruel. How about "cyclic switching"?

- If there's one background jobs, I can visit it with 2 `ctrl-z` (fg 1, bg)
- If there's two background jobs, I can visit two background jobs in turn with 4 `ctrl-z` (fg 1, bg, fg 2, bg)
- If there's three background jobs, I can visit all three background jobs in turn with 6 `ctrl-z` (fg 1, bg, fg 2, bg, fg 3, bg)
- ...

I’m not the maintainer, so consider this a casual suggestion please :D

P.S., At first, I thought that `ctrl-z` will switch between two jobs without "bg" i.e., fg 1 -> fg 2 -> fg 1 -> ...

whisperity:
> How about "cyclic switching"?
> 
>    * If there's one background jobs, I can visit it with 2 `ctrl-z` (fg 1, bg)
> 
>    * If there's two background jobs, I can visit two background jobs in turn with 4 `ctrl-z` (fg 1, bg, fg 2, bg)
> 
>    * If there's three background jobs, I can visit all three background jobs in turn with 6 `ctrl-z` (fg 1, bg, fg 2, bg, fg 3, bg)
> 
>    * ...

I don't think we can do this without a lot of additional work. First, `fancy-ctrl-z` would need a way to keep information which was the last visited job (in order to know at which point int the cycle we are at). But jobs can be terminated in the middle of the cycle (have 3 jobs, `fg %2`, then exit). If you terminate a job in the middle of a cycle, the shell with reuse the ID. So if you suspend 3 jobs and then exit the 2nd and create a new – essentially 4th – job, it will get ID 2, so the cycle is "broken" in a sense. But the shell doesn't give you any additional shortcuts apart from `%N` where $N$ is the ID, or `%+` / `%-` where the `%+` refers to the most recently suspended job, and `%-` refers to the second to most recently suspended one. We would have to reimplement a lot of and around the job control mechanism purely in a shell script, which would also not work well if the job control is interfaced with outside of the plug-in.

roeniss:
Thank you for such a deep answer! 

Rest is clear but one thing: Should we have "exactly 2" condition still? Why not just use "last 2"?

- a: added to bg
- b: added to bg
- c: added to bg
- fancy-ctrl-z: open c
- fancy-ctrl-z: c to bg
- fancy-ctrl-z: open b
- fancy-ctrl-z: b to bg
- fancy-ctrl-z: open c
- fancy-ctrl-z: b to bg
- ...

Rest of your words is clear to me, but I still get confused why this can't be activated with 3 background jobs.

whisperity:
@roeniss Thank you for the comments! Admittedly, this fell off my radar for quite some time. I only realised this PR is pending when I made my newer PR #13513 and found that on my machine the OMZ directory that was actually being used by the shell still had this PR checked out as opposed to the `master` branch 😬.

I've given it some thought and also found some bugs related to the current implementation. For example, if you did

```zsh
$ vim a.txt
<background>
$ vim b.txt
<background>
$ cd /tmp
$ ^Z
```

it only ever brought forth `vim b.txt`, which is **clearly a bug** and I'm on track of fixing it.

> Should we have "exactly 2" condition still? Why not just use "last 2"?
> 
> why this can't be activated with 3 background jobs.

Actually, I've given it some thought and tested a few things. With my previous mentions of `%+` and `%-`, it's clear that the Shell can give information about which was the "last 2" **suspended** jobs.
We can think about this like on TV remotes there used to be a button which switched you from the current channel to the channel you viewed prior. So if you went to channel 2 and then 8, pressing the button got you 2, 8, 2, 8, … again. But if after 2, 8, you went to channel 17, then 8, 17, 8, …; it forgot about the "2".

Several tools and utilities support using the `-` marker to explicitly intent switching to the "most recent" thing. For example, `cd -` will switch between your current and your most recent directory, while `git checkout -` will switch between your current and your most recently checked out branch.

And, in fact unlike TV remotes (😂), the shell keeps a stack of this information much better, as if we have 3 suspended jobs and one gets foregrounded and exits, it will return the `%+`, `%-` markers to how they have been previously. Importantly, as well, running background jobs do not "meddle" with this state.

>  * a: added to bg

I think we should be a bit clearer here for the record. There is a difference between "job in foreground/background" vs. "job is running/suspended". Likely 99% of the "foreground" jobs have to be "running", and I can't think of a reasonable case where a foreground job is suspended. However, there can be "background + running" jobs! Everything you start with a `&` or explicitly `bg %N` after a suspension is **running**, just in the background. The simplest explanation for "background" is that their standard stream operations (mounted to the pseudo-tty) would result in _SIGTTIN_ or _SIGTTOU_, which some processes handle well, others do not.

In fact, background-running processes also mislead and caused bugs in the implementation as it is in this patch now.

For example, `yes` can perfectly operate in the background (and it keeps spamming your terminal with `y` lines), but `cat` (no arguments, read from _stdin_, print to _stdout_!) immediately suspends itself when it is attempted to be backgrounded.

Here's a little test:

```zsh
~$ vim A.txt
<suspend>
~$ vim B.txt
<suspend>
~$ jobs
[1]  - suspended vim A.txt
[2]  + suspended vim B.txt
# Clearly, B is the last, and A is the second-to-last job here.
~$ yes >/dev/null &
~$ jobs
[1]  - suspended vim A.txt
[2]  + suspended vim B.txt
[3]    running   yes >/dev/null
~$ cd Desktop/
Desktop$ jobs
[1]  - suspended vim A.txt
[2]  + suspended vim B.txt
[3]    running   yes >/dev/null
Desktop$ jobs -d
[1]  - suspended vim A.txt
(pwd : ~)
[2]  + suspended vim B.txt
(pwd : ~)
[3]    running   yes >/dev/null
(pwd : ~)
Desktop$ cat &
Desktop$ jobs
[1]    suspended             vim A.txt
[2]  - suspended             vim B.txt
[3]    running               yes >/dev/null
[4]  + suspended (tty input) cat
# Now 'cat' is the last, and 'B' is the second-to-last job.
Desktop$ cd ~/
~$ jobs
[1]    suspended             vim A.txt
[2]  - suspended             vim B.txt
[3]    running               yes >/dev/null
[4]  + suspended (tty input) cat
~$ fg
# Equivalent to "fg %+", taking the last suspended job back to the foreground.
[4]    continued             cat
(pwd : ~/Desktop)
cat: stdin: interrupted system call
(pwd now: ~)
# 'cat' exited with 1.
~$ jobs
[1]  - suspended vim A.txt
[2]  + suspended vim B.txt
[3]    running   yes >/dev/null
```

So, yes, we can do "switch between last two suspended jobs" just fine, with what information the Shell can currently provide to us. Foregrounding a non-suspended background job is not that easy (for reasons mentioned, it would require a lot of additional scripting, maybe even additional features added to Zsh itself!) unambiguously. Having multiple jobs with an intermediate one exiting will leave a "slot" open which gets reused by a later-spawned job, so the `%N` ID of the jobs (neither their system PIDs) are good indicator of order:

```zsh
~$ jobs
[1]  - suspended vim A.txt
[2]  + suspended vim B.txt
[3]    running   yes >/dev/null
~$ vim C.txt
<suspend>
~$ jobs
[1]    suspended vim A.txt
[2]  - suspended vim B.txt
[3]    running   yes >/dev/null
[4]  + suspended vim C.txt
~$ fg %3
[3]    running   yes >/dev/null
^C
# yes exits with SIGINT (-2).
~$ jobs
[1]    suspended vim A.txt
[2]  - suspended vim B.txt
[4]  + suspended vim C.txt
~$ vim D.txt
<suspend>
~$ jobs
[1]    suspended vim A.txt
[2]    suspended vim B.txt
[3]  + suspended vim D.txt
[4]  - suspended vim C.txt
```

This patch will change _`fancy-ctrl-z`_ 's behaviour slightly, as the original intent was only to switch back to _the_ (?) background (slightly problematic terminology here again) job. But I think the use of _"fancy"_ is elevated greatly by this feature.

PR #13051, opened 2025-04-02
(enhancement)(themes): Source the default theme (`robbyrussell`) if the one specified in `$ZSH_THEME` does not exist

UnaTried:
## Standards checklist:

## Changes:
 - This change automatically sources the default theme (`robbyrussell`), if the one specified in `$ZSH_THEME` does not exist. If `robbyrussell` also doesn't exist, it gets downloaded using either wget or curl and then automatically sourced.

## Other comments:
  This expects the user to have curl installed, which they probably have.
...

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -239,7 +239,7 @@ if [[ -n "$ZSH_THEME" ]]; then
     elif is_default_theme "$ZSH/themes"; then
       source "$ZSH/themes/robbyrussell.zsh-theme"
     else
-      echo "oh-my-zsh] The default theme (robbyrussell) was not found, redownloading it!"
+      echo "[oh-my-zsh] The default theme (robbyrussell) was not found, redownloading it!"
       if ! command -v curl &> /dev/null; then
         wget -qO $ZSH/themes/robbyrussell.zsh-theme https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/refs/heads/master/themes/robbyrussell.zsh-theme
         source "$ZSH/themes/robbyrussell.zsh-theme"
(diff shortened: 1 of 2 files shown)


roeniss (commented):
Hi UnaTried. Here's my two cents:

1. Why download "default" theme?

A user wanted to download ZSH_THEME. It feels a bit weird to download default theme instead of the exact ZSH_THEME.

2. How about asking users before downloading the theme?

```
[oh-my-zsh] $ZSH_THEME is not found. Download now? [y/N]
```

UnaTried:
Is there a theme repository I don't know about? Or should I make it, that before downloading the default theme, checks if `ZSH_THEME` is equal to other preinstalled themes

UnaTried:
Ok, I just thought about downloading the theme without asking, so newer users wouldn't be wondering: "What is happening, why is my screen broken like that"
Then make a half assed Reddit post or GitHub issue about it

UnaTried:
Ok, I just thought about downloading the theme without asking, so newer users wouldn't be wondering: "What is happening, why is my screen broken like that"
Then make a half assed Reddit post or GitHub issue about it

This PR also only downloads robbyrussell when needed

roeniss:
I got what you mean. I think we have different opinion on "broken screen".

> So newer users wouldn't be wondering: "What is happening, why is my screen broken like that"

I think newer users would not perceive the screen as a broken screen because it'll look like just pure zsh.

carlosala:
Hi! Thanks for your contribution!

I think the current behaviour already informs the user that something is wrong, and they can adjust their config accordingly.
I'm not very keen on merging this one.
Thanks!

Issue #13059, opened 2025-04-11
asdf plugin fails to initialize when asdf is installed locally in ~/.asdf


muhammedogz:
### Describe the bug

After the recent update ([2 weeks ago](https://github.com/ohmyzsh/ohmyzsh/blob/a78d006faa022a5104b542d7dec5a3439e72046e/plugins/asdf/asdf.plugin.zsh) ) removing support for legacy <0.16 asdf setups, the plugin now short-circuits with:

```bash
(( ! $+commands[asdf] )) && return
```

This causes the plugin to exit before setting $PATH or completions, even if the user has asdf properly installed in ~/.asdf, which is the standard install path per official docs.

This breaks valid setups where .asdf is not sourced globally (e.g., via .zshrc) but expected to work via plugin loading.

Expected behavior: if .asdf directory exists, the plugin should initialize asdf even if not yet in $PATH.

Proposed fix: Replace the early command check with a check for the install directory, e.g.:

```bash
if [[ ! -d "$HOME/.asdf" ]]; then
  return
fi
```

At least it is working for me. I am not expert for oh my zsh plugins but when I change this, it is working fine on my end.

### Steps to reproduce

1. install asdf
2. enable asdf plugin with oh my zsh
3. try run `asdf --version`

### Expected behavior

Should initialize asdf without issue

### OS / Linux distribution

Ubuntu 22.04.5 LTS

### Zsh version

5.8.1

### Oh My Zsh version

master (750d3ac)

### Terminal emulator

Windows Terminal

### If using WSL on Windows, which version of WSL

WSL2

### Additional context

I am using `wsl`

muhammedogz:
I opened a PR for my proposal fix. 
Also related
- https://github.com/ohmyzsh/ohmyzsh/pull/13026

carlosala:
Hey! You should already have `asdf` in the path before sourcing `oh-my-zsh.sh`. I'm sure you added `asdf` to the PATH down in your `zshrc`. Just move that line above `source oh-my-zsh.sh`, and it'll pick it up properly.
Thanks!

PR #13060, opened 2025-04-11
fix(asdf): allow plugin to initialize if asdf is installed in ~/.asdf

muhammedogz:
Issue: #13059

Previously, the plugin exited early if the `asdf` command was not already in $PATH:

```bash
(( ! $+commands[asdf] )) && return
```

This broke valid setups where `asdf` is installed manually in ~/.asdf and not yet sourced globally (e.g., via .zshrc). Such installs are still recommended in the official asdf documentation.

This change replaces the early exit with a check for the presence of ~/.asdf, so the plugin only runs if the user has a modern asdf setup, while avoiding premature return before `$PATH` and completions are set up.

This preserves the new plugin behavior (removal of legacy <0.16 logic), while restoring compatibility for the common manual install path.

## Standards checklist:

## Changes:

- update `plugins/asdf/asdf.plugin.zsh`

### DIFF
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -1,4 +1,7 @@
-(( ! $+commands[asdf] )) && return
+# Only continue if .asdf directory exists (manual install) or fallback logic can apply
+if [[ ! -d "$HOME/.asdf" ]]; then
+  return
+fi
 
 export ASDF_DATA_DIR="${ASDF_DATA_DIR:-$HOME/.asdf}"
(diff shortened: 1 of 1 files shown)


muhammedogz:
Also related
- https://github.com/ohmyzsh/ohmyzsh/pull/13026

Probably you are better solution @kartikvashistha I was trying some basic things.
Any feedback appreciated 🙏🏻

carlosala:
See https://github.com/ohmyzsh/ohmyzsh/issues/13059#issuecomment-2809614813

PR #13068, opened 2025-04-15
Added new --prune-tags to git fetch alias

dmetzler1988:
## Standards checklist:

Appended the "new" `--prune-tags` option to `git fetch` (`gfa`) alias.  
It is available since git version 2.17.
See [documentation](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---prune-tags) for it.

## Changes:

- Extend the alias `gfa` with "new" `--prune-tags` option


carlosala:
Hey! I don't think this should be the default. `--prune-tags` removes tags from your git repo if they are not on the remote. If you create a tag and use `--prune-tags`, it gets removed.
I lean towards not merging this one.
Thanks for your contribution!

PR #13075, opened 2025-04-20
feat(command-not-found): added pkgfile presence check for arch

WilfredoN:
## Standards checklist:

## Changes:

- added check if pkgfile is exist in system for arch, so it will not be confusing for user if he forgot to install it and didn't notice the changes when `source .zshrc`

### DIFF
--- a/plugins/command-not-found/command-not-found.plugin.zsh
+++ b/plugins/command-not-found/command-not-found.plugin.zsh
@@ -1,8 +1,18 @@
 ## Platforms with a built-in command-not-found handler init file
 
+# Arch Linux: required 'pkgfile'
+if [[ -f /etc/arch-release ]]; then
+  if command -v pkgfile >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


plugins/command-not-found/command-not-found.plugin.zsh
```diff
@@ -1,8 +1,18 @@
 ## Platforms with a built-in command-not-found handler init file
 
+# Arch Linux: required 'pkgfile'
+if [[ -f /etc/arch-release ]]; then
+  if command -v pkgfile >/dev/null 2>&1; then
```
Vinfall: `if (( ${+commands[pkgfile]} )); then`is significantly faster and the check for `/etc/arch-release` is unnecessary (we have to check the presense of command-not-found handler anyway) and slow down everyone else.

carlosala:
I don't think this check is needed. The user needs to install `command-not-found` in order to use it. In Arch, this is done through `pkgfile` package, the user should do it by himself. It is specified in the readme!
Thanks for your contribution!

PR #13077, opened 2025-04-21
added zsh-autosuggestions

killertofus:
taken from here https://github.com/zsh-users/zsh-autosuggestions

## Standards checklist:

## Changes:

- [...]

## Other comments:

...

### DIFF
--- a/plugins/zsh-autosuggestions/README.MD
+++ b/plugins/zsh-autosuggestions/README.MD
@@ -1,0 +1,1 @@
+As you type commands, you will see a completion offered after the cursor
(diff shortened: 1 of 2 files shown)


plugins/zsh-autosuggestions/README.MD
```diff
@@ -0,0 +1 @@
+As you type commands, you will see a completion offered after the cursor
```
roeniss, line 1: I think we can make a better README than this. Some users won't know that this has been "adopted" from outside. For them, this is all they see. 

Look at the original repo: https://github.com/zsh-users/zsh-autosuggestions. The README contains so many things that I don't think just a line can cover them.

carlosala:
Hey! We're planning to add the plugins with our dependencies system, stay tuned, coming soon :)

PR #13086, opened 2025-04-24
feat(installer): add confirmation before overwriting existing .zshrc

olwooz:
## Standards checklist:

## Issue:

The current Oh My Zsh installation process silently overwrites the user's existing `.zshrc` file without explicit warning or confirmation. This has been a long-standing issue that affects many users, as evidenced by the continued activity on issue #4390 (originally opened in 2015 and still receiving comments in 2025).

The current implementation:
1. Backs up the existing `.zshrc` to `.zshrc.pre-oh-my-zsh`
2. If `.zshrc.pre-oh-my-zsh` already exists, it creates a timestamped backup

However, this happens without user confirmation, which can lead to:
- Users losing their custom configurations unexpectedly
- Confusion about where their original configuration went
- Need to manually restore from backup files
- Potential loss of work if users don't realize their file was backed up

## Changes:

- Added user confirmation prompt before overwriting existing `.zshrc` file
- Added clear messaging about backup location (.zshrc.pre-oh-my-zsh)
- Added option to skip overwrite and keep existing `.zshrc`
- Improved user experience by making the installation process more transparent and safer


killertofus:
will this work with --unattended

olwooz:
@killertofus Yes it does. I've added how I tested it in the PR comment fyi.

christoph-meier-offerup:
PLEASE merge this.

mcornella (changes requested):
OK, it seems we'll finally need to make this change, so there's no way around it.

Some things I'd add/modify:

- We have a minimal zshrc template (see templates directory). Could you add a message showing that this needs to be added to the zshrc file to make set up the zshrc file? Make sure to cat the file path instead of hardcoding the changes. 
- We already have a user prompt style in https://github.com/ohmyzsh/ohmyzsh/blob/1ed17ac0523ebc4d12e4e5324fd5f5ce5dafc668/tools/install.sh#L398, could you update the code to match it?

Then we'll have to update instructions and FAQs, but I can do that myself on top of your PR.

Thanks for the contribution!

mcornella (approved):
Looks good, thanks!

olwooz:
@carlosala Hi! Following up on this PR after 3 months of inactivity. This addresses a long-standing UX issue (#4390) with a minimal, non-breaking change that adds confirmation before overwriting `.zshrc` files.
Since @mcornella has already approved and the implementation is straightforward, could we move this forward? Happy to address any concerns if needed.
Thanks!

carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for your contribution, and sorry for the delay!

PR #13101, opened 2025-05-04
"Add act and runpy aliases to Python plugin"

Hanashiko:
## Standards checklist:

## Changes:

- Added the `act` alias to quickly activate a Python virtual environment.
- Added the `runpy` alias to easily run `main.py` with Python.

## Other comments:

- These aliases simplify the workflow for Python developers by reducing the number of commands needed to activate virtual environments and run the main script.
- The aliases do not duplicate existing functionality and should be useful for most Python projects that use virtual environments.


roeniss:
I recommend to use another alias because of the existence of https://github.com/nektos/act

roeniss:
and how about `pyr` instead of `runpy`?

Hanashiko:
of the `runpy' -> `pyr' trigger, poured the corresponding commit

carlosala:
Hi! Regarding `act` we already have `vrun` with the same purpose. Regarding `pyr`, it seems to be very particular to your use case; not everyone has `main.py` files at the root of their projects.
Thanks for your contribution!

PR #13121, opened 2025-05-18
feat(termsupport): add WezTerm window/tab title support (OSC 1, 2)

tobii-dev:
This adds TERM=wezterm to the list of supported terminals.
https://wezterm.org/recipes/passing-data.html?h=osc#user-vars

WezTerm uses TERM=xterm-256color by default, but it endorses users to use TERM=wezterm in their configuration:
https://wezterm.org/config/lua/config/term.html

With this patch, oh-my-zsh will still support setting the window and tab name.

## Standards checklist:

## Changes:

- add TERM=wezterm to the list of supported terminals in [lib/termsupport.zsh](https://github.com/tobii-dev/ohmyzsh/blob/78cfcccce4ae5c8c551e72e8f32eec11776c0dc6/lib/termsupport.zsh#L20)

### DIFF
--- a/lib/termsupport.zsh
+++ b/lib/termsupport.zsh
@@ -17,7 +17,7 @@ function title {
   : ${2=$1}
 
   case "$TERM" in
-    cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty*|st*|foot*|contour*)
+    wezterm*|cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty*|st*|foot*|contour*)
       print -Pn "\e]2;${2:q}\a" # set window name
       print -Pn "\e]1;${1:q}\a" # set tab name
       ;;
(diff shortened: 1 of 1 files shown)


lib/termsupport.zsh
```diff
@@ -17,7 +17,7 @@ function title {
   : ${2=$1}
 
   case "$TERM" in
-    cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty*|st*|foot*|contour*)
+    wezterm*|cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty*|st*|foot*|contour*)
```
carlosala, line 20:
```suggestion
    cygwin|xterm*|putty*|rxvt*|konsole*|ansi|mlterm*|alacritty*|st*|foot*|contour*|wezterm*)
```

carlosala (approved):
LGTM! 🚀

PR #13128, opened 2025-05-20
feat(cli) Add flag to `omz plugin list` to show only enabled plugins

glektarssza:
## Standards checklist:

## Changes:

- Added a flag to `omz plugin list` which allows users to list only enabled plugins by category (built-in vs. custom)

## Other comments:

Should be relatively efficient due to use of regex comparison as opposed to double `for` loops.

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -449,30 +449,22 @@ function _omz::plugin::info {
 
 function _omz::plugin::list {
   local -a custom_plugins builtin_plugins
-  custom_plugins=("$ZSH_CUSTOM"/plugins/*(-/N:t))
-  builtin_plugins=("$ZSH"/plugins/*(-/N:t))
 
-  # If --enabled is provided, filter plugins by what's enabled
+  # If --enabled is provided, only list what's enabled
   if [[ "$1" == "--enabled" ]]; then
-    # echo $plugins[@]
     local plugin
-    local -a new_custom_plugins new_builtin_plugins
-    for plugin in "${custom_plugins[@]}"; do
-      # Spaces are to ensure we don't match substrings of other plugins
-      # This avoids the need for a second, inner loop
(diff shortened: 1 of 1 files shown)


mcornella (changes requested):
Hi, thanks for the contribution. I think it's a valid use case, but I'd approach the code differently:

If `--enabled` is provided, skip listing all custom and built-in plugins and instead:

- Iterate over every plugin in `$plugins`
- If `$plugin` is found in `$ZSH_CUSTOM` add to custom
- If found in `$ZSH` add to built-in
- If none, skip plugin from the list

Could you make this change or do you need additional guidance? Thanks!

glektarssza:
I can definitely make those changes. No further guidance required as the implementation requirements you're looking for are quite.

I will submit an update later tonight once I get off work and have some time to implement.

lib/cli.zsh
```diff
@@ -449,8 +453,23 @@ function _omz::plugin::info {
...
+  if [[ "$1" == "--enabled" ]]; then
+    local plugin
+    for plugin in "${plugins[@]}"; do
+      if [[ -d "${ZSH_CUSTOM}/plugins/${plugin}" ]]; then
+        custom_plugins+=("${plugin}")
+      elif [[ -d "${ZSH}/plugins/${plugin}" ]]; then
+        builtin_plugins+=("${plugin}")
+      fi
+    done
+  elif [[ -n "$1" ]]; then
```
mcornella, line 467: This case is not necessary, it does nothing.

mcornella (changes requested):
Looks good, but please remove that unnecessary elif. Thanks!

lib/cli.zsh
glektarssza: Apologies, forgot to remove that! Resolved in 76047d16!

PR #13143, opened 2025-05-25
feat(grep): Use ggrep via Homebrew if it is installed

einstein95:
## Standards checklist:

## Changes:

- This checks if `ggrep` (GNU Grep) is installed via Homebrew and uses that for the grep aliases

## Other comments:

I am willing to admit that this may not get merged due to forcing the user into using grep installed via Homebrew, which at time of writing is `ggrep (GNU grep) 3.12`, over the native macOS version which is `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD`. This PR is mainly for my own uses, as I was sick of lacking `-P` for more powerful regex grepping and lacking colored output when using `grep`.

### DIFF
--- a/lib/grep.zsh
+++ b/lib/grep.zsh
@@ -5,8 +5,14 @@ __GREP_ALIAS_CACHES=("$__GREP_CACHE_FILE"(Nm-1))
 if [[ -n "$__GREP_ALIAS_CACHES" ]]; then
     source "$__GREP_CACHE_FILE"
 else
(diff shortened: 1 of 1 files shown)


carlosala:
Hey! I'm leaning towards not merging it. You can easily create your own alias after `source oh-my-zsh.sh` in your `zshrc`. Polluting that file with platform-specific logic isn't desirable IMO.

PR #13149, opened 2025-06-01
refactor(git)!: Update alias and function deprecations

rivenirvana:
## Standards checklist:

## Changes:

- Remove deprecated `git pull --rebase` aliases
- Start deprecating `ggpull`/`ggpur`/`ggpush` for their alternative functions
- Start deprecating `current_branch` function in favor of what's in core
- Update README to reflect above changes and undocument some long-removed aliases

---

* Closes/supercedes #9544 
* Closes/supercedes #11227

## Other comments:

I'm gonna be honest, I only wanted to just be able to use [gup](https://github.com/nao1215/gup/) without `unalias`-ing it first in my zshrc. But the code for deprecation warnings seemed like a waste to just remove, and after taking a look it felt like some parts needed updates too, so why not make use of it so we can proceed with a system where we can actually transition functionalities from soft deprecation to removal? Some of these have been sitting here for a while now, after all (>2 years). Appreciate any feedback/conversation regarding this.

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -238,12 +238,10 @@ receive further support.
 | :------- | :-------------------------------------------------------- | :-----------------------------------------------------|
 | `gap`    | `git add --patch`                                         | New alias: `gapa`                                     |
 | `gcl`    | `git config --list`                                       | New alias: `gcf`                                      |
-| `gdc`    | `git diff --cached`                                       | New alias: `gdca`                                     |
 | `gdt`    | `git difftool`                                            | No replacement                                        |
 | `ggpull` | `git pull origin $(git_current_branch)`                   | New function: `ggl` (`ggpull` is now aliased to this) |
 | `ggpur`  | `git pull --rebase origin $(git_current_branch)`          | New function: `ggu` (`ggpur` is now aliased to this)  |
 | `ggpush` | `git push origin $(git_current_branch)`                   | New function: `ggp` (`ggpush` is now aliased to this) |
-| `gwc`    | `git whatchanged -p --abbrev-commit --pretty=medium`      | New alias: `gwch`                                     |
 
 ## Functions
 


rivenirvana:
Is nobody interested in this change?

Kludex:
I've been using `ggpull` since the beginning of my career. Where is the discussion related to this?

yilas:
To be honest, I don’t really understand the rationale behind deprecating `ggpull` and `ggpush`.

rivenirvana:
Hmm. At the time, I was under the impression that those 3 aliases were to be replaced by their corresponding functions that have also been there for quite a long time now. Was following up on this commit https://github.com/ohmyzsh/ohmyzsh/commit/207d29b716301c77392ae140b1482eeba44e9102 that started deprecation notices, which seemed like the perfect opportunity to proceed with it. `ggpur` was already just an alias for the function `ggu`. The aliases are still there, I just had them prefixed with the deprecation text, is behavior still inconsistent outside of the extra message, like not passing flags? Completion should work just fine too.

Honestly, I didn't think this PR would get merged. Been bumping this regularly for months to no reply, but I've been fine staying on the branch locally. If users really feel strongly about this, then that's my bad for misinterpreting the maintainers' intentions and user concensus. Ultimately though, these discussions can only happen when the changes are merged, so comments are always appreciated.

Kludex:
> If users really feel strongly about this

I do, to the point that after I wrote the comment, I've added the shortcut on my `.zshrc` after loading `ohmyzsh`. I don't have the warning anymore, and when it's removed, I still have the command.

Anyway, I think it's an unnecessary change. People usually prefer new features, but never like when things they use are removed. If maintaining `ggpull` is not much of work, maybe it can be added back? Otherwise, I already have my workaround... 🤷‍♂️

mcornella:
Hi @Kludex! As far as I remember ggl and the rest were made to support many different workflows, including ggpull, etc. This was added a while back (years ago), and we only documented the deprecation but not made the code change.

I can understand how that can be hard to get used to, especially when the aliases have been long working. I'm OK with reverting the deprecation of those 3 aliases.

@rivenirvana could you PR a revert of that change, i.e. undeprecate the aliases? `current_branch` is OK to be deprecated, it's usually a set it and forget it kind of thing.

rivenirvana:
#13485

Thoughts on `alias ggpull='ggl'` and `alias ggpush='ggp'` (just without the notice), or are they actually incompatible? In any case, PR is ready.

forddavis:
Guess I'm beyond late to the party here, but the update message still says `ggpull` and `ggpush` are deprecated? Are they being removed?

+1 in favor of not deprecating from me.

Kludex:
It was solved in #13485, not sure it was released yet (I'm on the phone).

mcornella:
Correct, the revert is already released. Force an update with `omz update`. Thanks!

forddavis:
Ah very nice. Thank you!

PR #13158, opened 2025-06-07
fix(open_command,wsl): Escape cmd.exe special chars for URLs in open_command when using WSL

jwblangley:
## Standards checklist:

- [x] If the code introduces new aliases, I provide a valid use case for all plugin users down below: N/A

## Context

Use `open_command` on WSL to open a URL in the browser.

```zsh
open_command "https://example.com/test?arg=value&key=value"
```

Currently, this is executed with `cmd.exe`, which considers `&` (amongst others) as special characters that are not escaped.
The result of this is that the actual URL that is opened is truncated at the first special character. 
A prevalent case where this occurs is with multiple URL arguments, as shown in the above example.

## Changes:

- If detecting WSL, look for URLs (using the same definition for using `BROWSER`), and escape any of the following characters: `&|()<>^` with the `^` escape character (for cmd.exe)

## Questions for reviewers

* I am making use of `sed -E` here, are there any requirements about if `sed` with extended regular expressions is an expected dependency or not?

### DIFF
--- a/lib/functions.zsh
+++ b/lib/functions.zsh
@@ -23,6 +23,7 @@ function open_command() {
     linux*)   [[ "$(uname -r)" != *icrosoft* ]] && open_cmd='nohup xdg-open' || {
                 open_cmd='cmd.exe /c start ""'
                 [[ -e "$1" ]] && { 1="$(wslpath -w "${1:a}")" || return 1 }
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #13162, opened 2025-06-08
fix(genpass): follow Apple password specification requirements

ryanvo:
## Standards checklist:

## Changes:

- Modifies genpass-apple to adhere to the spec and match the password format generated by Apple
- Password is now 20 characters long
- Digit are only be placed on either side of hyphen or at the end of password
- Syllables only begin with a consonant
- Consonant list does not include 'L' character

### DIFF
--- a/plugins/genpass/README.md
+++ b/plugins/genpass/README.md
@@ -19,19 +19,20 @@ passwords in a script:
 ## genpass-apple
 
 Generates a pronounceable pseudoword passphrase of the "cvccvc" consonant/vowel
-syntax, inspired by [Apple's iCloud Keychain password generator][1]. Each
-password has exactly 1 digit placed at the edge of a "word" and exactly 1
-capital letter to satisfy most password security requirements.
+syntax based on [Apple Passwords’ Generated Strong Password Format][1]. Each
+password has exactly 1 digit placed on other either side of the hyphen or at the
+end of the password and exactly 1 capital letter to satisfy most password security
+requirements.
 
     % genpass-apple
-    gelcyv-foqtam-fotqoh-viMleb-lexduv-6ixfuk
+    nyvtyv-mYhmob-xyqby4
 
     % genpass-apple 3
-    japvyz-qyjti4-kajrod-nubxaW-hukkan-dijcaf
-    vydpig-fucnul-3ukpog-voggom-zygNad-jepgad
-    zocmez-byznis-hegTaj-jecdyq-qiqmiq-5enwom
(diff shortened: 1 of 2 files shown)


atoponce:
The `genpass-*` generators are all specifically targeting 128 bits of symmetric security, this why `genpass-apple` is not 20 characters. Rather, the standard 20 characters only provides about 72 bits: log2(17 * 19^11 * 6^6 * 10 * 6) ≈ 72.231.

However, it does adhere to [Apple's password rules validation tool](https://developer.apple.com/password-rules/), just enforcing the pronounceable `cvvcvc` structure..

ryanvo:
> The `genpass-*` generators are all specifically targeting 128 bits of symmetric security, this why `genpass-apple` is not 20 characters. Rather, the standard 20 characters only provides about 72 bits: log2(17 * 19^11 * 6^6 * 10 * 6) ≈ 72.231.
> 
> However, it does adhere to [Apple's password rules validation tool](https://developer.apple.com/password-rules/), just enforcing the pronounceable `cvvcvc` structure..

The spec is described [here](https://rmondello.com/2024/10/07/apple-passwords-generated-strong-password-format/) by the developer at Apple who designed it. They target 71 bits. Why shouldn't this plugin behave like the apple password generator?

atoponce:
Because I designed it to provide security for cryptographic symmetric keys which is at least 128 bits.

ryanvo:
What if we made the number of words a parameter? My use case is to generate passwords exactly like the Apple Passwords app in a non-Mac environment. I understand your requirement, but the Apple spec is a practical balance.

atoponce:
Outside of the argument to generate multiple passwords, I'd rather not implement any additional command line parsing. The beauty of the way it's designed now is it is both convenient _and_ secure. There are no foot guns in the plugin that weaken or compromise security. If you want a weaker password, you have to make the conscious decision to do that outside of the tool.

The password is hyphenated, so it should be trivial to pipe it to something like `sed(1)` or `awk(1)` in your scripts to trim it down if you want a shorter version that strictly adheres with the 72-bit/20-character implementation generated by iOS and Safari.

carlosala:
> The password is hyphenated, so it should be trivial to pipe it to something like `sed(1)` or `awk(1)` in your scripts to trim it down if you want a shorter version that strictly adheres with the 72-bit/20-character implementation generated by iOS and Safari.

I'd rather say this is a good point, let's stick to the current implementation.
Thanks for your contribution!

PR #13163, opened 2025-06-09
Adding gdno for git diff name only to list changed files

jfarcher:
## Standards checklist:

## Changes:
 Adds gdno alias for the git plugin, git diff --name-only lists the changed files
- [...]

## Other comments:

...


drjaska:
The addition here is a good idea in my opinion but you forgot to document your new alias in the plugin's README.md file.

Btw you typo'd your markdown checklist, `- [x]` is valid but `- [ x ]` is not and thus it didn't format properly.

PR #13173, opened 2025-06-17
feat(plugin): dockolor - docker ps with color based on status

bouteillerAlan:
## Standards checklist:

- [x] The code is mine, or it's from somewhere with an MIT-compatible license.
- [x] The code is stable, and I have tested it myself, to the best of my abilities.

## Changes:

- Set a function `dockolor` that colorizes `docker ps` output based on container status. 
- Safely overrides `dps`, `dpsa` alias from [the docker plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/docker) for a more unified Docker experience.
- If `awk` is not installed, the script returns the `docker ps` command has is with a message `You have installed dockolor but you don't have awk installed`.

## Use case

I'm sometime lost in my `docker ps` list, so we can improve the readability of the output just with colored line:

before
image

after
image

## Other comments:

I first started to do this plugin in go, but for keeping this code simple as possible without unnecessary dependencies, I move it to plain sh.

Tell me if something is wrong, if something need a refactor or else :)

### DIFF
--- a/plugins/dockolor/README.md
+++ b/plugins/dockolor/README.md
@@ -49,7 +49,7 @@ source ~/.zshrc
 
 ## Usage
 
-- `dockolor` — Same as `docker ps --no-trunc`, but colorized.
-- `dps` — Replaced with `dockolor_dps`, with color.
-- `dpsa` — Replaced with `dockolor_dps -a`, showing all containers, with color.
+- `dockolor` — Same as `docker ps`, but colorized.
+- `dps` — Replaced with `dockolor_dps`, same as `dockolor` or `docker ps`.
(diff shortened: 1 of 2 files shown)


carlosala:
Very interesting plugin! Nevertheless, it seems very specific in its goal, I'd say it is well suited to our External Plugins list in the [wiki](https://github.com/ohmyzsh/wiki).
If then it gets traction, we can upstream it here.
Thanks for your contribution!

bouteillerAlan:
I have to open a PR on the wiki repo to propose it to be there, right?

PR #13175, opened 2025-06-18
feat(ubuntu): add apt-fast support for faster package management

AmrElsayyad:
## Standards checklist:

## Changes:

• Enhanced the ubuntu plugin to detect and use apt-fast when available, providing faster package operations
• Maintains backward compatibility by falling back to apt or apt-get when apt-fast is not installed

### Test plan

- [x] Test with apt-fast installed - verify aliases use apt-fast
- [x] Test without apt-fast but with apt - verify aliases use apt  
- [x] Test on older systems with only apt-get - verify aliases use apt-get
- [x] Verify all existing aliases continue to work as expected

### DIFF
--- a/plugins/ubuntu/README.md
+++ b/plugins/ubuntu/README.md
@@ -10,10 +10,11 @@ plugins=(... ubuntu)
 
 ## Aliases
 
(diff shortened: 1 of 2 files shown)


carlosala:
Thanks for your contribution!

PR #13185, opened 2025-06-25
fix(rbw): do not wipe `rbwpw` items if newer are copied

carlosala:
## Standards checklist:

## Changes:

- Only wipe `rbwpw` content if it is the latest one to be added. This avoids getting passwords wiped when chaining requests.


mcornella (approved):
Added some improvements, LGTM like this

krismawan4:
awesome work thank you

Issue #13106, opened 2025-05-06
keychain: Update plugin to be compatible with 2.9.0


marcofranssen:
### If the feature request is for a plugin or theme, specify it here.

keychain

### If the feature solves a problem you have, specify it here.

Since keychain 2.9.0 the `--agents` option is deprecated, based on the keys it will automatically start the correct agents.

See: https://github.com/funtoo/keychain/releases/tag/2.9.0

### Describe the proposed feature.

Update the plugin to not use the `--agents` argument anymore.

marcofranssen:
I don't have the time to do so at the moment. Might be able in few weeks 🤞 

Feel free to pick up the issue and fix it in the meantime if you have time on your side.

ghost:
@zack6849 Thanks for the patch. However, the logic seems backwards. If is-at-least 2.9 shouldn't be adding the --agents flag, but the other way around. I have Keychain 2.9.5, and it warns me every time. (I copied the updated file into custom and just reversed two invocations beneath the if statement, and all is well.)

zack6849:
@todfox-net was my original code on the pr correct? https://github.com/ohmyzsh/ohmyzsh/pull/13189 I wrote it a bit simplistically and it was improved to use a better approach, i'm wondering if that logic was backwards in the improved version

ghost:
Yes, the logic seems reversed in the new pr.

Originally it had -
	"if (( major < 2 || (major == 2 && minor < 9) )); then"
which checks if it's _less than_ 2.9 and applies the line.

In the new script pr it has -
	"if is-at-least 2.9 "$version_string"; then"
which checks if it's _at least_(_greater than or equal to_) 2.9 and applies the line.

However, as the code inside the if/then/else block wasn't similarly flipped, the --agents code is being applied against 2.9 and not against the earlier versions.

carlosala:
Logic reversed, sorry for the inconvenience caused.

AnonymousX86:
Issue still exists. Installed `ohmyzsh` and `keychain` on a fresh Fedora 43, and it spits out error about `--agents is deprecated`.

carlosala:
@AnonymousX86 could you share the output of `keychain --version`?

AnonymousX86:
Sorry, forgot to include.
```
$ keychain --version

* keychain 2.9.2 ~ http://www.funtoo.org/Funtoo:Keychain

[...]

PR #13189, opened 2025-06-27
[#13106] Update keychain plugin to support 2.9.0+ without warnings

zack6849:
Keychain versions 2.9.0 and higher removed support for the --agents option and squawk in console constantly if the option isn't removed, now we check to make sure that the keychain version is less than 2.9.0 before passing the argument

## Standards checklist:

## Changes:

- Changed the plugin to only pass --agents on versions lower than 2.9.0 

## Other comments:

I tested this with keychain v 2.9.5 on arch and it seems to work, however i'm not sure if the version output is the same on other operating systems, we could also just change this to not pass the option at all, or only pass the agents option if some flag is set or something, open to suggestions / criticisim

### DIFF
--- a/plugins/keychain/keychain.plugin.zsh
+++ b/plugins/keychain/keychain.plugin.zsh
@@ -21,12 +21,9 @@ function {
 
 	# Check keychain version to decide whether to use --agents
 	local version_string=$(keychain --version 2>&1 | head -n 2 | tail -n 1 | cut -d ' ' -f 4)
-	local -a version_parts=(${(s:.:)version_string})
-	local major=${version_parts[1]:-0}
-	local minor=${version_parts[2]:-0}
-
-	# start keychain, only use --agents for versions below 2.9.0
-	if (( major < 2 || (major == 2 && minor < 9) )); then
+  # start keychain, only use --agents for versions below 2.9.0
+	autoload -Uz is-at-least
+	if is-at-least 2.9 "$version_string"; then
 		keychain ${^options:-} --agents ${agents:-gpg} ${^identities} --host $SHORT_HOST
 	else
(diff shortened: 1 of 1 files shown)


zack6849:
@marcofranssen any chance you could review this when you get a chance?

marcofranssen (approved):
LGTM 🚀

plugins/keychain/keychain.plugin.zsh
```diff
@@ -19,8 +19,18 @@ function {
...
-	# start keychain...
-	keychain ${^options:-} --agents ${agents:-gpg} ${^identities} --host $SHORT_HOST
+	# Check keychain version to decide whether to use --agents
+	local version_string=$(keychain --version 2>&1 | head -n 2 | tail -n 1 | cut -d ' ' -f 4)
+	local -a version_parts=(${(s:.:)version_string})
+	local major=${version_parts[1]:-0}
+	local minor=${version_parts[2]:-0}
+
+	# start keychain, only use --agents for versions below 2.9.0
+	if (( major < 2 || (major == 2 && minor < 9) )); then
```
carlosala, line 29:
```suggestion
  # start keychain, only use --agents for versions below 2.9.0
	autoload -Uz is-at-least
	if is-at-least 2.9 "$version_string"; then
```

carlosala (dismissed):
I like it, just a quick idiomatic fix!

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

zack6849:
@carlosala just a heads up https://github.com/ohmyzsh/ohmyzsh/issues/13106#issuecomment-3094448973

PR #13194, opened 2025-07-01
feat(gh-aliases): add GitHub CLI aliases plugin

3dyuval:
## Summary

Add gh-aliases plugin providing convenient aliases for GitHub CLI commands.

## What it does

This plugin adds short, memorable aliases for common GitHub CLI operations:
- Repository management (clone, create, view, etc.)
- Issues and pull requests 
- Workflows and releases
- Authentication and configuration
- Helper functions for common workflows

## Why add this

GitHub CLI has become essential for many developers' workflows. This plugin makes those commands faster to type and easier to remember, similar to how the git plugin provides aliases for git commands.

## Examples

```bash
ghrc user/repo     # gh repo clone user/repo
ghpc               # gh pr create  
ghil               # gh issue list
ghwl               # gh workflow list
```

The plugin includes 100+ aliases covering the full GitHub CLI feature set, plus helper functions like `ghrc-cd` (clone and cd into repo).

### DIFF
--- a/plugins/gh-aliases/README.md
+++ b/plugins/gh-aliases/README.md
@@ -1,0 +1,371 @@
+# gh-aliases
+
+Provides extensive aliases for the [GitHub CLI](https://cli.github.com/) (`gh`) commands, covering repositories, issues, pull requests, workflows, and more. This plugin makes GitHub CLI operations faster and more convenient by providing memorable short aliases for common tasks.
+
+To use it, add `gh-aliases` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... gh-aliases)
+```
+
+## Requirements
+
+- [GitHub CLI](https://cli.github.com/) (`gh`) must be installed
+- You must be authenticated with GitHub (`gh auth login`)
+
+## Aliases
+
+### Authentication Commands (`gh auth`)
+
+| Alias | Command | Description |
+| :---- | :------ | :---------- |
+| `ghal` | `gh auth login` | Login to GitHub |
+| `ghas` | `gh auth status` | Show authentication status |
+| `ghaw` | `gh auth switch` | Switch between GitHub accounts |
+| `ghao` | `gh auth logout` | Logout from GitHub |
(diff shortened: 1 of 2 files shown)


3dyuval:
📚 I've also submitted a corresponding documentation PR to add this plugin to the wiki plugins list: https://github.com/ohmyzsh/wiki/pull/74

adamchainz:
Hi, thanks for your work !

I'm not a big contributor to oh-my-zsh, but I did some work and review for the `git` plugin.

I would like to ship some `gh` aliases, too! However, I think we can refine the approach here.

First, you've aliased most of the `gh` subcommands, but I think we should limit the list to the most popular 10 or so aliases, especially at first. The `gh auth login` command, for example, is pretty useless, as it only needs to be run once, when setting up `gh`. I also think most of the shell functions could be removed.

Second, some key aliases are missing, at least `gh issue develop --checkout`. That's a very neat workflow command for going from issue number to local branch ready to make a PR.

Third, I think placing the aliases in the existing `gh` plugin would be fine.

3dyuval:
Yes, thanks! I mostly agree.
It's very personal. I actually have many usecases for ghat, just tody I did this:
```
docker run -i --rm \
  -e GITHUB_PERSONAL_ACCESS_TOKEN=$(ghat) \
  -e GITHUB_DYNAMIC_TOOLSETS=1 \
  ghcr.io/github/github-mcp-server
```

Looking at the full plugin, I see an opportunity for customization. 
Do you know of any plugins that allows to opt-in/out of features?

adamchainz:
> It's very personal. I actually have many usecases for ghat, just tody I did this:

Yeah, that one seems useful.

> Do you know of any plugins that allows to opt-in/out of features?

No, I believe the goal of oh-my-zsh is to provide minimal configuration.

carlosala:
These are for sure too many aliases, and if a couple of them are actually necessary, they should go to `gh` plugin. Thanks for your contribution!

PR #13204, opened 2025-07-09
feat(kubectl): change alias from kl to klog

PixiBixi:
## Standards checklist:

## Changes:

- kl is a Kube ecosystem software (https://github.com/robinovitch61/kl) 

## Other comments:

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -129,7 +129,7 @@ alias kga='kubectl get all'
 alias kgaa='kubectl get all --all-namespaces'
 
 # Logs
-alias kl='kubectl logs'
+alias klog='kubectl logs'
 alias kl1h='kubectl logs --since 1h'
 alias kl1m='kubectl logs --since 1m'
 alias kl1s='kubectl logs --since 1s'


PixiBixi:
@mcornella ? :)

PixiBixi:
@carlosala dont know if ure legit to approve it but its open for 2 months for a simple change 😅

mcornella:
Hey sorry for the delay here, I really appreciate the contribution. Thanks!

jessegoodier:
Why was this changed? I used it. 

You didn't change `klf` which is now inconsistent with `klog`

mcornella:
Hi @jessegoodier, kl is a k8s tool, we have a soft policy to not conflict with commands. You can use `klf`, it's not conflicting with anything else.

ciudilo:
Breaking a common workflow for an obscure binary conflict seems like the wrong trade-off here. It disrupts the muscle memory of a massive number of users to accommodate a relatively niche tool. Given that the Oh My Zsh alias almost certainly predates that tool, the priority should probably be on stability for existing users.

PixiBixi:
Obscure binary for you. 

Every binary is "obscure" before being well used 

I'm not sure breaking kl is a game changer tbh

mcornella:
We don't have usage metrics, so I'm not sure to revert or not. I could either revert, keep the change or add logic to only rename if kl is installed.

It's true that kl seems less popular and less standardized than other commands for which we typically make these changes, but I'm also not sure if the alias was ingrained in users' workflows.

I'll wait for more feedback to make a decision. Feel free to share this PR with any people you know that might be affected by the change. Thanks

jessegoodier:
IMO, it is the responsibility of the new tool to handle the conflict.
`kl` was added in 2018!

I even have such a tool: https://github.com/jessegoodier/kpf?tab=readme-ov-file#installation

Just have the users `unalias kl` in their .zshrc after the plugin section.

I would revert this PR. Though i've alraedy updated my alias to fix this breaking change.

If you don't revert, I would at least make the aliases consistent. 
image

bjschafer:
This was super ingrained in my workflow and was a jarring change. My vote would be to revert this PR.

cprivitere:
This is an interesting discussion. Probably needs to result in some sort of threshold a tool needs to pass to be relevant enough to move aliases around. 

For example, it turns out `klog` is ALSO an obscure binary that people use:

https://github.com/jotaen/klog/

(as well as a go library, but that's probably not gonna cause problems here).

mimparat132:
This would be a jarring change for me. Working in the k8s space I'd wager using vanilla "kubectl logs" over the kl utility is a much more common workflow.

TomasTokaMrazek:
I've spend half an hour trying to figure out, why `kl` stopped working, while `klf` doesn't. I don't think that changing aliases under our hands is a good idea. I get, that the soft rule is to avoid conflicts, however I see two issues with how things currently are. First, if `kl` was renamed, then `klf` should be renamed too. Second, with the sheer amount of aliases for `kubectl` commands, there is bound to be a conflict with something anyway. I also vote for reverting the PR.

mcornella:
Thank you all for the feedback, I've created #13496 to revert this, and we'll take this into consideration for future alias conflicts.

mcornella:
Merged, you can reset to old behavior with `omz update`

PR #13212, opened 2025-07-13
perf(git): improve git_main_branch() to use remote HEAD symbolic ref

ElisarEisenbach:
## Standards checklist:

## Changes:
Fall back to `git symbolic-ref` to query the default branch from remote HEAD
after checking common branch names. This approach is more accurate,
and respects Git's own understanding of the default
branch regardless of naming conventions.


cybercoder-naj:
Thank you, i was going to create a PR for this myself. Hopefully it is merged soon

plugins/git/git.plugin.zsh
```diff
@@ -31,16 +31,27 @@ function git_develop_branch() {
...
   done
+  
+  # Fallback: try to get the default branch from remote HEAD symbolic refs
+  for remote in origin upstream; do
+    ref=$(command git symbolic-ref --quiet refs/remotes/$remote/HEAD 2>/dev/null)
+    if [[ -n $ref ]]; then
+      echo ${ref#refs/remotes/$remote/}
+      return 0
+    fi
+  done
```
cybercoder-naj, line 54: What do you think of replacing it with [this answer](https://stackoverflow.com/a/28669194/11292068)?

ElisarEisenbach: no problem, i think that under the hood its the same logic, but I agree its cleaner approach.
my question is do we want to replace the fallback to be the first option? as people might set custom br for their main br, even if they have a br called main
wdyt?

carlosala (dismissed):
LGTM!

gomdopi:
I see this has already been merged, but could we check the remote HEAD symbolic refs first?

Or include `dev || develop` on this line: https://github.com/ohmyzsh/ohmyzsh/pull/13212/files#diff-bd72f3b7bb82839b6e8f34b96e7fc5ec3c923d24b79bec12427df62f07136678R40

That way, `master` isn't chosen over `develop`. There are some repos that may have a `master` branch defined but use `develop` as the "main" branch for example.

carlosala:
Actually, that function is meant to find the `main` branch. I think it is difficult to guess which branch is the appropriate one for every specific repository. This symbolic ref thingy was added just as a fallback, I wouldn't change the logic anymore.

jabari-max (approved):

jabari-max (approved):

jabari-max:
👍

PR #13228, opened 2025-07-24
feat(git-commit): add extra scopes

jonas-grgt:


carlosala:
Hey! Both `feat` and `test` are duplicated, and `ops` doesn't seem to be a standard scope in any conventional commits I found.
Thanks for your contribution!

PR #13253, opened 2025-08-07
Fix zsh completion for aws plugin.

mijoharas:
This is following the documentation for completion from aws here https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-completion.html. I imagine this may work if other plugins are executed before this that perform this action already, but it was broken for me.

## Standards checklist:

## Changes:

- Fix zsh completion for aws plugin.

## Other comments:

This is following the documentation for completion from aws here
https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-completion.html. I
imagine this may work if other plugins are executed before this that perform
this action already, but it was broken for me.

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -296,6 +296,7 @@ fi
 # AWS CLI v2 comes with its own autocompletion. Check if that is there, otherwise fall back
 if command -v aws_completer &> /dev/null; then
   autoload -Uz bashcompinit && bashcompinit
+  autoload -Uz compinit && compinit
   complete -C aws_completer aws
 else
   function _awscli-homebrew-installed() {


carlosala:
Hi! Oh My Zsh is already handling `zsh` completion after sourcing all plugins. I tried myself and it is working fine, it shouldn't be a problem that compinit is executed after the `complete` command. Is it possible that you're sourcing the plugin without `oh-my-zsh` core? If that's the case, you should take care of your own compinit.

mijoharas:
Ahhh... That's probably it. I'd assumed it was an ordering issue, but I use `antigen` and load oh my zsh plugins, which must not include the core. (I can see where it's loaded now)

I can manually load compinit myself if we don't expect this plugin to be able to be loaded separately and I'll keep that in mind for the future.

carlosala:
Nice then! Thanks for your contribution!

PR #13268, opened 2025-08-15
bug(asdf): don't add $ASDF_DATA_DIR/shims to the path if already present

phealy:
## Standards checklist:

## Changes:

-  don't add `$ASDF_DATA_DIR/shims` to the path if already present

## Other comments:

This is just a minor change to prevent `$ASDF_DATA_DIR/shims` from being added to the path multiple times (for example, if you `exec zsh`, your shims directory will keep getting added).

@RobLoach

### DIFF
--- a/plugins/asdf/asdf.plugin.zsh
+++ b/plugins/asdf/asdf.plugin.zsh
@@ -1,7 +1,9 @@
 (( ! $+commands[asdf] )) && return
 
 export ASDF_DATA_DIR="${ASDF_DATA_DIR:-$HOME/.asdf}"
-path=("$ASDF_DATA_DIR/shims" $path)
+
+# Only add to the path if not already present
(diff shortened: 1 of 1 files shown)


carlosala:
I like it! Nevertheless, it might be interesting to keep it in the first place of the PATH. Could you modify the logic to be:
- If it's present, remove it.
- Add it as usual.

phealy:
> I like it! Nevertheless, it might be interesting to keep it in the first place of the PATH. Could you modify the logic to be:
> 
> * If it's present, remove it.
> * Add it as usual.

How about this?

`path=("$ASDF_DATA_DIR/shims" ${path:#$ASDF_DATA_DIR/shims})`

I updated the assignment so that it will prepend the shims entry to the path array with any entries matching the shims path removed. Your suggestion was great, and I think this is even cleaner than what I had before. This ensures that there aren't duplicate entries in the path and that it's the first entry in the array.

carlosala (approved):
LGTM! 🚀

carlosala:
Thanks for your contribution!

PR #13275, opened 2025-08-19
fix(install): ensure `--unattended` is respected

carlosala:
## Standards checklist:

## Changes:

Ensure `--unattended` is respected in the installer.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -344,7 +344,7 @@ setup_zshrc() {
       return
     fi
     
-    if [[ $OVERWRITE_CONFIRMATION != "no" ]]; then
+    if [ $OVERWRITE_CONFIRMATION != "no" ]; then
       # Ask user for confirmation before backing up and overwriting
       echo "${FMT_YELLOW}Found ${zdot}/.zshrc."
       echo "The existing .zshrc will be backed up to .zshrc.pre-oh-my-zsh if overwritten."


tools/install.sh
```diff
@@ -25,9 +25,10 @@
 #   BRANCH  - branch to check out immediately after install (default: master)
 #
 # Other options:
-#   CHSH       - 'no' means the installer will not change the default shell (default: yes)
-#   RUNZSH     - 'no' means the installer will not run zsh after the install (default: yes)
-#   KEEP_ZSHRC - 'yes' means the installer will not replace an existing .zshrc (default: no)
+#   CHSH                   - 'no' means the installer will not change the default shell (default: yes)
+#   RUNZSH                 - 'no' means the installer will not run zsh after the install (default: yes)
+#   KEEP_ZSHRC             - 'yes' means the installer will not replace an existing .zshrc (default: no)
+#   OVERWRITE_CONFIRMATION - 'no' means the installer will not ask for confirmation to overwrite the existing .zshrc (default: yes)
```
mcornella, line 31: Can we use KEEP_ZSHRC instead? We want to make the default to not overwrite zshrc, therefore for unattended we can just turn this variable on.

tools/install.sh
```diff
@@ -342,21 +344,23 @@ setup_zshrc() {
...
-    echo "----------------------------------------"
-    printf '%sDo you want to overwrite it with the Oh My Zsh template? [Y/n]%s ' \
-      "$FMT_YELLOW" "$FMT_RESET"
-    read -r opt
-    case $opt in
-      [Yy]*|"") ;;
-      [Nn]*) echo "Overwrite skipped. Existing .zshrc will be kept."; return ;;
-      *) echo "Invalid choice. Overwrite skipped. Existing .zshrc will be kept."; return ;;
-    esac
+    if [[ $OVERWRITE_CONFIRMATION != "no" ]]; then
```
mcornella, line 347: For the installer we should keep to using only POSIX style conditionals, with single brackets `[`. Some `sh` don't support this syntax.

tools/install.sh
carlosala: Makes sense to have two separate configs, as discussed offline! 😊

PR #13288, opened 2025-09-04
feat(systemd): preserve editor env for sudo commands

iyzana:
## Standards checklist:

## Changes:

- preserve editor env for sudo commands

## Other comments:

When running `systemctl` commands that require sudo, the users' configured editor environment variables were not preserved. This resulted in systemctl opening a default editor.

By instructing sudo to preserve the common environment variables for configuring an editor, systemctl can see these variables and open the users' preferred editor.

This is mostly relevant for the `sc-edit` alias.


mcornella:
Hi there, thank you for the contribution.

After some thought, I belive this introduces some undesired risk of privilege escalation, for example if the user is in an unsafe environment that has access to modify env variables.

We're choosing not to include this, and instead such cases should be modified in configuration files.

Thanks

PR #13298, opened 2025-09-11
doc:Added a Note stating User also needs to install tldr

suJayant:
I think it is a good idea to keep uniformity as many other plugins do mention if a package is need to installed separately as well or not.

## Changes:

- I added a new Note saying that tldr needs to be installed


carlosala:
Thanks for your contribution!

PR #13312, opened 2025-09-19
feat(git): add gstam for git stash push --message

Ricinix:
## Standards checklist:

## Changes:

- add a new alias "gstam" in plugins/git/git.plugin.zsh
- improve plugins/git/README.md for the new alias

## Other comments:

I think `gstam` for `git stash push --message` is the best and short alias.


carlosala:
Hey! I'd rather say this is not needed. `gstam` is not that different from `gsta -m`, just a couple of strokes less.
Thanks for your contribution!

PR #13314, opened 2025-09-19
fix(bgnotify): Use lsappinfo and add support of Ghostty and Alacritty terminals

denysdovhan:
## Standards checklist:

## Changes:

Recently I tried to migrate from iTerm to Ghostty terminal. I noticed `bgnotify` doesn't work for Ghostty at all, so I tried to dig into that problem.

So the issue is `osascript` fails to detect the app id for Ghostty (probably for other apps too) with this error:

```
➜ osascript -e "tell application id \"com.mitchellh.ghostty\" to get the {id, frontmost, id of front window, visible of front window}"
71:73: execution error: Ghostty got an error: Can’t get id of window 1. (-1728)

➜ osascript -e 'tell application id "com.mitchellh.ghostty" to properties of front window'
47:57: execution error: Ghostty got an error: Can’t get every property of window 1. (-1728)

➜ osascript -e 'tell application id "com.mitchellh.ghostty" to get {id, frontmost}'
com.mitchellh.ghostty, true
```

I fixed that by using `lsappinfo` command, and it now works for iTerm, Ghostty, Alacritty and probably more other terminals.

## Other comments:

Works well for me.

In case of Alacritty, `TERM_PROGRAM` is empty string, so `bgnotify_programid` only adds Ghostty.

### DIFF
--- a/plugins/bgnotify/bgnotify.plugin.zsh
+++ b/plugins/bgnotify/bgnotify.plugin.zsh
@@ -60,8 +60,10 @@ function bgnotify_formatted {
 }
 
 function bgnotify_appid {
-  if (( ${+commands[osascript]} )); then
-    osascript -e "tell application id \"$(bgnotify_programid)\"  to get the {id, frontmost, id of front window, visible of front window}" 2>/dev/null
+  if (( ${+commands[lsappinfo]} )); then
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #13315, opened 2025-09-19
fix(plugins/copyfile): don't fail with empty filename

jnooree:
clipcopy interprets empty filename as stdin, but the current "error handling" breaks its behavior.

Relevant code: https://github.com/ohmyzsh/ohmyzsh/blob/9e23925b8581d22033f07f1983587412d3761494/lib/clipboard.zsh#L54-L100

This bug was introduced in ohmyzsh/ohmyzsh#13248.

## Standards checklist:

## Changes:

- Don't fail when copyfile is invoked with empty filename.

## Other comments:

N/A

### DIFF
--- a/plugins/copyfile/copyfile.plugin.zsh
+++ b/plugins/copyfile/copyfile.plugin.zsh
@@ -1,11 +1,12 @@
-# Copies the contents of a given file to the system or X Windows clipboard
+# Copies the contents of a given file or stdin to the system or X Windows
+# clipboard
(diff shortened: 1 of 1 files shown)


carlosala:
Hey! Instead of using copyfile, you can use `clipcopy` directly; you don't need this plugin. The only point of `copyfile` is to prevent you from copying the stdin. Otherwise it'd be just an alias for clipcopy.

PR #13349, opened 2025-10-04
fix(extract): extract to location of archive file (#13348)

### LINKED ISSUE #13348: Extract plugin extracts to current working directory instead of location of archive file
### Describe the bug

If running extract from a different directory than the location of the archive file, the files are extracted to the working directory instead of the location of the archive file.

### Steps to reproduce

1. Enable plugin
2. From user's home directory, extract an archive from a different directory (e.g. `~$ extract ~/Downloads/archive.tar`)
3. Note that files have been created in a folder in the user's home directory rather than in the directory where the archive file was located (e.g. files are extracted to `~/` rather than `~/Downloads/`)

### Expected behavior

Files should be extracted to the same location as the archive file that they came from

_No response_

### OS / Linux distribution

Linux Mint 22.2

### Zsh version

5.9

### Oh My Zsh version

master (d57775d)

### Terminal emulator

GNOME Terminal

None

### Additional context

_No response_

mdpenguin:
Extract to the same directory as the archive file instead of the current working directory

## Standards checklist:

## Changes:

- Set extract_dir variable to use full path of the archive instead of just the filename

## Other comments:
closes #13348
...

### DIFF
--- a/plugins/extract/extract.plugin.zsh
+++ b/plugins/extract/extract.plugin.zsh
@@ -28,7 +28,7 @@ EOF
 
     local success=0
     local file="$1" full_path="${1:A}"
-    local extract_dir="${1:t:r}"
+    local extract_dir="${1:A:r}"
 
     # Remove the .tar extension if the file name is .tar.*
     if [[ $extract_dir =~ '\.tar$' ]]; then


carlosala:
Closing, as stated in https://github.com/ohmyzsh/ohmyzsh/issues/13348#issuecomment-3452872975

Issue #13339, opened 2025-09-29
^[OA input with flutter plugin


Vonarian:
### Describe the bug

When using the flutter plugin: `plugins=(git flutter yarn brew zsh-autosuggestions zsh-syntax-highlighting)`, multiple default keybinds insert key codes instead of showing the result of the keybind, e.g arrow up shows ^[QA instead of the previously used command.
Please note that after pressing arrow up and then enter, the previous command is ran, but terminal shows `^[QA`.

### Steps to reproduce

1. Enable `flutter` plugin
2. Open a fresh terminal
3. Use the arrow keys, tab, etc.
4. You will see `^[OA^[OA^[OA^[OB^[OB^[OC^[OD` depending on the keyboard input.

### Expected behavior

When using some keys, for example arrow keys on a fresh terminal:

- Arrow up to show previous used command
- Arrow left/right to move cursor
- Tab to auto-complete a command, etc.

### OS / Linux distribution

MacOS 26.0

### Zsh version

5.9

### Terminal emulator

iTerm2, Apple Terminal

### If using WSL on Windows, which version of WSL

None

### Additional context

Doing Ctrl + C once solves the behaviour.

bitmorty:
Same here. Zsh 5.9 on Fedora and ghostty.

ashokkumarmuthu:
Hey folks, can someone share more context on this issue. Like to contribute here

sunjayaali:
Yes me too, also missing auto suggestion when opening the terminal for the first time. I need to run any command first before the auto suggestion works again.

PR #13350, opened 2025-10-05
fix(flutter): remove backgrounding from completion script generation (#13339)

Vonarian:
## Standards checklist:

## Changes:

The command was running in the background and causing keyboard input failure (see https://github.com/ohmyzsh/ohmyzsh/issues/13339)

Fixed:
```shell
flutter zsh-completion > "$ZSH_CACHE_DIR/completions/_flutter" 2>/dev/null
```

## Other comments:
closes https://github.com/ohmyzsh/ohmyzsh/issues/13339

### DIFF
--- a/plugins/flutter/flutter.plugin.zsh
+++ b/plugins/flutter/flutter.plugin.zsh
@@ -26,4 +26,4 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_flutter" ]]; then
   _comps[flutter]=_flutter
 fi
 
-flutter zsh-completion >| "$ZSH_CACHE_DIR/completions/_flutter" &|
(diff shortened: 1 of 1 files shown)


mcornella:
Hi @Vonarian, thank you for the PR. Running the command synchronously will add time to the init step, making it slower. 

From the original issue, it looks like `flutter` captures the keyboard for a little bit. Could you test redirect stdin to `/dev/null` instead?

Something like this, on the previous code:
```
< /dev/null
```

Vonarian:
Hi @mcornella , I can confirm your suggestion works and is better than the previous sync one,
```shell
flutter zsh-completion < /dev/null >| "$ZSH_CACHE_DIR/completions/_flutter" &|
```

I made the required change. Thank you for your time and review on this.

PR #13367, opened 2025-10-14
Fix task plugin README title

uthark:


uthark:
@carlosala Could you please take a look? Thank you!

PR #13370, opened 2025-10-17
fix(git): make aliases ggu, ggl, gp, ggfl, ggp more correct

hanskr:
## Standards checklist:

## Changes:

- correct `compdef` functions
- force local variable
- do not reassign variable when falling back (change `:=` to `:-`)

## Other comments:

Current behavior of `ggp` has a bug:

```
# run with a typo
$ ggp --forge
error: unknown option `forge'

# then run correctly after the typo
$ ggp --force
error: unknown option `forge'
```

the typo ends up "cached", because `b` is
1) potentially assigned in the expression `${b:=$1}`, and
2) not declared `local` there

Also, any global value of `b` would be used in `${b:=$1}`, so the bug can be triggered by anything in the environment.

This seems to be the same issue for 5 functions.
All those functions seem to have the wrong function declared for compdef as well.

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -280,10 +280,11 @@ alias gpra='git pull --rebase --autostash'
 alias gprav='git pull --rebase --autostash -v'
 
 function ggu() {
-  [[ "$#" != 1 ]] && local b="$(git_current_branch)"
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #13372, opened 2025-10-18
refactor(jump): improve robustness of getmark function

sanjarcode:
This commit refactors the `getmark` function in the `jump` plugin to improve its robustness and consistency with the rest of the file.

The `LANG=` prefix is added to ensure that the output of the `realpath` command is in a consistent, predictable format. The `command` prefix is used to bypass any potential aliases for `realpath` and `echo`, ensuring that the script executes the intended commands.

### DIFF
--- a/plugins/jump/jump.plugin.zsh
+++ b/plugins/jump/jump.plugin.zsh
@@ -46,7 +46,7 @@ marks() {
 }
 
 function getmark {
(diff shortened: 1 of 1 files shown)


sanjarcode:
tested manually, works

carlosala (approved):
LGTM! 🚀

PR #13373, opened 2025-10-19
Update README.md

tahamhrnd:
| dcx, dcxb, dcxd | Docker Compose utility aliases for quickly (re)starting, rebuilding, and tearing down containers |


carlosala:
I'm afraid these aliases do not exist. Closing.

PR #13374, opened 2025-10-19
Update docker.plugin.zsh

tahamhrnd:
| dcx, dcxb, dcxd | Docker Compose utility aliases for quickly (re)starting, rebuilding, and tearing down containers |


carlosala:
We don't have any aliases for docker compose, and particularly these aliases seem rather specific for a usecase.
Closing.

PR #13389, opened 2025-10-23
ci(deps): ensure push permissions are available

carlosala:
## Standards checklist:

## Changes:

In #13232 we restricted the permissions of the dependencies workflow. Nevertheless, write permissions are required since we are pushing to the repo from `GITHUB_TOKEN`. The GitHub App has no push permissions.
I'm not fully sure if this is the best way to tackle this, maybe we could provide the app with higher permissions, but this option seems more granular.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -39,7 +39,6 @@ jobs:
           GIT_APP_NAME: ohmyzsh[bot]
           GIT_APP_EMAIL: 54982679+ohmyzsh[bot]@users.noreply.github.com
           TMP_DIR: ${{ runner.temp }}
-          DRY_RUN: 1
         run: |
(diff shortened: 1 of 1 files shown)


mcornella:
Extra fixes for another time: 

- We can remove the use of the GitHub app and rely only on the permissions of the the workflow.
- We can sign our commits with a GPG key.

carlosala:
> * We can remove the use of the GitHub app and rely only on the permissions of the the workflow.

I think it is cool to open PRs as ohmyzsh bot, rather than default github actions.

> * We can sign our commits with a GPG key.

I'm not sure there is any way to make GitHub understand that it is a validated signature for that bot, we can investigate.

mcornella (approved):

mcornella:
> I think it is cool to open PRs as ohmyzsh bot, rather than default github actions.

Ah that's true, 👍

PR #13392, opened 2025-10-24
fix(completion): prevent script execution during tab completion

### LINKED ISSUE #13310: Fuzzy matching causes completion candidates to not display properly
### Discussed in https://github.com/ohmyzsh/ohmyzsh/discussions/13301

<div type='discussions-op-text'>

<sup>Originally posted by **rgoldberg** September 13, 2025</sup>
Fuzzy matching causes completion candidates to not display properly using the latest (and versions since at least 1.5 years ago) Oh My Zsh on zsh 5.9 included with macOS 15.6.1 (& with earlier macOS versions) or installed from Homebrew core (I presume that all distributions of zsh 5.9, as well as older versions of zsh,  suffer from the issue).

See <https://github.com/apple/swift-argument-parser/issues/631#issuecomment-3287779901>.

I don't know the exact requirements for the failure, but it seems like, if you try to complete after a single `-`, and if a short & long option designate the same option (it might occur for all shorts & longs being distinct, or for multiple shorts or multiple longs designating the same option, etc.), then the display of the candidates below the command line is visually messed up, either way each long, each short, and each description on its own line, repeated multiple times (I've seen 3 repeats, but maybe the repeat count differs under circumstances).

Candidates seem to be displayed correctly if you try to complete after `--`.

The problem goes away if you disable Oh My Zsh or all Oh My Zsh matchers for completions, e.g. via:

```shell
zstyle ':completion:*' matcher-list ''
```

I've attached a completion script that suffers from this problem: [_mas.txt](https://github.com/user-attachments/files/22309822/_mas.txt) (I had to append `.txt` to attach it).

Is there any way that existing completion scripts can be made compatible with Oh My Zsh fuzzy matching? 

If so, how?

Either way, is there a GitHub issue that tracks fixing this?

Oh My Zsh should be robust enough to not mess up display of valid zsh completion candidates. If the issue cannot be fixed, then should fuzzy matching be removed? Either way, should fuzzy matching be disabled by default until the problem is fixed?</div>

### LINKED ISSUE #13366: pressing `tab` after writing `git` for some reason trigger `bin/test`
### Describe the bug

I'm working on a Rails app and I'm seeing a weird behavior: when I type `git` and then press tab, I can see all the available subcommands but if one of them is partially typed, like `git work` for `worktree`, pressing tab triggers `bin/test` script, and it runs indefinitely 

without anything typed:

Image

with something partially typed:

```
❯ git woRunning RuboCop...
Inspecting 95 files
...............................................................................................

95 files inspected, no offenses detected
Running TS Lint...
```
and it runs forever

### Steps to reproduce

have a `.bin/test` executable script and try autosuggest the rest of a command

### Expected behavior

my script doesn't run

_No response_

### OS / Linux distribution

macOS 26.1 Beta (25B5057f)

### Zsh version

5.9

### Terminal emulator

Ghostty

None

### Additional context

This is nothing new, I've been seeing this behavior for quite some time now but just now got pissed enough to open an issue

My ZSH config is avaialble [here](https://github.com/luizkowalski/dotfiles/blob/master/files/.zshrc)

pfrederiksen:
Fixes #13366 - Tab completion after 'git' triggers bin/test execution 
Fixes #13310 - Duplicate completion candidates

I was able to duplicate this.

The 'l:|=* r:|=*' matcher pattern in completion.zsh was causing:
1. Unintended script execution during tab completion
2. Duplicate completion candidates
3. Overly permissive matching

This pattern allowed left-side anchor matching which could cause zsh to execute scripts like bin/test when searching for completions.

## Standards checklist:

## Changes:

Removed 'l:|=* r:|=*' from all three configuration branches while preserving essential functionality:
- Case-insensitive matching: m:{[:lower:][:upper:]}={[:upper:][:lower:]}
- Right-side fuzzy matching: r:|=*
- Hyphen-insensitive matching: m:{[:lower:][:upper:]-_}={[:upper:][:lower:]_-}

Testing shows that completion continues to work correctly without the security and performance issues.

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -14,13 +14,19 @@ bindkey -M menuselect '^o' accept-and-infer-next-history
 zstyle ':completion:*:*:*:*:*' menu select
 
 # case insensitive (all), partial-word and substring completion
+# NOTE: 'l:|=* r:|=*' matcher removed due to causing unexpected behavior:
+#   - Can trigger unintended execution of scripts during tab completion (issue #13366)
+#   - Causes duplicate completion candidates (issue #13310)
+#   - Creates overly permissive matching that degrades user experience
+# The remaining matchers (case-insensitive + right-side anchor) provide all essential
+# completion functionality without the problematic side effects.
 if [[ "$CASE_SENSITIVE" = true ]]; then
-  zstyle ':completion:*' matcher-list 'r:|=*' 'l:|=* r:|=*'
+  zstyle ':completion:*' matcher-list 'r:|=*'
 else
   if [[ "$HYPHEN_INSENSITIVE" = true ]]; then
-    zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]-_}={[:upper:][:lower:]_-}' 'r:|=*' 'l:|=* r:|=*'
+    zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]-_}={[:upper:][:lower:]_-}' 'r:|=*'
   else
-    zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]}={[:upper:][:lower:]}' 'r:|=*' 'l:|=* r:|=*'
+    zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]}={[:upper:][:lower:]}' 'r:|=*'
   fi
 fi
 unset CASE_SENSITIVE HYPHEN_INSENSITIVE


mcornella:
Fuzzy matching of substrings in the middle and right of completion candidate is a core feature of OMZ. Disabling this doesn't just fix the bug, it turns it off so it doesn't happen.

Before merging this I'd like to understand why it happens, and fix the root cause.

Starting point from this Copilot answer:
```
Short answer
- The matcher "l:|=* r:|=*" creates left‑ and right‑anchored substring matchers that let completion try matches starting at arbitrary offsets inside filenames/words. The completion engine (while building candidate “parts”) used those matchers to construct program/command names that happened to point at a local executable (e.g. bin/test). Those constructed candidates were then passed into the completion driver that invokes completion helper code — which can run the program — so bin/test got executed. Removing the left-side matcher prevents building those spurious candidates.

Key places in zsh source that implement this and how it happens
1) Parsing / installing matchers
- parse_cmatcher / set_gmatcher → compctl.c / complete.c
  - Matcher strings from zstyle/compctl are parsed into internal matcher objects (cmatcher) via parse_cmatcher and stored as a global matcher list (cmatcher / bmatchers).
  - (See: parse_cmatcher and set_gmatcher in Src/Zle/complete.c and Src/Zle/compctl.c.)

2) Adding matchers used when building completion parts
- add_bmatchers / bmatchers list → Src/Zle/compmatch.c
  - The code copies the configured matchers into the bmatchers list used when breaking the input into parts to test for matches.
  - This is where the left/right matcher flags (CMF_LEFT / CMF_RIGHT) and the left/right pattern lengths (llen / ralen) are recorded and later used.

3) Building candidate “parts” using left/right matchers
- bld_parts(...) in Src/Zle/compmatch.c
  - This is the place that actually looks at the bmatchers and tries to match the configured left/right patterns against substrings of the input word:
    - It checks mp->flags for CMF_RIGHT/CMF_LEFT and then calls pattern_match(mp->right, ...) and pattern_match(mp->left, ...) to see whether a candidate match exists at a given offset.
  - When the matcher is effectively “match anything on the left and/or right” (what l:|=* r:|=* does), the condition in bld_parts can succeed for substrings inside filenames/paths — so the engine will build completion candidates from non‑prefix positions in directory entries (for example the filename bin/test).

4) Invoking completion helper code that can execute programs
- runshfunc(...) called inside completion flow → Src/Zle/complete.c and callcompfunc(...) in Src/Zle/compcore.c
  - After building candidates and deciding which completion helper to call, the completion machinery invokes runshfunc(prog, w, name) (complete.c). That code hands off to whichever completion generator is appropriate.
  - If a candidate resolves to a “program” or completion helper that is not purely a safe static list, running that generator can execute external code (shell functions or external programs). callcompfunc and runshfunc are the places that bridge from the completion engine into running completion functions/helpers.
  - In the failure mode caused by the left-side matcher, the constructed candidate string happened to be a path to an executable in the repo (bin/test), so the completion machinery attempted to run it as a helper and thereby executed bin/test.

Why the left-side matcher is the root cause
- The right-side matcher 'r:|=*' alone makes suffix/fuzzy matches (try matches that end at the end of the word) but keeps the matching anchored on one side. The left-side matcher (the 'l:|=*' part) makes the matcher try substrings starting not at the beginning of the word. That extra permissiveness is what allowed the completion engine to treat internal substrings (including a path like bin/test) as valid candidate program names to probe. When the engine probes those candidates it may invoke them as completion helpers, causing execution.

Concrete references (files & functions)
- Src/Zle/complete.c
  - parse_cmatcher / set_gmatcher — parsing matcher-list strings into cmatcher objects
  - call site: runshfunc(prog, w, name) — where the completion generator is invoked
- Src/Zle/compctl.c
  - set_gmatcher / add_bmatchers — placing matchers into the bmatchers list used while building parts
- Src/Zle/compmatch.c
  - bld_parts(...) — uses mp->flags, mp->left/mp->right and calls pattern_match(...) to find matches at offsets; this is the code that allows substring matches when left/right matchers are present
- Src/Zle/compcore.c / Src/Zle/compresult.c
  - do_completion(...) and surrounding code — orchestrates building matches and handing them to completion functions

Summary / practical takeaway
- The left-side matcher allowed completion to construct candidates from substrings (not just prefixes/suffixes). Those candidates included local executable paths (bin/test) which the completion machinery then invoked as completion helpers. The change in the PR removes that left-side matcher so completion no longer builds those spurious candidates and so no longer calls bin/test during tab completion.
```

PR #13398, opened 2025-10-28
feat: add completion for fail2ban-client

makamo:
## Standards checklist:

- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below. *(Not applicable)*

## Changes:

- Added a new plugin `fail2ban` that provides Zsh completion for the `fail2ban-client` command.
- Supports subcommand completion (`start`, `stop`, `status`, `set`, `get`, etc.).
- Dynamically completes active jails.
- Context-aware completion for `set` and `get` options (e.g., `bantime`, `maxretry`, `ignoreip`).
- Automatically uses `sudo` when required.

## Other comments:

This plugin improves usability for users managing Fail2Ban via CLI.  
Tested on Zsh 5.8+ with various jail configurations.  
Let me know if you'd like me to add usage examples to the main README or provide a demo GIF.

### DIFF
--- a/plugins/fail2ban/README.md
+++ b/plugins/fail2ban/README.md
@@ -1,0 +1,37 @@
+# fail2ban-client
+
+## Description  
+
+Zsh completion for the `fail2ban-client` command.  
+This plugin provides:  
+
+- Completion for subcommands (`start`, `restart`, `reload`, `stop`, `unban`, `status`, `set`, `get`, `add`, etc.)
+- Dynamic completion of active jails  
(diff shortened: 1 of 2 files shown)


mcornella:
Hi Maxime, thank you for the contribution. For completion-only plugins, we're switching our approach a little bit.

I'd rather have this built-in for the upstream project, so that all fail2ban users can take advantage of it, or otherwise include it in [zsh-completions](https://github.com/zsh-users/zsh-completions) so that it's available for any zsh user, not only from Oh My Zsh.

I suggest you to propose it on the [main project](https://github.com/fail2ban/fail2ban).

PR #13402, opened 2025-10-30
feat(git): add alias listing latest branches

efiShtain:
add glb function to git plugin which
allows listing the latest X branches (10 by default) make it easy to work with multiple branches
when it is hard remembering the names

## Standards checklist:

## Changes:

- Add a function to the git plugin that allows listing the latest used branches
This is useful in cases where you work on multiple branches and don't remember the names 
making it hard to going between multiple branches

## Other comments:

...

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -268,6 +268,7 @@ receive further support.
 | `grename <old> <new>`    | Renames branch `<old>` to `<new>`, including on the origin remote.                                              |
 | `gbda`                   | Deletes all merged branches                                                                                     |
 | `gbds`                   | Deletes all squash-merged branches (**Note: performance degrades with number of branches**)                     |
+| `glb`                    | List latest switched to branches (default 10, can use glb X to get latest X branches)
 
 ### Work in Progress (WIP)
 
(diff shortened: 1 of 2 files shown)


robbyrussell (changes requested):
@efiShtain Appreciate the contribution here. I'm curious, with this functionality, would you like to see any branches that have since been deleted or if you git checkout a specific commit? (at the moment, this approach would show those in the list of 'branches')

Aside from that, given that this is just listing branches that you recently switched to, how about we do use: `grvb` for "recently visited branches"

efiShtain:
@robbyrussell 
I'm kind of used to glb now, so I'm biased, 
I also prefer shorter options if possible
and I don't see any other conflicts with the current suggestion

but I tend to agree your name is better
so you suggest I'll rename it?

I didn't think the delete option and to be honest I use it most of the times for low number of branches I lately use
so current implementation works really well for that

efiShtain:
@robbyrussell 
hey, just checking in on this one

efiShtain:
@robbyrussell 
Is this relevant from your side? should I close the PR?

PR #13403, opened 2025-10-31
refactor(git): remove unnecessary quotes around `$#` special parameters

artyom-fedosov:
## Standards checklist:

## Changes:

- I replaced all `"$#"` to just `$#`. It is possible because `$#` expands to the number of positional parameters in decimal, so this is guaranteed to be one word. This replacement also a little bit speeds up execution.

## Other comments:

...

### DIFF
--- a/plugins/git/git.plugin.zsh
+++ b/plugins/git/git.plugin.zsh
@@ -102,7 +102,7 @@ function work_in_progress() {
 alias grt='cd "$(git rev-parse --show-toplevel || echo .)"'
 
 function ggpnp() {
-  if [[ "$#" == 0 ]]; then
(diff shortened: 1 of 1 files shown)


mcornella:
Hi Artyom, thank you for the contribution.

In general as a maintainer, things like enforcing a style make more sense when done for the whole codebase, and when able to be automated. We have other instances of this, and also it'd be worth it to change from `==` to `-eq` or other numerical conditionals.

But just this once, I'm going to merge the PR as a `chore`. Thanks

mcornella (approved):

artyom-fedosov:
Hello Marc,

Thank you for your feedback. I will try to apply that you said to the whole project in my next PR.

PR #13421, opened 2025-11-12
fix(alias-finder): use `?` over `{0,1}` to support BSD

### LINKED ISSUE #13416: alias-finder: BSD grep incompatibility with '{0,1} pattern causes errors on macOS
### Describe the bug

The `alias-finder` plugin generates `grep: invalid repetition count(s)` errors on macOS when the `autoload` option is enabled. This
  occurs on EVERY command execution due to BSD grep's incompatibility with the `'{0,1}` regex pattern.

### Steps to reproduce

1. Enable alias-finder plugin in `~/.zshrc`:
     ```bash
     plugins=(alias-finder)
     zstyle ':omz:plugins:alias-finder' autoload yes
  2. Reload shell: source ~/.zshrc
  3. Run any command (e.g., ls, cd /tmp, echo hello)
  4. Observe error appearing before command output:
  grep: invalid repetition count(s)

### Expected behavior

Commands should execute without grep errors.

<img width="967" height="728" alt="Image" src="https://github.com/user-attachments/assets/02661a11-92c7-473d-a613-53988e857e49" />

### OS / Linux distribution

macOS 15.1 (Sequoia) - affects all macOS versions

### Zsh version

### Terminal emulator

iTerm2 / Terminal.app (affects all terminal emulators)

None

### Additional context

### Root cause:

  The plugin uses this pattern in line 2 of alias-finder.plugin.zsh:
  `wordEnd="'{0,1}$"`

  Later combined into:
 ```
 finder="'{0,1}$cmd$wordEnd"
  alias | grep -E "=$finder"

```
  BSD grep 2.6.0 (shipped with macOS since 2010) cannot parse '{0,1} - the leading single quote confuses its regex parser, causing it to
  reject the repetition count {0,1}.

###   Workaround:

  Install GNU grep and ensure it's in PATH before Oh My Zsh loads:

 ```
 # Install GNU grep
  brew install grep
 # Add to ~/.zshrc BEFORE sourcing oh-my-zsh.sh
  export PATH="/opt/homebrew/opt/grep/libexec/gnubin:$PATH"
```


 ###  Proposed solutions:

1.   Recommended: Use BSD-compatible regex pattern:
`wordEnd="'\\?$"`  # Use \? for zero-or-one (equivalent to {0,1})
3.   Alternative: Document GNU grep requirement in README with installation instructions
4.   Existing pattern: Already checks for rg (ripgrep) - could prioritize it over grep when available

  ### Distinction from existing issues:

  This issue is distinct from:
  - #13016 (1-character command errors) - fixed via early return
  - #13039 (256+ character errors) - fixed via ripgrep preference

  Those fixes addressed edge cases but didn't resolve the fundamental BSD grep incompatibility with the wordEnd pattern that affects all
  macOS users running system grep.

  The '{0,1} pattern fails universally on BSD grep regardless of command length, making this a more fundamental compatibility issue.

###   Impact:

  - Affects all macOS users with default system grep
  - Appears on every command when autoload is enabled
  - Has existed since plugin creation (BSD grep unchanged since 2010)

###   grep version:
  grep (BSD grep, GNU compatible) 2.6.0-FreeBSD

carlosala:
Closes #13416

## Standards checklist:

## Changes:

Prefer `?` to `{0,1}`, to ensure compatibility with BSD grep as well.

### DIFF
--- a/plugins/alias-finder/alias-finder.plugin.zsh
+++ b/plugins/alias-finder/alias-finder.plugin.zsh
@@ -1,5 +1,5 @@
 alias-finder() {
-  local cmd=" " exact="" longer="" cheaper="" wordEnd="'{0,1}$" finder="" filter=""
+  local cmd=" " exact="" longer="" cheaper="" wordEnd="'?$" finder="" filter=""
 
   # build command and options
   for c in "$@"; do
@@ -31,7 +31,7 @@ alias-finder() {
 
   # find with alias and grep, removing last word each time until no more words
   while [[ $cmd != "" ]]; do
-    finder="'{0,1}$cmd$wordEnd"
+    finder="'?$cmd$wordEnd"
 
     # make filter to find only shorter results than current cmd
     if [[ $cheaper == true ]]; then


mcornella (commented):
Hi there! We're also using the syntax in line 43, and it's more difficult to replace there.

mcornella:
I'm not able to reproduce, let me revise the root cause.

mcornella:
Closing because it doesn't address the issue, see https://github.com/ohmyzsh/ohmyzsh/issues/13416#issuecomment-3533504371

Issue #13412, opened 2025-11-08
Partial Search + Case Insensitive strange behaviour


JSchoreels:
### Describe the bug

Hello, if I use completion on "search", that folder is not found : `Anki-Search-Stats-Extended`.

If I do it on "Search" it is matching it.

But case insensitivity is still set to true and it does work for basic case like "Anki" matching things like `anki*` or `Anki*`

As a workaround I added

```
zstyle ':completion:*' matcher-list \
  'm:{[:lower:][:upper:]}={[:upper:][:lower:]}' \
  'r:|=*' \
  'l:|=* r:|=*' \
  'r:|?=** m:{[:lower:][:upper:]}={[:upper:][:lower:]}'
```

But it means "sea" also match things like `tickticker_obsidian_cleaner/`,  basically all files iwth a "s.*ea" pattern.

Any idea ?

### Steps to reproduce

mkdir tmp
cd tmp
touch Anki-Search-Stats-Extended/
cd search<TAB>
cd Search<TAB>

### Expected behavior

When case insensitivity is set to TRUE, partial match should partial match with case insensitivity 

### OS / Linux distribution

macos 26.0.1

### Zsh version

5.9

### Oh My Zsh version

master (90a22b6)

### Terminal emulator

iTerm2

### If using WSL on Windows, which version of WSL

None

Issue #13419, opened 2025-11-11
File name completion endless mmap loop: \xc2\xa0something then xyz.som


larry-cdn77:
### Describe the bug

File name autocompletion gets stuck in a very particular scenario

### Steps to reproduce

1. Create an empty file named:

`\xc2\xa0something` (or `\302\240something`)

2. Create a second empty file that comes later alphabetically than the first file and whose name ends in an extension made by the first three letters of ‘something’, eg:

`xyz.som`

3. Try to autocomplete ‘som’, eg:

`ls som<TAB>`

### Expected behavior

Completion

### Screenshots and recordings

Instead, zsh process enters an endless loop doing:

`mmap(NULL, 16384, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0)`

### OS / Linux distribution

Debian GNU/Linux 11 (bullseye)

### Zsh version

5.8

### Terminal emulator

macOS Terminal (remote)

### Additional context

Not sure if actually an Oh My issue but have not recreated it so far in plain Zshell

mcornella:
Hi Larry, you can try to reproduce this with barebones zsh.

To do so, run `zsh -f`, and enable completion with 
```
autoload -Uz compinit && compinit
```
Then try to reproduce the issue, see if the loop still happens.

If it doesn't, it's probable that it may have something to do with matcher-list zstyle variables, check whichever you have set up via
```
zstyle -L | grep matcher-list
```
and run it in the barebones zsh session.

Those are defined in `lib/completion.zsh` btw. If this still doesn't do it, maybe there is another completion related zstyle setting (check `zstyle -L` for anything that's suspicious).

You can also run `functions -t _main_complete` before completing, in your main shell, to get a trace output of what's specifically running in zsh code when it gets to that loop.

larry-cdn77:
Thank you for the reply

I tried `zsh -f` and `autoload -Uz compinit && compinit`: issue does not occur (the completion attempt simply generates a bell)

The output of `zstyle -L | grep matcher-list` is:

```
zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|=*' 'l:|=* r:|=*'
```

And a similar line is where it gets stuck with the `functions -t _main_complete` tracing enabled:

```
...
+_path_files:688> zstyle -t :completion::complete:ls:argument-rest:paths ambiguous
+_path_files:690> [[ som = */* ]]
[...]

Issue #12762, opened 2024-10-15
No completion for dir name with underscore followed by a capital letter


smartinellimarco:
### Describe the bug

When attempting to use tab-completion in Zsh for directory or file names that contain a dash (-) followed by a capital letter (e.g., my-Folder), no suggestions appear when the completion begins on the capital letter itself, but in lowercase.

### Steps to reproduce

Steps to reproduce

mkdir my-Directory
touch my-File
cd dir<TAB>
cat fi<TAB>

<No completions are shown>

### Expected behavior

Should offer completions

### OS / Linux distribution

Macos 14.6.1

### Zsh version

5.9

### Terminal emulator

kitty 0.36.4

### If using WSL on Windows, which version of WSL

None

smartinellimarco:
Forgot to add that this is the setting im using: https://github.com/ohmyzsh/ohmyzsh/blob/067558da9274986d326533daca047bbcb01f2451/lib/completion.zsh#L21

ambuj-1211:
@smartinellimarco is anyone working on this one or can I contribute to it?

smartinellimarco:
@ambuj-1211 Not that I know of!

PR #13436, opened 2025-11-22
feat: add `zstyle` completion matching settings and fuzzy match

mcornella:
## Standards checklist:

## Changes:

- Use `zstyle` setting for configuring `matcher-list`
- Add `fuzzy-match` for allowing fuzzy matching completion (`ad` matches `abcdef`, characters don't need to be consecutive)
- Allow turning off substring matching, which causes #13419

### DIFF
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -13,17 +13,72 @@ setopt always_to_end
 bindkey -M menuselect '^o' accept-and-infer-next-history
 zstyle ':completion:*:*:*:*:*' menu select
 
-# case insensitive (all), partial-word and substring completion
-if [[ "$CASE_SENSITIVE" = true ]]; then
-  zstyle ':completion:*' matcher-list 'r:|=*' 'l:|=* r:|=*'
-else
-  if [[ "$HYPHEN_INSENSITIVE" = true ]]; then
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM! Could you update the docs in the README, and also in the wiki?
Very good improvement!

mcornella:
Let's wait, I want to incorporate a couple other improvements about matcher-list. Particularly:
- [ ] #13412
- [ ] #13444
- [ ] #13392
- [ ] #12762 
- [ ] Add docs in both wiki and README

robbyrussell:
@mcornella — heads up on a collision. #13823 fixes #13412 and #12762 by folding the `m:` case rule into each partial-word spec; this PR keeps `m:` as its own spec, so it'd carry the same bug forward and conflict with the fix.

Land #13823 first and rebase this on it, or fold the composition in here? Either's fine — just flagging so one doesn't quietly undo the other.

mcornella:
I'd manually confirm #13823 fixes what it says it fixes first, then rebase this one on top. 

Since you're looking into it, I'd also like to see:
- an automated testing script done for the fixes (i.e. some script that inputs the characters and confirms the completion returns what we expect based on the matches)
- that issues and PRs in comment https://github.com/ohmyzsh/ohmyzsh/pull/13436#issuecomment-3580355064 including closed ones are also reviewed

PR #13451, opened 2025-11-28
feat(clipboard): tmux: send clipboard to client when using clipcopy

bretello:
## Standards checklist:

## Changes:

By adding `-w`, we can get `load-buffer` (and `clipcopy`) to also send the copied buffer to the client, allowing for easy copying tmux buffers into the local client, which is useful in scenarios such as ssh+tmux remote sessions.

From [`man tmux`](https://www.man7.org/linux/man-pages/man1/tmux.1.html):

```man
load-buffer [-w] [-b buffer-name] [-t target-client] path
             (alias: loadb)
       Load the contents of the specified paste buffer from path.  If -w is given, the buffer is also sent to the clipboard
       for target-client using the xterm(1) escape sequence, if possible.  If path is ‘-’, the contents are read from stdin.
```


tobiasbaehr:
LGTM

carlosala:
Thanks for your contribution!

PR #13478, opened 2025-12-11
fix(git): git_status_prompt respects spaces to properly match prefixes

### LINKED ISSUE #10909: git_prompt_status always returns empty in theme
### Describe the bug

I try to make my own theme and currently ran into a weird problem. 

My goal: To have the git branch in my prompt in a color based on the current status of the repository (modified, added, ...).
I use `$(git_prompt_status)` to get the current status and a case statement with regex to look up whats currently changed.

The problem: The case statement will never return anything different than the "default action" (`*)`) because `$(git_prompt_status)` always returns an empty string. 

Even weirder: It does not work when used in my theme file. But when using the theme and calling `my_git_color` in bash it will always show the correct output (the correct color string)!

Here's the full theme:
```bash
ZSH_THEME_GIT_PROMPT_ADDED="A"
ZSH_THEME_GIT_PROMPT_MODIFIED="M"
ZSH_THEME_GIT_PROMPT_DELETED="M"
ZSH_THEME_GIT_PROMPT_RENAMED="M"
ZSH_THEME_GIT_PROMPT_UNMERGED=""
ZSH_THEME_GIT_PROMPT_UNTRACKED="U"

function my_git_color() {
    local state="$(git_prompt_status)"
    case "$state" in
        *"A"*) echo -n "%B%F{green}";;
        *"M"*) echo -n "%B%F{yellow}";;
        *"U"*) echo -n "%F{141}";;
        *) echo -n "%F{red}";;
    esac
}

ZSH_THEME_GIT_PROMPT_PREFIX="%F{blue} [$(my_git_color)"
ZSH_THEME_GIT_PROMPT_SUFFIX="%b%F{blue}]%f "
ZSH_THEME_GIT_PROMPT_DIRTY=""
ZSH_THEME_GIT_PROMPT_CLEAN=""
ZSH_THEME_VIRTUALENV_PREFIX="%F{white}[%B%F{green}"
ZSH_THEME_VIRTUALENV_SUFFIX="%b%F{white}]%f "

PROMPT_CHAR="%B%F{grey}\$%b%f"
PROMPT='%B%(?:%F{green}→:%F{red}→)%b%f %F{cyan}%c%f $PROMPT_CHAR%f '
RPROMPT='$(virtualenv_prompt_info)$(git_prompt_info)'
```
You can safely ignore the virtualenv lines. The error already was there before I added them.

### Steps to reproduce

1. Use the theme above
2. `cd` into a local git repository with changes
3. See that the color is always red, no matter what you do
4. Call my_git_color to see it _should_ work

### Expected behavior

The git branch should have different colors based on the current changes in the repository

_No response_

### OS / Linux distribution

Windows 11, Ubuntu WSL

### Zsh version

5.8

### Oh My Zsh version

master (8f56a8bd)

### Terminal emulator

Windows Terminal Preview

WSL2

### Additional context

_No response_

tDwtp:
## Standards checklist:

## Changes:

`git_status_prompt` was always empty as whitespace around the keys for `$prefix_constant_map` were trimmed. The in list of the for loop must be enclosed in `"` to avoid trimming, which in return requires the iterator specifier to contain an `@` to avoid concatenation of all keys. This is important for the regex to work properly.

## Other comments:

fixes #10909

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -117,7 +117,7 @@ function _omz_git_prompt_status() {
   fi
 
   # For each status prefix, do a regex comparison
-  for status_prefix in ${(k)prefix_constant_map}; do
+  for status_prefix in "${(@k)prefix_constant_map}"; do
     local status_constant="${prefix_constant_map[$status_prefix]}"
     local status_regex=$'(^|\n)'"$status_prefix"
 


carlosala:
Thanks for your contribution, very good catch!

XusanDev07:
good

PR #13479, opened 2025-12-13
feat(tmux): add 'to' alias for session attachment

danrneal:
## Standards checklist:

## Changes:

- Added the to (Tmux Open) alias to the tmux plugin.
- Added documentation for to in the plugin's README.
- Hooked to into the existing _tmux completion system.

### Use Case / Justification:
The tmux plugin currently provides ta (attach) and ts (new-session), but both are "brittle" in common workflows: ta fails if a session doesn't exist, and ts fails if it already does.

The to (Tmux Open) alias introduces an idempotent "attach-or-create" workflow using the tmux new-session -A -s command.

### Workflow Example:
A user working across multiple contexts (e.g., main, logs, deploy) can use to <name> to instantly jump into that specific context without needing to check if the session is already running.

- `to main`: Attaches to main if it exists, or creates it if it doesn't.
- `to`: Quickly creates or attaches to a default anonymous session.

This reduces cognitive load and command failures during fast context switching, making it a high-value addition for the majority of tmux users. By using the `_build_tmux_alias` helper, the alias also benefits from full session-name tab completion.

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -81,6 +81,7 @@ alias tmuxconf='$EDITOR $ZSH_TMUX_CONFIG'
 
 _build_tmux_alias "ta" "attach" "-t"
 _build_tmux_alias "tad" "attach -d" "-t"
+_build_tmux_alias "to" "new-session -A" "-s"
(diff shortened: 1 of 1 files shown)


carlosala:
Thanks for your contribution!

PR #13484, opened 2025-12-16
feat(terraform): Added terraform apply and destroy aliases with parallelism=1 for certain Terraform providers

leggatron:
## Standards checklist:

## Changes:

- Added two new aliases for the Terraform plugin
- - `tfap = terraform apply parallelism=1`
- - `tfdp = terraform destroy parallelism=1`

These are useful because some Terraform providers are hitting APIs that have checks in place to prevent DDoS attacks or to generally try to mitigate for higher-than-normal request levels. In these cases, you'll often want to set the parallelism all the way down to 1 to ensure that each resource being created happens one-at-a-time. 

For this reason, I use this command on a daily basis. I can see how others might find it useful to have as an option with this plug in. 

## Other comments:

I found no other PRs with this change and no issues referencing this change. If I have missed anything, please let me know and I will amend or retract if necessary.


mcornella (changes requested):
It's unfortunate that this option is not settable at the provider level, so this looks like the best option against an env variable. 

Only one nitpick: could you order the aliases in their alphabetic order (below apply and destroy aliases)?

Thanks!

leggatron:
No problem at all - change has been made. 

Thanks!

PR #13500, opened 2026-01-04
fix minimal theme

amnay-mo:
The `minimal` theme was broken as it didn't show git branch info in the prompt

This is an attempt to fix it, I can confirm that it works on my machine :)


carlosala:
Hey! Instead, you can set up your git prompt to appear. Use the `force` section. https://github.com/ohmyzsh/ohmyzsh#async-git-prompt

Thanks for your contribution!

PR #13512, opened 2026-01-11
Update social media link from X to Bluesky

mattkime:
Simply updating the social media link in update complete message to the network that ohmyzsh is actively using.

## Standards checklist:

- [x] The code is mine.


carlosala:
We're actually using more X at this point, but having both sounds good! Could you update the PR to fit both?
I'd leave X first, since it's the most active one.

mattkime:
@carlosala Haha, I saw a few old posts near the top on X and didn't look any closer. Guess its another algorithm "optimization"

Anyway, it should be addressed

robbyrussell:
@mattkime Oops, I forgot that we have links to X in multiple places. Can you also update the `README` and `install.sh` scripts?

PR #13536, opened 2026-01-22
Update functions.zsh

Jemose:
Explicitly state wrapper function is redirecting to 'omz update' and not running any deprecated code.


carlosala:
I think the wording is good as it is!
Thanks for your contribution though!

PR #13556, opened 2026-02-08
fix(appearance): ensure arg separation in `test-ls-args`

rangaNP:
## Standards checklist:

## Changes:

- Use `shift` + `"$@"` instead of `"$args"` in `test-ls-args` to pass flags separately. 
- Code quality improvement, no functional change.

## Other comments:

Issue spotted with GitHub Copilot.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -40,9 +40,9 @@ if [[ -z "$LS_COLORS" ]]; then
 fi
 
 function test-ls-args {
-  local cmd="$1"  # ls, gls, colorls, ...
-  shift            # remove the command, leaving only the arguments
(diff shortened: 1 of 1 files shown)


mcornella:
Well spotted, a case of trying to be too clever. I simplified it to its simplest form, since we're not changing the argument order. Because of how we used it we never triggered the bug, so I'll merge this as a refactor. Thanks for the contribution!

rangaNP:
> Well spotted, a case of trying to be too clever. I simplified it to its simplest form, since we're not changing the argument order. Because of how we used it we never triggered the bug, so I'll merge this as a refactor. Thanks for the contribution!

Thanks for the review and appreciate the simplification and the feedback.

PR #13566, opened 2026-02-14
docs(jj): clarify git async-prompt limitation with wrapper functions

### LINKED ISSUE #13555: jj prompt override doesn't work when using git async-prompt
### Describe the bug

The Oh-My-Zsh [jj plugin installation doc](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/jj) recommends this override to add a jj status to the prompt:

```
_my_theme_vcs_info() {
  jj_prompt_template 'self.change_id().shortest(3)' \
  || git_prompt_info
}

PROMPT='$(_my_theme_vcs_info) $'
```

However, this override breaks when git async prompt is enabled, which is the default (that is, `zstyle ':omz:alpha:lib:git' async-prompt` is explicitly enabled or empty).

In that case, `git_prompt_info` just stops working, and returns nothing.

## Note

This is also the case when simply adding an extra level of indirection around `git_prompt_info` :

```
_my_git_prompt_info() {
  git_prompt_info
}

PROMPT='$(_my_git_prompt_info) $'
```

In this case, when async-prompt is enabled, `git_prompt_info` stops working, and returns nothing.

### Steps to reproduce

1. Edit the robbyrussel theme,
2. Replace `PROMPT+=' $(git_prompt_info)'` by:
    ```
    _my_git_prompt_info() {
      git_prompt_info
    }
    PROMPT+=' $(_my_git_prompt_info)'
    ```
3. Notice how the git info is no longer displayed in a git directory.

### Expected behavior

1. The `git_prompt_info` should work even in a wrapper function,
2. OR the jj-plugin installation doc should mention the need to disable git async-mode.

_No response_

### OS / Linux distribution

macOS 15.7.3

### Zsh version

5.9

### Oh My Zsh version

master (9df4ea0)

### Terminal emulator

iTerm2

None

## Workaround

Manually disabling the git async prompt makes `git_prompt_info` work again – even in a wrapper function:

```
zstyle ':omz:alpha:lib:git' async-prompt false
```

aviu16:
## Summary

This PR adds documentation to clarify a known limitation when using the recommended jj/git wrapper function pattern with Oh My Zsh's async git prompt feature.

## Issue

When wrapping `git_prompt_info` in a custom function (as shown in the jj plugin README for jj/git fallback), the git async prompt feature (enabled by default on Zsh ≥5.0.6) fails to detect the usage and doesn't register async handlers. This causes `git_prompt_info` to return nothing.

The async detection code in `lib/git.zsh` (lines 165-175) uses string pattern matching to detect `$(git_prompt_info)` or backticks directly in prompt variables. When wrapped like `$(_my_theme_vcs_info)`, the pattern doesn't match.

## Changes

Added a note to the jj plugin README explaining:
- The limitation when using wrapper functions with async prompt
- Two workarounds:
  - Disable git async prompt: `zstyle ':omz:alpha:lib:git' async-prompt no`
  - Force async handlers to always register: `zstyle ':omz:alpha:lib:git' async-prompt force`
- Link to the issue for additional context

## Testing

- Verified the markdown renders correctly
- Confirmed the workarounds are accurate based on the implementation in `lib/git.zsh`

Fixes #13555

### DIFF
--- a/plugins/jj/README.md
+++ b/plugins/jj/README.md
@@ -77,6 +77,15 @@ _my_theme_vcs_info() {
 PROMPT='$(_my_theme_vcs_info) $'
 ```
 
+> **Note:** When using a wrapper function around `git_prompt_info`, the git async prompt feature
+> (enabled by default on Zsh ≥5.0.6) will not work because it only detects the literal pattern
+> `$(git_prompt_info)` in prompt variables. To fix this, either:
+>
+> - Disable git async prompt: `zstyle ':omz:alpha:lib:git' async-prompt no`
+> - Force async handlers to always register: `zstyle ':omz:alpha:lib:git' async-prompt force`
+>
+> See [issue #13555](https://github.com/ohmyzsh/ohmyzsh/issues/13555) for details.
+
 You can find an example
 [here](https://github.com/nasso/omzsh/blob/e439e494f22f4fd4ef1b6cb64626255f4b341c1b/themes/sunakayu.zsh-theme).
 


aviu16:
closing this for now, gonna revisit later

PR #13567, opened 2026-02-15
Fix unquoted parameter expansion in zsh-navigation-tools

aviu16:
the terminfo values in n-list-draw werent quoted, so if they contained spaces or special chars the conditionals could break. just added double quotes around them


carlosala:
Thanks!

PR #13568, opened 2026-02-15
Fix unquoted password variable in rbw plugin

aviu16:
the password variable in the rbw plugin wasnt quoted, so passwords with spaces would get word split. just added quotes around it


aviu16:
closing this for now, gonna revisit later

carlosala:
`echo` has no problems with multiple arguments.

PR #13573, opened 2026-02-16
Fix unquoted variable expansions in pow plugin

aviu16:
## Summary
Quote variable expansions in pow plugin to prevent word splitting and globbing issues.

## Changes
- Line 35: `echo $basedir` → `echo "$basedir"`
- Line 40: `basename $basedir` → `basename "$basedir"`

## Why
Unquoted variables in shell can cause word splitting and globbing. If `basedir` contains spaces or special characters, the unquoted form would break.

## Test Plan
Existing functionality preserved - commands work correctly with paths containing spaces.

### DIFF
--- a/plugins/pow/pow.plugin.zsh
+++ b/plugins/pow/pow.plugin.zsh
@@ -32,12 +32,12 @@ rack_root(){
 
   builtin cd "$orgdir" 2>/dev/null
   [[ ${basedir} == "/" ]] && return 1
-  echo $basedir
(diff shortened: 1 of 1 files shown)


aviu16:
Friendly bump on this PR. When maintainers have a moment, could someone please take a look? Happy to make any requested changes quickly. Thanks!

carlosala:
Closing! It has a submodule included, feel free to reopen without it.

PR #13580, opened 2026-02-17
feat(update): default to auto update without prompt

HaraldNordgren:
Seems like most people would benefit from having this on by default. This would make ohmyzsh more zero-config.

I have been running like this for a while and it simplifies my life:

```bash
DISABLE_UPDATE_PROMPT=true
```

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -6,17 +6,15 @@ fi
 # Get user's update preferences
 #
 # Supported update modes:
-# - prompt (default): the user is asked before updating when it's time to update
-# - auto: the update is performed automatically when it's time
+# - auto (default): the update is performed automatically when it's time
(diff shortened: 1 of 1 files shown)


robbyrussell:
Appreciate the pitch for this, but I've intentionally set it to require a prompt so people know this is about to happen and can consent... especially given the uptick in supply chain vulnerabilities.

What I might be open to is having another option during the auto-update prompt that says, "(a) enable auto-upgrades" or something along those lines that then points people to how they can enable that in their config file, but I would prefer it to be an intentional decision they know they enabled and are likely to disable it. 

Otherwise, I worry about a) security and b) people complaining about slower boot times without knowing it's because of some extra activity happening behind the scenes.

HaraldNordgren:
> especially given the uptick in supply chain vulnerabilities.

Yeah that's fair!

PR #13588, opened 2026-02-21
feat(containerenv): New plugin for /run/.containerenv interpretation (podman & docker)

james-nesbitt:
## Standards checklist:

## Changes:

* new containerenv plugin, based off of the toolbox plugin
  * interpret the /run/.containerenv file for parametrization
  * add a prompt info function that can be used to put container info into the prompt
  * no new aliases are added.

## Other comments:

* coded with help using cursor.

I used to use an os-blue OS that relied on toolbox, and enjoyed the toolbox rprompt info; I recently switched to a different OS and now use distrobox, and wished to have the same prompt information.
I checked the toolbox code added in https://github.com/ohmyzsh/ohmyzsh/pull/10685, and saw that it uses `/run/.containerenv` for its source. 

The `/run/.containerenv`  file will be created by many different container runtimes, under various scenarios (docker and podman primarily.) Because the information source is generic, it didn't make sense to add the interpretation to any existing plugin (such as podman), so I made a new plugin.

I've checked the docker and podman container runtimes to see what values they put into the file, and there is a lot of consistency, so the standardized functions (_id, _name etc) made sense. The interpretation is fault tolerant, so it doesn't cause issues.

The interpretation of the file contents runs a grep exec for every function call.  This is not the most efficient approach if a prompt is using many of the function calls, but it shouldn't really affect the performance. There was initially a sed exec that was dropped for a pair of string interpretations.

This plugin adds no new aliases, but does add functions to interpret the contents of the file in a standardized way.  The plugin could replace the toolbox plugin functions, but I did not include any removals in this PR, as that could break existing usage.

The icons used in the prompt info might not be the best choice, and perhaps should be dropped.  Let me know.

### DIFF
--- a/plugins/containerenv/README.md
+++ b/plugins/containerenv/README.md
@@ -1,0 +1,82 @@
+# containerenv plugin
+
+A lightweight plugin that provides convenient functions to access container environment variables from `/run/.containerenv`. Useful in containerized environments (Podman, Docker, Toolbox, etc.).
+
+To use it, add `containerenv` to the plugins array in your `.zshrc` file:
+
+```zsh
+plugins=(... containerenv)
+```
+
+## Usage
+
+### Accessor functions
+
+Each function reads one key from `/run/.containerenv` and prints its value to stdout. On error (not in a container or missing key), a message is printed to stderr and the function returns 1.
+
+| Function | Key | Description |
+|----------|-----|-------------|
+| `containerenv_engine` | `engine` | Container runtime (e.g. `podman`, `docker`) |
+| `containerenv_name` | `name` | Container name |
+| `containerenv_id` | `id` | Container ID |
+| `containerenv_image` | `image` | Image name/reference |
+| `containerenv_imageid` | `imageid` | Image ID |
+| `containerenv_rootless` | `rootless` | Whether the container is rootless |
+| `containerenv_graphrootmounted` | `graphRootMounted` | Whether the graph root is mounted |
+
+**Examples:**
+
+```bash
+# Show container name
+containerenv_name
+
+# Use in a conditional or variable
+if name=$(containerenv_name 2>/dev/null); then
+  echo "Container: $name"
+fi
+```
+
+### Show all container env vars
+
+```bash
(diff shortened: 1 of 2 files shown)


james-nesbitt:
Hmmm,  in a busybox env (wolfi) I get the following issue:

```➜  ~ containerenv_name
grep: unrecognized option '--exclude-dir=.bzr'
BusyBox v1.37.0 (2026-02-02 03:30:07 UTC) multi-call binary.

Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A|B|C N] { PATTERN | -e PATTERN... | -f FILE... } [FILE]...

Search for PATTERN in FILEs (or stdin)

	-H	Add 'filename:' prefix
	-h	Do not add 'filename:' prefix
	-n	Add 'line_no:' prefix
	-l	Show only names of files that match
	-L	Show only names of files that don't match
	-c	Show only count of matching lines
	-o	Show only the matching part of line
	-q	Quiet. Return 0 if PATTERN is found, 1 otherwise
	-v	Select non-matching lines
	-s	Suppress open and read errors
	-r	Recurse
	-R	Recurse and dereference symlinks
	-i	Ignore case
	-w	Match whole words only
	-x	Match whole lines only
	-F	PATTERN is a literal (not regexp)
	-E	PATTERN is an extended regexp
	-m N	Match up to N times per file
	-A N	Print N lines of trailing context
	-B N	Print N lines of leading context
	-C N	Same as '-A N -B N'
	-e PTRN	Pattern to match
	-f FILE	Read pattern from file
Error: Key 'name' not found in /run/.containerenv
```

The function call fails, and no name is returned. There is a `run/.containerenv` file, and it does contain a "name" key.

I believe that this issue comes from the zsh grep lib, and not from the new code. This doesn't mean that the new code should not be considered a part of the problem.

james-nesbitt:
I pushed a small change which standardizes the rootless detection.

I force pushed an amended change. Let me know if this is an unacceptable approach.  It is a force of habit from my gerrit days.

Issue #11705, opened 2023-05-22
vi-mode visual mode cursor


Subjective:
### Describe the bug

When switching out of visual mode, the cursor does not change back to a block.

### Steps to reproduce

Minimal config:
```zsh
# .zshrc
...
plugins+=(vi-mode)

VI_MODE_SET_CURSOR=true
...
```
1. Enable the `vi-mode` plugin: `plugins+=(vi-mode)`
2. Set `VI_MODE_SET_CURSOR=true` in `.zshrc`
3. Enter visual select and hit escape
4. The cursor doesn't change back to a block as it should in normal mode

### Expected behavior

I would expect the cursor to remain a block when entering visual mode (or exiting visual mode) instead of switching into a vertical line. 

### Screenshots and recordings

![Screen Recording 2023-05-21 at 6 05 46 PM](https://github.com/ohmyzsh/ohmyzsh/assets/56745535/7bd0af0a-6172-45e8-9282-b0cf6c63dec3)

### OS / Linux distribution

macOS 13.3

### Zsh version

5.9

### Oh My Zsh version

master (b06663d)

### Terminal emulator

kitty

### If using WSL on Windows, which version of WSL

None

### Additional context

This seems to have been introduced by the fix for #11586 as removing the following lines resolves the issue:
```zsh
function _visual-mode {
  typeset -g VI_KEYMAP=visual
  _vi-mode-set-cursor-shape-for-keymap "$VI_KEYMAP"
  zle .visual-mode
}
zle -N visual-mode _visual-mode
```

okapia:
If you want to detect visual mode reliably, use the `zle-line-pre-redraw` hook and test for `[[ $KEYMAP = vicmd ]] && (( REGION_ACTIVE ))`. That maps to how zsh defined visual mode internally. Wrapping widgets should be a last resort because it tends to lose the effect of widget flags. That #11586 "fix" misses the `visual-line-mode` widget. And a user might use an emacs region widget or a custom widget. And it only detects entering visual-mode, not leaving it.

The cursor is a bit of a tricky issue with visual mode. If the cursor is done in reverse video and reverse is used for the selection they cancel each other out. Otherwise, a block cursor works best for vi-command mode including visual mode. But this user has configured a vertical bar for visual mode.

If you really want to wrap the `visual-mode` widget, consider making it conditional depending on whether the cursor is even configured to be different.

[...]

PR #13597, opened 2026-02-24
fix(vi-mode): re-render cursor after exisiting visual mode (ohmyzsh#11705)

KevinWMatthews:
Redraw the cursor after exiting visual or visual line mode.

Fixes 11705.

## Standards checklist:

_I didn't see a wiki or a style guide but did try follow this [style indication](https://github.com/KevinWMatthews/ohmyzsh/blob/master/CONTRIBUTING.md#style) in `CONTRIBUTING.md`_

## Changes:

Following [this suggestion](https://github.com/ohmyzsh/ohmyzsh/issues/11705#issuecomment-1557342576) from #11705:
- remove the `visual-mode` widget
- add a `zli-line-pre-redraw` hook ([special widget](https://zsh.sourceforge.io/Doc/Release/Zsh-Line-Editor.html#Special-Widgets))

This hook checks the highlight state, and if necessary updates the local understanding of the keymap and changes the cursor shape.

## Other comments:

This contribution is AI-assisted, and I take full responsibility for the code in this PR. (To be honest, I hadn't planned on learning anything about zsh but it with some help it isn't so bad :) ).

My current understanding is that zsh doesn't maintain a separate keycode for vi's visual and visual-line modes, but instead highlights part of the text. The `vi-mode` plugin emulates these vi modes by checking current highlight state and the current key.

The [REGION_ACTIVE](https://zsh.sourceforge.io/Doc/Release/Zsh-Line-Editor.html#index-REGION_005fACTIVE) variable indicates which characters/regions are to be highlighted, and it has values of:
- 0: no highlight
- 1: character/region highlight active
- 2: line highlight active

Note that the rprompt is not modified to indicate when visual mode is active. I think this could be a useful feature but is outside the scope of this PR. I plan to create a separate issue and post a follow-up PR for this. (edit: see #13625 and #13626).

### DIFF
--- a/plugins/vi-mode/vi-mode.plugin.zsh
+++ b/plugins/vi-mode/vi-mode.plugin.zsh
@@ -31,24 +31,32 @@ function _vi-mode-set-cursor-shape-for-keymap() {
   # https://vt100.net/docs/vt510-rm/DECSCUSR
   local _shape=0
   case "${1:-${VI_KEYMAP:-main}}" in
-    main)    _shape=$VI_MODE_CURSOR_INSERT ;; # vi insert: line
-    viins)   _shape=$VI_MODE_CURSOR_INSERT ;; # vi insert: line
-    isearch) _shape=$VI_MODE_CURSOR_INSERT ;; # inc search: line
-    command) _shape=$VI_MODE_CURSOR_INSERT ;; # read a command name
-    vicmd)   _shape=$VI_MODE_CURSOR_NORMAL ;; # vi cmd: block
-    visual)  _shape=$VI_MODE_CURSOR_VISUAL ;; # vi visual mode: block
-    viopp)   _shape=$VI_MODE_CURSOR_OPPEND ;; # vi operation pending: blinking block
-    *)       _shape=0 ;;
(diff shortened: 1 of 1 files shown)


KevinWMatthews:
Ah, I forgot to put "fixes" in the commit message. Fixed via a force-push.

robbyrussell (changes requested):
Sorry this sat for a while. The redraw approach fixes both visual and visual-line cursor transitions on zsh 5.9, but I found three things we need to fix before merging:

1. Please register a prefixed callback through `add-zle-hook-widget` instead of replacing `zle-line-pre-redraw` with `zle -N`. As written, this overwrites an existing redraw-hook owner, so plugins such as `history-substring-search` stop receiving redraws when they load first.
2. Please keep the current `visual-mode` wrapper as a fallback before zsh 5.3. Oh My Zsh still supports older zsh versions, where `zle-line-pre-redraw` is never invoked.
3. Please limit the `REGION_ACTIVE` checks to the `vicmd` keymap. The parameter also represents emacs and custom regions, which this currently records as vi visual mode.

The core fix works in my zsh 5.9 tests, so I think this is the right direction. Happy to take another look once those compatibility cases are covered.

PR #13604, opened 2026-03-02
feat(ruff): add ruff plugin

gorewilliams:
## Standards checklist:

## Changes:

- Adds ruff plugin (both autocompletions and aliases)

## Other comments:

- I used Claude Code in making this PR
- Solves #13607

### DIFF
--- a/plugins/ruff/README.md
+++ b/plugins/ruff/README.md
@@ -1,0 +1,22 @@
+# ruff plugin
+
+This plugin automatically installs [ruff](https://github.com/astral-sh/ruff)'s completions for you,
+and keeps them up to date. It also adds convenient aliases for common usage.
+
+To use it, add `ruff` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... ruff)
+```
+
+## Aliases
+
+| Alias | Command              | Description                                                  |
+| :---- | -------------------- | :----------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


robbyrussell (changes requested):
Thanks for this — ruff is a good fit, and the plugin structure is right (guard
first, `_comps` registration, README table).

Two fixes before it can merge:

1. **`ruup='ruff self update'` isn't a real command.** Ruff has no `self`
   subcommand — it gives `error: unrecognized subcommand 'self'`. I think it came
   across from the `uv` plugin, where `uv self update` does exist. Please drop the
   alias and its README row.

2. **Please rebase and use the atomic completion write.** We changed this repo-wide
   in #14000: the `>|` redirect empties `_ruff` at startup and refills it in the
   background, so a shell starting in that window gets no completions. Ruff's file
   is ~685 KB, so the window is noticeable.

   ```zsh
   zmodload -F zsh/files b:zf_mv
   () {
     local TMPPREFIX="$ZSH_CACHE_DIR/completions/_ruff"
     zf_mv -f -- =( ruff generate-shell-completion zsh ) "$TMPPREFIX"
   } &|
   ```

   `plugins/uv/uv.plugin.zsh` on master is the closest example.

Ping me when it's updated and I'll take another look.

gorewilliams:
Done @robbyrussell

robbyrussell (approved):

PR #13609, opened 2026-03-03
feat(git): add `gan` and `ganpa` aliases for `git add --intent-to-add`

cdunda-perchwell:
## Description

Adds two new aliases to the git plugin:

- `gan` → `git add --intent-to-add` — marks untracked files as intent-to-add so they can be used with `git diff` and `git add --patch`
- `ganpa` → `git add --intent-to-add && git add --patch` — convenience combo that marks all untracked files then enters interactive patch mode

## Motivation

`git add --patch` (`gapa`) skips untracked files entirely. The standard workaround is to first run `git add --intent-to-add` (or `-N`) to register them as empty tracked files, then `git add --patch` shows their full content as new hunks.

This is a common workflow for developers who review every change before staging. Having `gan` as a composable primitive (e.g. `gan . && gapa`) and `ganpa` as a one-shot combo covers both use cases.

## Prior art

Relates to #6871 (closed) and #6901 (open, stale 4+ years). This PR takes a simpler approach: `gan` is a direct alias for the flag (composable), and `ganpa` is the combined workflow.


robbyrussell:
Thanks for this! One thing to fix before it goes in.

`ganpa` needs a pathspec. Without one, `git add --intent-to-add` doesn't do anything, so the alias ends up being just a slower `gapa`:

```console
$ git status --porcelain
?? untracked.txt

$ git add --intent-to-add
Nothing specified, nothing added.
hint: Maybe you wanted to say 'git add .'?

$ git status --porcelain
?? untracked.txt          # nothing happened

$ git add --intent-to-add .
$ git status --porcelain
 A untracked.txt          # this is what you want
```

So `ganpa` should be:

```zsh
alias ganpa='git add --intent-to-add . && git add --patch'
```

`gan` on its own is fine as-is, since you type paths after it.

For what it's worth, there was an older PR for this same alias (#6901, from 2018) that had the `.` in it. I've closed that one in favor of yours... just want to make sure the working version is the one we keep.

robbyrussell:
Let's also credit [karlhorky](https://github.com/karlhorky) on this

PR #13610, opened 2026-03-05
feat: allow backspace correction in update prompt

rlaope:
## Standards checklist:

## Changes:

- Changed `read -r -k 1` to `read -r` in the update prompt so that users can correct their input with backspace before pressing Enter, instead of the input being accepted immediately after a single keypress.

## Other comments:

Previously, when prompted for an update (`[Y/n]`), `read -r -k 1` captured input immediately after one character with no chance to correct a mistake. By switching to `read -r`, users can now type, use backspace to fix typos, and
confirm with Enter. The default behavior (pressing Enter without input) still triggers the update.

No.. 🥺🥺
image


carlosala:
The idea behind that prompt is to be the least intrusive possible. Adding an extra type is overkill in my opinion, especially for something that is not hazardous or anything like that.
Thanks for your contribution!

PR #13615, opened 2026-03-05
FIX[#13616]: tools/update.sh -- remove redundant variable and local declaration

berniegreen:
# NIT: tools/update.sh -- fix minor redundant variable declaration

Issue #13616 

## Standards checklist:

## Changes:

- Removed `local` keyword from `ret=0` declaration at line 4 of `tools/upgrade.sh`. The `local` builtin is only valid inside a function — using it at script (top-level) scope is incorrect. `zsh` silently accepts it as an extension, which masked the bug; running the script with `sh` exposed the error.
- Removed a duplicate `local ret=0` declaration at line 228 (now line 226) of the same file. This declaration was unreachable dead code — `ret` was already declared and in scope from the top of the script, making this a redundant re-declaration with no effect.

## Other comments:

The `local` misuse was discovered by accidentally running `tools/upgrade.sh` with `sh upgrade.sh` instead of `zsh upgrade.sh`. The POSIX `sh` shell correctly rejects `local` outside of a function, surfacing the issue. Because `zsh` silently promotes `local` at script scope to a normal variable assignment, the bug had no visible impact in normal use — but it represents incorrect shell scripting practice.

AI disclosure: Claude Code (claude-sonnet-4-6) assisted in drafting this PR description.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env zsh
 set +u # disable nounset
 
-local ret=0 # exit code
+ret=0 # exit code
 
 # Protect against running with shells other than zsh
 if [ -z "$ZSH_VERSION" ]; then


mcornella:
Hi, we're generally expecting things like this done as part of other fixes, features or bigger redactors. This doesn't have any impact to default functionality, so I'm closing the PR. Thanks

PR #13617, opened 2026-03-06
refactor(bundler): Remove bundle_install

halilim:
## Standards checklist:

## Changes:

- Remove the `bundle_install` function, which should only be needed for deprecated Bundler versions (4 years and older)

## Other comments:

- `commands[bundle]` check: The shell should provide feedback for unrecognized commands
- `_within-bundled-project` check: Bundler returns `Could not locate Gemfile`
- `1.4.0` check: It's 12+ years old and beyond deprecated
- `BUNDLE_JOBS`: It's been enabled by default for all platforms since `2.3` (the first non-deprecated -but still legacy- version, 4+ years old), and before that, for non-Windows

Supported versions (from https://bundler.io/man/bundle-install.1.html):

Screenshot 2026-03-06 at 08 39 57

### DIFF
--- a/plugins/bundler/README.md
+++ b/plugins/bundler/README.md
@@ -11,18 +11,18 @@ plugins=(... bundler)
 
 ## Aliases
 
-| Alias  | Command                              | Description                                                                              |
-|--------|--------------------------------------|------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


robbyrussell:

PR #13623, opened 2026-03-08
fix(last-working-dir): use explicit return 0 in early-exit guards

aecant:
## Standards checklist:

## Changes:
- Replace bare `return` with explicit `return 0` in the two early-exit guards at the bottom of `last-working-dir.plugin.zsh`

## Other comments:
The bare `return` statements inherit the exit code of the last executed command, which can be non-zero depending on the shell environment. Plugin managers that check the exit code of `source` (e.g. zcomet) incorrectly report a sourcing failure as a result, even though the plugin loads and functions correctly. Using `return 0` makes the intent explicit and avoids the false error.

Tested on macOS with Ghostty and Kitty terminals. AI-assisted (Claude helped diagnose the root cause).


carlosala:
Thanks for your contribution!

PR #13626, opened 2026-03-10
Feat/visual mode indicator

KevinWMatthews:
## Standards checklist:

Add a customizable indicator for the `vi-mode` plugin's:
- visual mode
- visual line mode

These can be configured with the variables:
- `VISUAL_MODE_INDICATOR`
- `VISUAL_LINE_MODE_INDICATOR`

The default is **<**<< in white, similar to but contrasting normal mode's default setting of **<**<< in the red.

## Changes:

- add two environment variables for configuring the prompt in visual and visual line modes
- extend `vi_mode_prompt_info()` to account for these new prompts
- reset the prompt on mode change, considering that visual mode is tracked by the active region

## Other comments:

I originally used Claude to generate this code, and then reviewed and reduced the change set manually.

Here is an example of the default behavior:
vi-mode-default

and an example that mirrors vim's behavior (`-- INSERT --`, `-- VISUAL --`, `-- VISUAL LINE --`):
vi-mode-vi-like

This latter example exposes a pre-existing bug: that the insert indicator is not cleared when enter is pressed (ticket not yet filed).


KevinWMatthews:
Ah, this change depends on those in #13597. I had meant to stack this PR on that one, but that doesn't seem to be possible?

PR #13631, opened 2026-03-13
feat(kube-ps1): add environment indicator to the prompt layout

dolefulowl:
## Standards checklist:

## Changes:
Add an optional global variable `KUBE_PS1_ENV_CTX_ENABLE` (defaults to `false`) to get environment markers from `KUBE_PS1_CONTEXT` (like dev/test/staging/prod) and display them in a separate block before the main layout; e.g `testing-mycluster` becomes `[test] mycluster`. The main idea for this feature is to be able to manage the colours of your cluster environments independently of cluster names. It also adds many various global variables, so that everyone can tune the feature to their liking and setup.


dolefulowl:
@mcornella Hi there 👋 I would be really glad if you could review my PR :)

dolefulowl:
@carlosala @mcornella Hi guys, I would really appreciate it if you could give some feedback on it

mcornella:
> @carlosala @mcornella Hi guys, I would really appreciate it if you could give some feedback on it

Hi @dolefulowl, we're managing kube-ps1 as a pull dependency from repo: jonmosco/kube-ps1

Ideally we'd like to keep divergences to zero. Please suggest this feature upstream and we'll pull it in once it's released there.

My only 2 cents: you're copying logic for multiple environments. A better approach would be using an associative array for label, color and re.

mcornella:
Closing for now, please notify us here if there's an update, even though updates should be pulled in automatically once a week. Thanks!

PR #13663, opened 2026-04-03
feat(juju): add native zsh completion and fix plugin utilities

iyiguncevik:
feat(juju): add native zsh completion and fix plugin utilities

## Standards checklist:

## Changes:

- Add _juju: native zsh completion script replacing bash-completion sourcing;
supports subcommand, flag, model, and controller tab completion including
"controller:model" format
- Fix missing `local controller` declaration in jclean()
- Respect $JUJU_DATA env var in jcontroller() and jmodel()
- Fix jmodel() to handle controller names with special chars and "null" yq
output
- Change wjst() default poll interval from 5s to 1s; guard against missing juju
binary

## Other comments:

- Some parts of the code was written with the help of Claude or GPT-Codex

### DIFF
--- a/plugins/juju/_juju
+++ b/plugins/juju/_juju
@@ -1,0 +1,236 @@
+#compdef juju
+(( $+functions[compdef] )) && compdef _juju juju
+
+# zsh completion for juju                                  -*- shell-script -*-
+
+__juju_debug()
+{
+    local file="$BASH_COMP_DEBUG_FILE"
+    if [[ -n ${file} ]]; then
+        echo "$*" >> "${file}"
+    fi
+}
+
+__juju_help_options()
+{
+    local out line token cleaned desc f
+    local -a opts pending
+    typeset -U opts
+
+    __juju_debug "[options] called with args: $*"
+    out=$(command juju help "$@" 2>/dev/null)
+    local rc=$?
+    __juju_debug "[options] juju help exit code: $rc, output length: ${#out}"
+    (( rc )) && return 1
+
+    while IFS= read -r line; do
+        if [[ "$line" =~ '^[[:space:]]{0,3}-' ]]; then
+            for f in "${pending[@]}"; do opts+=("$f"); done
+            pending=()
+            for token in ${(z)line}; do
+                cleaned="${token%%,*}"
+                cleaned="${cleaned%%;*}"
+                cleaned="${cleaned%%]*}"
+                cleaned="${cleaned%%)*}"
+                cleaned="${cleaned%%=<*}"
+                cleaned="${cleaned%%=*}"
+                cleaned="${cleaned%%<*}"
+                cleaned="${cleaned%%\[*}"
+                cleaned="${cleaned%%\(*}"
+                [[ "$cleaned" == --* || "$cleaned" == -[[:alnum:]] ]] || continue
+                [[ "$cleaned" == "-" || "$cleaned" == "--" ]] && continue
+                __juju_debug "[options] found flag: $cleaned"
+                pending+=("$cleaned")
+            done
+        elif (( ${#pending} )) && [[ -n "$line" ]]; then
+            desc="${line#"${line%%[![:space:]]*}"}"
+            desc="${desc//:/\\:}"
+            __juju_debug "[options] desc for ${pending[*]}: $desc"
+            for f in "${pending[@]}"; do opts+=("${f}:${desc}"); done
+            pending=()
+        elif [[ -z "$line" ]]; then
+            for f in "${pending[@]}"; do opts+=("$f"); done
+            pending=()
+        fi
+    done < <(printf "%s\n" "$out")
+
+    for f in "${pending[@]}"; do opts+=("$f"); done
+    __juju_debug "[options] total opts: ${#opts}, first few: ${opts[1]} ${opts[2]} ${opts[3]}"
+
+    printf "%s\n" "${opts[@]}"
+}
+
+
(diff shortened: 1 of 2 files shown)


iyiguncevik:
@carlosala Can I get a review? :pray:

plugins/juju/_juju
carlosala, line 1: Where does this completion come from? Is it generated? If it is, please follow the same structure as `deno` plugin, so that each user generates its own completion.

plugins/juju/juju.plugin.zsh
```diff
@@ -3,12 +3,14 @@
...
-[[ -f "$completion_file" ]] && source "$completion_file"
-unset completion_file
+source "${0:A:h}/_juju"
+
+# group-name '' enables visual separation between completion groups (e.g. models
+# vs controllers in juju switch <TAB>). This is a safe global setting that
+# improves completion display for all commands.
+zstyle ':completion:*' group-name ''
+# Show group headers only for juju completions.
+zstyle ':completion::complete:juju:*' format '%B%d%b'
```
carlosala: This section does not belong here. Each user should decide their own format.

plugins/juju/juju.plugin.zsh
```diff
@@ -165,27 +168,32 @@ jreld() {
 
 # Return Juju current controller
 jcontroller() {
-  local controller="$(awk '/current-controller/ {print $2}' ~/.local/share/juju/controllers.yaml)"
-  if [[ -z "$controller" ]]; then
-    return 1
-  fi
+  local file=${JUJU_DATA:=~/.local/share/juju}/controllers.yaml
```
carlosala:
```suggestion
  local file="${JUJU_DATA:-~/.local/share/juju}/controllers.yaml"
```

plugins/juju/juju.plugin.zsh
```diff
@@ -165,27 +168,32 @@ jreld() {
...
+  local controller="$(awk '/current-controller/ {print $2}' "$file")"
+  [[ -z "$controller" ]] && return 1
 
   echo $controller
   return 0
 }
 
 # Return Juju current model
 jmodel() {
+  local file=${JUJU_DATA:=~/.local/share/juju}/models.yaml
```
carlosala:
```suggestion
  local file="${JUJU_DATA:-~/.local/share/juju}/models.yaml"
```

plugins/juju/juju.plugin.zsh
```diff
@@ -194,9 +202,10 @@ jmodel() {
   return 0
 }
 
-# Watch juju status, with optional interval (default: 5 sec)
+# Watch juju status, with optional interval (default: 1 sec)
 wjst() {
-  local interval="${1:-5}"
+  command -v juju >/dev/null 2>&1 || return 1
+  local interval="${1:-1}"
```
carlosala, line 194: What is the rationale behind this change?

carlosala (changes requested):

plugins/juju/juju.plugin.zsh
iyiguncevik: You're right, this was overriding any format user might have defined. I changed it so that any user setting is not overridden. I didn't want to completely remove it because it provides a very good default value and makes the completion with groups looks very intuitive. I hope that's alright.

plugins/juju/juju.plugin.zsh
iyiguncevik: This alias is used while waiting for status command until juju is done with its changes. This usually takes a few seconds. Having 5 seconds wait duration is way too long for that.

People typically uses a one second refresh (not even the 2 seconds default). I don't know why the default was 5 seconds, it's too long for juju and doesn't make so much sense.

Let me know if you have concerns with backwards compatibility etc.

(the other line is to ensure juju command exist before running watch, otherwise it'll keep printing out "command not found")

plugins/juju/_juju
iyiguncevik: This is not generated. I've written it myself. Unfortunately, juju doesn't have a builtin completions like in `deno` or `lxc` (it's not even in the backlog).

I've addressed all other comments. Thanks for the review!

plugins/juju/juju.plugin.zsh
```diff
@@ -3,12 +3,21 @@
...
+# (grouped completions with bold headers) without overriding personal preferences.
+# To customise, set these zstyles in your .zshrc before or after loading oh-my-zsh:
+#   zstyle ':completion:*:*:juju:*' group-name ''
+#   zstyle ':completion:*:*:juju:*' format '%B%d%b'
+local _juju_tmp
+zstyle -s ':completion:*:*:juju:*' group-name _juju_tmp \
+    || zstyle ':completion:*:*:juju:*' group-name ''
+zstyle -s ':completion:*:*:juju:*' format _juju_tmp \
+    || zstyle ':completion:*:*:juju:*' format '%B%d%b'
+unset _juju_tmp
```
carlosala:
```suggestion
```

1. There is no need to source the completion. Should work out of the box.
2. I really prefer not to force completion styles inside the plugin, even if it is only for `juju`.

carlosala (commented):
Just a small change, looks good!

plugins/juju/juju.plugin.zsh
iyiguncevik: Done. Thanks!

carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #13671, opened 2026-04-05
test(git): add regression tests for non-UTF-8 bytes in git status output

### LINKED ISSUE #13330: _omz_git_prompt_status:65: regex matching error: illegal byte sequence
### Describe the bug

Our git branch name contain Chinese String open Zsh tips
_omz_git_prompt_status:65: regex matching error: illegal byte sequence

### Steps to reproduce

You can git branch name  中文-1.0.0-中文
cd to route tips _omz_git_prompt_status:65: regex matching error: illegal byte sequence

### Expected behavior

You can git branch name  中文-1.0.0-中文
cd to route tips _omz_git_prompt_status:65: regex matching error: illegal byte sequence

_No response_

### OS / Linux distribution

macOS 15.2

### Zsh version

### Oh My Zsh version

master (763aab3)

### Terminal emulator

Terminal

None

### Additional context

_No response_

0dayfall:
## Summary

Two things changed since this PR was opened, so I've rewritten it.

**1. The fix landed upstream.** #13878 (`898b579c`, merged 2026-08-05) closed #13330 with `local -x LC_ALL=C` at the top of `_omz_git_prompt_status`. That's a better fix than the manual save/restore this PR originally proposed — it covers the whole function including early returns, restores automatically, and is exported for child processes. I've **dropped the `lib/git.zsh` change** here; nothing about the fix needs revisiting.

**2. My original root-cause description was wrong, and so were the tests.** I claimed `[^ ]+` rejects multibyte sequences in a UTF-8 locale. It doesn't — valid UTF-8 matches fine:

```zsh
$ LC_ALL=en_US.UTF-8 zsh -c '[[ "## 中文-1.0.0 [ahead 1]" =~ "^## [^ ]+ \[(.*)\]" ]]; echo rc=$?'
rc=0
```

The actual trigger is bytes that are **invalid in the active locale's encoding** — the C library regex returns `REG_ILLSEQ`. A GBK-encoded branch name or file path under a UTF-8 locale does it:

```zsh
$ LC_ALL=en_US.UTF-8 zsh -c '[[ $'"'"'## \xd6\xd0\xce\xc4-br [ahead 1]'"'"' =~ "^## [^ ]+ \[(.*)\]" ]]'
zsh: regex matching error: illegal byte sequence
```

The comment on `898b579c` has this right. Consequence: my original tests used valid UTF-8 Chinese, so they **passed with or without the fix** — they proved nothing.

## What this PR now adds

`lib/tests/git.test.zsh`, rewritten so it genuinely reproduces the bug:

- Branch-tracking line with non-UTF-8 bytes → no stderr (the `^## [^ ]+ \[(.*)\]` match)
- File path with non-UTF-8 bytes → no stderr (the per-prefix `status_text` loop, a separate regex)
- `[ahead 1]` still parsed for such a branch — proves `LC_ALL=C` keeps parsing working, not just silences the error
- ASCII `[behind 3]` control

The byte-sequence tests need a multibyte locale to reproduce, so the file picks the first UTF-8 locale from `locale -a` and skips them with a notice if the system has none. Follows the `lib/tests/cli.test.zsh` conventions (`#!/usr/bin/zsh -df`, `print -u2` result lines, non-zero exit on failure).

## Test plan

Verified in both directions against current `master`:

- [x] `zsh lib/tests/git.test.zsh` → 4/4 pass, exit 0
- [x] With `local -x LC_ALL=C` deleted from `lib/git.zsh` → 3 failures, exit 3, ASCII control still passing

Happy to close this if you'd rather not carry tests for `lib/git.zsh`.

> AI-assisted: Claude was used to rewrite these tests and verify them. The changes were reviewed and verified manually.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -103,6 +103,11 @@ function _omz_git_prompt_status() {
   local status_lines
   status_lines=("${(@f)${status_text}}")
 
+  # Use C locale for regex matching to avoid "illegal byte sequence" errors
+  # when branch names or file paths contain non-ASCII characters (e.g. Chinese)
+  local _omz_lc_all=$LC_ALL
+  LC_ALL=C
+
   # If the tracking line exists, get and parse it
   if [[ "$status_lines[1]" =~ "^## [^ ]+ \[(.*)\]" ]]; then
     local branch_statuses
@@ -125,6 +130,8 @@ function _omz_git_prompt_status() {
       statuses_seen[$status_constant]=1
     fi
   done
+
+  LC_ALL=$_omz_lc_all
 
   # Display the seen statuses in the order specified
   local status_prompt
--- a/lib/tests/git.test.zsh
+++ b/lib/tests/git.test.zsh
@@ -1,0 +1,74 @@
+#!/usr/bin/zsh -df
+
+# Regression tests for lib/git.zsh
+
+local -i _failures=0
+
+run_test() {
+  local description="$1"
+  local got="$2"
+  local expected="$3"
+
+  print -u2 "Test: $description"
+  if [[ "$got" == "$expected" ]]; then
+    print -u2 "\e[32mSuccess\e[0m"
+  else
+    print -u2 "\e[31mError\e[0m"
+    print -u2 "  expected: ${(q)expected}"
+    print -u2 "  got:      ${(q)got}"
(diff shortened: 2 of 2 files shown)


0dayfall:
Whats happening?

PR #13675, opened 2026-04-07
chore(archlinux): simplify pacman helper implementations

WH-2099:
## Standards checklist:

## Changes:

This PR simplifies several implementations in the `archlinux` plugin and makes them more idiomatic to zsh.

- `paclist`:
  the current implementation spawns one `pacman -Qs` per package via `xargs -P0`, which is unnecessarily expensive. It also does not make much use of zsh's own pattern-matching features. This is now handled in a single pass over `pacman -Qei`, parsed directly in zsh.

- `pacdisowned`:
  the previous implementation relied on temporary files and cleanup logic for a comparison that does not really need them. This now uses process substitution directly with `comm`, which is shorter and avoids unnecessary filesystem work.

- `pacweb`:
  the previous implementation parsed `pacman -Si` output through multiple external `grep` calls. This now parses the output once in zsh while keeping the same early-return behavior when package lookup fails.

- `upgrade`:
  the previous implementation repeatedly parsed command output with external tools and had a more repetitive command dispatch path. This now reads the required fields once in zsh, compares versions directly, and keeps the final helper execution path simpler.

- locale handling:
  the explicit `LANG=C` usage is kept intentionally. If the reason for that looks surprising, see `#11398` for earlier context. I still believe it is a necessary safeguard here, since I have run into odd `pacman` behavior under unusual locale settings, and these helpers depend on predictable command output.

- small follow-up cleanups:
  while touching the file, I also normalized a few style inconsistencies I noticed, such as quote style, helper formatting, a redundant spacing issue, and some argument quoting in helper functions.

## Other comments:

- Also smoke-tested the updated functions in a fresh `archlinux` container environment.
- Additional stubbed checks were used during development for command dispatch paths that are awkward to exercise directly.
- AI-assisted: Used AI tools(Gemini) to generate this PR description and assist with code review, but all the code changes were written manually.

### DIFF
--- a/plugins/archlinux/archlinux.plugin.zsh
+++ b/plugins/archlinux/archlinux.plugin.zsh
@@ -22,45 +22,49 @@ alias pacfileupg='sudo pacman -Fy'
 alias pacfiles='pacman -F'
 alias pacls='pacman -Ql'
 alias pacown='pacman -Qo'
-alias pacupd="sudo pacman -Sy"
+alias pacupd='sudo pacman -Sy'
 
 function paclist() {
-  pacman -Qqe | xargs -I{} -P0 --no-run-if-empty pacman -Qs --color=auto "^{}\$"
+  emulate -L zsh -o extendedglob
+
+  local line
+  local -A info
+  while IFS= read -r line; do
+    [[ $line == (#b)(Name|Version|Description)[[:space:]]#:[[:space:]]#(*) ]] || continue
+    info[$match[1]]=$match[2]
+
+    [[ $match[1] == Description ]] || continue
+    print -r -- "local/${info[Name]} ${info[Version]}"
+    print -r -- "    ${info[Description]}"
+  done < <(LANG=C pacman -Qei)
 }
 
 function pacdisowned() {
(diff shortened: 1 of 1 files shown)


aakashdev26 (approved):

ratijas:
Disclaimer: I'm not using oh-my-zsh itself anymore and barely use Linux these days where zsh would be relevant

With that said, I'm not sure what's the purpose of `emulate -L zsh -o extendedglob` in this context, and whether these additions can be called simplifications.

```zsh
[[ $line == (#b)([^:]##[^[:space:]])[[:space:]]#:[[:space:]]#(*) ]] || continue
```

Not saying that changes are bad. Maybe it's a good optimization, but even by looking at diff size it can hardly be called simplification.

---

https://zsh.sourceforge.io/Doc/Release/Options.html#index-EXTENDEDGLOB

> EXTENDED_GLOB
>     Treat the ‘#’, ‘~’ and ‘^’ characters as part of patterns for filename generation, etc. (An initial unquoted ‘~’ always produces named directory expansion.)

We aren't even working with file names here, so why bother?

PR #13686, opened 2026-04-12
Add pipenv argcomplete compatibility for >= 2026.5.0

### LINKED ISSUE #13719: Pipenv Changed Shell Completions in >= 2026.5.0
### Describe the bug

The pipenv plugin will throw errors when the user has a version installed that is >= `2026.5.0`. The terminal output is:

```
Shell completion has changed in pipenv >= 2026.5.0.
The _PIPENV_COMPLETE environment variable is no longer supported.

To enable shell completion, install the completion extra and use:
  pip install pipenv[completion]
  eval "$(register-python-argcomplete pipenv)"

For more details, see: https://pipenv.pypa.io/en/latest/shell.html#shell-completion
```

### Steps to reproduce

1. Have pipenv installed at >= `2026.5.0`
2. enable plugin `pipenv`
3. Open new shell
4. observe error

### Expected behavior

Completions are automatically setup, and no extra terminal output occurs from the plugin detailing the need to upgrade completion implementation.

_No response_

### OS / Linux distribution

macOS 15.7.4

### Zsh version

5.9

### Oh My Zsh version

master (2d584174)

### Terminal emulator

vscode, ghostty

None

### Additional context

_No response_

matheusfelipeog:
Hey. The pipenv completion system has changed and now uses the argcomplete package for versions >= 2026.5.0.

ref 1: https://github.com/pypa/pipenv?tab=readme-ov-file#shell-completion
ref 2: https://pipenv.pypa.io/en/latest/shell.html#shell-completion

I think this fix is enough to keep both old and new versions compatible.

### OMZ log:

```zsh
[WARNING]: Console output during zsh initialization detected.

...

-- console output produced during zsh initialization follows --

Shell completion has changed in pipenv >= 2026.5.0.
The _PIPENV_COMPLETE environment variable is no longer supported.

To enable shell completion, install the completion extra and use:
  pip install pipenv[completion]
  eval "$(register-python-argcomplete pipenv)"

For more details, see: https://pipenv.pypa.io/en/latest/shell.html#shell-completion
```

---

resolve #13719

### DIFF
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -2,15 +2,33 @@ if (( ! $+commands[pipenv] )); then
   return
 fi
 
-# If the completion file doesn't exist yet, we need to autoload it and
-# bind it to `pipenv`. Otherwise, compinit will have already done that.
-if [[ ! -f "$ZSH_CACHE_DIR/completions/_pipenv" ]]; then
-  typeset -g -A _comps
-  autoload -Uz _pipenv
-  _comps[pipenv]=_pipenv
-fi
+# Compatibility note:
+# pipenv < 2026.5.0 used Click-based shell completion driven by the
+# _PIPENV_COMPLETE environment variable.
+#
+# pipenv >= 2026.5.0 removed this mechanism and switched to argcomplete-based
+# completion using register-python-argcomplete instead.
 
-_PIPENV_COMPLETE=zsh_source pipenv >| "$ZSH_CACHE_DIR/completions/_pipenv" &|
+if (( $+commands[register-python-argcomplete] )); then
+  # pipenv >= 2026.5.0 (argcomplete-based completion)
+  autoload -U bashcompinit
+  bashcompinit
+
+  eval "$(register-python-argcomplete pipenv)"
+
+else
+  # pipenv < 2026.5.0 (legacy Click-based completion via _PIPENV_COMPLETE)
+
+  # If the completion file doesn't exist yet, we need to autoload it and
+  # bind it to `pipenv`. Otherwise, compinit will have already done that.
+  if [[ ! -f "$ZSH_CACHE_DIR/completions/_pipenv" ]]; then
+    typeset -g -A _comps
+    autoload -Uz _pipenv
+    _comps[pipenv]=_pipenv
+  fi
+
+  _PIPENV_COMPLETE=zsh_source pipenv >| "$ZSH_CACHE_DIR/completions/_pipenv" &|
+fi
 
 if zstyle -T ':omz:plugins:pipenv' auto-shell; then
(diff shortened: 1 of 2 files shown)


carlosala (commented):
`register-python-argcomplete` seems something common for more than one python library. We should check for pipenv version, if the user has `pipenv[completion]` package installed, etc. Could you figure that out?

matheusfelipeog:
It is now checking the specific pipenv version, @carlosala

plugins/pipenv/pipenv.plugin.zsh
```diff
@@ -2,15 +2,37 @@ if (( ! $+commands[pipenv] )); then
...
+# Compatibility note:
+# pipenv < 2026.5.0 used Click-based shell completion driven by the
+# _PIPENV_COMPLETE environment variable.
+#
+# pipenv >= 2026.5.0 removed this mechanism and switched to argcomplete-based
+# completion using register-python-argcomplete instead.
+
+autoload -Uz is-at-least
+
+_pipenv_version="${$(pipenv --version 2>/dev/null)#pipenv, version }"
```
carlosala: It'd be nice to cache as well this command, it is very slow, around 300ms in my machine. Could you do it as well? You can save the result in a `$ZSH_CACHE_DIR/pipenv_version`, for instance, and use it to figure it out, and refresh it async, just like we generate completions.

carlosala (changes requested):

carlosala:
Hey! Were you able to check it out? Thanks!

plugins/pipenv/pipenv.plugin.zsh
matheusfelipeog: @carlosala You're right, that is indeed slow. I implemented a solution as you suggested. Thanks.

matheusfelipeog:
> Hey! Were you able to check it out? Thanks!

@carlosala Sorry for the delay. I had forgotten about this. 🤦‍♂️

I implemented a solution to cache the version now. Ready for your review, please.

---

p.s. I also added a section about the cache behavior in the `README.md`, [similar to the one in GitHub CLI](https://github.com/ohmyzsh/ohmyzsh/blob/df34d2b8d575777465aed8ae9b7cd90d63fdcd6e/plugins/gh/README.md) — which, by the way, I think should be removed there, since version caching was removed in this commit: [3f21432](https://github.com/ohmyzsh/ohmyzsh/commit/3f214329d631b3ae39d6b283262c77819a0078de#diff-d293de76bc2488c1027313b4785e79932fadda0ce4581b8223500b137e4e79d6) 🤔

matheusfelipeog:
Just a friendly ping in case you forgot about this PR @carlosala

carlosala (approved):

mcornella (approved):

matheusfelipeog:
Thank you for your time 😄

PR #13715, opened 2026-04-27
fix(z): prevent stack overflow from recursive Tab binding on re-source

### LINKED ISSUE #13714: Stack overflow problem with z plugin introduced in #13710
### Describe the bug

#13710 merged yesterday seems to be causing a stack overflow problem with `z` plugin. The exact error is `_zshz_zle_completion_widget:34: maximum nested function level reached; increase FUNCNEST?` The same error doesn't have with commit #13712 - the immediate preceding commit.

### Steps to reproduce

1. Enable plugin 'z'
2. Type `cd` and press `tab`

### Expected behavior

`cd [tab]` should auto complete the files from current directory

_No response_

### OS / Linux distribution

Mac OS Tahoe 26.3.1

### Zsh version

zsh 5.9 (arm64-apple-darwin25.0)

### Oh My Zsh version

master (278ee10)

### Terminal emulator

iTerm2

None

### Additional context

_No response_

maheshpec:
## Summary
- Fixes a stack overflow (`maximum nested function level reached`) introduced in #13710
- The new `_zshz_zle_completion_widget` saves the current `^I` binding and delegates to it, but if the plugin is sourced more than once, the saved binding points to the widget itself, causing infinite recursion on any Tab press
- Guards against self-reference by only capturing the binding on first load and falling back to `expand-or-complete` if the saved value is our own widget

Fixes #13714

## Test plan
- [ ] Enable `z` plugin and open a new shell
- [ ] Type `cd ` and press Tab — verify directory completion works without errors
- [ ] Source `.zshrc` again (to re-source the plugin), then repeat Tab completion — verify no stack overflow
- [ ] Type `z foo bar` and press Tab — verify multi-word completion still works

### DIFF
--- a/plugins/z/z.plugin.zsh
+++ b/plugins/z/z.plugin.zsh
@@ -960,8 +960,15 @@ add-zsh-hook chpwd _zshz_chpwd
 
 (( ${fpath[(ie)${0:A:h}]} <= ${#fpath} )) || fpath=( "${0:A:h}" "${fpath[@]}" )
 
-# Save the existing Tab binding
-ZSHZ[TAB_BINDING]="${$(bindkey -M main '^I')##* }"
+# Save the existing Tab binding (only on first load to avoid self-reference
+# if the plugin is sourced more than once)
+if [[ -z ${ZSHZ[TAB_BINDING]} ]] || \
+   [[ ${ZSHZ[TAB_BINDING]} == _zshz_zle_completion_widget ]]; then
+  ZSHZ[TAB_BINDING]="${$(bindkey -M main '^I')##* }"
+  # If we somehow captured our own widget (race condition), fall back to default
+  [[ ${ZSHZ[TAB_BINDING]} == _zshz_zle_completion_widget ]] && \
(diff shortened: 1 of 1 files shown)


agkozak:
@maheshpec Good catch! Would you like to submit your fix to the upstream repository? ([agkozak/zsh-z](https://github.com/agkozak/zsh-z)).

agkozak:
@maheshpec Actually, hold off on submitting that PR. I may have found a solution I prefer. But you'll get full credit for diagnosing the bug.

maheshpec:
thank you @agkozak . Would me like to retain this PR or close?

agkozak:
@maheshpec You should leave it up for now. If the Oh My Zsh maintainers have an opinion about which fix is preferable, obviously I'm interested. :) Thanks again for your very timely discovery of that infinite loop.

carlosala:
We prefer to stick to upstream to avoid maintaining the codebase. Thank you @maheshpec, adding you as coauthor to the update :)

PR #13722, opened 2026-05-01
feat(brew): Add option to disable aliases

jahirvidrio:
**Changes:**

   - Added a `zstyle :omz:plugins:brew:aliases` check in `brew.plugin.zsh`.
   - Wrapped existing aliases and the brews function in this check.
   - Updated README.md with instructions on how to disable aliases.


carlosala:
This option is [built-in in ohmyzsh](https://github.com/ohmyzsh/ohmyzsh#skip-aliases).

Thanks for your contribution!

PR #13723, opened 2026-05-02
perf: deduplicate fpath before comparing cached zcompdump metadata

### LINKED ISSUE #13720: oh-my-zsh.sh compinit cache is invalidated on every startup when $fpath contains entries from other shell init files (brew, pyenv, conda, etc.)
### Problem

Oh-My-Zsh's compinit caching logic in `oh-my-zsh.sh` (approx. lines 113–132) compares the current `$fpath` against a stored metadata line in `$ZSH_COMPDUMP` using `grep -Fx` (exact whole-line match). In many common setups, the `$fpath` string differs subtly between sessions, causing OMZ to **delete the dump and do a full compinit rebuild on every shell startup**.

The underlying `.zcompdump` file is correct — OMZ just never trusts it, wasting 500–1500 ms per new shell.

### How to reproduce

Minimal reproduction on a fresh OMZ install:

1. Add a line to `~/.zprofile` that modifies `$fpath`. Any of the following will do:
   ```sh
   eval "$(/opt/homebrew/bin/brew shellenv)"     # macOS Homebrew
   # or
   eval "$(pyenv init -)"
   # or
   eval "$(conda shell.zsh hook)"
   ```

2. Measure two consecutive shell startups:
   ```sh
   time zsh -i -c exit     # First run
   time zsh -i -c exit     # Second run
   ```

3. Confirm the dump is being rewritten each time:
   ```sh
   stat -f "%Sm" ~/.zcompdump-*    # macOS
   # or: stat -c "%y" ~/.zcompdump-*    # Linux
   zsh -i -c exit
   stat -f "%Sm" ~/.zcompdump-*    # timestamp changed
   ```

### Correct behavior

After the first shell startup builds the dump, subsequent startups should reuse it (via `compinit -C`) until either:
- The dump is older than a configured TTL, OR
- The set of completion sources has actually changed

Second-run startup time should be dramatically faster than the first run.

### Actual behavior

Second-run startup time is identical to the first run. The `.zcompdump` file is deleted and rebuilt every time, because OMZ's fpath metadata check fails.

On a representative OMZ install with ~13 plugins (git, brew, fzf, kubectl, npm, etc.):

| Configuration | Startup | compinit | compdef calls |
|---|---|---|---|
| Current OMZ (cache always invalidated) | ~1100–1300 ms | ~1100 ms | ~820 |
| With `compinit -C` (cache trusted) | ~130 ms | ~15 ms | ~17 |

### Root cause

Current OMZ logic on `master`:

```zsh
# Construct zcompdump OMZ metadata
zcompdump_revision="#omz revision: $(builtin cd -q "$ZSH"; git rev-parse HEAD 2>/dev/null)"
zcompdump_fpath="#omz fpath: $fpath"

# Delete the zcompdump file if OMZ zcompdump metadata changed
if ! command grep -q -Fx "$zcompdump_revision" "$ZSH_COMPDUMP" 2>/dev/null \
   || ! command grep -q -Fx "$zcompdump_fpath" "$ZSH_COMPDUMP" 2>/dev/null; then
  command rm -f "$ZSH_COMPDUMP"
  zcompdump_refresh=1
fi
```

The fpath check uses `grep -Fx` against `$fpath` joined as a single string. This is brittle because:

1. **Order matters.** `a b c` does not match `a c b`. Tools that prepend vs. append to `fpath` can produce different orderings between sessions.
2. **Duplicates matter.** If a path appears twice in one session and three times in another, the strings don't match even though the effective fpath is identical.
3. **`fpath` is mutable before OMZ runs.** `.zprofile`, `.zshenv`, Apple's `/etc/zshenv`, `brew shellenv`, pyenv/conda/nvm/mise hooks all inject entries with variable timing.

Example of the observed drift between sessions on a Homebrew setup:

```
Session N stored: #omz fpath: .../site-functions .../zsh-completions .../plugins/git ...
Session N+1 live: #omz fpath: .../site-functions .../site-functions .../zsh-completions ...
                               ^^^ duplicate injected by brew shellenv + brew plugin
```

The `grep -Fx` fails, dump is deleted, full rebuild runs, dump is rewritten with session N+1's fpath, and the cycle repeats next session.

### Environment

Please replace with your actual values when filing:

```
OMZ version:    061f773dd356df52a8bccd5e73377c012f97ef14 (2026-04-13)
Zsh version:    zsh 5.9 (arm64-apple-darwin25.0)
OS / kernel:    Darwin 25.4.0 Darwin Kernel Version 25.4.0 arm64 (macOS, Apple Silicon)
Terminal:       Warp v0.2026.04.22
Relevant tools: Homebrew 5.1.7, pyenv 2.6.1, conda 26.3.2, mise 2025.4.0, direnv 2.35.0
```

### Self-contained reproduction

Attached as a gist: [`reproduce.sh`](https://gist.github.com/rndjams/c7a3e85348ecb348f2d937c81628e49a) — a ~160-line zsh script
that isolates the bug in a `mktemp -d` sandbox (does not touch your
real `$HOME`). It:

1. Clones a fresh OMZ into `$TMPDIR`
2. Writes a minimal `.zshrc` (only the `git` plugin) and a `.zprofile`
   that perturbs `$fpath` between sessions
3. Runs `zsh -i -c exit` several times
4. Reports whether the dump was rewritten between runs

Expected behavior on current master: dump rewritten on every run where
fpath differs from the stored metadata (several distinct mtimes across
runs). Expected behavior with the proposed fix: one mtime across all
runs.

Sample output included in the gist as `SAMPLE-OUTPUT.md`, showing both
the bug on master and the expected result after the fix.

### Related issues and PRs

Searched before filing (per CONTRIBUTING guidelines):

- #8536 — "Why my Oh My Zsh startup is slow" (Jan 2020, open, `Performance` label). Symptoms match this root cause but the issue never identified the fpath-string comparison as the culprit.
- #7642 — "After enabling a plugin, autocompletion requires manual rm of a `.zcompdump` file" (2019, open). Related but the *opposite* problem (cache not invalidated when it should be).
- [forivall:oh-my-zsh @ `use-cached-compdump`](https://github.com/ohmyzsh/ohmyzsh/compare/master...forivall:oh-my-zsh:use-cached-compdump) — Nov 2024 branch by @forivall adding an opt-in `ZSH_COMPINIT_CACHE=true` flag. Has not been filed as a PR as of April 2026.
- [ctechols gist](https://gist.github.com/ctechols/ca1035271ad134841284) — 213-star community workaround from 2015. The canonical "`compinit -C` if dump is fresh" pattern.
- [mattmc3/ez-compinit](https://github.com/mattmc3/ez-compinit) — alternative plugin that replaces OMZ's compinit handling entirely.

### Proposed solutions (for discussion)

Three options, increasing in invasiveness:

**Option A — Normalize fpath before comparison.** Sort and dedupe the fpath string so ordering and duplicates don't invalidate the cache:
```zsh
zcompdump_fpath="#omz fpath: ${${(@onu)fpath}}"
```
Preserves current semantics ("invalidate when fpath content changes"), fixes the common case silently. Existing dumps invalidated once on upgrade.

Note on syntax: the nested `${...}` is required. Single-form `${(onu)fpath}` silently does not sort/dedupe in a string context — the flags need array-element expansion via `@`, then re-wrapping for interpolation.

**Option B — Opt-in time-based cache (forivall's approach).** Add `ZSH_COMPINIT_CACHE=true` that short-circuits to `compinit -C -d "$ZSH_COMPDUMP" -D` when the dump is less than 24 h old. Zero impact on existing users.

**Option C — Combine A and B.** Fix the common case for all users (A), and offer power users the fastest path (B).

### Would a PR be welcome?

If so, which option (A/B/C or something else) would be preferred? I'm happy to open a PR once maintainers indicate a direction. Drafting is already in progress, with credit to @forivall for the prior-art branch on Option B.

cc @forivall (prior work on `use-cached-compdump` branch)

---

**AI-assisted disclosure (per CONTRIBUTING.md):** This issue was drafted with AI assistance during root-cause investigation on a real Homebrew+OMZ setup. All code references and measurements were verified locally. The author understands the fix and can explain/defend each part on request.

rndjams:
## Summary

Fixes silent compinit cache invalidation where `.zcompdump` is deleted
and rebuilt on every shell startup in any setup where another init
file injects duplicate entries into `$fpath` (macOS Homebrew, pyenv,
conda, nvm, mise, direnv, etc.). The underlying dump is functionally
valid in these cases; the existing `grep -Fx` metadata check just
fails to recognize it.

Closes #13720.

## What changed

`zcompdump_fpath` now uses `${${(@u)fpath}}` (dedup via array
expansion, preserving first-occurrence order) instead of raw `$fpath`.
Sessions that differ only in fpath duplicate counts now compare equal,
so the cache is preserved across sessions where `brew shellenv` or a
similar hook re-prepends an already-present directory.

**Preserving first-occurrence order is deliberate.** zsh's `compinit`
resolves colliding completion files (two directories both shipping
`_foo`) by taking whichever appears first in `$fpath`. Two fpaths
containing the same set of entries in different orders can therefore
register different `compdef` bindings and are *not* semantically
equivalent. An earlier revision of this patch used `${${(@onu)fpath}}`
(sort + dedup) and was withdrawn after review feedback from @mologie
on the issue thread, who correctly noted that sort-normalizing would
mask legitimate invalidation in that case. Dedup-only cannot cause
that regression: removing later duplicates never changes which entry
is first, so the collision winner is identical before and after dedup.

No new settings. No behavior change for users whose fpath was already
duplicate-free. No change to insecure-directory warnings, compfix
behavior, zrecompile, or any public variable.

Users will see exactly one cache rebuild on first upgrade (because the
stored metadata was in the old non-deduped form). After that, cache
hits work as originally intended.

## Scope of the fix (what this PR does and does not cover)

This PR fixes the dominant real-world cause of spurious invalidation:
**duplicate entries injected by shell init hooks.** It does not claim
to fix the narrower case where two sessions produce fpaths with the
same *set* of entries in a legitimately different order (e.g. plugin
manager A prepends while plugin manager B appends, with no duplicates
on either side). In that case the current `grep -Fx` invalidation is
preserved and the rebuild is unnecessary but correct.

For a design that handles the reorder-of-non-colliding-entries case
automatically, see the "Future work" note below on per-fpath-hashed
dump paths. That is intentionally left for a follow-up PR to keep this
diff minimal and low-risk.

## Measurements

Benchmarked with `hyperfine --warmup 3 --min-runs 15 --max-runs 30 'zsh -i -c exit'`
on macOS (arm64, Apple Silicon, zsh 5.9), driven by a `.zprofile` that
alternates between injecting 1 and 2 copies of the same fpath entry to
mimic real-world drift (e.g. `brew shellenv` run once vs. twice across
login/interactive shells). The "1-plugin" row uses `plugins=(git)`; the
"13-plugin" row uses a representative realistic set (git, brew, fzf,
kubectl, npm, node, aws, docker, history, colored-man-pages,
command-not-found, sudo, virtualenv).

| Configuration | 1 plugin | 13 plugins |
|---|---|---|
| `master` — cache invalidated every session | 2104 ms ± 144 ms | 3442 ms ± 190 ms |
| This PR (`${${(@u)fpath}}`) — cache preserved | **279 ms ± 25 ms** | **348 ms ± 23 ms** |
| Speedup | **7.5×** | **9.9×** |

For reference, I also benchmarked the rejected sort+dedup form
(`${${(@onu)fpath}}`) at 1 plugin: **308 ms ± 25 ms**. Dedup-only is
marginally faster than sort+dedup because `(u)` does less work than
`(onu)`, and it is strictly more correct (see Alternatives below).

## Correctness validation

Two reproduction scripts run against master and both patch variants
confirm dedup-only is the only form that passes both cases:

| Variant | duplicate-injection case | collision-reorder case |
|---|---|---|
| `master` (unpatched) | ✗ cache rewritten every run | ✓ correctly invalidates |
| `${${(@u)fpath}}` (**this PR**) | ✓ cache reused | ✓ correctly invalidates |
| `${${(@onu)fpath}}` (rejected) | ✓ cache reused | ✗ **stale dump retained** |

The "duplicate-injection case" deterministically alternates between 1
and 2 copies of the same fpath entry and checks that the stored
`.zcompdump` mtime is stable across runs. The "collision-reorder"
case creates two directories each shipping a `_git` completion file,
flips their order in `$fpath` between runs, and checks that the dump
is rewritten when the first-occurrence (and therefore the `compdef`
winner) changes. The sort+dedup form fails this case because its
stored metadata is identical across both orderings, so OMZ serves the
stale dump matched against the wrong `_git` winner.

Reproduction scripts: `reproduce-duplicates.sh` and
`reproduce-collision-reorder.sh` (attached to the issue; both run in a
`mktemp -d` sandbox and do not touch the user's real `$HOME`).

## Tested on

- macOS 15 (arm64), zsh 5.9, OMZ commit on master at time of testing
- Plugins tested: minimal (`git` only) and representative
  13-plugin set (git, brew, fzf, kubectl, npm, node, aws, docker,
  history, colored-man-pages, command-not-found, sudo, virtualenv)

Automated verification against both reproduction scripts (see
"Correctness validation" above). Manual verification:
- First shell builds dump once; subsequent shells reuse it
- `compdef` works (no "command not found: compdef" errors)
- Tab completion for brew, git, kubectl, aws, npm all functional
- `omz update` correctly invalidates via the revision check
  (verified by flipping `zcompdump_revision` on disk)
- Adding a new plugin to `plugins=(...)` correctly invalidates via
  the fpath check (verified by adding/removing a test plugin and
  confirming new completions are picked up without manual
  `rm .zcompdump*`)
- Reordering colliding completion files is still detected: the
  `reproduce-collision-reorder.sh` script exercises exactly the
  scenario @mologie flagged — two directories each shipping `_git`,
  with fpath order flipping between runs. With this patch the dump
  is correctly rewritten on every flip, so the completion winner
  changes accordingly. (This is the failure mode a sort+dedup form
  would introduce; dedup-only is strictly safe.)

## Prior art

- [ctechols gist](https://gist.github.com/ctechols/ca1035271ad134841284)
  (2015, 213 stars) — the original "`compinit -C` if dump is fresh"
  community workaround. This PR solves the same problem but through
  the safer path of making OMZ's existing invalidation logic work
  correctly.
- [forivall:use-cached-compdump](https://github.com/ohmyzsh/ohmyzsh/compare/master...forivall:oh-my-zsh:use-cached-compdump)
  (Nov 2024) — unfiled branch proposing a `ZSH_COMPINIT_CACHE=true`
  opt-in. That approach has trade-offs (see "Alternatives considered"
  below).

## Alternatives considered

I looked at what other popular zsh plugin managers do and at
@romkatv's [zsh-bench](https://github.com/romkatv/zsh-bench) analysis
before choosing this approach.

**Sort + dedup (`${${(@onu)fpath}}`, rejected after review):** an
earlier revision of this patch used this form. Withdrawn after
@mologie pointed out on #13720 that zsh's `compinit` resolves
colliding completion files by first-occurrence in `$fpath`, so two
fpaths with the same entries in a different order can legitimately
register different `compdef` functions. Sort-normalizing would mask
that invalidation and could serve a stale dump matched against the
wrong winner. The current dedup-only form preserves first-occurrence
order and is strictly safe under zsh's actual semantics.

**Opt-in time-based cache (`ZSH_COMPINIT_CACHE=true`, rejected for
this PR):** short-circuit to `compinit -C -d "$ZSH_COMPDUMP" -D` when
the dump is less than 24 h old. This is the pattern @forivall proposed
and is what Prezto does by default. Per
[zsh-bench's "Cutting Corners" section](https://github.com/romkatv/zsh-bench/tree/master/README.md#cutting-corners)
and the Prezto issue
[#2010](https://github.com/sorin-ionescu/prezto/issues/2010), this
pattern has a real UX cost: newly installed tool completions won't
appear for up to 24 h until the user manually runs `rm ~/.zcompdump*`.
zcomet and zim both moved away from this pattern after @romkatv's
review.

**Per-directory mtime check (zsh4humans/zim style, follow-up
candidate):** replace string comparison with iteration over each
directory in `$fpath` comparing mtime to the dump. Strictly more
accurate than string comparison — catches completions added to
existing directories, which the current logic misses. Worth exploring
as a follow-up once this PR lands.

**Per-fpath-hashed dump paths (follow-up candidate, per @mologie's
suggestion on #13720):** hash the live `$fpath` into the dump filename
so each distinct fpath gets its own cache and reorders of colliding
non-duplicate entries produce distinct caches automatically. Strictly
more correct than any normalization approach and auto-recovers when a
user toggles between fpath configurations. Trade-offs: cache files
accumulate (needs LRU or age-based cleanup), larger diff, and invites
bikeshedding about cache directory layout and XDG compliance. Better
as a separate follow-up PR once the duplicate-injection bleeding is
stopped.

Options B, C, and D are intentionally out of scope for this PR; see
the issue for full discussion.

## AI-assisted disclosure

Per the
[AI-assisted contributions section of CONTRIBUTING.md](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#a-note-on-ai-assisted-contributions):

This patch was developed with AI assistance during root-cause
investigation of a real OMZ slow-startup problem on a Homebrew-enabled
macOS install. The AI helped diagnose the `grep -Fx` failure and
draft the patch and PR description.

Two QA iterations were performed during patch development:

1. An initial attempt used the single-form `${(onu)fpath}` which
   silently does not apply sort/unique flags to the fpath array in a
   double-quoted string context. The reproduction script (attached to
   the issue) caught this on first run — the stored fpath was not
   actually normalized — and the patch was corrected to the nested
   form.

2. The nested form was initially written as `${${(@onu)fpath}}` (sort
   + numeric-sort + unique). After review feedback from @mologie on
   the issue thread pointed out that fpath order is semantically
   meaningful to `compinit`'s collision resolution, the patch was
   narrowed to `${${(@u)fpath}}` (dedup only, preserving
   first-occurrence order). This change is explicitly called out so
   reviewers know the final form was arrived at through review, not
   by initial design.

All code references, measurements, and verification steps were
performed locally. The author understands every line of the change,
reviewed the Zsh parameter expansion flags documentation, and can
defend or revise any part on request.

cc @forivall — thank you for the prior branch exploring the
`ZSH_COMPINIT_CACHE` direction. Your analysis was part of the
reasoning process even though this PR takes a different approach.

cc @mologie — thank you for the order-sensitivity review on the
issue. The patch is narrower and more defensible as a result.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -110,9 +110,14 @@ if [[ -z "$ZSH_COMPDUMP" ]]; then
   ZSH_COMPDUMP="${ZDOTDIR:-$HOME}/.zcompdump-${SHORT_HOST}-${ZSH_VERSION}"
 fi
 
-# Construct zcompdump OMZ metadata
+# Construct zcompdump OMZ metadata. fpath is deduplicated (preserving
+# first-occurrence order) before comparison so that sessions which
+# differ only in duplicate counts produce identical metadata. Order is
+# preserved because `compinit` resolves colliding completion files by
+# first-occurrence in fpath, so two fpaths with the same entries in a
+# different order can legitimately produce different completions.
 zcompdump_revision="#omz revision: $(builtin cd -q "$ZSH"; git rev-parse HEAD 2>/dev/null)"
-zcompdump_fpath="#omz fpath: $fpath"
+zcompdump_fpath="#omz fpath: ${${(@u)fpath}}"
 
 # Delete the zcompdump file if OMZ zcompdump metadata changed
 if ! command grep -q -Fx "$zcompdump_revision" "$ZSH_COMPDUMP" 2>/dev/null \


mologie:
I do not endorse or otherwise have anything to do with this PR despite my name appearing here ~~three~~ six times. I think this is a bad idea for reasons outlined in the linked issue.

Also feels bad talking to an LLM and having my points dismissed as "being interesting for a follow-up" - while missing the actual point (fix the root cause in one‘s setup) completely.

carlosala:
See https://github.com/ohmyzsh/ohmyzsh/issues/13720#issuecomment-4366042629

PR #13747, opened 2026-05-16
fix(tmux): respect user's tmux config when fixterm is enabled

### LINKED ISSUE #12247: When oh-my-zsh tmux plugin is enabled, the 'tmux' alias will cause my oh-my-tmux .tmux.conf.local to be ignored
### Describe the bug

I'm using both .oh-my-zsh and .oh-my-tmux.  Without the zsh tmux plugin enabled, all works fine for tmux to parse my .tmux.conf.local from oh-my-tmux.  When I have the zsh tmux plugin enabled, the configuration from .tmux.conf.local is completely ignored UNLESS I have ZSH_TMUX_AUTOSTART=true in my .zshrc.  Disabling the zsh tmux plugin make everything work as expected. 

I did verify that with the tmux plugin enabled if I path to the tmux binary or use any of the aliases that are defined by the tmux EXCEPT for 'tmux', then all works as expected.

### Steps to reproduce

1. Don't the 'tmux' plugin.
2. Install oh-my-tmux
3. cp the .tmux/.tmux.conf.local file to the home directory and enable the theme
4. Exit the shell
5. Start the shell (just to get everything loaded properly)
6. Type 'tmux' to start tmux, you should see the theme that is configured in .tmux.conf.local
7. Enable the oh-my-zsh 'tmux' plugin
8. source / restart your shell
9. Type 'tmux' and you'll see that instead of getting the theme as configured in the .tmux.conf.local, you are getting the default green status bar of tmux.
10. type 'ts' to start with a session and all seems okay.
11. type 'ta' to attach to the session  and all seems okay. Just the 'tmux' command that appears to be broken.

NOTE: If I have ZSH_TMUX_AUTOSTART=true in my .zshrc, then the proper theme is loaded from the .tmux.conf.local file as well.

### Expected behavior

use of oh-my-tmux should be unaffected by the zsh tmux plugin

_No response_

### OS / Linux distribution

MacOS Sonoma

### Zsh version

5.9

### Oh My Zsh version

master (0008534)

### Terminal emulator

kitty, iTerm2

None

### Additional context

_No response_

MD-Mushfiqur123:
## Description

When the tmux plugin is enabled with `ZSH_TMUX_FIXTERM=true` (default), the wrapper function uses `if/elif` logic that picks ONLY the fixed config (`tmux.extra.conf` or `tmux.only.conf`) and ignores the user's custom tmux config entirely. This causes custom configurations (e.g. from oh-my-tmux with `.tmux.conf.local`) to be disregarded unless `ZSH_TMUX_AUTOSTART=true`.

### Root Cause

Lines 146-150 in `tmux.plugin.zsh`:
```zsh
if [[ "$ZSH_TMUX_FIXTERM" == "true" ]]; then
  tmux_cmd+=(-f "$_ZSH_TMUX_FIXED_CONFIG")
elif [[ -e "$ZSH_TMUX_CONFIG" ]]; then
  tmux_cmd+=(-f "$ZSH_TMUX_CONFIG")
fi
```

When `ZSH_TMUX_FIXTERM=true` (the default), only the fixed config is passed via `-f`. The user's config at `$ZSH_TMUX_CONFIG` is completely skipped.

### Fix

1. Always pass the user's config first (when it exists) via `-f`, then apply the fixed config on top.
2. Always export `ZSH_TMUX_CONFIG` so tmux can read it via `$ZSH_TMUX_CONFIG` in `tmux.extra.conf` (which does `source-file $ZSH_TMUX_CONFIG`).

Fixes #12247

### DIFF
--- a/plugins/tmux/tmux.plugin.zsh
+++ b/plugins/tmux/tmux.plugin.zsh
@@ -100,8 +100,12 @@ fi
 0="${${(M)0:#/*}:-$PWD/$0}"
 
 # Set the correct local config file to use.
-if [[ "$ZSH_TMUX_ITERM2" == "false" && -e "$ZSH_TMUX_CONFIG" ]]; then
+# Always export ZSH_TMUX_CONFIG so tmux can source it when processing
+# the fixed config (tmux.extra.conf) via $ZSH_TMUX_CONFIG.
+if [[ -e "$ZSH_TMUX_CONFIG" ]]; then
   export ZSH_TMUX_CONFIG
+fi
+if [[ "$ZSH_TMUX_ITERM2" == "false" && -e "$ZSH_TMUX_CONFIG" ]]; then
   export _ZSH_TMUX_FIXED_CONFIG="${0:h:a}/tmux.extra.conf"
 else
   export _ZSH_TMUX_FIXED_CONFIG="${0:h:a}/tmux.only.conf"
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

Issue #13763, opened 2026-05-23
updater: tag-based sync uses the configured branch and misses add/delete-only updates


fallintoplace:
## Summary

I found three correctness bugs in `.github/workflows/dependencies/updater.py` that all affect the same vendoring pipeline:

1. tag-tracked dependencies can update the recorded version to a tag while copying files from the configured branch head instead of that tag
2. updates that only add new files can be treated as "clean" and skipped
3. upstream deletions are not propagated because files are overlaid instead of synced delete-aware

I think these belong in one issue because they are all in the same fetch/copy/detect-changes flow.

## Prior art / history

I checked the recent history around this updater before filing:

- `#12109` introduced the dependency updater
- `#12756` made commit subjects use the tag version when available
- `#13454` adjusted the logic to only open a PR if there are relevant changes

That history matches the intent, but there still seem to be correctness gaps in the actual sync path.

## 1. Tag-based updates copy the configured branch, not the resolved tag

In `update_or_notify()`, tag-tracked dependencies are detected with `version.startswith("tag:")`, and `check_newer_tag()` returns a newer tag. `new_version` is then set to that tag version.

However, `__apply_upstream_changes()` is called without any resolved ref, and inside that method the clone step still uses `self.values["branch"]`:

- `update_or_notify()` resolves the new tag and then calls `self.__apply_upstream_changes()`
- `__apply_upstream_changes()` reads `branch = self.values["branch"]`
- `Git.clone()` then clones `-b <branch>`

This means the workflow can:

- update `.github/dependencies.yml` to `tag:<new tag>`
- title the PR/commit as an update to that tag
- but vendor files from branch HEAD instead of the tag itself

That looks especially real for current entries such as:

- `plugins/gitfast` (`branch: master`, `version: tag:v2.2`)
- `plugins/wd` (`branch: master`, `version: tag:v0.10.1`)

### Expected

If the updater decides the new version is tag `vX.Y.Z`, the copied files should come from that exact tag/ref.

### Suggested fix

Pass the resolved ref into the copy step. For tag-based dependencies, clone/checkout the tag; for branch/SHA tracking, use the compared branch/SHA explicitly.

## 2. Add-only updates can be skipped because untracked files are ignored

After copying files, the updater decides whether anything changed via `Git.repo_is_clean()`.

Right now that helper uses only:

```bash
git diff --exit-code
```

That ignores untracked files.

[...]

PR #13764, opened 2026-05-23
ci(deps): use resolved tag when syncing dependencies

fallintoplace:
## Summary

This fixes tag-tracked dependency updates so the vendored files are copied from the resolved tag, not from the dependency's configured branch.

## Root cause

The updater already detects tag-tracked dependencies and resolves a newer tag version, but the copy step still cloned `self.values["branch"]`.

That let the workflow:
- update `.github/dependencies.yml` to a new `tag:<version>`
- describe the update as that tag in the commit and PR
- but vendor files from branch HEAD instead of the tag itself

## What changed

- compute a `source_ref` during update processing
- pass the resolved tag to `__apply_upstream_changes()` for tag-tracked dependencies
- keep using the configured branch for branch-tracked dependencies
- add unit coverage for both tag-tracked and branch-tracked ref selection

## Validation

- `/tmp/ohmyzsh-updater-tests/bin/python /Users/hoangvu/Code/OSS/ohmyzsh/.github/workflows/dependencies/test_updater_ref_selection.py`
- `/tmp/ohmyzsh-updater-tests/bin/python -m py_compile .github/workflows/dependencies/updater.py .github/workflows/dependencies/test_updater_ref_selection.py`

Related to #13763.


carlosala (dismissed):
I like it! Could you remove the tests though in another commit? We don't have test runners for python at the moment, let's leave it here in the PR just in case we need them in the future.

fallintoplace:
@carlosala I will take a look later today. Thank you for your attention.

mcornella (approved):
LGTM

carlosala:
Thanks!

PR #13765, opened 2026-05-23
ci(deps): detect add-only vendored changes

fallintoplace:
## Summary

This makes the dependency updater treat add-only vendoring updates as real changes by including untracked files in the repository cleanliness check.

## Root cause

The updater used `git diff --exit-code` to decide whether the repository was clean after copying vendored files.

That misses untracked files, so an upstream update that only adds new files can be copied into the tree and still be treated as "clean", which skips the YAML update, commit, and PR creation.

## What changed

- replace the `git diff --exit-code` cleanliness check with `git status --porcelain --untracked-files=normal`
- treat any porcelain output as a dirty repository
- add unit coverage for both a clean repo and a repo with an untracked file

## Validation

- `/tmp/ohmyzsh-updater-tests/bin/python /Users/hoangvu/Code/OSS/ohmyzsh/.github/workflows/dependencies/test_updater_repo_is_clean.py`
- `/tmp/ohmyzsh-updater-tests/bin/python -m py_compile .github/workflows/dependencies/updater.py .github/workflows/dependencies/test_updater_repo_is_clean.py`

Related to #13763.


carlosala (commented):
Hey! Could you remove the tests (we don't have test runner for python at the moment, we might take them in the future) and review copilot's comments?

fallintoplace:
@carlosala I will take a look later today. Thank you for your attention.

carlosala (approved):

carlosala:
Thanks!

PR #13777, opened 2026-05-28
fix: escape % characters in all git prompts

mcornella:
## Standards checklist:

## Changes:

This patch adds missing % character escaping for custom git prompts used in a few themes. It also includes escaping for git-prompt.sh.

In combination with CVE-2021-45444, this could allow code execution when displaying branch information in cloned malicious git repositories. However, zsh 5.8.1 and newer are largely the default zsh versions, and on those supported distributions with older zsh versions, the CVE has been found to be also patched.

For this reason, this doesn't qualify as a security patch, but a bug fix for proper printing of git branches.


carlosala (dismissed):
LGTM

carlosala (approved):

mcornella:
Published INFORMATIONAL level advisory: https://github.com/ohmyzsh/ohmyzsh/security/advisories/GHSA-x96c-8w82-wf96

PR #13780, opened 2026-05-29
fix(brew): add sbin to PATH

### LINKED ISSUE #13704: "Homebrew's "sbin" was not found in your PATH" after installing brew plugin
### Describe the bug

The brew plugin does not add `/opt/homebrew/sbin` to PATH.

When running `bdr` (which is an alias used by the OMZ plugin) you get the following error:
```
Warning: Homebrew's "sbin" was not found in your PATH but you have installed
formulae that put executables in /opt/homebrew/sbin.
Consider setting your PATH for example like so:
  echo 'export PATH="/opt/homebrew/sbin:$PATH"' >> ~/.zshrc
```

### Steps to reproduce

1. Enable the `brew` plugin and reload shell
2. Install a formula that requires /opt/homebrew/sbin (such as `brew install mtr`
3. Run the `mtr` command
4. Run the `bdr` command

### Expected behavior

The `mtr` application should run as expected.

Running `bdr` should _not_ return the following error:
```
Warning: Homebrew's "sbin" was not found in your PATH but you have installed
formulae that put executables in /opt/homebrew/sbin.
Consider setting your PATH for example like so:
  echo 'export PATH="/opt/homebrew/sbin:$PATH"' >> ~/.zshrc
```

_No response_

### OS / Linux distribution

MacOS 26.4.1 (25E253)

### Zsh version

5.9

### Oh My Zsh version

master (349b9e4)

### Terminal emulator

iTerm2

None

### Additional context

_No response_

Ininsico:
## Description

Some Homebrew formulae (e.g. \mtr\) install executables to \/opt/homebrew/sbin\ (Apple Silicon) or \/usr/local/sbin\ (Intel). Homebrew's \shellenv\ only adds \in\ directories to PATH, so \rew doctor\ warns that \sbin\ is missing when these formulae are installed.

This fix adds \\/sbin\ to PATH when:
- The \sbin\ directory exists under the Homebrew prefix
- It's not already present in PATH (avoids duplicates)

## Related Issue
Fixes #13704

## Verification Steps
1. Enable the \rew\ plugin in \.zshrc\ and reload the shell
2. Install a formula that puts executables in sbin: \rew install mtr\
3. Run \dr\ (brew doctor) — the sbin warning should no longer appear

### DIFF
--- a/plugins/brew/README.md
+++ b/plugins/brew/README.md
@@ -17,6 +17,12 @@ If `brew` is not found in the PATH, this plugin will attempt to find it in commo
 In case you installed `brew` in a non-common location, you can still set `BREW_LOCATION` variable pointing to
 the `brew` binary before sourcing `oh-my-zsh.sh` and it'll set up the environment.
 
+### sbin directory
+
+This plugin also adds `$HOMEBREW_PREFIX/sbin` to the PATH if the directory exists and isn't already present.
+Some Homebrew formulae (e.g. `mtr`) install executables to `sbin`, which `brew doctor` checks for. This
+ensures the `bdr` alias runs without warnings.
+
 ## Aliases
 
 | Alias    | Command                                 | Description                                                           |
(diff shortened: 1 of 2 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #13789, opened 2026-06-03
add alias to make signed commit with verbose output

lavigne958:
## Standards checklist:

## Changes:

- add a new git alias that creates a signed commit with verbose output

## Other comments:

The new alias is very useful, when creating a new git commit, (I don't have exact numbers) a majority of the industry is using signed commits (when you work for a company at least). and creating a signed commit being the standard in most companies, you regularly want to have the `--verbose` output of it, you can already see that the default git commit alias includes the `--verbose` option. 

So adding this `--verbose` option to the signed commit is a standard to me to match what a non signed commit behavior has.

and I use it every days, for now, years. I hope others do too and will benefit from it !

thank you for your consideration.


carlosala:
I wouldn't merge this one. Doing `gcss -v` doesn't add many keystrokes, and we can keep as minimal as possible the plugin.
Thanks for your contribution!

PR #13796, opened 2026-06-06
fix(deno): remove deprecated aliases and add modern ones

fauzan171:
## Summary

This PR updates the Deno plugin to remove deprecated commands and add aliases for newer Deno features.

### Removed aliases
- `db` (`deno bundle`): The `bundle` subcommand was deprecated in Deno 1.x and removed in Deno 2.0. Users should use `deno compile` or external bundlers instead.
- `dru` (`deno run --unstable`): The `--unstable` flag has been deprecated in favor of granular `--unstable-*` flags (e.g., `--unstable-worker-options`).

### Added aliases
- `dck` (`deno check`): Type-check modules without executing them.
- `dsv` (`deno serve`): Run an HTTP server (introduced in Deno 1.37, stabilized in Deno 2.0).

### Updated
- README table to reflect the alias changes.

## Checklist
- [x] The plugin aliases have been tested (manually verified alias expansions)
- [x] README has been updated accordingly

PR #13800, opened 2026-06-06
fix(deno): move guard check before aliases and remove deprecated commands

fauzan171:
## Summary

This PR fixes two issues in the deno plugin:

1. **Guard check placement**: The command existence check (`if (( ! $+commands[deno] ))`) was placed after the aliases, meaning aliases like `db`, `dc`, `drn`, etc. were created even when deno was not installed. This is inconsistent with other plugins (bun, volta, gh, etc.) that check for the command first.

2. **Removed deprecated aliases**:
   - `db` → `deno bundle`: The `bundle` command was deprecated in Deno 1.x and removed in Deno 2.0 (see https://docs.deno.com/runtime/manual/tools/bundler)
   - `dru` → `deno run --unstable`: The blanket `--unstable` flag was replaced by granular `--unstable-*` feature flags in Deno 2.0

## Changes

- Moved the `if (( ! ${+commands[deno]} ))` guard to the top of the plugin file
- Removed the `db` alias (`deno bundle` - removed in Deno 2.0)
- Removed the `dru` alias (`deno run --unstable` - replaced by granular flags)
- Updated README to reflect the removed aliases and added standard usage instructions

## Testing

- Verified the plugin loads correctly when deno is installed
- Verified no aliases are created when deno is not installed

### DIFF
--- a/plugins/deno/README.md
+++ b/plugins/deno/README.md
@@ -2,19 +2,23 @@
 
 This plugin sets up completion and aliases for [Deno](https://deno.land).
 
+To use it, add `deno` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... deno)
+```
+
 ## Aliases
 
-| Alias | Full command        |
-| ----- | ------------------- |
-| db    | deno bundle         |
-| dc    | deno compile        |
-| dca   | deno cache          |
-| dfmt  | deno fmt            |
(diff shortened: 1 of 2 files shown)


carlosala (approved):

carlosala:
This was basically done in #13796

PR #13799, opened 2026-06-06
feat(terraform): add missing commonly-used aliases

fauzan171:
## Summary

This PR adds aliases for frequently used Terraform commands that were missing from the plugin.

### New aliases

| Alias  | Command                           | Use case                              |
|--------|-----------------------------------|---------------------------------------|
| `tffck` | `terraform fmt -check -recursive` | Check formatting in CI pipelines    |
| `tfg`   | `terraform graph`                  | Visualize resource dependency graph |
| `tfib`  | `terraform init -backend=false`    | Initialize without backend config   |
| `tfip`  | `terraform import`                 | Import existing infrastructure      |
| `tfpo`  | `terraform plan -out tfplan`       | Save plan to file for apply         |
| `tfpr`  | `terraform providers`              | List provider dependencies          |
| `tfr`   | `terraform refresh`                | Refresh state from real resources   |
| `tfsr`  | `terraform state rm`               | Remove resource from state          |
| `tfwn`  | `terraform workspace new`          | Create a new workspace              |

### Other changes
- Sorted aliases alphabetically in the plugin file for easier maintenance
- Updated README with all new aliases

These are all standard Terraform CLI commands that are commonly used in day-to-day infrastructure management workflows.

PR #13801, opened 2026-06-06
fix(terraform): add command guard and document missing aliases

fauzan171:
## Summary

This PR adds a command existence guard check to the terraform plugin and documents two missing aliases in the README.

### Changes

1. **Added command guard check**: The terraform plugin was missing a command existence check at the top. This means all aliases (`tf`, `tfa`, `tfp`, etc.) and functions (`tf_prompt_info`, `tf_version_prompt_info`) were created even when terraform was not installed. This is inconsistent with other plugins like `bun`, `volta`, `gh`, `helm`, etc. that all guard with `if (( ! ${+commands[terraform]} )); then return; fi`.

2. **Documented missing aliases in README**: The `tfapp` (`terraform apply tfplan`) and `tfpo` (`terraform plan -out tfplan`) aliases exist in the plugin but were missing from the README documentation table.

## Testing

- Verified the plugin loads correctly when terraform is installed
- Verified no aliases/functions are created when terraform is not installed
- Confirmed `tfapp` and `tfpo` aliases match their documented commands

### DIFF
--- a/plugins/terraform/README.md
+++ b/plugins/terraform/README.md
@@ -21,6 +21,7 @@ plugins=(... terraform)
 | `tfa`   | `terraform apply`                      |
 | `tfa!`  | `terraform apply -auto-approve`        |
 | `tfap`  | `terraform apply -parallelism=1`       |
+| `tfapp` | `terraform apply tfplan`               |
 | `tfc`   | `terraform console`                    |
 | `tfd`   | `terraform destroy`                    |
 | `tfd!`  | `terraform destroy -auto-approve`      |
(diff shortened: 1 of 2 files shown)


mcornella:
Hi @fauzan171, thank you for the contribution. 

I'll close this as won't do, and the reason is that depending on how people are loading terraform, it's possible that the command won't be in `$PATH` when the plugin loads (for example if using `mise`, and it being loaded later on). 

The documentation of missing aliases will be fixed in #13799. Thanks!

PR #13814, opened 2026-06-12
feat(updater): add cooldown option to delay applying updates (#13813)

### LINKED ISSUE #13813: Introduce a cooldown feature for the auto-updater

Core updater tools

I've been thinking about supply-chain attacks and reflecting on possible approaches to providing users with a little more protection. Currently, users can configure the frequency of the auto-updater, which is more of a UX feature in that it'll let you wait a few weeks before it prompts you to consider updating. At face value, that seems like you're not fetching the latest, all the time... but the updater will actually fetch the most recent code in our master branch right up to the second you say YES.

It'd be nice to offer users the ability to only fetch merged code as X day(s) ago should they and/or their organization require them to wait a little longer.

Introduce a configuration option that will only pull in code changes as of X days ago, defaulting to zero days. Add documentation that explains this. 

_No response_

### Additional context

_No response_

### Related Issues

_No response_

robbyrussell:
Hey folks... wanted to share some context on the approach here before you dig in.

My first instinct was to use `git rebase `to land on a specific historical commit, since the existing updater already uses `--rebase`. But I got a bit nervous about introducing a bare `git rebase <SHA>` into the workflow. It's less predictable than rebasing to a branch tip, and if a user's local copy somehow ended up ahead of the cooldown ref (e.g. they'd manually updated previously), we'd be attempting to rebase onto an ancestor which could leave things in a weird state.

Instead I went with `git fetch` + `git merge --ff-only` to the target commit. Fast-forward only means if you're already at or past the cooldown ref, it's a clean no-op... nothing moves. It also feels closer in spirit to what `git pull` does, just with a specific commit as the target rather than the branch tip.

One other thing worth calling out: I'm filtering by committer timestamp (`%ct`) rather than author date. Since our workflow is PR based, the committer date is when GitHub actually merged the code which is what we care about for "how old is this change really." Author dates are self-reported and could be backdated.

Would love to see some folks help us kick the tires on this before we mark it ready. Open to feedback!

## Standards checklist:

## Changes:

- Introduces a `zstyle ':omz:update' cooldown <days>` setting that limits the
  updater to only apply commits that are at least N days old. Defaults to 0
  (current behavior — always pull latest).
- When cooldown is active, the updater fetches the remote branch, finds the most
  recent commit whose committer timestamp is at least N days old, and applies it
  via `git merge --ff-only`. If the local copy is already at or past the cooldown
  ref, nothing changes.
- `tools/upgrade.sh`: reads cooldown zstyle; replaces `git pull --rebase` with
  `git fetch` + `git merge --ff-only` when cooldown > 0
- `README.md`: documents the new setting under "Getting Updates"
- `templates/zshrc.zsh-template`: adds a commented-out cooldown example alongside
  frequency, with rephrased comments so the two read as a pair

## Other comments:

Claude Code was used to assist with the implementation of this feature.

Closes #13813

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -47,6 +47,26 @@ function is_update_available() {
   remote=${"$(builtin cd -q "$ZSH"; git config --local oh-my-zsh.remote)":-origin}
   remote_url=$(builtin cd -q "$ZSH"; git config remote.$remote.url)
 
+  local cooldown_days
+  zstyle -s ':omz:update' cooldown cooldown_days || cooldown_days=0
+  [[ $cooldown_days == <-> ]] || cooldown_days=0
+
+  if (( cooldown_days > 0 )); then
+    local local_head cutoff_epoch cooldown_ref
+    local_head=$(builtin cd -q "$ZSH"; git rev-parse $branch 2>/dev/null) || return 0
+    (builtin cd -q "$ZSH"; LANG= git fetch --quiet $remote $branch) || return 1
+    zmodload zsh/datetime
+    cutoff_epoch=$(( EPOCHSECONDS - cooldown_days * 86400 ))
+    cooldown_ref=$(builtin cd -q "$ZSH"; git log --format="%H %ct" FETCH_HEAD \
+      | awk -v c="$cutoff_epoch" '$2 <= c { print $1; exit }')
+    [[ -n "$cooldown_ref" ]] || return 1
+    [[ "$cooldown_ref" != "$local_head" ]] || return 1
+    if (builtin cd -q "$ZSH"; git merge-base --is-ancestor "$cooldown_ref" "$local_head" 2>/dev/null); then
+      return 1
+    fi
+    return 0
+  fi
+
   local repo
   case "$remote_url" in
   https://github.com/*) repo=${${remote_url#https://github.com/}%.git} ;;
@@ -109,8 +129,10 @@ EOD
 }
 
 function update_ohmyzsh() {
-  local verbose_mode
+  local verbose_mode cooldown_days
   zstyle -s ':omz:update' verbose verbose_mode || verbose_mode=default
+  zstyle -s ':omz:update' cooldown cooldown_days || cooldown_days=0
+  [[ $cooldown_days == <-> ]] || cooldown_days=0
 
   # Force verbose mode to silent if p10k instant prompt is enabled
   if [[ ${POWERLEVEL9K_INSTANT_PROMPT:-off} != "off" ]]; then
@@ -118,13 +140,13 @@ function update_ohmyzsh() {
   fi
 
   if [[ "$update_mode" != background-alpha ]] \
-    && LANG= ZSH="$ZSH" zsh -f "$ZSH/tools/upgrade.sh" -i -v $verbose_mode; then
+    && LANG= ZSH="$ZSH" zsh -f "$ZSH/tools/upgrade.sh" -i -v $verbose_mode -c $cooldown_days; then
     update_last_updated_file
     return $?
   fi
 
   local exit_status error
-  if error=$(LANG= ZSH="$ZSH" zsh -f "$ZSH/tools/upgrade.sh" -i -v silent 2>&1); then
+  if error=$(LANG= ZSH="$ZSH" zsh -f "$ZSH/tools/upgrade.sh" -i -v silent -c $cooldown_days 2>&1); then
     update_last_updated_file 0 "Update successful"
   else
     exit_status=$?
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -57,7 +57,7 @@ function is_update_available() {
     (builtin cd -q "$ZSH"; LANG= git fetch --quiet $remote $branch) || return 1
     zmodload zsh/datetime
     cutoff_epoch=$(( EPOCHSECONDS - cooldown_days * 86400 ))
-    cooldown_ref=$(builtin cd -q "$ZSH"; git log --format="%H %ct" FETCH_HEAD \
+    cooldown_ref=$(builtin cd -q "$ZSH"; git log --first-parent --format="%H %ct" FETCH_HEAD \
       | awk -v c="$cutoff_epoch" '$2 <= c { print $1; exit }')
     [[ -n "$cooldown_ref" ]] || return 1
     [[ "$cooldown_ref" != "$local_head" ]] || return 1
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -231,6 +231,10 @@ local ret=0
 remote=${"$(git config --local oh-my-zsh.remote)":-origin}
 branch=${"$(git config --local oh-my-zsh.branch)":-master}
 
+# cooldown: minimum age (in days) of commits to apply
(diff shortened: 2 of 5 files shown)


tools/upgrade.sh
```diff
@@ -231,6 +231,10 @@ local ret=0
 remote=${"$(git config --local oh-my-zsh.remote)":-origin}
 branch=${"$(git config --local oh-my-zsh.branch)":-master}
 
+# cooldown: minimum age (in days) of commits to apply
+local cooldown_days
+zstyle -s ':omz:update' cooldown cooldown_days || cooldown_days=0
```
mcornella: Since `tools/upgrade.sh` gets called in a forked command from `check_for_upgrade.sh` (it's called directly with `zsh tools/upgrade.sh` and some args), the zstyle value will not be propagated from the user's set value in .zshrc.

We need to make this test in `check_for_upgrade.sh` and pass a specific argument to this script, similarly as we do for the verbosity option.

tools/upgrade.sh
```diff
@@ -242,7 +246,20 @@ last_commit=$(git rev-parse "$branch")
 if [[ $verbose_mode != silent ]]; then
   printf "${BLUE}%s${RESET}\n" "Updating Oh My Zsh"
 fi
-if LANG= git pull --quiet --rebase $remote $branch; then
+if {
```
mcornella: Let's extract this logic into a separate function, this looks weird

tools/upgrade.sh
```diff
@@ -242,7 +246,20 @@ last_commit=$(git rev-parse "$branch")
...
+  if (( cooldown_days > 0 )); then
+    zmodload zsh/datetime
```
mcornella: Ideally let's declare module loads at the top of the script.

tools/upgrade.sh
```diff
@@ -242,7 +246,20 @@ last_commit=$(git rev-parse "$branch")
...
+    local cutoff_epoch cooldown_ref
+    cutoff_epoch=$(( EPOCHSECONDS - cooldown_days * 86400 ))
+    LANG= git fetch --quiet $remote $branch && {
+      cooldown_ref=$(git log --format="%H %ct" "$remote/$branch" \
+        | awk -v c="$cutoff_epoch" '$2 <= c { print $1; exit }')
+      [[ -z "$cooldown_ref" ]] || LANG= git merge --ff-only --quiet "$cooldown_ref"
```
mcornella: This order of operations can also be improved. Putting this into its own function will make this easier as well.

We'd also need to return some error code if this fails, so that the outer if gets computed correctly, as we'd do in the default case if no cooldown is set and `git pull` is called instead.

tools/upgrade.sh
```diff
@@ -242,7 +246,20 @@ last_commit=$(git rev-parse "$branch")
...
+    }
+  else
+    LANG= git pull --quiet --rebase $remote $branch
+  fi
+}; then
   # Check if it was really updated or not
   if [[ "$(git rev-parse HEAD)" = "$last_commit" ]]; then
     message="Oh My Zsh is already at the latest version."
```
mcornella: If cooldown is set, and this check passes, it doesn't necessarily mean we're at the latest version; probably we're N days behind the version, which may be several commits, but still we "did not move to newer commits". I believe we can put this check separately in each branch of the decision (with cooldown set vs. using the default), and then the logic will become clearer plus we can give slightly different messages:

> already at the latest version

or

> already at a version N days old

or some message like this

mcornella (changes requested):
I think this script is long overdue for a good refactoring and style cleanup, but I think these suggestions alone will be enough to merge this feature ✨

robbyrussell:
Thanks for the review, @mcornella — I think this round covers your notes.

- Cooldown is read in `check_for_upgrade.sh` / `omz update` and passed into `upgrade.sh` as `-c`, same pattern as verbosity. `upgrade.sh` no longer reads the zstyle itself.
- The nested `if {` is now `perform_update` / `update_with_cooldown`, with fetch/merge failures returned to the outer `if`.
- `zmodload zsh/datetime` is at the top of `upgrade.sh`.
- Already-up-to-date copy is split: latest vs the actual age of HEAD when cooldown is set.
- Auto-update only prompts when a cooldown-eligible commit would actually move HEAD, so you shouldn't get an empty "update" every cycle.
- Direct `upgrade.sh` users can pass `-c`; that's documented next to the zstyle.

Would you take another look when you have a minute?

robbyrussell:
@copilot review

The latest commit addresses the previous review by restricting both cooldown-ref searches to the remote branch’s first-parent history.

tools/check_for_upgrade.sh
```diff
@@ -47,6 +47,26 @@ function is_update_available() {
...
+
+  if (( cooldown_days > 0 )); then
+    local local_head cutoff_epoch cooldown_ref
+    local_head=$(builtin cd -q "$ZSH"; git rev-parse $branch 2>/dev/null) || return 0
+    (builtin cd -q "$ZSH"; LANG= git fetch --quiet $remote $branch) || return 1
+    zmodload zsh/datetime
+    cutoff_epoch=$(( EPOCHSECONDS - cooldown_days * 86400 ))
+    cooldown_ref=$(builtin cd -q "$ZSH"; git log --format="%H %ct" FETCH_HEAD \
+      | awk -v c="$cutoff_epoch" '$2 <= c { print $1; exit }')
+    [[ -n "$cooldown_ref" ]] || return 1
```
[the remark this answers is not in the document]
mcornella: @copilot this line uses `git log --first-parent` which only returns commits belonging actually to the remote branch, ignoring commits from merged branches. 

Sources:
- an example blog post from 2015: https://marcgg.com/blog/2015/08/04/git-first-parent-log/
- official docs for first-parent: https://git-scm.com/docs/git-log#Documentation/git-log.txt---first-parent 

Confirm this information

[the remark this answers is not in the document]
mcornella: Confirmed this offline, not applicable to the code here

mcornella (approved):
LGTM. 

Small note: we're using commit date as a source of truth for the cooldown. This date is spoofable, but since direct push to master is forbidden, and all updates happen via Pull Requests strictly based on our branch protections, this commit date will always be written by GitHub (trustable entity).

robbyrussell:
Merged, thanks for the reviews!

For anyone landing here later... a bit about what this does and why you might want it.

**Why:** Oh My Zsh ships straight from `master`. If you'd rather let others kick the tires on new commits for a few days before they reach your machine, `cooldown` does that. You still get everything, just a little later.

**How to enable:**

1. Run `omz update` to pick up this change (this first update pulls latest, same as before).
2. Add this to your `.zshrc`:
   ```sh
   zstyle ':omz:update' cooldown 10   # only apply commits at least 10 days old
   ```
3. Open a new shell. From then on, both `omz update` and the auto-updater stop at the newest commit that's at least that many days old.

If you call `$ZSH/tools/upgrade.sh` directly, pass `-c 10` instead... that script doesn't read your `.zshrc`.

**Feedback welcome.** This touches the updater, so if anything looks off... stuck on "up to date" when it shouldn't be, an update prompt with nothing to apply, a fetch or merge error... please open an issue and mention #13814. The output of `omz update` plus your cooldown value is all we'd need.

PR #13825, opened 2026-06-16
perf(kubectx)!: load context asynchronously

olegg-oai:
## Standards checklist:

## Changes:

- Run `kubectl config current-context` through the existing Oh My Zsh async
  prompt API instead of blocking every prompt render.
- Register the async handler only when `kubectx_prompt_info` is referenced
  directly by a prompt variable.
- Preserve synchronous behavior on unsupported zsh versions and through the
  `async-prompt no` setting.
- Support indirect prompt wrappers through the `async-prompt force` setting.
- Keep the existing context mapping and prompt-escaping behavior unchanged.
- Handle the period before the first async result without errors when
  `nounset` is enabled.

## Other comments:

The current implementation starts `kubectl` synchronously every time the
prompt is rendered. With a large kubeconfig, this measured approximately 129
ms per prompt on macOS. The proposed renderer takes approximately 0.71 ms in
the foreground; the existing async callback updates and redraws the prompt
when `kubectl` finishes in the background.

This follows the implementation and compatibility behavior already used by
the async Git prompt. The deferred registration hook runs before
`_omz_async_request` and removes itself after the initial prompt inspection.

This changes `kubectx_prompt_info` to asynchronous behavior by default on
supported zsh versions. Users who require the previous synchronous behavior
can configure:

```zsh
zstyle ':omz:alpha:plugins:kubectx' async-prompt no
```

Themes that call `kubectx_prompt_info` indirectly can configure:

```zsh
zstyle ':omz:alpha:plugins:kubectx' async-prompt force
```

Validated locally with:

- default deferred async registration and callback
- forced async registration
- synchronous fallback
- direct and mapped Kubernetes context rendering
- `nounset` before and after async completion
- concurrent Git and kubectx async handlers
- a single deferred hook and correct hook ordering
- `zsh -n plugins/kubectx/kubectx.plugin.zsh`

No aliases are introduced by this change.

AI-assisted: OpenAI Codex assisted with profiling, implementation, test-case
development, repository-history research, and PR preparation. I reviewed and
iterated on the resulting change and validation.

### DIFF
--- a/plugins/kubectx/README.md
+++ b/plugins/kubectx/README.md
@@ -22,6 +22,21 @@ RPS1='$(kubectx_prompt_info)'
 PROMPT="$PROMPT"'$(kubectx_prompt_info)'
 ```
 
+The context is loaded asynchronously on supported versions of zsh so that
+`kubectl` does not block the prompt. To restore synchronous behavior, add this
+before Oh My Zsh is sourced:
+
+```zsh
+zstyle ':omz:alpha:plugins:kubectx' async-prompt no
+```
+
+If your theme calls `kubectx_prompt_info` indirectly through another function,
+force registration of the async handler instead:
+
+```zsh
+zstyle ':omz:alpha:plugins:kubectx' async-prompt force
+```
+
 ### Custom context names
 
 You can rename the default context name for better readability or additional formatting.
--- a/plugins/kubectx/kubectx.plugin.zsh
+++ b/plugins/kubectx/kubectx.plugin.zsh
@@ -1,6 +1,6 @@
 typeset -g -A kubectx_mapping
 
-function kubectx_prompt_info() {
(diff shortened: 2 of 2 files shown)


olegg-oai:
@mcornella any chance you can take a look at this PR/point to someone else who can?
Thank you!

olegg-oai:
cc @carlosala sorry for tagging, could you please suggest on how to get this PR reviewed?
Thank you!

olegg-oai:
@carlosala @mcornella any chance to get it reviewed?
Thank you!

mcornella (approved):
It works, but I'll add the disclaimer to use the new function name if called somewhere else not in the prompt.

mcornella:
Thank you for your patience!

PR #13836, opened 2026-06-24
feat(ai-session-search): add plugin

jarvan1:
## Standards checklist:

- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below. (N/A — no aliases)

## Changes:

Adds a new `ai-session-search` plugin: an fzf-style fuzzy picker over the
session histories of several AI CLIs — **Claude Code, Codex, Copilot CLI, and
Gemini CLI**.

- Bound to <kbd>Ctrl-X</kbd> <kbd>Ctrl-W</kbd> (chosen so it does **not** clobber
  zsh's built-in <kbd>Ctrl-W</kbd> backward-kill-word); also runnable as the
  `ai-sessions` command with an optional initial filter term.
- In the picker you can fuzzy-filter by tool, preview the first real user message
  of each session, open the raw `.jsonl` in `$EDITOR`, rescan, and — on Enter —
  **resume the session in its original working directory**.
- Each provider's sessions are read directly from disk with `jq`; resume dispatches
  to the right CLI (`claude --resume`, `codex resume`, `copilot --resume=`, or just
  re-opening the dir for gemini-cli, which has no resume-by-id).
- Scan locations and a few behaviors are configurable via `AISS_*` env vars
  (documented in the plugin README).

### Why

There's no existing Oh My Zsh plugin that browses/resumes AI CLI session
histories (I searched the `plugins/` tree — the closest by name, `fzf` and
`zsh-interactive-cd`, operate on files/cd, not AI sessions). As people use these
CLIs across many project directories, finding and resuming the right past session
becomes painful; this fills that gap.

### Requirements

Runtime deps are `fzf` (the picker) and `jq` (parsing); the plugin checks for them
lazily and prints a single actionable hint if missing, without nagging on every
shell startup. The individual AI CLIs are only needed to actually *resume* a
session — browsing/previewing work without them.

### Testing

- `zsh -n` passes on both `ai-session-search.plugin.zsh` and `aiss-helper.zsh`.
- Verified end-to-end against real local session data for all four providers:
  scanning, the preview pane (header + colored USER/ASSISTANT turns), and the
  generated resume command (correct `cd` into the original dir + provider-specific
  resume invocation).

## Other comments:

**AI-assisted disclosure (per CONTRIBUTING.md):** This plugin was developed with
significant help from Claude (Anthropic's Claude Code). The AI assisted with
writing the zsh widget, the `jq` filters for each session format, and the README.
I directed the design, reviewed every line, debugged the behavior on my own
machine (including a tricky fzf-preview hang caused by using `read -r -d ''` for a
heredoc inside fzf's background process group — fixed by switching to
`$(cat <<'JQ' …)`), and tested it against my actual session histories. I
understand how it works and can explain/maintain any part of it.

No new aliases are introduced.

### DIFF
--- a/plugins/ai-session-search/README.md
+++ b/plugins/ai-session-search/README.md
@@ -1,0 +1,78 @@
+# ai-session-search
+
+fzf-style fuzzy search across the session histories of multiple AI CLIs:
+**Claude Code / Codex / Copilot / Gemini**. Hit a hotkey to pop up a list, type
+`claude` / `codex` / `copilot` / `gemini` to filter by tool, and press Enter to
+`resume` the session right in its original directory.
+
+To use it, add `ai-session-search` to the plugins array in your zshrc file:
+
+```zsh
+plugins=(... ai-session-search)
+```
+
+## Requirements
+
+This plugin requires [`fzf`](https://github.com/junegunn/fzf) (the picker, 0.35+
+recommended) and [`jq`](https://stedolan.github.io/jq/) (to parse the session
+files):
+
+```sh
+brew install fzf jq          # macOS
+sudo apt install fzf jq      # Debian/Ubuntu
+```
+
+Browsing and previewing work with just `fzf` + `jq` — scanning reads the session
+files directly. To actually *resume* a session you also need the matching CLI on
+your `PATH` (`claude` / `codex` / `copilot` / `gemini`).
+
+## Usage
+
+Default keybinding: <kbd>Ctrl-X</kbd> <kbd>Ctrl-W</kbd> (does not clobber zsh's
+built-in <kbd>Ctrl-W</kbd> backward-kill-word). You can also run it as a command:
+`ai-sessions` or `ai-sessions codex` (with an initial filter term).
+
+In the picker:
+
+| Key     | Action |
+|---------|--------|
+| type    | fuzzy-filter (the provider name is part of each row, so typing `claude` narrows to Claude) |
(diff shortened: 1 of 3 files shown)


robbyrussell:
Thanks for this, and sorry it sat so long without an answer.

I'm going to close it, and the reason is the approach rather than the code. The plugin works by reading four vendors' on-disk session formats — Claude Code, Codex, Copilot and Gemini — none of which are documented or stable. When any of them changes its layout this breaks quietly for everyone who has it enabled, and we have no way to notice: we can't ship fixtures for four proprietary formats, and our CI never runs plugin code. That is a maintenance commitment we can't make in the core repo.

For what it's worth, it also isn't working today. Three things I checked:

- `aiss-helper.zsh:242` — `${f:t:r#rollout-*-}` is not a valid zsh modifier and aborts the shell, so any Codex session that reaches that fallback kills the whole scan and silently drops your Copilot and Gemini results.
- `aiss-helper.zsh:261` — the Gemini filter `try (...) // "" catch ""` is a jq syntax error, so Gemini previews are always blank.
- Gemini rows use the hash directory name as the working directory, so resuming one runs `mkdir -p <hash> && cd <hash>` and leaves a junk directory wherever you were.

The idea is a good one, and it is a natural fit for a standalone plugin — people can install it with their plugin manager or drop it in `$ZSH_CUSTOM/plugins`, and you can move as fast as the vendors change their formats instead of waiting on us. Happy to link it from a discussion if you publish it.

Issue #13838, opened 2026-06-24
Minor robustness/cleanup issues in install.sh, completion.zsh, key-bindings.zsh, diagnostics.zsh


hponukumati:
While reviewing the codebase I found a few small but concrete issues. Related to #13193 (OpenSSF SAST score) but these are specific, actionable items not called out there.

1. **`tools/install.sh:54`** — `HOME="${HOME:-$(eval echo ~$USER)}"` combines `eval` with an unquoted `$USER`. Low-risk in practice (env var, not user input) but worth quoting/avoiding `eval`.
2. **`tools/install.sh:347`** — `if [ $OVERWRITE_CONFIRMATION != "no" ]` uses an unquoted variable; breaks if the value is ever empty. Should be `"$OVERWRITE_CONFIRMATION"`.
3. **`tools/install.sh:483`** — `if [ $? -ne 0 ]` appears far from the multi-line `chsh` block it's checking, making it fragile to accidental reordering during future edits. Consider capturing the exit code immediately after the command.
4. **`lib/completion.zsh:1`** — long-standing author comment "fixme - the load process here seems a bit bizarre" with no follow-up; worth either resolving or removing/clarifying.
5. **`lib/diagnostics.zsh:49,167`** and **`lib/git.zsh:145`** — TODO references a CentOS 7 workaround "remove when CentOS 7 reaches EOL." CentOS 7 reached EOL in June 2024, so this is now actionable.
6. **`lib/key-bindings.zsh:128-145`** — ~18 lines of dead, commented-out code that could be removed for clarity.

## Suggested fix
Small, independent PRs per item (quoting fixes in install.sh, drop CentOS 7 workaround, remove dead code in key-bindings.zsh). Happy to submit a PR for any/all of these if maintainers are open to it.

hponukumati:
Disclosure per the contributing guidelines: this issue was drafted with AI assistance (Claude) — I used it to review the codebase and flag these specific lines, then verified them myself before filing.

cc @mcornella — you signed the CentOS 7 TODO in `lib/git.zsh:145` and you're assigned on the related #13193, so flagging this in case it's relevant to either.

carlosala:
Closing. Feel free to open PRs if you want, they are small changes, no need to have a tracking issue. Thanks!

PR #13840, opened 2026-06-25
chore(install): quote vars and defensive programming

hponukumati:
Fixes part of #13838.

- Quote `$USER` in the `eval echo ~$USER` fallback used to determine `$HOME`.
- Quote the `$OVERWRITE_CONFIRMATION` comparison so it doesn't break if the value is ever empty.
- Capture `chsh`'s exit status immediately after each call (in both the `sudo` and non-`sudo` branches) instead of relying on `$?` after an unrelated `if/else`, which is fragile to future edits.

No behavior change for the documented/default flows — these are quoting/robustness fixes.

AI-assisted: drafted with help from Claude (code review identified the lines, fixes were written and verified by me, `sh -n` syntax-checked).


carlosala (approved):

carlosala:
Thanks!

PR #13619, opened 2026-03-06
fix: preserve user .zshrc when no backup exists during uninstall

xin700:
## Description

Fixes #13156
Fixes #13328

Currently, when uninstalling Oh My Zsh, if no original backup (.zshrc.pre-oh-my-zsh) exists, the user's .zshrc is deleted permanently. This happens when users installed Oh My Zsh years ago on a different machine.

## Changes

- When no original backup exists, restore the current .zshrc with Oh My Zsh lines removed
- Use grep to filter out 'source.*oh-my-zsh.sh' lines
- Keep the timestamped backup file for safety

## Impact

Users will no longer lose their .zshrc configuration when uninstalling, even if the original backup doesn't exist.

## Example

Before: .zshrc deleted, no backup available
After: .zshrc restored with Oh My Zsh lines removed

## Standards checklist

- [x] The PR title is descriptive
- [x] The PR doesn't replicate another PR which is already open
- [x] The code follows the code style guide
- [x] The code is stable and tested

PR #13849, opened 2026-07-02
fix(uninstall): prompt before restoring pre-OMZ zshrc backup

### LINKED ISSUE #13156: Uninstalling Oh My Zsh deletes .zshrc
### Describe the bug

I have a .zshrc I keep in a github repo. And before you ask why - it's for good reasons, like the one below.

I share my shell config and aliases among all my computers. I've been using it for the last couple of years, gradually accumulating useful stuff. I use the same setup on many computers - seven computers currently, and many, many others previously.

I have installed Oh My Zsh the very first time I started using zsh several years ago.

Today I tried uninstalling Oh My Zsh.

You can see where this is going; it all went swimmingly well. Because as you might have guessed, I _do not_ have on any of my current computers the backup the Oh My Zsh installer created several years and multiple computers ago.

And frankly, _even_ if I had that backup from several years ago... does anyone really, truly believes restoring it and losing all the config changes since then would be the right thing to do?

And that folks is why I keep my stuff in git.

### Steps to reproduce

1. Install Oh My Zsh.
2. Commit your .zshrc to github repo.
3. Change job, return old computer, and get new computer.
4. Clone the github repo.
5. Observe everything is configured as you expect.
6. Uninstall Oh My Zsh.
7. Observe your .zshrc is obliterated.

### Expected behavior

Do not delete .zshrc, expecting a backup that was created several years and multiple computers ago to exist or if it does, to be relevant and useful.

_No response_

### OS / Linux distribution

Windows 11 / WSL / Ubuntu 24, OS X, Raspbian

### Zsh version

5.9

### Oh My Zsh version

Latest (can't really check, omz is uinstalled)

### Terminal emulator

iTerm2, Warp, xterm, Windows Terminal

WSL2

### Additional context

_No response_

### LINKED ISSUE #13328: uninstall.sh drop all my settings and tokens stored in .zshrc
### Describe the bug

My carelessness and false expectations led to the funniest situation in the last 10 years. (It would have been tragic if I didn’t have a backup.)

After uninstall.sh finished running, I started reading through its log lines. I see the line:
`Found ~/.zshrc -- Renaming to /Users/a0s/.zshrc.omz-uninstalled-2025-09-23_12-41-12`
and I think, great, I can delete it, since it’s just a backup that includes source oh-my-zsh.sh. So I run rm (yes, I rushed). Then I read:
`Found /Users/a0s/.zshrc.pre-oh-my-zsh -- Restoring to ~/.zshrc`
and start feeling a bit confusing, because it’s been a very long time since I first installed oh-my-zsh. And then I open `~/.zshrc` and see an almost empty file. All my configs and tokens have gone to Valhalla. Thanks for the awesome uninstaller! :) 

I fully realize that in the current situation I’m at fault, but overwriting the current config with a file that’s many, many years old also seems unreasonable.

### Steps to reproduce

1. Run `sh ~/.oh-my-zsh/tools/uninstall.sh`

### Expected behavior

It should try to find the line with "source oh-my-zsh.sh" and try to remove it.

```shell
> sh ~/.oh-my-zsh/tools/uninstall.sh
Are you sure you want to remove Oh My Zsh? [y/N] Y
Removing ~/.oh-my-zsh
Found ~/.zshrc -- Renaming to /Users/a0s/.zshrc.omz-uninstalled-2025-09-23_12-41-12
Looking for original zsh config...
Found /Users/a0s/.zshrc.pre-oh-my-zsh -- Restoring to ~/.zshrc
Your original zsh config was restored.
Thanks for trying out Oh My Zsh. It's been uninstalled.
Don't forget to restart your terminal!
```

### OS / Linux distribution

macOS 15.6.1 (24G90)

### Zsh version

5.9

### Oh My Zsh version

unknown

### Terminal emulator

iTerm2

None

### Additional context

_No response_

hponukumati:
Fixes #13156, fixes #13328.

Two independent users lost years of zsh configuration because `uninstall.sh` silently restored `.zshrc.pre-oh-my-zsh` over their current config without any confirmation — even when that backup was created years ago at install time.

## What changed

**Before:** the pre-OMZ backup was restored unconditionally with no prompt, and the timestamped backup of the current config was labelled ambiguously ("Renaming to..."), leading users to delete it thinking it was a throwaway file.

**After:**
- The timestamped backup message now clearly states it contains the user's **current** zsh configuration
- The modification date of the pre-OMZ backup is shown before asking to restore it, so the user knows how old it is
- A `[y/N]` confirmation prompt (defaulting to **N**) is shown before restoring
- If the user declines, both file paths are printed so they can review and merge manually
- Cross-platform date detection: tries macOS `date -r` then GNU `stat -c %Y` with a graceful fallback

## Example output

```
Found ~/.zshrc -- Saving your current config to ~/.zshrc.omz-uninstalled-2026-07-02_10-30-00
  (this file contains your current zsh configuration)
Found ~/.zshrc.pre-oh-my-zsh (last modified: 2023-01-15)
This is the zsh config that existed before Oh My Zsh was installed.
Restore it as ~/.zshrc? [y/N]
```

If the user says N:
```
Skipping restore.

Your files are at:
  Current config : ~/.zshrc.omz-uninstalled-2026-07-02_10-30-00
  Pre-OMZ backup : ~/.zshrc.pre-oh-my-zsh
You can review and merge them manually.
```

A previous attempt at fixing this (#13297) was closed; this PR takes a different, simpler approach — rather than changing what gets restored, it just asks the user first.

AI-assisted: drafted with Claude, reviewed and tested manually, `sh -n` syntax-checked.

### DIFF
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -23,18 +23,40 @@ fi
 
 if [ -e ~/.zshrc ]; then
   ZSHRC_SAVE=~/.zshrc.omz-uninstalled-$(date +%Y-%m-%d_%H-%M-%S)
-  echo "Found ~/.zshrc -- Renaming to ${ZSHRC_SAVE}"
+  echo "Found ~/.zshrc -- Saving your current config to ${ZSHRC_SAVE}"
+  echo "  (this file contains your current zsh configuration)"
   mv ~/.zshrc "${ZSHRC_SAVE}"
 fi
 
 echo "Looking for original zsh config..."
 ZSHRC_ORIG=~/.zshrc.pre-oh-my-zsh
 if [ -e "$ZSHRC_ORIG" ]; then
-  echo "Found $ZSHRC_ORIG -- Restoring to ~/.zshrc"
-  mv "$ZSHRC_ORIG" ~/.zshrc
-  echo "Your original zsh config was restored."
+  ZSHRC_ORIG_DATE=$(date -r "$ZSHRC_ORIG" "+%Y-%m-%d" 2>/dev/null || date -d "@$(stat -c %Y "$ZSHRC_ORIG" 2>/dev/null)" "+%Y-%m-%d" 2>/dev/null || echo "unknown date")
+  echo "Found $ZSHRC_ORIG (last modified: ${ZSHRC_ORIG_DATE})"
+  echo "This is the zsh config that existed before Oh My Zsh was installed."
+  printf "Restore it as ~/.zshrc? [y/N] "
+  read -r restore_confirmation
+  if [ "$restore_confirmation" = y ] || [ "$restore_confirmation" = Y ]; then
+    mv "$ZSHRC_ORIG" ~/.zshrc
+    echo "Your original zsh config was restored to ~/.zshrc."
+    if [ -n "$ZSHRC_SAVE" ]; then
+      echo "Your Oh My Zsh config is saved at ${ZSHRC_SAVE} if you need it."
+    fi
+  else
+    echo "Skipping restore."
+    if [ -n "$ZSHRC_SAVE" ]; then
+      echo ""
+      echo "Your files are at:"
+      echo "  Current config : ${ZSHRC_SAVE}"
+      echo "  Pre-OMZ backup : ${ZSHRC_ORIG}"
+      echo "You can review and merge them manually."
+    fi
+  fi
 else
-  echo "No original zsh config found"
+  echo "No original zsh config found."
+  if [ -n "$ZSHRC_SAVE" ]; then
+    echo "Your current config is saved at ${ZSHRC_SAVE}."
+  fi
 fi
 
 echo "Thanks for trying out Oh My Zsh. It's been uninstalled."
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -45,18 +45,25 @@ if [ -e "$ZSHRC_ORIG" ]; then
   else
     echo "Skipping restore."
     if [ -n "$ZSHRC_SAVE" ]; then
-      echo ""
-      echo "Your files are at:"
-      echo "  Current config : ${ZSHRC_SAVE}"
-      echo "  Pre-OMZ backup : ${ZSHRC_ORIG}"
-      echo "You can review and merge them manually."
+      mv "${ZSHRC_SAVE}" ~/.zshrc
+      echo "Your current zsh config has been kept at ~/.zshrc."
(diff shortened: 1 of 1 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry it sat so long. Two people losing their config is exactly the right thing to be fixing, and asking before restoring is the right shape for the fix.

One blocking problem. The `mv ~/.zshrc "${ZSHRC_SAVE}"` near the top runs unconditionally, and the new "Skipping restore" branch never moves it back, so declining leaves the user with no `~/.zshrc` at all. I ran the three ways to decline and they all end the same way:

| answer | result |
| --- | --- |
| `n` | no `~/.zshrc` |
| bare Enter (the default) | no `~/.zshrc` |
| no stdin, e.g. non-interactive | no `~/.zshrc` |

Since the prompt defaults to `N`, the safest-looking answer is the one that removes their config, and the next `zsh` start runs `zsh-newuser-install`. Could you move `${ZSHRC_SAVE}` back to `~/.zshrc` in that branch, so declining means "keep what I have now"?

Two smaller things:

1. The description says this fixes #13156, but the no-backup path still ends with no `~/.zshrc` — it only gains a message there. If the decline branch restores the file, the same one-line restore in the `else` branch fixes #13156 properly and this PR covers both issues. #13619 is currently trying to fix that half on its own.
2. Whatever gets restored still has `source $ZSH/oh-my-zsh.sh` in it, which will error on every new shell now the directory is gone. Not necessarily yours to solve here, but I'd like your read — commenting the line out, or just warning the user, both seem reasonable.

The date detection is nicely done, by the way. I checked `date -r` on macOS against a file with an old mtime and it reports the real modification date, with a graceful fallback.

hponukumati (commented):
Thanks for the thorough review and for catching the decline-leaves-no-zshrc bug — that's a real regression and I missed it completely.

Here's what I plan to fix:

1. **Blocking issue:** When the user declines, I'll `mv "$ZSHRC_SAVE" ~/.zshrc` to put their current config straight back. Declining will mean "keep what I have now" — which is what the default `N` should imply.

2. **`else` branch (no pre-omz backup):** Same fix — move `$ZSHRC_SAVE` back to `~/.zshrc` instead of leaving the user with nothing. This should properly close #13156 too.

3. **`source $ZSH/oh-my-zsh.sh` still in the file:** I'll go with a warning rather than auto-editing the file. Modifying someone's config without being explicitly asked feels too invasive — a clear message after uninstall pointing out the stale line leaves the decision with the user.

Will push the updated changes once confirmed.

PR #13853, opened 2026-07-05
fix(extract): properly handle xz archives

xs5871:
## Standards checklist:

- [x ] I have read the contribution guide and followed all the instructions.
- [ x] If I used AI tools (ChatGPT, Claude, Gemini, etc.) to assist with this contribution, I've disclosed it below.

## Changes:

replace `unxz` with `xzcat`

## Other comments:

`unxz` by itself extracts `path/file.xz` to `path/file`, removing the archive in the process. This clashes with the plugins behaviour to extract into the current directory and keeping the archive by default.

example:
- extraction in same directory doesn't work at all
- extraction from other directory doesn't extract to cwd and removes archive

edit: ping as per guideline @drjaska @mcornella

### DIFF
--- a/plugins/extract/extract.plugin.zsh
+++ b/plugins/extract/extract.plugin.zsh
@@ -102,7 +102,7 @@ EOF
       (*.tar.lrz) (( $+commands[lrzuntar] )) && lrzuntar "$full_path" ;;
       (*.gz) (( $+commands[pigz] )) && pigz -cdk "$full_path" > "${file:t:r}" || gunzip -ck "$full_path" > "${file:t:r}" ;;
       (*.bz2) (( $+commands[pbzip2] )) && pbzip2 -d "$full_path" || bunzip2 "$full_path" ;;
-      (*.xz) unxz "$full_path" ;;
+      (*.xz) xzcat "$full_path" > "${file:t:r}" ;;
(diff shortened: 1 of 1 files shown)


its-Sohan (commented):
Correct fix. `unxz` extracts in-place (destructive + wrong directory), while `xzcat > file` extracts to cwd and preserves the archive — matching every other format in the plugin.

One edge case worth noting: if a file named the same as `${file:t:r}` already exists in the cwd, the redirect will silently overwrite it. This is consistent with how the other formats behave in this plugin (e.g., `.gz` and `.bz2` have the same issue), so it's not a blocker. Just something to be aware of. LGTM.

xs5871:
> One edge case worth noting: if a file named the same as `${file:t:r}` already exists in the cwd, the redirect will silently overwrite it. This is consistent with how the other formats behave in this plugin (e.g., `.gz` and `.bz2` have the same issue), so it's not a blocker. Just something to be aware of. LGTM.

for clarification: As I understand, extraction is done in a temporary directory and collision with existing files or directories is already taken care of (see lines 69ff). That is consistent with the tests I've done. I might be misunderstanding your edge case description though.

carlosala (approved):
LGTM!

PR #13858, opened 2026-07-07
feat(plugin): Add project-aliases plugin

dvigo:
Adds the project-aliases plugin, which allows loading project-specific aliases dynamically and securely upon entering a project directory. It includes a security trust-verification model to prevent silent execution of untrusted aliases.

### DIFF
--- a/plugins/project-aliases/README.md
+++ b/plugins/project-aliases/README.md
@@ -1,0 +1,127 @@
+
+# Project Aliases (Zsh Plugin)
+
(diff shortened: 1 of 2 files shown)


robbyrussell:
@dvigo I appreciate you pitching this as a possible plugin to include. I'm going to decline this as a bundled plugin for now.

Why? It looks like the project is still relatively new; I suggest we document this as a possible custom plugin people can explore. You had good docs on your README for people to use set that up:

For anyone curious, check it out at:

- https://github.com/dvigo/project-aliases

PR #13860, opened 2026-07-09
feat(golang): update completions

lrh3321:
## Standards checklist:

## Changes:

- Updated shared Go build flag completions based on current `go help build` output.
- Added modern build flags including `-C`, `-msan`, `-asan`, `-cover`, `-covermode`, `-coverpkg`, `-asmflags`, `-buildmode`, `-buildvcs`, `-json`, `-mod`, `-modfile`, `-overlay`, `-pgo`, `-pkgdir`, `-trimpath`, and `-toolexec`.
- Added value completion for enum-style flags such as `-buildmode`, `-buildvcs`, `-compiler`, and `-mod`.
- Updated `go get` completion for current module-focused usage with `-t`, `-u`, `-u=patch`, `-tool`, and package arguments.
- Added `go version` completion for `-m`, `-v`, and file operands.
- Added `go work` completion with `edit`, `init`, `sync`, `use`, `vendor`, and `help` subcommands.
- Added command-specific completion for `go work edit`, `go work init`, `go work use`, and `go work vendor`.
- Removed duplicate `-v` definitions from `go build` and `go install`, since `-v` is now included in shared build flags.

## Other comments:

AI assistance disclosure: I used AI assistance to inspect the existing completion file, compare it against local `go help` output, and draft this PR description.

Testing:
- Checked the generated diff manually.
- Used local `go help build`, `go help get`, `go help version`, and `go help work` output as references.
- Verified the updated completions in another environment with `zsh` available.

No similar open PR was found for this change.

No new aliases are introduced by this change.

### DIFF
--- a/plugins/golang/_golang
+++ b/plugins/golang/_golang
@@ -81,6 +81,7 @@ _go() {
     'tool[run specified go tool]'
     'version[print Go version]'
     'vet[run go tool vet on packages]'
+    'work[workspace maintenance]'
   )
   if (( CURRENT == 2 )); then
     # explain go commands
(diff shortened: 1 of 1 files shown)


robbyrussell (commented):
Thanks for this, and sorry it sat two months without a word — that's on us.

I checked every flag you added against `go help build|get|version|work` on go1.25.5 and the shared build-flag list matches the real one exactly, in order. Dropping the old `-ccflags` and the 5c/6c/8c text is right too.

Three things before it lands:

1. `go work vendor <TAB>` completes nothing. That branch has only the `-v` and `-o` option specs and no operand spec, so there is nothing left to offer. `go mod vendor` just above it ends with `"*:flags"` — the same line here should do it.
2. `go version` is missing `-json`, which Go added in 1.24. `go help version` gives `[-m] [-v] [-json] [file ...]`.
3. A question rather than a finding: the `-u=[...]:mode:(patch)` spec sitting next to plain `-u` declares an option whose literal name is `-u=`. Did you mean that, or the `-u=patch` value form? I could not get a clean answer out of a test harness here, so I would rather ask than guess.

Also: please ignore the automated review from September 5. I checked, and `go test -v` and `go build -v` still complete correctly after you moved `-v` into the shared list.

lrh3321:
Thanks for the thorough review @robbyrussell . I’ve addressed the issues:

- Added `*:flags` to `go work vendor`, so argument completion now works.
- Added the missing `-json` flag to `go version`.
- Confirmed that `-u=patch` is intentional, matching Go’s documented syntax.

Would you like me to squash these changes together with the previous commit into a single commit?

carlosala (approved):

carlosala:
Thanks!

PR #13861, opened 2026-07-09
feat(systemadmin)!: replace clrz with pszombie

yilas:
## Standards checklist:

## Changes:

- Added `pszombie` alias to list zombie processes with ownership, PID, and parent PID (PPID) details.
- Removed (commented out) the `clrz` function, as attempting to directly `kill -9` (a xargs is missing, I guess) zombie processes is fundamentally ineffective in Linux (zombies are already dead and must be cleared via their parent process).

### Use case for `pszombie`

Zombie processes cannot be killed directly using standard signals like `kill -9`. To clear them from the process table, a system administrator needs to identify and kill or restart their parent process (PPID), or identify the owner (USER) responsible for them.

The new `pszombie` alias provides a quick, lightweight way for plugin users to monitor these defunct processes and immediately grab the exact fields (USER and PPID) required to take action and keep the process table clean.

### DIFF
--- a/plugins/systemadmin/systemadmin.plugin.zsh
+++ b/plugins/systemadmin/systemadmin.plugin.zsh
@@ -177,10 +177,10 @@ function getip() {
   fi
 }
 
-# Clear zombie processes
-function clrz() {
-  ps -eal | awk '{ if ($2 == "Z") {print $4}}' | kill -9
-}
+## Clear zombie processes
+#function clrz() {
+#  ps -eal | awk '{ if ($2 == "Z") {print $4}}' | kill -9
+#}
 
 # Second concurrent
 function conssec() {
--- a/plugins/systemadmin/systemadmin.plugin.zsh
+++ b/plugins/systemadmin/systemadmin.plugin.zsh
@@ -28,6 +28,8 @@ alias mkdir='mkdir -pv'
 # get top process eating memory
 alias psmem='ps -e -orss=,args= | sort -b -k1 -nr'
 alias psmem10='ps -e -orss=,args= | sort -b -k1 -nr | head -n 10'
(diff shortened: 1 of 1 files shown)


robbyrussell (dismissed):
Thanks for this, and sorry for the slow reply. You're right about `clrz`, and you're right about the shape of the fix.

For anyone reading later: `clrz` pipes PIDs into `kill -9`, but `kill` doesn't read stdin, so all it has ever done is print `kill: not enough arguments`. And it couldn't have worked anyway... you can't kill a zombie, it's already dead. You kill the parent, which is exactly why showing `ppid` is the right call here.

Couple of things to fix before it lands:

1. `ps -eo user,pid,ppid,state,cmd` doesn't work on macOS. BSD `ps` has no `cmd` keyword, so it just prints `ps: cmd: keyword not found`. `comm` works on both... I ran `ps -eo user,pid,ppid,state,comm | awk '$4=="Z"'` here and it found the zombies on my machine correctly.

2. `plugins/systemadmin/README.md:51` still has the `clrz` row. That wants replacing with a `pszombie` one.

3. The old function is commented out rather than removed. Please delete it... the git history keeps it if anyone ever wants it back.

One more thing: since this removes a documented function, the commit wants to be `feat(systemadmin)!:` with a `BREAKING CHANGE:` line in the body, so it shows up properly in the changelog. Given `clrz` never actually worked I don't think anyone will be hurt by it, but the changelog should still say so.

Worth knowing that #13948 proposed the same change a few weeks after yours, so you're not the only one who hit this.

Happy to take another look once those are in.

carlosala (approved):

carlosala:
Thanks for your contribution!

PR #13847, opened 2026-07-01
feat(git): add `gbcopy`

NunzioFornitto:
### Summary
Renames the helper function to `gbcopy` to clearly signal its functionality (copying the current Git branch name to clipboard via `clipcopy`).

### Changes
- Renamed function to `gbcopy`.
- Moved definition out of the WIP section to its proper alphabetical spot under `Functions Current`.
- Added `gbcopy` documentation row to `plugins/git/README.md`.

PR #13867, opened 2026-07-15
Add gcbc helper to copy current git branch name

### LINKED ISSUE #13843: add gcb helper to copy current git branch name

_No response_

No equivalent helper was found to quickly copy the current Git branch name directly to the clipboard.

Add a small helper to the Git plugin:
`gcb`
**Expected behavior:**
- Get current Git branch name
- Copy it to clipboard
**Example:**
```bash
$ git branch --show-current
```
```
feature/login-api
```
```bash
$ gcb
```
Clipboard:
```
feature/login-api
```
**Proposed implementation:**
```bash
gcb() {
  git branch --show-current | tr -d '\n' | xclip -selection clipboard
}
```

Initial implementation uses xclip as a starting point. Support for other clipboard tools (pbcopy, wl-copy, etc.) could be extended if needed.

_No response_

### Additional context

_No response_

### Related Issues

_No response_

yoshibase:
## Summary

Fixes #13843.

Add `gcbc` to copy the current Git branch name with Oh My Zsh's existing cross-platform `clipcopy` helper. The requested `gcb` name is already the alias for `git checkout -b`, so this uses the next unambiguous name.

Detached HEAD is handled explicitly: the function prints an error and returns 1 instead of copying an empty value.

## Test plan

- [x] `zsh -n plugins/git/git.plugin.zsh`
- [x] Diff is limited to the function and its Git plugin documentation

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -255,6 +255,7 @@ receive further support.
 | `git_develop_branch`     | Returns the name of the “development” branch: `dev`, `devel`, `development` if they exist, `develop` otherwise |
 | `git_main_branch`        | Returns the name of the main branch: `main` if it exists, `master` otherwise                                   |
 | `grename <old> <new>`    | Renames branch `<old>` to `<new>`, including on the origin remote                                              |
+| `gcbc`                   | Copies the current branch name to the system clipboard (uses `clipcopy`)                                       |
 | `gbda`                   | Deletes all merged branches                                                                                    |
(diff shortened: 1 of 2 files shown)


yoshibase:
Happy to adjust naming or behavior if maintainers prefer a different approach for the clipboard helper.

robbyrussell:
Thanks for this, and sorry for the slow reply.

#13847 is doing the same thing and got there a couple of weeks earlier... it's already been through a review round, so I'm going to carry that one forward. Worth following if you want to see where it lands.

Closing as a duplicate. Thanks for taking the time on it, and please do send more.

PR #13816, opened 2026-06-14
fix(git): support nounset option

1795771535y-cell:
## Standards checklist:

- [x] The PR title is descriptive.
- [x] The PR doesn't replicate another PR which is already open.
- [x] I have read the contribution guide and followed all the instructions.
- [x] The code follows the code style guide.

**AI-assisted**: OpenClaw assisted with root cause analysis and fix implementation.

## Problem

When `set -u` (nounset option) is enabled, opening a terminal with
ohmyzsh produces shell errors because `${RPROMPT}` and the right-side
prompt variables `${RPS1}`-`${RPS4}` may not be set.

```
_defer_async_git_register:2: RPROMPT: parameter not set
```

See #13701.

## Fix

Add `-` to the parameter expansion to provide an empty default
when these variables are unset:
- `${RPROMPT}` → `${RPROMPT-}`
- `${RPS1}` → `${RPS1-}` (same for RPS2-RPS4)

This matches the existing approach for `${PS1}`-`${PS4}` which
are always set by zsh and don't need this treatment.

Fixes #13701

PR #13869, opened 2026-07-16
fix(core): support nounset startup

### LINKED ISSUE #13854: set -eu, oh-my-zsh.sh, and unset variables
### Describe the bug

`set -eu` is considered good practice. However, `source $ZSH/oh-my-zsh.sh` cannot run with those parameters. A few issues can be fixed easily. E.g.:

    [[ -n "$ZSH_CUSTOM" ]] || ZSH_CUSTOM="$ZSH/custom"

Should be

    [[ -v ZSH_CUSTOM ]] || ZSH_CUSTOM="$ZSH/custom"

But there are tons of other variables that are not used safely.

### Steps to reproduce

Add `set -eu` to `.zshrc`

### Expected behavior

No error during startup

_No response_

### OS / Linux distribution

Arch Linux

### Zsh version

5.9.1

### Oh My Zsh version

19962ac

### Terminal emulator

Konsole

_No response_

### Additional context

_No response_

tianrking:
## Summary

- make core startup defaults safe when `set -u` is enabled
- preserve the existing unset-variable defaults for update settings, completion, history, terminal support, and theme loading
- add a CI smoke test that sources Oh My Zsh with unset optional configuration

The startup path previously read optional configuration variables directly. Under `nounset`, that aborted individual libraries and left the shell only partially initialized.

Fixes #13854

## Validation

- `zsh -df lib/tests/nounset.test.zsh`
- `zsh -df lib/tests/cli.test.zsh`
- `zsh -n` for every changed shell script
- `git diff --check`

AI-assisted: Codex helped investigate and implement the fix; the startup scenario was validated locally in Zsh 5.9.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,3 +41,5 @@ jobs:
                       ./themes/*.zsh-theme; do
             zsh -n "$file" || return 1
           done
+      - name: Check nounset startup
+        run: zsh -df lib/tests/nounset.test.zsh
--- a/lib/completion.zsh
+++ b/lib/completion.zsh
@@ -14,10 +14,10 @@ bindkey -M menuselect '^o' accept-and-infer-next-history
 zstyle ':completion:*:*:*:*:*' menu select
 
 # case insensitive (all), partial-word and substring completion
-if [[ "$CASE_SENSITIVE" = true ]]; then
+if [[ ${CASE_SENSITIVE-} = true ]]; then
   zstyle ':completion:*' matcher-list 'r:|=*' 'l:|=* r:|=*'
 else
-  if [[ "$HYPHEN_INSENSITIVE" = true ]]; then
+  if [[ ${HYPHEN_INSENSITIVE-} = true ]]; then
     zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]-_}={[:upper:][:lower:]_-}' 'r:|=*' 'l:|=* r:|=*'
   else
     zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]}={[:upper:][:lower:]}' 'r:|=*' 'l:|=* r:|=*'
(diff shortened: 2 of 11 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry it sat so long. The idea is one we want... we've merged targeted nounset fixes before (#7433, #10443, #11856, #13816), and listing what you ran in the description is genuinely helpful.

Couple of things to fix before it can land. The first one is the important one, because it makes things worse for people who aren't using `nounset` at all:

1. `oh-my-zsh.sh:100,103` — `SHORT_HOST="${HOST/.*/}"` became `"${HOST-}"`, which drops the domain-stripping. `$HOST` is a zsh builtin parameter and is always set, so the guard wasn't needed there. On Linux (where the `else` branch always runs) a machine called `web01.prod.example.com` gets `SHORT_HOST=web01.prod.example.com` instead of `web01`. That renames `.zcompdump` and forces a completion rebuild, and it also moves `~/.ssh/environment-$SHORT_HOST` (ssh-agent plugin) and `~/.keychain/$SHORT_HOST-sh` (keychain plugin), so both lose their cached agent. Reverting those two lines to `${HOST/.*/}` should be all it needs.

2. `oh-my-zsh.sh:90` — the fpath loop still does `for plugin ($plugins)` unguarded, so startup under `set -u` still aborts, just later than before. I ran master and your branch side by side with `plugins` unset: master stops at line 55, your branch stops at line 90. The new test doesn't catch it because it sets `plugins=()` at the top. If you drop that line from the test it should fail, and then guarding line 90 the same way you did line 203 fixes both.

Smaller, not blocking:

3. On a default setup (`plugins=(git)`, robbyrussell theme) the prompt still prints `_OMZ_ASYNC_OUTPUT[_omz_git_prompt_info]: parameter not set` on every draw, from `lib/git.zsh:170`. Worth pulling in while you're here.

4. `lib/history.zsh:38` — `[ -z ${HISTFILE-} ]` lost its quotes. It works, but the rest of the diff keeps them.

One other thing you should know: #13702 has been open since April doing much the same job, and it covers `lib/git.zsh` and the battery plugin, which yours doesn't. Yours covers the `oh-my-zsh.sh` guards that one misses. Neither is a superset of the other, so it may be worth the two of you comparing notes... cc @demccormack. I'd rather land one complete fix than half of each.

Happy to take another look once 1 and 2 are in.

PR #13872, opened 2026-07-18
fix(bundler): restore `bu` and add `bua` alias

### LINKED ISSUE #13871: Recent Bundler `bu` alias change breaks commands based on `bundle update`
### Describe the bug

Recently, Bundler plugin's `bu` alias was [changed](https://github.com/ohmyzsh/ohmyzsh/commit/ff2f16e8df7386d7198009566aef09cbbc0c8212) from pointing to `bundle update` to a more specific `bundle update --all` command that breaks compatibility with [other helpful commands](https://bundler.io/man/bundle-update.1.html) that use `bundle update` as their base. To name a few:

* `bu --ruby` – update the locked version of Ruby in `Gemfile.lock` to the current version of Ruby
* `bu --bundler` - update the locked Bundler version in `Gemfile.lock`
* `bu money` - update any specific gem's version

That said, the `bu` alias has essentially become a more specific command from a generic one it used to be. And generic aliases like `bu` should be kept intact unless there's a good reason not to. Specific commands like `bundle update --all`, on the other hand, should have specific aliases (if justified).

### Steps to reproduce

1. Enable the `bundler` plugin.
2. Create a new directory and `cd` into it: `mkdir omz-bu-issue-repro && cd $_`.
3. Create a `Gemfile`: `bundle init`.
4. Add the `money` gem to `Gemfile`: `ba money` (aliased to `bundle add`).
5. Try running `bu money`.
6. Get an error message: `Cannot specify --all along with specific options.`.

### Expected behavior

Provided a Ruby version or a gem is specified in your `Gemfile`, the commands above should succeed using the `bu` alias as before:

```sh
omz-bu-issue-repro λ bu money
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
Resolving dependencies...
Bundler attempted to update money but its version stayed the same
Bundle updated!
Gems in the group 'production' were not updated.
```

_No response_

### OS / Linux distribution

macOS 26.5.2

### Zsh version

5.9.2

### Terminal emulator

iTerm2

None

### Additional context

While I understand the intent of @eclectic-coding to add a non-generic command to the plugin, it should not have come at the expense of a generic command like `bundle update`.

If we'd like to keep the `bundle update --all` command, we could add a more specific alias like `bua`:

```sh
alias bua="bundle update --all"
```

with the description from [Bundler's `bundle update` man page](https://bundler.io/man/bundle-update.1.html): `Update all gems specified in Gemfile`.

More importantly, I believe we should revert the changes made to the `bu` alias.

agarwalrahul2702:
## Standards checklist:

## Changes:

- Fixes #13871.
- Restore `bu` to the generic `bundle update` alias so arguments pass through naturally:
  - `bu rake` -> `bundle update rake`
  - `bu --ruby` -> `bundle update --ruby`
  - `bu` -> `bundle update`
- Add `bua` as the explicit `bundle update --all` alias for users who want the all-gems update command added recently.
- Update the Bundler plugin README to document both aliases.

Bundler rejects `--all` when combined with specific gems or options. Since aliases append their arguments, `alias bu="bundle update --all"` turns `bu rake` into `bundle update --all rake`, which fails. Keeping `bu` as the generic alias fixes parameterized updates and follows the contribution guidance to prefer one generic alias.

The dedicated `bua` alias keeps the recent `bundle update --all` behavior available without making the generic `bu` alias incompatible with gem-specific updates.

## Verification:

- `zsh -n plugins/bundler/bundler.plugin.zsh`
- `zsh -fc 'source plugins/bundler/bundler.plugin.zsh; alias bu; alias bua'` reports `bu='bundle update'` and `bua='bundle update --all'`
- Fake `bundle` smoke test:
  - `bu rake` expands to `bundle update rake`
  - `bu --ruby` expands to `bundle update --ruby`
  - `bu` expands to `bundle update`
  - `bua` expands to `bundle update --all`
- `git diff --check`

## Other comments:

This update intentionally removes the function and `unalias` approach from the first PR version. That avoids assumptions about other plugins' aliases and keeps `bu` as the generic Bundler update command. The added `bua` alias covers the all-gems update use case separately.

AI disclosure: Claude Fable 5 (Claude Code) assisted with reproducing, diagnosing, implementing, and testing the original change. OpenAI Codex assisted with addressing the review feedback. I reviewed and understand the final code and verification results.

### DIFF
--- a/plugins/bundler/README.md
+++ b/plugins/bundler/README.md
@@ -22,7 +22,7 @@ plugins=(... bundler)
 | `bo`   | `bundle open`     | Opens the source directory for a gem in your bundle      |
 | `bout` | `bundle outdated` | List installed gems with newer versions available        |
 | `bp`   | `bundle package`  | Package your needed .gem files into your application     |
-| `bu`   | `bundle update --all` (no arguments) or `bundle update <args>` | Update all gems, or only the given gems/options (e.g. `bu rake`, `bu --ruby`) |
+| `bu`   | `bundle update`   | Update your gems to the latest available versions          |
 
 ## Gem wrapper
 
--- a/plugins/bundler/README.md
+++ b/plugins/bundler/README.md
@@ -23,6 +23,7 @@ plugins=(... bundler)
 | `bout` | `bundle outdated` | List installed gems with newer versions available        |
 | `bp`   | `bundle package`  | Package your needed .gem files into your application     |
 | `bu`   | `bundle update`   | Update your gems to the latest available versions          |
+| `bua`  | `bundle update --all` | Update all gems to the latest available versions      |
 
 ## Gem wrapper
 
--- a/plugins/bundler/README.md
+++ b/plugins/bundler/README.md
@@ -22,7 +22,7 @@ plugins=(... bundler)
 | `bo`   | `bundle open`     | Opens the source directory for a gem in your bundle      |
 | `bout` | `bundle outdated` | List installed gems with newer versions available        |
 | `bp`   | `bundle package`  | Package your needed .gem files into your application     |
-| `bu`   | `bundle update --all` | Update your gems to the latest available versions    |
+| `bu`   | `bundle update --all` (no arguments) or `bundle update <args>` | Update all gems, or only the given gems/options (e.g. `bu rake`, `bu --ruby`) |
 
 ## Gem wrapper
 
--- a/plugins/bundler/bundler.plugin.zsh
+++ b/plugins/bundler/bundler.plugin.zsh
@@ -9,7 +9,16 @@ alias bl="bundle list"
 alias bo="bundle open"
 alias bout="bundle outdated"
 alias bp="bundle package"
-alias bu="bundle update --all"
+unalias bu 2>/dev/null
+# `bundle update` needs --all to update everything, but rejects --all when
+# gems or options are given (e.g. `bu <gem>`, `bu --ruby`)
+bu() {
+  if [[ $# -eq 0 ]]; then
+    bundle update --all
+  else
+    bundle update "$@"
+  fi
+}
 
 ## Gem wrapper
 
--- a/plugins/bundler/bundler.plugin.zsh
+++ b/plugins/bundler/bundler.plugin.zsh
@@ -10,6 +10,7 @@ alias bo="bundle open"
 alias bout="bundle outdated"
 alias bp="bundle package"
 alias bu="bundle update"
+alias bua="bundle update --all"
 
 ## Gem wrapper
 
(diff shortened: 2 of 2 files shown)


dskecse:
Hey Rahul, thank you for your change!

Let me share a few thoughts on why I believe we may want to handle the issue differently.

The change now assumes the default behavior for a `bu` command typed in a terminal without options is a more specific `bundle update --all` rather than a generic `bundle update` one.

Also, note that the [`New plugin aliases` section](https://github.com/ohmyzsh/ohmyzsh/blob/master/CONTRIBUTING.md#new-plugin-aliases) of the project's `CONTRIBUTING.md` guide mentions that:

> ... we require that new aliases follow these conditions:
> 3. Prefer one generic alias over many specific ones.
> 5. If a command with the same name exists, look for a different alias name.

Another thing that stands out, although it's not as important, is the use of `unalias`. It seems like we start making assumptions of other plugins' aliases, and I'm wondering if that should be the case?

agarwalrahul2702:
Thanks for the feedback. I updated the PR in `35182d81` to use the simpler generic-alias approach:

- `bu` is now `alias bu="bundle update"`.
- The function and `unalias` are removed, so the plugin no longer makes assumptions about aliases from other plugins.
- The README now documents `bu` as the generic `bundle update` alias.

This means `bu rake` and `bu --ruby` pass arguments through naturally, while bare `bu` remains the generic `bundle update` command, which aligns with the contribution guidance to prefer one generic alias.

dskecse (approved):
Thank you for the quick update!

Would you please update the PR title as well?

I also noticed the `bundle update --all` is now completely gone. Since it was added only recently, I'm leaving this up to you to decide whether to add this command as a new alias, thanks!

agarwalrahul2702:
Thanks again. I updated the PR title to `fix(bundler): restore generic \`bu\` alias`.

On the optional `bundle update --all` alias: I am going to leave it out of this PR for now. The reported bug is fixed by making `bu` generic again, and adding a new alias would expand the change beyond the issue while also needing a separate alias-name discussion under the contribution guidelines. If maintainers prefer a dedicated alias for `bundle update --all`, I can add it in a follow-up or adjust this PR with the preferred alias name.

agarwalrahul2702:
I added the optional all-gems update alias in `1ff50dd7`:

- `bu` remains the generic `bundle update` alias, so `bu rake` and `bu --ruby` work.
- `bua` now maps to `bundle update --all` for users who want the all-gems update behavior explicitly.
- The README and PR description have been updated to document both aliases.

I also verified syntax and alias expansion for `bu`, `bu rake`, `bu --ruby`, and `bua`.

dskecse:
Great, thank you Rahul!

robbyrussell (approved):
This makes sense to me, I'll push this through today

PR #13873, opened 2026-07-19
perf: avoid duplicate completion security audit

### LINKED ISSUE #13870: Startup: compaudit runs its security scan twice on every shell init
### Describe the bug

I got curious what Claude (Fable 5) would notice with a fresh set of eyes on our init path, so I had it profile a new shell and see where the time goes. It flagged a few things, but the big one made me laugh a little: we've been running the `compaudit` security scan twice on every single startup.

In `oh-my-zsh.sh`:

```zsh
compinit -i -d "$ZSH_COMPDUMP"
# If completion insecurities exist, warn the user
handle_completion_insecurities &|
```

Looks like one audit, right? Turns out `compinit` always runs a full `compaudit` internally unless you pass `-u` or `-C`. The `-i` just tells it to quietly skip insecure dirs instead of prompting. Then `handle_completion_insecurities` immediately runs the same scan *again* so it can print a nicer warning. Same `fpath`, same dirs, same answer, twice, milliseconds apart. The xtrace timestamps:

```text
.833472  oh-my-zsh.sh:127> compinit -i ...
.835729  compaudit:60> _i_files=( ... )               <- scan #1, inside compinit
.858194  oh-my-zsh.sh:129> handle_completion_insecurities
.859539  handle_completion_insecurities:13> compaudit  <- scan #2
```

On my machine (Apple Silicon, zsh 5.9, minimal `.zshrc` with just the git plugin) the two scans are ~50ms of a ~100ms warm startup. In fairness, the second one is backgrounded so nobody's staring at it, but scan #1 puts ~18ms on the critical path for checking directories that basically never change between shells.

Two ideas, neither of which touches how anyone configures anything:

1. Run `compaudit` once ourselves and share the result: `compinit -C`/`-u` when it comes back clean, and hand the captured output to the warning function instead of re-scanning.
2. Maybe go further and cache a clean verdict keyed on `$fpath` + dir mtimes, like `lib/grep.zsh` already does for its checks.

Before anyone (me included) runs off and writes a patch: has this been tried before? This code has been touched carefully over the years and I'd love to hear if there's history here that the git log isn't telling me.

There were a few other smaller findings (startup forks we could cache, zcompile opportunities). I'll open those separately so this one stays about `compaudit`.

### Steps to reproduce

This was via claude/fable

### Expected behavior

Can we make this faster

_No response_

### OS / Linux distribution

MacOS

### Zsh version

5.9

### Oh My Zsh version

master

### Terminal emulator

Ghostty

None

### Additional context

_No response_

farlyfeifei:
## Summary

- skip the second `compaudit` pass on clean startup
- use the `_comp_secure` flag set by stock `compinit -i` when insecure completion entries are found
- keep the existing warning and filtering behavior for insecure setups

## Why

Oh My Zsh currently runs `compinit -i`, which performs a security audit, and then always backgrounds `handle_completion_insecurities`, which performs the same audit again to build its warning. On a clean machine the second pass is unnecessary overhead.

This keeps `compinit -i` in place, so completion directories are still audited on every normal startup. It does not restore the old `compinit -C` shortcut removed in #10672, which could trust stale dump state.

Behavior after this change:

- clean startup: one `compaudit`, with no warning handler process
- insecure startup: insecure entries are still excluded and the existing full warning is shown; the second audit remains to obtain the display list
- `ZSH_DISABLE_COMPFIX=true`: unchanged

## Checks

- `zsh -n oh-my-zsh.sh`
- controlled clean full-source fixture: completion initialized and `compaudit` called once
- controlled insecure full-source fixture: insecure path removed from `fpath`, probe completion not loaded, full warning emitted, and display audit retained
- verified `_comp_secure` behavior in stock zsh sources from 4.3.9 through 5.9

Fixes #13870

AI-assisted; the implementation was reviewed and tested manually, including clean and insecure `fpath` cases.

### DIFF
--- a/oh-my-zsh.sh
+++ b/oh-my-zsh.sh
@@ -124,9 +124,11 @@ fi
 if [[ "$ZSH_DISABLE_COMPFIX" != true ]]; then
   source "$ZSH/lib/compfix.zsh"
   # Load only from secure directories
+  # Reset the flag compinit sets when -i excludes insecure entries
+  unset _comp_secure
   compinit -i -d "$ZSH_COMPDUMP"
   # If completion insecurities exist, warn the user
-  handle_completion_insecurities &|
+  [[ "$_comp_secure" == yes ]] && handle_completion_insecurities &|
 else
   # If the user wants it, load from all found directories
   compinit -u -d "$ZSH_COMPDUMP"


carlosala (approved):
LGTM!

carlosala:
Fantastic solution! I didn't know about this `_comp_secure` and it is not documented.

farlyfeifei:
Thank you!



---- Replied Message ----
| From | Carlo ***@***.***> |
| Date | 08/05/2026 19:21 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [ohmyzsh/ohmyzsh] perf: avoid duplicate completion security audit (PR #13873) |

@carlosala approved this pull request.

LGTM!

carlosala:
Thanks for your contribution!

PR #13879, opened 2026-07-23
feat(web-search): add 8 new search engines

letavocado:
Add search aliases for:

- `maven` — Maven Central (Java/Kotlin packages)
- `mdn` — MDN Web Docs
- `pd` — pub.dev (Dart/Flutter packages)
- `pypi` — Python Package Index
- `aur` — Arch User Repository
- `archwiki` — Arch Wiki
- `hf` — Hugging Face (ML models & datasets)
- `wikipedia` — Wikipedia

All align with the project's alias guidelines: commonly used, general-purpose, single generic alias per service.

### DIFF
--- a/plugins/web-search/README.md
+++ b/plugins/web-search/README.md
@@ -51,6 +51,14 @@ Available search contexts are:
 | `npmpkg`              | `https://www.npmjs.com/search?q=`               |
 | `packagist`           | `https://packagist.org/?query=`                 |
 | `gopkg`               | `https://pkg.go.dev/search?m=package&q=`        |
+| `maven`               | `https://search.maven.org/search?q=`            |
+| `mdn`                 | `https://developer.mozilla.org/en-US/search?q=` |
+| `pd`                  | `https://pub.dev/packages?q=`                   |
+| `pypi`                | `https://pypi.org/search/?q=`                   |
+| `aur`                 | `https://aur.archlinux.org/packages?O=0&K=`     |
+| `archwiki`            | `https://wiki.archlinux.org/index.php?search=`  |
+| `hf`                  | `https://huggingface.co/search?q=`              |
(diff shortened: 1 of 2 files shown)


robbyrussell (changes requested):
Thanks for this, and sorry for the slow reply. Most of these are good additions... `pypi`, `archwiki` and `aur` in particular are ones I'd expect people to reach for.

A few need changing before it can go in. The first two are the important ones, because they'd break plugins people already have enabled:

1. `mdn` is already taken. `plugins/frontend-search` has `alias mdn='frontend mdn'` at line 19. Whichever plugin loads second wins, so anyone using both gets one of them silently broken.

2. `pd` is already taken too, by `plugins/perl` (`alias pd='perldoc'`, line 18). Same problem.

3. `hf` shadows the actual Hugging Face CLI, which is called `hf` as of huggingface_hub 0.34. An alias that hides a real command on someone's PATH is one we try hard to avoid. The URL also needs fixing regardless... `https://huggingface.co/search?q=` redirects to `/search/full-text` and drops the query on the way. I checked, and `https://huggingface.co/search/full-text?q=` works.

4. `maven` points at `search.maven.org`, which has been superseded by `https://central.sonatype.com/search?q=`.

5. Minor, your call: the plugin already has `alias wiki='web_search duckduckgo \!w'` on line 104, so `wikipedia` is a bit of an overlap. Not blocking if you think the direct one is worth having.

So: drop or rename 1, 2 and 3, fix the two URLs, and the rest looks good to me. If you'd rather just drop the three awkward names and keep the other five, that's fine too and I'd merge it.

letavocado:
Implemented the review changes in commit `9f21ea7c`: removed the conflicting `mdn`, `pd`, and `hf` aliases, updated Maven to `https://central.sonatype.com/search?q=`, and kept the remaining five search contexts.

robbyrussell (approved):
lgtm

PR #13900, opened 2026-07-27
fix(pm2): escape `+` in completion regex for BSD awk

EnzoDeg40:
## Standards checklist:

## Changes:

The `_pm2_id_names` completion passed `+--- ` as the regex argument to awk's sub(). An unescaped leading `+` is an illegal primary in BSD awk (macOS /usr/bin/awk), breaking `pm2 <cmd> <TAB>` completion. GNU awk tolerates it, so the bug only showed on macOS/BSD. Escape it to `\+`.

## Other comments:

- I used an LLM to identify the problem
- Contributor for pm2 plugin (due to the contribution guidelines):
  - @cuspymd Creator of plugin
  - @carlosala Last update completion

### DIFF
--- a/plugins/pm2/_pm2
+++ b/plugins/pm2/_pm2
@@ -314,7 +314,7 @@ _pm2_subcommands() {
 (( $+functions[_pm2_id_names] )) ||
 _pm2_id_names() {
   local app_list=$(pm2 list -m)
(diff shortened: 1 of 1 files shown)


carlosala (approved):

carlosala:
Good catch, thanks!

PR #13907, opened 2026-07-30
feat(git): add grbias alias for rebase --interactive --autosquash

### LINKED ISSUE #13865: Add Git autosquash alias

Git plugin

Frequently, when working with Git, I'm using fixup commits. When I go to squash all the fixups in, I have to type out `--autosquash`, which is a relatively long option to have to type. Git does not have a shorthand.

Add a Git alias that maps to `git rebase --interactive --autosquash`. This could be `grbia`, `grbas`, or `grbsq` (sq = squash).

From the contribution guidelines on suggesting new aliases:

> They will be used by many people, not just a few.
>
> The aliases will be used many times and for common tasks.

Committing and squashing fixups is a common and widely used technique. There's enough demand and frequency-of-use to justify `gcfu`. I would imagine that many individuals using that alias also use `--autosquash`.

Also, judging by the number of closed PRs for this, it seems that there's some demand for it.

> Prefer one generic alias over many specific ones.

This could be slightly more generic by removing the `--interactive` option. However, I find it useful to review the rebase before it happens. If a user doesn't want to review changes, exiting their editor of choice should be quick enough (pressing `ZZ` in vim is pretty much a reflex at this point).

> When justifying the need for an alias, talk about workflows where you'll use it, preferably in combination with other aliases.

When working on code changes, I'll pile up a bunch of fixups for commits I don't want to rebase right away. This may be necessary or useful for many reasons:

- The changes are still in flux. I don't want to rebase if I would have to do it again shortly.
- I can't be bothered to rebase at the moment.
- Me or other developers have other changes based on the original commit.
- The change is up for review and rebasing the branch from under the reviewer would be rude. 

The `--autosquash` alias makes it easy to rebase these fixups into the original commits.

- Creating my own alias: This is an option, but I think this would be a helpful alias for many others.
- Setting the Git config variable `rebase.autoSquash` to true: There are just as many times I'm trying to rebase without squashing, and writing `--no-autosquash` would also become irritating. This has been mentioned as an alternative solution in other PRs adding this alias, which is not a workable alternative for many users.

### Additional context

I couldn't find any specific maintainers listed for the Git plugin so I didn't mention anyone. Sorry!

### Related Issues

Related PRs:

- https://github.com/ohmyzsh/ohmyzsh/pull/12418: Judging by the comments, this PR was submitted at a time when new aliases were rejected wholesale. According to the current version of the contribution guidelines, this does not appear to be the case anymore.
- https://github.com/ohmyzsh/ohmyzsh/pull/9451: Bundled with other aliases. Also, the `rebase.autoSquash` was proposed (see above).
- https://github.com/ohmyzsh/ohmyzsh/pull/8934: Is less generic that what is suggested here. This uses specific branch names.
- https://github.com/ohmyzsh/ohmyzsh/pull/6030: Is less generic, still open.

MeGaurav4:
Adds `grbias` alias for `git rebase --interactive --autosquash`, complementing the existing `gcfu` (commit --fixup) and `grbi` (rebase --interactive) aliases.

Closes #13865.


robbyrussell:
Thanks for this, and for opening it against #13865 rather than just dropping the alias in.

We've turned this one down twice before, in #12418 and #9451. The reason both times was that `git config --global rebase.autoSquash true` makes `grbi` do exactly this, without spending another name in the git plugin.

Closing for the same reason... thanks for taking the time on it.

PR #13908, opened 2026-07-30
fix(git): use canonical `--remotes`

AnturGyffrous:
Git's long-option abbreviation allows `--remote` to work, but the [git-branch](https://git-scm.com/docs/git-branch) documentation specifies the plural form `--remotes`.

## Standards checklist:

## Changes:

- Updated the `gbr` alias in the Git plugin to use the canonical documented long option `--remotes` instead of the abbreviation `--remote`.
- Updated the Git plugin README.md to reflect this change.

## Other comments:

The original alias was added in commit 9a0196dbf72521319a0fa0db4ae84603c2c79868 on Aug 10, 2014 and it hasn't changed since. It works because Git's command-line parser accepts any long option as long as it uniquely identifies a single option.

For example, provided none of the prefixes are ambiguous, all of these are typically equivalent:

```
git branch --remotes
git branch --remote
git branch --remo
git branch --rem
```

If Git were ever to gain another `git branch` option beginning with `--remote...` (for example `--remote-only`), then `--remote` might become ambiguous and stop working, whereas `--remotes` would continue to work.

I ran this change passed ChatGPT to sense-check my understanding, and it said:

> This is a nice catch, and I think you’ve stumbled across one of Git’s more subtle command-line behaviours.

It went on to say:

> ### Was `--remote` ever official?
>
> As far as I can tell, no.
>
> The documentation consistently shows:
>
> ```
> -r
> --remotes
> ```
>
> and doesn't mention `--remote` as an alias.

https://chatgpt.com/share/6a6b6750-3b24-83ed-9c04-fbd95d0c7eb9

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -45,7 +45,7 @@ plugins=(... git)
 | `gbgD`                 | `LANG=C git branch --no-color -vv \| grep ": gone\]" \| cut -c 3- \| awk '"'"'{print $1}'"'"' \| xargs git branch -D`           |
 | `gbm`                  | `git branch --move`                                                                                                             |
 | `gbnm`                 | `git branch --no-merged`                                                                                                        |
-| `gbr`                  | `git branch --remote`                                                                                                           |
+| `gbr`                  | `git branch --remotes`                                                                                                          |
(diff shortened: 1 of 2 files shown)


carlosala (approved):

carlosala (approved):
LGTM!

carlosala:
Thanks!

PR #13911, opened 2026-07-31
feat(semver-tag): add plugin to bump semver git tags

katelovescode:
## Summary

Adds a `semver-tag` plugin that bumps a repo's latest `MAJOR.MINOR.PATCH` git tag and creates the next one.

- `gtM` / `gtm` / `gtp` bump major/minor/patch as a lightweight tag, with a confirmation prompt before creating anything
- `gtvM` / `gtvm` / `gtvp` are identical, except when bootstrapping a repo with no tags yet, they prefix the first tag with `v`
- "Latest" is determined by tag creation time (`git for-each-ref --sort=-creatordate`), not the highest version number and not `git describe`'s nearest-reachable-tag semantics, so it stays correct even if tags are created out of numeric order
- If the requested `v`/no-`v` format doesn't match what the repo's existing tags actually use, the confirmation prompt surfaces a mismatch warning instead of silently guessing - it always bumps using the real existing format
- Anything passed beyond the bootstrap flag is forwarded straight to `git tag` after the computed name, so `git tag`'s own flags (`--annotate`, `-m`, `-s` for signed tags, etc.) work without the plugin needing to reimplement them

Full behavior and examples are in the plugin's README.

## Test plan

- [x] Included test suite (`plugins/semver-tag/test.zsh`, run via `zsh test.zsh`) covers bootstrapping, decline/confirm, tag-format mismatch detection in both directions, invalid latest-tag format handling, creation-date-vs-version-number tag selection, and `git tag` flag pass-through - 31 assertions, all passing
- [x] Verified alias behavior in an actual interactive zsh session (not just the non-interactive test harness), confirming identical output/behavior

### DIFF
--- a/plugins/semver-tag/README.md
+++ b/plugins/semver-tag/README.md
@@ -1,0 +1,67 @@
+# semver-tag plugin
+
+Bumps a repo's latest `MAJOR.MINOR.PATCH` git tag and creates the next one, with a confirmation prompt before anything is created.
+
+To use it, add `semver-tag` to the plugins array in your `.zshrc` file:
+
+```zsh
+plugins=(... semver-tag)
+```
+
+## Commands
+
+| Command | Description |
+|---|---|
+| `gtM` | Bump the major version and create the tag (e.g. `1.4.2` -> `2.0.0`) |
+| `gtm` | Bump the minor version (e.g. `1.4.2` -> `1.5.0`) |
+| `gtp` | Bump the patch version (e.g. `1.4.2` -> `1.4.3`) |
+| `gtvM` | Same as `gtM`, but on first use in a repo with no tags, bootstraps with a `v` prefix (`v1.0.0`) |
+| `gtvm` | Same as `gtm`, `v`-prefixed bootstrap |
+| `gtvp` | Same as `gtp`, `v`-prefixed bootstrap |
+
+`M` = major, `m` = minor, `p` = patch (case-sensitive).
+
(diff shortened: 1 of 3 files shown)


robbyrussell:
Thanks for this, and it's clearly careful work... the README is thorough, there's no load-time cost, and the function names don't collide with anything in the tree. I checked.

I'm going to pass on it though, and it's not about quality.

Most plugins here wrap a tool someone already installed... aliases and completions, and when the tool changes we follow it. This one implements the semver logic itself, which means the project owns that logic and its bugs for as long as the plugin exists. We're a small group of volunteers, so the question we have to ask is who's maintaining this in three years, and for something that `npm version`, `standard-version`, `svu` and `cargo release` already do, I don't think we can answer it honestly.

One thing worth knowing either way, since it'd bite you in your own use: it picks the latest tag by creation date, but our git plugin's `gtl` sorts with `--sort=-v:refname`. So if you ever tag a backport, `gtM` will suggest a version that already exists.

This works really well as a custom plugin in `$ZSH_CUSTOM`, and it'll survive updates there. If you'd rather have it as a standalone repo, people can point their plugin manager straight at it.

Thanks again for taking the time... sorry it's a no.

katelovescode:
No worries at all, your explanation makes sense.  Thanks for the thoughtful reply!

robbyrussell:
Hope you contribute again!

PR #13931, opened 2026-08-11
fix: use fallback for login shell exec in omz reload

### LINKED ISSUE #13919: omz reload may fail or restart a different Zsh executable
`omz reload` uses `ZSH_ARGZERO` to restart the current shell:

```sh
local zsh="${ZSH_ARGZERO:-${functrace[-1]%:*}}"
[[ "$zsh" = -* || -o login ]] && exec -l "${zsh#-}" || exec "$zsh"
```

However, `ZSH_ARGZERO` describes how Zsh was invoked, not necessarily the path of the running executable.

For example:

* If Zsh was invoked through a relative path such as `./bin/zsh`, changing the working directory before `omz reload` can make `omz reload` **fail entirely** (or refer to a different executable).
* A login shell commonly has `ZSH_ARGZERO=-zsh`, so the reload resolves zsh again using the current `$PATH`.
* Likewise, `ZSH_ARGZERO=zsh` is resolved using the current `$PATH`, which may have changed since startup.

As a result, `omz reload` may fail or silently switch to a different Zsh executable/version.

In general, the current executable path cannot be reliably reconstructed from `ZSH_ARGZERO`, but `$SHELL` seems like a reasonable fallback for login shells, since it normally retains the configured shell path.

deepshekhardas:
﻿Uses a `\` fallback for login shell exec in `omz reload` so it works when the shell isn't the user's login shell.

Fixes #13919.

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -739,7 +739,11 @@ function _omz::reload {
   # Old zsh versions don't have ZSH_ARGZERO
   local zsh="${ZSH_ARGZERO:-${functrace[-1]%:*}}"
   # Check whether to run a login shell
-  [[ "$zsh" = -* || -o login ]] && exec -l "${zsh#-}" || exec "$zsh"
+  if [[ "$zsh" = -* || -o login ]]; then
+    exec -l "${${SHELL:-$zsh}#-}"
+  else
+    exec "$zsh"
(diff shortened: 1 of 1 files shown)


carlosala (commented):

PR #13932, opened 2026-08-11
fix(poetry-env): preserve venv in project subdirectories

### LINKED ISSUE #12377: poetry-env gets deactivated if changing to a sub directory
### Describe the bug

Previous versions of the poetry-env plugin allowed changing to a subdirectory of a poetry directory while keeping the virtual environment active. With the new version changing into a subdirectory will deactivate the environment

### Steps to reproduce

1. Enable plugin poetry-env
2. Create a folder with poetry environment (.venv folder and pyproject.toml file)
3. Change to the folder with the poetry environment: the environment gets activated
4. Create a subdirectory in this folder
5. Change to the subdirectory: the environment gets deactivated

### Expected behavior

The environment should stay activated even with subdirectories and only change when changing to a directory that is does not contain a poetry environment or is not a subdirectory of a directory that contains a poetry environment. 

_No response_

### OS / Linux distribution

Ubuntu 22.04

### Zsh version

5.8.1

### Oh My Zsh version

master (ee69c14)

### Terminal emulator

GNOME terminal

None

### Additional context

Related to #12346. 
Specifically [here](https://github.com/ohmyzsh/ohmyzsh/pull/12346/files) line 9. 
`in_poetry_dir` will always be 0 when going from the main poetry directory to a subdirectory and `poetry_active` will always be 1 in that case, so the environment gets deactivated.
Probably the or-check here should be an and-check: we need to make sure that we are not in a poetry directory and not in a subdirectory of the poetry directory we were in earlier before we deactivate the environment.

nrps9909:
## Standards checklist:

## Changes:

- Keep the active Poetry environment while navigating ordinary project subdirectories.
- Deactivate and switch when entering a valid nested Poetry project.
- Match project paths on a directory boundary, so a sibling such as `project2` is not treated as a child of `project`.
- Document the two marker files and the navigation behavior.

Fixes #12377.

## Testing:

- Ran the repository's CI syntax command locally: all 579 matched Zsh files passed `zsh -n`.
- Ran an isolated `zsh -f` navigation smoke test covering project activation, ordinary and incomplete nested directories, a valid nested Poetry project, its child, returning to the parent project, a same-prefix sibling project, paths with spaces, and leaving the project tree.
- Reproduced the original child-directory deactivation against `origin/master`; all eight navigation scenarios pass with this branch.

## Other comments:

No aliases are introduced, so the alias-use-case checklist item is not applicable.

AI-assisted: Codex analyzed the existing state transitions, drafted the condition, and assembled the test matrix. The final condition was cross-checked against the reported regression and edge cases, and the tests above were run locally.

cc @carlosala

### DIFF
--- a/plugins/poetry-env/README.md
+++ b/plugins/poetry-env/README.md
@@ -1,7 +1,9 @@
 # Poetry Environment Plugin
 
-This plugin automatically changes poetry environment when you cd into or out of the project directory.
-Note: Script looks for pyproject.toml file to determine poetry if its a poetry environment
+This plugin automatically activates a Poetry environment when you enter a
+directory containing both `pyproject.toml` and `poetry.lock`. It keeps the
+environment active in project subdirectories, switches to nested Poetry
+projects, and deactivates it when you leave the project tree.
 
 To use it, add `poetry-env` to the plugins array in your zshrc file:
 
--- a/plugins/poetry-env/poetry-env.plugin.zsh
+++ b/plugins/poetry-env/poetry-env.plugin.zsh
@@ -1,15 +1,23 @@
 _togglePoetryShell() {
   # Determine if currently in a Poetry-managed directory
   local in_poetry_dir=0
+  local in_active_poetry_dir=0
   if [[ -f "$PWD/pyproject.toml" && -f "$PWD/poetry.lock" ]]; then
     in_poetry_dir=1
   fi
 
-  # Deactivate the current environment if moving out of a Poetry directory or into a different Poetry directory
-  if [[ $poetry_active -eq 1 ]] && { [[ $in_poetry_dir -eq 0 ]] || [[ "$PWD" != "$poetry_dir"* ]]; }; then
-    export poetry_active=0
-    unset poetry_dir
-    (( $+functions[deactivate] )) && deactivate
+  # Deactivate when leaving the active project or entering a nested project
+  if [[ $poetry_active -eq 1 ]]; then
+    if [[ "$PWD" == "$poetry_dir" || "$PWD" == "$poetry_dir"/* ]]; then
+      in_active_poetry_dir=1
+    fi
+
+    if [[ $in_active_poetry_dir -eq 0 ]] ||
+      { [[ $in_poetry_dir -eq 1 ]] && [[ "$PWD" != "$poetry_dir" ]]; }; then
+      export poetry_active=0
+      unset poetry_dir
+      (( $+functions[deactivate] )) && deactivate
+    fi
   fi
 
   # Activate the environment if in a Poetry directory and no environment is currently active
--- a/plugins/poetry-env/poetry-env.plugin.zsh
+++ b/plugins/poetry-env/poetry-env.plugin.zsh
@@ -8,7 +8,8 @@ _togglePoetryShell() {
 
   # Deactivate when leaving the active project or entering a nested project
   if [[ $poetry_active -eq 1 ]]; then
-    if [[ "$PWD" == "$poetry_dir" || "$PWD" == "$poetry_dir"/* ]]; then
+    local project_dir="${poetry_dir%/}"
+    if [[ "$PWD" == "$project_dir" || "$PWD" == "$project_dir"/* ]]; then
       in_active_poetry_dir=1
     fi
 


Morningstar202604:
Heads-up while this is in flight: `plugins/poetry-env/poetry-env.plugin.zsh:9` has the same string-prefix-vs-path-component slip that #13932 is working around, and it produces a different symptom from the subdirectory one this PR addresses.

```zsh
if [[ $poetry_active -eq 1 ]] && { [[ $in_poetry_dir -eq 0 ]] || [[ "$PWD" != "$poetry_dir"* ]]; }; then
```

`"$PWD" != "$poetry_dir"*` is satisfied by *any* directory whose path merely starts with the project's text, so an unrelated sibling project is treated as "still inside the project" and the first environment is never released.

Reproduced on current master with two sibling projects `proj` and `proj-extra`, each with `pyproject.toml` + `poetry.lock` and its own venv, `poetry env info --path` stubbed to answer per `$PWD`:

```
start in proj       -> VIRTUAL_ENV=/tmp/.../venvs/proj
after cd proj-extra -> VIRTUAL_ENV=/tmp/.../venvs/proj     <- expected /tmp/.../venvs/proj-extra
                    -> poetry_dir=/tmp/.../proj
```

`poetry_dir` is still the old project and `VIRTUAL_ENV` still points into project A's venv, so commands run in project B are executed with project A's interpreter and dependencies. `in_poetry_dir` is 1 there (B has its own `pyproject.toml`/`poetry.lock`, just not the *same* one), so the `[[ $in_poetry_dir -eq 0 ]]` half does not fire and the prefix test is what lets it through.

Suggesting the component-boundary form rather than a prefix test — the same shape I landed for #14006 and used in #14021:

```zsh
local project_dir="${poetry_dir%/}"
if [[ "$PWD" != "$project_dir" && "$PWD" != "$project_dir"/* ]]; then
```

Truth table, `stay` = keep the environment:

| `poetry_dir` | `PWD` | current | with the above |
|---|---|---|---|
| `/x/proj` | `/x/proj` | stay | stay |
| `/x/proj` | `/x/proj/sub/dir` | stay | stay |
| `/x/proj` | `/x/proj-extra` | stay (bug) | deactivate |
| `/` | `/tmp` | stay | stay |

Not posting a competing patch since this PR is already open on the file — just flagging it so the fix can cover both cases if you want to fold it in.

nrps9909:
Thanks for the detailed boundary table. The existing PR head already used the component-aware `"$PWD" == "$poetry_dir"/*` check, and a fresh zsh probe confirmed that `/x/proj-extra` deactivates while `/x/proj/sub` stays active. Your `/` case exposed a remaining edge, though: an unnormalized root produced `//*`, which did not match `/tmp`. Signed follow-up `a7248eac` now normalizes the trailing slash before both exact and descendant comparisons. Revalidated states: sibling project -> deactivated once; ordinary child -> active with zero deactivations; root child `/tmp` -> active with zero deactivations. `zsh -n` and `git diff --check` pass.

Morningstar202604:
You are right about the scope, and thanks for chasing it down instead of letting it slide — my note mis-attributed the bug. The string-prefix slip is in **master's** `_togglePoetryShell`; I verified the reproduction against master and framed the note as if this PR's head also had it, which it did not. Sorry for the noise.

I have re-run the same stubbed-`poetry` probe against your current head (`a7248ea`), and all three behaviours you listed check out independently:

- `proj` -> `proj/subdir`, where the child has no `pyproject.toml`/`poetry.lock`: the environment **stays active** on this branch; on master the same `cd` drops it, which is the case this PR fixes.
- `proj` -> sibling `proj-extra` with its own project files and venv: the branch switches to the sibling's own environment rather than keeping the old one.
- leaving the project tree for an unrelated directory: deactivates.

I also exercised the containment condition from `a7248ea` directly rather than only through the transitions, including the unnormalized-root case you fixed: with `poetry_dir=/`, the normalized `project_dir` keeps `/`, `/tmp` and `/usr/local` inside the project; with `poetry_dir=/x/proj`, the exact path and children stay and a sibling like `/x/proj-docs` deactivates. That matches your commit message.

Nothing outstanding from me on this one.

Morningstar202604:
Clarification: The string-prefix bug I initially mentioned is in **master's** `_togglePoetryShell`, not in this PR. This PR correctly uses component-aware path matching (`"$PWD" == "$poetry_dir"/*`). The reproduction and fix scope are accurate — ready for review.

robbyrussell (approved):
Nice catch, and thanks for the writeup.

This one's more annoying than it looks in the diff... on master, walking into any ordinary subdirectory of your own project killed the venv, because `in_poetry_dir` only goes to 1 when the current directory itself has both marker files. So the fix isn't really about sibling directories, it's about the most normal thing you can do in a project.

I ran the decision table against both branches. Subdirectories stay active now, nested projects switch properly, and a project at `/` still behaves. Three of those cases were wrong on master, so this fixes more than it advertises.

One tiny thing, not for this PR: `venv_dir` on line 17 still isn't `local`, so it leaks into the global namespace. Pre-existing, and I'll pick it up separately.

Merging. Thanks!

robbyrussell:
Follow-up on the `venv_dir` note above: #14043.

PR #13933, opened 2026-08-12
feat(theme): Add JJ prompt to Agnoster theme

eunchan:
## Standards checklist:

## Changes:

- add prompt_jj() to agnoster theme and replace prompt_git in build_prompt to prompt_jj
- prompt_jj() to check if repo is jj repo then add the bookmark or change-id then falls back to prompt_git

### DIFF
--- a/themes/agnoster.zsh-theme
+++ b/themes/agnoster.zsh-theme
@@ -63,6 +63,12 @@ esac
 : ${AGNOSTER_GIT_DIRTY_FG:=black}
 : ${AGNOSTER_GIT_DIRTY_BG:=yellow}
 
+# Jujutsu related
+: ${AGNOSTER_JJ_CLEAN_FG:=${CURRENT_FG}}
+: ${AGNOSTER_JJ_CLEAN_BG:=green}
+: ${AGNOSTER_JJ_DIRTY_FG:=black}
+: ${AGNOSTER_JJ_DIRTY_BG:=yellow}
+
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for this, and welcome! Agnoster picking up jj is a reasonable thing to want.

Before going through it line by line, I think the approach wants a rethink... the jj plugin already has most of this built.

`plugins/jj/jj.plugin.zsh` ships `jj_prompt_template` and `jj_prompt_template_raw` (they're literally commented "convenience functions for themes"), and the plugin's README shows the git-fallback wrapper pattern. It also has a `:omz:plugins:jj ignore-working-copy` zstyle so people can skip jj's working-copy snapshot on every prompt. Building on those means you inherit all of that instead of maintaining a second copy of it.

The other thing is cost. `jj root` runs on every prompt in every directory for anyone who has jj installed, whether or not they're in a jj repo. That's a fork per prompt for a lot of people who'd never notice why their shell got slower. Have a look at `prompt_bzr` in the same file... it walks up looking for the directory marker before spawning anything. A `.jj` check first would do the same job here.

Smaller thing while you're in there: `prompt_segment` takes `(bg, fg, text)` and the call passes fg first, so the segment renders inverted. And `AGNOSTER_JJ_DIRTY_FG`/`_BG` get defined but never used, so there's no dirty state yet.

Happy to look again once it's built on the plugin's helpers.

Issue #10982, opened 2022-06-09
History Plugin: add alias with tail


SIMULATAN:
### If the feature request is for a plugin or theme, specify it here.

History (https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/history)

### If the feature solves a problem you have, specify it here.

If you try to search the last n commands including a certain string using the history plugin, you can easily get overwhelmed by too many results.

### Describe the proposed feature.

It would be great to have an additional alias that greps in the history with the tail command so only a certain amount of results are shown.

### Describe alternatives you've considered

Just use the output of the `hs` alias and pipe it into tail
This is a completely valid solution, however, because omz alias plugins are all about convenience and quick access, this might as well be a good addition.

mcornella:
That's not possible with an alias, it would have to be a function. One alternative that I use is a global alias (a global shorthand) that can be quickly added to a command without having to type the whole `| tail`:
```sh
alias -g :T='| command tail'
```
If we add the `| tail`, we would also need to add the `| head`, so I'm not sure this is so good to have. Maybe you could use [`fzf`](https://github.com/junegunn/fzf) instead of the `history` plugin and do `history | fzf`?

robbyrussell:
Hey @SIMULATAN, coming back to this one late, sorry.

Marc's reply still holds, and you'd already landed on the answer yourself: `hs pattern | tail -20` does the job, and `history -20` covers a plain last-N listing.

We're going to leave the plugin as it is. It's a deliberately small one — `h`, `hl`, `hs`, `hsi` — and doing this properly needs a function taking a count rather than an alias, at which point `| head` wants one too. Every alias we add is muscle memory we can't take back later.

Closing it out. Thanks for raising it.

🤖 Co-authored with Claude Code

PR #13936, opened 2026-08-13
feat(history): add aliases to show only the last 20 matches

deepshekhardas:
Adds \hst\ and \hsti\, which pipe \history\ search results through \	ail -20\ so a broad grep on a long history is not overwhelming.\n\nCloses #10982


carlosala (commented):

PR #13937, opened 2026-08-13
feat(zsh-interactive-cd): add opt-in zic_accept_last to auto-accept single match

### LINKED ISSUE #13131: zsh-interactive-cd does not automatically accept the last remaining match

zsh-interactive-cd

fzf contains a feature to automatically accept the last remaining option via `fzf --bind one:accept` but this plugin does not contain configuration to opt into that behaviour.

Add an opt-in configuration option for this feature.

The only alternatives in this situation are to not use the plugin or to fork it and source my own instead of from omz.

### Additional context

No additional context.

### Related Issues

https://github.com/ohmyzsh/ohmyzsh/issues/13130 already asks for one configuration option, it would save effort to add multiple of them at once.

deepshekhardas:
Adds an opt-in `zic_accept_last` configuration variable. When set to `true`, the plugin appends `one:accept` to the fzf bindings so the last remaining match is accepted automatically instead of requiring Enter.

```zsh
zic_accept_last=true
```

Closes #13131

### DIFF
--- a/plugins/zsh-interactive-cd/README.md
+++ b/plugins/zsh-interactive-cd/README.md
@@ -19,3 +19,11 @@ This plugin provides an interactive way to change directories in zsh using fzf.
 ## Usage
 
 Press tab for completion as usual, it'll launch fzf automatically. Check fzf’s [readme](https://github.com/junegunn/fzf#search-syntax) for more search syntax usage.
+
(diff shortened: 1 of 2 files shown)


carlosala (commented):

PR #13941, opened 2026-08-13
feat(dotenv): support loading multiple dotenv files via ZSH_DOTENV_FILES

### LINKED ISSUE #13833: DotEnv multiple files

DotEnv

I work in a few projects that have either 1 or sometimes 2 env files.

Could we get the ability to specify multiple env files to load in the order specified?

I can't really think of an alternative as the split wont realistically go away any time soon.

### Additional context

_No response_

### Related Issues

_No response_

deepshekhardas:
Adds a `ZSH_DOTENV_FILES` variable to load additional dotenv files in order, after `ZSH_DOTENV_FILE`:

```zsh
ZSH_DOTENV_FILES=".env.local .env.override"
```

Only existing files are loaded, and later files take precedence. Backward compatible: the default behavior (single file) is unchanged.

Closes #13833

### DIFF
--- a/plugins/claude/README.md
+++ b/plugins/claude/README.md
@@ -1,0 +1,20 @@
+# Claude Code plugin
+
+This plugin provides aliases for the [Claude Code](https://code.claude.com/docs/en/cli) CLI.
+
+To use it, add `claude` to the plugins array in your .zshrc file:
+
+```zsh
+plugins=(... claude)
+```
+
+## Aliases
+
+| Alias | Command                 | Description                               |
+|-------|-------------------------|-------------------------------------------|
+| `cc`  | `claude`                | Interactive prompt                         |
+| `ccx` | `claude --continue`     | Continue the most recent conversation      |
+| `ccr` | `claude --resume`       | Resume a specific conversation by session  |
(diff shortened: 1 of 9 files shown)


mcornella:
Hi @deepshekhardas, I don't know if I'm talking to a person or a bot. Please update this Pull Request to only contain changes related to the PR description, and don't override the Pull Request template when creating it. It is important for us to understand that the steps have been followed, and particularly that AI usage is disclosed.

Additional review comments:
1. The tests use `basic-parsing.zunit`, but it's better to create a separate one so that the additional fixture setup only happens for this part of the documentation; otherwise it's just a waste of resources.
2. The current change doesn't prompt for all env files; it should notify all files found.
3. The preparation step doesn't remove duplicates from the array; if a file is both in `$ZSH_DOTENV_FILE` and `$ZSH_DOTENV_FILES` it will be parsed twice.
4. The parsing logic doesn't take into account env variables found in multiple .env files, such that for example this is processed correctly:
```sh
# settings
ZSH_DOTENV_FILES=".env.override .env"

# .env
HOST="app.localhost:3000"
API_URL="http://${HOST}/api/v1"
# .env.override
HOST="app.example.com"
```

carlosala (commented):

PR #13944, opened 2026-08-14
feat(updater): add cooldown option to delay applying updates

### LINKED ISSUE #13813: Introduce a cooldown feature for the auto-updater

deepshekhardas:
Closes #13813

Introduces a new `zstyle ':omz:update' cooldown <days>` setting that limits the updater to only apply commits that are at least N days old. Defaults to 0 (current behavior - always pull latest).

When cooldown is set, the updater fetches the remote branch and finds the most recent commit whose committer timestamp is at least N days old, then applies it via `git merge --ff-only`.

## Motivation
Automated update runs (e.g. cron) can pull in a newly-broken commit before it is caught. A cooldown lets users delay adoption of fresh commits while still staying up to date.

## Changes
- `tools/upgrade.sh`: cooldown check + safe apply
- `templates/zshrc.zsh-template`: documented zstyle
- `README.md`: docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,15 @@ zstyle ':omz:update' frequency 7
 zstyle ':omz:update' frequency 0
 ```
 
+By default, updates always pull the latest changes. If you'd rather let others kick the tires first
+before an update reaches your machine, you can set a cooldown (in days). You'll still get everything —
+just a little later:
+
+```sh
+# Only apply updates that are at least 10 days old
+zstyle ':omz:update' cooldown 10
+```
+
 ### Updates Verbosity
 
 You can also limit the update verbosity with the following settings:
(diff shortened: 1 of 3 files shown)


mcornella:
It's rude / bad etiquette to duplicate other people's work. Duplicate of #13814

PR #13948, opened 2026-08-14
feat(systemadmin): replace clrz with pszombie alias

deepshekhardas:
Closes #13861

Replaces the broken `clrz` function with a new `pszombie` alias in the systemadmin plugin.

## Why
- The old `clrz` function used `ps -eal` output parsed with awk, then piped PIDs into `kill -9` — but zombie processes cannot be killed (their parent must reap them), so it was misleading.
- The new `pszombie` alias lists zombie processes with user, PID, PPID, state and command for actual diagnosis.
- Old `clrz` kept but commented out for reference.

## Change
- `plugins/systemadmin/systemadmin.plugin.zsh`: +6/-4


carlosala (commented):

PR #13951, opened 2026-08-14
fix(jj): deprecated destination arg with onto

JaKXz:
## Standards checklist:

## Changes:

- The `--destination` / `-d` arguments were renamed to `--onto` / `-o` in v36: https://docs.jj-vcs.dev/v0.36.0/changelog/#deprecations_1 so this PR updates the outdated use here :)


JaKXz:
Hi @carlosala sorry to ping directly but I think this is a quick win :)

robbyrussell:
Thanks for this! One question before I take it.

I went looking at jj's source and `-d`/`--destination` are still *visible* aliases of `-o`/`--onto` as of v0.45.1... no hidden flag, no deprecation warning. So `-d` works on every jj release, while `-o` only exists from v0.36 onward (December 2025). Switching would break `jjrbm` for anyone on a distro-packaged jj older than that.

So: did `-d` actually break for you somewhere, or is this getting ahead of a removal you saw coming? If upstream has said it's going away I'm happy to take it... I just don't want to raise the version floor for a rename that hasn't landed yet.

Either way, `plugins/jj/README.md` line 45 still documents `jjrbm` as `-d`, so that row needs to move in the same PR or the table and the alias will disagree.

Heads up too that #13962 touches this same file and is waiting on a rebase, so whichever goes second will need a small one.

JaKXz:
> If upstream has said it's going away I'm happy to take it... I just don't want to raise the version floor for a rename that hasn't landed yet.

Fair; This is just getting ahead of a removal (`-d` is working fine for now), so maybe better to wait! 

Also, sorry I forgot to update the readme, and then when I did my editor decided some formatting was in order. If you'd rather close that's totally fine, sorry for the noise!

joshka:
I suspect that the answer to this really comes down to working out what the targets where you might use 0.36 are. I'm a strong believer that you should be using generally up to date developer tools in all places that you work. This is a fairly opinionated view from someone that doesn't tend to work in spaces where they can't update their own tools though.

robbyrussell:
@copilot resolve the merge conflicts in this pull request

carlosala (approved):

carlosala:
Thanks!

PR #13962, opened 2026-08-16
feat(jj): add common command aliases

joshka:
## Standards checklist:

## Changes:

Add commonly used aliases for core jj commands and subcommands:

- `jjblt` → `jj bookmark list --tracked`
- `jjcfg` → `jj config`
- `jjcfgl` → `jj config list`
- `jjdst` → `jj diff --stat`
- `jjev` → `jj evolog`
- `jjf` → `jj file`
- `jjfl` → `jj file list`
- `jjg` → `jj git`
- `jjgi` → `jj git init`
- `jjgrl` → `jj git remote list`
- `jjop` → `jj op`
- `jjopl` → `jj op log`
- `jjor` → `jj op restore`
- `jjs` → `jj show`
- `jju` → `jj undo`
- `jjw` → `jj workspace`
- `jjwl` → `jj workspace list`

The README also documents the previously added `jjwa` and `jjwf` aliases.

These aliases provide short, consistent access to commands that are useful during normal jj workflows:

- Inspecting changes with `jjs`, `jjdst`, and `jjf`/`jjfl`
- Reviewing tracked bookmarks and configuration with `jjblt`, `jjcfg`, and `jjcfgl`
- Working with Git repositories and remotes using `jjg`, `jjgi`, and `jjgrl`
- Inspecting and recovering operations with `jjev`, `jjop`, `jjopl`, `jjor`, and `jju`
- Managing workspaces with `jjw`, `jjwl`, `jjwa`, and `jjwf`

The new aliases copy existing compositional conventions for symmetry. For example, the existing `jjb` maps to `jj bookmark`, while `jjbl` maps to `jj bookmark list`. This change applies the same pattern to `jjg`/`jjgf`, `jjw`/`jjwl`, and `jjop`/`jjopl` for Git, workspace, and operation commands.

This makes the aliases easier to remember without needing to recall which underlying jj commands use spaces or built-in command aliases.

## Other comments:

Review from previous jj plugin contributors would be especially valuable:

- [@nasso](https://github.com/nasso), the original jj plugin author, introduced the plugin in [#12292](https://github.com/ohmyzsh/ohmyzsh/pull/12292).
- [@imp](https://github.com/imp), who contributed several earlier jj aliases in [#13105](https://github.com/ohmyzsh/ohmyzsh/pull/13105) and [#13179](https://github.com/ohmyzsh/ohmyzsh/pull/13179).
- [@danvratil](https://github.com/danvratil), who added the existing workspace aliases in [#13850](https://github.com/ohmyzsh/ohmyzsh/pull/13850).
- [@laurensrouw](https://github.com/laurensrouw), who added the existing bookmark alias in [#13638](https://github.com/ohmyzsh/ohmyzsh/pull/13638).

[@martinvonz](https://github.com/martinvonz) may also be a useful reviewer for checking that the selected aliases match common jj usage and terminology, given the discussion in [this review comment](https://github.com/ohmyzsh/ohmyzsh/pull/13179#discussion_r2241244638).

Validation performed:

- `zsh -n plugins/jj/jj.plugin.zsh`
- Sourced the plugin and verified the new alias expansions
- Checked the relevant jj command help output
- Verified that every plugin alias is documented in the README
- Reviewed the resulting jj diff

AI-assisted: Codex helped draft and apply the alias and README changes and run validation. I reviewed the resulting diff and manually verified the aliases and command targets.


JaKXz:
I wouldn’t block on this but, while these are useful I would worry about adding too many aliases that are also very close to each other…

joshka:
> I wouldn’t block on this but, while these are useful I would worry about adding too many aliases that are also very close to each other…

Fair point, but the contrast is taking a look at https://github.com/ohmyzsh/ohmyzsh/blob/master/plugins/git/git.plugin.zsh for comparison. The main rationale for most of thsee is when I go to type an alias whether I'd anticipate that it should already be there when considering the other aliases already existing in the aliases. There's a strong amount of symmetry here. In particular, I regularly reach for `jj git init` for when I'm looking at a repo I've cloned previously with gh, `jj git remote list` to understand what the origin/upstream/fork remotes look like, `jj workspace` (I'd prefer this as `jjws`, but `jjwa` and `jjwf` already sort of established that workspace was w not ws.) and `jjwl` because I often work in workspaces for AI assisted parallel work.

The rest are mostly things that feel like they need to be there for symmetry.

robbyrussell:
Thanks for this, and thanks @JaKXz for raising the size question... that was the right thing to ask.

I'm good with it. These are mostly distinct subcommands rather than variations on aliases we already ship, and they follow the `jjb`/`jjbl` pattern the plugin already set. So it's filling out something we committed to rather than starting something new.

Only thing needed is a rebase... it conflicts on a trailing blank line at the end of `jj.plugin.zsh`, so it should be a quick one. Heads up that #13951 also touches this plugin, so whichever lands second will need another small rebase.

Ping me once it's rebased and I'll get it in.

JaKXz:
@joshka maybe it would be better for this PR to just supersede mine? If you're happy to make the update to `jjrb` so it uses `-o` instead of `-d`, then I can just close #13951

joshka:
@robbyrussell I've rebased on main

@JaKXz I think perhaps keep the -o/-d thing separate makes sense. I'm in general agreement with it, but would prefer to keep that separate to make it easy to revert if needed.

robbyrussell (approved):

robbyrussell:
lgtm, thjanks to you both for helping coordinate on this!

PR #13965, opened 2026-08-18
fix(eza): enhance hyperlink option handling in eza plugin

busybox11:
## Standards checklist:

## Changes:

- Updated the hyperlink configuration logic to handle the new eza parameter syntax, with a fallback to "auto" for existing setups

## Other comments:

`eza` commit implementing the new `--hyperlink` usage: https://github.com/eza-community/eza/commit/72e323171ca7466b62c5d2a5d0da8d1b5edc9593

### DIFF
--- a/plugins/eza/eza.plugin.zsh
+++ b/plugins/eza/eza.plugin.zsh
@@ -53,7 +53,12 @@ function _configure_eza() {
     _EZA_TAIL+=("--time-style='$_val'")
   fi
   if zstyle -t ":omz:plugins:eza" "hyperlink"; then
-    _EZA_TAIL+=("--hyperlink")
+    case "${_val:l}" in
+      yes|true|on|1) _val=auto ;;
+    esac
+    if [[ ${_val:l} == (always|auto|automatic|never) ]]; then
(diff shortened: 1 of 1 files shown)


robbyrussell:
Thanks for digging into this, and sorry for the slow reply.

I'm going to close this in favor of #14024, which fixes the same bug. Two things led me there:

The `hyperlink` style is read with `zstyle -t`, which only returns true for boolean values, so `auto`, `always` and `never` never make it past the outer `if` and into the `case`:

```
hyperlink yes     -> TRUE   block runs
hyperlink always  -> false  block SKIPPED
hyperlink never   -> false  block SKIPPED
```

`$_val` is also still holding the `time-style` value at that point, so in practice no `--hyperlink` flag gets emitted at all.

The other piece is old eza. Ubuntu 24.04 LTS ships 0.18.2 and Debian stable has 0.21.0, and neither accepts a WHEN value here, so we need to keep passing the bare flag on those.

That said, the idea underneath this — letting people pick the WHEN value instead of hardcoding one — is a good one, and I'd be happy to see it as a follow-up on top of #14024. It would need `zstyle -s` rather than `zstyle -t` to read the value.

PR #13967, opened 2026-08-21
fix(kubectl): avoid racing compinit with async completion generation

### LINKED ISSUE #13964: Some oh-my-zsh plugins like kubectl generate completions async in the background, racing with later compinit
### Describe the bug

With some plugins like kubectl, that generate the completions file asynchrnously in the background, they end up racing with a subsequent compinit which causes it to often read them as emtpy files, missing their `#compdef` lines and as such, not completing for them. The logic in this script to push it directly into comps doesn't work as compinit tramples that.

Please don't rule this out just cause this isn't the official directly install of omz, many users install omz and plugins via varying package manager, that have differing mechanisms for running compinit that all can hit this.

### Steps to reproduce

1. Use omz via antidote & mattmc3/antidote-use-omz
2. Enable the kubectl plugin of omz.
3. Delete the zcompdump anda restart the shell
4. Often kubectl will have no completions

### Expected behavior

With the kubectl plugin enabled, kubectl completions should just work.

_No response_

### OS / Linux distribution

macOS 26.6

### Zsh version

5.9

### Oh My Zsh version

master (97e11051)

### Terminal emulator

iTerm2

_No response_

### Additional context

Session transcript from Claude Code figuring this out [2026-08-17-161319-something-broke-my-zsh-completions-for-kubectl-t.txt](https://github.com/user-attachments/files/31144077/2026-08-17-161319-something-broke-my-zsh-completions-for-kubectl-t.txt)

deepshekhardas:
Fixes #13964

The kubectl plugin generated completions in the background directly into `$ZSH_CACHE_DIR/completions/_kubectl`. A `compinit` running after the plugin loads (common with package-manager installs like antidote, zinit, etc.) could read the file mid-write, cache it without its `#compdef` line, and leave kubectl without completions.

The generation now writes to a temp file and `mv`s it into place atomically, so concurrent compinit runs either see the previous complete file or the new one — never a partial one.

### DIFF
--- a/plugins/kubectl/kubectl.plugin.zsh
+++ b/plugins/kubectl/kubectl.plugin.zsh
@@ -10,7 +10,11 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_kubectl" ]]; then
   _comps[kubectl]=_kubectl
 fi
 
-kubectl completion zsh 2> /dev/null >| "$ZSH_CACHE_DIR/completions/_kubectl" &|
+# Generate the completion into a temporary file and move it into place
+# atomically, so a compinit running concurrently (package-manager setups
+# may run it after this plugin) never reads an empty or partial file.
+local _kubectl_tmp="$ZSH_CACHE_DIR/completions/_kubectl.tmp.$$"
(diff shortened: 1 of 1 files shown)


carlosala (commented):

PR #13969, opened 2026-08-21
fix(helm): avoid racing compinit with async completion regeneration

deepshekhardas:
Same race as #13964 (kubectl): the helm plugin regenerated its completion file in the background via `tee`, writing directly to `$ZSH_CACHE_DIR/completions/_helm`. A `compinit` running concurrently (common with package-manager oh-my-zsh installs) could read the file mid-write and cache it as empty.

Now writes to a temp file and `mv`s it into place atomically.


carlosala (commented):

PR #13970, opened 2026-08-21
fix(argocd): avoid racing compinit with async completion generation

deepshekhardas:
Same race as #13964 (kubectl): the argocd plugin generated completions in the background directly into `$ZSH_CACHE_DIR/completions/_argocd`. A `compinit` running concurrently could read a partial file and cache it as empty.

Now writes to a temp file and moves it into place atomically.


carlosala (commented):

PR #13981, opened 2026-08-21
fix(gh): avoid racing compinit with async completion generation

deepshekhardas:
fix(gh): avoid racing compinit with async completion generation

The gh plugin generated completions in the background directly into
$ZSH_CACHE_DIR/completions/_gh. A compinit running after the
plugin loads (common with package-manager installs like antidote,
zinit, etc.) could read the file mid-write, cache it without its
#compdef line, and leave gh without completions.

The generation now writes to a temp file and mv's it into place
atomically, so concurrent compinit runs either see the previous
complete file or the new one — never a partial one.


carlosala (commented):

PR #14000, opened 2026-08-22
fix: load completion files atomically

### LINKED ISSUE #13964: Some oh-my-zsh plugins like kubectl generate completions async in the background, racing with later compinit

carlosala:
## Standards checklist:

## Changes:

- Generate completion into tmp files before writing it to `cache/completion`.

Closes #13964

### DIFF
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -11,7 +11,7 @@ fi
 
 {
   local completion="$ZSH_CACHE_DIR/completions/_op" tmp
-  tmp=$(command mktemp "$completion.XXXXXX") || exit
+  tmp=$(command mktemp -t _omz_comp.XXXXXXXX) || exit
   op completion zsh >| "$tmp" && command mv -f "$tmp" "$completion"
   command rm -f "$tmp"
 } &|
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -9,11 +9,10 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_op" ]]; then
   _comps[op]=_op
 fi
 
-{
-  local completion="$ZSH_CACHE_DIR/completions/_op" tmp
-  tmp=$(command mktemp -t _omz_comp.XXXXXXXX) || exit
-  op completion zsh >| "$tmp" && command mv -f "$tmp" "$completion"
-  command rm -f "$tmp"
+zmodload -F zsh/files b:zf_mv
+() {
+  local TMPPREFIX="$ZSH_CACHE_DIR/completions/_op"
+  zf_mv -f -- =( op completion zsh ) "$TMPPREFIX"
 } &|
 
 # Load opswd function
--- a/plugins/1password/1password.plugin.zsh
+++ b/plugins/1password/1password.plugin.zsh
@@ -9,7 +9,12 @@ if [[ ! -f "$ZSH_CACHE_DIR/completions/_op" ]]; then
   _comps[op]=_op
 fi
 
-op completion zsh >| "$ZSH_CACHE_DIR/completions/_op" &|
+{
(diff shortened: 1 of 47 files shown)


mcornella (changes requested):
Hi Carlo, I've iterated over the final solution to avoid forks and keep it atomic, as likely we will introduce a lot of overhead if multiple plugins are enabled.

Decisions and explanation below (the anon function is necessary so that local works correctly). Can you test and confirm this does the same and apply the same pattern everywhere? Thanks! 

```sh
zmodload -F zsh/files b:zf_mv
() {
  local TMPPREFIX="$ZSH_CACHE_DIR/completions/_op"
  zf_mv -f -- =( op completion zsh ) "$TMPPREFIX"
} &|
```

### What Everything Does and the Decisions Taken

*   **`zmodload -F zsh/files b:zf_mv`**
    *   **Decision:** Safely load Zsh's builtin `mv` with a prefixed name.
    *   **Why:** Using `-F` and `b:zf_mv` injects *only* `zf_mv` into your interactive session forever. Your system's standard `/bin/mv` remains completely untouched, ensuring zero conflicts or unexpected behaviors for your everyday terminal use or other scripts.

*   **`local TMPPREFIX="$ZSH_CACHE_DIR/completions/_op"`**
    *   **Decision:** Consolidate the target path and the temporary file prefix into a single variable.
    *   **Why:** Zsh uses `TMPPREFIX` as the base string when generating temp files for `=()`. By setting it exactly to our desired final path, the temporary file is created in the exact same directory (e.g., `.../_op<random_suffix>`), guaranteeing a lightning-fast atomic `rename()` system call. We then reuse the same variable as the destination path for `zf_mv`, eliminating the need for a redundant `completion` variable.

*   **`=( op completion zsh )`**
    *   **Decision:** Rely on Zsh's native process substitution instead of `mktemp`.
    *   **Why:** The `=()` operator natively handles securely generating a unique temp file (using our `TMPPREFIX`), running the command, writing the output, and returning the temp file's path. This elegantly eliminates the need to fork `/bin/mktemp`.

*   **`zf_mv -f -- ... "$TMPPREFIX"`**
    *   **Decision:** Move the file using the builtin to our target destination.
    *   **Why:** `zf_mv` performs the atomic replacement instantly within the current Zsh process, avoiding a fork to the external `/bin/mv` binary. 

*   **Omission of `rm` (Cleanup)**
    *   **Decision:** Let Zsh's native garbage collection handle cleanup.
    *   **Why:** When `=()` finishes, Zsh automatically attempts to delete the temporary file it created. Because `zf_mv` already moved the file to its final destination, Zsh's cleanup silently does nothing. We drop the `/bin/rm` fork and save a line of code.

mcornella (approved):
LGTM

PR #14004, opened 2026-08-24
fix(docker): resolve completion path correctly in anonymous function

### LINKED ISSUE #14003: Docker legacy completion resolves bundled file path relative to PWD after #14000
### Describe the bug

fter updating Oh My Zsh to commit 830a5bc, loading a new Zsh session with the Docker plugin and legacy completion enabled produces the following error:

`cp: cannot stat './completions/_docker': No such file or directory
`

The bundled completion file exists at:

`$ZSH/plugins/docker/completions/_docker
`

However, the plugin tries to find it at:

./completions/_docker`

`The Docker plugin file has no local modifications.

### Steps to reproduce

1. Enable the Docker plugin in `.zshrc`:
```
plugins=(
  docker
)
```
2. Enable Docker legacy completion before sourcing Oh My Zsh:

`zstyle ':omz:plugins:docker' legacy-completion yes `

> source "$ZSH/oh-my-zsh.sh"

3. Start a new Zsh session:

> exec zsh

4. Observe the error:

`cp: cannot stat './completions/_docker': No such file or directory`

The same error also appears immediately after updating Oh My Zsh to 830a5bc.

### Expected behavior

The plugin should copy the bundled completion file from:

`$ZSH/plugins/docker/completions/_docker`

to:

`$ZSH_CACHE_DIR/completions/_docker`

No error should be displayed when starting or reloading Zsh.

## Actual behavior

The source path is resolved as:

`./completions/_docker`

This makes the path relative to the current working directory. When Zsh starts from the home directory, the plugin therefore looks for:

`$HOME/completions/_docker
`
instead of the completion file located in the Docker plugin directory.

##  Suspected regression

This appears to have been introduced by commit [830a5bc](https://github.com/ohmyzsh/ohmyzsh/commit/830a5bcfd29fd577fdcd5f3b8e98cbaf973421fa), through #14000.

Before that commit, the completion code was executed in a command group:

```
{
  # ...
  command cp "${0:h}/completions/_docker" \
    "$ZSH_CACHE_DIR/completions/_docker"
} &|
```

The commit changed the command group into an anonymous function:

```
zmodload -F zsh/files b:zf_mv

() {
  # ...
  command cp "${0:h}/completions/_docker" \
    "$ZSH_CACHE_DIR/completions/_docker"
} &|
```

The cp expression itself was not changed, but its execution context was.

Inside the anonymous function, $0 becomes (anon) on my system. Consequently:

`${0:h}
`
evaluates to:

`.`

A minimal Zsh test reproduces this behaviour:

```
setopt functionargzero
0=/tmp/example/plugins/docker/docker.plugin.zsh

print -r -- "outside: $0 -> ${0:h}"

() {
  print -r -- "inside: $0 -> ${0:h}"
}

```
Result:

```
outside: /tmp/example/plugins/docker/docker.plugin.zsh -> /tmp/example/plugins/docker
inside: (anon) -> .
```

The issue only becomes visible when the legacy completion branch executes:

> zstyle -t ':omz:plugins:docker' legacy-completion

In my environment, this command returns 0, confirming that legacy completion is enabled.

One possible solution would be to resolve the plugin directory before entering the anonymous function and pass it as an argument, instead of resolving ${0:h} inside the function.

For example:

```
() {
  local plugin_dir="$1"

  # ...
  command cp "$plugin_dir/completions/_docker" \
    "$ZSH_CACHE_DIR/completions/_docker"
} "${0:A:h}" &|
```

_No response_

### OS / Linux distribution

OS: Debian GNU/Linux 13 (trixie)

### Zsh version

Zsh: 5.9 (x86_64-debian-linux-gnu)

### Oh My Zsh version

master (830a5bc)

### Terminal emulator

xcfe Terminal

None

### Additional context

- Docker: `26.1.5+dfsg1`, `build a72d7cd
`- Docker legacy completion: enabled
- `ZSH_CACHE_DIR: $HOME/.oh-my-zsh/cache`
- ` plugins/docker/docker.plugin.zsh`: no local modifications

sony-level:
## Description

Fix the legacy Docker completion path resolution inside the background
anonymous function.

Inside the anonymous function, `$0` refers to the function itself rather
than `plugins/docker/docker.plugin.zsh`. As a result, using `$0` from
inside the function can resolve the bundled completion file from an
incorrect directory.

This change evaluates `${0:A:h}` before entering the anonymous function,
passes the absolute plugin directory as an argument, and stores it in the
local `plugin_dir` variable.

The legacy completion file is then resolved using:

```zsh
"${plugin_dir}/completions/_docker"
```

Fixes #14003

## Standards checklist:

## Changes:

- [fix/docker-completion-cache](https://github.com/sony-level/ohmyzsh/compare/fix/docker-completion-cache)

## Other comments:

Tested with:

- Debian GNU/Linux 13 (trixie)
- Zsh 5.9
- Docker 26.1.5+dfsg1
- Docker legacy completion enabled

Checks performed:

- Run `zsh -n plugins/docker/docker.plugin.zsh.`
- Started a new Zsh session with legacy Docker completion enabled.
- Verified that the completion cache was generated without the incorrect path error.
- Verified completion for Docker commands.

...

### DIFF
--- a/plugins/docker/docker.plugin.zsh
+++ b/plugins/docker/docker.plugin.zsh
@@ -58,14 +58,17 @@ fi
 
 zmodload -F zsh/files b:zf_mv
 () {
+
+  local plugin_dir="$1"
+
   # `docker completion` is only available from 23.0.0 on
   # docker version returns `Docker version 24.0.2, build cb74dfcd85`
   # with `s:,:` remove the comma after the version, and select third word of it
   if zstyle -t ':omz:plugins:docker' legacy-completion || \
     ! is-at-least 23.0.0 ${${(s:,:z)"$(command docker --version)"}[3]}; then
-        command cp "${0:h}/completions/_docker" "$ZSH_CACHE_DIR/completions/_docker"
+        command cp "${plugin_dir}/completions/_docker" "$ZSH_CACHE_DIR/completions/_docker"
       else
         local TMPPREFIX="$ZSH_CACHE_DIR/completions/_docker"
         zf_mv -f -- =( command docker completion zsh ) "$TMPPREFIX"
   fi
-} &|
+} "${0:A:h}" &|


mcornella (approved):
LGTM

PR #14006, opened 2026-08-24
fix(git): force C locale only while matching status regexes

### LINKED ISSUE #13985: Git prompt status causing "character not in range" error
### Describe the bug

I have a custom theme that sets characters for the different git states. Those characters are available in my nerd font and are valid unicode characters. After updating Oh My Zsh to commit [898b579](https://github.com/ohmyzsh/ohmyzsh/commit/898b579cf51026e95f0249f57e36387227e7a184), I'm getting the following error and my git status characters are not showing:
```
_omz_git_prompt_status:105: character not in range
```

### Steps to reproduce

Create the following .zshrc:
```sh
ZSH_THEME="my-custom-theme"
zstyle ':omz:alpha:lib:git' async-prompt no
source $HOME/.oh-my-zsh/oh-my-zsh.sh
```

Create a file for the custom theme:
```sh
touch $HOME/.oh-my-zsh/custom/themes/my-custom-theme.zsh-theme
```

Add the following content to the custom theme file:
```sh
ZSH_THEME_GIT_PROMPT_UNTRACKED="\U2731"
PROMPT='$(git_current_branch) $(git_prompt_status)'$'\n'
```

### Expected behavior

I expect my prompt to show the current branch and the [`U+2731` (Heavy Asterisk)](https://www.compart.com/en/unicode/U+2731) character if the current git repository has untracked files.

_No response_

### OS / Linux distribution

Ubuntu 24.04.4 LTS

### Zsh version

5.9

### Oh My Zsh version

master (97e1105)

### Terminal emulator

Ghostty

None

### Additional context

_No response_

Morningstar202604:
## Standards checklist:

## Changes:

Fixes #13985, without undoing what #13878 was trying to do.

The short version: `898b579` set `LC_ALL=C` for the whole `_omz_git_prompt_status` function to stop the regex from choking on invalid byte sequences (#13330). That works, but it's broader than intended — zsh re-runs `setlocale()` as soon as `LC_ALL` is assigned, so the rest of the function also runs under the C locale. When a theme uses a Unicode escape like `\U2731` for its prompt glyphs, the final `echo` can't map it to a character anymore and dies with `character not in range`, which is exactly the error reported in #13985.

So this moves the locale forcing into a tiny helper (`_omz_git_prompt_status_match`) that wraps each individual `[[ =~ ]]` check. The three call sites that parse porcelain output still get the C locale (that part of #13878 stays intact), while everything else — most importantly printing the theme symbols — keeps the user's own locale. `$match` etc. are global special parameters, so nothing changes about how matches are consumed afterwards.

For testing I ran this on an Ubuntu 24.04 container with zsh 5.9:

- the reporter's exact scenario now renders the glyph instead of erroring
- a repo with a non-UTF-8 filename no longer triggers "illegal byte sequence" (#13330 case)
- ahead / behind / diverged tracking-line parsing gives identical output to unmodified master
- full `oh-my-zsh.sh` startup smoke test passes; `zsh -n` is clean

## Other comments:

AI disclosure: this patch was prepared with the help of an AI coding agent (root-cause analysis and draft). I reproduced the bug locally, reviewed the change line by line, and ran the tests listed above myself before submitting.

### DIFF
--- a/lib/git.zsh
+++ b/lib/git.zsh
@@ -39,16 +39,27 @@ function _omz_git_prompt_info() {
   echo "${ZSH_THEME_GIT_PROMPT_PREFIX}${ref//\%/%%}${upstream//\%/%%}$(parse_git_dirty)${ZSH_THEME_GIT_PROMPT_SUFFIX}"
 }
 
-function _omz_git_prompt_status() {
-  # OHMYZSH-13330: avoid "regex matching error: illegal byte sequence".
-  # zsh's "=~" operator delegates to the C library regex, which aborts with
-  # REG_ILLSEQ when the subject contains an invalid byte sequence under a
-  # multibyte locale (e.g. a filename with non-UTF-8 bytes in `git status`
-  # output). Forcing the C locale makes every byte a valid character, so the
-  # regex matching below never fails that way. `git status --porcelain` is
-  # locale-independent, so this does not change the parsed output.
+# Match an extended regular expression against a subject, forcing the C locale
+# only for the duration of the call.
+#
+# OHMYZSH-13330: zsh's "=~" operator delegates to the C library regex, which
+# aborts with REG_ILLSEQ when the subject contains an invalid byte sequence
+# under a multibyte locale (e.g. a filename with non-UTF-8 bytes in `git
+# status` output). Forcing the C locale makes every byte a valid character,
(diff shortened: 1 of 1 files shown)


carlosala (approved):
LGTM!

carlosala:
Thanks for your contribution!

PR #14019, opened 2026-08-25
feat(core): add `omz diagnose` locale check

### LINKED ISSUE #10658: Autocompletion is broken in Oh My Zsh out of the box (clean env)
### Describe the bug

This is a bug for which I have found a lot of reports for a long time, but nothing has helped me solve it so far.

Using out-of-the-box Oh My Zsh, autocompletion glitches and duplicates characters at the prompt.

### Steps to reproduce

Here's a Dockerfile that will allow you to reproduce it in an environment that is clean of any past configuration:

```Dockerfile
FROM ubuntu:latest
RUN apt-get update && apt-get install -y curl git zsh
RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
ENTRYPOINT zsh
```

To build and run:

```sh
docker build . --progress=plain -t omz
docker run -it --rm omz
```

Once you're in, type `ls` then press Tab. If the bug happened you should end up with `lsls` displayed on your prompt., but you can only delete the second one.

Note that although I am testing in the Ubuntu Docker image, my host OS is macOS and I use the Terminal application.

### Expected behavior

The glitch should not happen (in the example above, `ls` should not be displayed twice).

https://user-images.githubusercontent.com/816901/152645367-5b23597d-4110-4dbe-b9f8-fee809e01aec.mov

### OS / Linux distribution

Ubuntu? macOS?

### Zsh version

5.8

### Oh My Zsh version

master (2d3bae96)

### Terminal emulator

Terminal.app

_No response_

### Additional context

_No response_

Morningstar202604:
## Standards checklist:

## Changes:

Fixes #10658.

Oh My Zsh already documents that non-UTF-8 locales cause glitches such as duplicated characters at the prompt (see the wiki FAQ), but nothing tells the user that their current environment is affected. As discussed in the issue, this adds a small startup check:

- After the emulation-mode check in `oh-my-zsh.sh`, run `locale charmap` (skipped entirely if the `locale` binary is missing).
- If the effective charmap is not `UTF-8`, print a short yellow warning pointing to the wiki FAQ, suggesting e.g. `export LANG=C.UTF-8`.
- Users can silence it with `DISABLE_LOCALE_WARNING=true`.

Design notes:

- The check catches all the flavors of this problem: unset `LANG` (POSIX fallback), locale names that aren't generated on the machine, and `LC_ALL=C`-style overrides — which is the same class of interaction I fixed for the git prompt in #14006.
- It reuses the `omz_f` ANSI helper before `unset -f omz_f`, so formatting is automatically suppressed when stdout is not a terminal.
- Wording is intentionally soft ("may cause glitches") so platforms where the default charset is already UTF-8 despite an unusual `$LANG` don't get scary messages.
- Cost is one `locale charmap` fork per interactive start. I measured it rather than assuming: 3.9 ms per `locale` invocation on this box, and `source oh-my-zsh.sh` (zsh 5.9, plugins=(git), 30 runs x 3 reps) goes 120.4 / 122.0 / 122.1 ms unparsed vs 125.2 / 124.6 / 125.6 ms with the check, i.e. a **+2.6 to +4.9 ms** delta. That is a real cost, and I'd previously stated ~1 ms here, which was wrong.
- The fork is not laziness, it is the only way to catch the case the feature exists for. `locale charmap` reports the *effective* charmap after glibc's silent fallback, whereas parsing `$LANG` only sees what the user *asked* for. Verified on a minimal container that ships just `C`, `C.utf8`, `POSIX`: `LANG=en_US.UTF-8 locale charmap` -> `ANSI_X3.4-1968`, and the same for `zh_CN.GBK` and `no_NO.ISO-8859-1`. A pure-string `$LANG` fast path would therefore print "UTF-8" on these machines and stay silent in exactly the broken setups from #10658. Happy to drop the check or gate it behind an opt-in style if the startup cost is judged not worth it.

## Testing:

Tested locally with zsh 5.9.2:

- `LANG=C` → `Warning: your current locale is not UTF-8 (charmap: ANSI_X3.4-1968)` shown (re-verified here on zsh 5.9)
- `LANG=no_NO.ISO-8859-1` → warning shown with `charmap: ISO-8859-1`
- `LANG=en_US.UTF-8 LC_ALL=C` → warning shown (the override wins)
- `LANG=C.UTF-8` / `LANG=en_US.UTF-8` → no output change
- `DISABLE_LOCALE_WARNING=true` with a bad locale → silent
- Both warning paths re-run against a real `${ZSH}/oh-my-zsh.sh` source (not just `zsh -n`): unset `LANG`, `LANG=C.UTF-8` (silent), `LANG=en_US.UTF-8` on an ungenerated locale (warns with the effective `ANSI_X3.4-1968`), and the disable flag. The `[[ "$FLAG" != true ]]` test matches the existing house convention in `oh-my-zsh.sh:124` (`ZSH_DISABLE_COMPFIX`) and `lib/termsupport.zsh:50` (`DISABLE_AUTO_TITLE`).
- `zsh -n oh-my-zsh.sh` passes

## Other comments:

AI disclosure: this patch was prepared with the help of an AI coding agent (root-cause research and draft). I reviewed every line, ran the test matrix above myself, and confirmed the behavior against the FAQ scenarios before submitting.

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -24,6 +24,7 @@ function _omz {
   local -a cmds subcmds
   cmds=(
     'changelog:Print the changelog'
+    'diagnose:Run configuration diagnostics'
     'help:Usage information'
     'plugin:Manage plugins'
     'pr:Manage Oh My Zsh Pull Requests'
@@ -163,12 +164,38 @@ function _omz::log {
 
 ## User-facing commands
 
+function _omz::diagnose {
+  # Check the runtime locale: non-UTF-8 charsets are known to cause glitches
+  # such as duplicated characters at the prompt after completions.
+  # https://github.com/ohmyzsh/ohmyzsh/wiki/FAQ#i-see-duplicate-typed-characters-after-i-complete-a-command
+  #
+  # The zsh/langinfo module is free of subprocess overhead; fall back to
+  # `locale charmap` when it is not available.
+  if zmodload -F zsh/langinfo b:langinfo 2>/dev/null; then
+    local codeset="$langinfo[CODESET]"
+  elif (( $+commands[locale] )); then
+    local codeset="$(locale charmap 2>/dev/null)"
+  else
+    local codeset="unknown"
+  fi
+
+  if [[ "$codeset" = "UTF-8" ]]; then
+    _omz::log info "locale charmap is UTF-8"
+  else
+    _omz::log warn "your locale is not UTF-8 (charmap: ${codeset:-unknown})"
+    _omz::log info "non-UTF-8 locales may cause glitches in completions and international characters"
+    _omz::log info "consider setting a UTF-8 locale, e.g.: export LANG=C.UTF-8"
+    return 1
+  fi
+}
+
 function _omz::help {
   cat >&2 <<EOF
 Usage: omz <command> [options]
 
 Available commands:
 
+  diagnose            Run configuration diagnostics
   help                Print this help message
   changelog           Print the changelog
   plugin <command>    Manage plugins


Morningstar202604:
Quick summary: This adds a startup warning when \locale charmap\ doesn't resolve to UTF-8, pointing users to the wiki FAQ. Helps prevent the duplicate-character issue that many users hit with non-UTF-8 locales. CI is green, ready for review.

robbyrussell:
Thanks for this, and sorry you didn't get an answer when you asked which direction to go... that one's on us.

Before anyone looks at the code, this is really a product call rather than a code one. The warning would fire for every macOS user with `LANG` unset, every SSH session without locale forwarding, and every minimal container... on every single shell. So the question is whether we want to nag that many people about a setup a lot of them chose on purpose.

@mcornella @carlosala, worth a decision here. Carlo, you mentioned in #10658 back in 2022 that some kind of hint might be worth doing. The way I see the options: take it roughly as-is, move it into `omz diagnose` where people go when something looks wrong, or make it a one-time notice instead of every startup.

Leaving this open until we settle that. Nothing needed from you in the meantime, and thanks for the writeup on what you tested.

mcornella:
I think it's best to put this in `omz diagnose` and use the langinfo module which is zero cost: https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html#The-zsh_002flanginfo-Module (I think CODESET is the right variable we need here).

But it does make sense to flag this on some theme that explicitly needs UTF8, when it's loaded for the first time. I think though that we'll need to invest in building some kind of standardized notification system that keeps track of seen / read notifications.

Morningstar202604:
Thanks @robbyrussell and @mcornella! I agree with the direction: `omz diagnose` is the right home for this — zero startup cost, and it surfaces exactly when someone is already debugging a broken setup. Using the `zsh/langinfo` module's `CODESET` also avoids the `locale charmap` subprocess entirely.

Happy to rework the PR that way once you've settled the product call — just say the word.

Morningstar202604:
Thanks both! Happy to leave the product call to you. If the decision lands on `omz diagnose`, I can rework this to check the locale there using the `zsh/langinfo` module (`$langinfo[CODESET]`) as @mcornella suggests — zero-cost and no startup nag. A one-time-notice variant would also be straightforward if that wins. Just let me know which direction to take and I will update the PR.

Morningstar202604:
Thanks both — happy to leave the product call to you. If the decision lands on `omz diagnose`, I can rework this to check the locale there using the `zsh/langinfo` module (`$langinfo[CODESET]`) as @mcornella suggests — zero-cost and no startup nag. A one-time-notice variant would also be straightforward if that wins. Just let me know the direction and I will update the PR.

Morningstar202604:
Reworked per the discussion — the check now lives in a new `omz diagnose` command instead of a startup warning. It uses `zsh/langinfo` CODESET (no subprocess at startup; `locale charmap` only as fallback when the module is unavailable) and exits non-zero when the locale is not UTF-8. The startup block in `oh-my-zsh.sh` is gone.

robbyrussell:
Thanks for reworking this... the opt-in `omz diagnose` shape is a lot better than the startup check, and it's what we were hoping for.

Before we get into the details, there's a question on our side I want to settle first: we already ship `omz_diagnostic_dump` in `lib/diagnostics.zsh`, and I'd like to check with @mcornella whether `omz diagnose` should be the front door for that rather than a locale-only command. That shapes where this function ends up living.

Sitting on it until we've talked it through... nothing needed from you in the meantime.

mcornella:
`omz diagnose` should be the front door where everything lives, but pulling the previous `omz_diagnostic_dump` should be a separate PR. Plus, that older function creates a dump file, this should be a separate option in the omz diagnose command (like -something).

Also this file is getting huge, if possible I'd like to see some system of autoloadable functions that get pulled in, and also a dynamic way of generating the help message and the completion entries for omz based on them. If we separate the functions into different files, we can also include unit testing more easily.

But for now this PR is fine as is.

PS. We should not forget documenting this, which currently is the README plus ohmyzsh/wiki

Morningstar202604:
Thanks @mcornella — that all makes sense: `omz diagnose` as the front door, with the dump-file behavior as a separate option and a separate PR; autoloadable per-function files plus generated help/completion would be a nice follow-up once this lands. This PR stays as-is (locale check only); I will document it in the README/wiki bit you mentioned.

PR #14021, opened 2026-08-25
fix(pipenv): keep activation and the project directory in sync

### LINKED ISSUE #10600: pipenv plugin error when using directory containing Pipfile
If you include the pipenv plugin in your zshrc and open a new terminal directly into a directory which contains a Pipfile, the plugin produces an error:-

> -- console output produced during zsh initialization follows --

> Launching subshell in virtual environment...
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/cli/command.py", line 393, in shell
    do_shell(
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/core.py", line 2340, in do_shell
    shell.fork_compat(*fork_args)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/shells.py", line 121, in fork_compat
    c.interact(escape_character=None)
  File "/opt/homebrew/Cellar/pipenv/2022.1.8/libexec/lib/python3.10/site-packages/pipenv/vendor/pexpect/pty_spawn.py", line 788, in interact
    mode = tty.tcgetattr(self.STDIN_FILENO)
termios.error: (19, 'Operation not supported by device')

This issue was first mentioned in #6679 .

### Steps to reproduce

1. Create a new directory
2. cd into it
3. pipenv install requests (or anything)
4. Set your (iTerm) profile to open new windows in that directory. (I am using Visual Studio code which does this automatically, but the same thing happens)
5. Open a new terminal window.

### Expected behavior

It should not show any errors and just set the relevant pipenv environment.

Screenshot 2022-01-19 at 12 36 36

### OS / Linux distribution

macOS 12.1

### Zsh version

5.8

### Oh My Zsh version

master (957dca6)

### Terminal emulator

iTerm2, VSCode

_No response_

### Additional context

_No response_

Morningstar202604:
## Standards checklist:

## Changes:

Fixes #10600, and fixes a second bug in the same hook that I found while testing it.

`_togglePipenvShell` gets its two halves wrong in ways that both leave the shell claiming a state it does not have. Two commits, one per bug.

**Commit 1 — activation is unguarded (#10600).** When a `Pipfile` exists but its virtualenv does not, `source "$(pipenv --venv)/bin/activate"` runs with no error handling, so:

- `pipenv --venv`'s complaint (in some versions a full Python traceback) goes straight to the shell's startup output;
- `$(pipenv --venv)` expands to empty and zsh tries to `source /bin/activate`, printing `no such file or directory`;
- `PIPENV_ACTIVE=1` and `pipfile_dir` are exported anyway although nothing was activated, which blocks every later activation attempt in that shell;
- and because `PIPENV_ACTIVE=1` was set anyway, the *next* `cd` out of the directory takes the deactivation branch and calls a `deactivate` function that no activate script ever defined.

Activation now happens only when `pipenv --venv` succeeds **and** the activate script exists. This also removes the follow-on `command not found: deactivate`, since that state is never entered.

**Commit 2 — the directory test uses string prefixes, not path components.** The deactivation guard is `[[ "$PWD" != "$pipfile_dir"* ]]`, which is true for subdirectories of the project (intended) but also for unrelated siblings that merely start with the same characters: `…/proj` → `…/proj-archive` or `…/projX` keeps the virtualenv active in a directory that has no `Pipfile` at all, and `PIPENV_ACTIVE` stays `1`. Fixed by requiring either an exact match or a `/` component boundary. Subdirectory behaviour is unchanged, and a project at `/` still matches everything.

Same class of slip as the one fixed for the git prompt in #14006. `poetry-env` has the identical prefix test on line 9 of its plugin (`"$PWD" != "$poetry_dir"*`), so it has the same bug — I left that file alone because #13932 is already open against it, but noted it there.

## Testing:

Stubbed `pipenv` driving each toolchain state, capturing stderr per step, run against this branch and `origin/master`; every cell reproduced on three consecutive runs. zsh 5.9 (the CI job is a single `ubuntu-latest` + apt zsh with no version matrix, so I am not claiming coverage beyond that one version).

| case | master | this branch |
|---|---|---|
| `Pipfile` present, venv missing (startup) | `pipenv` stderr leaks + `source:…: no such file or directory: /bin/activate` + `PIPENV_ACTIVE=1` while `VIRTUAL_ENV` unset | silent; `PIPENV_ACTIVE`/`pipfile_dir` stay unset |
| …then `cd` out of that directory | `command not found: deactivate` | silent |
| `--venv` prints a path with no `bin/activate` | same failure class | silent, stays inactive |
| healthy venv, start in project | `PIPENV_ACTIVE=1`, `VIRTUAL_ENV` set, `deactivate` defined | identical |
| healthy: project → elsewhere → back | activates → inactive → re-activates, stderr empty | identical |
| **project → unrelated `proj-archive` sibling** | ⚠️ stays active: `VIRTUAL_ENV=<proj venv> PIPENV_ACTIVE=1` | ✅ `VIRTUAL_ENV=<none> PIPENV_ACTIVE=<unset>` |
| **project → `…/proj-archive/empty`** | ⚠️ still active | ✅ still inactive |
| **deep subdir *inside* the project** (no `Pipfile` there) | stays active | stays active (unchanged) |

The boundary logic as a truth table (`stay` = keep virtualenv):

| `pipfile_dir` | `PWD` | master | this branch |
|---|---|---|---|
| `/x/proj` | `/x/proj` | stay | stay |
| `/x/proj` | `/x/proj/src` | stay | stay |
| `/x/proj` | `/x/proj-archive` | ⚠️ stay | ✅ deactivate |
| `/x/proj` | `/x/projX` | ⚠️ stay | ✅ deactivate |
| `/` | `/tmp` | stay | stay |

`zsh -n` passes on the plugin and the CI syntax sweep (`oh-my-zsh.sh`, `lib/*.zsh`, `plugins/*/*.plugin.zsh`, `plugins/*/_*`, `themes/*.zsh-theme`) is clean on this branch.

## Other comments:

If you would rather have these as two PRs, say the word and I will split commit 2 out onto its own branch — they are independent and each reverts cleanly on its own. They share a file and a function, which is why they are together here.

The two new `[[ … ]]` operands and the `local project_dir` add no measurable startup cost: the guard is a shell builtin pattern match on variables that are already set, and the previously measured `pipenv --venv` call still happens exactly once per activation attempt as before.

AI disclosure: this patch was prepared with the help of an AI coding agent (root-cause analysis, patch draft, and test harness). I reviewed every changed line and re-ran the tables above against both branches myself before submitting.

### DIFF
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -56,7 +56,11 @@ if zstyle -T ':omz:plugins:pipenv' auto-shell; then
     # deactivate shell if Pipfile doesn't exist and not in a subdir
     if [[ ! -f "$PWD/Pipfile" ]]; then
       if [[ "$PIPENV_ACTIVE" == 1 ]]; then
-        if [[ "$PWD" != "$pipfile_dir"* ]]; then
+        # Compare whole path components: `$pipfile_dir"*` also matches unrelated
+        # siblings that merely start with the same text (…/proj -> …/proj-docs),
+        # which would keep the virtualenv active outside the project.
+        local project_dir="${pipfile_dir%/}"
+        if [[ "$PWD" != "$project_dir" && "$PWD" != "$project_dir"/* ]]; then
           unset PIPENV_ACTIVE pipfile_dir
           deactivate
         fi
--- a/plugins/pipenv/pipenv.plugin.zsh
+++ b/plugins/pipenv/pipenv.plugin.zsh
@@ -63,12 +63,15 @@ if zstyle -T ':omz:plugins:pipenv' auto-shell; then
       fi
     fi
 
-    # activate the shell if Pipfile exists
+    # activate the shell if Pipfile exists and its virtualenv is usable
     if [[ "$PIPENV_ACTIVE" != 1 ]]; then
       if [[ -f "$PWD/Pipfile" ]]; then
-        export pipfile_dir="$PWD"
-        source "$(pipenv --venv)/bin/activate"
-        export PIPENV_ACTIVE=1
+        local venv_path
+        if venv_path="$(pipenv --venv 2>/dev/null)" && [[ -n "$venv_path" && -f "$venv_path/bin/activate" ]]; then
+          export pipfile_dir="$PWD"
+          source "$venv_path/bin/activate"
+          export PIPENV_ACTIVE=1
+        fi
       fi
     fi
   }


Morningstar202604:
Quick summary: Keeps pipenv activation and project directory in sync. The auto-shell hook now checks if the virtualenv exists before activating, preventing errors on fresh directories with Pipfile but no venv yet. Related to issue #10600. Tests pass, ready for review.

robbyrussell (approved):
Thanks — this is a good catch, and two of them. The prefix-match bug in the deactivate guard is the same slip we fixed in #14006, and the unguarded activation explains the `command not found: deactivate` reports in #10600.

I checked the boundary cases myself (`proj-archive`, `projX`, deep subdirs, a project at `/`) and the exit-status capture — declaring `local venv_path` on its own line is the detail that makes that guard work, so thanks for getting it right.

One thing, not blocking: `pipenv --venv 2>/dev/null` means a genuinely broken virtualenv now fails silently. Happy to merge as-is; if you want to add a one-line hint on first failure, I'll take that as a follow-up.

The test matrix was helpful. Merging.

PR #14024, opened 2026-08-26
fix(eza): pass an explicit WHEN value to --hyperlink

Pro100v:
Since eza 0.23 `--hyperlink` is an option with an optional WHEN value
instead of a boolean flag, so a bare `--hyperlink` at the end of the
alias consumes the following argument:

    $ ll ./tmp
    error: invalid value './tmp' for '--hyperlink [<WHEN>]'
      [possible values: always, auto, never]

Pass `--hyperlink=auto` (mirroring how the plugin already handles
`--icons=auto`) when eza accepts it, and fall back to the bare flag on
older releases.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01FD5F6nyEXwHsqtH7eT3HeZ

### DIFF
--- a/plugins/eza/eza.plugin.zsh
+++ b/plugins/eza/eza.plugin.zsh
@@ -53,7 +53,13 @@ function _configure_eza() {
     _EZA_TAIL+=("--time-style='$_val'")
   fi
   if zstyle -t ":omz:plugins:eza" "hyperlink"; then
-    _EZA_TAIL+=("--hyperlink")
+    # eza >= 0.23 turns --hyperlink into an option with an optional WHEN value,
+    # so a bare --hyperlink swallows the next argument (e.g. `ll ./tmp`).
+    if eza --hyperlink=auto --version &>/dev/null; then
+      _EZA_TAIL+=("--hyperlink=auto")
+    else
+      _EZA_TAIL+=("--hyperlink")
+    fi
   fi
 }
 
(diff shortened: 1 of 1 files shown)


robbyrussell (commented):
Thanks for tracking this down — the repro in the description made it easy to verify. Works as advertised on both old and new eza.

Keeping the version fallback is the right call. Ubuntu 24.04 LTS still ships eza 0.18.2 and Debian stable has 0.21.0, so bare `--hyperlink` will be around for a while yet.

One small thing — the comment is heavier than the rest of this file:

```zsh
  if zstyle -t ':omz:plugins:eza' 'hyperlink'; then
    # eza >= 0.23 needs an explicit WHEN value
    if command eza --hyperlink=auto --version &>/dev/null; then
      _EZA_TAIL+=("--hyperlink=auto")
    else
      _EZA_TAIL+=("--hyperlink")
    fi
  fi
```

Also switched to single quotes to match the rest of the function, and `command eza` so a user's own eza wrapper can't confuse the check.

robbyrussell:
Went ahead and pushed the tidy-up myself so this isn't waiting on a round trip — comment shortened, single quotes, and `command eza` on the probe. Behavior is unchanged; verified the alias still resolves correctly against both a 0.23.x and a pre-0.23 eza.

robbyrussell:
Passing to @mcornella for review/approval

mcornella (approved):

Issue #13949, opened 2026-08-14
aws plugin - asr() error message leaves terminal red and outputs to stdout instead of stderr


varshini03:
### Summary

hi, I noticed an edge case -
When asr is called with an invalid region, the error message is missing ${reset_color}, which leaves the terminal permanently red for all subsequent output. The same error goes to stdout instead of stderr, which is inconsistent with how the sibling function asp() handles the same case.

### Actual behavior

Terminal stays red post the error, error message goes to stdout

Happy to contribute a quick fix once confirmed by a maintainer!

### Steps to reproduce

1. Enable the aws plugin
2. Run asr with an INVALID_REGION
3. Observe the prompt and any text typed after

### Expected behavior

Error message goes to stderr
Terminal color resets after the error message

### OS / Linux distribution

macOS 26.5.2 (Tahoe)

### Zsh version

5.9

### Oh My Zsh version

b54a7197 (2026-08-11)

### Terminal emulator

Terminal.app

### If using WSL on Windows, which version of WSL

None

varshini03:
just a doubt, won't I be allowed to make changes and raise pr? because I had clearly mentioned I wanted to work on this
I would like some maintainer to pay heed to this since a PR has been raised despite me mentioning I would like to raise pr once confirmed.
or am I just expected to raise a PR without confirmation? @mcornella just tagging a maintainer to get this confimed, thank you

robbyrussell:
> won't I be allowed to make changes and raise pr? because I had clearly mentioned I wanted to work on this

@varshini03 We appreciate your desire to go ahead and work on this yourself. That said, we don't have a way to prevent other people from doing that before you and/or we chime in. 

We're also navigating folks having automated PR contributions being added. We're going to go ahead and approve the fix since it's okay, but we encourage you to go ahead and try to resolve an issue like this yourself and submit a PR, too

PR #13952, opened 2026-08-15
fix(aws): asr() prints error to stderr and resets terminal color

deepshekhardas:
Fixes #13949

PR #14027, opened 2026-08-29
fix(aws): send asr() error message to stderr and reset terminal color

Morningstar202604:
## Description

`asr()` currently prints its "Available regions" error to stdout without resetting the color, so after a failed `asr <invalid-region>` the terminal stays red and the diagnostics end up in piped output. The sibling helper `asp()` already does both things correctly (`>&2` and `${reset_color}`), so this just mirrors it.

Refs #13949 (the earlier PR on that ticket, #14022, was withdrawn by its author)

## How was this tested?

- Diff is a single line matching the existing `asp()` idiom in the same file
- oh-my-zsh CI runs the plugin checks on this repo

### DIFF
--- a/plugins/aws/aws.plugin.zsh
+++ b/plugins/aws/aws.plugin.zsh
@@ -69,7 +69,7 @@ function asr() {
   local -a available_regions
   available_regions=($(aws_regions))
   if [[ -z "${available_regions[(r)$1]}" ]]; then
-    echo "${fg[red]}Available regions: \n$(aws_regions)"
+    echo "${fg[red]}Available regions: \n$(aws_regions)${reset_color}" >&2
     return 1
   fi
 


robbyrussell (approved):
Thanks — this is correct, and mirroring `asp()` is the right call. I confirmed the behavior both ways: before, the red escape goes to stdout with no reset; after, the message is on stderr and ends with `${reset_color}`.

One optional thing while you're on this line: `available_regions` (line 70) already holds the list, so the message could use `${(F)available_regions}` instead of calling `$(aws_regions)` again. Since `aws_regions` makes an `ec2 describe-regions` API call, the current code costs two network round-trips on the error path. I checked and `${(F)…}` gives byte-identical output, so nothing visible changes. Happy to merge as-is if you'd rather keep the diff to one line.

Small thing: I couldn't find a #14022 in this repo — I think the earlier attempts at this were #14007, #13966 and #13952.

Issue #14025, opened 2026-08-27
git: add gbls alias to list local branch names


XxUnkn0wnxX:
### If the feature request is for a plugin or theme, specify it here.

git plugin

### If the feature solves a problem you have, specify it here.

`gb` runs `git branch`, which is useful for the normal branch listing, but sometimes I only want a quick, plain list containing one local branch name per line.

A small `gbls` alias would provide a convenient way to get that clean list without the normal `git branch` presentation.

### Describe the proposed feature.

Add a `gbls` alias to the Git plugin for quickly listing **local branch names only**.

For example:

```zsh
alias gbls='git for-each-ref --format="%(refname:short)" refs/heads/'
```

This would produce simple output such as:

```text
develop
feature/test
main
```

The alias is intentionally minimal: no dates, authors, remote branches, sorting by recent activity, or other metadata.

It would complement the existing `gb` alias rather than replace it:

- `gb` → normal `git branch` listing
- `gbls` → plain list of local branch names only

This is useful when you just want to quickly scan, copy, or pipe the local branch names without needing the normal branch listing.

### Describe alternatives you've considered

A user can define the alias manually in `.zshrc`, or run something such as:

```zsh
gb --format='%(refname:short)'
```

However, `gbls` would make this common minimal listing available as a short, consistent Git-plugin alias without having to remember or type the format each time.

### Related Issues

Related, but not duplicates:

- #8790 proposes `gbll` / `gblo` for listing branches **sorted by commit date**.
- #13402 proposes `glb` for listing the **latest X branches**.
- #11132 is about showing **local branches only in Git tab completion**.

None of these cover the same minimal use case as this request.

`gbls` would simply list all local branch names using the equivalent of:

`git for-each-ref --format='%(refname:short)' refs/heads/`

The intention is to provide a quick "just list my local branches" command alongside `gb`, while keeping `gb` unchanged for the normal `git branch` listing.

robbyrussell:
Thanks for the suggestion! Going to pass on this one.

`gb` already accepts the same flag, so `gb --format='%(refname:short)'` gets you the plain one-per-line list without a new alias. We try to keep to generic aliases rather than adding a specific one for each flag combination.

If you want the shortcut on your own machine, dropping it in `$ZSH_CUSTOM` works well and survives updates.

PR #14028, opened 2026-08-29
feat(git): add gbls alias to list local branch names

Morningstar202604:
## Description

Adds a `gbls` alias to the git plugin for a plain list of local branch names, one per line:

```console
% gbls
main
fix/aws-asr-stderr-13949
```

The alias, command and README entry follow the ticket's proposal; `gb` keeps its current behavior for the decorated view.

Refs #14025


robbyrussell:
Thanks for putting this together, and for linking it back to the issue.

Going to pass on this one though. `gb` already takes the same flag... `gb --format='%(refname:short)'` gives you the exact same list, so this ends up being a shortcut for a flag on a shortcut. We try to keep to generic aliases rather than adding a specific one for each flag combination, and that's what this runs into.

If you want it on your own machine, an alias in `$ZSH_CUSTOM` will survive updates just fine.

Thanks again!

Issue #10412, opened 2021-11-12
completions provided by custom plugins does not get loaded


felipecrs:
### Describe the bug

I have my own completions set at `$ZSH_CUSTOM/plugins/my-completions/src` and a `my-completions.plugin.zsh` to add `src` to fpath.

However, while the completions for the built-in plugins of OMZ seems to be working fine, my completions at my custom simply does not get loaded.

I have to manually run `compinit` (or add it after the `source .oh-my-zsh.zsh`, but it brings some consequences like not leveraging cache).

### Steps to reproduce

Here is my .zshrc:

https://github.com/felipecrs/dotfiles/blob/master/home/dot_zshrc

and here is my-completions:

https://github.com/felipecrs/dotfiles/tree/master/home/dot_oh-my-zsh/custom/plugins/exact_my-completions

both deno and volta does not work.

### Expected behavior

It's expected to simply work, without me having to re-run compinit.

### Screenshots and recordings

https://user-images.githubusercontent.com/29582865/141393793-2082836e-8777-44fc-a0ca-34aecb23ab58.mp4

### OS / Distro

Ubuntu 20.04

### Zsh version

5.8

### Terminal emulator

Windows Terminal

### If using WSL on Windows, which version of WSL

WSL2

### Additional context

In the [zsh-completions install guide for OMZ](https://github.com/zsh-users/zsh-completions#oh-my-zsh), they suggest adding `autoload -U compinit && compinit` but I guess that was not supposed to be needed.

mcornella:
The compinit happens before the plugin is loaded, so the `$fpath` modification happens afterwards. You'd need to manually call compinit, or hack around it by autoloading the files and running `compdef` after that.

The easier solution though is to move them outside the src folder.

felipecrs:
Shouldn't we switch this order? I mean, can't we make compinit be called only after all plugins are loaded?

mcornella:
This was a possibility, but a lot of plugins heavily use `compdef` which is requires compinit to be loaded.

felipecrs:
I could move my completions out of the `src` folder, but that does not solve the problem for zsh-completions. Should some alternative subfolders be supported, like `src` as zsh-completions, since it's a noticeably big project?

felipecrs:
Or, perhaps, a subfolder called `completions`, and then we can propose zsh-completions to setup a symlink in their repo from `completions` to `src`.

mcornella:
[...]

PR #14032, opened 2026-08-30
fix(cli): keep completion flag when loading multiple plugins

mxtymoshyk:
Inside the plugin loop, `has_completion` is assigned rather than accumulated:

```zsh
has_completion=$(( $#comp_files > 0 ))
```

Each iteration overwrites what the previous one concluded, so when the loop ends the flag reflects only the last plugin named on the command line.

`omz plugin load z docker` is enough to show it. `plugins/z/_z` sets the flag to 1 on the first iteration, then `plugins/docker/` — which has no top-level `_*` file, since its completions live in a `completions/` subdirectory — resets it to 0. `compinit` never runs, `_z` ends up in `$fpath` unregistered, and pressing Tab after `z ` does nothing with no error to explain why. Reversing the argument order makes it work, which is the tell that ordering matters when it shouldn't.

`(( has_completion )) || ...` short-circuits once the flag is set, so a later plugin without completions can no longer clear it. I kept it strictly 0/1 rather than summing, since it's consumed as a boolean by the `(( has_completion ))` test further down. Loading a single plugin is unaffected — with one iteration there's nothing to overwrite.

Simulating a load of one plugin with completions followed by one without:

```
before:  has_completion=0    (compinit skipped, _z never registered)
after:   has_completion=1
```

This might be related to #10412, but that one is about the startup `$fpath` path rather than `omz plugin load`, so I don't think this closes it.

## Standards checklist:

## Changes:

- Accumulate `has_completion` across the loop instead of overwriting it, so `compinit` re-runs whenever any loaded plugin ships completions.

## Other comments:

AI-assisted: I've been using Oh My Zsh for years and wanted to contribute something back, so I used Claude to help audit the tree for bugs. This was one of the things it turned up. I've read the surrounding code, confirmed the behaviour myself on zsh 5.9.2, and can explain the change.

### DIFF
--- a/lib/cli.zsh
+++ b/lib/cli.zsh
@@ -519,7 +519,7 @@ function _omz::plugin::load {
     # Check if it has completion to reload compinit
     local -a comp_files
     comp_files=($base/_*(N))
-    has_completion=$(( $#comp_files > 0 ))
+    (( has_completion )) || has_completion=$(( $#comp_files > 0 ))
 
     # Load the plugin
     if [[ -f "$base/$plugin.plugin.zsh" ]]; then


carlosala (approved):

carlosala:
Thanks!!!

PR #14035, opened 2026-08-31
fix(ssh-agent): create screen/tmux symlink atomically

mecampbellsoup:
## Standards checklist:

## Changes:

With `zstyle :omz:plugins:ssh-agent agent-forwarding yes`, opening a tmux window
that starts several panes at once prints one or more of these to stderr:

```
ln: /tmp/ssh-agent-<user>-screen: File exists
```

The symlink still ends up correct, so this is cosmetic, but it is noise on every
multi-pane startup and it looks like a real failure.

**Cause.** The symlink block runs on *every* shell startup — the guard above it
(`[[ ! -L "$SSH_AUTH_SOCK" ]]`) tests the source socket, not the destination, so
the link is torn down and rebuilt each time. `ln -sf` is not an atomic swap: it
unlinks the destination, then calls `symlink()`. Two shells starting at the same
moment interleave those steps — A unlinks, B unlinks (already gone), A creates,
B's `symlink()` returns `EEXIST`.

**Fix.** Stage the link under a PID-unique name and move it into place.
`rename(2)` replaces the destination atomically, so concurrent shells can never
collide, and the last writer still wins with a correct link.

## Other comments:

Measured on macOS 15 (Darwin 25.6.0), zsh 5.9, 200 concurrent invocations of each
variant (8 at a time, 25 rounds):

| variant | `File exists` errors |
| --- | --- |
| current `ln -sf` | 82 / 200 |
| this patch | 0 / 200 |

Also reproduced end-to-end with real shell startups before the fix — 60 startups
of my own interactive zsh, 6 concurrent at a time, produced 22 errors; the same
60 run strictly sequentially produced 0, isolating concurrency as the trigger.

After the patched run the link resolves to the correct socket and no `.$$`
staging files are left behind (verified with `find`, against a positive control).

No behavior change for the single-shell case.

AI disclosure: the diagnosis, the A/B benchmark, and this description were
produced with Claude (Anthropic) assisting; I reviewed and tested the result.

### DIFF
--- a/plugins/ssh-agent/ssh-agent.plugin.zsh
+++ b/plugins/ssh-agent/ssh-agent.plugin.zsh
@@ -101,10 +101,18 @@ if zstyle -t :omz:plugins:ssh-agent agent-forwarding \
    && [[ -n "$SSH_AUTH_SOCK" ]]; then
   if [[ ! -L "$SSH_AUTH_SOCK" ]]; then
     if [[ -n "$TERMUX_VERSION" ]]; then
-      ln -sf "$SSH_AUTH_SOCK" "$PREFIX"/tmp/ssh-agent-$USERNAME-screen
+      _omz_ssh_agent_link="$PREFIX"/tmp/ssh-agent-$USERNAME-screen
     else
-      ln -sf "$SSH_AUTH_SOCK" /tmp/ssh-agent-$USERNAME-screen
+      _omz_ssh_agent_link=/tmp/ssh-agent-$USERNAME-screen
     fi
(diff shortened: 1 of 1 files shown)


carlosala (dismissed):
LGTM!

carlosala (approved):

mcornella (approved):
This approach looks like it will occasionally waste resources. Probably we want to introduce some kind of lock approach (there are zsh native lock mechanisms).

But, for now, it's good enough.

PR #14036, opened 2026-08-31
feat(juju): add unit and application name completion

iyiguncevik:
This PR extends the juju plugin's zsh completion with unit and application name autocompletion, sourced live from `juju status`. It also fixes a bug where pressing `<TAB>` for the first time in a new shell produced no completions.

Typing `juju ssh <TAB>`, `juju config <TAB>`, `juju integrate <TAB>` and many other commands will now suggest real unit and application names from the current model. The `app:endpoint` and `unit:/path` colon-suffix patterns used by `integrate` and `scp` are handled correctly — completion stops at the colon and lets the shell take over for the remainder.

The first-tab fix resolves a subtle zsh autoloading issue: with `#compdef`, the file body is the function, so a nested `_juju() {}` definition at the end was replacing but never calling the real logic on the first invocation. The fix follows the same pattern used by `_git` in the zsh distribution.

## Standards checklist:

## Changes:

- `plugins/juju/_juju`: fix first-tab completion; add unit and application name completion

## Other comments:

- I'm one of the maintainers of juju in Canonical.
- AI was used as a coding assistant during development.

### DIFF
--- a/plugins/juju/README.md
+++ b/plugins/juju/README.md
@@ -1,7 +1,7 @@
 # juju plugin
 
-This plugin provides useful aliases and functions for [juju](https://juju.is/) (for TAB completion,
-refer to the [official repo](https://github.com/juju/juju/blob/develop/etc/bash_completion.d/juju)).
+This plugin provides useful aliases and functions for [juju](https://juju.is/), including
+zsh tab completion for juju commands, flags, models, controllers, applications, and units.
 
 To use this plugin, add `juju` to the plugins array in your zshrc file.
 
--- a/plugins/juju/_juju
+++ b/plugins/juju/_juju
@@ -132,6 +132,37 @@ __juju_controllers()
         | command jq -r '.controllers | keys | .[]' 2>/dev/null
 }
 
+# Output all unit names for the current model (e.g. "mysql/0", "mysql/leader").
+# Includes subordinate units that appear nested under principal units.
+__juju_units()
+{
+    command juju status --format=json 2>/dev/null \
+        | command jq -r '
+            .applications // {} | to_entries[] |
+            .value.units // {} | to_entries[] |
+            ( [.key] + ( .value.subordinates // {} | keys ) ) | .[]
+          ' 2>/dev/null
+}
+
+# Complete a unit token.  If the token already contains a colon (e.g. the user
+# has typed "mysql/0:" for an scp path) we return immediately so zsh can fall
+# through to its built-in file/path completion for the remote path portion.
+__juju_complete_unit()
+{
+    local current="$1"
+    local -a units
+    __juju_debug "[complete_unit] current='${current}'"
+
+    # Once a colon is present the unit name is already complete; let the shell
+    # handle whatever comes after (remote path, etc.).
+    [[ "$current" == *:* ]] && return 1
+
+    units=("${(@f)$(__juju_units)}")
+    __juju_debug "[complete_unit] units=${#units}"
+    (( ${#units} )) || return 1
+    compadd -S '' -- "${units[@]}"
+}
+
 _juju()
 {
     # Initialise lookup arrays here so they are set on the very first invocation.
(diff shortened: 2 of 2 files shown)


robbyrussell (changes requested):
Thanks for this — the first-tab fix is a good catch and the diagnosis is right. I checked zsh 5.9's own `_git` and it ends with the same bare `_git` call for exactly this reason.

I found four things worth fixing before merge. Two of them make behavior worse than it is today, so those are the ones I'd start with.

**1. No fallback when `juju status` fails** — `_juju:169,183`
`units=("${(@f)$(...)}")` on empty output gives a one-element array holding an empty string, not an empty array. So `(( ${#units} ))` always passes and the `return 1` never runs. When juju can't answer (offline, not logged in, no current model), `juju config <TAB>` now offers nothing, where before it offered flags. Dropping the `@` fixes it: `units=(${(f)"$(__juju_units)"})`.

**2. `run` breaks with a global flag** — `_juju:388`
The loop starts at `i = 3`, assuming the subcommand is always `words[2]`. `juju run <TAB>` works, but `juju --debug run <TAB>` doesn't — `run` itself gets counted as a positional arg. The subcommand loop above already finds the right index and could be reused.

**3. The colon fall-through doesn't hand off** — `_juju:352-356, 405-410`
The comments say completion falls through once a colon is present, but the code does a hard `return 1`. That doesn't trigger path completion — the next completers are `_ignored`/`_correct`/`_approximate`, not `_files`. So `juju scp mysql/0:/tmp/<TAB>` and `juju integrate mysql:db <TAB>` give nothing at all now. For `scp`, `_alternative` over units plus `_files` would cover this and also fix `juju scp ./local-file<TAB>`.

**4. `remove-saas` reads the wrong key** — `_juju:400`
It completes from `.applications`, but SAAS offers live under `.application-endpoints`, so it will always suggest local application names.

**Smaller, not blocking:**
- `compadd -S ''` removes the trailing space everywhere, but the colon syntax only matters for `integrate`/`relate`/`remove-relation`/`scp`. `juju config mysql<TAB>` now needs a manually typed space.
- No caching on `juju status` — a controller round trip now firing on ~35 commands. `_store_cache`/`_retrieve_cache` keyed on the current model would help.
- `_juju_unit_flags`/`_juju_app_flags` are checked for every subcommand, not just `exec` and `operations`. `-m`/`-c` are safe to treat globally in juju; `-a`/`-u` are more generic.
- `--apps`/`--applications` take comma-separated lists, but only the first name completes.
- nit: the comment on `_juju_unit_commands` mentions `juju ssh`, but ssh/scp are handled lower down.
- optional: the README still points to the official repo for tab completion.

One thing I couldn't verify: I don't have a live controller, so #4 is based on the status JSON schema rather than a real run. Correct me if it behaves differently on your setup.

The jq filters themselves are correct — I ran both against sample status output and subordinate units come through properly.

carlosala (changes requested):
Thanks — this commit addresses most of the previous feedback, including the empty-array representation, `remove-saas`, trailing spaces, local-file completion for `scp`, and the README.

I think a few correctness issues remain before merge:

1. **The fallback when `juju status` fails still does not work.**  
   The helper now correctly returns failure for an empty result, but the application/unit command branches immediately `return 1`. They never reach the generic option completion at the bottom of `_juju`. Consequently, `juju config <TAB>` still offers nothing when status is unavailable.

2. **Single-positional commands are not position-aware.**  
   Commands in `_juju_app_commands` and `_juju_unit_commands` invoke name completion for every non-flag argument, despite being described as accepting a name only in the first position. For example, `juju config mysql <TAB>` offers application names again instead of moving on to options or subsequent arguments.

3. **Global flags with values still confuse subcommand detection.**  
   Saving `subcmd_idx` fixes `juju --debug run <TAB>`, but `juju -m model run <TAB>` identifies `model` as the subcommand. The scan needs to skip values belonging to global options such as `-m`/`--model` and `-c`/`--controller`.

Please also verify `juju scp mysql/0:/tmp/<TAB>` interactively. The current branch invokes `_files`, which handles local paths; it is not clear that it produces candidates for the remote-path form from the original review.

I'd be happy with caching, comma-separated application flags, and the remaining smaller items being handled separately, but I think the fallback and argument parsing issues above should be fixed before merge. Some completion tests for these command lines would also make this much safer to maintain.

🤖 Codex helped me with this!

PR #14059, opened 2026-09-06
feat(composer): ask composer for completions

GromNaN:
Composer is a Symfony Console application. Since composer 2.4 it answers the `_complete` protocol, so it can describe its own commands, options and arguments. This plugin now asks it, using the frontend shipped by `symfony/console`.

### What it replaces

Two partial sources were covering composer, depending on the zsh version:

- The completion in this plugin, only active on zsh older than 5.7. It parsed the output of `composer` and `composer show -s` with awk and sed, and offered nothing beyond command names and the packages already required.
- The `_composer` zsh ships since 5.7, which is why this plugin steps aside on those versions. It is a hand written description of composer's interface, and its header says it does not complete custom commands, including script aliases.

Asking composer covers both and more:

```
$ cat composer.json
{"scripts":{"my-custom-task":"echo hi"}}
$ composer my<TAB>
$ composer my-custom-task

$ composer install --no-de<TAB>
$ composer install --no-dev

$ composer require symfony/http-f<TAB>
$ composer require symfony/http-foundation
```

The commands reported are those of the installed version, so commands added by composer plugins show up too, and the list does not need updating when composer changes.

### Opting out

Users who prefer zsh's builtin composer completion, which is faster for package names thanks to its disk cache, can opt out of the native one:

```zsh
zstyle ':completion:*:composer:*' native-completion no
```

The builtin only exists since zsh 5.7. When it is not available the autoload fails, and the native completion runs as before.

### Composer 2.2 keeps what it has

Composer 2.2 is the LTS line, maintained until at least 2026-12-31, and it has no `_complete` command:

```
$ composer2.2 _complete --no-interaction -sbash -S1 -c1 -icomposer -i' '
  [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command "_complete" is not defined.
$ echo $?
1
```

An installed composer with nothing to offer exits with 0 and no output, so the exit status tells the two apart. The previous completion of this plugin is kept as `_composer_legacy` and used when the request fails. Composer 2.2 users therefore keep exactly what they have today, on any zsh version, instead of only on zsh older than 5.7.

### Why the request uses `-sbash` and `-S1`

Composer bundles `symfony/console` 5.4, which only ships the bash completion output and names the version option `--symfony` (`-S`), the option later renamed to `--api-version` (`-a`). Checked against composer 2.10.3:

```
$ composer _complete --no-interaction -szsh -a1 -c1 -icomposer -i' '
  The "-a" option does not exist.
$ composer _complete --no-interaction -szsh -S1 -c1 -icomposer -i' '
                     # empty, no zsh output in symfony/console 5.4
$ composer _complete --no-interaction -sbash -S1 -c1 -icomposer -i' '
help
list
completion
```

The bash output is a plain newline separated list, which the zsh frontend reads just as well.

### On the frontend itself

It is adapted from `src/Symfony/Component/Console/Resources/completion.zsh` of the Symfony 6.4 branch, so it already carries the upstream fixes: the request is run as an array of arguments rather than through `eval`, and it runs with `SHELL_VERBOSITY=0`. The `symfony6` plugin still has the version from before those fixes, #14058 ports them there. The two PRs are independent and touch different files.

### How it was tested

`zsh -n` on the file, then in a real shell driven through `zpty`, against composer 2.10.3: the four cases above, plus `composer global req<TAB>` and the `cr symfony/http-f<TAB>` alias from this plugin, which exercises the alias resolution path. Against composer 2.2.25 put first in `PATH`, the fallback kicks in and `composer inst<TAB>` and `composer diag<TAB>` still complete. With `native-completion no` set, the builtin takes over and `composer require symfony/http-f<TAB>` completes from the disk cache.

AI-assisted: Claude Code helped write the plugin change and the `zpty` test harness, and ran the comparisons against zsh's builtin completion and composer 2.2. I reviewed and ran everything myself.

### DIFF
--- a/plugins/composer/README.md
+++ b/plugins/composer/README.md
@@ -4,6 +4,18 @@ This plugin provides completion for [composer](https://getcomposer.org/), as wel
 for frequent composer commands. It also adds Composer's global binaries to the PATH, using
 Composer if available.
 
+Completion asks composer itself, so it reports the commands of the installed version,
+including those added by composer plugins, the scripts declared in `composer.json`, and
+the options and arguments of each command. Composer 2.2, which cannot describe itself,
+falls back to the previous completion of this plugin.
+
+To opt out of the native completion and use zsh's builtin `_composer` instead (faster for
+package names thanks to its disk cache), add this to your zshrc before `compinit`:
+
+```zsh
+zstyle ':completion:*:composer:*' native-completion no
+```
+
 To use it add `composer` to the plugins array in your zshrc file.
 
 ```zsh
--- a/plugins/composer/composer.plugin.zsh
+++ b/plugins/composer/composer.plugin.zsh
@@ -1,34 +1,138 @@
-## Basic Composer command completion
-# Since Zsh 5.7, an improved composer command completion is provided
-if ! is-at-least 5.7; then
-  _composer () {
-    local curcontext="$curcontext" state line
-    typeset -A opt_args
-    _arguments '*:: :->subcmds'
+## Composer command completion
+#
+# Composer is a Symfony Console application, so since composer 2.4 it answers
+# the "_complete" protocol and can describe its own commands, options and
+# arguments. This is the frontend for it, adapted from the script shipped by
+# symfony/console:
+# https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Console/Resources/completion.zsh
+#
+# Composer 2.2, the LTS line maintained until at least 2026-12-31, has no
+# "_complete" command. _composer_legacy below is the previous completion of
+# this plugin, kept as a fallback for it.
+# Completion for composer older than 2.4, which cannot describe itself.
+# It parses the human readable output and only knows about command names and
+# the packages already required.
+_composer_legacy() {
+  local curcontext="$curcontext" state line
+  typeset -A opt_args
+  _arguments '*:: :->subcmds'
 
-    if (( CURRENT == 1 )) || ( (( CURRENT == 2 )) && [[ "$words[1]" = "global" ]] ); then
-      # Command list
-      local -a subcmds
-      subcmds=("${(@f)"$($_comp_command1 --no-ansi 2>/dev/null | awk '
-        /Available commands/{ r=1 }
-        r == 1 && /^[ \t]*[a-z]+/{
-          gsub(/^[ \t]+/, "")
-          gsub(/  +/, ":")
-          print $0
-        }
-      ')"}")
-      _describe -t commands 'composer command' subcmds
-    else
-      # Required list
-      compadd $($_comp_command1 show -s --no-ansi 2>/dev/null \
-        | sed '1,/requires/d' \
-        | awk 'NF > 0 && !/^requires \(dev\)/{ print $1 }')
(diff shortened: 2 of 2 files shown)


robbyrussell:
Thanks Jérôme! Completing custom scripts and plugin-added commands is something neither source covers today, so the upside is real.

One thing before I say yes, and it's a call on our side rather than anything wrong with the code. On zsh 5.7+ this plugin currently steps aside for zsh's own `_composer`. This takes that over for everyone with the plugin enabled, so every TAB would spawn composer. zsh's version avoided that on purpose... its header says Packagist completion "can be very slow; it may even hang the shell," and it gives people `zstyle ':completion:*:composer:*' fetch-packages no` as a way out. There's no equivalent here.

So... when you tested against 2.10.3, how did it feel? Roughly how long is a `composer <TAB>`, and what happens to `composer require sym<TAB>` offline or on a slow connection? I don't have composer set up here to measure it myself.

If it's fast, happy to take it. If there's real lag, I'd want caching or a way back to the builtin first. I'll get you an answer either way.

GromNaN:
Thanks Robby. I measured it against composer 2.10.3 on my machine.

| Completion | Native (`composer _complete`) | zsh builtin (`_composer`) |
| --- | --- | --- |
| `composer <TAB>` (command list) | 0.45s every call | 0.03s after first (disk cache) |
| `composer require sym<TAB>` (package names) | 1.75s every call | 0.03s after first (disk cache) |

The builtin uses `_store_cache` / `_retrieve_cache`, so the first call is slow (around 3.3s, building the cache) but every subsequent call is instant. The native completion has no caching: composer is spawned fresh on every TAB.

So there is real lag for package names, and the concern is fair. The command list at 0.45s is acceptable for a one-shot spawn, but `composer require sym<TAB>` at 1.75s on every keystroke is noticeably slower than the cached builtin.

Offline does not help. The 1.75s is composer reading its local package cache, not making network calls. The Packagist fetch only happens on the first run, and composer caches it on disk itself.

I added an opt-out in the latest push. Users who prefer the builtin can set:

```zsh
zstyle ':completion:*:composer:*' native-completion no
```

This autoloads and calls zsh's `_composer` instead, so they get the disk cache back. The default stays native, because that is the only source for scripts and plugin commands, but the way back is there. On zsh older than 5.7 where the builtin does not exist, the autoload fails and native runs as before.

Is the opt-out sufficient, or would you prefer caching on the native path as well?

robbyrussell (changes requested):
Thanks for measuring... that's exactly what I was after, even if the numbers aren't what either of us was hoping for.

Two things.

The opt-out doesn't fire. `autoload +X -Uz _composer` returns 1 when `_composer` is already defined, and by that point it is — we're inside it. So the guard always fails and the native path runs no matter what the zstyle says. The branch only becomes reachable after an `unfunction _composer`.

The bigger one is your own measurement: 0.45s for the command list and 1.75s for `composer require sym<TAB>`, on every TAB, with no caching. That's too slow to be the default for everyone who has the plugin enabled. Could you turn it around... keep the current completion as the default and put the native path behind the zstyle, for people who'd rather have accuracy than speed? It's the same code with the guard the other way round, and the opt-out problem disappears with it, since the zstyle becomes the thing that switches it on.

PR #14067, opened 2026-09-06
perf(appearance): cache the diff and ls capability probes

robbyrussell:
## Standards checklist:

## Changes:

- `diff --color` and `ls -G` (or `--color`, `gls`, `colorls` depending on the OS) were probed with a fork each on every startup.
- Remember the probe results in `$ZSH_CACHE_DIR/appearance-probes` for a day, the same way `lib/grep.zsh` caches its flag probe. `test-ls-args` becomes `test-cmd-args` and is used for the `diff` probe too.
- The ls section is wrapped in an anonymous function so the early return for `DISABLE_LS_COLORS` still runs the cleanup at the end of the file. Behaviour is otherwise unchanged.

## Other comments:

Part of a series of small startup-time PRs. Measured on macOS arm64, zsh 5.9: `lib/theme-and-appearance.zsh` goes from 5.7 ms to 0.7 ms per interactive start.

Verified: first run writes the cache and sets `ls='ls -G'` plus the `diff` function; second run forks neither `diff` nor `ls` and gives the same result; a cache older than a day is re-probed; `DISABLE_LS_COLORS=true` sets no ls alias and no `LSCOLORS`, still wraps `diff`, and leaves no helper function or variable behind; the `gnu-ls` zstyle path probes `gls` and caches its absence.

### DIFF
--- a/lib/theme-and-appearance.zsh
+++ b/lib/theme-and-appearance.zsh
@@ -20,9 +20,9 @@ __omz_probe_cache="$ZSH_CACHE_DIR/appearance-probes"
 __omz_probe_cached=("$__omz_probe_cache"(Nm-1))
 [[ -z "$__omz_probe_cached" ]] || source "$__omz_probe_cache"
 
-function test-cmd-args {
-  # Usage: test-cmd-args cmd args...
-  # e.g. test-cmd-args gls --color
+function __omz_test_cmd_args {
+  # Usage: __omz_test_cmd_args cmd args...
+  # e.g. __omz_test_cmd_args gls --color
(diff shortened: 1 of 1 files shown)


robbyrussell:
Also fixed the first suppressed comment in `b0934e56`: the doc comment now spells out that `/dev/null` is appended, so the explicit `/dev/null` in the `diff` probe reads as the second operand `diff` needs rather than a mistake.

The second one (`coreutils (which might be installed for )`) is a pre-existing typo on master, not something this PR introduced, so I've left it out to keep this a perf-only change.

carlosala (approved):

PR #14076, opened 2026-09-06
fix(juju): add missing `junex` alias

cnYui:
## What

Adds the missing `junex` alias (`juju unexpose`) to the `juju` plugin.

## Why

The plugin README has documented this alias since the plugin was first added in #10075:

| Alias   | Command         | Description                                                    |
|---------|-----------------|----------------------------------------------------------------|
| `junex` | `juju unexpose` | Remove public availability over the network for an application |

However, the alias was never actually defined in `juju.plugin.zsh`, so anyone following the README and typing `junex` just gets `command not found`. This has been a docs-vs-code inconsistency since the plugin was introduced.

`junex` is the natural counterpart of the existing `jex` (`juju expose`) alias, and `juju unexpose` is a real `juju` subcommand, so the fix is simply to add the missing definition next to `jex`.

## Change

```diff
 alias jex='juju expose'
+alias junex='juju unexpose'
 alias jh='juju help'
```

One line, placed right after its `jex` counterpart, matching the exact command string already documented in the README.

## Testing

- `git diff --check` is clean.
- The added line is syntactically identical to the 60+ existing `alias` definitions in the same file, and its command string exactly matches what the README documents.

## Note

AI-assisted: the fix was found and prepared with an AI coding assistant, then reviewed manually. The change is a single one-line alias that mirrors an existing adjacent alias, and I understand exactly what it does.


carlosala (approved):

carlosala:
Thanks!

PR #14077, opened 2026-09-07
feat(git): glom alias

jonas-grgt:
glom alias allows for show commits log for the main branch only.

## Standards checklist:

## Changes:

Add glom alias for the git plugin.


plugins/git/git.plugin.zsh
```diff
@@ -239,6 +239,7 @@ alias glod='git log --graph --pretty="%Cred%h%Creset -%C(auto)%d%Creset %s %Cgre
 alias glola='git log --graph --pretty="%Cred%h%Creset -%C(auto)%d%Creset %s %Cgreen(%ar) %C(bold blue)<%an>%Creset" --all'
 alias glols='git log --graph --pretty="%Cred%h%Creset -%C(auto)%d%Creset %s %Cgreen(%ar) %C(bold blue)<%an>%Creset" --stat'
 alias glol='git log --graph --pretty="%Cred%h%Creset -%C(auto)%d%Creset %s %Cgreen(%ar) %C(bold blue)<%an>%Creset"'
+alias glom='git log main --graph --pretty="%Cred%h%Creset -%C(auto)%d%Creset %s %Cgreen(%ar) %C(bold blue)<%an>%Creset"'
```
robbyrussell: @jonas-grgt Thanks for this, but I wonder if we can approach this differently so that we aren't hardcoding a specific branch name. It'd be the first one in the plugin to do so

robbyrussell (changes requested):

plugins/git/git.plugin.zsh
jonas-grgt: ooh that's a mistake on my end, I intended to use `$(git_main_branch)` instead. I'll update the PR.

jonas-grgt: Yes I do realize you probably want something like `glom my-branch`.

So here is what I suggest: 
-`glob` will log all comments from `$git_main_branch` 
- `glob my-branch` will log all comments from `my-branch`

This won't be an alias anymore but a function I suppose?

robbyrussell:
Thanks for the quick turnaround on `git_main_branch`... I saw the force-push, that's exactly it. The hardcoded-`main` concern is settled.

One thing to sort before it can land: `glom` is already claimed by #13743, open since May. Theirs is `git log --oneline --decorate $(git_main_branch)..` — that's `glo` applied to main, and since `glo` is `git log --oneline --decorate`, the name follows pretty directly. Yours uses `glol`'s format, so by the same logic it'd be `glolm`.

Could you rename it to `glolm`? That way both can live side by side and neither of you loses the work.

jonas-grgt:
@robbyrussell I renamed it to `glolm` thanks for the feedback!

robbyrussell (approved):

PR #14085, opened 2026-09-09
feat(cli): add interactive theme browser

robbyrussell:
Related issue: https://github.com/ohmyzsh/ohmyzsh/issues/14084

## Standards checklist:

## Changes:

- Add `omz theme preview <theme>` for isolated static prompt previews.
- Add `omz theme browse [filter]` for interactive filtering, navigation, and explicit use/save actions.
- Keep preview execution out of the active shell and restore terminal state after cancellation or failure.
- Add focused documentation and regression tests under `tools/theme-browser/`.

### Screen recording

https://github.com/user-attachments/assets/3a780d24-b713-4b26-8769-d5b57da29bed

Tested on macOS with zsh 5.9:

- `zsh -n lib/cli.zsh tools/theme-browser/*.zsh tools/theme-browser/tests/*_test.zsh`
- `zsh -df tools/theme-browser/tests/preview_test.zsh`
- `zsh -df tools/theme-browser/tests/browser_test.zsh`
- manually previewed the `robbyrussell` theme

Before marking this ready, I still need to resolve or explicitly gate older-zsh compatibility, decide whether to run the regression suites in CI, and test runtime behavior on Linux.

AI-assisted implementation and review. I reviewed and tested the resulting behavior locally; the known compatibility gap is disclosed above while this remains a draft.


robbyrussell:
I just added a short video to the PR description to show the theme browser working.

If you'd like to try it locally, start from the repository root and launch a disposable shell that doesn't read your normal zsh configuration:

```zsh
env ZSH="$PWD" zsh -dfi
```

Then load the CLI and try a few previews:

```zsh
source "$ZSH/lib/cli.zsh"

omz theme preview robbyrussell
omz theme preview agnoster
omz theme browse
omz theme browse agn
```

You can also run the regression suites from the repository root:

```zsh
zsh -df tools/theme-browser/tests/preview_test.zsh
zsh -df tools/theme-browser/tests/browser_test.zsh
```

The browser's **Use** action only changes the theme in the disposable session. Avoid **Save** unless you intend to update your real `~/.zshrc`.

robbyrussell (commented):
I'm holding this before merge. Both regression suites pass on macOS with zsh 5.9, but review found two confirmed cases they don't cover:

1. `tools/theme-browser/browser.zsh:49`: `print -Pn` turns literal backslash escapes such as `\e[2J` into terminal controls, undoing the backend's sanitization. The renderer needs `-r` and coverage for literal backslashes in filters and preview output.
2. `tools/theme-browser/browser.zsh:4-7,49`: `emulate -L zsh` preserves a disabled `PROMPT_PERCENT`. That exposes formatting markup and disables line clipping. The browser needs to enable the option explicitly.

Correction: I originally also reported that inherited EXIT-hook output corrupted Use/Save selections. That reproduction called the internal browser directly at top level, not through `omz theme browse`. Full CLI tests pass for Use, Save, and cancellation with both function and string EXIT traps on the unchanged code. I withdraw that finding; no trap-handling fix is needed.

Next step is to fix those cases and add regressions. Linux runtime testing and an explicit older-zsh compatibility decision are still outstanding. Current CI neither checks the new backend files nor runs these suites.

🤖 Co-authored with OpenCode

robbyrussell:
Addressed the two confirmed renderer issues in 76439de7 and added regression coverage. Both suites pass on macOS with zsh 5.9. I withdrew the EXIT-hook finding after full CLI tests showed it wasn't a bug; those cases are now covered too.

Linux runtime testing and older-zsh compatibility remain open.

🤖 Co-authored with OpenCode

PR #14089, opened 2026-09-10
refactor(github)!: deprecate legacy hub plugin

robbyrussell:
## Standards checklist:

- [x] This change does not introduce new aliases.

## Changes:

- Deprecate the `github` plugin, which supports the legacy `hub` CLI, and direct GitHub CLI users to the `gh` plugin.
- Keep existing `hub` behavior during the deprecation period while making the repository helpers use the current branch and stop after a failed directory change.
- Replace stale `github` documentation with migration guidance for `gh repo create`.
- Preserve the last valid `gh` completion cache when completion generation fails.
- Correct the documented `gh` completion cache location.

## Testing:

- Ran `zsh -n` against both plugin files.
- Ran `git diff --check`.
- Loaded both plugins through `oh-my-zsh.sh`.
- Generated and loaded completion using GitHub CLI v2.100.0.
- Simulated failed completion generation and verified the existing cache remained unchanged.
- Verified `exist_gh` does not call `hub` after a failed `cd` and pushes `HEAD` for a valid directory.

## Other comments:

GitHub CLI is not a drop-in Git wrapper, so this intentionally does not replace `alias git=hub` with `gh`. Existing `hub` users retain the old behavior until the deprecated plugin is removed.

AI-assisted: OpenCode assisted with research, implementation, and test setup; the resulting changes were reviewed and tested manually.

### DIFF
--- a/plugins/gh/README.md
+++ b/plugins/gh/README.md
@@ -15,9 +15,6 @@ This plugin does not add any aliases.
 This plugin caches the completion script and is automatically updated when the
 plugin is loaded, which is usually when you start up a new terminal emulator.
 
-The cache is stored at:
-
-- `$ZSH/plugins/gh/_gh` completions script
-
-- `$ZSH_CACHE_DIR/gh_version` version of GitHub CLI, used to invalidate
-  the cache.
+The completion script is stored at `$ZSH_CACHE_DIR/completions/_gh`. If an
+update fails, the plugin keeps the last successfully generated completion
+script.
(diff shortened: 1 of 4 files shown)


plugins/gh/gh.plugin.zsh
carlosala, line 1: These changes do not fit in this PR. Please remove.

plugins/gh/README.md
carlosala, line 1: Same as above.

carlosala (changes requested):

Issue #12307, opened 2024-03-27
Update prompt conflicts with VSCode python venv functionality


marcospgp:
### Describe the bug

VSCode python venv functionality runs `source venv/bin/activate` on terminal startup

Update prompt from oh my zsh captures the first character

![image](https://github.com/ohmyzsh/ohmyzsh/assets/3464445/b2f4b1c1-0105-44db-879f-5b5e8a5c946a)

### Steps to reproduce

See above

### Expected behavior

See above

### OS / Linux distribution

MacOS

### Zsh version

5.8.1

### Terminal emulator

VSCode

### If using WSL on Windows, which version of WSL

None

marcospgp:
Could a simple check for incoming input on shell startup suffice? I assume VSCode sends its command right away, faster than a human would, so a simple check for input before prompting user to update could solve the issue for VSCode and any other automated shell-creating programs that could interfere with ohmyzsh and have ohmyzsh interfere with them

carlosala:
The problem is that the input comes after the shell is initialized, that's why the first char gets cached by the update system.

carlosala:
You can choose that as well by setting the update mode `reminder`. You can read about it in the README!

marcospgp:
But given the wide adoption of VSCode as an editor is there a chance we could make the case for having that as a default?

mcornella:
I think that's an idea in the short run, let's add the check ourselves, but ideally in the long run I'd like to have a better solution for delayed initial input.

ka1amita:
I experienced the same issue with `cat > <newfile>` and `touch <newfile>` commands when propted for the update.

Ok imediatelly after update

mcornella:
I'm not able to reproduce this in VS Code (tried following [[1]] to no avail), and I don't have IntelliJ installed to reproduce this. Could you write a bullet point list of the steps to set this up, and confirm that you're in the latest OMZ version?

[1]: https://stackoverflow.com/questions/58433333/auto-activate-virtual-environment-in-visual-studio-code

marcospgp:
Why not just avoid capturing initial terminal output, which anecdotally conflicts with more than one application? I think a prompt that asks the user to run a command to update would work just as well in all instances, no?

Is there a big advantage in capturing input? Forcing the user to decide immediately whether they want to update (seems a bit intrusive)?

PR #14093, opened 2026-09-12
fix(updater): detect pending terminal input on macOS

pentaoa:
## Standards checklist:

- [ ] If the code introduces new aliases, I provide a valid use case for all plugin users down below. (Not applicable.)

## Changes:

`has_typed_input` currently returns false on macOS even when input is already queued: Apple's `/bin/stty` rejects `--save`, and the error is suppressed. The update prompt then consumes the first character of that input. For example, `source venv/bin/activate` becomes `ource venv/bin/activate`.

Use `stty -g` to save the terminal settings. This is supported by macOS/BSD and is also the [GNU short spelling of `--save`](https://www.gnu.org/software/coreutils/manual/html_node/stty-invocation.html). The existing detection and restoration logic can then run on macOS.

Related to #12307: this fixes the case where input has arrived **before** the update check. It does not address input arriving after the confirmation prompt is displayed.

## Other comments:

Tested the complete `tools/check_for_upgrade.sh` in a real macOS PTY with Zsh 5.9. Each case used a temporary HOME, cache, and Git repository without a remote, so no update was downloaded or executed:

| Input queued before sourcing | Before | After |
| --- | --- | --- |
| `source venv/bin/activate` plus newline | Prompt consumes `s` | Reminder; entire line preserved |
| Same input without newline | Prompt consumes `s` | Reminder; entire input preserved |
| No input | Confirmation prompt | Confirmation prompt |

Terminal settings were restored in all three cases (excluding macOS's transient `PENDIN` flag). `zsh -n tools/check_for_upgrade.sh` also passed.

To reproduce the complete-line case manually, run the following from the repository root in Zsh on macOS. During the three-second pause, type `source venv/bin/activate` and press Enter. The final `read` prints the remaining input without executing it:

```zsh
updater=$PWD/tools/check_for_upgrade.sh
test_dir=$(mktemp -d)
git init -q "$test_dir"
mkdir "$test_dir/log" "$test_dir/cache"
print 'LAST_EPOCH=0' > "$test_dir/cache/.zsh-update"
env HOME="$test_dir" ZSH="$test_dir" ZSH_CACHE_DIR="$test_dir/cache" \
  zsh -dfc 'zstyle ":omz:update" mode prompt
    sleep 3
    source "$1"
    IFS= read -r line
    print -r -- "Remaining input: $line"' zsh "$updater"
rm -rf -- "$test_dir"
```

AI-assisted: OpenAI Codex assisted with diagnosis, the change, and PTY validation.

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -185,7 +185,7 @@ function has_typed_input() {
   # Consider that no input can be typed if stty fails
   # (this might happen if stdin is not a terminal)
   local termios
-  termios=$(stty --save 2>/dev/null) || return 1
+  termios=$(stty -g 2>/dev/null) || return 1
   {
     # Disable canonical mode so that typed input counts
     # regardless of whether Enter was pressed


carlosala (approved):

carlosala:
Thanks for your contribution!

PR #14107, opened 2026-09-18
feat(eza): Modify eza aliases in eza.plugin.zsh

CedricBaaklini:
## Standards checklist:

## Changes:

Modified the following aliases of `eza` for better human readability and adding more important details:

`ll`, `la`, `lD`, `lDD`, `lsdl`

## Other comments:

This will show the headers, the octal version of the permission bits, the group that owns a file, and the links. Similar to how the OG `ls` show the contents of a folder/directory


CedricBaaklini:
Update: I just found out that many of these can be toggle by adding the settings in zshrc.

PR #14109, opened 2026-09-18
fix(tmux): sanitize `tds` session names for tmux target syntax

peat:
## Standards checklist:

## Changes:

`tds` names its session after the current directory, but `tmux new -A -s NAME` looks NAME up as a **target** when the session already exists. tmux's target syntax treats `.` and `:` as separators and a leading `$`, `@` or `%` as an id. So in a directory like `example.com`, `v1.2.3` or `.hidden` the first `tds` creates the session and every later one fails:

```
$ cd example.com && tds      # works, creates example.com-e71c75
$ tds                        # can't find pane: com-e71c75
```

This replaces those characters with `_` before building the name (`example.com` → `example_com-e71c75`). The hash suffix is untouched, so uniqueness per directory is preserved.

Tested on tmux 3.7c with `.`, `:`, leading `$`/`@`/`%`, spaces, unicode and plain names: all now create and re-find correctly.

## Other comments:

Only directories containing the affected characters get a different session name than before, and those could never be reattached anyway. The `ZSH_TMUX_AUTONAME_SESSION` path in `_zsh_tmux_plugin_run` has the same class of issue (`attach -t "$session_name"`); I left it alone to keep this focused, happy to extend if wanted.

AI disclosure: Claude Code helped investigate the tmux target-parsing behaviour and draft this description.


carlosala:
Good catch! @peat could you fix?

peat:
Good catch, thanks. Confirmed on tmux 3.7c: a leading `=` is stripped as the exact-match prefix, so `=work` created a session that `new -A -s` then couldn't find. Added `=` to the leading-character class in f2618abd.

While checking, I brute-forced every printable ASCII character at the leading, middle and trailing positions. The only characters that break reattach are the ones now handled (`.` `:` anywhere; `$` `@` `%` `=` leading). `#` gets format-expanded by tmux so it yields an uglier name, but it expands the same way on every call and still reattaches, so I left it alone.

carlosala (approved):

PR #14110, opened 2026-09-18
feat(chezmoi): added aliases for common Chezmoi commands

CedricBaaklini:
## Standards checklist:

## Changes:

Added aliases for common chezmoi commands. Including Chezmoi itself. There is a table in README.md that I added that briefly explains these commands.

`alias cm="chezmoi"`
`alias cma="chezmoi apply"`
`alias cmc="chezmoi cd"`
`alias cmd="chezmoi diff"`
`alias cme="chezmoi edit"`
`alias cmg="chezmoi git"`
`alias cmi="chezmoi init"`
`alias cms="chezmoi status"`
`alias cmu="chezmoi update"`

## Other comments:

FULL DISCLOSURE: I used Jetbrains' Junie to write the table on the README.


plugins/chezmoi/README.md
```diff
@@ -9,3 +9,17 @@ To use it, add `chezmoi` to the plugins array of your zshrc file:
 ```bash
 plugins=(... chezmoi)
 ```
+
+## Aliases
+
+| Alias | Command          | Description                                                    |
+|:------|:-----------------|:---------------------------------------------------------------|
+| `cz`  | `chezmoi`        | Chezmoi main command                                           |
```
axmrnv: Imo, "cz" as an alias for "chez moi" is counterintuitive; "cm" would fit better.

CedricBaaklini: In hindsight, I agree. I have changed it to "cm".

robbyrussell (changes requested):
Thanks for this. Two things before I can take it:

1. `plugins/chezmoi/chezmoi.plugin.zsh:23`: `chezmoi re-apply` isn't a chezmoi command (I checked the command reference and the source tree; there's `re-add`, but no `re-apply`). As written, `cmr` just prints an error. Drop it, or point it at `re-add` if that's what you meant.
2. This overlaps with #13598, which has been open since February and proposes a bigger set with different names (`cma` is `add` there and `apply` here). I'd like to merge one chezmoi alias set, not two. Can you take a look at that one and say why you'd pick your set over it? The contributing guide asks for the workflow behind new aliases, so a couple of sentences on how you use these day to day would settle it for me.

One more I'm less sure of: `cmd` is the Windows command prompt for people running Oh My Zsh under MSYS2 or Git Bash. I haven't tested that, but a different name would sidestep it.

CedricBaaklini:
1. I must've mixed up my commands because for some reason, I kept thinking of the word "apply" instead of "add". Oops.
2. Honestly, after looking at that PR, I'd say omer-inam did it much better than I ever could.

PR #14112, opened 2026-09-19
fix(frontend-search): use fallback search for dead dartlang docs URL

manysuq:
fix(frontend-search): use fallback search for dead dartlang docs URL

The hardcoded https://api.dartlang.org/apidocs/... viewer URL returns 404 - that host no longer serves API docs. Route dartlang searches through the existing _frontend_fallback helper (Google/DuckDuckGo site search over dart.dev), the same pattern already used by emberjs, reactjs, vuejs and others in this plugin.

Verified: zsh -n passes on the plugin file; _frontend_fallback dart.dev resolves to a working site search URL with default settings.

AI-assisted: used AI tooling to scan plugin URLs for dead links; the fix and verification were done and reviewed manually.


robbyrussell (approved):
lgtm

